PR #29, opened 2017-01-23
Add types to express all known attribute value types.

geeknoid:


mixer/v1/config/cfg.proto
```diff
@@ -129,3 +129,23 @@ message ClientConfig {
...
   repeated AspectRule rules = 3;
 }
+
+// Uri represents a properly formed URI.
+message Uri {
+  string uri = 1;
+}
+
+// IPAddr holds an IPv4 or IPv6 address.
+message IPAddr {
```
wora: IpAddress

mixer/v1/config/cfg.proto
```diff
@@ -129,3 +129,23 @@ message ClientConfig {
...
+  bytes ip_addr = 1;
+}
+
+// DNSNames holds a valid domain name.
+message DNSName {
```
wora: DnsName

mixer/v1/config/cfg.proto
```diff
@@ -129,3 +129,23 @@ message ClientConfig {
...
+  string dns_name = 1;
+}
+
+// EmailAddr holds a properly formatted email address.
+message EmailAddr {
```
wora: EmailAddress

mixer/v1/config/cfg.proto
```diff
@@ -129,3 +129,23 @@ message ClientConfig {
   string revision = 2;
   repeated AspectRule rules = 3;
 }
+
+// Uri represents a properly formed URI.
+message Uri {
```
wora: If we use Camel case for Uri, we should do the same for all names.

mixer/v1/config/cfg.proto
```diff
@@ -129,3 +129,23 @@ message ClientConfig {
...
+}
+
+// DNSNames holds a valid domain name.
+message DNSName {
+  string dns_name = 1;
+}
+
+// EmailAddr holds a properly formatted email address.
+message EmailAddr {
+  string email_addr = 1;
```
wora: For all well known types under google/protobuf, we use generic "value" inside the message. I suggest we do the same here.

geeknoid:
Updated per Hong's comments.

PR #35, opened 2017-01-29
Streamline QuotaDescriptor.

geeknoid:
This generalizes refresh windows and eliminates the API level distinction between allocation and rate quotas.

Allocated quota now just has an expiration period which determines when how long an allocation 'survives' until it is automatically freed. If you specify a expiration time of 0, then the quota is never freed.

This removes the notion of precise/imprecise quotas. That was just a leftover of how things are handled in GCP and isn't really an appropriate top-level concern.

### DIFF
--- a/mixer/v1/config/descriptor/quota_descriptor.proto
+++ b/mixer/v1/config/descriptor/quota_descriptor.proto
@@ -52,23 +52,8 @@ message QuotaDescriptor {
   // The default imposed maximum amount for values of this quota.
   int64 max_amount = 6;
 
-  // The type of quota behavior expected, defaults to ALLOCATION_LIMIT if not specified
-  QuotaKind kind = 7;
-
-  // Whether the quota's current value is tracked precisely or not.
-  // Precisely tracked quotas only allow a relatively modest level of
-  // scaling, whereas imprecise quotas can support a nearly unbounded
-  // level of scaling at the cost of potential accounting inaccuracies.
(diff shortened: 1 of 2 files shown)


mixer/v1/config/descriptor/quota_descriptor.proto
```diff
@@ -52,23 +52,8 @@ message QuotaDescriptor {
...
-    ALLOCATION_LIMIT = 0;
-
-    // Quota limit expresses a maximum amount over a rolling time interval
-    PER_SECOND_LIMIT = 1;
-
-    // Quota limit expresses a maximum amount over a rolling time interval
-    PER_MINUTE_LIMIT = 2;
-  }
+  // The amount of time allocated quota remains valid before it is
+  // automatically released. If this is 0, then allocated quota is
```
louiscryan: Are we happy using 0 as opposed to a wrapped value?

geeknoid:
By wrapped value, do you mean -1? If so, what do you think would be the
benefit?

louiscryan:
Nope I mean

https://github.com/google/protobuf/blob/master/src/google/protobuf/wrappers.proto#L83

i.e. nullable

geeknoid:
I don't see the benefit. An expiration window of 0 is semantically invalid
so it makes a fine sentinel value. If there's a style guide recommendation
to use a wrapper, I can certainly change it. But on the face of it, I think
0 is fine.

PR #36, opened 2017-02-06
Rework the metric_descriptor type

ZackButcher:
Rework the metric_descriptor to separate the description of the shape of the metric from the attributes used to produce values for the metric at run time.


ZackButcher:
https://github.com/istio/api/pull/37 will need to be merged in before this one.

mixer/v1/config/descriptor/label_descriptor.proto
```diff
@@ -0,0 +1,31 @@
+// Copyright 2016 Google Inc.
```
geeknoid: 2017

PR #44, opened 2017-02-11
Add MixerAttributeRule for proxy config

qiwzhang:


qiwzhang:
@mandarjog and @geeknoid, this is my proposal of proxy config for mixer attributes.  Could you review it?
Thanks

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,64 @@ message ProxyConfig {
...
+message QuotaAttributes {
+  // List attributes for quota call.
+  // It must have at least one attribute.
+  repeated string request_attributes = 1;
+
+  // List attributes proxy should use for calculating cache key.
+  // If absent, all attributes should be used for cache key.
+  repeated string cache_key_attributes = 2;
+
+  // Amount of quota should be allocated for this request.
```
wora: Please use formal English for comment.

// The amount of quota units should be allocated for the request.

- Use articles "a" and "the".
- This message is not a request, so not "this request".

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,64 @@ message ProxyConfig {
...
+
+// Describes attributes for quota call.
+message QuotaAttributes {
+  // List attributes for quota call.
+  // It must have at least one attribute.
+  repeated string request_attributes = 1;
+
+  // List attributes proxy should use for calculating cache key.
+  // If absent, all attributes should be used for cache key.
+  repeated string cache_key_attributes = 2;
```
wora: As a reader, I don't understand how to use this. If I want to allocate 5 VMs for my project, what does "cache key" means?

"List attributes", I don't know if list is noun or verb here. If verbs, it should be lists xxx. If noun, it should be "A list of attributes".

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,64 @@ message ProxyConfig {
...
+  // List attributes for Report calls.
+  // If absent, send all attributes that proxy can extract.
+  repeated string request_attributes = 1;
+}
+
+// Describes attributes for quota call.
+message QuotaAttributes {
+  // List attributes for quota call.
+  // It must have at least one attribute.
+  repeated string request_attributes = 1;
```
wora: The concept of request is unreferenced here.

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,64 @@ message ProxyConfig {
...
+
+// Describes attributes for check call.
+message CheckAttributes {
+  // List attributes for Check calls.
+  // If absent, send all attributes that proxy can extract.
+  repeated string request_attributes = 1;
+
+  // List attributes proxy should use for calculating cache key.
+  // If absent, all attributes should be used for cache key.
+  repeated string cache_key_attributes = 2;
```
wora: Why don't let mixer return this value? I think this is more dynamic .

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,64 @@ message ProxyConfig {
   // policies, failure recovery policies such as timeouts, retries, circuit
   // breakers), etc.
   repeated UpstreamCluster upstream_clusters = 3;
+
+  // specify attributes proxy extracted from request for mixer.
```
wora: Specifies the set of attributes that the proxy should provide when calling the mixer.

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,64 @@ message ProxyConfig {
...
+  repeated MixerAttributeRule mixer_attribute_rules = 4;
+}
+
+// Describes rules for proxy to extract attributes for a request.
+message MixerAttributeRule {
```
wora: The complexity of this config seems huge. Different mixer calls will have different attributes, it will be very hard to reason. Just a concern.

proxy/v1/config/cfg.proto
qiwzhang: Done. Thanks.

proxy/v1/config/cfg.proto
qiwzhang: Done. Thanks.

proxy/v1/config/cfg.proto
qiwzhang: Acked.

proxy/v1/config/cfg.proto
qiwzhang: Thanks. Updated the comment.

proxy/v1/config/cfg.proto
qiwzhang: Done. update the comment

kyessenov:
Wayne, do you mind splitting your addition to a separate file? We're no longer using the original proto. I think it makes sense to associate mixer config with upstream cluster and not the route rule, similarly how it's done for load balancing. However, mixer filter is shared for all clusters in the destination, so you need add something like "upstream_cluster" in the fault injection filter:
https://lyft.github.io/envoy/docs/configuration/http_filters/fault_filter.html

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,96 @@ message ProxyConfig {
...
   repeated UpstreamCluster upstream_clusters = 3;
+
+  // Specifies the set of rules on how the proxy should communicate
+  // with the mixer. For example, they specify what attributes the
+  // proxy should provide to the mixer and how the proxy should cache
+  // the requests.
+  repeated MixerAttributeRule mixer_attribute_rules = 4;
+}
+
+// Describes a rule on how the proxy should communicate with the mixer.
```
wora: The comment is not very readable to me.

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,96 @@ message ProxyConfig {
...
+message MixerAttributeRule {
+  // If a request matches the condition, the rule should apply.
+  oneof match {
+    L4MatchCondition l4 = 1;
+    HttpMatchCondition match = 2;
```
wora: Why match inside match?

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,96 @@ message ProxyConfig {
...
+message CheckRequestCache {
+  // A list of attributes the proxy can use to calculate the cache
+  // key. If absent, all attributes in the check request should be
+  // used for cache key.
+  repeated string key_attributes = 1;
+
+  // Specifies how long a cached request item should live in second.
+  // If the mixer response has cache time_to_live, it will override
+  // the one from the config.
+  int64 time_to_live = 2;
```
wora: ttl_seconds.

I seriously doubt you need int64 for this.

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,96 @@ message ProxyConfig {
...
+  // Specifies how the proxy should cache the check requests.
+  CheckRequestCache cache = 2;
+}
+
+// Specifies how the proxy should cache the check requests.
+message CheckRequestCache {
+  // A list of attributes the proxy can use to calculate the cache
+  // key. If absent, all attributes in the check request should be
+  // used for cache key.
+  repeated string key_attributes = 1;
```
wora: Explain how this is useful? Why don't just say used_attributes.

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,96 @@ message ProxyConfig {
...
+}
+
+// Describes the attributes for a quoa metric.
+message QuotaMetric {
+  // A list of attributes for a quota metric that the proxy should
+  // send to the mixer in the quota request call.
+  repeated string request_attributes = 1;
+
+  // The amount of quota units should be allocated for the request.
+  int64 amount = 2;
```
wora: Amount of what? If we have both request quota and bandwidth quota, what this amount means?

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,96 @@ message ProxyConfig {
...
+}
+
+// Describes the attributes the proxy should send for the report call.
+message ReportAttributes {
+  // A list of attributes the proxy should send for the report call.
+  // If absent, send all attributes the proxy can extract.
+  repeated string request_attributes = 1;
+}
+
+// Describes the attributes for a quoa metric.
```
wora: A typical metric has a name, a value, and some dimensions. This one doesn't look like a metric.

proxy/v1/config/cfg.proto
geeknoid: Can this be a Duration instead?

proxy/v1/config/cfg.proto
geeknoid: The proxy needs to know the set of attributes the mixer wants to hear about in general, and the set of attributes which affect the yes/no result of the precondition check. This will often be a subset of the first.

I think we'll eventually want to make both sets be returned dynamically from mixer so as to avoid needing the operator to specify this by hand. But for starters, we can be explicit for the time being.

proxy/v1/config/cfg.proto
```diff
@@ -87,6 +87,96 @@ message ProxyConfig {
...
+message QuotaMetric {
```
geeknoid: Let's not use "Quota" and "Metric" together in terms of names. In One Platform, we built quotas as limits on metrics, but this is not the case in Istio.

So maybe QuotaAlloc would be better.

proxy/v1/config/cfg.proto
geeknoid: It doesn't matter what this is counting. This is trying to say "whenever there is a request, please allocate N quota units from this quota".

geeknoid: Clearly, this will ultimately need to be more complex so as to support different quota for different API methods, for example.

proxy/v1/config/cfg.proto
qiwzhang: For now as starter,  we will have different config format for proxy(mixer client) and mixer server. This is defining one for Proxy (or mixer client).  At client, we only need to know metric name.

proxy/v1/config/cfg.proto
mandarjog: @geeknoid @qiwzhang  can we punt on all performance related configuration here including cache keys?

Mixer API will be augmented in the future to negotiate attribute list. Is there still value in having this here?

If we must have it here, @qiwzhang can you create a separate `mixer_client_cfg.proto`? file that deals with mixer specific config?

geeknoid:
We certainly can punt perf optimizations for the moment. But this is fairly
low cost and lets us get a better sense of perf until the real solution is
in place.

PR #45, opened 2017-02-14
Add definitions proto for vocabulary and other definitions

mandarjog:
Definitions proto provides an envelope to define the following using existing descriptor protos.

- attributes    ( vocabulary)
- metrics
- quotas
- labels
- log entries
- monitored resources
- principals.


wora:
Let's move all descriptor protos under //mixer/descriptor. We can introduce //mixer/descriptor/v2 if it is really needed, but I sincerely hope not.

mandarjog:
Done. PTAL.

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,42 @@
...
+
+package istio.mixer.descriptor;
+
+// Istio Definitions
+// An istio definition proto message may contain some or all definition types.
+// Messages are combined by runtime to form the effective definition.
+// In many cases attribute definitions (vocabulary) are available as separate messages. 
+// Every descriptor proto defined in this package should be represented here.
+message Definitions {
+  string subject = 1;
```
wora: Please add comments to each field. Can't tell what they do.

mandarjog:
@geeknoid @smawson I have moved 	v1/config/descriptor > descriptor, which effectively makes `attributes` and other `descriptors` unversioned.

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+// Mixer uses a lookup api to access these definitions.
+// Find(definition_name, definition_type)
+message Definitions {
+  // subject 
+  string subject = 1;
+  string revision = 2;
+  // attributes form the attribute vocabulary
+  // The effective vocabulary is a union of vocabulary presented by the proxy
+  // and additional attributes generated by the mixer.
+  repeated istio.mixer.v1.config.descriptor.AttributeDescriptor attributes = 3;
```
smawson: Need to update the packages here.

wora: I think we should move attribute and vocabulary to istio.attribute. Leave the rest under mixer/somewhere.

geeknoid: The Definitions message is intended to hold the definitions of the various 'object types' used by the operator.

The set of attributes supported by a deployment is not a property of the configuration state, it is a manifest associated with pieces of code (the proxy in particular). As such, I don't think this field should exist here.

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+import "mixer/descriptor/label_descriptor.proto";
+import "mixer/descriptor/metric_descriptor.proto";
+import "mixer/descriptor/quota_descriptor.proto";
+import "mixer/descriptor/log_entry_descriptor.proto";
+import "mixer/descriptor/monitored_resource_descriptor.proto";
+import "mixer/descriptor/principal_descriptor.proto";
+
+package istio.mixer.descriptor;
+
+// Istio Definitions
```
geeknoid: Please rephrase this comment to be in flowing prose with the intended audience to be the consumer of the message. That is, this shouldn't have things like "Every descriptor proto defined in this package should be represented here" as that is a comment for implementers of the package, not for its consumers.

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+// An istio definition proto message may contain some or all definition types.
```
geeknoid: "a definition may contain some or all definitions" isn't clear. Presumably, you mean to say something like

The effective set of descriptors understood by Istio is produced through the union of all Definition messages within the configuration state.

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+// Messages are combined by runtime to form the effective definition.
+// In many cases attribute definitions (vocabulary) are available as separate messages. 
+// Every descriptor proto defined in this package should be represented here.
+// Every proto message in definition has a unique name within the message class.
```
geeknoid: What's 'message class'?

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+// Mixer uses a lookup api to access these definitions.
+// Find(definition_name, definition_type)
```
geeknoid: Presumably, this is actually descriptor_name and descriptor_type, right?

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+message Definitions {
```
geeknoid: Considering this holds descriptors, why not call this Descriptors?

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+  // and additional attributes generated by the mixer.
+  repeated istio.mixer.v1.config.descriptor.AttributeDescriptor attributes = 3;
+  // labels are analogous to attributes. They are externalized representation of attributes used by aspects.
+  repeated istio.mixer.v1.config.descriptor.LabelDescriptor labels = 4;
+  // metrics defines available metrics types.
+  repeated istio.mixer.v1.config.descriptor.MetricDescriptor metrics = 5;
+  // quota defines available quota types.
+  repeated istio.mixer.v1.config.descriptor.QuotaDescriptor quotas = 6;
+  // logEntries define available log types.
+  repeated istio.mixer.v1.config.descriptor.LogEntryDescriptor logEntries = 7;
```
geeknoid: For proto fields, please use _ to separate words instead of camelCase.

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+// An istio definition proto message may contain some or all definition types.
+// Messages are combined by runtime to form the effective definition.
+// In many cases attribute definitions (vocabulary) are available as separate messages. 
+// Every descriptor proto defined in this package should be represented here.
+// Every proto message in definition has a unique name within the message class.
+// Mixer uses a lookup api to access these definitions.
+// Find(definition_name, definition_type)
+message Definitions {
+  // subject 
+  string subject = 1;
```
geeknoid: What's this for? The comment is not helpful.

mixer/descriptor/definitions.proto
```diff
@@ -0,0 +1,55 @@
...
+  string revision = 2;
```
geeknoid: What's this for, how is it used, what does it contain?

geeknoid:
Can this PR be closed?

mandarjog:
This is being addressed piece meal in other issue now.

PR #52, opened 2017-03-02
Use /** and /// comments for doc generation

rshriram:


proxy/v1/config/cfg.proto
```diff
@@ -379,46 +326,50 @@ message HttpFaultInjection {
...
+    // TODO: The semantics and syntax of the headers is undefined.
     string override_header_name = 5;
   }
 }
 
-// Faults can be injected into the L4 traffic forwarded by the proxy, for
-// testing the failure recovery capabilities of downstream services.
-// Faults include terminating established Tcp connections, throttling the
-// upstream/downstream bandwidth (for Tcp|Udp), or both.
+// Faults can be injected into the connections from downstream by the
```
ZackButcher: Should this have 3 slashes?

proxy/v1/config/cfg.proto
```diff
@@ -379,46 +326,50 @@ message HttpFaultInjection {
...
+    /// Error code to use to abort the Http request. Requests can be aborted
+    /// either with Http/1.1 status codes | http2 error codes or gRPC status
+    /// codes. Only http_status based errors are supported at the moment.
     oneof error_type {
       string grpc_status = 2;
       string http2_error = 3;
-      uint32 http_status = 4;
+      int32 http_status = 4;
     }
     // Specify abort code as part of Http request.
```
ZackButcher: ///

proxy/v1/config/cfg.proto
rshriram: Intentionally done. We dont want to generate docs for these publicly yet because they are not implemented. As far as proto is concerned, there are still comments.

proxy/v1/config/cfg.proto
rshriram: Actually, sorry. This was also left out intentionally, as we don't support these yet. We dont want users to get confused about stuff that is unimplemented.

PR #57, opened 2017-03-05
Move fault injection into route rule

rshriram:
Just moving fault injection into route rule as discussed during the calls. Retained all docs, etc.


proxy/v1/config/cfg.proto
```diff
@@ -158,7 +158,7 @@ message MatchCondition {
   /// Set of HTTP match conditions based on HTTP/1.1, HTTP/2, GRPC request metadata,
   /// such as "uri", "scheme", "authority".
   /// The header keys are case-insensitive.
-  map<string, StringMatch> http = 5;
+  map<string, StringMatch> httpHeaders = 5;
```
wora: Use snake_case please.

PR #58, opened 2017-03-06
Provide TTL values for Check and Quota responses.

geeknoid:
Quota had a TTL value expressed in terms of absolute time.
This unfortunately is useless in a distributed system.

Now both QuotaResponse and CheckResponse carry a Duration value
represent how long the response is valid for.

### DIFF
--- a/mixer/v1/check.proto
+++ b/mixer/v1/check.proto
@@ -16,6 +16,7 @@ syntax = "proto3";
 
 package istio.mixer.v1;
 
+import "google/protobuf/duration.proto";
 import "google/rpc/status.proto";
 import "mixer/v1/attributes.proto";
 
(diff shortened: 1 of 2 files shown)


mixer/v1/check.proto
```diff
@@ -34,4 +35,7 @@ message CheckResponse {
 
   // Indicates whether or not the preconditions succeeded
   google.rpc.Status result = 2;
+
+  // The amount of time for which this result can be considered valid, given the same inputs
+  google.protobuf.Duration expiration = 3;
```
douglas-reid: nit: style guide recommendation is to append `_duration`... so maybe `validity_duration` ?

geeknoid: @wora Hong, what's the rationale for sticking a _duration suffix on such a field? Seems completely redundant. We don't do name_string and port_int, so why have _duration here?

PR #61, opened 2017-03-08
Use gogoprotobuf code gen options to improve the Go versions of our A…

geeknoid:
…PI protos.

This makes for smaller & nicer Go code that runs faster due to decreased
GC pressure.

### DIFF
--- a/mixer/v1/check.proto
+++ b/mixer/v1/check.proto
@@ -16,9 +16,14 @@ syntax = "proto3";
 
 package istio.mixer.v1;
 
+import "gogoproto/gogo.proto";
 import "google/protobuf/duration.proto";
 import "google/rpc/status.proto";
 import "mixer/v1/attributes.proto";
+
+option (gogoproto.goproto_getters_all) = false;
+option (gogoproto.equal_all) = false;
+option (gogoproto.gostring_all) = false;
 
 // Used to verify preconditions before performing an action.
 message CheckRequest {
@@ -26,7 +31,7 @@ message CheckRequest {
   int64 request_index = 1;
 
   // The attributes to use for this request
-  Attributes attribute_update = 2;
+  Attributes attribute_update = 2 [(gogoproto.nullable) = false];
 }
 
 message CheckResponse {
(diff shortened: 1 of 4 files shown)


mixer/v1/config/cfg.proto
```diff
@@ -126,12 +120,6 @@ message GlobalConfig {
   string revision = 1;
   repeated Adapter adapters = 2;
 
-  // TODO: remove these in https://github.com/istio/api/pull/45
-  repeated istio.mixer.v1.config.descriptor.LogEntryDescriptor logs = 3;
```
mandarjog: Did you want to move this to some other place, or just remove it ?

The descriptors here are the first solution to add descriptors to global config.

geeknoid:
Oops, I temporarily removed these because other stuff wasn't building. That
shouldn't be in the PR, I'll restore that.

mixer/v1/check.proto
```diff
@@ -16,26 +16,41 @@ syntax = "proto3";
...
   // Index of the request this response is associated with
   int64 request_index = 1;
 
   // Indicates whether or not the preconditions succeeded
-  google.rpc.Status result = 2;
+  google.rpc.Status result = 2 [(gogoproto.nullable) = false];
 
   // The amount of time for which this result can be considered valid, given the same inputs
-  google.protobuf.Duration expiration = 3;
+  google.protobuf.Duration expiration = 3 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
```
mandarjog: does `protobuf.Duration` offer anything beyond nullability over `int64` ?

geeknoid: It's primarily a semantic thing. It clearly articulates it is a duration in a known format.

geeknoid:
The options used here are:

These three prevent generating utility functions we don't use. They prevent GetXXX accessors, Equal, and GoString from being generated:

```
option (gogoproto.goproto_getters) = false;
option (gogoproto.equal) = false;
option (gogoproto.gostring) = false;
```

These generate Marshal, Unmarshal, and Size methods which improve marshalling perf by replacing reflection-oriented logic with hand-written code.

```
option (gogoproto.marshaler) = true;
option (gogoproto.unmarshaler) = true;
option (gogoproto.sizer) = true;
```

And when this is applied to a field, it makes it so what would normally be a pointer to a struct is replaced with an inline struct instead.

```
(gogoproto.nullable) = false];
```

geeknoid:
Attached is an example of the code generated with these options.

[check.pb.go.txt](https://github.com/istio/api/files/828894/check.pb.go.txt)

geeknoid:
And here's that same file compiled without these new options.
[check.pb.go.txt](https://github.com/istio/api/files/828906/check.pb.go.txt)

geeknoid:
PTAL.

geeknoid:
Can I go ahead with this? Other work is piled on waiting for this change...

PR #64, opened 2017-03-12
Another gogo improvement: inline stringmap structs to avoid another u…

geeknoid:
…seless pointer.

### DIFF
--- a/mixer/v1/check.proto
+++ b/mixer/v1/check.proto
@@ -38,9 +38,12 @@ message CheckResponse {
   // Index of the request this response is associated with
   int64 request_index = 1;
 
+  // The attributes to use for this response
+  Attributes attribute_update = 2;
+
   // Indicates whether or not the preconditions succeeded
-  google.rpc.Status result = 2 [(gogoproto.nullable) = false];
+  google.rpc.Status result = 3 [(gogoproto.nullable) = false];
 
   // The amount of time for which this result can be considered valid, given the same inputs
-  google.protobuf.Duration expiration = 3 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
+  google.protobuf.Duration expiration = 4 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
 }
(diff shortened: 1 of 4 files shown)


mixer/v1/report.proto
```diff
@@ -39,4 +39,7 @@ message ReportResponse {
 
   // Indicates whether the report was processed or not
   google.rpc.Status result = 2 [(gogoproto.nullable) = false];
+
+  // The attributes to use for this response
+  Attributes attribute_update = 3;
```
ZackButcher: Can you make this one consistent with the others? Either make the attributes field #2 in this message, or #3 in the others. IMO check/quota should be updated to match this one.

geeknoid: I was my intent to make those all the same, but I guess my mind strayed somewhere along the way.

PTAL.

mixer/v1/check.proto
```diff
@@ -38,9 +38,12 @@ message CheckResponse {
   // Index of the request this response is associated with
   int64 request_index = 1;
 
+  // The attributes to use for this response
+  Attributes attribute_update = 2;
+
   // Indicates whether or not the preconditions succeeded
-  google.rpc.Status result = 2 [(gogoproto.nullable) = false];
+  google.rpc.Status result = 3 [(gogoproto.nullable) = false];
```
qiwzhang: "2" used to be "result", now it is used for attribute_update. This is bad.  It breaks the backward  compatibility.  You never re-use a proto index.

geeknoid: Sorry, given the phase the project is in I didn't think backward compat was an issue yet.

ZackButcher: We did this (and in general have been changing field indices) because these protos aren't being persisted anywhere today - the field number change should have no effect on anyone. We're still pre-Alpha so all bets are off w.r.t. backwards compatibility. Post official Alpha release I totally agree this kind of shuffling of fields must never happen.

qiwzhang: It is between proxy and mixer.  Without this change,  we can safely say all proxy versions works with all mixer versions.  Now with this change, we could not say that any more.  
The demo users have to be careful which version of proxy and mixer can work together.  They are not always using the latest.

PR #65, opened 2017-03-21
Add attribute descriptors to the set of descriptors in the global config

ZackButcher:


ZackButcher:
I _think_ we don't care about field numbers here since this config is strictly consumed by the mixer (no danger of mismatched version, and we don't persist the config as proto anywhere).

PR #67, opened 2017-03-31
Define proxy mesh global variables

kyessenov:
We would like to eliminate the use of flags from the manager images, since they are prone to errors and lead to inconsistency. The plan is to import the config values to components from a `ConfigMap`.

### DIFF
--- a/proxy/v1/config/cfg.proto
+++ b/proxy/v1/config/cfg.proto
@@ -48,13 +48,63 @@
 syntax = "proto3";
 
 import "google/protobuf/any.proto";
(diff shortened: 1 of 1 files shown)


kyessenov:
Let's not worry about Envoy-specificity and backwards compatibility for this proto. A lot of things fall apart once you switch to nginx, including the architectural choice to split manager and proxy agent.

kyessenov:
There is also a technical reason with ugly one-of translation in Go, and the fact that configmap should preferably be single-level and not nested.

PR #70, opened 2017-04-07
added egress proxy service to proxyconfig

GregHanson:
Field needed for https://github.com/istio/manager/pull/463

https://github.com/istio/manager/issues/67

### DIFF
--- a/proxy/v1/config/cfg.proto
+++ b/proxy/v1/config/cfg.proto
@@ -59,7 +59,7 @@ option go_package = "config";
 // service mesh.
 message ProxyMeshConfig {
   // Address of the egress proxy service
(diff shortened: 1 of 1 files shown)


proxy/v1/config/cfg.proto
```diff
@@ -58,25 +58,29 @@ option go_package = "config";
 // ProxyMeshConfig defines variables shared by all proxies in the Istio
 // service mesh.
 message ProxyMeshConfig {
+  // Address of the egress proxy service
```
kyessenov: (e.g. "istio-egress:80")

GregHanson: updated comment

PR #71, opened 2017-04-08
Introduce unary mixer APIs.

geeknoid:


geeknoid:
Never mind, this needs extra bake time...

PR #72, opened 2017-04-09
Introduce unary mixer APIs.

geeknoid:

This API updates the attribute protocol to use both a global connection-level dictionary and a
per-message dictionary, add supports for using the dictionary for attribute values in addition to
using it for attribute names. Finally, the new protocol doesn't provide delta
semantics, it's always all-or-nothing.

### DIFF
--- a/mixer/v1/config/descriptor/quota_descriptor.proto
+++ b/mixer/v1/config/descriptor/quota_descriptor.proto
@@ -16,7 +16,6 @@ syntax = "proto3";
 
 package istio.mixer.v1.config.descriptor;
 
-import "google/protobuf/duration.proto";
 import "mixer/v1/config/descriptor/value_type.proto";
 
 // Configuration state for a particular quota.
@@ -43,11 +42,6 @@ message QuotaDescriptor {
   // for a quota of this type.
   map<string, ValueType> labels = 4;
 
-  // The default imposed maximum amount for values of this quota.
-  int64 max_amount = 5;
-
-  // The amount of time allocated quota remains valid before it is
-  // automatically released. If this is 0, then allocated quota is
-  // not automatically released.
-  google.protobuf.Duration expiration = 6;
+  // Indicates whether the quota represents a rate limit or represents a resource quota.
+  bool rate_limit = 5;
 }


mixer/v1/attributes.proto
```diff
@@ -24,102 +24,54 @@ option (gogoproto.goproto_getters_all) = false;
...
-  // the same effect using multiple gRPC streams. The benefit of using streams
-  // would be that the mixer would be in control of the maximum number of streams
-  // it allows, whereas with the current model the proxy could overwhelm the
-  // mixer by creating too many contexts.
-  int32 attribute_context = 2;
+  // Attributes payload. All `sint32` values represent indices into
+  // one of the word dictionaries. Positive values are indices into the
+  // global connection-level dictionary, Negative values are indices into
+  // the message-level dictionary.
+  map<sint32, sint32> string_attributes2 = 14;
```
douglas-reid: are we using signed ints here for more efficient encoding?

mixer/v1/attributes.proto
```diff
@@ -24,102 +24,54 @@ option (gogoproto.goproto_getters_all) = false;
...
-  // TODO: temporary workaround for gogoprotobuf code gen error (https://github.com/gogo/protobuf/issues/273). Remove
-  // when issue is fixed.
-  map<int32, google.protobuf.Timestamp> timestamp_attributes_HACK = 13 [(gogoproto.nullable) = false, (gogoproto.stdtime) = false];
-  map<int32, google.protobuf.Duration> duration_attributes_HACK = 14 [(gogoproto.nullable) = false, (gogoproto.stdduration) = false];
 }
 
 // A map of string to string. The keys in these maps are from the current
 // dictionary.
+// TO BE REMOVED
 message StringMap {
```
douglas-reid: instead of defining a new StringMap message type, couldn't we just add a new field to the message called something like `signed_int_map`?  That would save some of this `2` ugliness.

mixer/v1/dictionary.proto
```diff
@@ -0,0 +1,39 @@
...
+
+option (gogoproto.goproto_getters_all) = false;
+option (gogoproto.equal_all) = false;
+option (gogoproto.gostring_all) = false;
+
+message DictionaryRequest {
+  // The set of words in the request dictionary.
+  //
+  // Once the request dictionary has been set, subsequent request messages for the connection can
+  // use dictionary indices to effeciently encode strings.
```
douglas-reid: typo here with `effeciently`.

mixer/v1/dictionary.proto
```diff
@@ -0,0 +1,39 @@
...
+  repeated string words = 1;
+}
+
+message DictionaryResponse {
+  // The set of words in the response dictionary.
+  //
+  // Once the request dictionary has been set, subsequent response messages for the connection can
+  // use dictionary indices to effeciently encode strings.
```
douglas-reid: same typo.

mixer/v1/service.proto
```diff
@@ -17,21 +17,30 @@ syntax = "proto3";
...
 import "mixer/v1/quota.proto";
 
-// The Istio Mixer API
+// The Istio mixer's API
 service Mixer {
   // Checks preconditions before performing an operation.
   // The preconditions enforced depend on the set of supplied attributes
   // and the active configuration.
-  rpc Check(stream CheckRequest) returns (stream CheckResponse) {}
+  rpc Check2(CheckRequest) returns (CheckResponse) {}
```
douglas-reid: instead of `2`, can we give these semantic (temporary) names?  like `CheckUnary`?

or, alternatively, should we just define a different service, called `MixerUnary` (or maybe just `Mixer` and rename the streaming to `MixerStreaming`?

Then the rpc names could be left as is.

mixer/v1/attributes.proto
geeknoid: Yes, the gRPC docs claim this provides better encoding when negative values are used.

mixer/v1/service.proto
geeknoid: I know it's ugly, but its extremely temporary and was the most expedient approach. I'd rather not have to change a bunch of stuff at the moment to use better names when I'll need to rename it all again once the streaming API is deleted.

mixer/v1/dictionary.proto
geeknoid: Seems that my Gogland spell-checker is on the fritz at the moment :-(

Typos are fixed.

mixer/v1/attributes.proto
```diff
@@ -24,102 +24,54 @@ option (gogoproto.goproto_getters_all) = false;
...
-  // Once a dictionary has been sent over, it stays in effect until
-  // a new dictionary is sent to replace it. The first request sent on a
-  // stream must include a dictionary, otherwise the mixer can't process
-  // any attribute updates.
-  //
-  // Dictionaries are independent of the attribute context and are thus global
-  // to each gRPC stream.
-  map<int32, string> dictionary = 1;
+  // The message-level dictionary.
+  repeated string words = 13;
```
qiwzhang: so, 0 is a valid index?

I prefer using positive index for message -level dictionary and negative for connection-level.  
For c++ client,  most likely, we could not set connection-level.   It is easier to use positive indexes.

mixer/v1/quota.proto
```diff
@@ -47,21 +44,22 @@ message QuotaRequest {
...
+  // TO BE REMOVED
   int64 request_index = 1;
+}
 
+message QuotaResponse {
   // The attributes to use for this response
-  Attributes attribute_update = 2;
-
-  // Indicates whether the quota request was successfully processed.
-  google.rpc.Status result = 3 [(gogoproto.nullable) = false];
```
qiwzhang: WE are not passing status?   using amount=0 to indicate failure?

mixer/v1/check.proto
```diff
@@ -27,23 +27,21 @@ option (gogoproto.gostring_all) = false;
...
-  // Indicates whether or not the preconditions succeeded
-  google.rpc.Status result = 3 [(gogoproto.nullable) = false];
+  Attributes attribute_update = 2 [(gogoproto.nullable) = false];
 
   // The amount of time for which this result can be considered valid, given the same inputs
   google.protobuf.Duration expiration = 4 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
+
+  // TO BE REMOVED
+  int64 request_index = 1;
+  google.rpc.Status result = 3 [(gogoproto.nullable) = false];
```
qiwzhang: I think we still need Status.

mixer/v1/service.proto
qiwzhang: I think it will be cleaner to have two different services with two set of request and response messages.

We can support two Services: Unary and streaming at the same time.

geeknoid: It's a lot of work on the mixer side to support two distinct services.

As soon as the proxy starts using the unary APIs, I will delete all the streaming code. So I don't want to do spend any time making the dual-mode look nice.

mixer/v1/attributes.proto
geeknoid: I was told by the gRPC Go guys that every gRPC implementation makes connection stuff visible to the caller. Basically, the caller has to explicitly connect to the server via some Dial call to produce a usable client object. Using this client object, you would do a SetDictionary call to set that connection's dictionary. So every time you connect to the mixer, you issue a single SetDictionary call. There is a 50-60% increase in RPC overhead between using a mix of connection-level and message-level dictionary compared to only using message-level.

In the current design, index 0..N is for the connection-level dictionary, and -1..-N is for the message-level dictionary. You have to use (-index-1) to go from an index number to the proper slot in the message-level dictionary.

Can you confirm whether you can or cannot use SetDictionary?

mixer/v1/quota.proto
geeknoid: I don't know how this is exposed in C++, but for the Go bindings all unary API calls return both a message response and a Go error value. The status ends up being encoded in the returned error.

How is a gRPC API modeled in C++ client code?

qiwzhang: I see.  we can use Status from Unary call.

mixer/v1/attributes.proto
qiwzhang: Stub is across tcp connection.   If server restarted, client will not be notified.  We could not Set global dict this way.

mixer/v1/service.proto
douglas-reid: I'm not sure it would be a ton of work to support two services.

But, if we don't choose to support that path, I suggest we do something like create a branch for the unary change. We would develop against this branch on mixerclient and mixer side in related branches, and then cut over when everything is working. That would avoid awkward naming and keep the intent clear (unary not supported yet, etc.).

PR #73, opened 2017-04-12
Introduce AttributeManifests and make mixer config more readable by removing LabelDescriptor

ZackButcher:
We remove `LabelDescriptor` and replace all usages with a `map<string, ValueType>`. This makes descriptors look like the adapter and aspect configs that use them. We also introduce the concept of an `AttributeManifest` and replace the global config's list of `AttributeDescriptor`s with a list of `AttributeManifest`s instead.

Example `MetricDescriptor` using `LabelDescriptors`:

```
name: request_count
kind: COUNTER
value: INT64
description: request count by source, target, service, and code
labels:
- name: source
  valueType: STRING
- name: target
  valueType: STRING
- name: method
  valueType: STRING
- name: response_code
  valueType: INT64
```

After this change, the `MetricDescriptor` will look like:
```
name: request_count
kind: COUNTER
value: INT64
description: request count by source, target, service, and code
labels:
  source: STRING
  target: STRING
  method: STRING
  response_code: INT64
```

Here's what aspect configuration, which is how a user tells us runtime values to use for a label, looks like today:
```
descriptorName: request_count
value: "1"
labels:
    source: source.name | "unknown"
    target: target.service | "unknown"
    method: request.method | "unknown"
    response_code: response.code | 200
```

Note how the aspect configuration and the descriptor are now nearly identical in structure.


mixer/v1/config/descriptor/log_entry_descriptor.proto
```diff
@@ -16,44 +16,44 @@ syntax = "proto3";
...
+
+  // The template that will be populated with labels at runtime to generate a log message; the labels describe the
+  // parameters for this template.
+  //
+  // The template strings must conform to go's text/template syntax.
+  // TODO: unify templates to use our expression language when its finalized
+  string log_template = 5;
+
+  // Labels describe the parameters of this log's template string. At runtime expressions will be evaluated to provide
+  // values for each label. Label names are mapped to expressions in the service config as part of configuring the
```
geeknoid: Could you scour and remove any reference to "service config" and as many references to "service" as possible in the protos?

mixer/v1/config/descriptor/metric_descriptor.proto
```diff
@@ -82,5 +81,5 @@ message MetricDescriptor {
   // Labels that dimension the data recorded by this metric. The metric definition allows the user to
   // map attribute expressions to actual values for these labels at run time; the result of the evaluation
   // must be of the type described by the kind for each label.
```
geeknoid: Could use a comment update....

ZackButcher:
Removed all of these occurrences (except in `cfg.proto`; `ServiceConfig` is still defined there and used correctly in that comment):
```
$ grep -rnw . -e "service config"
./mixer/v1/config/cfg.proto:32://# service config
./mixer/v1/config/descriptor/log_entry_descriptor.proto:56:  // values for each label. Label names are mapped to expressions in the service config as part of configuring the
./mixer/v1/config/descriptor/monitored_resource_descriptor.proto:39:  // names are mapped to expressions in the service config.
./mixer/v1/config/descriptor/principal_descriptor.proto:30:  // names are mapped to expressions in the service config.
```

Updated metric_descriptor. All of the other entries appear to be valid comments so I'm leaving them in place:
```
$ grep -rnw . -e "service"
<snip 103 entries in cfg.proto about services>
./mixer/v1/config/descriptor/attribute_descriptor.proto:29:// and consumed by different services. For example, a frontend service can
./mixer/v1/config/descriptor/attribute_descriptor.proto:30:// generate an authenticated user attribute and pass it to a backend service for
./mixer/v1/config/descriptor/metric_descriptor.proto:41:// This is provided in the service's config; using our above descriptor we might describe the metric as:
```

PR #76, opened 2017-04-13
proxy config: per retry timeout

rshriram:
This PR adds support for specifying timeout per retry in our route rule specification (corresponding support has been added in Envoy lyft/envoy#756). Together, we now have global timeout budget, per retry timeout with jitter between retries, max retries, and policies that govern when a retry should be attempted.

### DIFF
--- a/proxy/v1/config/cfg.proto
+++ b/proxy/v1/config/cfg.proto
@@ -303,9 +303,12 @@ message HTTPRetry {
     /// will be determined automatically (25ms+). Actual number of retries
     /// attempted depends on the http_timeout
     int32 attempts = 1;
+    /// Timeout per retry attempt for a given request. 
+    /// Specified in seconds.nanoseconds format.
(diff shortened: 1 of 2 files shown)


rshriram:
Yes, thats the Plan. I wanted separate commit IDs, so that I can separate the code changes in manager.

proxy/v1/config/cfg.proto
```diff
@@ -303,9 +303,12 @@ message HTTPRetry {
     /// will be determined automatically (25ms+). Actual number of retries
     /// attempted depends on the http_timeout
     int32 attempts = 1;
+    /// Timeout per retry attempt for a given request. 
+    /// Specified in seconds.nanoseconds format.
+    double per_try_timeout_seconds = 2;
```
douglas-reid: why not make the new field == 3 instead? Will this break any existing uses?

PR #79, opened 2017-04-14
double to protobuf.Duration

rshriram:


proxy/v1/config/cfg.md
```diff
@@ -172,7 +172,7 @@ Retry policy to use when a request fails.
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
 | attempts | [int32](#int32) | optional | Number of retries for a given request. The interval between retries will be determined automatically (25ms+). Actual number of retries attempted depends on the http_timeout |
-| per_try_timeout_seconds | [double](#double) | optional | Timeout per retry attempt for a given request. Specified in seconds.nanoseconds format. |
+| per_try_timeout | [google.protobuf.Duration](#double) | optional | Timeout per retry attempt for a given request. format: 1s/1ms/1us/1ns. |
```
kyessenov: i think you need to change `#double` as well.

ZackButcher:
Lets just delete the markdown file in this repo. istio/istio.github.io is the source of truth for these kinds of docs now anyway. When we pushed the markdown into this repo originally, istio.github.io and istio.io didn't exist yet.

proxy/v1/config/cfg.md
rshriram: @ZackButcher I added a new field (per_try_timeout) and then renamed all the duration fields ( s/_seconds//). If you plan to change these two things in the main docs, then feel free to delete this file.

rshriram: And ignore the #double .. It doesn't link out to anything. Artifact of the tool I used a while ago.

ZackButcher: I am planning on grabbing the latest protos and re-generating all the docs before our alpha (I've been putting it off in part because a bunch of that stuff is changing as we get close to alpha), so I will be updating the main repo in the near future. I can delete this MD file when I do that.

PR #80, opened 2017-04-14
Collapse AttributeDescriptor into a map<string, AttributeInfo>

ZackButcher:
This is the next step in making all of configs more readable and consistent. Today attributes are specified like:

```yaml
manifests:
  - name: istio-proxy
    revision: "1"
    attributes:
    - name: source.name
      valueType: STRING
      description: The name of the source
    - name: target.name
      valueType: STRING
      description: The name of the target
    - name: origin.ip
      valueType: IP_ADDRESS
      description: The IP address of the service that originated the request
```

with this change, they will be specified like:

```yaml
manifests:
  - name: istio-proxy
    revision: "1"
    attributes:
      source.name:
        valueType: STRING
        description: The name of the source
      target.name:
        valueType: STRING
        description: The name of the target
      origin.ip:
        valueType: IP_ADDRESS
        description: The IP address of the service that originated the request
```
<hr>

I'm not sure what we want to do with the comments that were on the `AttributeDescriptor` message before. AFAIK they've been superseded by the [Attribute concept page on istio.io](https://istio.io/docs/concepts/attributes.html).


geeknoid:
Would you mind renaming GlobalConfig.manifests to GlobalConfig.attribute_manifests? I expect to have at least one more kind of manifest in a few weeks to capture the event types that a service can produce.

wora:
I do not agree with this change yet.
- For all config schemas under google.api.*, the proto comments are the official spec and the documentation is generated from proto files.
- The descriptor schema is used by other Istio configs. I think having a map<name, info> is a fine choice, but we have to look at overall pros and cons. I don't think this PR provides enough justification.

I think we should review the overall impact before making this incremental change.

mandarjog:
Using a map conveys additional information about uniqueness of names.

However representing this struct in json schema is non standard. [ (key1,value1) ...] is preferred.

Using "iPhone 9";

geeknoid:
It sure makes for nicer looking YAML, which IMO trumps the other stuff...

kyessenov:
One difficulty I can see using maps in YAMLs is non-deterministic ordering. JSON does not guarantee the order of the fields/maps. This can make it difficult to implement PATCH in the longer run.

ZackButcher:
The documentation on the FieldMask proto states that repeated fields are overwritten by `PATCH`, so I don't think that should factor in to our decision:

>  If a repeated field is specified for an update operation, the existing repeated values in the target resource will be overwritten by the new values.

(I couldn't find a good link to the public proto definition file where that comment lives, but it's in the language specific docs generated from the .proto file; separately there's a bug about fieldmask in golang so it's not linkable.)

PR #81, opened 2017-04-19
Add ingress settings

kyessenov:


proxy/v1/config/cfg.proto
```diff
@@ -98,13 +98,20 @@ message ProxyMeshConfig {
   // Connection timeout used by the Envoy clusters
   google.protobuf.Duration connect_timeout = 10;
 
+  // Class of ingress resources to be processed by Istio ingress controller.
+  // This corresponds to the value of "kubernetes.io/ingress.class" annotation.
+  string ingress_class = 20;
+
+  // Use Istio ingress controller for unclassified ingress resources
+  bool ingress_default_controller = 21;
```
douglas-reid: would it be clearer to call this `use_ingress_default_controller`?

PR #82, opened 2017-04-20
Add distribution kind to metric_descriptor.

douglas-reid:
This is needed to provide instruction to backends and adapters to treat
values as observations within a histogram. This will enable
adapters/backends to more readily monitor values that are distributions
when they have native support for distribution values.

Example uses:
- prometheus (histograms:
https://prometheus.io/docs/concepts/metric_types/#histogram)
- statds (timings:
https://github.com/etsy/statsd/blob/master/docs/metric_types.md#timing)
- stackdriver (distributions:
https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TypedValue#Distribution)

The documentation for the bucket definitions (and actual proto messages)
are taken from existing stackdriver custom metrics protos for
distributions.

NOTE: individual reports can still be single-valued. Adapters that
support distributions will be responsible for appropriately
summarizing/transforming single reports. Adapters that do not support
distributions can treat them as counters or gauges (depending). At the
moment, we do not have any adapters that cannot support distributions.

### DIFF
--- a/mixer/v1/config/descriptor/metric_descriptor.proto
+++ b/mixer/v1/config/descriptor/metric_descriptor.proto
@@ -48,7 +48,7 @@ import "mixer/v1/config/descriptor/value_type.proto";
 //        api_method: $apiMethod | "unknown"
 //        # either the attribute named 'responseCode' or the literal int64 500; must eval to an int64
 //        response_code: $responseCode | 500
-//      
+//
 message MetricDescriptor {
   // The name of this descriptor. This is used to refer to this descriptor in other contexts.
   string name = 1;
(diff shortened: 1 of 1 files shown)


mixer/v1/config/descriptor/metric_descriptor.proto
```diff
@@ -82,4 +92,80 @@ message MetricDescriptor {
   // map attribute expressions to actual values for these labels at run time; the result of the evaluation
   // must be of the type described by the kind for each label.
   map<string, ValueType> labels = 6;
+
+  message BucketsDefinition {
```
ZackButcher: Nit: reorder ths message a bit: put the submessages first then the oneof. It feels weird to name types that haven't been introduced yet.

mixer/v1/config/descriptor/metric_descriptor.proto
```diff
@@ -82,4 +92,80 @@ message MetricDescriptor {
...
+
+      // The explicit buckets.
+      Explicit explicit_buckets = 3;
+    }
+
+    // Specifies a linear sequence of buckets that all have the same width
+    // (except overflow and underflow). Each bucket represents a constant
+    // absolute uncertainty on the specific value in the bucket.
+    //
+    // There are `num_finite_buckets + 2` (= N) buckets. Bucket `i` has the
```
ZackButcher: Can you update this comment and the similar comments on the other two messages to describe the `+ 2` buckets (for 0 and overflow, I assume)?

douglas-reid:
Thanks for the review!

PR #84, opened 2017-04-26
Add ingress service to the proxy mesh

ijsnellf:
Needed for ingress to expose its external IP. See: https://github.com/istio/manager/pull/562

### DIFF
--- a/proxy/v1/config/cfg.proto
+++ b/proxy/v1/config/cfg.proto
@@ -133,6 +133,9 @@ message ProxyMeshConfig {
 
   // Path to the secrets used by the authentication policy
   string auth_certs_path = 101;
+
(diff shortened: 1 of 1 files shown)


proxy/v1/config/cfg.proto
```diff
@@ -133,6 +133,9 @@ message ProxyMeshConfig {
 
   // Path to the secrets used by the authentication policy
   string auth_certs_path = 101;
+
+  // Optional service fronting the ingress controllers
```
kyessenov: You can drop "optional" since everything is optional in this file. Maybe move it closer to egress or ingress settings? You can renumber the fields since we don't need to keep wire format same.

ijsnellf: Done

PR #85, opened 2017-04-27
Update mixer API documentation

ZackButcher:
Update documentation on our API surface to line up with today's world. Various other small touchups.

### DIFF
--- a/mixer/v1/config/cfg.proto
+++ b/mixer/v1/config/cfg.proto
@@ -50,10 +50,10 @@ package istio.mixer.v1.config;
 message ServiceConfig {
   option deprecated = true;
 
-  // subject is unique for a config type
+  // Optional. Subject is unique for a config type.
   // 2 config with the same subject will overwrite each other
   string subject = 1;
-  // revision of this config. This is assigned by the server
+  // Optional. revision of this config. This is assigned by the server
   string revision = 2;
   repeated AspectRule rules = 3;
 }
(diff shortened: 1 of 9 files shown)


mixer/v1/config/cfg.proto
```diff
@@ -62,63 +58,100 @@ message ServiceConfig {
   repeated AspectRule rules = 3;
 }
 
-// AspectRules are intent based
+// An AspectRule is a selector and a set of intentions to be executed when the
+// selector is `true`. The selectors of the this rules child AspectRules are only
```
geeknoid: of the this rule, bad fragment.

mixer/v1/config/cfg.proto
```diff
@@ -62,63 +58,100 @@ message ServiceConfig {
...
+// ```
 message Aspect {
+  // Required. The kind of aspect this intent is targeting.
   string kind = 1;
-  string adapter = 2; // optional, allows specifying an adapter
-  // Struct representation of a proto defined by the aspect
+  // Optional. The name of the adapter this Aspect targets. If no name is provided,
+  // Mixer will use the adapter of the target kind named "default".
+  string adapter = 2;
+  // Struct representation of a proto defined by the each aspect kind.
```
geeknoid: the each

mixer/v1/config/cfg.proto
```diff
@@ -62,63 +58,101 @@ message ServiceConfig {
...
+//       src_consumer_id: source.user | source.uid
+//       target_response_status_code: response.code
+//       target_service_name: target.service
+// ```
 message Aspect {
+  // Required. The kind of aspect this intent is targeting.
   string kind = 1;
-  string adapter = 2; // optional, allows specifying an adapter
-  // Struct representation of a proto defined by the aspect
+  // Optional. The name of the adapter this Aspect targets. If no name is provided,
```
douglas-reid: should we only mark optional fields as `Optional` (and have the default be `Required`)?  Or only mark required fields as `Required` (and have optional be the default)?  Or do we really want both?

It seems we have a mix now of `Required`, `Optional`, and not annotated.

ZackButcher: I left off required where I thought the comment made it obvious is was required (e.g. `kind` in the various messages). I'll go back and mark everything explicitly.

PR #87, opened 2017-04-28
Decompose proxy proto into smaller files with additional docs

rshriram:
Added examples and reformatted docs in proxy proto and split it up into smaller files, to simplify doc generation. Using google.api.restriction to hide features from docs, that are not implemented in current manager/envoy.

No functional change. All fields and ordering of fields remain the same. Just splitting up and adding docs.


proxy/v1/config/routerule.proto
```diff
@@ -0,0 +1,274 @@
...
+    //(-- For proxies that support custom timeout policies --)
+    google.protobuf.Any custom = 2 [(google.api.field_visibility).restriction = "NOT_IMPLEMENTED"];
+  }
+}
+
+// Describes the retry policy to use when a HTTP request fails. For
+// example, the following rule sets the maximum number of retries to 3 when
+// calling ratings:v1 service, with a 2s timeout per retry attempt.
+//
+//     destination: "ratings.default.svc.cluster.local"
```
kyessenov: unnecessary quotes in destinations

proxy/v1/config/routerule.proto
```diff
@@ -0,0 +1,274 @@
...
+// only requests originating from "reviews:v2", with the "Cookie" header
+// containing "user=jason".
+//
+//     destination: ratings.default.svc.cluster.local
+//     match:
+//       source: reviews.default.svc.cluster.local
+//       sourceTags:
+//         version: v2
+//       httpHeaders:
+//         Cookie:
```
kyessenov: header is case insensitive, use "cookie"

proxy/v1/config/destpolicy.proto
```diff
@@ -0,0 +1,194 @@
...
+// for more details.
+message CircuitBreaker {
+  // Parameters to tune Envoy's circuit breaker configuration. A simple
+  // circuit breaker can be set based on a number of criteria such as
+  // connection and request limits. For example, the following destination
+  // policy sets a limit of 100 connections to "reviews" service version
+  // "v1" backends. 
+  //
+  //     destination: reviews.default.svc.cluster.local
+  //     - policy:
```
kyessenov: this is not right YAML I think

rshriram:
fixed. PTAL.

proxy/v1/config/routerule.proto
```diff
@@ -0,0 +1,283 @@
...
+
+syntax = "proto3";
+
+import "google/protobuf/any.proto";
+import "google/protobuf/duration.proto";
+
+package istio.proxy.v1alpha.config;
+option go_package = "config";
+
+import "httpfault.proto";
```
ZackButcher: Here and elsewhere the proto imports need to be relative to the root of this repo, so they should read: `import "proxy/v1/config/httpfault.proto";`. You can verify you got it right by compiling the protos from the root dir, like: `protoc proxy/v1/config/routerrule.proto --go_out=.`

proxy/v1/config/destpolicy.proto
```diff
@@ -0,0 +1,194 @@
...
+    // Maximum number of requests per connection to a backend. Setting this
+    // parameter to 1 disables keep alive.
+    int32 http_max_requests_per_connection = 7;
+
+    // Maximum % of hosts in the load balancing pool for the destination
+    // service that can be ejected by the circuit breaker. Defaults to
+    // 10%.
+    int32 http_max_ejection_percent = 8;
+  }
+  oneof cb_policy {
```
ZackButcher: Just `policy`, the "`cb`" is redundant given the context of `message CircuitBreaker`.

proxy/v1/config/destpolicy.proto
```diff
@@ -0,0 +1,194 @@
...
+    SimpleCircuitBreakerPolicy simple_cb = 1;
```
ZackButcher: Same as above, just `simple`.

proxy/v1/config/l4fault.proto
```diff
@@ -0,0 +1,79 @@
...
+    int64 downstream_limit_bps = 2;
+    // bandwidth limits in "bits" per second between Envoy and upstream
+    int64 upstream_limit_bps = 3;
+
+    oneof throttle_after {
+      // Wait a while after the connection is established, before
+      // starting bandwidth throttling. This would allow us to inject fault
+      // after the application protocol (e.g., MySQL) has had time to
+      // establish sessions/whatever handshake necessary.
+      google.protobuf.Duration throttle_after_period = 4;
```
ZackButcher: I think these names are a bit redundant: `throttle_after: throttle_after_period: "1s"`. I like `throttle_after: bytes: 1024` and `throttle_after: period: "1s"` better personally.

proxy/v1/config/l4fault.proto
```diff
@@ -0,0 +1,79 @@
...
+
+      // Alternatively, we could wait for a certain number of bytes to be
+      // transferred to upstream before throttling the bandwidth.
+      double throttle_after_bytes = 5;
```
ZackButcher: Should this be a double? I think `uint32` or `uint64` would make more sense here.

kyessenov:
@ZackButcher comments to be addressed in a subsequent PR, will file an issue

ZackButcher:
These aren't all nits: routerule.proto will not compile right now, and likely won't in your BUILD rules either.

proxy/v1/config/destpolicy.proto
rshriram: I would like to keep the field renaming for a separate iteration as it involves changes in manager code, examples in docs, etc.

proxy/v1/config/l4fault.proto
rshriram: It's one of. Not a message. So throttle_after doesn't appear in the yaml right ?

proxy/v1/config/l4fault.proto
rshriram: I had uint32 few months ago. I think the feedback was to avoid unsigned ints.

proxy/v1/config/l4fault.proto
ZackButcher: Ug, the proto3 spec doesn't prescribe exactly how a oneof looks in json. You're right, it shouldn't appear in the json. It will appear in golang code handling the oneof, though. Not a big deal either way.

proxy/v1/config/destpolicy.proto
rshriram: same as comment below I guess?

PR #89, opened 2017-05-01
Fix proxy config imports and packages

kyessenov:


proxy/v1/config/routerule.proto
```diff
@@ -16,12 +16,10 @@ syntax = "proto3";
...
 import "google/protobuf/duration.proto";
+import "proxy/v1/config/httpfault.proto";
+import "proxy/v1/config/l4fault.proto";
 
-package istio.proxy.v1alpha.config;
-option go_package = "config";
-
-import "httpfault.proto";
-import "l4fault.proto";
+package istio.proxy.v1.config;
```
ZackButcher: BTW you can keep the `v1alpha1` package and these imports still work as `proxy/v1/...` (if y'all weren't ready to go to v1 yet).

kyessenov: it's weird that package and directory are inconsistent, so let's fix it now.

PR #90, opened 2017-05-02
formatting fixes for proxy proto

rshriram:
Addresses most comments from #87


rshriram:
Trying to avoid all field renaming.

PR #95, opened 2017-05-04
Introduce unary mixer APIs.

geeknoid:

### DIFF
--- a/mixer/v1/attributes.proto
+++ b/mixer/v1/attributes.proto
@@ -24,102 +24,35 @@ option (gogoproto.goproto_getters_all) = false;
 option (gogoproto.equal_all) = false;
 option (gogoproto.gostring_all) = false;
 
-// An instance of this message is delivered to the mixer with every
+// An instance of this message is delivered to Mixer with many
 // API call.
 //
-// The general idea is to leverage the stateful gRPC streams from the
-// proxy to the mixer to keep to a minimum the 'attribute chatter'.
-// Only delta attributes are sent over, multiple concurrent attribute
-// contexts can be used to avoid thrashing, and attribute indices are used to
-// keep the wire protocol maximally efficient.
-//
-// Producing this message is the responsibility of the mixer's client
-// library which is linked into different proxy implementations.
-//
-// The processing order for this state in the mixer is:
-//
-//   * Any new dictionary is applied
-//
-//   * The requested attribute context is looked up. If no such context has been defined, a
-//     new context is automatically created and initialized to the empty state. When a gRPC
-//     stream is first created, there are no attribute contexts for the stream.
-//
-//   * If reset_context is true, then the attribute context is reset to the
-//     empty state.
-//
-//   * All attributes to deleted are removed from the attribute context.
-//
-//   * All attribute changes are applied to the attribute context.
-//
+// Within this message, strings are referenced using integer indices into
+// one of two string dictionaries. Positive integers index into the global
+// deployment-wide dictionary, whereas negative integers index into the message-level
+// dictionary instead. The message-level dictionary is carried by the
+// `words` field of this message, the deployment-wide dictionary is determined via
+// configuration.
 message Attributes {
-  // A dictionary that provides a mapping of shorthand index values to
-  // attribute names.
-  //
-  // This is intended to leverage the stateful gRPC stream from the
-  // proxy to the mixer. This dictionary is sent over only when a
-  // stream to the mixer is first established and when the proxy's
-  // configuration changes and different attributes may be produced.
-  //
-  // Once a dictionary has been sent over, it stays in effect until
-  // a new dictionary is sent to replace it. The first request sent on a
-  // stream must include a dictionary, otherwise the mixer can't process
-  // any attribute updates.
-  //
-  // Dictionaries are independent of the attribute context and are thus global
-  // to each gRPC stream.
-  map<int32, string> dictionary = 1;
+  // The message-level dictionary.
+  repeated string words = 1;
 
-  // The attribute context against which to operate.
-  //
-  // The mixer keeps different contexts live for any proxy gRPC stream. This
-  // allows the proxy to maintain multiple concurrent 'bags of attributes'
-  // within the mixer.
-  //
-  // If the proxy doesn't want to leverage multiple contexts, it just passes
-  // 0 here for every request.
-  //
-  // The proxy is configured to use a maximum number of attribute contexts in order
-  // to prevent an explosion of contexts in the mixer's memory space.
-  //
-  // TODO: Consider removing support for this feature. The proxy can achieve
-  // the same effect using multiple gRPC streams. The benefit of using streams
-  // would be that the mixer would be in control of the maximum number of streams
-  // it allows, whereas with the current model the proxy could overwhelm the
-  // mixer by creating too many contexts.
-  int32 attribute_context = 2;
-
-  // When true, resets the current attribute context to the empty state before
-  // applying any incoming attributes.
-  //
-  // Resetting contexts is useful to constrain the amount of resources used by
-  // the mixer. The proxy needs to intelligently manage a pool of contexts.
-  // It may be useful to reset a context when certain big events happen, such
-  // as when an HTTP2 connection into the proxy terminates.
-  bool reset_context = 3;
-
-  // Attributes being updated within the specified attribute context. These maps
-  // add and/or overwrite the context's current set of attributes.
-  map<int32, string> string_attributes = 4;
-  map<int32, int64> int64_attributes = 5;
-  map<int32, double> double_attributes = 6;
-  map<int32, bool> bool_attributes = 7;
-  map<int32, google.protobuf.Timestamp> timestamp_attributes = 8 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
-  map<int32, google.protobuf.Duration> duration_attributes = 9 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
-  map<int32, bytes> bytes_attributes = 10;
-  map<int32, StringMap> stringMap_attributes = 11 [(gogoproto.nullable) = false];
-
-  // Attributes that should be removed from the specified attribute context. Deleting
-  // attributes which aren't currently in the attribute context is not considered an error.
-  repeated int32 deleted_attributes = 12;
-
-  // TODO: temporary workaround for gogoprotobuf code gen error (https://github.com/gogo/protobuf/issues/273). Remove
-  // when issue is fixed.
-  map<int32, google.protobuf.Timestamp> timestamp_attributes_HACK = 13 [(gogoproto.nullable) = false, (gogoproto.stdtime) = false];
-  map<int32, google.protobuf.Duration> duration_attributes_HACK = 14 [(gogoproto.nullable) = false, (gogoproto.stdduration) = false];
+  // Attribute payload. All `sint32` values represent indices into
+  // one of the word dictionaries. Positive values are indices into the
+  // global deployment-wide dictionary, negative values are indices into
+  // the message-level dictionary.
+  map<sint32, sint32> strings = 2;
+  map<sint32, int64> int64s = 3;
+  map<sint32, double> doubles = 4;
+  map<sint32, bool> bools = 5;
+  map<sint32, google.protobuf.Timestamp> timestamps = 6 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
+  map<sint32, google.protobuf.Duration> durations = 7 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
+  map<sint32, bytes> bytes = 8;
+  map<sint32, StringMap> string_maps = 9 [(gogoproto.nullable) = false];
 }
 
-// A map of string to string. The keys in these maps are from the current
-// dictionary.
+// A map of string to string. The keys and values in this map are dictionary
+// indices (see the [Attributes][istio.mixer.v1.Attributes] message for an explanation)
 message StringMap {
-  map<int32, string> map = 1;
+  map<sint32, sint32> entries = 1;
 }
--- a/mixer/v1/check.proto
+++ b/mixer/v1/check.proto
@@ -18,7 +18,6 @@ package istio.mixer.v1;
 
 import "gogoproto/gogo.proto";
 import "google/protobuf/duration.proto";
-import "google/rpc/status.proto";
 import "mixer/v1/attributes.proto";
 
 option (gogoproto.goproto_getters_all) = false;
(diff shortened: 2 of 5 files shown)


mixer/v1/attributes.proto
```diff
@@ -24,102 +24,45 @@ option (gogoproto.goproto_getters_all) = false;
...
+  map<sint32, StringMap> stringMap_attributes = 11 [(gogoproto.nullable) = false];
 
-  // When true, resets the current attribute context to the empty state before
-  // applying any incoming attributes.
-  //
-  // Resetting contexts is useful to constrain the amount of resources used by
-  // the mixer. The proxy needs to intelligently manage a pool of contexts.
-  // It may be useful to reset a context when certain big events happen, such
-  // as when an HTTP2 connection into the proxy terminates.
+  // TO BE REMOVED
```
ZackButcher: What blocks these being removed now? I thought we only had them to be compatible with old version of the proxy's mixer client (which would now be very old).

mixer/v1/attributes.proto
```diff
@@ -24,102 +24,45 @@ option (gogoproto.goproto_getters_all) = false;
...
-  // Once a dictionary has been sent over, it stays in effect until
-  // a new dictionary is sent to replace it. The first request sent on a
-  // stream must include a dictionary, otherwise the mixer can't process
-  // any attribute updates.
-  //
-  // Dictionaries are independent of the attribute context and are thus global
-  // to each gRPC stream.
-  map<int32, string> dictionary = 1;
+  // The message-level dictionary.
+  repeated string words = 13;
```
ZackButcher: Nice, this is way easier than the map<i32, str> version.

mixer/v1/attributes.proto
```diff
@@ -24,102 +24,45 @@ option (gogoproto.goproto_getters_all) = false;
...
-  // the same effect using multiple gRPC streams. The benefit of using streams
-  // would be that the mixer would be in control of the maximum number of streams
-  // it allows, whereas with the current model the proxy could overwhelm the
-  // mixer by creating too many contexts.
-  int32 attribute_context = 2;
+  // Attribute payload. All `sint32` values represent indices into
+  // one of the word dictionaries. Positive values are indices into the
+  // global deployment-wide dictionary, negative values are indices into
+  // the message-level dictionary.
+  map<sint32, sint32> string_attributes2 = 14;
```
ZackButcher: nit: we can probably just drop the `_attributes` postfix here and elsewhere, I think it's implied by the context of the `Attributes` message.

mixer/v1/attributes.proto
```diff
@@ -24,102 +24,45 @@ option (gogoproto.goproto_getters_all) = false;
...
-  // TODO: temporary workaround for gogoprotobuf code gen error (https://github.com/gogo/protobuf/issues/273). Remove
-  // when issue is fixed.
-  map<int32, google.protobuf.Timestamp> timestamp_attributes_HACK = 13 [(gogoproto.nullable) = false, (gogoproto.stdtime) = false];
-  map<int32, google.protobuf.Duration> duration_attributes_HACK = 14 [(gogoproto.nullable) = false, (gogoproto.stdduration) = false];
 }
 
-// A map of string to string. The keys in these maps are from the current
-// dictionary.
+// A map of string to string. The keys and values in this map are dictionary
+// indices (see the `Attributes` message for an explanation)
```
ZackButcher: You can refer to the `Attributes` message by changing this to:

> see the [`Attributes`][istio.mixer.v1.Attributes] message for ...

and doc gen will link them up correctly.

mixer/v1/attributes.proto
ZackButcher: But we probably don't want to do this pre-alpha, since it breaks proxy.

mixer/v1/attributes.proto
geeknoid: It's easier for me to leave all the old stuff alone and remove it all in one shots once I know the proxy is using the new API exclusively.

mixer/v1/attributes.proto
geeknoid: I like the idea. In order to make it possible to upgrade the proxy to use these new names, I think I need to introduce the new names now and support both in the mixer for a little while. I think I can do this relatively easily without too much boilerplate in Mixer.

mixer/v1/attributes.proto
geeknoid: Look at you, Mr DocGen expert now... :-)

I'll change it, thanks.

mixer/v1/check.proto
```diff
@@ -27,23 +27,23 @@ option (gogoproto.gostring_all) = false;
...
 message CheckResponse {
-  // Index of the request this response is associated with
-  int64 request_index = 1;
-
   // The attributes to use for this response
-  Attributes attribute_update = 2;
-
-  // Indicates whether or not the preconditions succeeded
-  google.rpc.Status result = 3 [(gogoproto.nullable) = false];
+  Attributes attributes = 5 [(gogoproto.nullable) = false];
```
mandarjog: Is this the cache key being returned by the check request?
If not we need it now, correct?

mixer/v1/check.proto
```diff
@@ -27,23 +27,23 @@ option (gogoproto.gostring_all) = false;
...
 
   // The amount of time for which this result can be considered valid, given the same inputs
   google.protobuf.Duration expiration = 4 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
+
+  // TO BE REMOVED
```
mandarjog: just request_index is deprecated?  Are we continuing with the attribute update protocol? 

Otherwise please update to specify what is being removed.

mixer/v1/check.proto
geeknoid: How to generate a cache key must be known statically by the caller. Think of it, how do you look something up from the cache without knowing the key?

mixer/v1/check.proto
geeknoid: Here and elsewhere in this change everything below the 'TO BE REMOVED' comment is deprecated.

attribute_update is replaced with the attributes field above. It's really just a rename, but I don't want to break compat so I'm introducing a new field for it.

mixer/v1/service.proto
```diff
@@ -27,25 +27,30 @@ import "mixer/v1/quota.proto";
...
+//
+// - *Telemetry Reporting*. Enables services to report logging and monitoring.
+// In the future, it will also enable tracing and billing streams intended for
+// both the service operator as well as for service consumers.
 service Mixer {
   // Checks preconditions before performing an operation.
   // The preconditions enforced depend on the set of supplied attributes and
   // the active configuration.
-  rpc Check(stream CheckRequest) returns (stream CheckResponse) {}
+  rpc Check2(CheckRequest) returns (CheckResponse) {}
```
douglas-reid: i believe we've discussed this before, but why not create a separate service for these new methods and messages?

That seems like a cleaner way to transition and would avoid having all of these ugly *2 names.

It shouldn't be all that ugly in mixer to add handling for multiple entry points into the same dispatch code and should allow clean parallel development.

geeknoid:
I think it is simply not worth any effort here. This is transient and will
all get cleaned up in a couple weeks.

mixer/v1/service.proto
geeknoid: It's actually more involved. If I have two services with methods called Check but with different signatures, that means I'll need to have two different types implementing the interface. So a bunch of code in grpcServer.go would need to be split/replicated. It's just not worth the effort.

geeknoid:
PTAL

- Renamed fields per Zack's suggestion.

- Clarified the TO BE REMOVED comments per Mandar's comment.

- Will consider add reserved annotations when the old stuff is removed.  It's not clear whether this is useful at all, since removing the old stuff is a fully breaking change anyway, so we may as well just renumber all the fields anew and start with a clean slate. We'll see when we get there...

- Look at what it would take to avoid having a temporary 2 suffix on method names. It turns out to be  a pain, and I stil end up having the service name itself with a 2 suffix (Mixer2). So it's not worth the hassle, I'm leaving this alone.

mandarjog:
From the performance work that @costinm did, we have come across a distributed dead lock between the mixer client and the mixer server.

Both client and server are blocked on `stream.Write`, no one is reading and the stream is full.
There are several bugs that caused this problem that are being addressed, however having a clearer mixer api would have also helped.  

Since Check and Quota are so closely related they should be one call.  
```
PreRequest {
 CheckRequest
 ConditionalQuotaRequest   --> only executed if check succeeds 
}
```
This is not only more performant but it is also correctly expresses semantics correctly which are not apparent when Check and Quota are separate unrelated calls.
By encapsulating we are protecting a call with side effects by a conditional. 
It also makes for a cleaner caching story. 
@louiscryan 
This is orthogonal to batching unrelated requests.

Details of the deadlock.
https://github.com/istio/proxy/issues/300#issuecomment-300338197

costinm:
AFAIK there are 2 kinds of quota (at least): one applied per user or other
attributes
resulting from CheckRequest, but the other is to rate limit access to the
resource
and protect the CheckRequest itself.

Also quota has pretty different perf characteristics - you can use
batching, sampling,
buckets, etc.

It may be worth keeping it separate - in terms of cost it's not that much
if it's properly
optimized. Having rate limit/quota in front of authentication can really
save the day...

geeknoid:
​Mandar, can you explain why combining the two calls into one would avoid
the deadlock? I have a hard time believing that this is so.​

mandarjog:
ref [dead lock table](https://github.com/istio/proxy/issues/300#issuecomment-300345999).

The API design was not the *cause* of the deadlock, but on the mixer client side it introduced complexity of having 2 successive async calls to the mixer. It was this complexity that lead to the bug. 

If the common case is to use those 2 calls together, then that should be part of the API, making it simple to write the mixer client.

kyessenov:
hint hint: we need futures in GRPC.
i think the proxy change and deadlines in mixer are probably enough to fix deadlock.

mixer/v1/attributes.proto
```diff
@@ -24,102 +24,52 @@ option (gogoproto.goproto_getters_all) = false;
...
+  // global deployment-wide dictionary, negative values are indices into
+  // the message-level dictionary.
+  map<sint32, sint32> strings = 14;
+  map<sint32, int64> int64s = 15;
+  map<sint32, double> doubles = 16;
+  map<sint32, bool> bools = 17;
+  map<sint32, google.protobuf.Timestamp> timestamps = 18 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
+  map<sint32, google.protobuf.Duration> durations = 19 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
+  map<sint32, bytes> bytes = 20;
+  map<sint32, StringMap> stringMaps = 21 [(gogoproto.nullable) = false];
```
louiscryan: As discussed offline we can also do

message Value {
   sint32 name = 1;
   oneof value_entry {
     sint32 word_ref = 2;
     int64 int_value = 3;
     double double_value = 4;
     ...
  }
}

repeated Value values = 1;

If we wanted to avoid the object materialization costs associated with this we create a custom protobuf deserializer which is not particularly hard. The only concern that would remain is how quickly/easily clients can produce this structure.

mixer/v1/attributes.proto
```diff
@@ -24,102 +24,52 @@ option (gogoproto.goproto_getters_all) = false;
...
-  map<int32, google.protobuf.Timestamp> timestamp_attributes_HACK = 13 [(gogoproto.nullable) = false, (gogoproto.stdtime) = false];
-  map<int32, google.protobuf.Duration> duration_attributes_HACK = 14 [(gogoproto.nullable) = false, (gogoproto.stdduration) = false];
 }
 
-// A map of string to string. The keys in these maps are from the current
-// dictionary.
+// A map of string to string. The keys and values in this map are dictionary
+// indices (see the [Attributes][istio.mixer.v1.Attributes] message for an explanation)
 message StringMap {
+  map<sint32, sint32> entries = 2;
```
louiscryan: This should be a definite improvement, nice.

mixer/v1/check.proto
louiscryan: Whatever we decide is the policy for cache key generation it needs to be documented here.

mixer/v1/quota.proto
```diff
@@ -47,21 +44,21 @@ message QuotaRequest {
...
+}
 
-  // The attributes to use for this response
-  Attributes attribute_update = 2;
-
-  // Indicates whether the quota request was successfully processed.
-  google.rpc.Status result = 3 [(gogoproto.nullable) = false];
-
+message QuotaResponse {
   // The amount of time the returned quota can be considered valid, this is 0 for non-expiring quotas.
```
louiscryan: Document what null implies here

mixer/v1/report.proto
```diff
@@ -26,20 +26,30 @@ option (gogoproto.gostring_all) = false;
 
 // Used to report telemetry after performing an action.
 message ReportRequest {
-  // Index within the stream for this request, used to match to responses
+  // The attributes to use for this request.
+  //
+  // These attributes are delta-encoded. In other words, each individual
```
louiscryan: Do we really want sequential delta encoding like this, it seems like a lot of work for the client.

I think clients would rather just build up a single dictionary over the batch here and then emit independent Atrtibutes using that dictionary as the base. Sequential delta encoding will be much harder to implement for the client and the server and is unlikely to yield much benefit over a large shared dictionary if any.

mixer/v1/service.proto
louiscryan: Cut'n'paste is not the worst evil and it would make reviewing this in isolation easier.

mixer/v1/attributes.proto
geeknoid: Since I have this current API implemented, I was going to check that in first, and then evaluate whether this new approach is usable and performs well.

mixer/v1/report.proto
geeknoid: The model I defined here supports a shared dictionary.

If a client doesn't want to support delta-encoding, it can just send a full complement of attributes with every message. I added a comment to that effect.

I think this can be a big win. If we batch 1000 metrics where only a handful of attributes change from message to message, the payload will be substantially smaller using delta encoding.

mixer/v1/quota.proto
geeknoid: Expanded the comment.

mixer/v1/attributes.proto
louiscryan: Fair enough, we can keep iterating once there are some benchmarks

mixer/v1/report.proto
louiscryan: Really? I'd like to see some quantification of that. If every string has become an int reference into a common dictionary I suspect that delta encoding is (a) at best a very marginal improvement and (b) slower because of the increased complexity involved in handle add/removes

PR #99, opened 2017-05-15
Add configuration fields for sidecar injection

ayj:


ayj:
cc @kyessenov @andraxylia - This configuration will be consumed by k8s admission-controller/initializer and/or `kube-inject` to determine proper images to inject into end-user pod templates.

kyessenov:
I'm not a supporter of embedding infrastructure-level concerns into proxy mesh config. If you think this is the way forward, I'll let other decide.

ayj:
Alternative suggestions to make client-side `kube-inject` work short-term without istio.VERSION dependency? Or would you prefer continued use of sourcing istio.VERSION until `kube-inject` logic is moved server-side (pre-AC/initializer)?

kyessenov:
Hardcode hub and tag to be :0.1 and image pull policy always. Then retag 0.1.x as 0.1 after each release.

PR #101, opened 2017-05-18
Add language indicator to code snippets

kyessenov:


geeknoid:
I've added support this morning such that code blocks don't need to specify
a language and the rendering will look OK.

As far as how much indent to use in markdown to get proper generated HTML
for code blocks, it's a pain indeed.

PR #102, opened 2017-05-18
Fix copyright.

geeknoid:


ZackButcher:
We have different LICENSE file formats in different repos now. istio/mixer has the line you deleted at the top, for example.

PR #104, opened 2017-05-24
New config model (renames and restructuring)

guptasu:
Proto changes reflecting the new revised mixer configuration object model (https://docs.google.com/document/d/1elqeOTDCGSaKP7GI0mrNkZDtt0dq8TbarEt0jUNTk-M/edit).

### DIFF
--- a/mixer/v1/config/cfg.proto
+++ b/mixer/v1/config/cfg.proto
@@ -33,7 +33,7 @@ package istio.mixer.v1.config;
 // ```yaml
 // subject: "namespace:ns1"
 // revision: "1011"
-// action_rules:
+// actionRules:
 // - selector: target.service == "*"
 //   actions:
 //   - handler: prometheus-handler
@@ -75,12 +75,14 @@ message ServiceConfig {
 
   // This is about to get deprecated. The configuration will move to
   // [action_rules][istio.mixer.v1.config.ServiceConfig.action_rules].
-  repeated AspectRule rules = 3;
+  // (== deprecation_description Please use action_rules to express the intention associated with this ServiceConfig ==)
+  repeated AspectRule rules = 3 [deprecated = true];
 
   // Optional. List of constructors that can be referenced from [actions][istio.mixer.v1.config.Action.instances]
   repeated Constructor constructors = 4;
 
-  // Optional. List of actions that apply for this service.
+  // Optional. List of actions that apply for this service
+  // TODO: Rename this to rules once we delete the AspectRule field.
   repeated ActionRule action_rules = 5;
 }
 
@@ -154,7 +156,7 @@ message Aspect {
 // params:
 // ```
 message Adapter {
-  // Required, must be unique per `kind`. Used by [Aspect][istio.mixer.v1.config.Aspect]
+  // Required. must be unique per `kind`. Used by [Aspect][istio.mixer.v1.config.Aspect]
   // to refer to this adapter. The name "default" is special: when an Aspect does not
   // specify a name, the Adapter named "default" of the same `kind` is used to execute
   // the intention described by the [AspectRule][istio.mixer.v1.config.AspectRule]s.
@@ -166,7 +168,7 @@ message Adapter {
   // Required. The name of a specific adapter implementation. An adapter's
   // implementation name is typically a constant in its code.
   string impl = 3;
-  // Optional, depends on adapter implementation. Struct representation of a
+  // Optional. Depends on adapter implementation. Struct representation of a
   // proto defined by the implementation; this varies depending on `impl`.
   google.protobuf.Struct params = 4;
 }
@@ -196,7 +198,7 @@ message GlobalConfig {
   // Optional. List of handlers that can be referenced from [actions][istio.mixer.v1.config.Action.handler]
   repeated Handler handlers = 9;
 
-  // Optional. List of types that can be referenced from [Constructors][istio.mixer.v1.config.Constructor.type]
+  // Optional. List of types that can be referenced from [constructors][istio.mixer.v1.config.Constructor.type]
   repeated Type types = 10;
 }
 
@@ -295,11 +297,11 @@ message EmailAddress {
 }
 
 
-// An ActionRule is a selector and a set of intentions to be executed when the
-// selector is `true`.
+// An Rule is a selector and a set of intentions to be executed when the
+// selector is `true`
 //
 // The following example instructs Mixer to invoke 'prometheus-handler' handler for all services and pass it the object
-// constructed using the constructor 'RequestCountByService'
+// constructed using the 'RequestCountByService' constructor.
 //
 // ```yaml
 // action_rules:
@@ -309,9 +311,9 @@ message EmailAddress {
 //     instances:
 //     - RequestCountByService
 // ```
-message ActionRule {
+message Rule {
   // Required. Selector is an attribute based predicate. When Mixer receives a
-  // request it evaluates all selectors in scope and executes the ActionRule for all
+  // request it evaluates all selectors in scope and executes the `actions` for all
   // selectors that evaluated to true.
   //
   // A few example selectors:
@@ -328,7 +330,8 @@ message ActionRule {
 
   // Optional. Nested action rules; their selectors are evaluated if this selector
   // predicate evaluates to `true`.
-  repeated ActionRule action_rules = 3;
+  // TODO: Rename this to rules once we delete the AspectRule field.
+  repeated Rule action_rules = 3;
 }
 
 // Action describes which [Handler][istio.mixer.v1.config.Handler] to invoke and what data to pass to it for processing.
@@ -342,22 +345,24 @@ message ActionRule {
 //   - RequestCountByService
 // ```
 message Action {
-  // Required: The handler to invoke.
+  // Required: The handler to invoke. Must match the `name` of a [Handler][istio.mixer.v1.config.Handler.name] in scope.
   string handler = 2;
 
-  // Required: List of constructor names that will be evaluated during request execution, and the constructed object
-  // will be passed to the handler for processing.
+  // Required: The name must match the `name` of the [Constructor][istio.mixer.v1.config.Constructor.name]s in scope.
+  // Referenced constructors are evaluated by resolving the attributes/literals for all the fields.
+  // The constructed objects are then passed to the `handler` referenced within this action.
   repeated string instances = 3;
 }
 
-// Constructor tell Mixer how to create instances of particular types.
+// Constructor tell Mixer how to create instances of particular [Type][istio.mixer.v1.config.Type].
 //
 // Constructor is defined by the operator. Constructor is defined relative to a known
-// [Type][istio.mixer.v1.config.Type]. Their purpose is to tell Mixer how to use attributes or literals to produce
-// instances of the specified type.
+// `Type`. Their purpose is to tell Mixer how to use attributes or literals to produce
+// instances of the specified `Type` at runtime.
 //
-// The following example instructs Mixer to construct an object named 'RequestCountByService' for type 'RequestCounter'
-// and it also provides mapping between fields of the type and attributes/literals.
+// The following example instructs Mixer to construct an object for type 'RequestCounter'
+// and it also provides mapping between fields of the type and attributes/literals. This construtor
+// can be referenced by [Actions][istio.mixer.v1.config.Action] using name 'RequestCountByService'.
 //
 // ```yaml
 // - name: RequestCountByService
@@ -369,16 +374,18 @@ message Action {
 //       target_ip: destination.ip
 // ```
 message Constructor {
-  // Require. Name of the constructor
+  // Required. Name of the constructor
   //
   // Must be unique in the entire mixer configuration. Used by [Action][istio.mixer.v1.config.Action] to refer
   // to this constructor.
   string name = 1;
 
-  // Require. The name of the type this constructor intend to construct.
+  // Required. The value must match the `name` of one of the [Types][istio.mixer.v1.config.Type.name] in scope.
+  // It points to the `Type` this constructor intend to construct.
   string type = 2;
 
-  // Required. Struct representation of a proto associated with the 'type'.
+  // Required. Depends on referenced `Type`. Struct representation of a
+  // proto defined by the [Type][istio.mixer.v1.config.Type]; this varies depending on the value of field `type`.
   google.protobuf.Struct params = 3;
 }
 
@@ -394,25 +401,27 @@ message Constructor {
 // params:
 // ```
 message Handler {
-  // Required, Must be unique in the entire mixer configuration. Used by [Actions][istio.mixer.v1.config.Action.handler]
+  // Required. Must be unique in the entire mixer configuration. Used by [Actions][istio.mixer.v1.config.Action.handler]
   // to refer to this handler.
   string name = 1;
   // Required. The name of a specific adapter implementation. An adapter's
   // implementation name is typically a constant in its code.
   string adapter = 2;
-  // Optional, depends on adapter implementation. Struct representation of a
-  // proto defined by the implementation; this varies depending on `adapter`.
+  // Optional. Depends on adapter implementation. Struct representation of a
+  // proto defined by the adapter implementation; this varies depending on the value of field `adapter`.
   google.protobuf.Struct params = 3;
 }
 
-// Type describes the shape of the object that can be constructed using the
+// A `Type` describes the shape of an object that can be constructed using a
 // [Constructor][istio.mixer.v1.config.Constructor].
 //
-// In the following example we define a `RequestCounter` type. The type is based on 'MetricTemplate' template.
-// Template is a proto message that describes a general category of data that mixer can
-// pass to various adapters for processing. Template dictates how types associated with it should configure the
-// shape of the object that can be constructed. In this case, MetricTemplate requires all types associated with it to
-// declare a 'value' field with its required ValueType and dimensions field with their name and required ValueType.
+// A `Type` is based on a specific `Template`. A `Template` describes a category of `Type`s. Each of the types
+// conforming to the `Template` describes the shape of an object that can be constructed at runtime.
+//
+// In the following example we define a 'MetricTemplate' template, which requires that any `Type` associated with it to
+// declare ValueType associated with 'value' and 'dimensions' field. The example also declares a 'RequestCounter'
+// `Type` which is based on `Template` 'MetricTemplate'. 'RequestCounter' expresses that 'value' field be of type
+// 'INT64' and 'source' dimension be of type STRING.
 //
 // ```yaml
 // - name: RequestCounter
@@ -433,13 +442,14 @@ message Handler {
 // }
 // ```
 message Type {
-  // Required, Name of the type. Must be unique in the entire mixer configuration. Used by
+  // Required. Name of the type. Must be unique in the entire mixer configuration. Used by
   // [Constructor][istio.mixer.v1.config.Constructor.type] which tells the mixer how to construct this type.
   string name = 1;
-  // Required. The name of a specific template this type belongs to.
+  // Required. The name of a specific template this type belongs to. The value must match the name of the available
+  // template in scope.
   string template = 2;
 
-  // Required, depends on template referenced by this type. Struct representation of a
-  // proto defined for the template; this varies depending on `template`.
+  // Optional. Depends on referenced `template`. Struct representation of a
+  // proto defined by the Template; this varies depending on the value of field `template`.
   google.protobuf.Struct params = 3;
 }
--- a/mixer/v1/config/cfg.proto
+++ b/mixer/v1/config/cfg.proto
@@ -197,7 +197,7 @@ message GlobalConfig {
   repeated Handler handlers = 9;
 
   // Optional. List of types that can be referenced from [Constructors][istio.mixer.v1.config.Constructor.type]
-  repeated Types types = 10;
+  repeated Type types = 10;
 }
 
 // AttributeManifest describes a set of Attributes produced by some component
@@ -439,7 +439,7 @@ message Type {
   // Required. The name of a specific template this type belongs to.
   string template = 2;
 
-  // Optional, depends on template referenced by this type. Struct representation of a
+  // Required, depends on template referenced by this type. Struct representation of a
   // proto defined for the template; this varies depending on `template`.
   google.protobuf.Struct params = 3;
 }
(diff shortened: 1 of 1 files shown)


mixer/v1/config/cfg.proto
```diff
@@ -33,17 +33,34 @@ package istio.mixer.v1.config;
 // ```yaml
 // subject: "namespace:ns1"
 // revision: "1011"
-// rules:
+// action_rules:
```
kyessenov: yaml convention is to use camelCase, e.g. `actionRules`

mixer/v1/config/cfg.proto
```diff
@@ -33,17 +33,34 @@ package istio.mixer.v1.config;
 // ```yaml
 // subject: "namespace:ns1"
```
kyessenov: I believe `subject` and `revision` are no longer part of the config, right?

mixer/v1/config/cfg.proto
```diff
@@ -33,17 +33,34 @@ package istio.mixer.v1.config;
...
+//   params:
+//     value: 1
+//     dimensions:
+//       source: origin.service
+//       target_ip: destination.ip
+//
+// types:
+// - name: RequestCounter
+//   template: MetricTemplate
+//   params:
```
kyessenov: This feels very awkward, almost like proto AST written in proto. Can the system accept `MetricTemplate.proto` that just defines value and dimensions as an actual proto?

mixer/v1/config/cfg.proto
```diff
@@ -55,7 +72,16 @@ message ServiceConfig {
   string subject = 1;
   // Optional. revision of this config. This is assigned by the server
   string revision = 2;
+
+  // This is about to get deprecated. The configuration will move to
+  // [action_rules][istio.mixer.v1.config.ServiceConfig.action_rules].
   repeated AspectRule rules = 3;
```
kyessenov: Shouldn't this be the other way, each field below has a `service` field?

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+// ```yaml
+// - handler: prometheus-handler
+//   instances:
+//   - RequestCountByService
+// ```
+message Action {
+  // Required: The handler to invoke.
+  string handler = 2;
+
+  // Required: List of constructor names that will be evaluated during request execution, and the constructed object
```
kyessenov: Please rephrase. Names are resolved, selectors are evaluated, requests follow a data path.

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+// ```
+message Constructor {
+  // Require. Name of the constructor
+  //
+  // Must be unique in the entire mixer configuration. Used by [Action][istio.mixer.v1.config.Action] to refer
+  // to this constructor.
+  string name = 1;
+
+  // Require. The name of the type this constructor intend to construct.
+  string type = 2;
```
kyessenov: There are lot of uninterpreted strings around. Maybe some should be actual types that unions of enums and custom types?

kyessenov:
I'd hold on to merging this until Mixer code changes. There are other non-Mixer APIs that might require changes.

mixer/v1/config/cfg.proto
guptasu: can you give an example ?

mixer/v1/config/cfg.proto
guptasu: Unclear what you mean by 'each field below has a service field'

mixer/v1/config/cfg.proto
guptasu: We already have that, MetricTemplate is equivalent to [MetricDescriptor](https://github.com/istio/api/blob/master/mixer/v1/config/descriptor/metric_descriptor.proto#L56)

mixer/v1/config/cfg.proto
guptasu: will fix thanks.

mixer/v1/config/cfg.proto
guptasu: I can remove it, if we no longer want it. Proto has is, can mark it as deprecated too.

ZackButcher:
Please hold off on merging this till I'm back in the office and can regenerate docs. We don't want to launch and have all our config docs be out of date.

guptasu:
Just FYI, this change would get into unary branch. Eventually will get to
master through unary branch.

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+// In the following example we define a `metrics` adapter using the Mixer's prepackaged
+// prometheus adapter. This adapter doesn't require any parameters.
+//
+// ```yaml
+// name: prometheus-handler
+// adapter: prometheus
+// params:
+// ```
+message Handler {
+  // Required, Must be unique in the entire mixer configuration. Used by [Actions][istio.mixer.v1.config.Action.handler]
```
ZackButcher: We have a mixer of "Required, ..." and "Required. ..." we should pick one version and stick with it. I happen to prefer "Required. " but I don't really care which we pick.

mixer/v1/config/cfg.proto
douglas-reid: @kyessenov this is a bit awkward, but no less awkward than the current state of mixer, for better or worse. As @guptasu points out, we currently have a meta type of MetricDescriptor (a metric has labels that have ValueTypes) and then we have the actual instances of the descriptor (label foo is a string) and then we have the actual attribute binding (label foo is taken from attribute bar).

Is your suggestion to skip the meta type and just use a proto like:
```
message RequestCountMetric {
  int64 value;
  string foo;
  string bar;
  ...
}
```

or similar?

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+// - name: RequestCountByService
+//   type: RequestCounter
+//   params:
+//     value: 1
+//     dimensions:
+//       source: origin.service
+//       target_ip: destination.ip
+// ```
+message Constructor {
+  // Require. Name of the constructor
```
douglas-reid: instead of "Require." do we want "Required." here and elsewhere?

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+  string name = 1;
+  // Required. The name of a specific adapter implementation. An adapter's
+  // implementation name is typically a constant in its code.
+  string adapter = 2;
+  // Optional, depends on adapter implementation. Struct representation of a
+  // proto defined by the implementation; this varies depending on `adapter`.
+  google.protobuf.Struct params = 3;
+}
+
+// Type describes the shape of the object that can be constructed using the
```
ZackButcher: I think
> A `Type` describes the shape of an object that can be constructed using a `Constructor`.

flows a bit better. It's not clear which object or constructor you're referring to when you say "the object" and "the Constructor". Or specify that you mean the constructor whose `type` field matches the name of this `Type` (but it's pretty hard to say this succinctly).

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+// [Constructor][istio.mixer.v1.config.Constructor].
+//
+// In the following example we define a `RequestCounter` type. The type is based on 'MetricTemplate' template.
+// Template is a proto message that describes a general category of data that mixer can
+// pass to various adapters for processing. Template dictates how types associated with it should configure the
```
ZackButcher: > Template dictates how types associated with it should configure the shape of the object that can be constructed.

Like above, it's not entirely clear which shape and object are being referenced. You mean the Template referenced by name in the Type's `template` field. Unfortunately, like above, that's a pretty unwieldy phrase.

How about something like:
> A Template describes a set of types. Each of the types conforming to the Template describes the shape of an object that can be constructed at runtime.

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+//   ValueType value;
+//   map<string, ValueType> dimensions;
+// }
+// ```
+message Type {
+  // Required, Name of the type. Must be unique in the entire mixer configuration. Used by
+  // [Constructor][istio.mixer.v1.config.Constructor.type] which tells the mixer how to construct this type.
+  string name = 1;
+  // Required. The name of a specific template this type belongs to.
+  string template = 2;
```
ZackButcher: Same as above, make it clear this refers to the `name` field of a template.

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+
+  // Required, depends on template referenced by this type. Struct representation of a
+  // proto defined for the template; this varies depending on `template`.
```
ZackButcher: > Struct representation of a proto defined for this `Template`; the value required depends on [`template`][istio.mixer.v1.config.Template.template].

ZackButcher: We should also make sure that the comments on all the `params` fields use a parallel comment structure (meaning they're all phrased the same way, pointing out that the exact value depends on a proto defined by the thing we're referencing by name).

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+  // Required: List of constructor names that will be evaluated during request execution, and the constructed object
+  // will be passed to the handler for processing.
+  repeated string instances = 3;
+}
+
+// Constructor tell Mixer how to create instances of particular types.
+//
+// Constructor is defined by the operator. Constructor is defined relative to a known
+// [Type][istio.mixer.v1.config.Type]. Their purpose is to tell Mixer how to use attributes or literals to produce
+// instances of the specified type.
```
ZackButcher: > instances of the specified type **at runtime**.

mixer/v1/config/cfg.proto
```diff
@@ -55,7 +72,16 @@ message ServiceConfig {
   string subject = 1;
   // Optional. revision of this config. This is assigned by the server
   string revision = 2;
+
+  // This is about to get deprecated. The configuration will move to
```
ayj: `ServiceConfig` is already deprecated. Any reason it can't be removed altogether at this point to avoid depreciating fields in deprecated protobufs?

ZackButcher: +1, we can just delete it since this is a breaking change anyway.

ayj:
As @kuat said, there are likely other non-mixer API changes to be made _soon-ish_. Should all these major/breaking changes be maintained in a separate staging branch until we're satisfied with the overall refactoring/redesign? Or is it sufficient to merge to master (post-alpha release) and defer updating mixer/manager/etc commit references to istio/api?

mixer/v1/config/cfg.proto
```diff
@@ -33,17 +33,34 @@ package istio.mixer.v1.config;
...
-//       metrics: # defines metric collection across the board.
-//       - descriptorName: response_time_by_status_code
-//         value: response.time
-//         labels:
-//           statusCode: response.code
-// ```
+//   actions:
+//   - handler: prometheus-handler
+//     instances:
+//     - RequestCountByService
```
douglas-reid: fwiw, i still believe that ultimately, we will want this to be inlined. i'm not sure that we will need a reference to `RequestCountByService` outside of the scope for which it is useful.

Also, I'd also like to see us try to include namespaces in our examples, if possible.  Would this be `global/RequestCountByService` ?

mixer/v1/config/cfg.proto
kyessenov: Yes, precisely. You want a schema that defines "the shape of data", why isn't a proto good enough for you? You get strong typing and basic validation.

mixer/v1/config/cfg.proto
kyessenov: I am suggesting make each rule have a field called "service" instead of "service" holding many rules.

mixer/v1/config/cfg.proto
kyessenov: Yes, these referential constraints are easier to solve with enums. You probably have <10 types, and bake them in + add custom type.

mixer/v1/config/cfg.proto
kyessenov: I find that last sentence very difficult to parse.

mixer/v1/config/cfg.proto
douglas-reid: I think the general concern with having a `RequestCountMetric` proto is that you want an adapter to be able to handle an array of `*Metric`s. Having an abstraction layer above helps with making those calls, etc..

I'm not sure the approach proposed here differs much from, say, the prometheus model (https://github.com/prometheus/client_model/blob/master/metrics.proto), excepting that Prometheus makes all `labels` (or `dimensions` here) of type `string` (this simplifies the modeling a bit).

Consider: Prometheus' metric definition proto (~ Mixer `Template`) is distinct from both the variable declarations (~ Mixer `Type`) and the method calls that create values (~ Mixer `Constructor`).

To illustrate, in proto land, they have

```
...
message LabelPair {
  optional string name  = 1;
  optional string value = 2;
}
...
message Metric {
  repeated LabelPair label        = 1;
  optional Gauge     gauge        = 2;
  optional Counter   counter      = 3;
  optional Summary   summary      = 4;
  optional Untyped   untyped      = 5;
  optional Histogram histogram    = 7;
  optional int64     timestamp_ms = 6;
}
...
```

And when used in code:

```
var (
	rpcDurations = prometheus.NewSummaryVec(
		prometheus.SummaryOpts{
			Name:       "rpc_durations_seconds",
			Help:       "RPC latency distributions.",
			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
		},
		[]string{"service"},
	)
)
...
rpcDurations.WithLabelValues("uniform").Observe(v)
...
``` 

Mixer is trying to represent all three aspects of this system all in config. This makes this a bit awkward, without doubt.

mixer/v1/config/cfg.proto
ZackButcher: An adapter declares a `Template`, e.g. the `MetricsDescriptor`*. The `MetricsDescriptor` describes the class of valid inputs for an adapter that processes Metrics. The `MetricsDescriptor` allows for a lot of different `Type`s of metrics to be declared, though: RequestCount is a type of metric, and there can be many instances of that type of metric at runtime; similarly RequestLatency is a _different_ type of metric than RequestCount, and there can be many instances of that type of metric at runtime too. Both of these `Type`s has a unique, user defined name (RequestCount, RequestLatency). It is that unique user defined name that is being referenced here by the `type` string, _not_ the `Template` named `MetricDescriptor`.

If we were referring to a `Template` in this field I'd agree an enum could be a good solution. However, under the new mixer config model we allow adapter authors to define new Templates at will, so having a closed set (an enum) of templates that can be referenced in config is no good either (which is why `Type` refers to a `Template` by name, just like `Constructor` refers to its `Type` by name).


\* `MetricDescriptor` will be renamed to `MetricTemplate` soon.

mixer/v1/config/cfg.proto
guptasu: action_rules because we need selectors to be part of ActionRule and we can have 'actions' inside the ActionRule which would represent the true action expression the handler and the constructors.

mixer/v1/config/cfg.proto
kyessenov: I have to learn more about prometheus design. It seems like we want more complex metrics (with types) but still subscribe to dynamically typed labels in prometheus.

mixer/v1/config/cfg.proto
kyessenov: In lieu of the config simplification, we should target each particular config to be less than <50 lines of YAML. A config should be a minimal functional unit of Mixer. @mandarjog @geeknoid

kyessenov:
Feel free to merge since it's for non-master branch.

mixer/v1/config/cfg.proto
guptasu: Another argument against use of 'RequestCountMetric' proto is it needs to not just express types but also concrete values. Example:
message MetricTemplate {
  string name;
  ValueType value;
  map<string, ValueType> dimensions;
  MetricKind kind;
}

enum MetricKind {
DELTA,
GAUGE
}

There is no way to express DELTA or GAUGE if we represent RequestMetricCount as a proto message 
message RequestCountMetric {
  int64 value;
  ????  kind
}

mixer/v1/config/cfg.proto
guptasu: The service configuration is kind of a tree structure where Service configuration is the root.  Rules are meant to be child of their parent service configuration. Service configuration being primary key for this case provides isolation between multiple services.

mixer/v1/config/cfg.proto
guptasu: DONE

mixer/v1/config/cfg.proto
guptasu: Will leave it untouched for this PR since it is unrelated and I am not 100% sure on the story behind the subject and revision.

mixer/v1/config/cfg.proto
guptasu: As Zack mentioned, enums won't work here since all of these concepts (`Type`, `Constructor`, `*Template`) are defined by users. In theory except for *Template, user should be allowed to add remove `Type`s and `Constructor`s without requiring to recompile mixer.

mixer/v1/config/cfg.proto
guptasu: Done

mixer/v1/config/cfg.proto
guptasu: Deleting would result into cascading changes.. current mixer:unary repo depend on this unary repo and might also depend on this field. I would suggest until we delete all the code/tests depending on all of the old behaviors, it will be a real pain to delete individual pieces from the old model. Thoughts ?

mixer/v1/config/cfg.proto
```diff
@@ -55,7 +72,16 @@ message ServiceConfig {
...
+
+  // This is about to get deprecated. The configuration will move to
+  // [action_rules][istio.mixer.v1.config.ServiceConfig.action_rules].
   repeated AspectRule rules = 3;
+
+  // Optional. List of constructors that can be referenced from [actions][istio.mixer.v1.config.Action.instances]
+  repeated Constructor constructors = 4;
+
+  // Optional. List of actions that apply for this service.
+  repeated ActionRule action_rules = 5;
```
guptasu: Correct. As mentioned above, deleting in this PR is going to be tricky because of cascading affects in the mixer repo. Added a TODO to do the change in the followup PRs when all the existing tests have started using the new model. Renaming later will be tedious but easy.

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+//
+// ```yaml
+// action_rules:
+// - selector: target.service == "*"
+//   actions:
+//   - handler: prometheus-handler
+//     instances:
+//     - RequestCountByService
+// ```
+message ActionRule {
```
guptasu: Done

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+// The following example instructs Mixer to invoke 'prometheus-handler' handler and pass it the object
+// constructed using the constructor 'RequestCountByService'
+//
+// ```yaml
+// - handler: prometheus-handler
+//   instances:
+//   - RequestCountByService
+// ```
+message Action {
+  // Required: The handler to invoke.
```
guptasu: Done.

mixer/v1/config/cfg.proto
guptasu: Done

mixer/v1/config/cfg.proto
```diff
@@ -260,3 +293,153 @@ message DnsName {
...
+  // Require. Name of the constructor
+  //
+  // Must be unique in the entire mixer configuration. Used by [Action][istio.mixer.v1.config.Action] to refer
+  // to this constructor.
+  string name = 1;
+
+  // Require. The name of the type this constructor intend to construct.
+  string type = 2;
+
+  // Required. Struct representation of a proto associated with the 'type'.
```
guptasu: Done

mixer/v1/config/cfg.proto
guptasu: Done. Good eye

mixer/v1/config/cfg.proto
kyessenov: We can use annotations and a proto plugin or a descriptor visitor to analyze the proto. The tooling is already there. I'm just wondering if it's the right choice from the extensibility and end-user point of view.

mixer/v1/config/cfg.proto
guptasu: Rephrased

mixer/v1/config/cfg.proto
guptasu: Templates will not have name field, we will need to reflect on all templates that whatever the message name is will become the name of the template. Type will refer to those message names. Better ideas ?

mixer/v1/config/cfg.proto
guptasu: Will we have anything like istio.mixer.v1.config.Template.template ? I think templates are free flowing and we will not have any meta template type. We just know that any field that has ValueType are dynamic fields everything else is allowed. Thoughts ?

guptasu: made parallel construction.

mixer/v1/config/cfg.proto
```diff
@@ -33,17 +33,34 @@ package istio.mixer.v1.config;
...
-// ```
+//   actions:
+//   - handler: prometheus-handler
+//     instances:
+//     - RequestCountByService
+//
+// handlers:
+// - name: prometheus-handler
+//   adapter: prometheus
+//   params:
```
guptasu: action_rules because we need selectors to be part of ActionRule and we can have 'actions' inside the ActionRule which would represent the true action expression the handler and the constructors.

if we inline params for each Adapter, that would mean we will have to define multiple different proto messages for each adapter. Right ? We might be able to do it via codegen, but eventually somewhere in the code we will have to make this generic for shared code. Also, if we make things inline and create multiple messages, users will no longer be able to write a generic "handlers" block in yaml
handlers:
- ...
  ...
- ...
  ...

He will have to do something like
gcphandler:
  url: 
  timeout:

Unless we want to do our own yaml to proto conversion (no auto parsing of yaml into service config proto), then we can let users write however we want and we validate and fill our service config proto ourselves. May be an in person chat on this will be a good idea.

mixer/v1/config/cfg.proto
douglas-reid: @kyessenov we are trying to put together a few examples of all the moving pieces now -- both to sanity check our own assumptions and, hopefully, as a way to illustrate how the components play well together. Stay tuned, as we will definitely want your feedback.

PR #115, opened 2017-06-06
Rename DestinationPolicy to DestinationRule

kyessenov:


rshriram:
What about the Docs? :( .. I was hoping that we can enforce docs changes to go with API changes. If this is blocking, can you atleast create a PENDING_DOC_CHANGES.md file where you add a line stating DestinationPolicies needs to be replaced with DestinationRules?

kyessenov:
@ZackButcher did the doc gen tool got open sourced?

rshriram:
Yes, but its super insufficient. I had to fix a bunch of things by hand, in order for the text to make sense. this is related to our docs versioning (0.1.6 vs whatever).

kyessenov:
Alpha should have no guarantees on API stability. We have a lot of problems in API surface and freezing it at this stage is counter-productive.

rshriram:
This isn't freezing API. This is renaming bunch of fields that require widespread docs changes, and examples changes..

PR #116, opened 2017-06-06
Add name field to a rule

kyessenov:
This is necessary to make a route rule a self-contained object.

### DIFF
--- a/proxy/v1/config/route_rule.proto
+++ b/proxy/v1/config/route_rule.proto
@@ -72,6 +72,10 @@ package istio.proxy.v1.config;
 //       weight: 100
 //
 message RouteRule {
+  // REQUIRED: Route rules have unique names to allow multiple rules
+  // for the same destination, e.g. "my-rule".
+  string name = 11;
+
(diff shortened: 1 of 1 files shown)


proxy/v1/config/route_rule.proto
```diff
@@ -72,6 +72,10 @@ package istio.proxy.v1.config;
 //       weight: 100
 //
 message RouteRule {
+  // REQUIRED: Route rules have unique names to allow multiple rules
+  // for the same destination, e.g. "my-rule".
+  string name = 11;
```
mandarjog: You were against encoding the name inside the proto.
now it is double encoded `/route-rule/{destination}/{rule-name}`

kyessenov: The machinery to embed metadata fields into protos is difficult. There are no traits in proto lang or intefaces AFAIK. Also, not all config types have names.

kyessenov:
PTAL

kyessenov:
Destination policy does not need a name, so no changes there.

PR #118, opened 2017-06-07
add global_word_number field.

qiwzhang:


qiwzhang:
@geeknoid could you review it?

mixer/v1/attributes.proto
```diff
@@ -76,6 +76,13 @@ message Attributes {
   map<sint32, google.protobuf.Duration> durations = 7 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
   map<sint32, bytes> bytes = 8;
   map<sint32, StringMap> string_maps = 9 [(gogoproto.nullable) = false];
+
+  // The number of words in the global dictionary.
+  // Used to detect global dictionary out of sync between client and server.
+  // client sets the number of words in its global dictionary in this field.
+  // server compares with its global dictionary, if server is smaller, it
+  // should fail the request.
+  int global_word_number = 10;
```
geeknoid: I think this field should be in the XXXRequest messages, rather than in Attributes. Otherwise, it ends up being repeated N times in batch Report requests. And the value only makes sense from proxy to mixer, not from mixer to proxy.

Finally, I think it should be global_word_count instead of global_word_number.

qiwzhang:
Added global_word_count to all 3 requests.

qiwzhang:
@geeknoid could you submit it for me?  I don't have permission

PR #119, opened 2017-06-07
Global dictionary in yaml

qiwzhang:
I think it is better to check the global dictionary file here.
If listing them in a doc, you have to manually copy them from doc to code. 

By storing here, we will:
1) have version control, who/when the file is updated
2) can use script to convert this file to your code, better than manually copy

### DIFF
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -21,7 +21,7 @@
 # The main reason using yaml format is that it allows comments like this.
 #
 # Important update rules:
-# * never remove, only add new words.
+# * never remove, only add words at the end of file.
 # * http headers should be lower case.
 # * not duplication, all words have to be unique.
 #
@@ -77,6 +77,12 @@
 - :path
 - :scheme
 - :status
+- access-control-allow-origin
+- access-control-allow-methods
+- access-control-allow-headers
+- access-control-max-age
+- access-control-request-method
+- access-control-request-headers
 - accept-charset
 - accept-encoding
 - accept-language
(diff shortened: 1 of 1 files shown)


qiwzhang:
@geeknoid 

Could you review this? if good, submit it.

mixer/v1/global_dictionary.yaml
```diff
@@ -0,0 +1,133 @@
...
+################################################################################
+#
+#
+# This is the global dictionary used by mixer client and server
+# for sending mixer attributes.
+#
+# The main reason using yaml format is that it allows comments like this.
+#
+# Important update rules:
+# * never remove, only add new words.
```
qiwzhang: Done

mixer/v1/global_dictionary.yaml
```diff
@@ -0,0 +1,133 @@
...
+- vary
+- via
+- www-authenticate
+
+# HTTP methods
+- GET
+- POST
+- PUT
+- PATCH
+- DELETE
```
qiwzhang: Done

mixer/v1/global_dictionary.yaml
```diff
@@ -0,0 +1,160 @@
...
+- www-authenticate
+
+# HTTP methods
+- GET
+- POST
+
+# Common strings
+- http
+- envoy
+- '200'
```
geeknoid: Do we really expect strings like that with the single quotes?

qiwzhang: No, we don't.   This is for python yaml parser.   It will parse into 200 as number without quote. 
It may break the tool to generate code.

qiwzhang: python yaml parser will convert it into a 200 string without quote

PR #120, opened 2017-06-08
Introduce ingress rule

kyessenov:
This definition captures the effective use of the route rules in the ingress controller in Pilot. The route rules fields that are missing are simply ignored by the implementation. Hoping to get this merged so we can transition to one-proto-per-type in the config system.

### DIFF
--- a/proxy/v1/config/ingress_rule.proto
+++ b/proxy/v1/config/ingress_rule.proto
@@ -1,0 +1,66 @@
+// Copyright 2017 Istio Authors
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "proxy/v1/config/route_rule.proto";
+
+package istio.proxy.v1.config;
+
+// Ingress rules are routing rules applied to the mediating middle proxy. This
+// proxy serves as an ingress proxy for the entire mesh, but can also be
+// addressed from inside the mesh.  Each ingress rule defines a destination
+// service and port. Rules that do not resolve to a service or a port in the
+// mesh are ignored.
+//
+// The routing rules for the destination service are applied at the ingress
+// proxy. That means the routing rule match conditions are composed and its
+// actions are enforced. The traffic splitting for the destination service is
+// also effective.
+message IngressRule {
+  // REQUIRED: Ingress rules have unique names to allow multiple rules, e.g.
+  // "my-rule".
+  string name = 1;
+
+  // REQUIRED: Port on which the ingress proxy listens and applies the rule.
+  int32 port = 2;
+
+  // Optional TLS secret path to apply server-side TLS context on the port.
+  string tls_secret = 3;
+
+  // RECOMMENDED. Precedence is used to disambiguate the order of application
+  // of rules. A higher number takes priority. If not specified, the value is
+  // assumed to be 0.  The order of application for rules with the same
+  // precedence is unspecified.
+  int32 precedence = 4;
+
+  // Match condtions to be satisfied for the ingress rule to be
+  // activated.
+  MatchCondition match = 5;
+
+  // REQUIRED: Destination uniquely identifies the destination service. Value
+  // must be in fully qualified domain name format (e.g.,
+  // "my-service.default.svc.cluster.local").
+  string destination = 6;
+
+  // REQUIRED: Destination port identifies a port on the destination service for routing.
+  oneof destination_service_port {
+    // Identifies the destination service port by value
+    int32 destination_port = 7;
+
+    // Identifies the destination service port by name
+    string destination_port_name = 8;
+  }
+}
+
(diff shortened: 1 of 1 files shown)


proxy/v1/config/ingress_rule.proto
```diff
@@ -0,0 +1,66 @@
...
+
+import "proxy/v1/config/route_rule.proto";
+
+package istio.proxy.v1.config;
+
+// Ingress rules are routing rules applied to the mediating middle proxy. This
+// proxy serves as an ingress proxy for the entire mesh, but can also be
+// addressed from inside the mesh.  Each ingress rule defines a destination
+// service and port. Rules that do not resolve to a service or a port in the
+// mesh are ignored.
```
douglas-reid: should we change "are ignored" to "should be ignored"?

proxy/v1/config/ingress_rule.proto
```diff
@@ -0,0 +1,66 @@
...
+  // Optional TLS secret path to apply server-side TLS context on the port.
+  string tls_secret = 3;
+
+  // RECOMMENDED. Precedence is used to disambiguate the order of application
+  // of rules. A higher number takes priority. If not specified, the value is
+  // assumed to be 0.  The order of application for rules with the same
+  // precedence is unspecified.
+  int32 precedence = 4;
+
+  // Match condtions to be satisfied for the ingress rule to be
```
douglas-reid: s/condtions/conditions/

proxy/v1/config/ingress_rule.proto
```diff
@@ -0,0 +1,66 @@
...
+message IngressRule {
+  // REQUIRED: Ingress rules have unique names to allow multiple rules, e.g.
+  // "my-rule".
+  string name = 1;
+
+  // REQUIRED: Port on which the ingress proxy listens and applies the rule.
+  int32 port = 2;
+
+  // Optional TLS secret path to apply server-side TLS context on the port.
+  string tls_secret = 3;
```
ayj: Does *path* here mean filesystem path or reference to a secret?

proxy/v1/config/ingress_rule.proto
```diff
@@ -0,0 +1,66 @@
...
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "proxy/v1/config/route_rule.proto";
+
+package istio.proxy.v1.config;
+
+// Ingress rules are routing rules applied to the mediating middle proxy. This
```
rshriram: I would say, "applied to the Pool of frontend load balancers that receive traffic for the entire mesh. The load balancer can also be addressed directly from within the cluster if needed".

proxy/v1/config/ingress_rule.proto
```diff
@@ -0,0 +1,66 @@
...
+// proxy serves as an ingress proxy for the entire mesh, but can also be
+// addressed from inside the mesh.  Each ingress rule defines a destination
+// service and port. Rules that do not resolve to a service or a port in the
+// mesh are ignored.
+//
+// The routing rules for the destination service are applied at the ingress
+// proxy. That means the routing rule match conditions are composed and its
+// actions are enforced. The traffic splitting for the destination service is
+// also effective.
```
rshriram: We need an example of IngressRule here. PTAL at the RouteRule for reference.

proxy/v1/config/ingress_rule.proto
rshriram: We also need to provide justification for why Ingress rules are different from normal route rules. And how they relate to ingress rules in K8S. I don't think Mesos or Swarm or CF has any notion of Ingress Rule.

proxy/v1/config/ingress_rule.proto
rshriram: This is very vague. For a generic ingress rule, TLS secret would have a lot of things, such as CA certs, private keys, etc etc. The above case seems to imply K8S secrets, in which case, this ingress becomes specific to K8S.

proxy/v1/config/ingress_rule.proto
```diff
@@ -0,0 +1,66 @@
...
+  // "my-service.default.svc.cluster.local").
+  string destination = 6;
+
+  // REQUIRED: Destination port identifies a port on the destination service for routing.
+  oneof destination_service_port {
+    // Identifies the destination service port by value
+    int32 destination_port = 7;
+
+    // Identifies the destination service port by name
+    string destination_port_name = 8;
```
rshriram: Why do we need to worry about port_name here? This will be cause for confusion.

proxy/v1/config/ingress_rule.proto
kyessenov: done

proxy/v1/config/ingress_rule.proto
kyessenov: done

proxy/v1/config/ingress_rule.proto
kyessenov: it's meant to be vague since Istio doesn't have secret store API. I'm not going to design an API in this tiny commit.

proxy/v1/config/ingress_rule.proto
kyessenov: This is not even alpha. Once it stabilizes, we'll get examples.

proxy/v1/config/ingress_rule.proto
kyessenov: Because you do port translation here unlike route rule.

kyessenov: In k8s, ports are can be called by name.

proxy/v1/config/ingress_rule.proto
kyessenov: Done regarding the pool specifically for ingress.

kyessenov:
Placed a warning that this API is experimental.

PR #122, opened 2017-06-12
Define the Galley API

mandarjog:
This PR proposes an abstract config API service loosely based on the Kubernetes API server.
It adds a streaming watch to satisfy istio use cases efficiently, Stream watcher API follows etcdv3 watch API.


mandarjog:
@lavalamp Can you please take a look at this?

config/service.proto
```diff
@@ -0,0 +1,257 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.config;
```
ZackButcher: If all the api methods are `v1/...` then this package should be `istio.config.v1` too (so the gRPC service is hosted as `istio.config.v1.IstioConfig`, rather than `istio.config.IstioConfig`. On that note, the service name should be changed, making comment below.

Including the version in the proto package should save us pain on the gRPC side later.

config/service.proto
```diff
@@ -0,0 +1,257 @@
...
+
+// The Istio Config API service follows
+// the kubenertes API server delegation and resource model.
+// /apiGroup/version/objecttype/{namespace}/objectname  -->
+//
+// It generalizes the concept of namespace to an arbitrary grouping.
+//
+// /apiGroup/version/objecttype/{objectGroup}/objectname
+// + etcdv3 stream watch API
+service IstioConfig {
```
ZackButcher: `IstioConfig` -> `Config`. We're already in the package `istio.config`, the repeated info is redundant. If anything, you might call this `Service` so it's addressed as `istio.config.v1.Service`.

config/service.proto
```diff
@@ -0,0 +1,257 @@
...
+        };
+  };
+
+  // Get a list of objects.
+  // other arguments like meta.object_group, meta.name are optional
+  // This lets one list all object of a type, or all object of a type
+  // that is within an object group.
+  rpc ListObjects(ListObjectsRequest) returns (ObjectList) {
+        option (google.api.http) = {
+            get: "/v1/{meta.api_group}/{meta.object_type}" // ?meta.object_group=group1}"
```
ZackButcher: Leftover comment here? `// ?meta.object_group=group1}"`

config/service.proto
```diff
@@ -0,0 +1,257 @@
...
+            additional_bindings: {
+              get: "/v1/{meta.api_group}/{meta.object_type}/meta.object_group}"
```
ZackButcher: Missing a `{` for the `object_group`:

`"/v1/{meta.api_group}/{meta.object_type}/{meta.object_group}"`

config/service.proto
mandarjog: fixed naming issues.

config/service.proto
```diff
@@ -0,0 +1,258 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.v1.config;
```
ZackButcher: Are we sure we want `istio.v1.config` and not `istio.config.v1`? The first implies that this API surface can only change when Istio changes version, while the second implies that the config API can be versioned independently of Istio overall. I think we'll want to freedom to iterate on this API independently of Istio's overall version.

config/service.proto
```diff
@@ -0,0 +1,258 @@
...
+    DELETE = 1;
+  }
+  // type is the kind of event. If type is a PUT, it indicates
+  // new data has been stored to the key. If type is a DELETE,
+  // it indicates the key was deleted.
+  EventType type = 1;
+  // kv holds the KeyValue for the event.
+  // A PUT event contains current kv pair.
+  // A PUT event with kv.Version=1 indicates the creation of a key.
+  // A DELETE/EXPIRE event contains the deleted key with
```
jmuk: `DELETE/EXPIRE` -> `DELETE` only for now? Are we going to support expiration of configs?

config/service.proto
```diff
@@ -0,0 +1,258 @@
...
+    WatchCancelRequest cancel_request = 2;
+  }
+}
+
+message WatchCreateRequest {
+  // used to identify 
+  ObjectMeta key = 1;
+
+  // start watching from this revision.
+  int64 start_revision = 2;
```
jmuk: It seems no revisions are set in any of the responses. What does the revision means here?

mandarjog: versions are available in `ObjectMeta`

jmuk: Is it the revision of the storage? I was thinking it's per-object revision. Could you modify the comment in `ObjectMeta`?

It's probably better to use a single word for a concept, either `version` or `revision`, but not both.

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+  // watch_id is the ID of the watcher that corresponds to the response.
+  int64 watch_id = 1;
+  // created is set to true if the response is for a create watch request.
+  // The client should record the watch_id and expect to receive events for
+  // the created watcher from the same stream.
+  // All events sent to the created watcher will attach with the same watch_id.
+  bool created = 2;
+  // canceled is set to true if the response is for a cancel watch request.
+  // No further events will be sent to the canceled watcher.
+  bool canceled = 3;
```
lavalamp: Are all 4 combinations of created and canceled valid?

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+//
+// It generalizes the concept of namespace to an arbitrary grouping.
+//
+// /apiGroup/version/objecttype/{objectGroup}/objectname
+// + etcdv3 stream watch API
+service Service {
+  // Get a single object.
+  rpc GetObject(GetObjectRequest) returns (Object) {
+        option (google.api.http) = {
+            get: "/v1/{meta.api_group}/{meta.object_type}/{meta.object_group}/{meta.name}"
```
lavalamp: Kubernetes would have group/version, not version/group.

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+    };
+  }
+}
+
+message ObjectFieldSelector {
+  // if true object(s) should include data.
+  bool data = 1;
+
+  // if true object(s) should include source_data.
+  bool source_data = 2;
```
lavalamp: What is the difference between data and source_data? are all 4 combinations of these bools valid?

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+  // source_data is the data as it was specified in json / yaml format.
+  google.protobuf.Struct source_data = 2;
+}
+
+message GetObjectRequest {
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+
+  // select selects fields to include in the result.
+  ObjectFieldSelector select = 2;
```
lavalamp: A Kubernetes field selector is used to find an object(s), not to select which fields to return.

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+  // meta is the metadata associated with this list.
+  Meta meta = 1;
+  
+  // list of objects returned.
+  repeated Meta object_types = 2;
+}
+
+message ObjectList {
+  // meta is the metadata associated with this list.
+  Meta meta = 1;
```
lavalamp: K8s uses separate ObjectMeta and ListMeta types, the latter is much shorter.

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+  // list of objects returned.
+  repeated Object objects = 2;
+} 
+
+
+// Meta
+message Meta {
+  // api_group is the top level delegation.
+  // All messages from an api group are handled by the same delegation handler.
+  string api_group = 1;
```
lavalamp: It is not clear that it's useful to have proto versions of the api_group & api_version concepts, since definitionally clients must have the proto version compiled in to even talk to the endpoint.

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+  // For Mixer configuration object_group == mesh, svc:svc1, svc:svc2 ...
+  // For Proxy it can be mesh, destination. 
+  string object_group = 3;
+
+  // name of the object.
+  string name = 4;
+
+  // version
+  // version of the object.
+  int64 version = 5;
```
lavalamp: What kind of version? resource version, like in k8s? something else? Comment should probably explain.

config/v1/service.proto
```diff
@@ -0,0 +1,283 @@
...
+  // route-rule, metrics-type, metrics-ctor, ...
+  string object_type = 2;
+
+  // objectGroup is an arbitrary hierarchical grouping of resources.
+  // For Mixer configuration object_group == mesh, svc:svc1, svc:svc2 ...
+  // For Proxy it can be mesh, destination. 
+  string object_group = 3;
+
+  // name of the object.
+  string name = 4;
```
lavalamp: uid?

config/service.proto
```diff
@@ -0,0 +1,258 @@
...
+  // Update a single object. May return validation errors.
+  rpc UpdateObject(UpdateObjectRequest) returns (Object) {
+        option (google.api.http) = {
+            put: "/v1/{meta.api_group}/{meta.object_type}/{meta.object_group}/{meta.name}"
+            body: "source_data"
+        };
+  };
+
+  // Delete a single object. May return validation errors.
+  rpc DeleteObject(DeleteObjectRequest) returns (google.protobuf.Empty) {
```
ZackButcher: [API style guide mandates `Delete` methods return `Empty `.](https://cloud.google.com/apis/design/standard_methods#delete)

config/v1/service.proto
mandarjog: ok changed it to match k8s.

config/v1/service.proto
mandarjog: Changed name.

I will leave it out arbitrary query selection for now. This is an optimization.

config/v1/service.proto
mandarjog: added.

config/v1/service.proto
mandarjog: Meta could describe any part of the hierarchy. I would like to keep it this way and then specialized it later.

config/v1/service.proto
mandarjog: created = 1 and canceled = 1 is invalid.  (rest are valid.)

lavalamp: An enum might be better, then.

config/v1/service.proto
lavalamp: I see, interesting. (comments should probably be expanded unless there's API documentation elsewhere?)

config/v1/service.proto
```diff
@@ -0,0 +1,292 @@
...
+
+  // source_data is the data as it was specified in json / yaml format.
+  google.protobuf.Struct source_data = 2;
+
+  // data is the binary encoded protobuf data
+  // Both representations are useful. 
+  // For example:
+  // If `souce_data` contains an expressions in text form 
+  // `data` may contain a parsed AST
+  bytes data = 3;
```
jmuk: If it's assumed to be protobuf-encoding, should this be `google.protobuf.Any`?

config/v1/service.proto
mandarjog: Added comments.

config/v1/service.proto
mandarjog: `google.protobuf.Any` has repercussions for the grpc transcoding proxy. It needs to have access to the underlying protos defined by the internal `type` field of `any`.
That defeats the purpose of having an opaque proto message that only the code that converts it needs to understand.

config/v1/service.proto
mandarjog: Created an enum.

config/v1/service.proto
```diff
@@ -0,0 +1,305 @@
...
+    option (google.api.http) = {
+      get: "/{meta.api_group}/{meta.api_group_version}"
+      additional_bindings: {
+        get: "/"
+      };
+    };
+  };
+
+  // Create an object. This may result in configuration validation errors.
+  // Referrential integrity is not generally guranteed, however individual validators
```
ayj: s/guranteed/guaranteed/

config/v1/service.proto
```diff
@@ -0,0 +1,305 @@
+// Copyright 2017 Istio Authors
+// vim: set expandtab ts=2 sw=2:
```
ayj: Accidentally left in vim modeline?

config/v1/service.proto
```diff
@@ -0,0 +1,305 @@
...
+
+  repeated Event events = 4;
+}
+
+message Event {
+  enum EventType {
+    UPDATE = 0;  // ADD and UPDATE
+    DELETE = 1;
+  }
+  // type is the kind of event. If type is a PUT, it indicates
```
ayj: s/PUT/UPDATE here and lines 297 and 298

config/v1/service.proto
```diff
@@ -0,0 +1,305 @@
...
+  // `include` is a reserved word. so not used here.
+  ObjectFieldInclude incl = 2; 
+}
+
+message UpdateObjectRequest {
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+  
+  // source_data is the data as it was specified in json / yaml format.
+  google.protobuf.Struct source_data = 2;
```
ayj: Would we ever want to allow users to create/update objects using the explicit proto message type vs. Struct to avoid unnecessary conversions?

config/v1/service.proto
ayj: What happens if grpc transcending proxy doesn't have access to underlying protos? Does it fail in a reasonable way (i.e. do nothing) or does it corrupt data, etc?

config/v1/service.proto
```diff
@@ -0,0 +1,305 @@
...
+message WatchCreateRequest {
+  // used to identify 
+  Meta key = 1;
+
+  // start watching from this revision.
+  int64 start_revision = 2; 
+
+  enum FilterType {
+    // filter out put event.
+    NOPUT = 0;
```
ayj: Use UPDATE instead of PUT to be consistent with `EventType` below. Also, can you use CAPITALIZED_NAMES_WITH_UNDERSCORES to be consistent with https://cloud.google.com/apis/design/naming_convention?

config/v1/service.proto
```diff
@@ -0,0 +1,305 @@
...
+  };
+
+  // Get a list of object types.
+  // Lists all object types or all object types 
+  // within an api group.
+  rpc ListObjectTypes(ListObjectTypesRequest) returns (ObjectTypeList) {
+    option (google.api.http) = {
+      get: "/{meta.api_group}/{meta.api_group_version}"
+      additional_bindings: {
+        get: "/"
```
jmuk: I've tried the grpc-gateway and met an error on this line. It seems grpc-gateway assumes the path not ending with `/`.

```
--grpc-gateway_out: segment neither wildcards, literal or variable: expected "{" but got "\x00": /
```

maybe add a special word for this section, like `"/__all__"`?

config/v1/service.proto
```diff
@@ -0,0 +1,305 @@
...
+// /apiGroup/version/objecttype/{objectGroup}/objectname
+// 
+// api_group = /mixer/v1  or
+// api_group = /proxy/v1
+// + etcdv3 stream watch API
+service Service {
+  // Get a single object.
+  rpc GetObject(GetObjectRequest) returns (Object) {
+    option (google.api.http) = {
+      get: "/{meta.api_group}/{meta.api_group_version}/{meta.object_type}/{meta.object_group}/{meta.name}"
```
jmuk: This doesn't allow to specify `incl`, and since booleans are false by defualt, the server can't respond any resources. Possibly you might want to reconsider it from `include` to `exclude`.

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.config.v1;
```
geeknoid: config -> galley

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+service Service {
+  // Get a single object.
+  rpc GetObject(GetObjectRequest) returns (Object) {
+    option (google.api.http) = {
+      get: "/{meta.api_group}/{meta.api_group_version}/{meta.object_type}/{meta.object_group}/{meta.name}"
+    };
+  };
+
+  // Get a list of objects.
+  // other arguments like meta.object_group, meta.name are optional
```
geeknoid: Other

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  // This lets one list all object of a type, or all object of a type
```
geeknoid: object -> objects

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+    option (google.api.http) = {
+      get: "/{meta.api_group}/{meta.api_group_version}"
+      additional_bindings: {
+        get: "/"
+      };
+    };
+  };
+
+  // Create an object. This may result in configuration validation errors.
+  // Referrential integrity is not generally guranteed, however individual validators
```
geeknoid: I think we want a more aggressive stance here. Referential integrity should be maximally enforced, except where not possible.

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+    option (google.api.http) = {
+      delete: "/{meta.api_group}/{meta.api_group_version}/{meta.object_type}/{meta.object_group}/{meta.name}"
+    };
+  };
+};
+
+
+// Watcher service is designed to efficiently watch mutiple subtrees of the resource tree.
+service Watcher { 
+  // Watch method is borrowed from etcdv3 bi-directional streaming watches.
```
geeknoid: Remove this line.

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  // Watch watches for events happening or that have happened. Both input and output
+  // are streams; the input stream is for creating and canceling watchers and the output
```
geeknoid: cancelling

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  Meta meta = 1;
+}
+
+message ListObjectsRequest {
+  // meta is the metadata associated with the root where listing begins.
+  Meta meta = 1;
+
+  // incl(ude) selects fields to include in the result. 
+  // `include` is a reserved word. so not used here.
+  ObjectFieldInclude incl = 2;
```
geeknoid: incl -> included_fields

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  google.protobuf.Struct source_data = 2;
+}
+
+message GetObjectRequest {
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+
+  // incl(ude) selects fields to include in the result.
+  // `include` is a reserved word. so not used here.
+  ObjectFieldInclude incl = 2;
```
geeknoid: incl -> included_fields

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+
+  // source_data is the data as it was specified in json / yaml format.
+  google.protobuf.Struct source_data = 2;
+
+  // data is the binary encoded protobuf data
+  // Both representations are useful. 
+  // For example:
+  // If `souce_data` contains an expressions in text form
```
geeknoid: souce -> source

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  repeated Object objects = 2;
+} 
+
+
+// Meta
+message Meta {
+  // api_group is the top level delegation.
+  // All messages from an api group are handled by the same delegation handler.
+  // api group itself is versioned.
+  string api_group = 1;
```
geeknoid: As mentioned above, we shouldn't be categorizing our config by "api_group". That's in fact an anti-pattern for us.

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  // For Mixer configuration object_group == mesh, svc:svc1, svc:svc2 ...
+  // For Proxy it can be mesh, destination. 
+  string object_group = 4;
+
+  // name of the object.
+  string name = 5;
+
+  // uid assigned to the object. Only used when a new object is created 
+  // in the same place as before.
+  string uid = 6;
```
geeknoid: What's the scope of this ID? Should this in fact be uuid?

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+
+  // revision
```
geeknoid: Why repeat the field name in the comment, here and below?

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  // revision of the repository, the last time Object was updated.
+  int64 revision = 7;
```
geeknoid: What's 'repository'? The whole config store as a whole? Can we expand the word 'revision' to be more descriptive?

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+
+  // labels
+  // labels associated with the object.
+  map<string, string> labels = 8;
+}
+
+// watchers follows etcdv3 watch pattern.
```
geeknoid: Remove reference to external stuff and document the pattern here instead. It's OK to go steal comments from k8s if you need to.

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+message WatchRequest {
+  // request_union is a request to either create a new watcher or cancel an existing watcher.
```
geeknoid: "is a request to either" -> "indicates whether to"

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+  oneof request_union {
```
geeknoid: Why this weird pattern? Shouldn't we just have different methods, one to create a watch and one to delete it, rather than having this union?

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
...
+    WatchCreateRequest create_request = 1;
+    WatchCancelRequest cancel_request = 2;
+  }
+}
+
+message WatchCreateRequest {
+  // used to identify the watched subtree
+  Meta subtree = 1;
+
+  // start watching from this revision.
```
geeknoid: Revision of what? The repository?

geeknoid:
I'm not thrilled with this API. I think we're going too far in being generic, the API provides little semantic clarity over what parameters are expected and how they interact. 

I realize we have some open-ended config stuff to deal with. But at the same time, there are various chunks of well-known config that we will handle (routing rules for example, or all of Mixer's well-known templates). I believe we should have dedicated strongly-typed APIs for those things we know, and only fallback to generic bland API for the open-ended stuff.

ayj:
re: @geeknoid's concerns, it would be good to understand and document the motivation for a more generic API. We can probably live with some churn while the API settles down considering we'll be auto-generating the more tedious boilerplate code.

config/v1/service.proto
mandarjog: done.

config/v1/service.proto
mandarjog: good point, I will add it here.

config/v1/service.proto
mandarjog: done

config/v1/service.proto
mandarjog: done

config/v1/service.proto
mandarjog: it fails hard, if the underlying protos are not available. This establishes an unwanted coupling.

config/v1/service.proto
mandarjog: done.

config/v1/service.proto
mandarjog: // Create and cancel are part of the same message because the WatchRequest message
// is used in a streaming API. The client may add new watchers and remove old watchers
// on an existing stream.

Added comments.

config/v1/service.proto
mandarjog: removed filter types for now.

config/v1/service.proto
mandarjog: yes repository. updated comment.

config/v1/service.proto
mandarjog: done.

config/v1/service.proto
```diff
@@ -0,0 +1,322 @@
+// Copyright 2017 Istio Authors
```
mandarjog: done.

config/v1/service.proto
mandarjog: Left it in while it is being edited.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+message ListObjectTypesRequest {
+  // meta is the metadata associated with the root where listing begins.
+  Meta meta = 1;
+}
+
+message ListObjectsRequest {
+  // meta is the metadata associated with the root where listing begins.
+  Meta meta = 1;
+
+  // incl(ude) selects fields to include in the result.
```
geeknoid: Update comment.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+  // incl(ude) selects fields to include in the result. 
+  // `include` is a reserved word. so not used here.
+  ObjectFieldInclude include_fields= 2; 
+}
+
+message ObjectRequest {
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+
+  // oneof source_data or data
```
geeknoid: Remove this three line comment.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+  // Note: an actual oneof is not used here because
+  // grpc gateway does not handle oneoff in request body
+ 
+  // source_data is the data as it was specified in json / yaml format.
+  google.protobuf.Struct source_data = 2;
+ 
+  // data is protos encoded as bytes only specify
```
geeknoid: Fix comment.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+ 
+  // data is protos encoded as bytes only specify  
+  bytes data = 3; 
+}
+
+message GetObjectRequest {
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+
+  // incl(ude) selects fields to include in the result.
```
geeknoid: Fix comment.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+  // `include` is a reserved word. so not used here.
+  ObjectFieldInclude include_fields= 2; 
+}
+
+message DeleteObjectRequest {
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+}
+
+// Object
```
geeknoid: Say something useful here.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+message Object {
+  // meta is the metadata associated with the object.
+  Meta meta = 1;
+
+  // source_data is the data as it was specified in json / yaml format.
+  google.protobuf.Struct source_data = 2;
+
+  // data is the binary encoded protobuf data
```
geeknoid: The comment here and on ObjectRequest.data probably should match.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+  // list of objects returned.
+  repeated Meta object_types = 2;
+}
+
+message ObjectList {
+  // meta is the metadata associated with this list.
+  Meta meta = 1;
+  
+  // list of objects returned.
+  repeated Object objects = 2;
```
geeknoid: We probably need pagination support for any list API.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+  string uid = 6;
+
+  // revision of the repository, the last time Object was updated.
+  int64 revision = 7;
+
+  // labels associated with the object.
+  map<string, string> labels = 8;
+}
+
+// WatchRequest creates on cancels a watch.
```
geeknoid: Probably worth putting the watcher definitions into their own proto file.

galley/v1/service.proto
```diff
@@ -0,0 +1,311 @@
...
+// Galley follows
+// the Kubenertes API server delegation and resource model.
+// /apiGroup/version/objecttype/{namespace}/objectname  -->
+//
+// It generalizes the concept of namespace to an arbitrary grouping.
+//
+// /apiGroup/version/objecttype/{objectGroup}/objectname
+// 
+// api_group = /mixer/v1  or
+// api_group = /proxy/v1
```
geeknoid: I remain unconvinced that api_group is a good thing. It encourages component-based config state. Having "core" as an api group and putting everything in there doesn't seem useful. Why have "core" then?

geeknoid:
As I mentioned in my previous review, I'm not a fan of this API. This is really just a totally generic "store protos in a hierarchy" API. Is this all we can or should do? Do we expect to create a usability layer on top of this which has stronger typing?

ZackButcher:
> Do we expect to create a usability layer on top of this which has stronger typing?

I think this is the only way this API can work - as I said earlier, I'm not confident I could correctly `curl` it. 

With respect to adding strong types for things like well-known templates in Mixer, I'd really prefer we don't do that. In my opinion we should not special case things just because we happen to ship them instead of some other third party. On the other hand, I want a usable API. But I don't want to have to rebuild everything, even my CLI tools, every time I change my Mixer deployment. It's a hard balance to strike.

Is it an option for us to have the fully generic API for the reasons that've been laid out (mostly to avoid proto dependencies everywhere), but _also_ ship a more strongly typed API? As a strawman, consider generating a strongly typed API server from a set of inputs (the templates from Mixer, whatever is required by Envoy, etc) that acts as a wrapper dispatching to this generic API. Our CLI tools can go directly against the generic API and still provide a good experience, but we can also offer a reasonable experience for operators not using our tooling. Such a wrapper API would likely make building a UI easier as well.

geeknoid:
I'm dubious whether the generic API can lead to a usable CLI tools. It's
likely that the CLI tools would need to be as generic as the API and so let
users add/delete/edit documents in a hierarchy, with no clear semantics on
what these documents are. You'd just have to pass a blob and see if errors
come back.

So what if the model is that we have this generic API and then all the
stuff above this API depends on a plug-in model for usability? So we'd have
istioctl which dynamically loads auto-generated plug-ins based on things
like Mixer templates?

lavalamp:
Take a look at the kubernetes command line tool. It provides CRUD against resources it doesn't have compiled in. (I will leave it to you to judge whether it is "usable" or not :) )

To make this work we are opinionated about the resource data format (must have consistent metadata). A proto-first system like the one in this PR is going to make this hard, because proto is not self-describing. Maybe you can have rules about the tag number of the metadata field. You absolutely must have some way of identifying the type of a blob of data; in kubernetes, our protos have an envelope layer that declares the type and format (i.e., compressed, encrypted, json, proto).

lavalamp:
I am happy to get in a room with all of you and give a brief explanation of how Kubernetes solves these problems if that would be helpful, btw.

mandarjog:
https://docs.google.com/document/d/1I78MyoLwA6lrecPc82i_IzyntUDVA9TIMf2ZuY4Eqzo/edit#

galley/v1/service.proto
```diff
@@ -0,0 +1,371 @@
...
+      body: "*"
+    };
+  };
+};
+
+
+// Object validator service validates objects before they are committed to storage.
+// Galley maintains a map of object_types to validators. A single validator may validate many types.
+// For example MixerValidator Service should validate all Mixer resources.
+service ObjectValidator {
```
guptasu: Can we rename this to ObjectValidatorAndConverter ?

galley/v1/service.proto
```diff
@@ -0,0 +1,371 @@
...
+
+  // uuid assigned to the object. Only used when a new object is created 
+  // in the same place as before.
+  string uuid = 6;
+
+  // revision of the repository, the last time Object was updated.
+  int64 revision = 7;
+
+  // labels associated with the object.
+  map<string, string> labels = 8;
```
guptasu: For the case of mixer, Is this where the templateName will be passed   as a key to the map ?

geeknoid: I believe that labels are entirely owned by the user. The user assigns and manages labels on pieces of state. So we shouldn't be using those for system-level state.

mandarjog: That is correct. System related information is encoded in resource URI.
As we talked about before,  templateName is am implementation detail of how Mixer implements `mesh Functions`
```
/core/metrics/v1/mesh/request_count ==>
templateName = metrics
subject = mesh
constructorName = request_count 
```
labels are owned by operator. I also think labels may be used by UI to record metadata.

galley/v1/service.proto
```diff
@@ -0,0 +1,378 @@
...
+    // Server sends periodic messages of progress when no actual watches fire.
+    // The client knows it is caught up to a certain revision of the repository.
+    WatchProgress progress = 6; 
+  };
+};
+
+
+message Event {
+  enum EventType {
+    PUT = 0;  // ADD and UPDATE
```
ZackButcher: Why do we use HTTP verbs here? If `PUT` represents an `ADD` or an `UPDATE` why don't we just have variants `ADD`, `UPDATE`, `DELETE` - is there a reason to constrain this set to HTTP-like things?

xiaolanz:
When is the message Validation API intended for use? Does it serve as a compromise API usability and generalization?

galley/v1/service.proto
```diff
@@ -0,0 +1,378 @@
...
+//
+// It generalizes the concept of namespace to an arbitrary grouping.
+//
+// /apiGroup//objecttype/version/{objectGroup}/objectname
+// 
+// api_group = /core/
+// api_group = /vendor1
+service Galley {
+  // Get a single object.
+  rpc GetObject(GetObjectRequest) returns (Object) {
```
xiaolanz: "Object" is a very generic word? Can we change to Config?

galley/v1/service.proto
```diff
@@ -0,0 +1,378 @@
...
+  repeated Object objects = 2;
+
+  // If next_page_token is not empty, this is a paged result.
+  // use this value in the next request.
+  string next_page_token = 3;
+} 
+
+
+// Meta
+message Meta {
```
xiaolanz: Does message Meta effectively serve as "key" for object? It feels more readable to just call it Key.

galley/v1/service.proto
xiaolanz: Do you allow a client sends in a any number of keys which does not exist yet?

galley/v1/service.proto
```diff
@@ -0,0 +1,378 @@
...
+
+message ListObjectsRequest {
+  // meta is the metadata associated with the root where listing begins.
+  Meta meta = 1;
+
+  // paged result, set to empty on first page.
+  string page_token = 2;
+}
+
+message ObjectRequest {
```
xiaolanz: should ObjectRequest includes a message Object instead of duplicating the fields?

galley/v1/service.proto
```diff
@@ -0,0 +1,378 @@
...
+  // Mixer validator should check 
+  // 1. expression syntax is valid
+  // 2. expression uses known attributes
+  // 3. Rule refers to a known handler
+  // 4. Any other semantic check that is required for full validation.
+  //
+  // It should convert untyped proto into typed proto and return binary encoding of it in Object.data.
+  // 
+  // On validation failure, it should return a validation error with text.
+  rpc ValidateAndConvert(ValidationRequest) returns (Object) {
```
xiaolanz: When is the message Validation API intended for use? Does it serve as a compromise API usability and generalization?

galley/v1/service.proto
geeknoid: I agree that's it awfully generic. I would suggest ConfigEntry

galley/v1/service.proto
```diff
@@ -0,0 +1,378 @@
...
+message WatchCanceled {
+}
+
+// Indicates that a watch was successfully created.
+message WatchCreated {
+  // returns the initial_state of the specified subtree. Watch stream will begin 
+  // at the revision.
+  repeated Object initial_state = 1;
+
+  // Revision of the tree described in the initial state.
```
geeknoid: Should this be revision of the repository?

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
+// Copyright 2017 Istio Authors
+// vim: set expandtab ts=2 sw=2:
```
geeknoid: Remove

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+package istio.galley.v1;
+
+import "google/rpc/status.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/empty.proto";
+import "google/api/annotations.proto";
+
+// Galley follows
+// the Kubenertes API server delegation and resource model.
+// /apiGroup/version/objecttype/{namespace}/objectname  -->
```
geeknoid: Remove -->

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+//
+// It generalizes the concept of namespace to an arbitrary grouping.
+//
+// /apiGroup//objecttype/version/{objectGroup}/objectname
```
geeknoid: apiGroup// -> apiGroup/

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+// 
+// api_group = /core/
```
geeknoid: Above it says apiGroup, as opposed to api_group

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+    option (google.api.http) = {
+      get: "/{key.api_group}/{key.object_type}/{key.object_type_version}/{key.object_group}/{key.name}"
+    };
+  };
+
+  // Get a list of objects.
+  // Other arguments like query.object_group, query.name are optional
+  // This lets the caller list all objects of a type, or all objects of a type
+  // that is within an object group.
+  rpc ListObjects(ListObjectsRequest) returns (ObjectList) {
```
geeknoid: Does your API pattern follow proper naming conventions? In particular, should the return value of this function be a ListObjectsResponse? Similar for GetObject object, should it return a GetObjectResponse? Hong would be the authority on this.

@wora

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+      get: "/{meta.api_group}"
+      additional_bindings: {
+        get: "/"
+      };
+    };
+  };
+
+  // Create an object. This may result in configuration validation errors.
+  // Referential integrity should be maximally enforced, except where not possible or desirable.
+  rpc CreateObject(ObjectRequest) returns (ConfigObject) {
```
geeknoid: How are errors returned here and elsewhere? We need a rich model for quality error results.

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+
+  // Update a single object. May return validation errors.
+  rpc UpdateObject(ObjectRequest) returns (ConfigObject) {
+    option (google.api.http) = {
+      put: "/{key.api_group}/{key.object_type}/{key.object_type_version}/{key.object_group}/{key.name}"
+      body: "source_data"
+    };
+  };
+
+  // Delete a single object. May return validation errors.
```
geeknoid: So it's possible that Delete will fail due to consistency constraints. That's always difficult for the user. The errors we return will need to include the steps needed to be able to delete.

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+  rpc DeleteObject(DeleteObjectRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      delete: "/{meta.api_group}/{meta.object_type}/{meta.object_type_version}/{meta.object_group}/{meta.name}"
+    };
+  };
+};
+
+
+// Watcher service is designed to efficiently watch mutiple subtrees of the resource tree.
```
geeknoid: Can we put watcher into a separate proto file?

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+  rpc Watch(stream WatchRequest) returns (stream WatchResponse) {
+    option (google.api.http) = {
+      post: "/events/v1:watch"
+      body: "*"
+    };
+  };
+};
+
+
+// Object validator service validates objects before they are committed to storage.
```
geeknoid: How about we just talk in terms of Transformers? Transform user input into storage format. The transformation will fail if the input is invalid.

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+// For example MixerValidator Service should validate all Mixer resources.
+service ObjectValidatorAndConverter {
+  // Validate the resource and convert it to typed proto if applicable
+  // if Object.source_data is specified, it should be converted 
+  // to the appropriate proto representation.
+  // Every attempt should be made to do a deep validation.
+  // If full validation requires referential integrity checks, this service should use the
+  // GalleyWatch Service to maintain current view of configuration.
+  //
+  // For example A Mixer rule consists of a selector expression and a named handler amongst other things.
```
geeknoid: A Mixer -> a Mixer

galley/v1/service.proto
```diff
@@ -0,0 +1,384 @@
...
+  // if Object.source_data is specified, it should be converted 
+  // to the appropriate proto representation.
+  // Every attempt should be made to do a deep validation.
+  // If full validation requires referential integrity checks, this service should use the
+  // GalleyWatch Service to maintain current view of configuration.
+  //
+  // For example A Mixer rule consists of a selector expression and a named handler amongst other things. 
+  // Mixer validator should check 
+  // 1. expression syntax is valid
+  // 2. expression uses known attributes
```
geeknoid: How does Mixer know the valid attributes? We need a flow to deliver attribute manifests.

galley/v1/service.proto
ZackButcher: +1

galley/v1/service.proto
mandarjog: ok

galley/v1/service.proto
mandarjog: ok

galley/v1/service.proto
mandarjog: ok

galley/v1/service.proto
mandarjog: ok

galley/v1/service.proto
mandarjog: I think it misses the point about Validator. 

I can make it `ValidatorAndTransformer`.

galley/v1/service.proto
mandarjog: done

galley/v1/service.proto
mandarjog: I will take that as a followup. 
We don't have a mechanism to deliver attribute manifests today either. So that problem needs to be solved orthogonally. 

At present we assume that Mixer has access to the attribute manifest somehow, because it does validations today. This will continue until a process supersedes it.

galley/v1/service.proto
mandarjog: ok

galley/v1/service.proto
mandarjog: At present errors are google.rpc.Status.
We can add validation errors, however given that the errors are for human consumption this may suffice.

galley/v1/service.proto
mandarjog: per API guide, 
``` rpc GetBook(GetBookRequest) returns (Book) ```
So unless there is a need to have other information Book is the return value and on failure we return `google.rcp.Status`

Comment (ZackButcher) [inline galley/v1/service.proto]: That rule is specific to `GetFoo`. Lists wrap their response in a FooListResponse object: `rpc ListBooks(ListBooksRequest) returns (ListBooksResponse);`

https://cloud.google.com/apis/design/naming_convention#list_response

**Comment by geeknoid** [comment]
I'm worried that just returning a google.rpc.status for errors in
insufficient, and so the general API structure may need to change as a
result.

If there are several validation errors, just returning a single
google.rpc.Status will not create a good experience for the user. I think
we want a more structured thing that returns individual errors in an array,
with some well-known metadata for things like filenames and/or line numbers.

**Comment by ZackButcher** [comment]
Status allows us to return details per failure via it's [`details` field](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto#L91). We can return a single status representing the logic result of the user's operation, and additionally return one detail per failure. We can then use the existing [error_details proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) as the payloads, or define our own.

File: galley/v1/service.proto [hunk 71]
```diff
@@ -0,0 +1,263 @@
...
+
+  // If non zero, response should have at most these number of entries.
+  int32 max_page_size = 3;
+}
+
+message ObjectRequest {
+  // key associated with the requested object.
+  Meta key = 1;
+
+  // oneof source_data or data
```
geeknoid: Delete this comment.

galley/v1/service.proto
```diff
@@ -0,0 +1,263 @@
...
+  Meta query = 1;
+
+  // paged result, set to empty on first page.
+  string page_token = 2;
+
+  // If non zero, response should have at most these number of entries.
+  int32 max_page_size = 3;
+}
+
+message ObjectRequest {
```
geeknoid: Probably should be CreateObjectRequest

galley/v1/service.proto
```diff
@@ -0,0 +1,263 @@
...
+  string next_page_token = 3;
+} 
+
+
+// Meta is the metadata associated with an object.
+message Meta {
+  // api_group is the top level delegation and extension point.
+  string api_group = 1;
+
+  // objectType is the type of an object.
```
geeknoid: object_type

galley/v1/service.proto
```diff
@@ -0,0 +1,263 @@
...
+  // It corresponds to a behavior being configured.
+  // route-rule, metrics, logs, ...
+  string object_type = 2;
+  
+  // every object type is independently versioned.  
+  string object_type_version = 3;
+
+  // objectGroup is an arbitrary hierarchical grouping of resources.
```
geeknoid: object_group

galley/v1/service.proto
```diff
@@ -0,0 +1,263 @@
...
+  // For Mixer configuration object_group == mesh, svc:svc1, svc:svc2 ...
```
geeknoid: Delete the mixer/proxy-specific comment.

galley/v1/service.proto
```diff
@@ -0,0 +1,263 @@
...
+  repeated ConfigObject objects = 2;
+
+  // If next_page_token is not empty, this is a paged result.
+  // use this value in the next request.
+  string next_page_token = 3;
+} 
+
+
+// Meta is the metadata associated with an object.
+message Meta {
```
geeknoid: Do we want a payload_type enum as suggested yesterday in our conversation with Daniel?

galley/v1/watcher.proto
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// WatchProgress message is sent periodically.
+message WatchProgress {
+  // Revision of the repository when this event was sent.
+  int64 current_revision = 1;
+}
+
+
+// The client should WatchResponse
```
geeknoid: Not sure what this comment is trying to say.

galley/v1/watcher.proto
```diff
@@ -0,0 +1,142 @@
...
+  EventType type = 1;
+
+  // kv holds the KeyValue for the event.
+  // A PUT event contains current kv pair.
+  // A PUT event with kv.Version=1 indicates the creation of a key.
+  // A DELETE event contains the deleted key with
+  // its modification revision set to the revision of deletion.
+  ConfigObject kv = 2;
+
+  // prev_kv holds the key-value pair before the event happens.
```
geeknoid: happens -> happened

galley/v1/watcher.proto
```diff
@@ -0,0 +1,142 @@
...
+    };
+  };
+};
+
+
+// WatchRequest creates on cancels a watch.
+// Create and cancel are part of the same message because the WatchRequest message
+// is used in a streaming API. The client may add new watchers and remove old watchers
+// on an existing stream.
+message WatchRequest {
```
lizan: There are known issues to implement bidi streaming for HTTP1 (See [RFC6202](https://tools.ietf.org/html/rfc6202)), and combines multiple watches in one streaming API adds too much complexity to both server and client implementation.

It sounds like implementing multiplexing by hand.

Since gRPC comes with multiplexing and flow-control, it is better to just use 1 stream per WatchCreateRequest, and use the stream state to handling cancel etc.

Is there any strong reason not to do so?

mandarjog: Having multiple watch subscriptions arrive serialized thru the same pipe is the semantic that we are want here. The main consumer of this API is a grpc client.

I don't think it adds complexity at the point of use. Stream provides a good place to maintain subscription state .

Will bidi streaming simply not work with http1 ?

ZackButcher: Where is multiplexing occurring here? The client initiates listening to a stream of updates by informing the server which updates its interested in. The client can subsequently inform the server that it's interested in more (or less) updates without breaking its existing connection to the server. From the perspective of both the client and server the stream of changes is in order.

If anything, I'd argue that streaming is simpler for the client. By forcing the client to handle a stream for each subset of events it cares about, the client now has to open N connections to the server rather than one, and has to handle listening to all of them. Chances are the client will wire up the same listener to each of the N connections - at that point, what's the advantage of multiple connections vs one connection? Further, gRPC supports flow control on streaming APIs.

ZackButcher: And is HTTP1 support that important? We expect the primary consumers of this API to use gRPC. We can expose a streaming REST interface over HTTP2. Is it really a huge loss to not be able to stream listening to changes over HTTP1?

lizan: Even the main consumer is gRPC client, can't multiple one directional streaming API calls in same gRPC channel provide same semantics? (i.e. let gRPC library manage streams state instead of clients)

Bidi streaming rarely works well with HTTP1, it requires client stack to be stream aware.

lizan: @ZackButcher client never has to open N connections (for gRPC or HTTP2) client. And for HTTP1 it has to but still better than bidi.

The watch_id here is logically stream_id and you are re-implementing multiplexing here, no? I'm not saying this should be unary, but 1 stream per WatchCreateRequest sounds more reasonable.

ZackButcher: > The watch_id here is logically stream_id and you are re-implementing multiplexing here, no? I'm not saying this should be unary, but 1 stream per WatchCreateRequest sounds more reasonable.

That's a good point, interpreted that way the stream is multiplexed. However I think the hard parts of multiplexing are lining up the (request, response) pairs and handling per-message cancellations and deadlines (none of which gRPC supports). This API doesn't expose any of those semantics. Instead, I think of it as allowing the client to expand or restrict the stream of changes the server is sending to it.

Suppose instead of using an ID, the cancel request looked like:
```
message WatchCancelRequest {
  Meta subtree = 1;
}
```

That's effectively what the `watch_id` is a shorthand for. At that point, I'd argue that the API doesn't look nearly as much like its attempting to multiplex.

> client never has to open N connections (for gRPC or HTTP2) client. And for HTTP1 it has to but still better than bidi.

Sorry, I was unclear: I mean that complexity for the client was increased because it has to manage N logical (application level) connections to the server, regardless of how gRPC chooses to transmit the data over the network.

lizan: > However I think the hard parts of multiplexing are lining up the (request, response) pairs and handling per-message cancellations and deadlines (none of which gRPC supports). This API doesn't expose any of those semantics. Instead, I think of it as allowing the client to expand or restrict the stream of changes the server is sending to it.

Sorry I don't understand that part. Suppose we have one directional streaming API like

```
service Watcher { 
  rpc Watch(WatchCreateRequest) returns (stream WatchResponse) {
    option (google.api.http) = {
      post: "/events/v1:watch"
      body: "*"
    };
  };
}

message WatchResponse {
  oneof response_union {
    WatchCreated created = 3;
    WatchEvents events = 4;
    WatchProgress progress = 6; 
  };
}
```

So each streaming call maps to a `watch_id` here. Since gRPC has cancellation support, client can cancel any stream. 
- Open a new stream instead of sending WatchCreateRequest to bidi stream.
- Cancel existing stream instead of sending WatchCancelRequest.

Isn't gRPC taking care of the hard parts you mentioned?

> Sorry, I was unclear: I mean that complexity for the client was increased because it has to manage N logical (application level) connections to the server, regardless of how gRPC chooses to transmit the data over the network.
It really depends on how client are going to use the data watcher API. If you are watching multiple same kind (object_type) of ConfigObjects that maybe true, but if client are watching different kind of ConfigObjects, the client has to handle them differently.

This just reminds me when we started mixer API with bidi streaming and eventually moved to unary. I'm skeptical that the bidi would work efficiently here, given the complexity that you have to maintain watch_id in both server and client.

kyessenov:
Also please elaborate on the flow control on the watcher. How do we batch
updates and rate limit them?

lizan:
I just meant having multiple one directional streaming let gRPC provide more fine grained flow-control (flow control are per gRPC stream).

galley/v1/watcher.proto
ZackButcher: I meant that the hard parts of a bidi multiplexed stream are those things, which gRPC does not handle for you in its implementation of streaming. But that that is fine because we don't _use_ those things in this API.

If the intent of this API is for the client to dispatching to handlers based on the `watch_id` then I agree that your proposed API, where the client opens a stream per thing it cares about, is better. However, I do not think that that is the intent of this API. I think it's more likely that the client dispatches based on the `response_union`, and doesn't particularly care about the `watch_id`.

lizan: > we're streaming because we really do have a logical stream of changes.

OK, then the intent should be clearly commented, whether the client should send a `WatchCreateRequest` to existing gRPC streaming call or start a new gRPC streaming call.

Though I'm not completely convinced that we really need the ability to change the logical stream dynamically. (How often and how many `WatchCreateRequest` or `WatchCancelRequest` we send per gRPC stream?) If you look at K8s watch API, it doesn't provide such ability but works for most cases.

galley/v1/service.proto
```diff
@@ -0,0 +1,249 @@
...
+      };
+    };
+  };
+
+  // Get a list of object types.
+  // Lists all object types or all object types 
+  // within an api group.
+  rpc ListObjectTypes(ListObjectTypesRequest) returns (ListObjectTypesResponse) {
+    option (google.api.http) = {
+      get: "/{meta.api_group}"
```
jmuk: nit: `meta.api_group` -> `query.api_group`

galley/v1/service.proto
```diff
@@ -0,0 +1,249 @@
...
+    option (google.api.http) = {
+      put: "/{key.api_group}/{key.object_type}/{key.object_type_version}/{key.object_group}/{key.name}"
+      body: "source_data"
+    };
+  };
+
+  // Delete a single object. May return validation errors.
+  rpc DeleteObject(DeleteObjectRequest) returns (DeleteObjectResponse) {
+    option (google.api.http) = {
+      delete: "/{meta.api_group}/{meta.object_type}/{meta.object_type_version}/{meta.object_group}/{meta.name}"
```
jmuk: nit: `meta` -> `key`

galley/v1/validator.proto
```diff
@@ -0,0 +1,89 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.galley.v1;
+
+import "google/rpc/status.proto";
+import "google/api/annotations.proto";
+
+import "service.proto";
```
jmuk: nit: import line should specify the path from the root. i.e. `galley/v1/service.proto`

galley/v1/watcher.proto
```diff
@@ -0,0 +1,141 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.galley.v1;
+
+import "google/rpc/status.proto";
+import "google/api/annotations.proto";
+
+import "service.proto";
```
jmuk: ditto: `galley/v1/service.proto`

galley/v1/service.proto
```diff
@@ -0,0 +1,249 @@
...
+  };
+
+  // Get a list of object types.
+  // Lists all object types or all object types 
+  // within an api group.
+  rpc ListObjectTypes(ListObjectTypesRequest) returns (ListObjectTypesResponse) {
+    option (google.api.http) = {
+      get: "/{meta.api_group}"
+      additional_bindings: {
+        get: "/"
```
jmuk: As I commented previously, grpc-gateway can't compile with this spec (see https://github.com/grpc-ecosystem/grpc-gateway/issues/414). Consider removing this additional binding for the time being.

galley/v1/service.proto
```diff
@@ -0,0 +1,249 @@
...
+      };
+    };
+  };
+
+  // Create an object. This may result in configuration validation errors.
+  // Referential integrity should be maximally enforced, except where not possible or desirable.
```
jmuk: I'm not sure what this line means. Could you elaborate more?

galley/v1/watcher.proto
```diff
@@ -0,0 +1,141 @@
...
+    DELETE = 1;
+  }
+  // type is the kind of event. If type is a PUT, it indicates
+  // new data has been stored to the key. If type is a DELETE,
+  // it indicates the key was deleted.
+  EventType type = 1;
+
+  // kv holds the KeyValue for the event.
+  // A PUT event contains current kv pair.
+  // A PUT event with kv.Version=1 indicates the creation of a key.
```
jmuk: `ConfigObject` does not have `version` property.

galley/v1/watcher.proto
```diff
@@ -0,0 +1,141 @@
...
+  // A DELETE event contains the deleted key with
+  // its modification revision set to the revision of deletion.
```
jmuk: It's hard to ensure to deliver the exact revision when the deletion happened -- some compaction or bundled event delivery would lose such information. Maybe, we can just say that the revision means that the kv doesn't exist at that point.

galley/v1/watcher.proto
```diff
@@ -0,0 +1,141 @@
...
+  repeated ConfigObject initial_state = 1;
+
+  // Revision of the repository when the initial state was produced.
+  int64 current_revision = 2;
+}
+
+// WatchEvents indicates that this message contains events from the watch stream.
+message WatchEvents {
+  // returns events for the specified watch id. 
+  repeated Event events = 4;
```
jmuk: Can it also has the current_revision too?
Otherwise the watcher client can't tell what's the exact revision right now, it needs to wait for the next WatchProgress event. That sounds like redundant.

galley/v1/service.proto
```diff
@@ -0,0 +1,249 @@
...
+}
+
+message ValidationError {
+  // The key of the object where error was detected.
+  // This may not be the object that introduced the error.
+  string key = 1;
+  // The primary field that was in error, if available.
+  string field = 2;
+  // The line number on which error occurred, if available.
+  int32 line_number = 3;
```
jmuk: Since the config object has no information about the original source (`source_data` is already converted into `Struct` proto), it's probably impossible to respond with line number in the original source. Maybe we should keep the original source string in the config object too?

geeknoid:
We have to maintain provenance information. Every layer that performs a
transformation on some textual input needs to keep track of the line the
input came from.

galley/v1/service.proto
```diff
@@ -0,0 +1,249 @@
...
+}
+
+message GetObjectRequest {
+  // key associated with the object to fetch.
+  Meta key = 1;
+}
+
+message DeleteObjectRequest {
+  // key associated with the requested object.
+  Meta key = 1;
```
elevran: any validation that the meta.version, emta.revision, etc. matches the current state before deletion?

galley/v1/service.proto
```diff
@@ -0,0 +1,249 @@
...
+  // route-rule, metrics, logs, ...
+  string object_type = 2;
+  
+  // every object type is independently versioned.  
+  string object_type_version = 3;
+
+  // object_group is an arbitrary hierarchical grouping of resources.
+  // For Mixer configuration object_group == mesh, svc:svc1, svc:svc2 ...
+  // For Proxy it can be mesh, destination. 
+  string object_group = 4;
```
elevran: the comment above seems to suggest that string could have a format (e.g., svc:svc1, etc.) that indicates the grouping.
Is this the intent? Is Galley aware of the grouping and/or syntax used?

galley/v1/validator.proto
```diff
@@ -0,0 +1,89 @@
...
+// Validation repesents individual change to be validated.
+message Validation {
+  oneof validation_union {
+    // This object should be validated and converted to binary encoded proto.
+    ConfigObject object = 1;
+
+    // If this object was deleted.
+    // In this case Validator should check referential integrity.
+    // Only metadata about the object being deleted is sent.
+    Meta deleted = 2;
```
elevran: And deny deletion if it could result in loss of referential integrity?
What about cross-component referential integrity - needed, possible?

galley/v1/watcher.proto
elevran: It would be helpful to describe a concrete use case, where a client needs (or does not need) to change subscription of a watch. While useful in theory, in many cases I would expect a component to establish a watch "near the (logical) top of its hierarchy", possibly with filtering by type, so it gets all events of interest without granular control.

galley/v1/watcher.proto
elevran: Maybe comment should refer to `ConfigObject.Meta.revision`?

mandarjog:
@elevran I am abandoning this PR in favor of https://github.com/istio/galley/pull/51

PR #124, opened 2017-06-13
Combine Quota functionality into the Check method.

geeknoid:
Check is now expressed as a series of conditions to check.
It can check preconditions (what we've traditionally associated
with Check), and it can check/allocate quotas.

This approach lets us reduce the RPC overhead from 2 calls to 1 and lets Mixer
execute precondition checks and quota allocations in parallel.

THIS WILL BE A BREAKING API CHANGE.


mixer/v1/check.proto
```diff
@@ -25,24 +25,49 @@ option (gogoproto.goproto_getters_all) = false;
...
+
+// An individual work item for `CheckRequest`.
+message Condition {
+  // Name of the condition to check.
+  //
+  // This is either:
+  //
+  //   - `precondition` to indicate a generic precondition check.
+  //   - `quota:<quota name>` to indicate a quota check.
+  string name = 1;
```
douglas-reid: since this isn't a map of <name, condition> or similar, what is the expected behavior for multiple `Conditions` with the `name` of "precondition" with differing amounts?  Does mixer take the first? the last? the minimum?

mixer/v1/service.proto
```diff
@@ -38,14 +37,11 @@ import "mixer/v1/quota.proto";
...
 // both the service operator as well as for service consumers.
 service Mixer {
-  // Checks preconditions before performing an operation.
+  // Checks preconditions and allocate quota before performing an operation.
   // The preconditions enforced depend on the set of supplied attributes and
   // the active configuration.
   rpc Check(CheckRequest) returns (CheckResponse) {}
 
-  // Quota allocates and releases quota.
-  rpc Quota(QuotaRequest) returns (QuotaResponse) {}
```
douglas-reid: for my understanding: are we getting rid of Quota altogether, or just for now? is the idea that resource quotas would use the unified Check() too?

mixer/v1/check.proto
```diff
@@ -53,31 +78,25 @@ message CheckResponse {
...
+// `Check` call can be reused within the parameters specified by the result.
+//
+// Caching should only take place when the `Check` method itself returns success,
+// indicating a successful end-to-end RPC.
+message ConditionResult {
+  // name of the condition, as specified in `Condition.name`
+  string name = 1;
+
+  // The amount of time for which this result can be considered valid.
+  google.protobuf.Duration duration = 2 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
```
douglas-reid: maybe prepend `valid_` here, since we no longer have the context of CachabilityInfo?  so, `valid_duration` and `valid_use_count` ?

mixer/v1/check.proto
```diff
@@ -53,31 +78,25 @@ message CheckResponse {
...
-  // can lead to changes in the result of `Check`. Determining this precise subset
-  // of attributes is handled by the Istio configuration system, and made available
-  // out-of-band to clients.
-  //
-  // Caching should only take place when the `Check` method itself returns success,
-  // indicating a successful end-to-end RPC. Both positive responses (when `CheckResponse.status.Code` is OK)
-  // and negative responses (when `CheckResponse.status.Code is not OK) can be cached.
-  CachabilityInfo cachability = 3 [(gogoproto.nullable) = false];
+  // Provides information regarding each individual tested condition.
+  repeated ConditionResult results = 3;
```
qiwzhang: I think status should be inside ConditionResult too?   Each condition may have different status result.  Server should not combine them, or client cache may not be able to use it.

mixer/v1/check.proto
geeknoid: I made it a map, that way I don't need to answer you :-)

mixer/v1/check.proto
```diff
@@ -25,24 +25,49 @@ option (gogoproto.goproto_getters_all) = false;
...
+  //
+  //   - `precondition` to indicate a generic precondition check.
+  //   - `quota:<quota name>` to indicate a quota check.
+  string name = 1;
+
+  // The number of desired uses of the returned value.
+  //
+  // This is primarily meaningful for quotas and indicates the number of quota tokens
+  // desired. For non-quota conditions, this can be set to 0.
+  int64 amount = 2;
```
geeknoid: Done.

mixer/v1/check.proto
geeknoid: Done.

mixer/v1/service.proto
geeknoid: Yes, I think that even resource quotas would eventually go through Check.

For the foreseeable future, I expect that the intermediated API will only support rate quotas, and resource quotas will only be exposed through the disintermediated API.

mixer/v1/check.proto
geeknoid: My intent was to not support partial failures. If a precondition check fails, then we shouldn't allocate any quotas.

geeknoid:
Updated per Doug's comments.

Wayne, you make a good point that an efficient cache will inherently reduce the RPC overhead of Check+Quota. However, the overhead will not be negligible and this new API would cut this overhead by almost 50%. Additionally, it's worth considering the behavior of the system when check and quota results aren't in the cache. In order to maintain correct quota accounting, the client will need to issue Check, wait for Check to return, and then issue Quota. This means that the latency for uncached accesses is in fact 2x the cost of a single API. Ouch. Combining these two operations together avoids this double wait.

Do you really feel the API is that complicated? CheckRequest and CheckResponse remain fairly small structure with narrow semantics.

mixer/v1/check.proto
qiwzhang: How about precondition passed, but quota fails,  how can client tell a failed status is from which condition?   For just Check and Quota,  if Quota failure has a fixed status code,  we may be able to work around this problem.   But semantic wise,  a repeated conditions may means a lot of conditions to check,  client like to know the result of each condition.

mixer/v1/check.proto
qiwzhang: we need:

bool best_effort = 3;

geeknoid: Do we have use cases where best_effort=false applies?

geeknoid:
@qiwzhang In terms of usability, I think we need to discuss this in person. Previously you had to do:

1. Fill in CheckRequest
2. Make Check call
3. Wait for Check call to complete
4. Fill in QuotaRequest
5. Make Quota call
6. Repeat #4 and #5 for however many quotas there are
7. Wait for all Quota calls to complete

Now, you supply all the same information in a single CheckRequest and you get the same things back in response. It's actually simpler to use since there is no synchronization between two API calls.

qiwzhang:
Your steps are without cache.  With Check and Quota cache, the steps are:

1) call CheckCache(attributes,   check_transport_cb,   check_done_cb);
2) call QuotaCache(attiributes ,  quota_transport_cb,  quota_done_cb);

check_transport_cb is making the actual remote call.   it is cache to determine when to call it since they have different expiration for check and quota.

mixer/v1/check.proto
```diff
@@ -25,24 +25,49 @@ option (gogoproto.goproto_getters_all) = false;
...
   // Mixer's configuration determines how these attributes are used to
-  // determine the result returned in the response.
+  // establish the result returned in the response.
   Attributes attributes = 1 [(gogoproto.nullable) = false];
 
-  // The number of words in the global dictionary.
-  // To detect global dictionary out of sync between client and server.
+  // The number of words in the global dictionary, used with to populate the attributes.
+  // This value is used as a quick way to determine whether the client is using a dictionary that
+  // the server understands.
```
ZackButcher: Why did we decide to use number of words in the dictionary as the differentiator rather than something even marginally stronger like a checksum?

As I've said in other places, I think we need to plan for better handling of the dictionary. Why not mandate a version number for the dictionary and use that here instead?

mixer/v1/check.proto
ZackButcher: `best_effort` makes sense for quota, but not check. I already didn't love `amount` here since it's only used for half the call. If we want that extra info, why not:

```proto
message Condition {
  string name = 1;
  
  message QuotaParams {
    int64 amount = 1;
    bool best_effort = 2;
  }
  message CheckParams {
    // none are required today
  }

  oneof operation {
    QuotaParams quota = 2;
    CheckParams check = 3;
  }
}
```

qiwzhang: quota-prefetch will need it.

geeknoid:
PTAL.

This uses a more general-purpose approach for conditions. Maps are gone, and instead simpler arrays are used instead. The best_effort functionality is back for quota.

This continues to have a single deduplication_id per request, as opposed to per condition. Similarly, this continues to have a single status result per response, instead of one per condition. I don't believe we can reasonably support partial failures of this API. As it is defined now, it's all or nothing, which is easier for consumers to deal with.

mixer/v1/check.proto
```diff
@@ -25,24 +25,54 @@ option (gogoproto.goproto_getters_all) = false;
...
+    bool best_effort = 2;
+  }
+
+  oneof param {
+    PreconditionParams precondition = 3;
+    QuotaParams quota = 2;
+  }
 }
 
 message CheckResponse {
```
ZackButcher: As I understand it, the long term goal is for Mixer to return the set of attributes it used to determine the result, and for Envoy to cache on those attributes. Should we include an `Attributes` value on the `CheckResponse` to use for that? Right now the comments on `ConditionResult` seem to indicate that all attributes sent to Mixer will be used to construct the cache key, which will result incorrect cache misses.

mixer/v1/check.proto
```diff
@@ -53,31 +83,23 @@ message CheckResponse {
...
-  // of attributes is handled by the Istio configuration system, and made available
-  // out-of-band to clients.
-  //
-  // Caching should only take place when the `Check` method itself returns success,
-  // indicating a successful end-to-end RPC. Both positive responses (when `CheckResponse.status.Code` is OK)
-  // and negative responses (when `CheckResponse.status.Code is not OK) can be cached.
-  CachabilityInfo cachability = 3 [(gogoproto.nullable) = false];
+  // Provides information regarding each individual tested condition.
+  // There's one result per tested condition as specified in `CheckRequest.conditions`
+  repeated ConditionResult results = 3;
```
ZackButcher: `CheckRequest.conditions[i]` corresponds to `CheckResponse.results[i]`, given that the results have no name? We should make note of that in the comments.

mixer/v1/check.proto
```diff
@@ -25,24 +25,54 @@ option (gogoproto.goproto_getters_all) = false;
...
 }
 
 message CheckResponse {
-  // The thumbs-up/thumbs-down decision for the precondition check.
+  // The thumbs-up/thumbs-down decision for the overall condition check.
   //
-  // A status code of OK indicates all preconditions were satisfied. Any other code indicates not
-  // all preconditions were satisfied.
+  // A status code of OK indicates all conditions were satisfied. Any other code indicates not
+  // all conditions were satisfied.
```
qiwzhang: if client has separate check cache and quota cache,  it needs to know an error status from which condition?   How can the client know which condition results in the error?

mixer/v1/check.proto
qiwzhang: count() and expiration() also applied to failed condition.   Even a failed condition can be cached. 
Is that correct?

mixer/v1/check.proto
```diff
@@ -25,24 +25,54 @@ option (gogoproto.goproto_getters_all) = false;
...
+  // This value is used as a quick way to determine whether the client is using a dictionary that
+  // the server understands.
   uint32 global_word_count = 2;
+
+  // The various things the `Check` call is expected to verify.
+  repeated Condition conditions = 3;
+
+  // Used for deduplicating `Check` calls in the case of failed RPCs and retries. This should be a UUID
+  // per call, where the same UUID is used for retries of the same call.
+  string deduplication_id = 4;
```
geeknoid: Yes, it needs to be set in all uses of the API. I'll probably add it to Report as well. This is just a way to ensure that mutating operations only occur once. It matters for quota, and it matters for thing like Report for billing events.

So whenever you send a message, just set the dedup id. If the RPC fails, you can retry the same message with the same dedup ID.

mixer/v1/check.proto
geeknoid: Yes, we'll need something like that. I'm deferring that design to a later PR.

geeknoid:
OK, this is yet another variation of the API surface.  It's simpler overall than before and supports the semantics Wayne needs for proper caching.

PTAL.

mixer/v1/check.proto
```diff
@@ -53,31 +73,28 @@ message CheckResponse {
...
-  // The amount of time for which a `Check` response can be considered valid.
-  google.protobuf.Duration duration = 3 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
+  message QuotaResult {
+    // The amount of time for which this result can be considered valid.
+    google.protobuf.Duration valid_duration = 1 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
 
-  // The number of uses for which a `Check` response can be considered valid.
-  int32 use_count = 4;
+    // The number of uses for which this result can be considered valid.
+    int32 valid_use_count = 3;
```
qiwzhang: should be granted_amount  to match "amount" in the request.

qiwzhang: it should be 0 if best_effort = false, and amount is not enough.  it can be less than "amount" if best_effort is true.

qiwzhang: type is int64 too to match "amount" type in the request.

geeknoid: Done

geeknoid:
PTAL.

PR #131, opened 2017-06-29
Combine Quota functionality into the Check method.

geeknoid:


qiwzhang:
@geeknoid I have to submit it since I am waiting for this PR.

PR #133, opened 2017-07-20
Add support for caching hints to CheckResponse.

geeknoid:


mixer/v1/check.proto
```diff
@@ -91,3 +97,31 @@ message CheckResponse {
...
+
+    // The kind of match against the attribute value
+    Condition condition = 2;
+
+    // The matching regex is the case of condition = REGEX
+    string regex = 3;
+  }
+
+  // The message-level dictionary.
+  repeated string words = 1;
```
qiwzhang: not need to sent this back, client still have the whole request proto.

geeknoid: I don't think Mixer is in a good position to reuse the request's dictionary. It definitely fits better to regenerate the message-level dictionary if needed. The problem is that by the time the response is produced, Mixer no longer has the request around. Returning an explicit dictionary means a bit less coupling between the request and response messages.

If having this here makes it harder for the client to recover (since the ID space for the requests is now different than the respones), then I could remove it and deal with the extra complication in Mixer.

qiwzhang: Client has it.  If it will be easier for mixer to generate a new one,  client can work with that too.

mixer/v1/check.proto
```diff
@@ -91,3 +97,31 @@ message CheckResponse {
...
+  // How an attribte's value was matched
+  enum Condition {
+    EXACT = 0;
+    PREFIX = 1;
+    REGEX = 2;
+  }
+
+  message AttributeMatch {
+    // The name of the attribute.
+    int32 index = 1;
```
rshriram: Reword the comment perhaps? Name of attribute doesn't match with "index"

mixer/v1/check.proto
```diff
@@ -91,3 +97,31 @@ message CheckResponse {
...
+
+    // The kind of match against the attribute value
+    Condition condition = 2;
+
```
rshriram: Nit: We have a similar enum in route rule. I think it's called StringMatch. Might want to sync with that for sake of uniformity.

mixer/v1/check.proto
rshriram: May be I am missing some context here. What are the values of this dictionary ? Or is it describing the set of attribute names that the mixer used for matching along with the match type (regex or not), etc.

geeknoid: I'll add better comments here. This is a message-level word dictionary, in the same model as we use in the Attributes message (see attributes.proto)

mixer/v1/check.proto
geeknoid: It's an index into the word dictionary where the name is actually encoded. I'll provide a better comment.

mixer/v1/check.proto
geeknoid: I think I'd rather keep those distinct for the time being.

mixer/v1/check.proto
```diff
@@ -74,6 +74,9 @@ message CheckResponse {
     // ferry new attributes that Mixer derived based on the input set of
     // attributes and its configuration.
     Attributes attributes = 4 [(gogoproto.nullable) = false];
+
+    // The attributes that were used in producing the result
+    RelevantAttributes relevant_attributes = 5 [(gogoproto.nullable) = false];
```
douglas-reid: We might consider renaming this to be something like: `ContributingAttributes` or `ConsumedAttributes`, or perhaps even `UsedRequestAttributes` (to distinguish that these are only relevant attributes that were in the request and not generated attributes).

Relevant requires some context to understand and can imply some value judgement. I'm a bit concerned a quick first reaction would be "why would i ever use anything but the relevant attributes?"

mixer/v1/check.proto
```diff
@@ -74,6 +74,9 @@ message CheckResponse {
     // ferry new attributes that Mixer derived based on the input set of
     // attributes and its configuration.
     Attributes attributes = 4 [(gogoproto.nullable) = false];
+
+    // The attributes that were used in producing the result
```
douglas-reid: Perhaps expand this comment to indicate that these are the subset of attributes used in producing the result -- they include on the attributes that were sent to Mixer, not any used that were generated by Mixer.

mixer/v1/check.proto
```diff
@@ -91,3 +97,33 @@ message CheckResponse {
   // The resulting quota, one entry per requested quota.
   map<string, QuotaResult> quotas = 3 [(gogoproto.nullable) = false];
 }
+
+// Describes the attributes that were used in order to determine the response.
+// This can be used to construct a response cache.
+message RelevantAttributes {
+  // How an attribte's value was matched
```
douglas-reid: s/attribte/attribute

PR #135, opened 2017-08-01
More API tweaks.

geeknoid:
Turns out we need to also match on "doesn't exist" for caching to work
properly. So add an "ABSENCE" match condition. Also, fold PREFIX into REGEX.


geeknoid:
Correct. Consider something like:

if doesnt.exist == 42 {
  return true
}
if request.size == 32 {
  return true
}
return false

Even if the proxy didn't send doesnt.exist to Mixer, that attribute is still influencing the decision Mixer makes. And if the proxy suddenly did define doesnt.exist, the proxy should query Mixer anew.

PR #136, opened 2017-08-02
Add common strings to global dictionary.

qiwzhang:


qiwzhang:
For https://github.com/istio/istio/issues/523

PR #140, opened 2017-08-08
Experimental support for go dep and godep 

costinm:
Checkin generated artifacts - will be ignored by bazel, but used if go dep is used.

This is intended as a proof of concept and to compare ease of build, speed, etc.


kyessenov:
you need to add a script and a test to make sure these stay up to date.
cc @geeknoid

costinm:
Yes, of course - ideally on prow.

Note that I'm not trying to have the exact same version as WORKSPACE - but
use semantic versioning wherever is possible,
and only pin to SHA where absolutely needed ( to match common/recommended
practices in go and go dep).

Regarding artifacts update - yes, will have a script to copy from bazel
genfiles initially.

PR #143, opened 2017-08-15
Rename fields (tags->labels, source->src, destination|target->dst)

rshriram:
Renaming all fields and references to fields in the doc comments.
Order of update in dependent repos could be (to maintain e2e green build)
1. Proxy adds src and dst fields in addition to source/target (as it doesn't depend on this repo)
2. Pilot adds src and dst fields to mixer config in addition to existing source/target (without updating proto)
3. Mixer switches to src/dst while updating the SHA for istio/api (also update istio/istio and e2e test)
4. Pilot removes source/target from mixer config, switches to new proto (also update istio/istio, e2e test)
5. Proxy removes source/target from the attributes it sends
6. Update docs in istio.github.io

### DIFF
--- a/mixer/v1/config/cfg.proto
+++ b/mixer/v1/config/cfg.proto
@@ -34,7 +34,7 @@ package istio.mixer.v1.config;
 // subject: "namespace:ns1"
 // revision: "1011"
 // rules:
-// - selector: target.service == "*"
+// - selector: dst.service == "*"
 //   aspects:
 //   - kind: metrics
 //     params:
@@ -45,7 +45,7 @@ package istio.mixer.v1.config;
 //           statusCode: response.code
 //
 // actionRules:
-// - selector: target.service == "*"
+// - selector: dst.service == "*"
 //   actions:
 //   - handler: prometheus-handler
 //     instances:
@@ -62,8 +62,8 @@ package istio.mixer.v1.config;
 //   params:
 //     value: 1
 //     dimensions:
-//       source: origin.service
-//       target_ip: destination.ip
+//       src: origin.service
+//       dst_ip: destination.ip
 // ```
 //
 // (== deprecation_description ServiceConfig is deprecated, see the Config API's
@@ -100,7 +100,7 @@ message AspectRule {
   //
   // * an empty selector evaluates to `true`
   // * `true`, a boolean literal; a rule with this selector will always be executed
-  // * `target.service == ratings*` selects any request targeting a service whose
+  // * `dst.service == ratings*` selects any request targeting a service whose
   // name starts with "ratings"
   // * `attr1 == "20" && attr2 == "30"` logical AND, OR, and NOT are also available
   string selector = 1;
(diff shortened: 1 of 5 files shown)


frankbu:
I tend to agree with @geeknoid. I prefer "source" and "destination" to "src" and "dst".

Also, there are a bunch of additional changes that are part of aligning mixer and pilot config names. I've been trying to keep track of them here: 

https://github.com/istio/istio/pull/558

Note the 2 open question at the bottom of the doc.

smawson:
I don't like "destination", I find it long and hard to type. I'd be fine with either target, dst, or dest.

source may be fine leaving as source and not shortening.

frankbu:
@smawson Because Istio configuration is a major breaking change this release it makes sense to get all the name changes done now. We already decided not to make the changes in this PR, but we did agree to use the same names for attributes in Mixer and Pilot config. The resulting agreement involves lots of changes to the Pilot config (see: https://github.com/istio/istio/blob/5125ef2f5ee1e0b95d2e28a1fc372ca0f8768116/samples/apps/bookinfo/rules/rule-changes.md), but we also agreed to change "target" to "destination" in the attribute vocabulary, because it is the proper name used for the target in networking terminology.

mixer/v1/config/cfg.proto
```diff
@@ -62,8 +62,8 @@ package istio.mixer.v1.config;
 //   params:
 //     value: 1
 //     dimensions:
-//       source: origin.service
-//       target_ip: destination.ip
+//       src: origin.service
+//       dst_ip: destination.ip
 // ```
```
mandarjog: Why src and dst? 
I thought we had agreed on using full words.

frankbu: @mandarjog Yes, we've agreed to not merge this PR and use full names instead. We're just using this PR as a discussion forum now :-)

rshriram:
```
We already went through a fairly long process of discussing what attribute namespaces to use, we shouldn't be quick to throw those away.
```

We went through the same on the routing side multiple times. 

```
We have documentation and configuration using the existing namespaces that would need to be changed, this is a breaking change for early adopters.
```

The mixer's configuration has changed a lot. Isn't that a breaking change already?

```
destination is long and hard to type compared to target, decreasing system usability and increasing file sizes (and runtime config size).
```

A difference of 5 characters. The argument about decreasing system usability because of a 5 character increase in one attribute name is weak given the complexity of the configuration we have. And even if I were to use the word destination 100 times in a config, I am looking at a 500 byte increase in config. As you yourself said in another document/PR, even the biggest/craziest configurations in youtube are a few hundred kilobytes in size (not gigabytes as @geeknoid had thought). 

```
We already have agreement with other teams within Google to standardize on these attribute names (for example Firebase), changing them here will require re-negotiating those agreements.
```

Living with unintuitive names is worse than having to do a one time rename of a single attribute.

```
So just generally speaking, this is a lot of churn for no clear benefit, and I suggest we stick with the existing names.

I also suggest we change route rule configuration to use "target" instead of "destination" in its field names for consistency, since that is a smaller change, but I also understand if we want to wait until the next release cycle to discuss.
```

It is no small change. It affects the format of routing rules (not to mention informing other blogs like these)
[https://medium.com/@jeffzzq/linkerd-vs-istio-my-2-1e16557891a6
http://dougbtv.com/nfvpe/2017/06/05/istio-deploy/
https://blogs.technet.microsoft.com/livedevopsinjapan/2017/07/19/istio-cicd-pipeline-for-vsts/]

I could replay the same arguments - changing things in routing rules breaks early adopters systems, makes it hard to understand the config, and involves changes in code that other integrators have started doing. 

Either way, we must make the call now. We have changed configs across the system and this would be the chance to unify things.

frankbu:
Should we close this PR and then open another one with the changes in https://github.com/istio/istio/blob/5125ef2f5ee1e0b95d2e28a1fc372ca0f8768116/samples/apps/bookinfo/rules/rule-changes.md?

frankbu:
Closing this PR, replaced by https://github.com/istio/api/pull/152

PR #144, opened 2017-08-16
mark websocket routes

rshriram:


proxy/v1/config/route_rule.proto
```diff
@@ -123,6 +123,11 @@ message RouteRule {
 
   //(-- L4 fault injection policy applies to Tcp/Udp (not HTTP) traffic --)
   L4FaultInjection l4_fault = 10;
+
+  // Indicates that a HTTP/1.1 client connection to this particular route
+  // should be allowed (and expected) to upgrade to a WebSocket connection.
+  // The default is false.
+  bool use_websocket = 11;
```
kyessenov: This looks odd relative to the rest of API...

rshriram: meaning? No structures or fancy fu? 

Websockets are per route and not for entire service. This is kind of what we ended up doing in Envoy as well. It cannot be treated as a different protocol because a client can talk to the server on a given port over standard HTTP/1.1 req-resp and then do a websocket upgrade for a particular path. E.g., server on port 80, where I do auth etc.. and then do ws://server:80/

rshriram: In fact, thats how most websocket clients operate (java/nodejs/so on). No dedicated port for websockets.

proxy/v1/config/route_rule.proto
```diff
@@ -123,6 +123,13 @@ message RouteRule {
 
   //(-- L4 fault injection policy applies to Tcp/Udp (not HTTP) traffic --)
   L4FaultInjection l4_fault = 10;
+
+  // Indicates that a HTTP/1.1 client connection to this particular route
+  // should be allowed (and expected) to upgrade to a WebSocket connection.
+  // The default is false. Envoy expects the first request to this route
+  // to contain the WebSocket upgrade headers. Otherwise, the request
+  // will be rejected.
+  bool websocket_upgrade = 11;
```
esbie: When I run protoc on this file, I get:
`proxy/v1/config/route_rule.proto:132:28: Field number 11 has already been used in "istio.proxy.v1.config.RouteRule" by field "name".`

kyessenov: Argh, @rshriram you broke this build too :) We need a presubmit test here @sebastienvas that at least checks that protos are valid.

rshriram: yikes.. why is name numbered 11!!

PR #150, opened 2017-08-18
Bazel build for sanity checking protos

kyessenov:


kyessenov:
We need some sort of checks to prevent malformed protos. Bazel seems to be most sensible choice given we use it everywhere. I can't seem to import this build file directly, however, so I'd expect repos to override the workspace.

lizan:
Can we use bazel's native [`proto_library`](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_library) instead of `go_proto_library` which has tons of dependencies which aren't used by C++? Also a `BUILD.bazel` in subdirectories are not overridable.

kyessenov:
We want to optimize for the go case since most users are golang at the moment.
By the way, https://github.com/istio/pilot/blob/master/WORKSPACE#L302 seems to override it.

lizan:
> By the way, https://github.com/istio/pilot/blob/master/WORKSPACE#L302 seems to override it.

Yes that's only for `go_repository`, it won't be overridable from mixerclient...

kyessenov:
@lizan @geeknoid would committing the generated pb.go files be a problem for mixerclient? this repo is the only one that does not follow go conventions of having pb.go next to the protos in pilot's workspace dependencies.

lizan:
I don't think so as long as it doesn't have a BUILD file, it can be easily excluded from mixerclient.

kyessenov:
@lizan we still want a way to generate those files with bazel. if I use root BUILD, would that let you override it?

lizan:
yes root BUILD only would be easier

rshriram:
can we merge this or not? @kyessenov ??

kyessenov:
It's out of date. Hold on to merging.

PR #152, opened 2017-08-18
Unified config api

frankbu:
Changes as per design: https://docs.google.com/document/d/1fGZpgFWJZhNRlQoBlCW815aOFR-ewfxKhla0CcPRsBM/edit#

Migration instructions and bookinfo sample changes tracked here: https://github.com/istio/istio/pull/558

### DIFF
--- a/mixer/v1/config/cfg.proto
+++ b/mixer/v1/config/cfg.proto
@@ -62,8 +62,8 @@ package istio.mixer.v1.config;
 //   params:
 //     value: 1
 //     dimensions:
-//       source: origin.service
-//       target_ip: destination.ip
+//       source: source.service
+//       destination_ip: destination.ip
 // ```
 //
 // (== deprecation_description ServiceConfig is deprecated, see the Config API's
@@ -117,8 +117,8 @@ message AspectRule {
 // aspect: each kind of aspect defines its own `params` proto.
 //
 // The following example instructs Mixer to populate a metric named "response_time"
-// that was declared to have three labels: src_consumer_id, target_response_status_code,
-// and target_service_name. For each label and the metric's `value` we provide
+// that was declared to have three labels: src_consumer_id, dst_response_status_code,
+// and dst_service_name. For each label and the metric's `value` we provide
 // an expression over Istio's attributes. Mixer evaluates these expressions for
 // each request.
 //
@@ -130,8 +130,8 @@ message AspectRule {
 //     value: response.time  # from the set of canonical attributes
 //     labels:
 //       src_consumer_id: source.user | source.uid
-//       target_response_status_code: response.code
-//       target_service_name: destination.service
+//       dst_response_status_code: response.code
+//       dst_service_name: destination.service
 // ```
 message Aspect {
   // Required. The kind of aspect this intent is targeting.
@@ -371,8 +371,8 @@ message Action {
 //   params:
 //     value: 1
 //     dimensions:
-//       source: origin.service
-//       target_ip: destination.ip
+//       source: source.service
+//       destination_ip: destination.ip
 // ```
 message Instance {
   // Required. The name of this instance
--- a/mixer/v1/config/cfg.proto
+++ b/mixer/v1/config/cfg.proto
@@ -34,7 +34,7 @@ package istio.mixer.v1.config;
 // subject: "namespace:ns1"
 // revision: "1011"
 // rules:
-// - selector: target.service == "*"
+// - selector: destination.service == "*"
 //   aspects:
 //   - kind: metrics
 //     params:
@@ -45,7 +45,7 @@ package istio.mixer.v1.config;
 //           statusCode: response.code
 //
 // actionRules:
-// - selector: target.service == "*"
+// - selector: destination.service == "*"
 //   actions:
 //   - handler: prometheus-handler
 //     instances:
@@ -100,7 +100,7 @@ message AspectRule {
   //
   // * an empty selector evaluates to `true`
   // * `true`, a boolean literal; a rule with this selector will always be executed
-  // * `target.service == ratings*` selects any request targeting a service whose
+  // * `destination.service == ratings*` selects any request targeting a service whose
   // name starts with "ratings"
   // * `attr1 == "20" && attr2 == "30"` logical AND, OR, and NOT are also available
   string selector = 1;
@@ -131,7 +131,7 @@ message AspectRule {
 //     labels:
 //       src_consumer_id: source.user | source.uid
 //       target_response_status_code: response.code
-//       target_service_name: target.service
+//       target_service_name: destination.service
 // ```
 message Aspect {
   // Required. The kind of aspect this intent is targeting.
@@ -305,7 +305,7 @@ message EmailAddress {
 //
 // ```yaml
 // actionRules:
-// - selector: target.service == "*"
+// - selector: destination.service == "*"
 //   actions:
 //   - handler: prometheus-handler
 //     instances:
@@ -320,7 +320,7 @@ message Rule {
   //
   // * an empty selector evaluates to `true`
   // * `true`, a boolean literal; a rule with this selector will always be executed
-  // * `target.service == ratings*` selects any request targeting a service whose
+  // * `destination.service == ratings*` selects any request targeting a service whose
   // name starts with "ratings"
   // * `attr1 == "20" && attr2 == "30"` logical AND, OR, and NOT are also available
   string selector = 1;
--- a/proxy/v1/config/dest_policy.proto
+++ b/proxy/v1/config/dest_policy.proto
@@ -26,15 +26,16 @@ package istio.proxy.v1.config;
 //
 //     metadata:
 //       name: reviews
+//       namespace: default # optional (default is "default" or value specified with "kubectl -n")
 //     spec:
 //       policy:
 //       - loadBalancing:
 //           name: RANDOM
 //
-// A single DestinationPolicy resource includes configuration for all service versions. ONLY
-// ONE policy can be defined per service version. Policy CANNOT be empty.
+// A single DestinationPolicy resource includes configuration for all service versions. Only
+// one policy can be defined per service version. Policy CANNOT be empty.
 message DestinationPolicy {
-  // REQUIRED. List of policies, one per service version.
+  // Required. List of policies, one per service version.
   repeated DestinationVersionPolicy policy = 2;
 }
 
@@ -56,7 +57,7 @@ message DestinationPolicy {
 //           name: RANDOM
 //
 // If labels are omitted, the policy applies for all versions of the
-// service. Policy CANNOT BE empty.
+// service. Policy CANNOT be empty.
 // *Note:* Destination policies will be applied only if the corresponding
 // tagged instances are explicity routed to. In other words, for every
 // destination policy defined, atleast one route rule must refer to the
--- a/proxy/v1/config/dest_policy.proto
+++ b/proxy/v1/config/dest_policy.proto
@@ -55,9 +55,11 @@ message DestinationPolicy {
 //       namespace: default
 //     spec:
 //       policy:
-//       - labels:
-//           env: prod
-//           version: v1
+//       - destination:
+//           name: reviews # Optional: default is metadata name
+//           labels:
+//             env: prod
+//             version: v1
 //         loadBalancing:
 //           name: RANDOM
 //
@@ -68,18 +70,19 @@ message DestinationPolicy {
 // destination policy defined, atleast one route rule must refer to the
 // service version indicated in the destination policy.
 message DestinationVersionPolicy {
-  // Optional set of labels that identify a particular version of the
-  // destination service. If omitted, the policy will apply to all versions
-  // of the service. (-- N.B. The map is used instead of
-  // pstruct due to lack of serialization support in golang protobuf
-  // library (see https://github.com/golang/protobuf/pull/208) --)
-  map<string, string> labels = 1;
+  // Optional: Destination uniquely identifies the destination service associated
+  // with this policy.
+  MatchService destination = 1;
+
+  // Optional: Source uniquely identifies the source service associated
+  // with this policy.
+  MatchService source = 2;
 
   // Load balancing policy.
-  LoadBalancing load_balancing = 2;
+  LoadBalancing load_balancing = 3;
 
   // Circuit breaker policy.
-  CircuitBreaker circuit_breaker = 3;
+  CircuitBreaker circuit_breaker = 4;
 
   //(-- Other custom policy implementations --)
   google.protobuf.Any custom = 100 ;
@@ -143,8 +146,10 @@ message CircuitBreaker {
   //       namespace: default
   //     spec:
   //       policy:
-  //       - labels:
-  //           version: v1
+  //       - destination:
+  //           name: reviews
+  //           labels:
+  //             version: v1
   //         circuitBreaker:
   //           simpleCb:
   //             maxConnections: 100
@@ -160,8 +165,10 @@ message CircuitBreaker {
   //       namespace: default
   //     spec:
   //       policy:
-  //       - labels:
-  //           version: v1
+  //       - destination:
+  //           name: reviews
+  //           labels:
+  //             version: v1
   //         circuitBreaker:
   //           simpleCb:
   //             maxConnections: 100
--- a/proxy/v1/config/dest_policy.proto
+++ b/proxy/v1/config/dest_policy.proto
@@ -20,22 +20,26 @@ package istio.proxy.v1.config;
 
 // DestinationPolicy defines client/caller-side policies that determine how
 // to handle traffic bound to a particular destination service. The policy
-// specifies configuration for load balancing and circuit breakers.  For
+// specifies configuration for load balancing and circuit breakers. For
 // example, a simple load balancing policy for the reviews service would
 // look as follows:
 //
 //     metadata:
 //       name: reviews
-//       namespace: default # optional (default is "default" or value specified with "kubectl -n")
+//       namespace: default # optional (default is "default" or value specified with "-n" argument when created)
 //     spec:
 //       policy:
 //       - loadBalancing:
 //           name: RANDOM
 //
+// The FQDN of the destination service is composed from the metadata name and namespace fields, along with
+// a platform-specific domain suffix
+// (e.g. on Kubernetes, "reviews" + "default" + "svc.cluster.local" -> "reviews.default.svc.cluster.local").
+//
 // A single DestinationPolicy resource includes configuration for all service versions. Only
 // one policy can be defined per service version. Policy CANNOT be empty.
 message DestinationPolicy {
-  // Required. List of policies, one per service version.
+  // REQUIRED. List of policies, one per service version.
   repeated DestinationVersionPolicy policy = 2;
 }
 
@@ -48,6 +52,7 @@ message DestinationPolicy {
 //
 //     metadata:
 //       name: reviews
+//       namespace: default
 //     spec:
 //       policy:
 //       - labels:
@@ -87,6 +92,7 @@ message DestinationVersionPolicy {
 //
 //     metadata:
 //       name: reviews
+//       namespace: default
 //     spec:
 //       policy:
 //       - loadBalancing:
@@ -134,6 +140,7 @@ message CircuitBreaker {
   //
   //     metadata:
   //       name: reviews
+  //       namespace: default
   //     spec:
   //       policy:
   //       - labels:
@@ -150,6 +157,7 @@ message CircuitBreaker {
   //
   //     metadata:
   //       name: reviews
+  //       namespace: default
   //     spec:
   //       policy:
   //       - labels:
--- a/proxy/v1/config/dest_policy.proto
+++ b/proxy/v1/config/dest_policy.proto
@@ -26,7 +26,7 @@ package istio.proxy.v1.config;
 //
 //     metadata:
 //       name: reviews
-//       namespace: default # optional (default is "default" or value specified with "-n" argument when created)
+//       namespace: default # optional (default is "default")
 //     spec:
 //       policy:
 //       - loadBalancing:
--- a/proxy/v1/config/dest_policy.proto
+++ b/proxy/v1/config/dest_policy.proto
@@ -25,22 +25,41 @@ package istio.proxy.v1.config;
 // look as follows:
 //
 //     metadata:
-//       name: reviews
+//       name: reviews-lb-policy
 //       namespace: default # optional (default is "default")
 //     spec:
+//       destination:
+//         name: reviews
 //       policy:
 //       - loadBalancing:
 //           name: RANDOM
 //
-// The FQDN of the destination service is composed from the metadata name and namespace fields, along with
+// The FQDN of the destination service is composed from the destination name and meta namespace fields, along with
 // a platform-specific domain suffix
 // (e.g. on Kubernetes, "reviews" + "default" + "svc.cluster.local" -> "reviews.default.svc.cluster.local").
 //
 // A single DestinationPolicy resource includes configuration for all service versions. Only
 // one policy can be defined per service version. Policy CANNOT be empty.
 message DestinationPolicy {
-  // REQUIRED. List of policies, one per service version.
-  repeated DestinationVersionPolicy policy = 2;
+  // Optional: Destination uniquely identifies the destination service associated
+  // with this policy.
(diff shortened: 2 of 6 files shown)


proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,160 @@ package istio.proxy.v1.config;
...
+
+// MatchSource identifies the service and optionally service version initiating the request.
+message MatchSource {
+  // The short name of the service such as "foo".
+  string name = 1;
+
+  // Optional namespace of the source service. Defaults to value of metadata.namespace.
+  string namespace = 2;
+
+  // One or more labels that uniquely identify the source service version. In
```
geeknoid: Here and potentially elsewhere, let's not put k8s-specific content in our generic API/config surface please.

proxy/v1/config/dest_policy.proto
```diff
@@ -24,51 +24,50 @@ package istio.proxy.v1.config;
...
 //
-//     destination: reviews.default.svc.cluster.local
-//     policy:
-//     - tags:
-//         env: prod
-//         version: v1
-//       loadBalancing: 
-//         name: RANDOM
+//     metadata:
+//       name: reviews
```
kyessenov: It's probably a good idea to add "namespace: default" to a concrete example. I know that we can fill it up in kubectl, but this makes the example depend on the implicit kubectl namespace context.

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,160 @@ package istio.proxy.v1.config;
...
+  // Kubernetes, labels correspond to the labels associated with pods.
+  map<string, string> labels = 3;
+}
 
+// MatchRequest specifies propeties of the service request.
```
geeknoid: properties

mixer/v1/config/cfg.proto
```diff
@@ -131,7 +131,7 @@ message AspectRule {
 //     labels:
 //       src_consumer_id: source.user | source.uid
 //       target_response_status_code: response.code
```
geeknoid: We don't want to change these two instances of target_*?

proxy/v1/config/dest_policy.proto
```diff
@@ -24,51 +24,50 @@ package istio.proxy.v1.config;
...
+//     spec:
+//       policy:
+//       - loadBalancing:
+//           name: RANDOM
 //
-// Policies are applicable per individual service versions. ONLY
+// A single DestinationPolicy resource includes configuration for all service versions. ONLY
```
geeknoid: Can we get rid of the SHOUTING in this comment, it's not really necessary.

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,160 @@ package istio.proxy.v1.config;
...
 
   //(-- L4 fault injection policy applies to Tcp/Udp (not HTTP) traffic --)
-  L4FaultInjection l4_fault = 12;
+  L4FaultInjection l4_fault = 11;
+}
+
+// Destination uniquely identifies the destination service.
+// The FQDN of a service is composed from the name, namespace, and implementation-specific domainSuffix
+// (e.g. on Kubernetes, "reviews" + "default" + "svc.cluster.local" -> "reviews.default.svc.cluster.local")
+message Destination {
```
kyessenov: Maybe worth adding `domain` as a provision to custom `domain` suffixes in the future.

proxy/v1/config/dest_policy.proto
```diff
@@ -24,51 +24,50 @@ package istio.proxy.v1.config;
...
 // ONE policy can be defined per service version. Policy CANNOT be empty.
 message DestinationPolicy {
-  // REQUIRED. Service name for which the service version is defined. The
-  // value MUST BE a fully-qualified domain name,
-  // e.g. _my-service.default.svc.cluster.local_.
-  string destination = 1;
-
   // REQUIRED. List of policies, one per service version.
```
geeknoid: Please search-and-replace all REQUIRED and OPTIONAL to merely Required and Optional so we have global consistency.

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,160 @@ package istio.proxy.v1.config;
...
   // (-- Set of layer 4 match conditions based on the IP ranges --)
-  L4MatchAttributes udp = 4 ;
+  L4MatchAttributes udp = 3 ;
+
+  // Properties of the request.
+  MatchRequest request = 4 ;
+}
+
+// MatchSource identifies the service and optionally service version initiating the request.
+message MatchSource {
```
kyessenov: similar, add `domain` here if you add it to another place.

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,160 @@ package istio.proxy.v1.config;
...
+  map<string, string> labels = 3;
+}
 
+// MatchRequest specifies propeties of the service request.
+message MatchRequest {
   // Set of HTTP match conditions based on HTTP/1.1, HTTP/2, GRPC request
   // metadata, such as _uri_, _scheme_, _authority_.  The header keys must be
   // lowercase and use hyphen as the separator, e.g. _x-request-id_.
   //
+  // Header values are case-sensitive and formatted as follows:
```
kyessenov: Let's not use the operator syntax. We can use a custom serialize for the string match proto (see https://github.com/golang/protobuf/pull/325) and serialize exact match directly to a string. E.g.:
```
header:
  exact: x
```
would be:
```
header: x
```

mixer/v1/config/cfg.proto
frankbu: Wasn't sure what we want to do with this one. I was thinking if we changed it from target_response_status_code it would probably be dst_response_status_code (instead of destination_*) to match the short form of src_consumer_id. I left it alone for now, since it's not part of the attribute vocabulary unification problem. Let me know if you want to change it.

proxy/v1/config/route_rule.proto
frankbu: Do you mean keep StringMatch in as the value in the proto?
```
map<string, StringMatch> http_headers = 1;
```
and then the YAML syntax will be:
```
headers:
   foo: "exact value"
   bar:
     prefix: "prefix value"
   baz:
     regex: "regex value"
```
Is that what you are suggesting?

kyessenov: Yes, it's more explicit and covers majority of cases (exact match).

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
 // "reviews" service to version "v1" can be specified as follows:
 //
-//     destination: reviews.default.svc.cluster.local
+//   metadata:
 //     name: my-rule
+//     namespace: default # optional (default is "default" or value specified with "kubectl -n")
+//   spec:
+//     destination:
+//       name: reviews
+//       namespace: my-namespace # optional (default is metadata.namespace)
```
rshriram: metadata.namespace.domain where domain is??

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
-  HTTPFaultInjection http_fault = 11;
+  HTTPFaultInjection http_fault = 10;
 
   //(-- L4 fault injection policy applies to Tcp/Udp (not HTTP) traffic --)
-  L4FaultInjection l4_fault = 12;
+  L4FaultInjection l4_fault = 11;
+}
+
+// Destination uniquely identifies the destination service.
+// The FQDN of a service is composed from the name, namespace, and implementation-specific domainSuffix
```
rshriram: nit (clarify): ```...composed from name (service name), optional namespace and the domain field. In certain platforms such as kubernetes, the domain suffix is automatically inferred, if omitted.``` 

[This would allow us to use normal domains even in k8s]

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
+// (e.g. on Kubernetes, "reviews" + "default" + "svc.cluster.local" -> "reviews.default.svc.cluster.local")
+message Destination {
+  // The short name of the service such as "foo".
+  string name = 1;
+
+  // Optional namespace of the service. Defaults to value of metadata.namespace.
+  string namespace = 2;
+
+  // (-- Domain suffix used to construct the service FQDN in implementations that
```
rshriram: why the hide this field with (-- ? This field needs to be exposed. Even in kubernetes, I might have use cases where I might want to forward to non k8s domains.

Also, we could just state that this is the domain used to construct the FQDN of the service. (no need to mention that this is applicable only for implementations taht support such. We might need this in k8s as well).

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
+// route rule to be applied to the connection or HTTP request. The
 // condition provides distinct set of conditions for each protocol with the
 // intention that conditions apply only to the service ports that match the
 // protocol. For example, the following route rule restricts the rule to
 // match only requests originating from "reviews:v2", accessing ratings
 // service where the URL path starts with /ratings/v2/ and the request
 // contains a "cookie" with value "user=jason",
 //
-//     destination: ratings.default.svc.cluster.local
+//   metadata:
```
rshriram: space mismatch

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
+  // Optional namespace of the source service. Defaults to value of metadata.namespace.
+  string namespace = 2;
+
+  // (-- Domain suffix used to construct the service FQDN in implementations that
+  // support such specification. --)
+  string domain = 3;
+
+  // One or more labels that uniquely identify the source service version.
+  map<string, string> labels = 4;
+}
```
rshriram: This is not clean :( . We have destination and matchsource that have almost all fields overlapping. I understand the reason for duplication (to avoid nesting) but the config is going to look confusing, with one form having 3 fields and another having an extra labels field.

Alternative:

```yaml
Destination source
map<string, string> source_labels
```

nomenclature is still consistent (use of labels).

rshriram: fwiw, we had this 8 months ago and we reverted to the current single level nesting.

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
+
+// MatchRequest specifies properties of the service request.
```
rshriram: Specifies the attributes of a HTTP request to be used for matching a request.

rshriram: and also need an example code snippet. Otherwise the config section looks very odd and unintelligible. That is the reason I added code snippets to almost every section

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
+message MatchRequest {
   // Set of HTTP match conditions based on HTTP/1.1, HTTP/2, GRPC request
   // metadata, such as _uri_, _scheme_, _authority_.  The header keys must be
   // lowercase and use hyphen as the separator, e.g. _x-request-id_.
```
rshriram: This should go into the main comment for MatchRequest (outside)

proxy/v1/config/route_rule.proto
```diff
@@ -65,125 +65,167 @@ package istio.proxy.v1.config;
...
   //
+  // Header values are case-sensitive and formatted as follows:
+  //
+  // *exact: "value"* or just *"value"* for exact string match
+  //
+  // *prefix: "value"* for prefix-based match
+  //
+  // *regex: "value"* for ECMAscript style regex-based match
+  //
```
rshriram: need examples.. May be use the ones you and kuat discussed above

proxy/v1/config/route_rule.proto
```diff
@@ -193,29 +235,30 @@ message MatchCondition {
 // instances with the "v2" tag and the remaining traffic (i.e., 75%) to
 // "v1".
 //
-//     destination: reviews.default.svc.cluster.local
+//   metadata:
```
rshriram: spacing issue?

proxy/v1/config/route_rule.proto
```diff
@@ -193,29 +235,30 @@ message MatchCondition {
...
+  Destination destination = 1;
 
   // Service version identifier for the destination service.
   // (-- N.B. The map is used instead of pstruct due to lack of serialization support
   // in golang protobuf library (see https://github.com/golang/protobuf/pull/208) --)
-  map<string, string> tags = 2;
+  map<string, string> labels = 2;
 
-  // REQUIRED. The proportion of traffic to be forwarded to the service
+  // Required. The proportion of traffic to be forwarded to the service
```
rshriram: skip

proxy/v1/config/route_rule.proto
```diff
@@ -241,12 +284,15 @@ message L4MatchAttributes {
 // requests for /v1/getProductRatings API on the ratings service to
 // /v1/bookRatings provided by the bookratings service.
 //
-//     destination: ratings.default.svc.cluster.local
+//   metadata:
```
rshriram: spacing?

proxy/v1/config/route_rule.proto
```diff
@@ -268,16 +314,19 @@ message HTTPRedirect {
 // demonstrates how to rewrite the URL prefix for api call (/ratings) to
 // ratings service before making the actual API call.
 //
-//     destination: ratings.default.svc.cluster.local
+//   metadata:
```
rshriram: spacing

proxy/v1/config/route_rule.proto
```diff
@@ -268,16 +314,19 @@ message HTTPRedirect {
...
+//   spec:
+//     destination:
+//       name: ratings
 //     match:
-//       httpHeaders:
-//         uri:
-//          prefix: /ratings
+//       request:
+//         headers:
+//           uri: ^/ratings
```
rshriram: old stuff? isn't this supposed to be `uri: prefix: /ratings` ?

proxy/v1/config/route_rule.proto
```diff
@@ -305,18 +354,21 @@ message StringMatch {
 // Describes HTTP request timeout. For example, the following rule sets a
 // 10 second timeout for calls to the ratings:v1 service
 //
-//     destination: ratings.default.svc.cluster.local
+//   metadata:
```
rshriram: spacing

proxy/v1/config/route_rule.proto
```diff
@@ -305,18 +354,21 @@ message StringMatch {
...
+//     - labels:
 //         version: v1
 //     httpReqTimeout:
 //       simpleTimeout:
 //         timeout: 10s
 //
 message HTTPTimeout {
   message SimpleTimeoutPolicy {
-    // REQUIRED. Timeout for a HTTP request. Includes retries as well.  Default
+    // Required. Timeout for a HTTP request. Includes retries as well.  Default
```
rshriram: skip

proxy/v1/config/route_rule.proto
```diff
@@ -336,10 +388,13 @@ message HTTPTimeout {
 // example, the following rule sets the maximum number of retries to 3 when
 // calling ratings:v1 service, with a 2s timeout per retry attempt.
 //
-//     destination: ratings.default.svc.cluster.local
+//   metadata:
```
rshriram: spacing

proxy/v1/config/route_rule.proto
```diff
@@ -348,7 +403,7 @@ message HTTPTimeout {
 //
 message HTTPRetry {
   message SimpleRetryPolicy {
-    // REQUIRED. Number of retries for a given request. The interval
+    // Required. Number of retries for a given request. The interval
```
rshriram: skip

proxy/v1/config/dest_policy.proto
```diff
@@ -24,51 +24,51 @@ package istio.proxy.v1.config;
...
 //
-// Policies are applicable per individual service versions. ONLY
-// ONE policy can be defined per service version. Policy CANNOT be empty.
+// A single DestinationPolicy resource includes configuration for all service versions. Only
+// one policy can be defined per service version. Policy CANNOT be empty.
 message DestinationPolicy {
-  // REQUIRED. Service name for which the service version is defined. The
-  // value MUST BE a fully-qualified domain name,
-  // e.g. _my-service.default.svc.cluster.local_.
-  string destination = 1;
```
rshriram: if we remove this, where are we explaining to users that metadata.name must be the short name of the service? 

Is there a schema for metadata? Also, that schema needs to have domain field, in addition to name and namespace.

mixer/v1/config/cfg.proto
```diff
@@ -131,7 +131,7 @@ message AspectRule {
...
-//       target_service_name: target.service
+//       target_service_name: destination.service
```
frankbu: Same as above discussion about `target_response_status_code`, I think it's a Mixer question, outside of the scope of this PR, which is just addressing alignment of the attribute vocabulary.

proxy/v1/config/route_rule.proto
frankbu: This would be the one place that we don't maintain the "a.b" -> "a: b" (nesting) pattern. Personally, I'd prefer the duplication. It would be nice if proto had inheritance.

proxy/v1/config/route_rule.proto
frankbu: The example is just above in the MatchCondition comments.

mixer/v1/config/cfg.proto
kyessenov: Out of scope. The rename to destination has huge implications for mixer and will take a while to sort out.

proxy/v1/config/dest_policy.proto
kyessenov: This is documented in standard vocabulary. All we do to split FQDN into parts to allow short domain names, which is a standard practice in DNS world.

proxy/v1/config/dest_policy.proto
```diff
@@ -24,51 +24,51 @@ package istio.proxy.v1.config;
...
-
-  // REQUIRED. List of policies, one per service version.
+  // Required. List of policies, one per service version.
```
kyessenov: @ZackButcher what's the state of proto documentation tooling? still manual?

proxy/v1/config/route_rule.proto
kyessenov: The domain name is not used in k8s world since it all defaults to the same value (cluster-domain in kubeadm setup). We will document this domain for VMs and change the code accordingly.

proxy/v1/config/route_rule.proto
kyessenov: @rshriram we agreed to use standard attribute namespaces. This is just executing on our plan. Let's not digress again away from the decision. "source.labels" is an attribute, and we translate "." into proto nesting, not underscores.

proxy/v1/config/route_rule.proto
kyessenov: +1 for snippets (I wish we had better tooling around these)

kyessenov:
ExternalName services are regular k8s services, so we apply a policy in the same way (name: httpbin, namespace: default). Besides, we don't really support external services policies even now (except auth).

proxy/v1/config/dest_policy.proto
rshriram: where is the standard vocabulary? It is not evident that one must supply a rule with name same as that of the destination service's name. Earlier we had a destination field.. But now, its part of metadata. Previous examples used arbitrary names for rule name, but this one has a specific requirement. That needs to be documented right?

kyessenov: Yes, this needs to be documented for destination policies.

rshriram:
I am not talking about external name services. In the light of egress proposals and forwarding egress calls to egress proxy without needing to use the clumsy httpbin.default.svc.local, we are looking at Envoy needing to forward calls to external destinations (e.g., google.com) without needing to use any namespace. And these destinations are just like any other destination. They will have load balancing policies, circuit breakers, etc. How are we going to specify those in the destination policy? I do not see a domain field under metadata. 

Are we going to make a hack in the metadata field, such that the name contains FQDN for external services (with namespace being empty?) ? Does that imply that the destination policy here is universal across all namespaces? what if one accidentally puts the namespace in the metadata field, with the assumption that they want to apply one (Aggressive) circuit breaker setting for google.com from service foo in namespace foo-ns, and a different circuit breaker setting for google.com from service bar in namespace bar-ns ?

proxy/v1/config/route_rule.proto
```diff
@@ -65,20 +65,20 @@ package istio.proxy.v1.config;
...
-//     destination:
-//       name: reviews
-//       namespace: my-namespace # optional (default is metadata.namespace)
-//     route:
-//     - labels:
-//         version: v1
-//       weight: 100
+//     metadata:
+//       name: my-rule
+//       namespace: default # optional (default is "default" or value specified with "-n" argument when created)
```
rshriram: nit. we can stop with default is #default .. the -n does not have a context here. Also, the code box in our webpage would be huge due to long line.

proxy/v1/config/route_rule.proto
```diff
@@ -133,17 +133,16 @@ message RouteRule {
 }
 
 // Destination uniquely identifies the destination service.
-// The FQDN of a service is composed from the name, namespace, and implementation-specific domainSuffix
-// (e.g. on Kubernetes, "reviews" + "default" + "svc.cluster.local" -> "reviews.default.svc.cluster.local")
+// The FQDN of a destinatin service is composed from the name, namespace, and implementation-specific domain suffix
```
rshriram: typo. destination

proxy/v1/config/route_rule.proto
frankbu: I was trying to say that there will be a way to set the value when you create the resource using kubectl, istioctl, or whatever, but I guess you're saying that should be documented with the CLI command, not here?

rshriram: yeah.

frankbu:
@rshriram I kind of think of your external http example as multiple versions of httpbin. Would it make sense to set `name: httpbin`, `namespace: your-ns (or istio-defaults for all)`, and then set the version label, or maybe we add another optional field to the DestinationVersionPolicy, to the dns suffix of the external service?

kyessenov:
For external services, we'd have to include destination policy into the egress rule. This is a special case, and can only support a subset of all policies in the proto definition.

rshriram:
I just confirmed with @ccamaran . Lyft is using custom circuit breakers per source to the same destination (albeit with runtimes).

rshriram:
The egress rule we have today is just an aggregation rule and it is probably going to hit a dead end because envoy does not have data plane dns resolution.
Either way, why do we need a separate proto definition for this? UX wise, I should be able to define circuit breakers for a destination. Does it matter if that destination is in cluster or outside the cluster? The destination policies are very generic (lb, circuit breakers, outliers). These are shared properties across destinations. So, shouldn't we add provisions for expressing such destinations as part of the new model? The old model certainly allowed this (`destination: httpbin.org`). Dropping such a support is effectively a regression, for the cosmetic gain of being able to simplify how one specifies namespaces.

kyessenov:
We did not support external destinations. If a service wants to participate in the mesh policies, it is reasonably to ask to have to a resource to represent the service in the mesh to get all the features.

mixer/v1/config/cfg.proto
rshriram: Lets change this as well.. If we are changing across the board, lets not leave things out.

mixer/v1/config/cfg.proto
rshriram: The amount of changes we are doing here has huge implications for Pilot. Everything has to change. Esp all this stuff about destination policies, namespaces, etc. Especially, after Kelsey published a tutorial on routing rules. So, if we are going down this path, we might as well fix everything and not special case mixer because of implementation pain.

mixer/v1/config/cfg.proto
frankbu: @geeknoid, @kyessenov: are you OK with changing these aspect and template labels/parms as well, to align with the changed vocabulary. There seem to be 3 examples in the proto doc (`target_response_status_code`, `target_service_name`, `target_ip`) but I'm guessing there may be other places that the word target is used to refer to the destination.

PR #162, opened 2017-08-31
Option to disable mixer checks

rshriram:
Allow user to disable checks (but keep reports on) for mixer.

### DIFF
--- a/proxy/v1/config/proxy_mesh.proto
+++ b/proxy/v1/config/proxy_mesh.proto
@@ -30,6 +30,11 @@ message ProxyMeshConfig {
   // Address of the mixer service (e.g. _istio-mixer:9090_).
   string mixer_address = 3;
 
+  // Disable policy checks by the mixer service. Metrics will still be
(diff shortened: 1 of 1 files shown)


proxy/v1/config/proxy_mesh.proto
```diff
@@ -30,6 +30,11 @@ message ProxyMeshConfig {
   // Address of the mixer service (e.g. _istio-mixer:9090_).
   string mixer_address = 3;
 
+  // Disable policy checks by the mixer service. Metrics will still be
+  // reported to the mixer for HTTP requests and TCP connections. Default
+  // is false, i.e. mixer policy check is enabled by default.
+  bool disable_policy_checks = 25;
+
```
frankbu: Why 25? I notice there's also another gap between 15 and 20.

rshriram: There is also one with 100. Ideally, we can simply fix all the numbers before release.

PR #165, opened 2017-09-06
Use one domain per egress rule

rshriram:
Convert from multiple domains per egress rule to a single domain per egress rule. Reuse the destination structure defined in route_rule and cleanup the description.

### DIFF
--- a/proxy/v1/config/egress_rule.proto
+++ b/proxy/v1/config/egress_rule.proto
@@ -14,6 +14,8 @@
 
 syntax = "proto3";
 
+import "proxy/v1/config/route_rule.proto";
+
 package istio.proxy.v1.config;
 
 // Egress rules describe the properties of a service outside Istio. When transparent proxying
@@ -32,7 +34,8 @@ package istio.proxy.v1.config;
 //     metadata:
 //       name: foo-egress-rule
 //     spec:
-//       destination: *.foo.com
+//       destination:
+//         service: *.foo.com
 //       ports:
 //         - port: 80
 //           protocol: http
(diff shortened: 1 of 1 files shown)


proxy/v1/config/egress_rule.proto
```diff
@@ -14,53 +14,56 @@
...
-  // Note that the wildcard will not match the empty string. e.g. “*-bar.foo.com” will match “baz-bar.foo.com” 
-  // but not “-bar.foo.com”.  Additionally, a special entry “*” is allowed which will match any host/authority header.
-  repeated string domains = 2;
+  // REQUIRED: Hostname or a wildcard domain name associated with the external service.
+  // ONLY the name field is taken into consideration. Namespace and domain fields are ignored.
+  // Wildcard domain specifications must conform to format allowed by Envoy's Virtual Host specification,
+  // such as “*.foo.com” or “*-bar.foo.com”. The character '*' indicates a non-empty string.
+  // So, a wildcard domain of form “*-bar.foo.com” will match “baz-bar.foo.com” 
+  // but not “-bar.foo.com”.
+  IstioService destination = 2;
```
frankbu: since you are only using the name field, why not just have a string field instead of IstioService?

proxy/v1/config/egress_rule.proto
```diff
@@ -16,51 +16,50 @@ syntax = "proto3";
...
+// are allowed to access. A subset of routing rules and all destination policies can be applied
+// on the service targeted by an egress rule. The destination of an egress rule is allowed to
+// contain wildcards (e.g., *.foo.com). Currently, only HTTP-based services can be expressed
+// through the egress rule. If TLS origination from the sidecar is desired, the protocol
+// associated with the service port must be marked as HTTPS, and the service is expected to
+// be accessed over HTTP (e.g., http://gmail.com:443). The sidecar will automatically upgrade
+// the connection to TLS when initiating a connection with the external service.
 //
-// An example rule - allow traffic to *cnn.com and *cnn.it domains that are not part of the mesh.
+// For example, the following egress rule describes set of services hosted in the *.foo.com domain
```
frankbu: the set

PR #167, opened 2017-09-07
Update Mixer rule.selector to rule.match per design doc

mandarjog:
1. Update Mixer rule.selector to rule.match per design doc
2. Update comments.

### DIFF
--- a/mixer/v1/config/cfg.proto
+++ b/mixer/v1/config/cfg.proto
@@ -304,33 +304,29 @@ message EmailAddress {
 // instance constructed using the 'RequestCountByService' instance.
 //
 // ```yaml
-// actionRules:
-// - selector: destination.service == "*"
+// rule:
+// - match: destination.service == "*"
 //   actions:
 //   - handler: prometheus-handler
 //     instances:
 //     - RequestCountByService
 // ```
 message Rule {
-  // Required. Selector is an attribute based predicate. When Mixer receives a
-  // request it evaluates all selectors in scope and executes the `actions` for all
-  // selectors that evaluated to true.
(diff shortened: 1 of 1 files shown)


mixer/v1/config/cfg.proto
```diff
@@ -304,33 +304,29 @@ message EmailAddress {
 // instance constructed using the 'RequestCountByService' instance.
 //
 // ```yaml
-// actionRules:
-// - selector: destination.service == "*"
+// rule:
```
guptasu: this should be rules right since it is repeated ?

guptasu: Or every rule is different and pushed to config server as independent unit ?

guptasu: If that is the case then we need to remove this line completely and just start from 
match: ...
actions:

mixer/v1/config/cfg.proto
```diff
@@ -304,33 +304,29 @@ message EmailAddress {
...
-  string selector = 1;
+  string match = 1;
 
-  // Optional. The actions that will be executed when selector evaluates to `true`.
+  // Optional. The actions that will be executed when match evaluates to `true`.
   repeated Action actions = 2;
-
-  // Optional. Nested action rules; their selectors are evaluated if this selector
-  // predicate evaluates to `true`.
-  repeated Rule rules = 3;
```
guptasu: nice

mixer/v1/config/cfg.proto
mandarjog: ok I will remove that line.

PR #168, opened 2017-09-08
Separate individual static proxy variables from dynamic mesh variables

kyessenov:


proxy/v1/config/proxy_mesh.proto
```diff
@@ -134,8 +135,18 @@ message ProxyMeshConfig {
...
+  string zipkin_address = 12;
+
+  // File address for the proxy access log (e.g. /dev/stdout).
+  // Empty value disables access logging.
+  string access_log_file = 13;
 
-  // Path to the secrets used by the authentication policy.
-  string auth_certs_path = 101;
+  // Default proxy config for injected proxies
+  ProxyConfig default_config = 14;
```
rshriram: What's this for? Need more comments here

proxy/v1/config/proxy_mesh.proto
```diff
@@ -18,49 +18,15 @@ import "google/protobuf/duration.proto";
...
-
-  // Port on which Envoy should listen for administrative commands.
-  int32 proxy_admin_port = 6;
-
-  // Port on which Envoy should listen for HTTP PROXY requests if set.
-  int32 proxy_http_port = 16;
+// ProxyConfig defines variables for individual Envoy instances.
+message ProxyConfig {
+  // Path to the generated configuration file directory
+  string config_path = 1;
```
rshriram: Config path or config file or both , to allow custom proxy configs ?

proxy/v1/config/proxy_mesh.proto
```diff
@@ -74,33 +40,68 @@ message ProxyMeshConfig {
...
+  string zipkin_address = 8;
+
+  // Connection timeout used by Envoy for supporting services. (MUST BE >=1ms)
+  google.protobuf.Duration connect_timeout = 9;
+
+  // IP Address and Port of a statsd UDP listener (e.g. _10.75.241.127:9125_).
+  string statsd_udp_address = 10;
+
+  // Port on which Envoy should listen for administrative commands.
+  int32 proxy_admin_port = 11;
```
rshriram: How about we add xDS refresh interval (may be one per DS) and another config for inbound http port?

kyessenov: eds, lds, cds refresh rates separately? OK.
Beware these cannot be changed dynamically. Only rds refresh rate.

proxy/v1/config/proxy_mesh.proto
kyessenov: This is not adding support for config file (no need to go through pilot for that).

proxy/v1/config/proxy_mesh.proto
kyessenov: For kube-inject/initializer to apply settings to each new proxy.

proxy/v1/config/proxy_mesh.proto
```diff
@@ -134,8 +135,18 @@ message ProxyMeshConfig {
 
   // Authentication policy defines the global switch to control authentication
   // for Envoy-to-Envoy communication.
-  AuthPolicy auth_policy = 100;
+  AuthPolicy auth_policy = 10;
+
+  // Polling interval for RDS (MUST BE >=1ms)
+  google.protobuf.Duration rds_refresh_delay = 11;
```
ayj: How is this different than the xDS refresh delays in ProxyConfig?

kyessenov: For unknown reasons, RDS refresh rate is specified through LDS, so it's dynamically updatable.

proxy/v1/config/proxy_mesh.proto
```diff
@@ -74,33 +40,68 @@ message ProxyMeshConfig {
...
+  google.protobuf.Duration parent_shutdown_duration = 5;
+
+  // Address of the discovery service exposing xDS (e.g. _istio-pilot:8080_).
+  string discovery_address = 6;
+
+  // Polling interval for service discovery (EDS, CDS, LDS). (MUST BE >=1ms)
+  google.protobuf.Duration discovery_refresh_delay = 7;
+
+  // Address of the Zipkin service (e.g. _zipkin:9411_).
+  string zipkin_address = 8;
```
ayj: Why is zipkin_address duplicated in Proxy and Mesh config? Seems like ProxyConfig should only contain bootstrap config necessary to reach Pilot.

ayj:
The PR includes static vs. dynamic distinction. It would be good to document the same in the proto itself.

proxy/v1/config/proxy_mesh.proto
```diff
@@ -18,49 +18,15 @@ import "google/protobuf/duration.proto";
```
costinm: Since today is code freeze date - and this while good doesn't seem critical change - can you keep the old config
around and add code to migrate/copy to the new one ? (i.e. follow standard process for API or config changes - 
deprecate old, migrate to new, remove old in next release). It's likely we have tests or configs using old format, this
may result in lots of eng spending time.

proxy/v1/config/proxy_mesh.proto
kyessenov: For unknown reasons, parts of zipkin are static and parts are dynamic. This is the ugliness in xDS leaking to our proto.

proxy/v1/config/proxy_mesh.proto
kyessenov: There are only two places mesh configs are used - in pilot (fixed there in another PR) and in istio (which I'll have to fix).

proxy/v1/config/proxy_mesh.proto
```diff
@@ -18,49 +18,15 @@ import "google/protobuf/duration.proto";
...
+// ProxyConfig defines variables for individual Envoy instances.
+message ProxyConfig {
+  // Path to the generated configuration file directory
+  string config_path = 1;
 
-  // The time in seconds that Envoy will drain connections during a hot
-  // restart. MUST be >=1s (e.g., _1s/1m/1h_)
-  google.protobuf.Duration drain_duration = 7;
+  // Path to Envoy binary
+  string binary_path = 2;
```
costinm: Keep in mind for raw VM we use environment variables - cobra flags can be overriden by env, not sure if equivalent is possible for config files.

proxy/v1/config/proxy_mesh.proto
```diff
@@ -74,33 +40,68 @@ message ProxyMeshConfig {
...
+  // Polling interval for service discovery (EDS, CDS, LDS). (MUST BE >=1ms)
+  google.protobuf.Duration discovery_refresh_delay = 7;
+
+  // Address of the Zipkin service (e.g. _zipkin:9411_).
+  string zipkin_address = 8;
+
+  // Connection timeout used by Envoy for supporting services. (MUST BE >=1ms)
+  google.protobuf.Duration connect_timeout = 9;
+
+  // IP Address and Port of a statsd UDP listener (e.g. _10.75.241.127:9125_).
```
costinm: We want to switch to statsd TCP if possible, maybe 'statsd_address=udp:.... | tcp:....'  ?
( so we don't have to change config again ) ? 

Or we'll add a statsd_tcp_address later ? What is closest to envoy config ?

proxy/v1/config/proxy_mesh.proto
```diff
@@ -74,33 +40,68 @@ message ProxyMeshConfig {
...
+  string statsd_udp_address = 10;
+
+  // Port on which Envoy should listen for administrative commands.
+  int32 proxy_admin_port = 11;
+}
+
+
+// ProxyMeshConfig defines mesh-wide variables shared by all Envoy instances in the
+// Istio service mesh.
```
costinm: Would be good to add comments on where it is expected to live.

Does mixer and CA read this file ?

proxy/v1/config/proxy_mesh.proto
```diff
@@ -74,33 +40,68 @@ message ProxyMeshConfig {
...
+  // Disable policy checks by the mixer service. Metrics will still be
+  // reported to the mixer for HTTP requests and TCP connections. Default
+  // is false, i.e. mixer policy check is enabled by default.
+  bool disable_policy_checks = 3;
 
-  // File address for the proxy error and debug log (e.g. /dev/stderr)
-  string error_log_file = 15;
+  // Port on which Envoy should listen for incoming connections from
+  // other services.
+  int32 proxy_listen_port = 4;
```
costinm: Note that for each of those settings, we will likely need annotations/overrides in 0.3 - while in k8s ports are 'free', 
in raw VM some ports may be in use. 
That includes all ports, mixer address (VMs may need FQDN), and possibly auth policy. 

(maybe add it as TODO)

proxy/v1/config/proxy_mesh.proto
costinm: I guessed this - but please add it to the field comment. 
Can it be a different top level resource ? 
( i.e. the default proxy config will look like a normal proxy config )

proxy/v1/config/proxy_mesh.proto
kyessenov: `--config "binaryPath: $ENVOY"` works, no?

proxy/v1/config/proxy_mesh.proto
kyessenov: Need to follow up on the status of TCP support for statsd in envoy...

proxy/v1/config/proxy_mesh.proto
kyessenov: No, but they should. This will be another CRD, like a route rule.

kyessenov: No but they should. This is meant to be another core config unit, a CRD.

proxy/v1/config/proxy_mesh.proto
kyessenov: This is precisely the point of separating. We can provide overrides per proxy over mesh-wide configs.

kyessenov: Mixer is being configured remotely, not locally. So you need to define a "pod" notion for LDS to provide custom values.

proxy/v1/config/proxy_mesh.proto
rshriram: Actually ignore my comment. I found that we have a global refresh rate config since 0.1.6.

proxy/v1/config/proxy_mesh.proto
rshriram: It already supports tcp statsd. We are waiting on mixer updating its image - which depends on tcp statsd to prometheus exporter

rshriram:
Please fix the comment for config_path.

kyessenov:
I think I addressed most concerns. Let's merge this, and file separate issues to improve upon defaults, existing behavior. I have tried to limit the scope to static/dynamic distinction for now.

proxy/v1/config/proxy_mesh.proto
ldemailly: we should have made that change before merging as anyway istio/api isn't automatically used by anyone
(pilot is still using Aug 28th version)

PR #171, opened 2017-09-11
Rename bytes_total -> total_bytes

douglas-reid:
cc: @wora 

Prefer readability of "total bytes" to "bytes total", per api style guide
> Field names should also avoid using postpositive adjectives (modifiers placed after the noun)

### DIFF
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -181,9 +181,9 @@
 # connection attributes
 - connection.id
 - connection.received.bytes
-- connection.received.bytes_total
+- connection.received.total_bytes
 - connection.sent.bytes
-- connection.sent.bytes_total
+- connection.sent.total_bytes
 - connection.duration
 
 # context attributes


douglas-reid:
What backwards compatibility are we trying to maintain?  0.1?  I feel like a breaking change here is ok (connection.sent.*) attributes aren't used by anything yet, right?

We need to trim cruft from this file. There is a bunch of stuff that is going to be confusing.

qiwzhang:
We don't know which version of Mixer server connecting with which version of client.  By design, global dictionary can only be appended.  Not need to trim this global dictionary file.  There are just some indexes, not using much of memory.

kyessenov:
@costinm @ldemailly this has implications throughout Istio...

douglas-reid:
We don't care about supporting old versions of Mixer server or old versions of Mixer client, right?

There is a ton of confusion around whether or not to use attributes like:
- `context.timestamp` (vs. `context.time`)
- `target.namespace` (vs. `destination.namespace`)

If we can't correct mistakes for unused attributes now, this problem will only compound.

Mixer generates its word list from this file, no longer maintaining a static list. Mixerclient/proxy should as well. Old mixer servers do not need to be supported. I think we can say the same for old mixer clients. That would leave us with just needing to get mixer and proxy on the 0.2.X tag of istio/api before the 0.2.X release, right?

qiwzhang:
During development, there could be some small version mismatch between mixer client and server, (different SHA),  not about major version, such as 0.1 or 0.2.   If we just change global dictionary dramatically, then some problems  such as attribute A suddenly become attribute B, will be very difficult to debug.  For such reason, we insist that global dictionary should be always appended.

It may has some unused words there. But the problems it is causing are much smaller than ones caused by mismatched dictionary.

ZackButcher:
The only link between attributes and the global dictionary is the protocol between Mixerclient and Mixer itself: if the string is in the dictionary we don't put those literal bytes on to the wire (in this case the string is the attribute name), and if the string is not in the dictionary then we must send them. You can freely add attributes that do not exist in the dictionary to the system, you just pay for it in the form of extra bytes on the wire.

No human being should ever consume the word list. [We have a set of canonical attributes that lives here](https://github.com/istio/istio.github.io/blob/master/_docs/reference/config/mixer/attribute-vocabulary.md). It is totally unrelated to the global word list. There's no reason to remove strings from the list.

kyessenov:
Just add another item to the list, and obfuscate the existing slot.

ldemailly:
What's conclusion on this ? it's not changeable now that we released 0.2 is it ?

wora:
We should rename the attributes and add mixer hacks to auto translate old names to new names for fixed period of time, such as 3 or 6 months. Then we can remove the hacks.

We need to stay close to the style guide for long term consistency. One specific issue here is all names shall be prefix order, e.g collected_items, not items_collected, max_size, not size_max. We need to prevent mixed order of phrases.

We also need to build mixer client and server to support renaming, since the attribute name is only used to map name to slot, so renaming should be easy as long as we let multiple names map to the same slot.

geeknoid:
So what's up with this PR? Keep or kill?

douglas-reid:
I need to refactor it, but I think we need it (or something just like it). Adding it to my list.

wora:
It is desired by the style guide to avoid post adjective names. :)

geeknoid:
Doug, since you're fiddling with attributes right now, do you think it's a good time to knock this one out?

douglas-reid:
Sure, I'll update the PR and push.

douglas-reid:
Closing this in favor of the update in #514

PR #175, opened 2017-09-14
Use proper destination field type in DestinationWeight

rshriram:
From `string destination` to `IstioService destination`

### DIFF
--- a/proxy/v1/config/route_rule.proto
+++ b/proxy/v1/config/route_rule.proto
@@ -251,7 +251,7 @@ message MatchRequest {
 message DestinationWeight {
   // Optional destination uniquely identifies the destination service. If not
   // specified, the value is inherited from the parent route rule.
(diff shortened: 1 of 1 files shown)


sebastienvas:
Is this for 0.2 ?

sebastienvas:
Anyone making  a change to API, need to do the following update:
mixer
mixerclient -> proxy -> pilot
 
Note that pilot also direclty depends on api

kyessenov:
This change only affects pilot. I propose we break this repo apart if the requirement now is to have a single SHA across istio.

ldemailly:
the basic sanity requirement is indeed to have all internal repos releasing the same sha/version/tag

https://github.com/istio/istio/issues/724 should eventually document the complete list and ideally we'd have a simple script/tool to update it (or integrate it with the update deps bot)

PR #178, opened 2017-09-19
DON'T SUBMIT: update quota api. 

qiwzhang:
Please see design document at https://docs.google.com/document/d/1sHd5CDXg2bnfPoxlW7pmRB5BbvmwP1p2mIex8BFAxM4/edit#


qiwzhang:
This is for after 0.2.   Will not submit, here is for discussion.

qiwzhang:
Design changed.  It is not needed.

PR #179, opened 2017-09-20
Add Template Extension proto inside istio.io/api/mixer/v1/template

guptasu:
In order to support templates to be defined outside Mixer repo, we need to move out the TemplateExtension.proto from https://github.com/istio/mixer/blob/master/pkg/adapter/template/TemplateExtensions.proto into istio.io/api/mixer/v1/template. If the TemplateExtensions.proto is in Mixer repo, it was causing cyclic includes when trying to compile a Template outside Mixer repo that depends on Mixer.

There will be one following PR in the Mixer repo that changes all of the the existing defined templates to use the one from this new location. It will also delete the old TemplateExtensions.proto inside the Mixer repo.

### DIFF
--- a/mixer/v1/template/extensions.proto
+++ b/mixer/v1/template/extensions.proto
@@ -1,0 +1,19 @@
+syntax = "proto3";
+
+package istio.mixer.v1.config.template;
+import "google/protobuf/descriptor.proto";
+
+enum TemplateVariety {
+    TEMPLATE_VARIETY_CHECK = 0;
+    TEMPLATE_VARIETY_REPORT = 1;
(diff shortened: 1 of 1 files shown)


guptasu:
@rshriram Doc this will be updated https://docs.google.com/document/d/1rKPt2Z2acy4pRwcvScPa-Na-EnU-6poGR-Rj_fZtaIc/edit# . This document is under development and will be converted to markdowna and be part of Mixer repo.

guptasu:
@rshriram and @ZackButcher so, can we submit this PR ? All tests are passing and Google and IBM reps approve it.

PR #181, opened 2017-09-25
Add source.service to global dictionary

mandarjog:
Without this change, we get flooded with
```
0923 02:31:03.459056       1 protoBag.go:68] Attribute 'source.service' not in either global or message dictionaries
``` 
in Mixer logs.

### DIFF
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -218,3 +218,4 @@
 - destination.labels
 - destination.user
 
+- source.service


mandarjog:
@ldemailly

ldemailly:
@ZackButcher this fixes 1 error per request in the mixer logs

kyessenov:
How did an attribute get passed without either a local or a global dictionary?

qiwzhang:
if a string is not in global dictionary,  it will be added to per-message dictionary.  
Global dictionary is not for better optimization.

PR #185, opened 2017-10-03
Add quota runtime config.

qiwzhang:
This is just a proposal.

Please see detail design in:
https://docs.google.com/document/d/15g3iYbDwd6KHOGOq1vroo9cALXnrYxrs9Ilm7iDX1xM/edit?ts=59d12c03#

### DIFF
--- a/proxy/v1/config/quota.proto
+++ b/proxy/v1/config/quota.proto
@@ -30,13 +30,11 @@ package istio.proxy.v1.config;
 //         namespace: my-namespace # optional (default is metadata namespace field)
 //       config:
 //         - rules
-//           match:
-//             api.operation:
-//               exact: create_book
-//           quota: request_count
-//           amount: 1
+//           - quotas:
+//               quota: request_count
+//               amount: 1
 //
-// Example2, For HTTP POST requests and path are prefixed with "//books",
+// Example2, For HTTP POST requests and path are prefixed with "/books" or "/shelves"
 //    charge two quotas:
 //    1) write_count of 1 mount
 //    2) request_count of 5 amount.
@@ -46,35 +44,39 @@ package istio.proxy.v1.config;
 //         name: foo-service
 //         namespace: my-namespace # optional (default is metadata namespace field)
 //       config:
-//         - rules
-//           match:
-//             request.headers.path:
-//               prefix: //books
-//             request.http_method:
-//               exact: POST
-//           quota: write_count
-//           amount: 1
-//         - rules
-//           match:
-//             request.headers.path:
-//               prefix: //books
-//             request.http_method:
-//               exact: POST
-//           quota: request_count
-//           amount: 5
-//
+//         - rules:
+//           - match_or:
+//               match_and:
+//                 request.headers.path:
+//                   prefix: /books
+//                 request.http_method:
+//                   exact: POST
+//           - match_or:
+//               match_and:
+//                 request.headers.path:
+//                   prefix: /shelves
+//                 request.http_method:
+//                   exact: POST
+//           - quotas:
+//               quota: write_count
+//               amount: 1
+//           - quotas:
+//               quota: request_count
+//               amount: 5
 
-// Quota specifies quota runtime config stored in CRD.
-message Quota {
-  // REQUIRED: Destination uniquely identifies the destination service targeted
-  // for this quota rule.
+// QuotaSpec specifies quota runtime specification stored in CRD.
+message QuotaSpec {
+  // REQUIRED: Destination uniquely identifies the destination service
+  // targeted for this quota spec.
   // For Ingress proxy, it is the external DNS name for the mesh.
   // For side car porxy, it is the backend service of the proxy.
-  // Pilot uses this field to distribute the Quota config to Proxy.
+  // Pilot uses this field to identify which proxy to distribute the
+  // Quota config to.
+  // It is defined in route_rule.proto
   IstioService destination = 1;
 
   // Quota config.
-  QuotaConfig config = 2
+  QuotaConfig config = 2;
 }
 
 // Specifies quota config used by Proxy.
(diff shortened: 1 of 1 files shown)


proxy/v1/config/quota.proto
```diff
@@ -46,35 +44,39 @@ package istio.proxy.v1.config;
...
+//           - quotas:
+//               quota: request_count
+//               amount: 5
 
-// Quota specifies quota runtime config stored in CRD.
-message Quota {
-  // REQUIRED: Destination uniquely identifies the destination service targeted
-  // for this quota rule.
+// QuotaSpec specifies quota runtime specification stored in CRD.
+message QuotaSpec {
```
geeknoid: In the comments, generally don't talk about Pilot, Proxy or Mixer. This is configuration state, independent of where it is delivered and how it will be implemented.

proxy/v1/config/quota.proto
```diff
@@ -86,16 +88,28 @@ message QuotaConfig {
 
 // QuotaRule specifies a quota rule
 message QuotaRule {
+  // If empty, match all request.
+  // The final result is "OR" of all match_or results.
+  repeated MatchAnd match_or = 1;
```
geeknoid: Strange to have a field called match_or of type MatchAnd. Is that right? Can we improve this?

proxy/v1/config/quota.proto
```diff
@@ -46,35 +44,39 @@ package istio.proxy.v1.config;
...
-//           match:
-//             request.headers.path:
-//               prefix: //books
-//             request.http_method:
-//               exact: POST
-//           quota: request_count
-//           amount: 5
-//
+//         - rules:
+//           - match_or:
```
geeknoid: Seems a bit strange. Could we just have:

 match:
   clause:
     ...

So replace match_or with just "match" and replace "match_and" with "clause"?

qiwzhang: Done

proxy/v1/config/quota.proto
qiwzhang: Removed

proxy/v1/config/quota.proto
qiwzhang: Changed

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,113 @@
...
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "proxy/v1/config/route_rule.proto";
+
+package istio.proxy.v1.config;
+
+// Specifies a runtime quota rule.
+//  * Use Istio attributes to match match requests
```
geeknoid: match match -> match

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,113 @@
...
+//         namespace: my-namespace # optional (default is metadata namespace field)
+//       config:
+//         - rules
+//           - quotas:
+//               quota: request_count
+//               amount: 1
+//
+// Example2, For HTTP POST requests and path are prefixed with "/books" or "/shelves"
+//    charge two quotas:
+//    1) write_count of 1 mount
```
geeknoid: mount -> amount

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,113 @@
...
+//               quota: request_count
+//               amount: 5
+
+// QuotaSpec specifies quota runtime specification stored in CRD.
+message QuotaSpec {
+  // REQUIRED: Destination uniquely identifies the destination service
+  // targeted for this quota spec.
+  // For Ingress proxy, it is the external DNS name for the mesh.
+  // For side car porxy, it is the backend service of the proxy.
+  // It is defined in route_rule.proto
```
geeknoid: No need to state where this is defined.

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,113 @@
...
+//           - quotas:
+//               quota: request_count
+//               amount: 5
+
+// QuotaSpec specifies quota runtime specification stored in CRD.
+message QuotaSpec {
+  // REQUIRED: Destination uniquely identifies the destination service
+  // targeted for this quota spec.
+  // For Ingress proxy, it is the external DNS name for the mesh.
+  // For side car porxy, it is the backend service of the proxy.
```
geeknoid: side car -> sidecar

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,113 @@
...
+  // It is defined in route_rule.proto
+  IstioService destination = 1;
+
+  // Quota config.
+  QuotaConfig config = 2;
+}
+
+// Specifies quota config used by Proxy.
```
geeknoid: Don't mention this is for proxy or "Envoy Mixer config". Also, you want to avoid saying effectively "QuotaConfig is the config for quota". How about:

Determines the quota used for individual requests to a service in the mesh.

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,113 @@
...
+// This message will be send to Proxy in Envoy Mixer config.
+message QuotaConfig {
+  // A list of Quota rules.
+  repeated QuotaRule rules = 1;
+}
+
+// QuotaRule specifies a quota rule
```
geeknoid: Can we say something not self-referential?

proxy/v1/config/quota.proto
qiwzhang: Done

proxy/v1/config/quota.proto
qiwzhang: Done

proxy/v1/config/quota.proto
qiwzhang: Done

proxy/v1/config/quota.proto
qiwzhang: Done

proxy/v1/config/quota.proto
qiwzhang: Done

proxy/v1/config/quota.proto
qiwzhang: Done

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// Determines the quotas used for individual requests.
+message QuotaConfig {
+  // A list of Quota rules.
+  repeated QuotaRule rules = 1;
+}
+
+// Specifies a rule with list of matches and list of quotas
+// If any of matches is satisfied, the list of quotas will be used.
```
geeknoid: If any clause match, the list...

PR #187, opened 2017-10-04
Add service IDL proto.

qiwzhang:
istioctl will convert OpenAPI spec to this format.  
The data in this format will be passed to Proxy to generate API attributes at runtime.


kyessenov:
cc @linsun for release approval

qiwzhang:
@kyessenov this is not for 0.2.  Do we still need release approval?

linsun:
maybe hold off merging till 0.2 is shipped (next week)?

proxy/v1/config/service_idl.proto
```diff
@@ -0,0 +1,75 @@
...
+// Determines the quotas used for individual requests.
+message ServiceIDL {
+  // service name, for api.service attribute
+  // For OpenAPI, it is from "title" inside "Info" object.
+  string name = 1;
+
+  // service version, for api.version attribute
+  // For OpenAPI, it is from "version" inside "Info" object.
+  string version = 2;
+
```
ayj: Can't a service implement more than one API? It seems like we'd want to group API attributes with the match condition. 

Also, to support general HTTP use case beyond OpenAPI it seems like we'd want to further generalize the pattern to support custom methods, prefix/regex matching, etc. against more than just the path.

For example,

```protobuf
message PathPattern {
    // attributes that should be generated when all match condition are true
    map<string, string> attributes = 1; 

    // List of match conditions against HTTP headers. Pseudo-Headers are used for method, path, etc.
     istio.proxy.v1.config.MatchRequest match_request = 2;
}

message ServiceIDL {
   IstioService destination = 1;
   repeated PathPattern patterns = 2;
}
```

rshriram:
What exactly does this do? and why shoudl this be in proxy/v1/config ?

rshriram:
In other words, the openAPI spec - how do you expect end users to supply this info? and how does it get to Envoys? How will this be surfaced in user interaction?

qiwzhang:
@linsun this will merge into api-mgmt-poc branch.  will not into master.

proxy/v1/config/service_idl.proto
```diff
@@ -0,0 +1,75 @@
...
+  // For OpenAPI, it is from "version" inside "Info" object.
+  string version = 2;
+  
+  // A list of Path rules.
+  repeated PathPattern paths = 3;
+}
+
+// Specifies a URL path pattern to map to an API operation.
+message PathPattern {
+  // If the path pattern is matched, the api.operation value.
```
geeknoid: , this will be the value used for the api.operation attribute.

proxy/v1/config/service_idl.proto
qiwzhang: @ayj Thanks for the input.   

I think the speed of lookup is very important since Proxy is in the data path.  We try to achieve O(1) instead of O(n) if we have n matches.

In my design,  basically map<url_path,  info> is used to achieve O(1).  This approach was used by Google cloud endpoints.

qiwzhang:
@ayj Updated to support multiple APIs.  PTAL

ayj:
This should be sufficient to bootstrap some API management features as a proof of concept. Its tempting to generalize this IDL to support exact/prefix/regex-style conditions per HTTP header, but that level of complexity may not be needed. I'll document both options in [Istio API spec and attribute generation](goo.gl/aw9M1w) where we can collect more community feedback.

PR #188, opened 2017-10-10
Add attribute proto.

qiwzhang:
This one is for Proxy config.  Pilot will use it to set mesh attributes to Istio Proxy.

### DIFF
--- a/proxy/v1/config/attributes.proto
+++ b/proxy/v1/config/attributes.proto
@@ -1,0 +1,61 @@
+// Copyright 2017 Istio Authors
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+package istio.proxy.v1.config;
+
+// Specifies Istio Mixer attributes for Envoy Mixer filter.
+// Following places will use this messages
+//  * Istio Pilot to set mesh attributes to Proxy.
+//  * Client Proxy to pass attributes to server Proxy.
+//
+
+// Specifies attributes to config Istio Proxy.
+// It represents the same data as ::istio:;mixer::v1::attributes defined in
+// mixer/v1/attributes.proto but in different formats.
+// ::istio:;mixer::v1::attributes was defined for transport optimizated.
+// It uses global dictionary and per-message dictionary to compress the data
+// for data transport.
+// This one is targeted for Proxy configuration, not need to compress data,
+// it should be easy to use.
+message Attributes {
+  // A list of attributes.
+  repeated Attribute attributes = 1;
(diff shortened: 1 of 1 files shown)


proxy/v1/config/attributes.proto
```diff
@@ -0,0 +1,61 @@
...
+
+package istio.proxy.v1.config;
+
+// Specifies Istio Mixer attributes for Envoy Mixer filter.
+// Following places will use this messages
+//  * Istio Pilot to set mesh attributes to Proxy.
+//  * Client Proxy to pass attributes to server Proxy.
+//
+
+// Specifies attributes to config Istio Proxy.
```
geeknoid: I think we can cut down on this comment somewhat, no need for all the back story. How about just:

Defines attributes for use in proxy configuration.

proxy/v1/config/attributes.proto
```diff
@@ -0,0 +1,61 @@
...
+
+  oneof value {
+    string string_value = 2;
+    int64 int64_value = 3;
+    double double_value = 4;
+    bool bool_value = 5;
+    bytes bytes_value = 6;
+    google.protobuf.Timestamp timestamp_value = 7;
+    google.protobuf.Duration duration_value = 8;
+    StringMap string_map_value = 9;
```
kyessenov: you could probably avoid embedding a map and use `map<string, string>` directly here?

proxy/v1/config/attributes.proto
```diff
@@ -0,0 +1,61 @@
...
+// Specifies Istio Mixer attributes for Envoy Mixer filter.
+// Following places will use this messages
+//  * Istio Pilot to set mesh attributes to Proxy.
+//  * Client Proxy to pass attributes to server Proxy.
+//
+
+// Specifies attributes to config Istio Proxy.
+// It represents the same data as ::istio:;mixer::v1::attributes defined in
+// mixer/v1/attributes.proto but in different formats.
+// ::istio:;mixer::v1::attributes was defined for transport optimizated.
```
kyessenov: please correct `:;` in the text. _optimization_

proxy/v1/config/attributes.proto
```diff
@@ -0,0 +1,61 @@
...
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+package istio.proxy.v1.config;
+
+// Specifies Istio Mixer attributes for Envoy Mixer filter.
+// Following places will use this messages
+//  * Istio Pilot to set mesh attributes to Proxy.
+//  * Client Proxy to pass attributes to server Proxy.
```
ayj: Wouldn't we want to use the compressed version over-the-wire between client and server proxy?

kyessenov: That would require negotiation about _largest known global word index_.
Local context is also difficult to achieve since requests can be balanced to different servers.

qiwzhang: Yes, it will be in the HTTP header as:

x-istio-attributes: base64_encoded(proto_seriazlied(Attributes))

proxy/v1/config/attributes.proto
qiwzhang: Tried.  I got this error:
Map fields are not allowed in oneofs

kyessenov: Ok, fair enough. Must be a limitation of protobuf.

kyessenov:
I just realized this is for a different branch. What's the motivation for branching Istio? I think this PR is generally useful.

ayj:
We created a temp branch to unblock some iterative development during the 0.2.x code freeze. This and https://github.com/istio/api/pull/190 are not specific to API mgmt and could be upstreamed soon-ish. Hopefully the IDL specific parts can be merged shortly after that.

PR #190, opened 2017-10-10
Added mixer_filter_config

qiwzhang:
It will be replacing

https://github.com/istio/pilot/blob/master/proxy/envoy/mixer.go#L63

### DIFF
--- a/proxy/v1/config/mixer_filter_config.proto
+++ b/proxy/v1/config/mixer_filter_config.proto
@@ -20,32 +20,54 @@ import "proxy/v1/config/service_idl.proto";
 
 package istio.proxy.v1.config;
 
-// Specifies config for Envoy mixer filter.
-message MixerFilterConfig {
-  // Send these attributes to Mixer.
-  Attributes mixer_attributes = 1;
+// Defines mixer control config.
+// It can be defined Envoy route opaque config.
+// This is a per-route config. Defferent upstream routes can have different
+// config.
+// This will be distributed by Envoy RDS.
+message MixerControlConfig {
+  // The flag to call Mixer Check
+  bool call_mixer_check = 1;
+  // The flag to call Mixer Report
+  bool call_mixer_Report = 2;
+  // The flag to forward attributes to upstream.
+  bool enable_forward_attributes = 3;
+
+  // Send these attributes to Mixer in both Check and Report.
+  Attributes mixer_attributes = 4;
 
   // Forward these attributes to upstream.
-  Attributes forward_attributes = 2;
+  Attributes forward_attributes = 5;
 
+  // Service IDL to generate API attributes.
+  ServiceIDL service_idl = 6;
+
+  // Quota config
+  QuotaConfig quota_config = 7;
+}
+
+// Specifies config for Envoy mixer filter.
+// This is a global config for each Mixer filter.
+// This will be distributed by Envoy LDS.
+message MixerFilterConfig {
   // The flag to disable check cache.
-  bool disable_check_cache = 3;
+  bool disable_check_cache = 1;
   // The flag to disable quota cache.
-  bool disable_quota_cache = 4;
+  bool disable_quota_cache = 2;
   // The flag to disable report batch.
-  bool disable_report_batch = 5;
+  bool disable_report_batch = 3;
 
   // Specifies the policy when failed to connect to Mixer server.
   // valid values are: [open|close]
-  string network_fail_policy = 6;
+  string network_fail_policy = 4;
 
   // The flag to disable check/quota calls for TCP filter.
-  bool disable_tcp_check_calls = 7;
+  bool disable_tcp_check_calls = 5;
 
-  // Service IDL to generate API attributes.
-  ServiceIDL service_idl = 8;
-
-  // Quota config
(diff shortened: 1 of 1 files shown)


proxy/v1/config/mixer_filter_config.proto
```diff
@@ -0,0 +1,51 @@
...
+
+  // Specifies the policy when failed to connect to Mixer server.
+  // valid values are: [open|close]
+  string network_fail_policy = 6;
+
+  // The flag to disable check/quota calls for TCP filter.
+  bool disable_tcp_check_calls = 7;
+
+  // Service IDL to generate API attributes.
+  ServiceIDL service_idl = 8;
```
kyessenov: why is IDL here?

qiwzhang: It is defined here
https://github.com/istio/api/blob/api-mgmt-poc/proxy/v1/config/service_idl.proto#L116

To generate api attributes

kyessenov: Are you planning to branch pilot as well?

ayj: I've got a private branch that integrates some of these changes. Now that master is unfrozen I should be able to push development their and avoid unnecessary branches.

re: `ServiceIDL`, I think this needs to be repeated to support multiple APIs from the same server proxy. Even though its perhaps a corner case, it is still possible for a pod (proxy) to implement multiple services.

qiwzhang: If service_idl is a "repeated" field, how does Proxy choose which service_idl to use?  
If it uses path prefix, it assumes different services should have different path prefixes.  In that case, they can be merged into one service_idl.

ayj: Sorry, I missed that you used `ServiceIDL` and not `ServiceIDLSpec`, so the host isn't known. We'd need another message that wraps ServiceIDL with host (or address:port equivalent to LDS) that can be repeated. Proxy would match request against the small list of hosts (where size is typically 0 or 1) before path matching. This would support the corner case of a user wanting to implement multiple APIs/services with the same pod. We don't need to implement support for this in the proxy (yet), but it would be hard to update the API in the future without breaking backwards compat to support the use case.

proxy/v1/config/mixer_filter_config.proto
```diff
@@ -0,0 +1,73 @@
...
+
+import "proxy/v1/config/attributes.proto";
+import "proxy/v1/config/quota.proto";
+import "proxy/v1/config/service_idl.proto";
+
+package istio.proxy.v1.config;
+
+// Defines mixer control config.
+// It can be defined Envoy route opaque config.
+// This is a per-route config. Defferent upstream routes can have different
```
ayj: Does envoy provide a persistent hook for per-route opaque config? As far as I could tell, filters could only use the immediate value via `route_entry` which would have perf implications for per-request deserialization/decode of protobuf fields.

proxy/v1/config/mixer_filter_config.proto
```diff
@@ -0,0 +1,73 @@
...
+  bool enable_forward_attributes = 3;
+
+  // Send these attributes to Mixer in both Check and Report.
+  Attributes mixer_attributes = 4;
+
+  // Forward these attributes to upstream.
+  Attributes forward_attributes = 5;
+
+  // Service IDL to generate API attributes.
+  ServiceIDL service_idl = 6;
```
ayj: @kyessenov, do you know the status of service association w.r.t pods supporting multiple services? Per https://istio.io/docs/setup/kubernetes/sidecar-injection.html that is not supported _as of now_. Is that a temporary limitation related to source attribute derivation, or another reason?

proxy/v1/config/mixer_filter_config.proto
qiwzhang: Mixer http filter needs to do some caching.  One idea of caching is to use route_entry pointer as cache key.  The problem is how to collect garbage; which cache items are not used.

kyessenov: We discussed this yesterday:
1. We're moving to xDS v2 which now has `metadata` on all xDS responses instead of an opaque config.
2. The new xDS has versions attached to responses.
3. Attaching HTTP config to LDS causes unnecessary re-initialization of LDS (draining and replacing sockets). We gain perf improvement in pre-processing per-route config, but lose in causing LDS churn.

proxy/v1/config/mixer_filter_config.proto
kyessenov: It's a P1 TODO item to handle multi-service pods properly with respect to routing and source workload derivation. Envoy sidecar has some tech debt with assigning a sidecar to a unique service cluster, that is causing some problems.

ayj:
> what is the benefit of externalizing the mixer filter's options?

The interface between Pilot and Proxy should be standardized. It mostly is via Envoy APIs. The mixer specific config,  up until this point, was not. 

> It also has reference to Attributes, ServiceIDL and QuotaConfig which are not defined.

This PR is for `api-mgmt-poc` branch. The dependent messages were also introduced into this branch in previous PRs.

PR #192, opened 2017-10-11
Proposal: add ingress exposure as part of the routing rule

kyessenov:
follow up from @rshriram, @ijsnellf discussion

### DIFF
--- a/proxy/v1/config/route_rule.proto
+++ b/proxy/v1/config/route_rule.proto
@@ -134,6 +134,27 @@ message RouteRule {
 
   //(-- L4 fault injection policy applies to Tcp/Udp (not HTTP) traffic --)
   L4FaultInjection l4_fault = 11;
+  
+  // Ingress specification exposes a service externally to the cluster
+  Ingress ingress = 12;
+}
+
+// Ingress specification for exposing a destination port externally
+message Ingress {
+  // REQUIRED: Port to expose the destination service on.
+  int32 port = 1;
+  
+  // Optional TLS secret path to apply on the exposed port.
+  string tls_secret = 2;
+
+  // REQUIRED: Destination port identifies a port on the destination service for routing.
+  oneof destination_service_port {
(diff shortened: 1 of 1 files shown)


proxy/v1/config/route_rule.proto
```diff
@@ -134,6 +134,28 @@ message RouteRule {
...
+}
+
+// Ingress specification for exposing a destination port externally
+message Ingress {
+  // REQUIRED: Port to expose the destination service on.
+  int32 port = 1;
+  
+  // Optional TLS secret path to apply on the exposed port.
+  string tls_secret = 2;
+
```
rshriram: might want to add client_secret and server_secret as well, to support mTLS on ingress.

frankbu:
@rshriram does the ingress rule have a match (patch/host) clause in your proposal, or do we still need to create a rule to block unwanted paths and rules to allow selected paths?

kyessenov:
The port mapping is important for cloud provider LBs. They typically don't allow exposing arbitrary ports on the cluster. Without port translation in route rules, it would be impossible to expose a non-default port.

rshriram:
Yes, but my point is that this port mapping does not have to be part of the route rule spec.. i.e., if service wants to listen on port 80 because the provider allows port 80, then all is fine. If service insists on port 9080, then the user has to find another way of mapping.

Alternatively, you can imagine specifying something like this:

```yaml
tlsMappings:
- host:
    name: foo.bar.com
    serverSecret: baz
    clientSecret: bar
...
applyRules:
- port: 80
   protocol: http (includes gRPC/H2)
   ruleLabels:
      foo: bar
- port: 443
   protocol: https
  ruleLabels:
     foo: baz
- port: 9379
   protocol: tcp
   service: myTCPservice
```

rshriram:
The above is no different from the HTTP proxy configuration

rshriram:
closing in favor of #254 and #243

PR #193, opened 2017-10-11
Build set-up for api

kyessenov:
Self-contained API repo with build files for generating go/gogo proto sources.
The clients of this repo are expected to do the following:
```
git_repository(
    name = "io_istio_api",
    commit = ISTIO_API_SHA,
    remote = "https://github.com/istio/api.git",
)
load("@io_istio_api//:api.bzl", "go_istio_api_dependencies")
go_istio_api_dependencies()
```
For local development against an API change:
```
bazel build //... --override_repository=io_istio_api=$(pwd)/../api
```


kyessenov:
PTAL: rewrote googleapis protos to `istio.io/api/google/rpc` for easier import of gogoslick dependency.

mixer/v1/BUILD
```diff
@@ -0,0 +1,51 @@
+load("@org_pubref_rules_protobuf//gogo:rules.bzl", "gogoslick_proto_library")
+
+gogoslick_proto_library(
+    name = "go_default_library",
+    importmap = {
+        "gogoproto/gogo.proto": "github.com/gogo/protobuf/gogoproto",
+        "google/rpc/status.proto": "istio.io/api/google/rpc",
```
douglas-reid: we were previously mapping this to: `"github.com/googleapis/googleapis/google/rpc"`, iirc.  I don't know that the change is that big of a deal, but it might be worth adding a comment / explanation somewhere.

mixer/v1/config/descriptor/BUILD
```diff
@@ -0,0 +1,31 @@
...
+    with_grpc = False,
+    deps = [
+        "@com_github_gogo_protobuf//sortkeys:go_default_library",
+        "@com_github_gogo_protobuf//types:go_default_library",
+    ],
+)
+
+filegroup(
+    name = "protos",
+    srcs = glob(["*.proto"]),
```
douglas-reid: This probably OK for now, but I generally prefer avoiding `glob`s where possible. We want to dramatically trim the set of files in this dir anyways, so we can revisit this soon.

mixer/v1/BUILD
kyessenov: I think it's important to highlight the gogoslicked version of these protos, rather than standard `googleapis/googleapis` generated ones. There's also `google/go-genproto` version of these, and they don't compile against the generated protos in `istio/api`.

mixer/v1/config/descriptor/BUILD
kyessenov: Ok, I'll enumerate protos.

qiwzhang:
@kyessenov this breaks Istio/proxy.

You could not put BUILD file in sub-folders, such as mixer/v1.   

You can only put BUILD file in the top level.   Istio/proxy can replace it with its own BUILD file.

Please revert this change

kyessenov:
Can you move proxy definitions to API repo instead? We need a way to share protobuf/go rules and this is the right place for them to exist.

qiwzhang:
Then we need to use same BUILD file for go and c++ proto generation
We tried it long time ago.  It is not easy to have same BUILD file for both go and c++.

kyessenov:
We need consistency in dependencies and go/proto rules instead of copying BUILD files from place to place. How do I reproduce the failure?

qiwzhang:
How could we submit something which breaks existing build?

kyessenov:
There has been no tests whatsoever here, until I established the build.

qiwzhang:
This change should be reverted.  The build is broken

kyessenov:
By the way, envoy has the same build file for python and cc. I don't buy that it's impossible.

kyessenov:
cc @lizan can you take a look and tell me how do I reproduce the failure? Is this just mixerclient?

qiwzhang:
Yes, if if update mixer client with this SHA.  it will not break.

PR #195, opened 2017-10-11
API changes for securing istio components

rkpagadala:
These changes are needed to pass parameters to pilot agent at startup

### DIFF
--- a/proxy/v1/config/proxy_mesh.proto
+++ b/proxy/v1/config/proxy_mesh.proto
@@ -73,6 +73,10 @@ message ProxyConfig {
   // Envoy as a sidecar in Kubernetes, this flag must be one of the availability
   // zones assigned to a node using failure-domain.beta.kubernetes.io/zone annotation.
   string availability_zone = 12;
+
+  // Authentication policy defines the global switch to control authentication
+  // for Envoy-to-Envoy communication for infra components Mixer and Pilot.
+  AuthPolicy infra_auth_policy = 13;
 }
 
 
--- a/proxy/v1/config/proxy_mesh.proto
+++ b/proxy/v1/config/proxy_mesh.proto
@@ -86,11 +86,11 @@ message ProxyConfig {
   // for Envoy-to-Envoy communication for infra components Mixer and Pilot.
   AuthPolicy infra_auth_policy = 13;
 
-  // Address of the mixer service (e.g. _istio-mixer:9090_).
+  // Address of the mixer service (e.g. _localhost:9090_).
   // Used by the proxy in front of mixer
   string mixer_address = 14;
 
-  // Pilot subject alternate name used for mutula TLS
+  // Pilot subject alternate name used for mutual TLS
   repeated string pilot_san = 15;
 }
 
--- a/proxy/v1/config/proxy_mesh.proto
+++ b/proxy/v1/config/proxy_mesh.proto
@@ -85,6 +85,10 @@ message ProxyConfig {
   // Authentication policy defines the global switch to control authentication
   // for Envoy-to-Envoy communication for infra components Mixer and Pilot.
   AuthPolicy infra_auth_policy = 13;
+
+  // Address of the mixer service (e.g. _istio-mixer:9090_).
+  // Used by the proxy in front of mixer
+  string mixer_address = 14;
 }
 
 
--- a/proxy/v1/config/proxy_mesh.proto
+++ b/proxy/v1/config/proxy_mesh.proto
@@ -91,7 +91,7 @@ message ProxyConfig {
   string mixer_address = 14;
 
   // Pilot subject alternate name used for mutula TLS
-  repeated string pilot_san = 14;
+  repeated string pilot_san = 15;
 }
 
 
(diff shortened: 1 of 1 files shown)


proxy/v1/config/proxy_mesh.proto
```diff
@@ -73,6 +81,17 @@ message ProxyConfig {
...
   // zones assigned to a node using failure-domain.beta.kubernetes.io/zone annotation.
   string availability_zone = 12;
+
+  // Authentication policy defines the global switch to control authentication
+  // for Envoy-to-Envoy communication for infra components Mixer and Pilot.
+  AuthPolicy infra_auth_policy = 13;
+
+  // Address of the mixer service (e.g. _istio-mixer:9090_).
+  // Used by the proxy in front of mixer
+  string mixer_address = 14;
```
kyessenov: isn't it localhost for the proxy in front of mixer?

proxy/v1/config/proxy_mesh.proto
```diff
@@ -73,6 +81,17 @@ message ProxyConfig {
...
+
+  // Pilot subject alternate name used for mutula TLS
```
kyessenov: _mutual_

proxy/v1/config/proxy_mesh.proto
```diff
@@ -18,6 +18,14 @@ import "google/protobuf/duration.proto";
 
 package istio.proxy.v1.config;
 
+enum AuthPolicy {
```
rshriram: AuthenticationPolicy

proxy/v1/config/proxy_mesh.proto
```diff
@@ -73,6 +81,18 @@ message ProxyConfig {
   // Envoy as a sidecar in Kubernetes, this flag must be one of the availability
   // zones assigned to a node using failure-domain.beta.kubernetes.io/zone annotation.
   string availability_zone = 12;
+
+  // Authentication policy defines the global switch to control authentication
+  // for Envoy-to-Envoy communication for infra components Mixer and Pilot.
```
rshriram: what the heck is "infra" ? Please remember that this is what ends up coming out in public documents. So lets avoid internal lingos.
```Authentication policy governing communication between data plane (sidecar) and control plane components (E.g., Pilot, Mixer)```

File: proxy/v1/config/proxy_mesh.proto [hunk 5]
```diff
@@ -73,6 +81,18 @@ message ProxyConfig {
...
+  AuthPolicy infra_auth_policy = 13;
+
+  // Address of the mixer service (e.g. __istio-mixer:15004_).
+  // Used by the proxy in front of mixer as the port to listen to,
+  // The value must be the same as MeshConfig.mixer_address
+  string mixer_address = 14;
```
rshriram: If its the same, why do we need a separate field in the first place? Why not just read it directly from the mesh config? Otherwise, this is going to be very confusing for folks manually adding the proxy configuration to their pods. If this is for internal purposes, this should not be present here.

proxy/v1/config/proxy_mesh.proto
```diff
@@ -73,6 +81,18 @@ message ProxyConfig {
   // Envoy as a sidecar in Kubernetes, this flag must be one of the availability
   // zones assigned to a node using failure-domain.beta.kubernetes.io/zone annotation.
   string availability_zone = 12;
+
+  // Authentication policy defines the global switch to control authentication
+  // for Envoy-to-Envoy communication for infra components Mixer and Pilot.
+  AuthPolicy infra_auth_policy = 13;
```
rshriram: please rename this infra_auth... to something sensible. control_plane_auth_policy

proxy/v1/config/proxy_mesh.proto
```diff
@@ -73,6 +81,18 @@ message ProxyConfig {
...
+
+  // Address of the mixer service (e.g. __istio-mixer:15004_).
+  // Used by the proxy in front of mixer as the port to listen to,
+  // The value must be the same as MeshConfig.mixer_address
+  string mixer_address = 14;
+
+  // Pilot subject alternate name used for mutual TLS
+  repeated string pilot_san = 15;
```
rshriram: this could use more explanation. Also, why is this an array? Secondly, have you considered how CI/CD systems that want to manually inject the sidecar templates (or companies that have a policy to have all kube config in GIT (ergo, you need to have the app+sidecar in one YAML in git) will encode these SANs etc. ?

rkpagadala:
Hi Shriram,
    Will rethink and send another PR to fix.
Thanks
Krishna

proxy/v1/config/proxy_mesh.proto
costinm: Made a comment on the other PR - this should be named "allow insecure", having pilot or mixer unsecured should be very clearly explained. 

For services having auth policy = none will be used to indicate the service does its own TLS. 

I assume for 0.3 'production ready' we can't ship with pilot/mixer not using HTTPS.

proxy/v1/config/proxy_mesh.proto
costinm: +1 - SAN is a specific term that few people know. 
We may as well use JWT tokens - where there is no SAN. This is badly named

PR #197, opened 2017-10-11
Add attributes proto for proxy config.

qiwzhang:


proxy/v1/config/attributes.proto
```diff
@@ -0,0 +1,55 @@
...
+
+package istio.proxy.v1.config;
+
+// Specifies Istio Mixer attributes for Envoy Mixer filter.
+// Following places will use this messages
+//  * Istio Pilot to set mesh attributes to Proxy.
+//  * Client Proxy to pass attributes to server Proxy.
+//
+
+// Defines attributes for use in proxy configuration.
```
ayj: A bit of context may help explain why these attributes are needed by proxy and why it's different from istio.mixer.v1.Attributes, e.g.

"The proxy produces attributes using information from traffic
requests and connections (e.g. HTTP). Some attributes are 
configured per-proxy as part of the proxy configuration, 
e.g. source.uid, destination.uid. Attributes may be 
forwarded  from the client to server proxy.

Attributes is a list of attributes. It is suitable for
configuring the proxy with sets of attributes to be included in
each request to the mixer and forwarded to the server proxy. It
differs from other [attribute][istio.mixer.v1.Attributes]
representations in that it is stateless and does not require
maintaining a global or per-connection dictionary of words."

proxy/v1/config/attributes.proto
```diff
@@ -0,0 +1,55 @@
...
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/timestamp.proto";
+
+package istio.proxy.v1.config;
+
+// Specifies Istio Mixer attributes for Envoy Mixer filter.
```
ayj: https://github.com/istio/api/blob/master/mixer/v1/attributes.proto already provides an explanation of attributes. That said, it still might be nice to have a quick blurb about what an attribute is for those browsing generating docs without needed to refer to a similarly named but slightly different file with its own attribute definition.

"Attributes represents a set of typed name/value pairs. An Istio
attribute carries a specific piece of information such as the
error code of an API request, the latency of an API request, or
the original IP address of a TCP connection. For example:

 ```
 request.path: xyz/abc
 request.size: 234
 request.time: 12:34:56.789 04/17/2017
 source.ip: 192.168.0.1
 target.service: example
 ```

The common baseline set of attributes available in most Istio
deployments is defined [here](https://istio.io/docs/reference/config/mixer/attribute-vocabulary.html)."

kyessenov: Can we have on definition imported from mixer and not have two definitions for attributes?

ayj:
Are you referring to the docs or the messege itself?

kyessenov:
Message itself. I understand that mixer definition is transport-optimized. So we need another definition in mixer proper to hold non-optimized attribute proto.

proxy/v1/config/attributes.proto
rshriram: Yes please. why not use the one from mixer? This seems like duplication. Also, there is no restriction that all proxy specific stuff should be under proxy/v1/config. proxy is bad name (should have been routing). Mixer protos are still used by Envoy (especially the check stuff).

qiwzhang:
I think Mixer doesn't need a proper attributes defined in proto format.   Probably proxy config is the first place to define a proper attributes proto. @geeknoid

proxy/v1/config/attributes.proto
ayj: @rshriram - I asked this question earlier. The answer is that the mixer version of `Attributes` is optimized through the use of global and per-connection dictionaries. The former requires negotiating the dictionary version. The latter is stateful. Apply the same mechanism for proxy-to-proxy requires solving the same problems. Also, we'd like to use the non-optimized version of `Attributes` for a subsequent PR which introduces a new IDL configuration for generating API related attributes based on path, etc. We would like to use the unoptimized version of `Attributes` for this purpose.

> Can we have on definition imported from mixer and not have two definitions for attributes?

Moving this and some of the other forthcoming mixer related proxy related protos under mixer/v1 sounds reasonable. We'll need another name for the types to avoid conflict with the optimized version. `UncompressedAttribute`, `ExpandedAttribute`, `ProxyAttribute`?

qiwzhang:
We can move this into mixer/v1/attributes.proto.   Rename the one over
there as CompressedAttributes.   It may break the build,  but we can the
build breakage.

-Wayne

rshriram:
AttributesUnOpt

geeknoid:
This is not a Mixer definition, it doesn't belong with Mixer stuff.

qiwzhang:
PTAL
Call it "UnCompressedAttributes" and moved into mixer/v1/attributes.proto.

ayj:
> Yes please. why not use the one from mixer? This seems like duplication. Also, there is no restriction that all proxy specific stuff should be under proxy/v1/config. proxy is bad name (should have been routing). Mixer protos are still used by Envoy (especially the check stuff

> This is not a Mixer definition, it doesn't belong with Mixer stuff.

Organizing config by component breaks down when config spans both components, e.g. mixer_filter_config for proxy.

ayj:
Neither location is ideal given our component-level organization of config and the fact that proxy_mixer_filter_config logically spans both components. Can we follow the existing conventions of putting configuration used by the proxy (and distributed by pilot) under `v1/proxy/`?

mixer/v1/attributes.proto
```diff
@@ -54,6 +54,36 @@ option (gogoproto.gostring_all) = false;
 // Each type of value is encoded into one of the so-called transport types present
 // in this message.
 //
+// Defines a list of attributes in uncompressed format.
+message UnCompressedAttributes {
```
ayj: s/UnCompressedAttributes/UncompressedAttributes

qiwzhang: Done

mixer/v1/attributes.proto
```diff
@@ -54,6 +54,36 @@ option (gogoproto.gostring_all) = false;
 // Each type of value is encoded into one of the so-called transport types present
 // in this message.
 //
+// Defines a list of attributes in uncompressed format.
```
rshriram: Please define the doc tags to make sure the doc gen does not document this file.. Take a look at other files for reference. Welcome to the docs madness.

qiwzhang: @rshriram 
The change is added to an existing file.   Do I still need to define the doc tag? 
Beside,  I don't find any doc tags in this file.  What does it look like?

mandarjog:
@qiwzhang where / how do you plan to use the uncompressed attributes ?

rshriram:
Might be too late. There’s no doc “tag”. Ptal at route_rule.proto to see
format we followed. Every proto line has a comment and messages have
examples, message fields have explanations as they become tables in html.

PR #199, opened 2017-10-12
C++ bazel rules for API

kyessenov:
Unfortunately, the fix to expose WKT for C++ was merged on Sep 5 and has missed protobuf 3.4.1.
Pointing to master as a temporary workaround.

I checked that both mixerclient and mixer build fine against this PR.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -2,8 +2,7 @@ package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])
 
-load("@io_bazel_rules_go//go:def.bzl", "gazelle", "go_prefix")
+load("@io_bazel_rules_go//go:def.bzl", "go_prefix")
 
 go_prefix("istio.io/api")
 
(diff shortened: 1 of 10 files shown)


kyessenov:
@chxchx @sebastienvas this is using wrong version of bazel. Please install 0.5.4 in this job.

rshriram:
Can we use whatever Envoy repo uses? I think @htuch has his own fork of the repo, for some hacks.

kyessenov:
We don't need a fork yet. The mainline protobuf has the fix that works with recent bazels.

sebastienvas:
We don't have 0.5.4, can we use 0.6.1 instead ?

kyessenov:
Yes, that's also fine.

sebastienvas:
https://github.com/istio/test-infra/pull/545

kyessenov:
Woohoo

PR #204, opened 2017-10-13
Add quota config proto file for Proxy

qiwzhang:
This is quota run-time config for Proxy.  Please see this design doc for detail.

https://goo.gl/zPbyML


kyessenov:
This should be moved to mixer proper as well. Is it hard to implement a basic expression parser that matches Mixer's grammar?

ayj:
This configuration is distributed to the proxy. The convention has been to put such types of config under `proxy/v1`.  And configuration that the Mixer proper uses is put under `mixer/v1`.

kyessenov:
`AttributeMatch` is more general than routing or proxy related config. Which is why it belongs to wherever attributes are defined. It happens that mixer defines attributes.

rshriram:
Was this doc ever shared publicly before hand? I can't find any reference in my emails. Since this is a new doc, please give me a few days to read/review. Will be back to the PR after that.

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,105 @@
...
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.proxy.v1.config;
+
+import "mixer/v1/attribute_match.proto";
+import "proxy/v1/config/route_rule.proto";
```
rshriram: The semantics of configuration are different between Mixer and Pilot. While terminology is same, mixer allows lots of expressions. We don't. I would prefer to not mix and match the two. Secondly, how will this be exposed to end user and shipped? If its via standard rules syntax in Pilot, then this whole thing has to change.. You have multiple rules per spec, while Pilot restricts things to a single rule per sec. Match does not have this clause field (but it might be okay to add. For certain, this "string_map_prefix" is super non standard.

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,105 @@
...
+  // A list of Quota rules.
+  repeated QuotaRule rules = 1;
+}
+
+// Specifies a rule with list of matches and list of quotas.
+// If any clause matched, the list of quotas will be used.
+message QuotaRule {
+  // If empty, match all request.
+  // If any of match is true, it is matched.
+  repeated istio.mixer.v1.AttributeMatch match = 1;
```
rshriram: Can we not use the match format in the route rule, if this is going to be in the proxy/v1/config ?

proxy/v1/config/quota.proto
qiwzhang: I think we need a consistent attribute match semantics across Istio components.  This attribute_match is an intent to achieve that.  
Current Pilot is using StringMatch on HTTP headers.  It is limited, not a generic attribute match. We should not restrict our-self by that semantic. 
Beside, proxy config could be used by either Pilot or Envoy mixer filter.   Quota config will be used by Envoy mixer filter.

proxy/v1/config/quota.proto
qiwzhang: match format in the route_rule only supports StringMatch.   Istio Attribute has other value types that we need to support match rule.

proxy/v1/config/quota.proto
kyessenov: There are two issue with using attribute language in routing rules.
1). Customers who are incrementally adopting Istio and not interested in Mixer, are forced to use Mixer terminology instead of the more familiar HTTP terms.
2). This puts a burden on all alternative proxy implementations to be compliant with Istio control plane. Each API addition comes at a cost of complexity in "mixerfilters" for other base proxies.

qiwzhang: This PR will not force route role (Pilot) to use attribute match semantics.  It can keep using HTTP teams, such as HTTP Headers.  
There are two different components here: Pilot and Envoy Mixer filter.  Pilot may not have Istio attributes.  But Envoy mixer filter has to use Istio attributes; API attributes,  JWT token attributes, request attributes, all have been unified into Istio attributes.  Quota has to use attribute match.

kyessenov: From this position, these terms belong mixer namespace, while pilot simply stuffs a verbatim blob into proxy configs. We should have an extension in route rules to embed config blobs for custom filters, while mixer config/spc specifies the content of the config blob.

qiwzhang: We will define MixerFilterConfig proto in a separate PR.  It should be in proxy/v1/config namespace since it is generated by Pilot and used by Envoy Mixer filter.  It has nothing to do with mixer.

qiwzhang:
@rshriram What are the main concerns?

proxy/v1/config/quota.proto
rshriram: @kyessenov that is a good point. That basically lets people configure the full gamut of Envoy filters outside whatever we support, in addition to giving the mixer folks flexibility in defining a variety of filters.

rshriram: So if this has nothing to do with route rules and this is a configuration artifact that is meant to be enforced via the Mixer, why is this residing in the proxy/v1/config ? The proxy/v1/config is badly named. It should be Pilot/v1/config. I guess what I am struggling to understand is why is Quota, a completely mixer concept, being pushed through Pilot, while some other configurations are being pushed through Mixer?

proxy/v1/config/quota.proto
```diff
@@ -0,0 +1,108 @@
...
+//           clause:
+//             request.headers:
+//               string_map_prefix:
+//                 sub_key: path
+//                 value: /books
+//             request.http_method:
+//               string_exact: POST
+//       - match:
+//           clause:
+//             api.operation:
```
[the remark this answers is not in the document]
qiwzhang: There could be other components in Istio/proxy generating Istio attributes.  e.g. API-Spec parser will generate api attribures, such as api.operation.   we may have JWT auth component in Proxy too.  it will generate auth attributes, such as auth.end_user.   

Users should be able to use any Istio attributes to config quota.   That is why quota match should be based on Istio attribute match.

proxy/v1/config/quota.proto
qiwzhang: @rshriram quota.proto file is moved to "mixer/v1/config/client" folder.  

Quota is mixer concept.  But for client cache,  it has to be processed in mixer client side.

PR #205, opened 2017-10-13
Add APISpec and APISpecBinding to enable API attribute generation

ayj:
See https://goo.gl/aw9M1w for a design proposal on "Istio API spec and attribute generation".

See https://goo.gl/3WTUfW for the list of API attributes.

NOTE: this depends on https://github.com/istio/api/pull/197 for UncompressedAttribute definition.


proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+  string http_method = 2;
+
+  // URI template to match against. Defined by
+  // https://tools.ietf.org/html/rfc6570.
+  string uri_template = 3;
+}
+
+// APISpecBinding defines the binding between APISpecs and one or more
+// IstioService.
+message APISpecBinding {
```
qiwzhang: I don't understand this one.  Is it used by CRD?
Is this specific to ApiSpec?  or other CRD items, such as quota, can use it too?
If so,  should we extract it out so it can be shared?

ayj: Yes, `APISpec` and `APISpecBinding` would be expressed as their own CRDs. 

This is specific to APISpec, but it might be possible to generalize for other types of config, e.g. quota. Alternatively, you could create QuotaBinding that is specific to quota.

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.proxy.v1.config;
+
+// APISpec defines the canonical configuration for generating API
```
geeknoid: API-related

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+// related attributes from HTTP request based on the method and uri
```
geeknoid: request -> requests

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+// templated path matches. It is sufficient for defining the API
+// surface of a service for the purposes of API attribute
+// generation. It is not intended to represent auth, quota,
+// documentation, or other information commonly found in other API
+// specification, e.g. OpenAPI.
```
geeknoid: specifications

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+//
+// Existing standards that define operations (or methods) in terms of
+// HTTP methods and paths can be normalized to this format for use in
+// Istio. For example, a simple petstore api described by OpenAPIv2
```
geeknoid: api -> API

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+//         uriTemplate: /api/pets/{id}
+//       - attributes:
+//           api.operation: deletePet
+//         httpMethod: delete
+//         uriTemplate: /api/pets/{id}
+//
+message APISpec {
+  // List of attributes that are generated when *any* of the HTTP
+  // patterns match. This list typically includes the "api.service"
+  // and "api.version".
```
geeknoid: ... "api.version" attributes.

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+
+// APISpecHTTPPattern defines a single pattern to match against
+// incoming HTTP requests. The per-pattern list of attributes is
+// generated if both the http_method and uri_template match. In
+// addition, the top-level list of attributes in the APISpec is also
+// generated.
+message APISpecHTTPPattern {
+  // List of attributes that are generated if the HTTP request matches
+  // the specified http_method and uri_template. This typically
+  // includes the "api.operation".
```
geeknoid: "api.operation" attribute.

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+  repeated UncompressedAttributes attributes = 1;
+
+  // HTTP request method to match against. Defined by
+  // https://tools.ietf.org/html/rfc7231#page-21.
```
geeknoid: Please make this a valid markdown URL reference
Please provide some examples directly inline.

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+  string http_method = 2;
+
+  // URI template to match against. Defined by
+  // https://tools.ietf.org/html/rfc6570.
```
geeknoid: Please make this a valid markdown URL reference
Please provide some examples directly inline.

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+  string uri_template = 3;
```
geeknoid: Why is this a template and not a regex? Templates are better used to produce URLs, not to pattern-match against them.

qiwzhang: If it is regex,  we have to eval each one.  the speed will be O(n) if there are n matches.  

url_template can be O(1).  you can think it as:
map<url_template + http_method,  attributes>

we support url_template with wild cards, such as /bar/*/foo or /foo/**

proxy/v1/config/api_spec.proto
qiwzhang: Who will create this binding?  and who will use it?

proxy/v1/config/api_spec.proto
geeknoid: Wayne, I'm sorry but I don't understand what you're saying.

It appears to me that for each template, you will need to issue a matchTemplate(template, url) call. You can just as easily do matchRegex(regex, url). Same overall performance complexity. What am I missing?

I don't know what the state of the art is in terms of regex handling in C++. Ideally, the regex can be parsed and tokenized once at config time, and then a faster regex match can occur at runtime.

proxy/v1/config/api_spec.proto
ayj: The `APISpec` owner could create the binding for existing services. Alternatively, the service owner could create the binding for an already defined `APISpec`. Pilot uses the bindings to distribute the `APISpec` to required proxies.

qiwzhang: So it will be stored in the same CRD as APISepc.  Now I understand it,  I will create a similar one for Quota too.

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+// Existing standards that define operations (or methods) in terms of
+// HTTP methods and paths can be normalized to this format for use in
+// Istio. For example, a simple petstore api described by OpenAPIv2
+// [here](https://github.com/googleapis/gnostic/blob/master/examples/v2.0/yaml/petstore-simple.yaml)
+// can be represented with the following APISpec.
+//
+//     apiVersion: config.istio.io/v1alpha2
+//     kind: APISpec
+//     metadata:
+//       name: petstore
```
smawson: add namespace for completeness?

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+//       - attributes:
+//           api.operation: deletePet
+//         httpMethod: delete
+//         uriTemplate: /api/pets/{id}
+//
+message APISpec {
+  // List of attributes that are generated when *any* of the HTTP
+  // patterns match. This list typically includes the "api.service"
+  // and "api.version".
+  UncompressedAttributes attributes = 1;
```
smawson: Minor annoyance, UncompressedAttributes should have been named Attributes, and Attributes should have been named CompressedAttributes :)

Maybe we can do the renaming at some point.

proxy/v1/config/api_spec.proto
smawson: We can build a single state machine to match all templates at once, effectively O(length of string). Thankfully we already have the code since it is fairly complex to write.

Otherwise you get O(length of string * number of regexes) matching, which can be slow if you have a large API.

proxy/v1/config/api_spec.proto
smawson: I think the mapping to CRD is flexible right? This is just the internal data format passed around, and doesn't require us to use a particular layout of CRDs?

I think it is probably OK to support this as a CRD but would prefer an annotation mechanism if it is more convenient. That would then just generate instances of this proto.

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+  // REQUIRED. One or more services to map the listed APISpec onto.
+  repeated IstioService services = 1;
+
+  message APISpecReference {
+    // REQUIRED. The short name of the APISpec. This is the resource
+    // name defined by the metadata name field.
+    string name = 1;
+
+    // Optional namespace of the APISpec. Defaults to the value of the
+    // metadata namespace field.
```
smawson: You mean it defaults to the value of the namespace the binding is in, right?

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,109 @@
...
+  string uri_template = 3;
+}
+
+// APISpecBinding defines the binding between APISpecs and one or more
+// IstioService.
+message APISpecBinding {
+  // REQUIRED. One or more services to map the listed APISpec onto.
+  repeated IstioService services = 1;
+
+  message APISpecReference {
```
smawson: Is there any reason we don't have a standard namespace+name proto we can use to name resources?

Seems silly to create different types to wrap namespace+name in various places, I assume we need this a lot (reference to some config object of a particular type).

proxy/v1/config/api_spec.proto
ayj: This could be encoded as an annotation on the service in which case we would remove (or omit)`IstioService` since that is implied.

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
ayj: done

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,118 @@
...
+  // HTTP request method to match against as defined by
+  // [rfc7231](https://tools.ietf.org/html/rfc7231#page-21). For
+  // example: GET, HEAD, POST, PUT, DELETE.
+  string http_method = 2;
+
+  // URI template to match against as defined by
+  // [rfc6570](https://tools.ietf.org/html/rfc6570). For example, the
+  // following are valid URI templates:
+  //
+  //    /pets
```
geeknoid: Extra space

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,118 @@
...
+  // REQUIRED. The short name of the APISpec. This is the resource
+  // name defined by the metadata name field.
+  string name = 1;
+
+  // Optional namespace of the APISpec. Defaults to the encompassing
+  // APISpecBinding's metadata namespace field.
+  string namespace = 2;
+}
+
+// APISpecBinding defines the binding between APISpecs and one or more
```
rshriram: Examples

proxy/v1/config/api_spec.proto
```diff
@@ -0,0 +1,118 @@
...
+  //
+  //    /pets
+  // 	/pets/{id}
+  // 	/dictionary/{term:1}/{term}
+  // 	/search{?q*,lang}
+  //
+  string uri_template = 3;
+}
+
+message APISpecReference {
```
rshriram: Missing top level comment for docs. And examples

rshriram:
And can you please double check the spacing (compare with route rule proto). Bad spacing results in more broken output.

qiwzhang:
@ayj I renamed UncompressedAttributes to Attributes.

PR #206, opened 2017-10-16
Rename Attributes to CompressedAttributes.

qiwzhang:
So that UncompressedAttributes can be called Attributes. 

Now it is good time to name it properly.  It will be much harder to do it once they are used all over the places.


douglas-reid:
cc: @geeknoid

costinm:
Isn't rename a backward incompatible change ? I thought we agreed that API changes will be announced/discussed and avoided if possible, and also that we'll attempt to maintain some discipline around API and backward compat. 

Is this absolutely necessary  ?

PR #209, opened 2017-10-17
secure istio components, address code review comments

rkpagadala:


proxy/v1/config/proxy_mesh.proto
```diff
@@ -83,16 +83,12 @@ message ProxyConfig {
...
-
-  // Address of the mixer service (e.g. __istio-mixer:15004_).
-  // Used by the proxy in front of mixer as the port to listen to,
-  // The value must be the same as MeshConfig.mixer_address
-  string mixer_address = 14;
-
-  // Pilot subject alternate name used for mutual TLS
-  repeated string pilot_san = 15;
+  // for Envoy-to-Envoy communication for istio components Mixer and Pilot.
+  AuthenticationPolicy control_plane_auth_policy = 13;
```
rshriram: How will sidecars use this configuration feature? Everything in this proto is for use by sidecars (read by istioctl and populated during sidecar injection).

rshriram: If its not going to be used by sidecars, I would basically get rid of this proto addition completely.

proxy/v1/config/proxy_mesh.proto
```diff
@@ -83,16 +83,12 @@ message ProxyConfig {
...
+  
+  // File path of custom proxy configuration, currently used by proxies
+  // in front of Mixer and Pilot.
+  string custom_config_file = 14;
```
rshriram: same as above.

proxy/v1/config/proxy_mesh.proto
rkpagadala: Thanks Shriram. All the changes are needed in the proxy. You can see them when I send out a PR on pilot repo.
Please hold off on approving until you see it.

PR #211, opened 2017-10-17
Add service_name field to check and report requests

mandarjog:
service identity is foundational for istio and Mixer.
This elevates `service_name` to a protocol field.

### DIFF
--- a/mixer/v1/check.proto
+++ b/mixer/v1/check.proto
@@ -53,6 +53,12 @@ message CheckRequest {
 
   // The individual quotas to allocate
   map<string, QuotaParams> quotas = 4 [(gogoproto.nullable) = false];
+
+  // Service_name is the canonical name of the destination service.
+  // It must be unique for a particular mesh.
+  // Service_name determines the configuration resources that are inspected during the request.
+  // If the caller can unambiguously determine the destination service, service_name must be specified.
+  string service_name = 5;
 }
 
 message CheckResponse {
--- a/mixer/v1/report.proto
+++ b/mixer/v1/report.proto
@@ -54,6 +54,12 @@ message ReportRequest {
   // The number of words in the global dictionary.
   // To detect global dictionary out of sync between client and server.
   uint32 global_word_count = 3;
+
+  // Service_name is the canonical name of the destination service.
+  // It must be unique for a particular mesh.
+  // Service_name determines the configuration resources that are inspected during the request.
+  // If the caller can unambiguously determine the destination service, service_name must be specified.
+  string service_name = 4;
 }
 
 message ReportResponse {


costinm:
I hope the implementation will keep backward compatibility, and if filed is missing load from attributes ?

To avoid a rename later - is it 'source', destionation, target  ? From comments it seems to be the destination.

Also: did you check with hybrid design docs - will it work for services that span multiple clusters, and 
what will be the 'canonical name' in that case ? 

It would be great if API changes would be reviewed in a more formal way..,.

mixer/v1/check.proto
```diff
@@ -53,6 +53,12 @@ message CheckRequest {
 
   // The individual quotas to allocate
   map<string, QuotaParams> quotas = 4 [(gogoproto.nullable) = false];
+
+  // Service_name is the canonical name of the destination service.
+  // It must be unique for a particular mesh.
```
geeknoid: You can remove this line, the uniqueness of the name is not relevant in this context.

mixer/v1/check.proto
```diff
@@ -53,6 +53,12 @@ message CheckRequest {
...
+  // Service_name determines the configuration resources that are inspected during the request.
+  // If the caller can unambiguously determine the destination service, service_name must be specified.
+  string service_name = 5;
```
geeknoid: Given the attribute is called destination.service, seems like it would be clearer to call this destination_service.
It would also be useful to mention the fact this field's value should not (or should it?) be specified as an attribute.

It would be worth providing an explanation why this field exists, given that the same value may also be provided as an attribute.

In case it's not clear by my comments here, I fear the introduction of this field will lead to puzzling questions for the user. When is this field used, as opposed to the attribute? What happens if this field is different than the attribute? Do things still work if the attribute is not specified? Etc...

qiwzhang: I need to know how mixer client can extract service name.   From Pilot?   Just extract "destination.service" attribute from Pilot and put it here?

ayj: A proxy may serve multiple services. How does mixerclient map the proxy identify to a specific service? 

Alternatively, the proxy/client could provide a client_identify and defer service-mapping to the mixer.

geeknoid:
I don't think we actually have a use for this feature at this point. Given all the open questions on this topic, and the uncertain value, I would close this PR for now. We can come back to it later if need be.

geeknoid:
I think we should retire this PR.

The service name should be obtained from the Mixer's sidecar. This would make it secure, based directly on the certs used to communicate with Mixer.

geeknoid:
Ping.

PR #213, opened 2017-10-18
add mixer_filter_config

qiwzhang:


mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+
+  // Send these attributes to Mixer in both Check and Report.
+  Attributes mixer_attributes = 4;
+
+  // Forward these attributes to upstream.
+  Attributes forward_attributes = 5;
+
+  // API specification to generate API attributes.
+  APISpec api_spec = 6;
+
```
ayj: Each service may implement a number of APIs. See [here](https://docs.google.com/document/d/16UII9dhIAJnijRpirMC-Ehsz6Q2MV06axQJFESPF838/edit#heading=h.x9snb54sjlu9)

`api_spec` needs to be repeated unless the configuration component that provides runtime config (this PR) merges multiple top-level APIspec into aggregate spec. The latter requires moving the per-spec [global attribute list](https://github.com/istio/api/blob/93848bb1c495ff6ebc1a938e2e0a403d52d345c1/mixer/v1/config/client/api_spec.proto#L67)  into the per-pattern attribute lists.

mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+  // Quota specification to generate quota requirement.
+  QuotaSpec quota_spec = 7;
```
ayj: This may need to be repeated. See my comment on `api_spec`.

mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+  bool call_mixer_check = 1;
+
+  // If true, call Mixer Report.
+  bool call_mixer_Report = 2;
+
+  // If true, disable forwarding any attributes to upstream.
+  bool disable_forward_attributes = 3;
+
+  // Send these attributes to Mixer in both Check and Report.
+  Attributes mixer_attributes = 4;
```
ayj: Any use case for differentiating attributes used for Check and Report?

mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+import "mixer/v1/config/client/api_spec.proto";
+import "mixer/v1/config/client/quota.proto";
+
+package istio.mixer.v1.config.client;
+
+// Defines mixer control config.
+// It is per-route for upstream if it is from Envoy route opaque config.
+message MixerControlConfig {
+  // If true, call Mixer Check.
+  bool call_mixer_check = 1;
```
ayj: The naming of boolean fields here isn't consistent. Can these be called `enabled_mixer_{check,report}` to match `disable_forward_attributes`?

mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+
+  // The flag to disable quota cache.
+  bool disable_quota_cache = 2;
+
+  // The flag to disable report batch.
+  bool disable_report_batch = 3;
+
+  // Specifies the policy when failed to connect to Mixer server.
+  // valid values are: [open|close]
+  string network_fail_policy = 4;
```
ayj: Any reason this can't be an enum isntead of a string?

mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+// It is per-route for upstream if it is from Envoy route opaque config.
+message MixerControlConfig {
+  // If true, call Mixer Check.
+  bool call_mixer_check = 1;
+
+  // If true, call Mixer Report.
+  bool call_mixer_Report = 2;
+
+  // If true, disable forwarding any attributes to upstream.
+  bool disable_forward_attributes = 3;
```
ayj: Why is this needed? Wouldn't this be equivalent to not adding any attributes to `forward_attributes`?

mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+  bool disable_report_batch = 3;
+
+  // Specifies the policy when failed to connect to Mixer server.
+  // valid values are: [open|close]
+  string network_fail_policy = 4;
+
+  // Specifies the default Mixer control config.
+  // If one is defined in upstream route table, their fields should be combined
+  // with following rules:
+  // * bool fields will be over-write; route config one over-write default one.
```
ayj: "Per-route mixer control config takes precedence over the default mixer control config. For example, ..."

mixer/v1/config/client/mixer_filter_config.proto
qiwzhang: Done. call it  "enable" for Check and Report.

mixer/v1/config/client/mixer_filter_config.proto
qiwzhang: it is more clearer with a bool field than just using empty forward_attributes field.   For example, default_control_config may have non-empty forward_attributes,  but per-route one is empty,  we may want still want to forward the attributes defined in "default".

mixer/v1/config/client/mixer_filter_config.proto
qiwzhang: For now, all Check attributes go to Report.  Report is super set of Check.

mixer/v1/config/client/mixer_filter_config.proto
qiwzhang: Change to enum

mixer/v1/config/client/mixer_filter_config.proto
qiwzhang: Done

mixer/v1/config/client/mixer_filter_config.proto
qiwzhang: Done

mixer/v1/config/client/mixer_filter_config.proto
qiwzhang: Done

mixer/v1/config/client/mixer_filter_config.proto
ayj: Makes sense - thanks for the explanation.

mixer/v1/config/client/mixer_filter_config.proto
```diff
@@ -0,0 +1,72 @@
...
+  // If true, disable forwarding any attributes to upstream.
+  bool disable_forward_attributes = 3;
+
+  // Send these attributes to Mixer in both Check and Report.
+  Attributes mixer_attributes = 4;
+
+  // Forward these attributes to upstream.
+  Attributes forward_attributes = 5;
+
+  // API specification to generate API attributes.
```
ayj: s/specification/specifications here and line 45

qiwzhang: Done

PR #214, opened 2017-10-19
Change proto Attributes to use map.

qiwzhang:
To support attribute update.  Update may replace or erase an attribute. 
Mixerclient plans to use this message to replace its Attributes class.

### DIFF
--- a/mixer/v1/attributes.proto
+++ b/mixer/v1/attributes.proto
@@ -54,7 +54,7 @@ option (gogoproto.gostring_all) = false;
 // Each type of value is encoded into one of the so-called transport types present
 // in this message.
 //
-// Defines a list of attributes in uncompressed format.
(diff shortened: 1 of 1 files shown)


mixer/v1/attributes.proto
```diff
@@ -60,14 +60,11 @@ option (gogoproto.gostring_all) = false;
...
 // 3) Forward attributes from client proxy to server proxy for HTTP requests.
 message Attributes {
-  // A list of attributes.
-  repeated Attribute attributes = 1;
-
-  // Specifies one attribute with name and value.
-  message Attribute {
-    // The attribute name.
-    string name = 1;
+  // A map of attribute name to its value.
```
ayj: Please update comments that reflect "map of attributes" instead of "list of attributes".

PR #225, opened 2017-11-03
commit generated protos

rshriram:
copied from istio.io/istio/vendor/api (whatever bazel generated there).


kyessenov:
You should also add dictionary file (attribute list)

PR #226, opened 2017-11-03
Vendor deps

rshriram:
Googleapi files..


rshriram:
cc @costinm / @mandarjog .. need this in order for dep ensure in istio/istio to work.

mandarjog:
related to https://github.com/istio/istio/pull/1354

kyessenov:
there is a lot of unnecessary code here, but if it makes things compile, doesn't hurt :)

ldemailly:
should this only checkin *.go ? 
how is the commit generated and how will we update it next ?

PR #227, opened 2017-11-06
[WIP] Gateway (standalone proxy) configuration

rshriram:
Multiple match clauses with OR semantics
Separate fault rule
top level destination field renamed to host to align with the ingress design (internal destination fields remain the same)

port/cidr in Istioservice to allow matching rules based on ports


rshriram:
@louiscryan based on our chat earlier on match being too limited.

proxy/v1/config/route_rule.proto
```diff
@@ -70,24 +70,24 @@ package istio.proxy.v1.config;
 //       name: my-rule
 //       namespace: default # optional (default is "default")
 //     spec:
-//       destination:
+//       host:
```
kyessenov: having namespace in host is weird. destination was supposed to be the normalized name of the service, not a specific host/authority header that was used.

proxy/v1/config/route_rule.proto
```diff
@@ -96,11 +96,17 @@ message RouteRule {
...
+  // DEPRECATED. Match condtions to be satisfied for the route rule to be
   // activated. If match is omitted, the route rule applies only to HTTP
   // traffic.
   MatchCondition match = 3;
 
+  // One or more match clauses to be used with OR semantics.
+  // i.e. match is true if one of the clauses match. A clause is said to
+  // match if all the conditions in the clause are true. Only one of match
+  // or multimatch can be specified.
+  repeated MatchCondition multimatch = 15;
```
kyessenov: can we not call it multimatch? can we do some proto tricks to change `match`?

proxy/v1/config/route_rule.proto
```diff
@@ -160,20 +166,24 @@ message IstioService {
   // The short name of the service such as "foo".
   string name = 1;
 
-  // Optional namespace of the service. Defaults to value of metadata namespace field.
+  // DEPRECATED. Optional namespace of the service. Defaults to value of metadata namespace field.
   string namespace = 2;
```
kyessenov: Why deprecate namespace?

proxy/v1/config/route_rule.proto
```diff
@@ -160,20 +166,24 @@ message IstioService {
...
 
   // Domain suffix used to construct the service FQDN in implementations that support such specification.
   string domain = 3;
 
   // The service FQDN.
-  string service = 4;
+  string fqdn = 4;
```
kyessenov: This is going backwards from using istio attributes? Should we still care about that or not?

proxy/v1/config/route_rule.proto
```diff
@@ -160,20 +166,24 @@ message IstioService {
...
 
   // Optional one or more labels that uniquely identify the service version.
-  //
-  // *Note:* When used for a RouteRule destination, labels MUST be empty.
-  //
   map<string, string> labels = 5;
+
+  // The port on which the service is listening for inbound connections.
+  int32 port = 6;
```
kyessenov: Why is port in service?

proxy/v1/config/route_rule.proto
frankbu: isn't this a gratuitous breaking change anyway? Even if it is a host, it is the/a host of the request destination, so why not leave the field as it is?

proxy/v1/config/route_rule.proto
frankbu: +1 to trying to keep it as `match` and figuring out how to support the single ones for backward compat.

proxy/v1/config/route_rule.proto
frankbu: again, isn't this a gratuitous breaking change?

proxy/v1/config/route_rule.proto
```diff
@@ -493,3 +503,34 @@ message CorsPolicy {
...
+  // resolution for HTTP traffic as well as IP-based resolution for
+  // TCP/UDP traffic.
+  IstioService destination = 1;
+
+  // RECOMMENDED. Precedence is used to disambiguate the order of
+  // application of rules for the same destination service. A higher number
+  // takes priority. If not specified, the value is assumed to be 0.  The
+  // order of application for rules with the same precedence is
+  // unspecified.
+  int32 precedence = 2;
```
frankbu: Are FaultRule precedences and rule evaluation separate from other rules? If not, what is the advantage of creating a special resource?

proxy/v1/config/route_rule.proto
louiscryan: Its back compat in proto, is the same true in YAML ?

kyessenov: Could be, but impl is far from perfect. There is a way to modify de-serializer if it becomes absolutely required. Or we just rewrite our configs.

rshriram:
@kyessenov @ZackButcher @louiscryan @9len @mccv @frankbu PTAL. Based on individual conversations and docs, etc., here is a stab at a few things:

1. ability to reuse pieces across route rules (rewrites, timeouts, retries, faults, cors, etc.). The idea here is that we would have first class resource called route property which would contain some of these reusable elements. Route rules can simply refer to a route property and the specific piece of the property they want to reuse.

2. I have removed the rule selector from the Ingress Rule. Instead, route rules need to be defined with the same destination as the one used in the ingress rule. There is a provision for a default host. You can pin a rule to specific LBs using the source match.

3. The ingress rule is a superset of EgressRule, with two notable differences:
    1. Egress is almost expressing a service, while Ingress is describing properties of a LB exposing multiple services.
    2. Egress carries an implicit assumption that the traffic goes to the original destination (orig_dst cluster in envoy). 
IF we can reconcile these two, we would have a much nicer representation

4. The fact that route rules need to be defined with specific hosts that match the ingress rules allows us to reproduce the kube ingress behavior (404 if not found).

I still haven't managed to collapse the circuit breakers, etc. But I hope this provides more clarity.

proxy/v1/config/route_rule.proto
```diff
@@ -143,37 +146,70 @@ message RouteRule {
...
-  // Additional HTTP headers to add before forwarding a request to the
+  // DEPRECATED. Additional HTTP headers to add before forwarding a request to the
   // destnation service.
   map<string, string> append_headers = 14;
+
+  // An array of key value pairs that specify the list of properties applicable to
+  // this route. Keys are well defined property names (e.g., rewrite, timeout, retry,
+  // fault, cors) and values are names of the property rule. If the property does not
+  // exist, it will be ignored. The valid property names are rewrite, timeout, retry,
+  // fault, cors.
```
frankbu: I don't follow how these keys work. If the value is the name of a property rule, which presumably can have more than one property (e.g., a timeout and a fault), what would the key be? Or did you mean for the fields in the property rule to be oneof? Even so, why do you need the key (e.g., timeout) if the value contains the timeout field?

rshriram: Its like
```yaml
properties:
  timeout: global-policy
  retry: global-policy
```

and a RouteProperties like
```yaml
metadata:
 name: global-policy
spec:
  timeout
    ...
   retry:
     ...
```

The key indicates the type (of somethingy) and the value indicates the route property name that contains these thingies. Its easy to create dedicated resources for Timeout, Retry, Rewrite, etc. WE would have too many API resources. Instead, I thought it would be easier if we can put all of these into one file and then essentially say that pick the timeout from this file, retry from the other file, etc.
The intended usage is that there will be a global setting that everyone will use. If someoen wants to override this, they would define a custom property file with the special timeout setting and then they tweak the route rule to specify that the timeout must be picked up from their custom property file instead of the global one.

proxy/v1/config/route_rule.proto
```diff
@@ -143,37 +146,70 @@ message RouteRule {
...
+  // For example, when running on Kubernetes, the domain field could include the namespace of the service
+  // along with the kubernetes domain suffix: "myns.svc.cluster.local". When running with Consul as the service
+  // registry, the domain suffix could be "service.consul".
   string domain = 3;
 
-  // The service FQDN.
+  // DEPRECATED. The service FQDN.
   string service = 4;
 
+  // A list of HTTP authorities or FQDNs that comprise the composite
```
frankbu: Do you really mean authorities here? Can the value optionally include a port? I'm also finding the mixed use of the terms "host", "authority" and "fqdn" confusing. They seem to be used more-or-less interchangeably.

proxy/v1/config/route_rule.proto
ZackButcher: I dislike string keys here. We have a limited set of valid properties, I'd rather see a well-typed message used to refer to them.

```proto
message RouteRules {
...
    message Properties {
        string retry = 1;
        string timeout = 2;
        ...
    }
    Properties properties = 15;
}
```

This also opens the door for us to allow in-line declaration of policies with a small change:

```proto
message Properties {
    oneof retry {
        string name = 1;
        HTTPRetry inline = 2;
    }
    oneof timeout {
        string name = 3;
        HTTPTimeout inline = 4;
    }
    ...
}
```

kyessenov: Another issue is lack of referential integrity and bundled updates. We should prescribe what happens when the name is not associated with a policy in a fashion, consistent with the rest of Istio. We've avoided this problem in the current design, preferring scoped overrides, instead of explicit sharing.

proxy/v1/config/route_rule.proto
kyessenov: Can you clarify why namespace is deprecated? K8s uses it extensively to qualify service names for cluster-scoped resources such as ingress.

rshriram: Sorry I missed this. The idea is to get rid of namespace from inside the route rule as it increasingly looks like a k8s leak. The namespace can always be encoded as part of the domain. And it can also be derived from the metadata namespace if we are using this for a k8s service.

proxy/v1/config/route_rule.proto
rshriram: With regard to string types, I am toying with the idea. The motive is to find a way to allow people to enable custom filters or such in Envoy. If there is a better way, i would love to hear it. The inlining part is cool. 

I don’t want to over engineer the ref integrity aspect. The reason being these are all optional settings. Stuff still works if timeouts are not set.

proxy/v1/config/route_rule.proto
```diff
@@ -215,8 +253,30 @@ message MatchCondition {
...
+    L4MatchAttributes udp = 3 ;
+
+    // Attributes of an HTTP request to match.
+    MatchRequest request = 4 ;
+
+  }
+
+  // One or more match clauses to be used with OR semantics.
+  // i.e. match is true if one of the clauses match.
+  repeated MatchClause clause = 5;
```
frankbu: Do we really want another level of nesting in the structure? I.e., this example:
```
match:
  source:
    name: reviews
    labels:
      version: v2
  request:
    headers:
      ...
```
becomes:
```
match:
- clause:
    source:
      name: reviews
      labels:
        version: v2
    request:
      headers:
        ...
```

proxy/v1/config/http_fault.proto
```diff
@@ -43,17 +43,23 @@ message HTTPFaultInjection {
   // service.
   //
   //     metadata:
```
frankbu: For clarity, add:
```
//     apiVersion: config.istio.io/v1alpha2
//     kind: RouteProperties
```

frankbu: also:
```
//     apiVersion: config.istio.io/v1alpha2
//     kind: RouteRule
```
below, and in following example

proxy/v1/config/route_rule.proto
rshriram: Yes and its inevitable. You can do single matches or if you want multiple matches, you need some way of delineating multiple match clauses. We could just have an array of matches, but I don't know how to go about converting a singleton element in current config to an array in a backward compatible manner.

frankbu: Hmm, from @kyessenov 's  earlier comment, I had the impression that if we change `match` to a repeated field, then the new syntax will be:
```
spec:
- match:
    source:
      name: reviews
      ...
```
BUT, with some deserializer tricks we can support the old syntax for backward copat:
```
spec:
  match:
    source:
      name: reviews
      ...
```
Did I misunderstand?

kyessenov: We can override deserialization semantics in protobuf if we're willing to play that game.

ZackButcher: It would be way nicer to not have to play that game. Extra level of nesting is way better than paying the price of supporting a non-standard proto ingestion pipeline.

kyessenov: Can we just increment the version and provide proto translation mechanism?

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,239 @@ message IngressRule {
...
+    // If set to true, the load balancer will send a 302 redirect for all
+    // http connections, asking the clients to use HTTPS.
+    bool https_redirect = 1;
+
+    // Optional: Indicates whether connections to this port should be secured using TLS.
+    // The value of this field determines how TLS is enforced. If set to "simple", Envoy will
+    // secure connections with standard TLS semantics (server certs only). If set to "mutual",
+    // Envoy will validate client's identity using mTLS authentication. If set to "passthrough",
+    // Envoy will forward the connection to the upstream server as is.
+    string mode = 2;
```
ZackButcher: Lets add an enum here please.

```proto
enum TLSMode {
    PASSTHROUGH = 0;
    SIMPLE = 1;
    MUTUAL = 2;
}
```
We should pick the default value carefully; I think passthrough is the most reasonable since it makes the proxy most transparent by default.

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,245 @@ message IngressRule {
...
+  // within the mesh. ONLY ONE OF rules or backend is allowed.
+  Backend backend = 5;
+}
+
+// Backend describes a service with a port that is the recipient of TCP
+// connections arriving at the ingress load balancer. Backends are
+// typically used in the context of TCP services such as Mongo/Redis/plain
+// TCP connections.
+message Backend {
+  // REQUIRED: Name of the service inside the mesh for which connections
```
kyessenov: Need namespace and port name as an alternative to port value.

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,245 @@ message IngressRule {
...
+//           port: 2379
+//
+message Server {
+  // REQUIRED: The Port on which Envoy should listen for incoming
+  // connections
+  int32 port = 1;
+
+  // REQUIRED: The protocol associated with this port. The protocol should
+  // be one of HTTP|HTTP2|GRPC|MONGO|REDIS|TCP.
+  string protocol = 2;
```
kyessenov: Why isn't this inherited from backend port?

proxy/v1/config/route_rule.proto
```diff
@@ -153,27 +153,58 @@ message RouteRule {
...
+  // metadata namespace field.
   string namespace = 2;
 
-  // Domain suffix used to construct the service FQDN in implementations that support such specification.
+  // Domain suffix used to construct the service FQDN in implementations
+  // that support such specification.  For example, when running on
+  // Kubernetes, the domain field could include the namespace of the
+  // service along with the kubernetes domain suffix:
+  // "myns.svc.cluster.local". When running with Consul as the service
+  // registry, the domain suffix could be "service.consul".
```
kyessenov: As I mentioned, the namespace field acts as a wildcard. Without it, expressing a policy over a namespace requires using `*.svc.cluster.local` as a domain while these fields are not supposed to be wildcards.

proxy/v1/config/route_rule.proto
kyessenov: Also, why multiple fqdns? What happens when they collide?

proxy/v1/config/route_rule.proto
```diff
@@ -153,27 +153,58 @@ message RouteRule {
...
-  //
   map<string, string> labels = 5;
+
+  // The ports on which the service is listening for inbound connections.
+  // Comma separated list of names or port numbers. If names are used, the
+  // same name must be used when defining the service in the orchestration
+  // platform.  If ports are omitted, all HTTP ports associated with the
+  // service will be subjected to the associated route rules. Similar
+  // principle applies to TCP ports as well.
+  repeated string port = 6;
```
kyessenov: Why is port here and not inherited from platform port specification?

proxy/v1/config/route_rule.proto
```diff
@@ -153,27 +153,58 @@ message RouteRule {
...
+// FQDN of the service is composed from the name and
+// implementation-specific domain suffix. When used in the context of a
+// route rule with meta data fields, the namespace associated with the
+// metadata will be automatically inherited as a domain suffix if its not
+// overridden by the domain field in the IstioService
+// specification. Alternatively, a fully qualified domain name can be used
+// to identify the service. Either name+domain or fqdn should be used, but
+// not both.  if the (e.g. on Kubernetes, "reviews" +
+// "default.svc.cluster.local" -> "reviews.default.svc.cluster.local").
 message IstioService {
```
kyessenov: The intention for this was a match condition not a declaration. We probably need a separate declaration message that is essentially `model.Service` sans labels.

proxy/v1/config/ingress_rule.proto
rshriram: I am reluctant to put the namespace here. How about we have hostname or ip as string?

rshriram: why is the port name significant here? We don’t have named ports in most platforms. Secondly, the port and protocol are sufficient enough to identify the Envoy filter to use.

proxy/v1/config/route_rule.proto
rshriram: Multiple fqdns because we might want to apply the same rule for multiple services.

proxy/v1/config/ingress_rule.proto
rshriram: This is front end server.

proxy/v1/config/route_rule.proto
rshriram: I just collapsed namespace and domain into one thing. you can always do myns.svc.cluster.local as domain

proxy/v1/config/route_rule.proto
rshriram: If omitted, it will inherit from platform and apply to all ports of same protocol. I put ports here only to help people differentiate between multiple http ports (80/8080/etc). But I agree that we should create a port struct to allow specifying named ports.

proxy/v1/config/route_rule.proto
rshriram: We are not using it as a declaration are we? It’s about identifying the service only

proxy/v1/config/ingress_rule.proto
kyessenov: What's the relation to the backend port? Why would it be HTTP if backend port is MONGO?

proxy/v1/config/ingress_rule.proto
kyessenov: There is a mismatch with k8s if you don't put those two things. Those are standard and used features of k8s.

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,245 @@ message IngressRule {
...
+// connections arriving at the ingress load balancer. Backends are
+// typically used in the context of TCP services such as Mongo/Redis/plain
+// TCP connections.
+message Backend {
+  // REQUIRED: Name of the service inside the mesh for which connections
+  // are being proxied
+  string name = 1;
+
+  // REQUIRED: Port number to which the connections should be forwarded to.
+  int32 port = 2;
```
ZackButcher: Should this port be a string (to allow named ports)? In IstioService we define port as string.

proxy/v1/config/ingress_rule.proto
```diff
@@ -173,8 +170,8 @@ message K8SIngressRule {
 //       - destination:
 //           name: homepage
 //
-message IngressRule {
-  // REQUIRED: Details about the port on which the ingress proxy listens and applies rules
+message Gateway {
```
costinm: AFAIK the keyword 'IngressRule' is not present in any of the .yaml or API surface, so it should have 
no impact in upgrade or migration.

However it would be good to double check and have clear release notes on anything that may be backward
incompatible.

proxy/v1/config/ingress_rule.proto
kyessenov: It should be `IntOrString` which is (sadly) difficult to represent in proto.

kyessenov:
I'm clearing my PR block but I don't think this is ready to merge yet.

proxy/v1/config/ingress_rule.proto
rshriram: good catch. I can just change this to a string. and I omitted protocol here because backend is tied to the server spec that has protocol defined there.

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,245 @@ message IngressRule {
...
+// of the user to ensure that external traffic to these ports are allowed into the
+// mesh.
+//
+//     metadata:
+//       name: my-gateway
+//     spec:
+//       servers:
+//       - port: 80
+//         protocol: HTTP
+//         authorities:
```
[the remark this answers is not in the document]
rshriram: This is a totally different spec right? Anyway, I don't mind as long as Monsieur @louiscryan agrees. :). He seemed to prefer authority over host.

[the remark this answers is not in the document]
costinm: If the new ingress spec will also use authority ( and accept host for backward compat ) - I mostly agree.

Still not sure what 'authority' means for TCP (or UDP) :-)

proxy/v1/config/ingress_rule.proto
rshriram: Backend does not have a protocol. Backend port is just a string. The idea is that the backend inherits the server's protocol. if server is http, then we apply route rules (which by default pick http ports). if server is using tcp/mongo/redis, then we allow a backend to be specified (the protocol is inferred). Make sense?

proxy/v1/config/ingress_rule.proto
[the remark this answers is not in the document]
frankbu: +1 also because an `authority` is syntactically `host[:port]`, but the values here do not include an optional port.

proxy/v1/config/route_rule.proto
rshriram: do we need destination policies over an entire namespace?

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,257 @@ message IngressRule {
...
+    TLSmode mode = 2;
+  }
+
+  // Set of TLS related options that govern the server's behavior. Use
+  // these options to control if all http requests should be redirected to
+  // https, and the TLS modes to use.
+  TLSOptions tls = 2;
+
+  // Forward TCP connections on the ingress port to a specific service
+  // within the mesh. ONLY ONE OF rules or tcp_backend is allowed.
```
ZackButcher: Missing `rules` field, and lets go ahead and make it a formal `oneof`.

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,257 @@ message IngressRule {
...
+  TCPBackend tcp_backend = 5;
```
ZackButcher: What happens when I use `protocol = TCP` but supply `rules` instead of a `tcp_backend`?

kyessenov: I was looking for HTTPBackend but there's none. Why special case this?

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,257 @@ message IngressRule {
...
+}
+
+// TCPBackend describes a service with a port that is the recipient of TCP
+// connections arriving at the ingress load balancer. TCPBackends are
+// typically used in the context of TCP services such as Mongo/Redis/plain
+// TCP connections.
+message TCPBackend {
+  oneof target {
+    // Name of the service for which connections are being proxied.
+    string hostname = 1;
```
kyessenov: `hostname` in one place and `authority` in another. We should be consistent. Is this FQ or short or short+namespace?

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,257 @@ message IngressRule {
...
+    // IP address of the service being proxied.
+    string address = 2;
```
kyessenov: Why static address in the rule? Who assigns this address?

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,257 @@ message IngressRule {
...
+  }
+
+  // REQUIRED: Port number or name to which the connections should be
+  // forwarded to. The protocol used for this port will be automatically
+  // derived from the protocol used by the associated Server that received
+  // the connection from a client.
+  string port = 3;
```
kyessenov: This needs to be `oneof`. You can name port "80" of value 8080 unfortunately in k8s.

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,257 @@ message IngressRule {
...
+//       certs:
+//       - authorities:
+//           - foo.bar.com
+//         serverCertificatet: server.crt
+//         clientCABundle: client.ca-bundle
+//       - authorities:
+//           - example.com
+//         serverCertificate: server.crt
+//
+message TLSConfig {
```
kyessenov: Can we import data-plane-api config for all the TLS settings?

proxy/v1/config/proxy_mesh.proto
```diff
@@ -155,6 +155,13 @@ message MeshConfig {
     // a secondary ingress controller (e.g., in addition to a
     // cloud-provided ingress controller).
     STRICT = 2;
+
+    // Istio ingress controller will not process any Kubernetes ingress
+    // resources.  Instead it will only process Istio specific
+    // IngressRules. Use this mode if Istio ingress controller will be a
+    // deployed in non-kubernetes environments, or if you wish to use Envoy
+    // as a standalone load balancer managed by Istio.
+    CUSTOM = 3;
```
kyessenov: This is an orthogonal mode dimension to strict/etc.

proxy/v1/config/ingress_rule.proto
ZackButcher: Unclear to me too; does this also imply that we need to add `UDPBackend` down the line? I think we want something more general here.

proxy/v1/config/ingress_rule.proto
rshriram: I don't know how to do that in proto. I tried once and it just barfed due to some bazel error. I can import the TLS config as is, but there are some language differences

rshriram:
@costinm this is not ready for review yet. I am just putting it out here to visualize how this will look.

proxy/v1/config/proxy_mesh.proto
ZackButcher: Yeah, I think `OFF` covers this case fine: we turn off the Pilot k8s ingress controller; we won't touch ingress resources. The fact that ingress to the cluster is being controlled via a service with Gateway rules is orthogonal to "is pilot the ignress controller?"

proxy/v1/config/ingress_rule.proto
kyessenov: We don't have bazel here. Just run `protoc` manually with the write imports and worry about CI/CD later.

proxy/v1/config/ingress_rule.proto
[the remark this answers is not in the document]
louiscryan: I'm not fanatical here. I was just trying to use one term consistently while we discussed the design principles

proxy/v1/config/ingress_rule.proto
rshriram: Sorry. that was typo. I fixed the comment.

proxy/v1/config/ingress_rule.proto
rshriram: In case people want to use ip address of service instead of hostname.

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +56,263 @@ message IngressRule {
...
+//             name: internal-redis
+//       certs:
+//       - authorities:
+//           - *.bookinfo.com
+//         serverCert: server.crt
+//         clientCABundle: client.ca-bundle
+//
+// The following routing rule exposes uk.bookinfo.com, eu.bookinfo.com via the gateway
+// and routes all calls to /reviews to the reviews service. Since a port
+// list is missing, this rule will apply to all http ports of the reviews service.
```
frankbu: What "port list" are you referring to here? Route rule destinations don't include ports, do they?

Also, in the bookinfo example, the gateway path `/reviews` (e.g., `http://eu.bookinfo.com/reviews:80`) needs be mapped to port 9080 of the reviews service, right? Where is this 80->9080 mapping specified? It used to be in the K8s Ingress.

proxy/v1/config/route_rule.proto
```diff
@@ -493,3 +493,15 @@ message CorsPolicy {
   // Access-Control-Allow-Credentials header.
   google.protobuf.BoolValue allow_credentials = 6;
 }
+
+// Port describes the properties of a specific port of a service.
+message Port {
+  // A valid non-negative integer port number.
+  int32 port = 1;
```
frankbu: Maybe `number` would be a better name for this field, to avoid having 2 nested port fields, that is:
```
    - port:
        number: 80
        name: http
        protocol: HTTP
```
instead of this:
```
    - port:
        port: 80
        name: http
        protocol: HTTP
```

kyessenov: Maybe `value`? 
```yaml
- port:
    value: 80
```

ZackButcher: I like `number` personally, it reads how you'd talk about it e.g. "what port number do I need to use to call this service?"

rshriram: done. created separate PR

proxy/v1/config/ingress_rule.proto
rshriram: done

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +65,263 @@ message IngressRule {
...
+//         serverCertificatet: server.crt
+//         clientCABundle: client.ca-bundle
+//       - hosts:
+//           - example.com
+//         serverCertificate: server.crt
+//
+message TLSConfig {
+  // REQUIRED: A set of hostnames associated with the server
+  // certificate. For wildcard hostname, use *
+  repeated string hosts = 1;
```
ZackButcher: I thought we wanted `domains` instead of `hosts`?

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +65,263 @@ message IngressRule {
...
+
+  // Set of TLS related options that govern the server's behavior. Use
+  // these options to control if all http requests should be redirected to
+  // https, and the TLS modes to use.
+  TLSOptions tls = 3;
+
+  // Forward TCP connections on the ingress port to a specific service
+  // within the mesh. TCP backend cannot be used for HTTP services. Use
+  // route rules to configure routing for HTTP services.
+  TCPBackend tcp_backend = 4;
```
ZackButcher: Why do we special case this to be TCP only? Backend is generally useful, and what's more the TCPBackend contains a `Port` field which can have an arbitrary protocol. What happens when I write?

```yaml
server:
  port:
    name: foo
    protocol: TCP
  hosts:
  - foo.bar.com
  TcpBackend:
    port:
       name: bar
       protocol: HTTP
```

I liked the idea of a default backend for the Gateway no matter what, and for HTTP that default backend becomes the destination I write route rules against. I also think this model scale better as we add more protocols. We don't want `TCPBackend`, `UDPBackend`, `MongoBackend`, `RedisBackend`, etc, the explosion becomes unmaintainable.

proxy/v1/config/ingress_rule.proto
louiscryan: right

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +65,263 @@ message IngressRule {
...
+//           - *
+//       - port:
+//           number: 2379 #to expose internal service via external port 2379
+//           protocol: REDIS
+//           name: redis
+//         tls:
+//           mode: mtls
+//         hosts:
+//           - redis.bookinfo.com
+//         tcpBackend:
```
louiscryan: This needs to live in RouteRule somehow otherwise we have an API alignment issue. 

Question here is how to do the binding. Features that are available to match on include:

1 gateway domain vs route domain
2 declared gateway protocol vs service port protocol
3 declared gateway port name vs service port name
4 declared gateway port number vs service port number

I suspect we want to allow matching on any of these properties but strongly encourage matching on domain, protocol and port to start with

@ZackButcher 
@kyessenov 
@frankbu 

thoughts ?

proxy/v1/config/ingress_rule.proto
```diff
@@ -65,3 +65,263 @@ message IngressRule {
...
+message Server {
+  // REQUIRED: The Port on which Envoy should listen for incoming
+  // connections
+  Port port = 1;
+
+  // REQUIRED: A list of hosts exposed by this load balancer. While
+  // typically applicable to HTTP services, it can also be used for TCP
+  // services using TLS with SNI.  Wildcards are allowed in the authority
+  // values. A special wildcard value of "*" stands for the default host.
+  repeated string hosts = 2;
```
louiscryan: Should be 'domains'

proxy/v1/config/ingress_rule.proto
ZackButcher: So working under the rule "route rules are exceptions to typical traffic flow", I thought the gateway specifying a default backend for traffic to the exposed domains was reasonable. Then we use regular old RouteRules to target that default destination to make exceptions to the typical flow (where the typical flow is that requests go to the default backend).

ZackButcher: But yes, I think matching based on domain, protocol, and port is the right starting point.

rshriram:
On Tue, Nov 14, 2017 at 3:55 AM Zack Butcher <notifications@github.com>
wrote:

> *@ZackButcher* commented on this pull request.
>
> In proxy/v1/config/ingress_rule.proto
> <https://github.com/istio/api/pull/227#discussion_r150684577>:
>
> > +//           protocol: HTTPS
> +//         tls:
> +//           mode: simple #enables HTTPS on this port
> +//         hosts:
> +//           - uk.bookinfo.com
>
> +//           - eu.bookinfo.com
> +//           - *
> +//       - port:
> +//           number: 2379 #to expose internal service via external port 2379
> +//           protocol: REDIS
> +//           name: redis
> +//         tls:
> +//           mode: mtls
> +//         hosts:
> +//           - redis.bookinfo.com
> +//         tcpBackend:
>
> So working under the rule "route rules are exceptions to typical traffic
> flow", I thought the gateway specifying a default backend for traffic to
> the exposed domains was reasonable. Then we use regular old RouteRules to
> target that default destination to make exceptions to the typical flow
> (where the typical flow is that requests go to the default backend).
>
>
Sorry I didn’t get a chance to update fully. But here is the gist:

We have a gap in our route rules impl today - we are not matching on
L4MatchAttributes and converting it into tcp proxy route configs. In order
to be able to do that reliably, we need a port based (not name as port
names are not globally unique) match in the destination, in addition to
domain or name.

Once we fix this gap, we should be able to write route rules (except
weighted routing) for tcp services: for port 8888 route to foo, labels x y
z.

Now assuming that exists, we can completely get rid of tcpBackend from
gateway. It would become purely a listener and virtual host spec. (One then
wonders why even have gateway resource and why not borrow Envoy listener,
vhost struct).

At this point, we have this nice decoupling: platform specifies the gateway
setup (ie kube/mesos/cf service spec), gateway config specifies the
properties of ports (tls, domains, type of listeners, etc), and route rules
define the routing (once we have weighted routing for tcp, things become
fun).

The reason I omit the httpBackend or the default backend for http stuff is
because it is equivalent to the current ingress stuff we do. Merge internal
route rules with ingress backends. That becomes kind of inelegant  when I
want to route to different backends based on url path, host header, etc. I
also don’t get the 404 by default property out of the box, because default
backend implies route all paths to backend.
Since We are pinning route rules to domains and gateway’s servers to
domains, we could define routerules based on domains and a single rule for
wildcard destination on the port, where there is no domain.

Thoughts?

proxy/v1/config/ingress_rule.proto
louiscryan: I've added some notes about this in the design doc.

proxy/v1/config/route_rule.proto
louiscryan: I think we can tolerate using a single field. First match by name, then treat as number if possible and match by that

kyessenov:
It makes a difference in JSON since 80 and "80" are different values.

rshriram:
closing in favor of #254 . Will send smaller PRs for other changes.

PR #228, opened 2017-11-07
add mixerclient auth policy for JWT and API-KEY

ayj:
This PR introduces the proxy/mixerclient runtime configuration for JWT validation and API-KEY attribute generation. The proposed mixerclient EndUserAuthenticationPolicy serves two purposes:

1) internal config that is generated/distributed by Pilot and consumed by Proxy.
2) user-facing config that user could define themselves and/or be derived from something like OpenAPI.

(2) should be considered temporary and we should instead use SecurityPolicy when its available. For short term to unblock API management development we'll overload use of (1) for (2).

Furthermore, overtime some aspects of mixerclient configuration might be moved to Envoy, e.g. JWT config if the JWT auth filter is upstreamed to Envoy.

JWT config needs is based on our existing JWT filter (see  https://github.com/istio/proxy/blob/master/src/envoy/auth/http_filter_config.cc)

API-KEY config is based on https://goo.gl/TdRL8a.


mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,172 @@
...
+  // system wide default is applied if no duration is explicitly
+  // specified.
+  google.protobuf.Duration public_key_cache_duration = 5;
+}
+
+// Determines how to apply auth policies for individual requests.
+message AuthPolicySpec {
+  // If empty, apply auth policy to all requests.
+  // If any of match is true, it is matched.
+  repeated AttributeMatch match = 1;
```
qiwzhang: There is a performance problem with conbining "api-key" and "jwt-auth" into one policy.   Since we are using "AttributeMatch",   we have to sequentially match each one.   In module design,  api-key code may not be in the same module as JWT auth code.    When api-key module wants to find out how to extract api-key,  it has to match all rules with JWT auth too since it doesn't know which one is api-key.

I feel at high level we should separate them out.  So api-key code only check apikey spec and jwt-auth code checks auth spec.

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,172 @@
...
+  // https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.1
+  // Usually a URL or an email address.
+  //
+  // Example: https://securetoken.google.com
+  // Example: 1234567-compute@developer.gserviceaccount.com
+  //
+  string issuer = 1;
+
+  // The list of JWT
+  // [audiences](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.3).
```
lizan: nit: link to RFC 7519, not the draft.

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,172 @@
...
+  // that are allowed to access. A JWT containing any of these
+  // audiences will be accepted.
+  //
+  // Any audience will be accepted if the audience list is empty.
```
lizan: Note the service name will be accepted if emtpy, based on today's end-user auth WG discussion.

PTAL:
https://docs.google.com/document/d/14Zqj5EAOYdy6GyBzXUo3txnVEapsnEC6-04pJD-QMwk/edit#heading=h.aflyta6as7rn

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,172 @@
...
+  // Optional if the key set document can either (a) be retrieved from
+  // [OpenID
+  // Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html
+  // of the issuer or (b) inferred from the email domain of the issuer
+  // (e.g. a Google service account).
+  //
+  // Example: https://www.googleapis.com/oauth2/v1/certs
+  string jwks_uri = 3;
+
+  // Value to insert into `sec-istio-auth-userinfo` header (or
```
lizan: On the second thought, let's just has a bool flag, and not turned it into `sec-istio-auth-userinfo`, since the whole encoded JWT is not userinfo but JWT.

We need a generic mechanism from mixer to propagate attributes to the app.

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,172 @@
...
+  // If any of match is true, it is matched.
+  repeated AttributeMatch match = 1;
+
+  message Policy {
+    oneof policy {
+      APIKey api_key = 1;
+      JWT jwt = 2;
+    }
+  }
+  // The list of auth policies to apply.
```
ayj: Copied from another thread, @lizan asked
> I expected that repeated in AuthPolicySpec is OR, no? needs clarification.

ayj: I think in the case of `APIKey` and `JWT` it is additive, e.g. we'd want to extract API key _and_ validate JWT. Do we need to extract AND/OR similar to https://github.com/envoyproxy/data-plane-api/pull/199/files#diff-1aaa6dc6fbc06b34d6a5bbb956032377R7?

mixer/v1/config/client/auth.proto
ayj: I was assuming mixerclient controller (config) would parse this out into respective sub-module config that is most efficient. Is that not possible with the current proposal?

mixer/v1/config/client/auth.proto
liminw: I think having AND/OR condition would be useful. As an example, API key may be required, but although you can specify multiple JWT issuers (e.g., google. firebase), we only require one of the JWT issuers. In this case, the policy would be (api_key and google) or (api_key and firebase).

liminw:
A high level question, is this going to be a Pilot config? It would be useful if you can provide some example configuration.

mixer/v1/config/client/auth.proto
ayj: It is current form `api-key` is just for extraction.

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,170 @@
...
+message AuthPolicySpec {
+  // If empty, apply auth policy to all requests.
+  // If any of match is true, it is matched.
+  repeated AttributeMatch match = 1;
+
+  message Policy {
+    oneof policy {
+      APIKey api_key = 1;
+      JWT jwt = 2;
+    }
```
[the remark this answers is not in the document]
ayj: Given that this is proxy (mixerclient) configuration I think mTLS would be configured through normal envoy filter config.

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,170 @@
...
+  // The list of auth policies to apply.
+  repeated Policy policies = 1;
+}
+
+// AuthPolicySpecBinding defines the binding between AuthPolicySpecs
+// and one or more IstioService.
+message AuthPolicySpecBinding {
+  // REQUIRED. One or more services to map the listed AuthPolicySpecs
+  // onto.
+  repeated IstioService services = 1;
```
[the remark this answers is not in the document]
ayj: The `api-key` config expressed here is more similar to attribute extraction a. la HTTP API spec. Does it make sense to scope that by mesh and namespace or even be included as part of auth policy? Maybe we should represent JWT and API-KEY separately and only scope JWT by namespace/mesh?

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// EndUserAuthPolicySpecBinding defines the binding between EndUserAuthPolicySpecs
+// and one or more IstioService.
+message EndUserAuthPolicySpecBinding {
+  // REQUIRED. One or more services to map the listed EndUserAuthPolicySpecs
+  // onto.
+  repeated IstioService services = 1;
+
+  message EndUserAuthPolicySpecReference {
```
spikecurtis: Missing a field that instantiates this message.

ayj:
/cc @apicl @spikecurtis 

I removed the `AttributeMatch` from the EndUserAuthenticationPolicy. AuthN is now scoped to the service only. Method-level validation is enforced by providing a list of JWT that can be matched against with `issuer`. The mixer is responsible for verifying whether method-level request includes appropriate auth attributes as defined by http://goo.gl/9dokCX.

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,137 @@
...
+  // system wide default is applied if no duration is explicitly
+  // specified.
+  google.protobuf.Duration public_key_cache_duration = 5;
+}
+
+// Determines how to apply auth policies for individual requests.
+message EndUserAuthenticationPolicySpec {
+  // List of JWT rules to valide.
+  //
+  // If the request includes a JWT it must match one of the JWT listed
```
qiwzhang: so config will not say if JWT is required or not for a method?   Only token is in the request, we will verify it,  if no,  we are OK, request will pass.   Is this OK?   Config for ESP can tell if JWT is required or not for a method.

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,137 @@
...
+  //
+  //     request.auth.presenter - The authorized presenter of the
+  //     credential. This value should reflect the optional Authorized
+  //     Presenter (`azp`) claim within a JWT
+  //
+  // If no match is found the request is rejected with HTTP status
+  // code 401.
+  //
+  // JWT validation is skipped if the user's traffic request does not
+  // include a JWT.
```
qiwzhang: Do we want users to specify where to extract JWT?   It is not always in "Authorization Bear".   For example,  IAP use a special HTTP header.

mixer/v1/config/client/auth.proto
spikecurtis: Whether or not JWT is required for a particular method should be part of the authorization layer, not authentication layer.

ayj: Correct, proxy/mixerclient only verifies user identity and returns 401 if that identify is not valid. ESP equivalent behavior should be handled by mixer adapter verifying correct identify is presented for particular method (or service, etc.).

mixer/v1/config/client/auth.proto
ayj: I added a location specification to JWT message to indicate where in HTTP request token should be extracted from.

mixer/v1/config/client/auth.proto
```diff
@@ -83,6 +83,29 @@ message JWT {
   // system wide default is applied if no duration is explicitly
   // specified.
   google.protobuf.Duration public_key_cache_duration = 5;
+
+  // Defines where to extract the JWT from an HTTP request.
+  //
+  // The Authorization header using the Bearer schema will be used if
+  // no explicit location is specified, e.g. Authorization: Bearer
+  // <token>.
```
qiwzhang: If not specified, we also check "access_token" in the query too

mixer/v1/config/client/auth.proto
```diff
@@ -83,6 +83,29 @@ message JWT {
...
+  message Location {
+    oneof scheme {
+      // JWT is sent in a request header. `header` represents the
+      // header name.
+      //
+      // For example, `header=x-goog-iap-jwt-assertion`
```
qiwzhang: If header=x-goog-iap-jwt-assertion,  the header format will be:
x-goog-iap-jwt-assertion: <token>

mixer/v1/config/client/auth.proto
```diff
@@ -83,6 +83,29 @@ message JWT {
...
+      //
+      string header = 1;
+
+      // JWT is sent in a query parameter. `query` represents the
+      // query parameter name.
+      //
+      // For example, `query=access_token`.
```
qiwzhang: access_token is default.   let us use other query as example
if query=jwt_token,   the format will be   ?jwt_token=<token>

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,160 @@
...
+message JWT {
+  // Identifies the principal that issued the JWT. See
+  // https://tools.ietf.org/html/rfc7519#section-4.1.1
+  // Usually a URL or an email address.
+  //
+  // Example: https://securetoken.google.com
+  // Example: 1234567-compute@developer.gserviceaccount.com
+  //
+  string issuer = 1;
+
```
apicl: Can we support regular expression?

mixer/v1/config/client/api_spec.proto
```diff
@@ -123,6 +136,47 @@ message HTTPAPISpecPattern {
...
+  oneof key {
+    // API Key is sent as a query parameter. `query` represents the
+    // query string parameter name.
+    //
+    // For example, `query=api_key` should be used with the
+    // following request:
+    //
+    //     GET /something?api_key=abcdef12345
+    //
+    string query = 1;
```
apicl: change name to query_parameter?

mixer/v1/config/client/api_spec.proto
```diff
@@ -123,6 +136,47 @@ message HTTPAPISpecPattern {
   }
 }
 
+// APIKey defines the explicit configuration for generating the
+// `request.api_key` attribute from HTTP requests.
+//
+// See https://swagger.io/docs/specification/authentication/api-keys
+// for a general overview of API keys as defined by OpenAPI.
+message APIKey {
```
apicl: ApiKeyLocation?

mixer/v1/config/client/auth.proto
```diff
@@ -0,0 +1,160 @@
...
+  // Optional namespace of the
+  // EndUserAuthenticationPolicySpec. Defaults to the value of the
+  // metadata namespace field.
+  string namespace = 2;
+}
+
+// EndUserAuthenticationPolicySpecBinding defines the binding between
+// EndUserAuthenticationPolicySpecs and one or more IstioService.
+message EndUserAuthenticationPolicySpecBinding {
```
apicl: What is this used for?

mixer/v1/config/client/api_spec.proto
ayj: This naming mirrors API-KEY terminology which is derived from OpenAPI. See "Describing API Keys" in https://swagger.io/docs/specification/authentication/api-keys.

mixer/v1/config/client/auth.proto
ayj: Bindings associate an EndUserAuthenticationPolicy with a particular. It is structured this way so that a single policy could be applied to multiple services. This approach was used for consistency with HTTPAPISpec and HTTPAPISpecBinding.

mixer/v1/config/client/auth.proto
ayj: @lizan @qiwzhang

mixer/v1/config/client/auth.proto
apicl: Should this be part of the authz and enforced in mixer?

mixer/v1/config/client/api_spec.proto
apicl: The doc says that the key can be sent in the query string, but does not say where.  I think query_parameter is an accurate name.

ayj: The method of describing the API-KEY location uses `query` (see excerpt below).

```
# 1) Define the key name and location
components:
  securitySchemes:
    ApiKeyAuth:        # arbitrary name for the security scheme
      type: apiKey
      in: header       # can be "header", "query" or "cookie"
      name: X-API-KEY  # name of the header, query parameter or cookie
```

mixer/v1/config/client/auth.proto
ayj: Binding is used by Pilot to distribute the policy to the correct set of proxies. Without this Pilot would need to distribute _all_ auth policies to _all_ proxies which doesn't scale.

mixer/v1/config/client/api_spec.proto
apicl: SG

mixer/v1/config/client/auth.proto
apicl: I am OK to check in this as first version. We can extend the support later.

mixer/v1/config/client/auth.proto
apicl: SG.

ayj:
Introduces proxy/mixerclient runtime configuration for JWT validation and API-KEY attribute generation. 

JWT config is based on existing JWT filter (see https://github.com/istio/proxy/blob/master/src/envoy/auth/http_filter_config.cc)

API-KEY config is based on https://goo.gl/TdRL8a and the OpenAPI spec.

EndUserAuthenticationPolicy services two purposes:

1) internal config that is generated/distributed by Pilot and consumed by Proxy.
2) user-facing config that user could define themselves and/or be derived from something like OpenAPI. 

(2) is temporary until a overall SecurityPolicy is available.

PR #229, opened 2017-11-09
Add blacklisted services for mTLS to mesh config.

diemtvu:
These fields is to (always) disable outbound authentication policy for some services (and all services in some namespace).

Candidate implementation to use these fields is https://github.com/istio/istio/pull/1607


proxy/v1/config/proxy_mesh.proto
```diff
@@ -190,4 +190,16 @@ message MeshConfig {
...
   // at runtime and config gets distributed dynamically.
   ProxyConfig default_config = 14;
+
+  // List of names of services that should not have auth enable. Typically,
+  // this is a list of control services (e.g k8s API server) (for regular
+  // user services, use service annotation to overwrite the authentication
+  // policy). These services don't have envoy in front, so the configuration
+  // will affect the client side only (i.e outbound clusters)
+  // This field has no effect if the policy is already 'NONE' (auth disabled).
+  repeated string auth_blacklisted_destination_service = 15;
```
rshriram: rather than having a service and a namespace (and leaking namespace into the proxy/envoy code), why not have a single set of blacklisted service fqdns? you can have regular expressions here (or just simple wildcards like *.bookinfo.com. In Pilot, you can easily match a service's hostname with the regex to see if it matches and then disable.

proxy/v1/config/proxy_mesh.proto
```diff
@@ -190,4 +190,12 @@ message MeshConfig {
...
   // at runtime and config gets distributed dynamically.
   ProxyConfig default_config = 14;
+
+  // List of names of services that should not have auth enable. Typically,
+  // these are control services (e.g k8s API server) and doesn't have envoy in
+  // front. This field will affect the client side config only (i.e outbound
+  // clusters). It has no effect if the policy is already 'NONE' (auth disabled).
+  // Do NOT use this for services that have envoy (instead, please use service
+  // annotation to overwrite the authentication policy).
+  repeated string auth_blacklisted_destination_service = 15;
```
rshriram: rename to mtls_excluded_services .

proxy/v1/config/proxy_mesh.proto
```diff
@@ -190,4 +190,12 @@ message MeshConfig {
   // and remain constant for the duration of the pod. The rest of the mesh config can be changed
   // at runtime and config gets distributed dynamically.
   ProxyConfig default_config = 14;
+
+  // List of names of services that should not have auth enable. Typically,
```
rshriram: rephrasing comments.
```
+ // List of remote services for which mTLS authentication should not be expected by Istio . Typically,
 +  // these are control services (e.g kubernetes API server) that don't have Istio sidecar
// to transparently terminate mTLS authentication.
It has no effect if the authentication policy is already 'NONE'.
 +  // DO NOT use this setting for services that are managed by Istio (i.e. using Istio sidecar). Instead, use service-level annotations to overwrite the authentication policy.
```

andraxylia:
@diemtvu Can you please update the generated .go files as well? Changes in proto files require an update of the generated files, we use them for the go build.

diemtvu:
@andraxylia Sorry, I missed it. But should it be updated in https://github.com/istio/api/pull/233 ? I run scripts/generate-protos.sh on my machine but see no pb.go file changes.

andraxylia:
Np, it's a newer requirement. It's all good, Sunny just updated them.

PR #231, opened 2017-11-10
Separate http client config with tcp.

qiwzhang:
Currently TCP is using the same HTTP client config.  It is a hack. This PR will try to separate them.
The changes are:
* Renamed mixer_filter_config to client_config  in the file name and message name
* Rename MixerControlConfig as ServiceConfig.  It is under mixer::v1::config::client name space so it will not cause confusing.
* Add a new message TransportConfig for cache and network policy
* Create separate HttpClientConfig and TcpClientConfig.
* Add disable_report_calls for tcp too.

### DIFF
--- a/mixer/v1/config/client/client_config.proto
+++ b/mixer/v1/config/client/client_config.proto
@@ -105,4 +105,7 @@ message TcpClientConfig {
 
   // If set to true, disables mixer check calls.
   bool disable_report_calls = 4;
+
+  // Quota specifications to generate quota requirements.
+  QuotaSpec quota_spec = 5;
 }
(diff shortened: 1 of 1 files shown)


mixer/v1/config/client/client_config.proto
```diff
@@ -65,25 +64,45 @@ message MixerFilterConfig {
...
+// Defines the client config for HTTP.
+message HttpClientConfig {
+  // The transport config.
+  TransportConfig transport = 1;
 
   // Map of control configuration indexed by destination.service. This
   // is used to support per-service configuration for cases where a
   // mixerclient serves multiple services.
-  map<string, MixerControlConfig> control_configs = 5;
+  map<string, ServiceConfig> control_configs = 2;
```
ayj: s/control_configs/service_configs

qiwzhang: Done

qiwzhang:
Here is quota per connection.

PR #233, opened 2017-11-11
Add scripts to auto-generate protobuf golang code

ayj:
This is mostly copied over from istio.io/istio/mixer/ with tweaks to
adjust the paths so that *.pb.go files appears alongside the *.proto
files.

Developers can run `./script/generate.sh` to (re)generate the latest *.pb.go files. Next steps would be to  run the script from presubmit and verify that protos compile cleanly and there are no unstaged changes as result (e.g. developer didn't forget to refresh).


ayj:
/cc @mandarjog for mixer protos

@rshriram, what's needed to add the presubmit validation check to CircleCI?

rshriram:
Take a look at .circleci/config.yml. It’s just a series of commands. Just
call the validation check script as a separate step in the unittest job

qiwzhang:
@ayj This PR broke mixerclient.   Basically,  we could not put BUILD file into proto folders where c++ code is using.   C++ code is using different BUILD file.  Here is c++ code https://github.com/istio/mixerclient/blob/master/repositories.bzl#L397
It cause bazel to generate error:  crosses boundary of subpackage '@mixerapi_git//mixer/v1'

ayj:
Can we add the mixerclient cc_proto_library code to istio/api? This seems to be the approach taken by envoyproxy (see https://github.com/envoyproxy/data-plane-api/blob/master/api/BUILD and https://github.com/envoyproxy/data-plane-api/blob/master/bazel/api_build_system.bzl#L29).

qiwzhang:
@kyessenov has tried it.  He failed.  It is very complicated.  
@lizan and @kyessenov knows the detail why it failed.

kyessenov:
Hey, I didn't fail, rather abandoned it since I thought api is getting mono-repoized, too. Envoy merged the patch to simplify the bazel protobuf setup.

andraxylia:
@ayj thanks for adding this!

PR #234, opened 2017-11-12
cleanup: rename IngressRule to K8SIngressRule

rshriram:
NOTE: I have not updated the associated pb.go file. Working remotely, and I don't have a high bandwidth internet connection. Don't want to exhaust my quota running bazel.

### DIFF
--- a/proxy/v1/config/ingress_rule.proto
+++ b/proxy/v1/config/ingress_rule.proto
@@ -18,19 +18,9 @@ import "proxy/v1/config/route_rule.proto";
 
 package istio.proxy.v1.config;
 
-// Ingress rules are routing rules applied to the ingress proxy pool. The
-// ingress proxes serve as the receiving edge proxy for the entire mesh, but
-// can also be addressed from inside the mesh.  Each ingress rule defines a
-// destination service and port. Rules that do not resolve to a service or a
-// port in the mesh should be ignored.
-//
-// The routing rules for the destination service are applied at the ingress
-// proxy. That means the routing rule match conditions are composed and its
-// actions are enforced. The traffic splitting for the destination service is
-// also effective.
-//
-// WARNING: This API is experimental and under active development
-message IngressRule {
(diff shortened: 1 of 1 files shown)


ZackButcher:
Do we need this change atm? We have `Gateway`, `RouteRule(++)`, and `DestinationRule`. Hopefully we can do away with `IngressRule` entirely. I thought the main motivation for this was https://github.com/istio/api/pull/227, but we've shifted to calling that config object `Gateway` rather than `IngressRule` as originally proposed.

rshriram:
Don’t we still need to maintain compatibility with whatever Ingress implementation we have (for k8s ingress controller)? Essentially this should not be documented. It’s for internal code use in proxy/Envoy/ingress.go 

@kyessenov WDYT?

kyessenov:
Can we map k8s ingress into gateways and whatnot reliably? If the answer is yes, then we can deprecate the internal IngressRule.

rshriram:
That’s a good idea. I think we can if we stick to strict mapping without
trying to infer the semantics of additional ports opened in the ingress. ie
ingress is just a gateway with port 80/443 and routing rules based on
existing transformation (path and prefix in the combinePathPrefix
function). We should not try to be smart and map other ports in the ingress
lb to internal services implicitly as the entire concept of gateway is
about explicit declaration of what to expose.

And if we do this mapping, we can actually get rid of this entire ingress
rule struct as we retain functionally through other means. We use this
purely for coding  convenience.

linsun:
I don't understand why we need to introduce this change and what benefit we will gain from doing this.  Please explain.

kyessenov:
This seems unnecessary.

rshriram:
It is because this is an implementation/platform specific detail. We don't have a ingress resource. Having an Ingress rule leads to confusion for people browsing through the repo. "If" we do happen to have an ingress rule in future that is cross platform, this will conflict.

PR #235, opened 2017-11-12
cleanup: Create separate message for port spec

rshriram:
Format is similar to K8S structs with name, port and protocol. Once again, there is no proto commit here, as I don't have the bandwidth required to run bazel (it downloads go and other stuff).


vadimeisenbergibm:
Should the protocol be an ENUM? Related to https://github.com/istio/istio/issues/1664.
And what is the port name used for?

ZackButcher:
Protocol is a string for extensibility (as we add support for more protocols, or the user builds their own codecs, less has to change for things to work). I have mixed feelings about it; I like enums for the strong typing, but in this case I'm not sure they actually add much.

Name is to support environments with named ports, like k8s.

rshriram:
Someone needs to update the pb.go files after all these small mods. Otherwise you cannot import this api sha in Istio/istio.

mandarjog:
It’s too late now, but do not merge until the pb.go is updated for subsequent PRs.

rshriram:
may be put this in presubmit?

proxy/v1/config/route_rule.proto
```diff
@@ -493,3 +493,15 @@ message CorsPolicy {
...
+// Port describes the properties of a specific port of a service.
+message Port {
+  // A valid non-negative integer port number.
+  int32 number = 1;
+  // The protocol to communicate with the external services.
+  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|REDIS|TCP.
+  string protocol = 2;
+
+  // Name assigned to the port.
+  string name = 3;
```
kyessenov: there should be oneof constraint - name or number

proxy/v1/config/route_rule.proto
```diff
@@ -493,3 +493,15 @@ message CorsPolicy {
...
   google.protobuf.BoolValue allow_credentials = 6;
 }
+
+// Port describes the properties of a specific port of a service.
+message Port {
+  // A valid non-negative integer port number.
+  int32 number = 1;
+  // The protocol to communicate with the external services.
+  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|REDIS|TCP.
+  string protocol = 2;
```
kyessenov: A single string will probably not work either - what exactly is TCP over gRPC over HTTPS :)

rshriram:
On Tue, Nov 14, 2017 at 9:06 AM Kuat <notifications@github.com> wrote:

> *@kyessenov* commented on this pull request.
>
> In proxy/v1/config/route_rule.proto
> <https://github.com/istio/api/pull/235#discussion_r150730226>:
>
> > @@ -493,3 +493,15 @@ message CorsPolicy {
>    // Access-Control-Allow-Credentials header.
>    google.protobuf.BoolValue allow_credentials = 6;
>  }
> +
> +// Port describes the properties of a specific port of a service.
> +message Port {
> +  // A valid non-negative integer port number.
> +  int32 number = 1;
> +  // The protocol to communicate with the external services.
> +  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|REDIS|TCP.
> +  string protocol = 2;
> +
> +  // Name assigned to the port.
> +  string name = 3;
>
> there should be oneof constraint - name or number
>

We can’t because egress rule is effectively a port spec. The name field is
actually not necessary, but I thought I will include it for other uses.
OTOH port in destination or gateway is a condition match against ground
truth (from platform). So name or port makes sense there. May be we should
have two types of ports. One a spec and another a match.

vadimeisenbergibm:
@rshriram @mandarjog sorry, my bad, figured it out and submitted https://github.com/istio/api/pull/237.

bobbytables:
This breaks `istio/istio` at this line: https://github.com/istio/istio/blob/master/pilot/model/validation.go?utf8=%E2%9C%93#L868

rshriram:
Sorry, this was merged prematurely. We are reverting this PR.

bobbytables:
> Sorry, this was merged prematurely. We are reverting this PR.

@rshriram - When will Pilot get these changes merged in? We're building something against Pilot's service endpoint right now and a proto change would cause problems for us so knowing what the plan is here would be great if you can.

rshriram:
It will be done shortly. That said, Pilot has no API (or service endpoint/service model API). So, technically, you should not have to worry about API/proto inconsistencies. Can you give more details on  what you are building so that we can keep you informed if anything here changes?

bobbytables:
I think I just misspoke is all, we're building something that dynamically creates RouteRules for services using the packages in `istio/istio`

rshriram:
Ah.. you mean istio/api ? We have fixed the issue. Sorry for the trouble. Let me know if the error still persists.

PR #236, opened 2017-11-13
adding http port configuration on ingress #1618

linsun:
Change in pilot is coming soon. This is for https://github.com/istio/istio/issues/1618


proxy/v1/config/proxy_mesh.proto
```diff
@@ -198,4 +200,10 @@ message MeshConfig {
   // DO NOT use this setting for services that are managed by Istio (i.e. using Istio sidecar).
   // Instead, use service-level annotations to overwrite the authentication policy.
   repeated string mtls_excluded_services = 15;
+
+  // Port on which Istio Ingress should listen for inbound http traffic.
+  int32 ingress_http_port = 16;
+
+  // Port on which Istio Ingress should listen for inbound https traffic.
+  int32 ingress_https_port = 17;
```
rshriram: This is non standard. K8s ingress runs only on 80 and 443 afaik. If anything these settings should be picked up from the service spec.

We have been trying to define a generic load balancer configuration that allows specifying the ports to listen on. See the other wip PRs I have.

kyessenov: @linsun what is the context of this change?
@rshriram is right that k8s requires 80 and 443 only for ingress.

linsun: @rshriram @kyessenov this is before the new thing on standalone envoy as ingress controller is available and stable, we need a way to use nginx and envoy together, like someone indicated on the user's mailing list: https://groups.google.com/forum/#!topic/istio-users/EEhKrspuH10

linsun: Please note the default is still 80 and 443 for ingress.  This patch will simply enable istio adopters to adopt istio using their preferred way before the new standalone envoy as ingress controller is in place.

rshriram:
On Wed, Nov 15, 2017 at 8:38 AM Lin Sun <notifications@github.com> wrote:

> *@linsun* commented on this pull request.
>
> In proxy/v1/config/proxy_mesh.proto
> <https://github.com/istio/api/pull/236#discussion_r151024521>:
>
> > @@ -198,4 +200,10 @@ message MeshConfig {
>    // DO NOT use this setting for services that are managed by Istio (i.e. using Istio sidecar).
>    // Instead, use service-level annotations to overwrite the authentication policy.
>    repeated string mtls_excluded_services = 15;
> +
> +  // Port on which Istio Ingress should listen for inbound http traffic.
> +  int32 ingress_http_port = 16;
> +
> +  // Port on which Istio Ingress should listen for inbound https traffic.
> +  int32 ingress_https_port = 17;
>
> Please note the default is still 80 and 443 for ingress. This patch will
> simply enable istio adopters to adopt istio using their preferred way
> before the new standalone envoy as ingress controller is in place.
>

Sure but this is not the way to go about this. This is istio global
configuration that is going to last a while. We don’t want to add temporary
things that we are going to throw away.

The gateway resource we are working on will not be the same as ingress. It
will continue to work with existing ingress but it is meant to act as a
stand-alone Envoy.

With regard to that user, he does not need to tweak Envoy ingress nor does
he need to run Envoy in ingress mode. My suggestion to him was very simple
- use orig dst cluster and pass everything through. He just wanted metrics.

linsun:
@rshriram please suggest a better way to do this, I'm open for ideas.  We need this in ibm cloud to enable our nginx and envoy together like i explained to you earlier.  While this is temporary, I expect we'll run with this mode for ~6 months till istio ingress is proven to be useful and capable like nginx.

linsun:
@ZackButcher @ijsnellf and I had a discussion on this today, we think this is the most simple approach to enable the feature I needed for the moment before our istio-ingress is usable when comparing with nginx.  @ZackButcher suggested to use a different name to avoid confusion, which I think is a good idea.  I don't need https so I'll remove that too.   Essentially, this will be a small mesh configuration for people who wants to wire nginx with envoy.

linsun:
@rshriram PTAL.

rshriram:
If all you want is a http port to receive traffic for all internal services, just set the https://github.com/istio/api/pull/236/files#diff-8e43ca81cda6fe9a3cece9a939db8025R129 proxy_http_port to the required port value. Deploy Envoy in sidecar mode. You don’t even need ip tables for your setup.

linsun:
@rshriram i tried that before, i can't get that working.  

```istio-proxy@public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1-600196534-8rg4f:/$ curl localhost:15001/productpage
curl: (52) Empty reply from server```

The main thing I see is I don't see the ingress routed correctly in this mode for things like /productpage.  Note the 73c1a0e2ecccab948256b3f16766c3fc61ba3e07 cluster is the nginx+envoy itself.

```
istio-proxy@public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1-600196534-8rg4f:/$ curl istio-pilot.istio-system:8080/v1/routes/80/istio-proxy/sidecar~~public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1-600196534-8rg4f.istio-system~istio-system.svc.cluster.local
{
  "virtual_hosts": [
   {
    "name": "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system.svc.cluster.local|http",
    "domains": [
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system:80",
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system",
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system.svc:80",
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system.svc",
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system.svc.cluster:80",
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system.svc.cluster",
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system.svc.cluster.local:80",
     "public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1.kube-system.svc.cluster.local",
     "172.21.227.67:80",
     "172.21.227.67"
    ],
    "routes": [
     {
      "prefix": "/",
      "cluster": "out.73c1a0e2ecccab948256b3f16766c3fc61ba3e07",
      "decorator": {
       "operation": "default-route"
      }
     }
    ]
   }
  ]
```

```
istio-proxy@public-cr1de3810db0024cb4a078eea8ba7bd36c-alb1-600196534-8rg4f:/$ curl istio-pilot.istio-system:8080/v1/routes/15001/istio-proxy/sidecar~~public-cr1de3810db0024cb4a8eea8ba7bd36c-alb1-600196534-8rg4f.istio-system~istio-system.svc.cluster.local
null```

ZackButcher:
@linsun I didn't understand when we talked that the real concern was port collision (wanting to host two proxies on the same host); I thought we were talking about the higher level use case of nginx edge routing to an Istio (Envoy) mesh edge.

If the goal of this change is to enable the one-off use case of Envoy listening locally on a specific port, then I think it'd be much better to implement that by creating a custom istio_proxy image with a custom Envoy bootstrap config. That config should be static; just a listener on localhost on a specific port. All the normal routing via Pilot should work. The advantage to handling this with static config is that it'll save us having to wire things up in Pilot, and also from cluttering the mesh config for one-off use cases. It's hard for us to retire fields from global config like this if we're maintaining backwards compatibility, so this proposal is not an insignificant amount of technical debt.

rshriram:
@linsun you just need a custom pilot agent that omits LDS and uses RDS/CDS. Pilot agent is not really a core part of istio. Implementors are free to customize and deploy their own agents or use static configurations. So, for your use case, just create a custom pilot agent that generates configuration like this:

```json
{
  "listeners": [
   {
    "address": "tcp://0.0.0.0:8080",
    "name": "http_0.0.0.0_8080",
    "filters": [
     {
      "type": "read",
      "name": "http_connection_manager",
      "config": {
       "codec_type": "auto",
       "stat_prefix": "http",
       "generate_request_id": true,
       "tracing": {
        "operation_name": "egress"
       },
       "rds": {
        "cluster": "rds",
        "route_config_name": "80",
        "refresh_delay_ms": 10
       },
       "filters": [
        {
         "type": "",
         "name": "cors",
         "config": {}
        },
        {
         "type": "decoder",
         "name": "router",
         "config": {}
        }
       ],
       "access_log": [
        {
         "path": "/var/log/istio/envoy-access.log"
        }
       ]
      }
     }
    ],
    "bind_to_port": true
   }
  ],
  "admin": {
    "access_log_path": "/var/log/istio/envoy-admin.log",
    "address": "tcp://127.0.0.1:15000"
  },
  "cluster_manager": {
    "clusters": [
      {
        "name": "rds",
        "connect_timeout_ms": 1000,
        "type": "strict_dns",
        "lb_type": "round_robin",
        "hosts": [
          {
            "url": "tcp://istio-pilot.istio-system:15003"
          }
        ]
      },
      {
        "name": "zipkin",
        "connect_timeout_ms": 1000,
        "type": "strict_dns",
        "lb_type": "round_robin",
        "hosts": [
          {
            "url": "tcp://localhost:6000"
          }
        ]
      }
    ],
    "sds": {
      "cluster": {
        "name": "sds",
        "connect_timeout_ms": 1000,
        "type": "strict_dns",
        "lb_type": "round_robin",
        "hosts": [
          {
            "url": "tcp://istio-pilot.istio-system:15003"
          }
        ]
      },
      "refresh_delay_ms": 10
    },
    "cds": {
      "cluster": {
        "name": "cds",
        "connect_timeout_ms": 1000,
        "type": "strict_dns",
        "lb_type": "round_robin",
        "hosts": [
          {
            "url": "tcp://istio-pilot.istio-system:15003"
          }
        ]
      },
      "refresh_delay_ms": 10
    }
  },
  "tracing": {
    "http": {
      "driver": {
        "type": "zipkin",
        "config": {
          "collector_cluster": "zipkin",
          "collector_endpoint": "/api/v1/spans"
        }
      }
    }
  }
}
```

Most of this machinery is already there. You just need to generate static listeners. A more and simple hackish way is to just concatenate two JSONs (from lds-ingress.json and envoy-sidecar.json) and start up envoy with the appropriate service cluster

rshriram:
please get approval from @ZackButcher or @louiscryan for this use case. IF they are okay with this, I am okay as well.

linsun:
@rshriram @ZackButcher thanks for the suggestion.  Generating a static config and built into the image doesn't work with all various istio config, mTLS or not, mixer or not.  This is not an one off scenario... I expect people to have nginx or kong in front of envoy for a while before they can completely replace envoy with nginx or kong.

linsun:
Closing this, thank you @rshriram @ZackButcher @ijsnellf , I was able to get pilot agent configured with what I wanted.  We'll use that method instead.

PR #240, opened 2017-11-15
Add cc_proto into mixer api and config proto BUILD files.

qiwzhang:


mixer/v1/BUILD
```diff
@@ -1,4 +1,5 @@
 load("@org_pubref_rules_protobuf//gogo:rules.bzl", "gogoslick_proto_library")
+load("@com_google_protobuf//:protobuf.bzl", "cc_proto_library")
```
lizan: why not use native ones?

googleapis.bzl
```diff
@@ -84,10 +84,105 @@ filegroup(
     ],
 )
 
+load("@com_google_protobuf//:protobuf.bzl", "cc_proto_library")
+
+exports_files(glob(["google/**"]))
+
+cc_proto_library(
```
ayj: Are `servicecontrol`, `service_config`, and `cloud_trace` targets used by mixerclient or proxy? A quick grep didn't turn up any meaningful results.

mixer/v1/BUILD
qiwzhang: We needs two proto files; duration.proto and timestamp.proto from protobuf repo.  But protobuf repo is not using native cc_proto_library.

googleapis.bzl
qiwzhang: You are right.  Only rpc/status.proto is used. I cleaned it up.

qiwzhang:
@ayj PTAL

PR #243, opened 2017-11-16
Better routing and Gateway continued from Shrirams PR

louiscryan:
Cloned Shrirams and updated to largely match state of discussion doc and noted decisions.
Remaining work to cleanup per-protocol cruft

### DIFF
--- a/proxy/v1/config/egress_rule.proto
+++ b/proxy/v1/config/egress_rule.proto
@@ -43,6 +43,15 @@ package istio.proxy.v1.config;
 //           protocol: https
 //
 message EgressRule {
+  // Port describes the properties of a specific TCP port of an external service.
+  message Port {
+    // A valid non-negative integer port number.
+    int32 port = 1;
+    // The protocol to communicate with the external services.
+    // MUST BE one of HTTP|HTTPS|GRPC|HTTP2.
+    repeated string protocol = 2;
+  }
+
   // REQUIRED: Hostname or a wildcard domain name associated with the external service.
   // ONLY the "service" field of destination will be taken into consideration. Name,
   // namespace, domain and labels are ignored. Routing rules and destination policies that
--- a/proxy/v1/config/gateway.proto
+++ b/proxy/v1/config/gateway.proto
@@ -164,10 +164,15 @@ message Server {
   // connections
   Port port = 1;
 
-  // REQUIRED: A list of domains exposed by this gateway. While
+  // A list of domains exposed by this gateway. While
   // typically applicable to HTTP services, it can also be used for TCP
-  // services using TLS with SNI.  Standard DNS wildcard prefix syntax
+  // services using TLS with SNI. Standard DNS wildcard prefix syntax
   // is permitted.
+  //
+  // RouteRules that are bound to a gateway must having a matching domain
+  // in their default destination. Specifically one of the route rule
+  // destination domains is a strict suffix of a gateway domain or
+  // a gateway domain is a suffix of one of the route rule domains.
   repeated string domains = 2;
 
   message TLSOptions {
--- a/proxy/v1/config/gateway.proto
+++ b/proxy/v1/config/gateway.proto
@@ -1,0 +1,247 @@
+// Copyright 2017 Istio Authors
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "proxy/v1/config/route_rule.proto";
+
+package istio.proxy.v1.config;
+
+// Gateway describes a load balancer operating within the mesh that bridges
+// traffic across network boundaries. Supported use-cases include
+// ingress, egress and middle proxies within the mesh.
+
+// The specification describes a set of ports that are exposed, the protocols
+// allowed on each port and TLs configuration etc.
+//
+// For example, the following Gateway spec sets up Envoy to act as a
+// load balancer exposing port 80 (http), 443 (https), and port 2379 (TCP).
+// While Istio will configure Envoy to listen on these ports, it is the
+// responsibility mesh operator to ensure that external traffic can reach these
+// ports.
+//
+//     metadata:
+//       name: my-gateway
+//     spec:
+//       servers:
+//       - port:
+//           number: 80
+//           protocol: http
+//         domains:
+//           - uk.bookinfo.com
+//           - eu.bookinfo.com
+//           - * #default
+//         tls:
+//           httpsRedirect: true # sends 302 redirect for http requests
+//       - port:
+//           number: 443
+//           protocol: https
+//         tls:
+//           mode: simple #enables HTTPS on this port
+//         hosts:
+//           - uk.bookinfo.com
+//           - eu.bookinfo.com
+//           - *
+//       - port:
+//           number: 2379  #to expose internal service via external port 2379
+//           protocol: redis
+//         tls:
+//           mode: mtls
+//         hosts:
+//           - redis.bookinfo.com
+//       certs:
+//       - hosts:
+//           - *.bookinfo.com
+//         serverCert: server.crt
+//         clientCABundle: client.ca-bundle
+//
+// The following routing rule exposes uk.bookinfo.com, eu.bookinfo.com via the
+// gateway and routes all calls to /reviews path in the reviews service.
+//
+//     metadata:
+//       name: bookinfo-rule
+//     spec:
+//       destination:
+//         name: reviews.prod
+//         domains:
+//           - uk.bookinfo.com
+//           - eu.bookinfo.com
+//       match:
+//         request:
+//           headers:
+//             uri:
+//               prefix: /reviews/
+//       route:
+//       - destination:
+//           name: reviews.prod  # could be elided
+//         weight: 80
+//       - destination:
+//           name: reviews.qa
+//         weight: 20
+//
+// By default, if there is no wildcard, requests for unknown hosts
+// will be responded with a 404. When some default behavior is desired, add
+// a default host, "*", to the Gateway specification and specify its
+// behavior with an appropriate route rule. For example,
+//
+//     metadata:
+//       name: default-ingress
+//     spec:
+//       destination:
+//         domain:
+//           - *
+//       route:
+//       - destination:
+//           name: homepage
+//
+// Note that the routing rules above will be applicable inside the service
+// mesh as well, to all services invoking the reviews API via the HTTP host
+// uk.bookinfo.com, and eu.bookinfo.com. In addition, the wildcard rule
+// above will force all internal services to go to the homepage service. In
+// order to restrict the wildcard routing rule above to be applicable only
+// at the Ingress load balancer, use the source match option such that the
+// rule will be applicable only to instances belonging to the load
+// balancer.
+//
+// For example, let us assume that the ingress pods/VMs have a label called
+// role: ingress
+//
+//     metadata:
+//       name: default-ingress
+//     spec:
+//       destination:
+//         fqdn:
+//           - *
+//       match:
+//         source:
+//           labels:
+//             role: ingress # Restrict this rule to instances with this label only
+//       route:
+//       - destination:
+//           name: homepage
+//
+message Gateway {
+  // REQUIRED: Details about the port on which the proxy listens and applies rules
+  repeated Server servers = 1;
+
+  // Optional names of certificates for the server and the CA bundle to be used to validate
+  // the client's certificate (if mTLS is to be used).
+  repeated TLSConfig certs = 2;
+}
+
+// Server describes the properties of the proxy on a given load balancer port.
+// For example,
+//     metadata:
+//       name: my-ingress
+//     spec:
+//       servers:
+//       - port:
+//           number: 80
+//           protocol: HTTP2
+//
+// Another example
+//     metadata:
+//       name: my-tcp-ingress
+//     spec:
+//       servers:
+//       - port:
+//           number: 2379
+//           protocol: redis
+
+message Server {
+  // REQUIRED: The port on which the gateway should listen for incoming
+  // connections
+  Port port = 1;
+
+  // REQUIRED: A list of domains exposed by this gateway. While
+  // typically applicable to HTTP services, it can also be used for TCP
+  // services using TLS with SNI.  Standard DNS wildcard prefix syntax
+  // is permitted.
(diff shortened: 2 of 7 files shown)


louiscryan:
@ZackButcher @rshriram @frankbu @kyessenov 

Continuing here while Shriram is travelling

proxy/v1/config/egress_rule.proto
```diff
@@ -43,6 +43,15 @@ package istio.proxy.v1.config;
 //           protocol: https
 //
 message EgressRule {
+  // Port describes the properties of a specific TCP port of an external service.
+  message Port {
```
frankbu: This seems messed up. This one message supposed to be moved to top level in route_rule.proto.

proxy/v1/config/gateway.proto
```diff
@@ -0,0 +1,247 @@
...
+//       destination:
+//         name: reviews.prod
+//         domains:
+//           - uk.bookinfo.com
+//           - eu.bookinfo.com
+//       match:
+//         request:
+//           headers:
+//             uri:
+//               prefix: /reviews/
```
frankbu: I'm still wondering where the port mapping is specified for this example. The Gateway is exposing port 80 and forwarding `/review` requests to the reviews service, but where do we specify the port of the target service. In this particular example (bookinfo), the target port is 9080.

proxy/v1/config/egress_rule.proto
ZackButcher: It's in route_rule.proto today: https://github.com/istio/api/blob/master/proxy/v1/config/route_rule.proto#L498, we should re-use it.

Also the comment lost a few protocols: `HTTP|HTTPS|GRPC|HTTP2|MONGO|REDIS|TCP` in the original.

proxy/v1/config/route_rule.proto
```diff
@@ -153,74 +191,47 @@ message RouteRule {
...
+  IPMatchAttributes match = 2 ;
 
-  // Domain suffix used to construct the service FQDN in implementations that support such specification.
-  string domain = 3;
+  repeated DestinationWeight route = 3;
+}
 
-  // The service FQDN.
-  string service = 4;
+message WebsocketRoute {
```
ZackButcher: How does this differ from an `HttpRoute` with `websocket_upgrade = true`?

proxy/v1/config/route_rule.proto
```diff
@@ -79,50 +79,88 @@ package istio.proxy.v1.config;
 //         weight: 100
 //
 message RouteRule {
```
ZackButcher: I like this idea of a single route rule describing all of the variations of traffic to a specific destination. However, I'm not 100% on board with requiring a new message per route type. Ultimately I think it'll slow down our ability to support more protocols, and make it harder for users to prototype new protocols themselves. Below is a strawman alternative that combines the Routes. I'm not really happy with it either, but I'll keep it as food for thought.

<hr>

A lot of the new `*Route` messages are the same: `precedence`, `destination`, `condition`; can we extract that out to a `Route`, then use something like a `oneof` for the options? It still doesn't address the root concern of requiring quite a bit of wiring in Pilot to support new protocols.

```proto
message Route {
  int32 precedence = 1;
  MatchCondition match = 2;
  // dest port has a protocol on it, so we know how to handle the traffic
  Destination destination = 3;

  oneof protocol_options {
    HttpOptions http_options = 4;
    // potentially support a backdoor here that allows the user to insert Envoy config directly?
  }
}

message HttpOptions {
   // all the HTTP stuff in HttpRoute now
}

message RouteRule {
  Destination default = 1;
  repeated Route routes = 2;
}
```

proxy/v1/config/route_rule.proto
```diff
@@ -153,74 +191,47 @@ message RouteRule {
...
 }
 
-// IstioService identifies a service and optionally service version.
-// The FQDN of the service is composed from the name, namespace, and implementation-specific domain suffix
-// (e.g. on Kubernetes, "reviews" + "default" + "svc.cluster.local" -> "reviews.default.svc.cluster.local").
-message IstioService {
-  // The short name of the service such as "foo".
-  string name = 1;
+message TcpRoute {
+  // RECOMMENDED. Precedence is used to disambiguate the order of
```
ZackButcher: Lets drop "Recommended" here and elsewhere; I don't think it adds much.

proxy/v1/config/route_rule.proto
louiscryan: I discussed this pattern with Kuat a bit and it seemed like it had usability issues.

I think we will have a problem when there is a high degree of protocol diversity that needs to be supported. Most of the pain will be in the matching rules BTW so they need to be coupled to the protocol options I think.

From a usability perspective I think its useful to have specific schema for the highly used protocols (HTTP, TCP, UDP, Websocket) and then provide a generic catch-all that we think user-defined codecs could map into conceptually. Something like HTTP

Thoughts?

ZackButcher: SG. Tying matches to protocols makes sense. I think making the common usecases first class and providing a catch-all is a good compromise given we haven't (yet) been able to come up with abstractions that nicely unify things.

We should give some thought to what the catch-all looks like, but it's clearly not a blocker for the current PR and design (it's not like we're losing some catch-all that exists today with the new design).

proxy/v1/config/route_rule.proto
```diff
@@ -79,50 +79,88 @@ package istio.proxy.v1.config;
...
+
+  // Alternate domain name matches for the addressable service. Supports
+  // standard DNS wildcard prefix match semantics.
+  repeated string domains = 2;
+
+  // Selects the port on the destination that is being addressed. Many services
+  // only expose a single port or label ports with the protocols they support,
+  // in these cases it is not required to explicitly select the port. Note that
+  // selection priority is to first match by name and then match by number.
+  string port = 3;
```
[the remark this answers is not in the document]
louiscryan: Not sure that adds any value here. One is for declaring a port, the other for referencing.

proxy/v1/config/route_rule.proto
```diff
@@ -79,50 +79,88 @@ package istio.proxy.v1.config;
...
+
+  // Optional one or more labels that uniquely identify the service version.
+  //
+  // *Note:* When used for a RouteRule destination, labels MUST be empty.
+  //
+  map<string, string> labels = 5;
```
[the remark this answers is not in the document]
louiscryan: I've taken this out for the moment as I think we still have a pending question about how we want to apply filter constraints to destinations.

Mostly do we want to have named constraints or un-named.

I've created a Source resource where we needed a label filter mechanism anyway and that was less ambiguous

proxy/v1/config/egress_rule.proto
kyessenov: Why is protocol repeated? That would allow strange combinations like MONGO + GRPC

kyessenov: +1 for having one canonical port descriptor proto. Maybe make port uint32 and add `name` as an option for pure named ports? Basically copy this https://github.com/istio/old_pilot_repo/blob/master/model/service.go#L68 into a proto format.

proxy/v1/config/gateway.proto
```diff
@@ -0,0 +1,247 @@
...
+import "proxy/v1/config/route_rule.proto";
+
+package istio.proxy.v1.config;
+
+// Gateway describes a load balancer operating within the mesh that bridges
+// traffic across network boundaries. Supported use-cases include
+// ingress, egress and middle proxies within the mesh.
+
+// The specification describes a set of ports that are exposed, the protocols
+// allowed on each port and TLs configuration etc.
```
kyessenov: typo: TLS

proxy/v1/config/gateway.proto
```diff
@@ -0,0 +1,247 @@
...
+//
+// For example, the following Gateway spec sets up Envoy to act as a
+// load balancer exposing port 80 (http), 443 (https), and port 2379 (TCP).
+// While Istio will configure Envoy to listen on these ports, it is the
+// responsibility mesh operator to ensure that external traffic can reach these
+// ports.
```
kyessenov: This feel insufficient. There needs to be a way to associate a gateway with an external IP. Is a gateway = one external IP, or gateways share an IP?

proxy/v1/config/gateway.proto
```diff
@@ -0,0 +1,247 @@
...
+//           number: 2379  #to expose internal service via external port 2379
+//           protocol: redis
+//         tls:
+//           mode: mtls
+//         hosts:
+//           - redis.bookinfo.com
+//       certs:
+//       - hosts:
+//           - *.bookinfo.com
+//         serverCert: server.crt
```
kyessenov: Are these file references? Are these files supposed to be managed by Istio?

proxy/v1/config/gateway.proto
kyessenov: Where is the translation from 80 to port "http" on reviews.prod?

proxy/v1/config/gateway.proto
```diff
@@ -0,0 +1,247 @@
...
+
+    // TLS modes enforced by Envoy
+    enum TLSmode {
+      // If set to "passthrough", Envoy will forward the connection to the
+      // upstream server as is.
+      PASSTHROUGH = 0;
+
+      // If set to "simple", Envoy will secure connections with
+      // standard TLS semantics (server certs only).
+      SIMPLE = 1;
```
kyessenov: Shouldn't SIMPLE be default as it is the most common way to expose a service mesh?

proxy/v1/config/gateway.proto
```diff
@@ -0,0 +1,247 @@
...
+message TLSConfig {
+  // REQUIRED: A set of domains associated with the server
+  // certificate. Standard DNS wildcard prefix syntax is permitted.
+  repeated string domains = 1;
+
+  // REQUIRED: The name of the file holding the server-side TLS certificate
+  // to use.  It is the responsibility of the underlying platform to mount
+  // the certificate as a file under /etc/istio/ingress-certs with the same
+  // name as the specified in this field.
+  string server_certificate = 2;
```
kyessenov: Why not use SDS for this? We could reuse proto from envoy for DataSource to allow options.

proxy/v1/config/route_rule.proto
```diff
@@ -79,50 +79,88 @@ package istio.proxy.v1.config;
...
+  // of the service and therefore the Destination's labels field MUST be empty.
   //
-  IstioService destination = 1;
+  Destination destination = 1;
+
+  repeated HttpRoute http = 2;
+
+  repeated TcpRoute tcp = 3;
+
+  repeated WebsocketRoute websocket = 4;
```
kyessenov: websocket is weird as it starts as HTTP.

proxy/v1/config/route_rule.proto
```diff
@@ -79,50 +79,88 @@ package istio.proxy.v1.config;
...
+
+  repeated UdpRoute udp = 5;
+}
 
+// Destination models a network addressable service that can receive traffic.
+message Destination {
+  // The addressable service name. May be partially or fully qualified.
+  // Partially qualified names are resolved in the context of the client.
```
kyessenov: I think we want canonical name for destination. That is no client context required for resolution of the name. Otherwise, we'd need multiple duplicate rules for alias names as being used from the client (`reviews` and `reviews.default` would require distinct rules but are equivalent for clients in the `default` namespace).

proxy/v1/config/route_rule.proto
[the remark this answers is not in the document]
kyessenov: This is not good. In JSON numbers don't require quotes, and users will be writing port _name_ "80" instead of port _value_ "80". In fact, you have port called "80" that maps to 9090.

proxy/v1/config/route_rule.proto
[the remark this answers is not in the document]
kyessenov: Can you please indicate labels on _what_? Service or workload?

proxy/v1/config/route_rule.proto
```diff
@@ -153,74 +191,47 @@ message RouteRule {
...
-// request header must be specified.
-message MatchCondition {
-  // Identifies the service initiating a connection or a request.
-  IstioService source = 1;
+  repeated DestinationWeight route = 3;
+}
 
-  // (-- Set of layer 4 match conditions based on the IP ranges --)
-  L4MatchAttributes tcp = 2 ;
+message UdpRoute {
```
kyessenov: Let's not put this into API until Envoy supports UDP.

proxy/v1/config/route_rule.proto
```diff
@@ -494,14 +505,31 @@ message CorsPolicy {
...
-// Port describes the properties of a specific port of a service.
-message Port {
-  // A valid non-negative integer port number.
-  int32 number = 1;
-  // The protocol to communicate with the external services.
-  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|REDIS|TCP.
-  string protocol = 2;
+// Fault injection can be used to test the failure recovery capabilities
+// of downstream services.
+message FaultRule {
```
kyessenov: Not referenced anywhere?

proxy/v1/config/gateway.proto
ZackButcher: Ah sorry, we use the `Port` message, so it's actually:

```yaml
metadata:
  name: bookinfo-rule
spec:
  destination:
    name: reviews.prod
    port:
      number: 9090
      protocol: http
    domains:
      - uk.bookinfo.com
      - eu.bookinfo.com
  match:
  # ...
```

proxy/v1/config/route_rule.proto
louiscryan: I think you're probably right that Websocket is not sufficiently distinguished from HTTP to deserve having a distinct top-level resource.

We're unlikely to route the websocket payloads themselves without having moved over to the protocol being tunneled over websocket. I.e it's a protocol within a protocol. Which would drop into a secondary protocol selection and routing rule mechanism.

Will update.

proxy/v1/config/gateway.proto
rshriram: do we care?

proxy/v1/config/gateway.proto
rshriram: Not managed by Istio. User is responsible for these. We should just specify the full path.

proxy/v1/config/gateway.proto
kyessenov: Yeah, if we want feature parity with k8s ingress. At least some clarification how this works end-to-end.

proxy/v1/config/gateway.proto
kyessenov: In reality, we fetch these certs. How's user responsible to deliver files on a managed ingress controller?

proxy/v1/config/gateway.proto
rshriram: the thing is this is user visible config. The envoy one is too complex. I think we should have a one of that drops into the low level envoy stuff,

proxy/v1/config/gateway.proto
ZackButcher: I think we should be transparent if no config is provided.

andraxylia:
@rshriram Can you please add a link to the doc in this PR?

rshriram:
Closing this as all action has moved to #254 ..

Issue #249, opened 2017-11-21
istio/api needs a build


andraxylia:
Anything can be committed in a proto file and there is no validation this builds until istio/istio is built.

istio/api needs a bazel build because it needs to compile the proto files.

On top of this, the build can run scripts/generate-protos.sh to update the generated files.

PR #250, opened 2017-11-21
add a pre-commit script that generates pb.go files from our protos

ZackButcher:
Fixes https://github.com/istio/api/issues/249


andraxylia:
@ZackButcher thanks for adding this!

PR #252, opened 2017-11-21
add EndUserAuthenticationPolicySpec to mixerclient ServiceConfig

ayj:


mixer/v1/config/client/client_config.proto
```diff
@@ -43,6 +44,9 @@ message ServiceConfig {
 
   // Quota specifications to generate quota requirements.
   repeated QuotaSpec quota_spec = 5;
+
+  // End user authentication policy.
+  EndUserAuthenticationPolicySpec end_user_authn_spec = 6;
```
qiwzhang: MixerControl could not use it with current design (Auth library has Envoy code dependency, and it will be build into its own Envoy filter). 
We should not add it to ServiceConfig.

ayj: My understanding is that Envoy filter config and filter instance are decoupled. Mixerclient can instantiate multiple separate filters driven by a single block of configuration from Pilot. In NamedHttpFilterConfigFactory mixerclient would invoke addStreamDecoderFilter() for each dependent filter in the prescribed order. cc @lizan

lizan: Yes that is correct, as long as we have `EndUserAuthenticationPolicySpec` in during filter factory, we can instantiate what we need.

qiwzhang: As discussed,  at Envoy filter factory time,  mixer client will concat all auth_spec from all service_configs and pass them to auth_filter.    It will not be per-route, but for it should be fine for short term.

PR #257, opened 2017-11-22
Rename enable_mixer_check to disable_check_calls.

qiwzhang:
Proto boolean default value is fault.  Change the default to be enabled.
Also fixed a typo for default api_key

### DIFF
--- a/mixer/v1/config/client/client_config.proto
+++ b/mixer/v1/config/client/client_config.proto
@@ -28,10 +28,10 @@ option (gogoproto.gostring_all) = false;
 
 // Defines the per-service client configuration.
 message ServiceConfig {
-  // If true, not to call Mixer Check.
+  // If true, do not call Mixer Check.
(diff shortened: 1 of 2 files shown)


mixer/v1/config/client/client_config.proto
```diff
@@ -28,11 +28,11 @@ option (gogoproto.gostring_all) = false;
 
 // Defines the per-service client configuration.
 message ServiceConfig {
-  // If true, call Mixer Check.
-  bool enable_mixer_check = 1;
+  // If true, not to call Mixer Check.
```
ayj: "If true, do not call Mixer Check"

qiwzhang: Done. Thanks

ayj:
You need to run `scripts/generate-protos.sh` and commit the updated *pb.go files

PR #258, opened 2017-11-22
update generated files

ayj:


kyessenov:
how did `prow/api-presubmit` miss this?

ayj:
The Github settings weren't setup to gate on failure.

ayj:
Looks like the cached bazel-genfiles is causing presubmit to fail (see below). istio/istio will have the same problem. Do we need to update the wildcard match of *.pb.go to be more explicit? 

```
I1122 19:32:44.345] On branch test
I1122 19:32:44.345] Untracked files:
I1122 19:32:44.345]   (use "git add <file>..." to include in what will be committed)
I1122 19:32:44.345] 
I1122 19:32:44.345] 	proxy/v1/config/gateway.pb.go
I1122 19:32:44.345] 
I1122 19:32:44.345] nothing added to commit but untracked files present (use "git add" to track)
I1122 19:32:44.345] Repo has unstaged changes. Re-run ./scripts/generate-protos.sh
```

kyessenov:
you could do `bazel query attr(...)` but the simple thing is to nuke the cache.

sebastienvas:
For this repo we could disable the cache. Do you want me to do so ?

PR #259, opened 2017-11-27
move existing pilot protos to routing/v1alpha1

rshriram:
In lieu of changes to come to v1alpha2

	renamed:    proxy/v1/config/proxy_mesh.proto -> mesh/v1alpha1/config.proto
	renamed:    proxy/v1/config/ -> routing/v1alpha1/config/


ZackButcher:
How about this instead: `proxy/v1/config/proxy_mesh.proto` -> `mesh/v1/config.proto`? At a minimum keep it versioned (`global/v1/proxy_mesh.proto`), IMO `proxy_mesh.proto` isn't a great name, and finally `global` doesn't really fit the existing directory scheme. I think `mesh` fits better with `mixer`, `pilot`, and `broker` than `global`.

ayj:
+1 for versioning proxy_mesh.proto.

Also, all of pilot/proxy config is currently versioned under the same group `config`. It's not clear to me how (if?) we would extend this in the future to support additional/modified APIs. Maybe something like `proxy/config/core/v1/{...}.proto` for now and then additional groups could go under `proxy/config` in the future?

rshriram:
I have no objections to either.. @kyessenov ^^^ ?

kyessenov:
Yeah, call it `mesh` and version it. It's a global toggle switch, not just routing.

rshriram:
@ayj I created routing/v1alpha1/... We could add other sections later (e.g., apim/v1alpha1/...). Unless you want to push all sidecar related stuff under one root.

ZackButcher:
LGTM, I like routing over proxy.

routing/v1alpha1/config/dest_policy.proto
```diff
@@ -15,9 +15,9 @@ syntax = "proto3";
 
 import "google/protobuf/any.proto";
 import "google/protobuf/duration.proto";
-import "proxy/v1/config/route_rule.proto";
+import "routing/v1alpha1/config/route_rule.proto";
 
-package istio.proxy.v1.config;
+package istio.routing.v1alpha1.config;
```
ayj: The location of `config` segment looks wrong to me here. The repo is inherently about api/config already so it seems like it either shouldn't be needed at all or at least at the top-level to keep config seperate from build and scripts directories, e.g.

```
config/routing/v1alpha1/...
config/mixer/v1alpha1/...
scripts/...
prow/
```

ZackButcher:
I do not like a top level `config` dir - I'd be a bigger fan of `mixer/config/v1alpha1/...`, `routing/config/v1alpha1/...`. This scheme makes sense for Mixer because both the API definition and the configuration protos live in istio/api. It's a bit out of place for the Envoy related config since the API definitions live elsewhere and we'll only ever have config protos in this repo. Given that, IMO just drop the `config` entirely: `istio.routing.v1alpha1.RouteRule`, `istio.routing.v1alpha1.HttpFault`, etc.

routing/v1alpha1/config/dest_policy.proto
ZackButcher: I do not like a top level `config` dir - I'd be a bigger fan of `mixer/config/v1alpha1/...`, `routing/config/v1alpha1/...`. This scheme makes sense for Mixer because both the API definition and the configuration protos live in istio/api. It's a bit out of place for the Envoy related config since the API definitions live elsewhere and we'll only ever have config protos in this repo. Given that, IMO just drop the `config` entirely: `istio.routing.v1alpha1.RouteRule`, `istio.routing.v1alpha1.HttpFault`, etc.

ayj: The fact that API definitions don't live in the API repo is non-obvious, but that is outside the scope of this PR. Dropping `config` from the package name seems like a good first step and could be done in this PR. We can tackle the other organizational issues re: mixer separately.

ayj:
lgtm

PR #263, opened 2017-12-05
Added max_retries to circuit breaker config

pacevedom:
This PR is to add the max_retries option to the default priority circuit breaker configuration API.

It just adds the parameter to the .proto file and nothing else, under the name of "httpMaxRetries", and it maps to [Envoy circuit breaker config](https://www.envoyproxy.io/docs/envoy/latest/api-v1/cluster_manager/cluster_circuit_breakers.html?highlight=max_retries)


routing/v1alpha1/dest_policy.proto
```diff
@@ -201,6 +201,10 @@ message CircuitBreaker {
     // service that can be ejected by the circuit breaker. Defaults to
     // 10%.
     int32 http_max_ejection_percent = 8;
+
+    // Maximum number of retries that can be outstanding to all hosts in a
+    // cluster at a given time. Defaults to 3.
+    int32 http_max_retries = 9;
```
rshriram: Thank you for adding these. Can you explain your use case a bit more in detail? I had omitted this thinking that this might not be of much use.

pacevedom: Hi!

We are testing out Istio on our current deployment, which used to have just envoy, without all the added functionality that istio provides. While testing this, we are finding some bits of the envoy configuration that we miss, this is one of them. We have a high number of connections to upstream clusters, and we make quite a heavy use of circuit breaking, so we want to be able to control how many concurrent retries are being pushed to each cluster, not just the default number, which is 3.

BTW I just realized I pushed this to github using the wrong user.email, so I will post a new PR with the good one.

pacevedom:
Replacing this PR for [#268](https://github.com/istio/api/pull/268) since the commit had faulty metadata.

PR #264, opened 2017-12-05
Add standard message types that will be used inside templates

guptasu:
Add standard message types that will be used inside templates to represent fields that are

IP_ADDRESS, URI, DURATION, TIMESTAMP, EMAIl_ADDRESS, DNS_NAME

**NOTE:** Currently we have duration and timestamp already covered with google.protobuf.* types, but for consistency we should replace them too with types inside standard_type.proto


mixer/v1/config/descriptor/standard_type.proto
```diff
@@ -0,0 +1,37 @@
...
+message DNSName {}
+
+// EmailAddress is used inside templates for fields that are of ValueType "EMAIL_ADDRESS"
+message EmailAddress {}
+
+// Uri is used inside templates for fields that are of ValueType "URI"
+message Uri {}
+
+// Duration is used inside templates for fields that are of ValueType "DURATION"
+message Duration {}
```
geeknoid: Don't we already supported this by using the normal Timestamp and Duration types?

mixer/v1/config/descriptor/standard_type.proto
```diff
@@ -0,0 +1,37 @@
+// Copyright 2017 Istio Authors
```
geeknoid: Maybe standard_types.proto would be better (plural form)?

mixer/v1/config/descriptor/standard_type.proto
guptasu: Yes, but I think there is no benefit to keep it that way.. everything in template can be expressed in a self contained type system as oppose to a mix of standard_types and google.protobug.*.. From standard_types we generate nice types in `Go` and `proto`; for example for standard_types#Duration we will generate google.protobuf.Duration and time.Duration inside the generated files. 

I mainly added those two types mainly to achieve consistency.

geeknoid: OK, I buy that.

guptasu:
@geeknoid alright, the types are now moved to /template dir..

the usage would look something like:
`istio.mixer.v1.template.IPAddress source_ip = 1;`

PR #254, opened 2017-11-22
Istio Mesh LB/Gateway + RouteRule v1alpha2

rshriram:
This proposal suggests a new Gateway configuration resource that describes a set of ports that should be exposed bridging two networks. A Gateway can act as an ingress allowing external traffic into the mesh and it can act as an egress allow traffic in the mesh to exit. A Gateway can also be used to model a proxy that is entirely internal to the mesh and implemented either by sidecars or by a middle proxy. Behaviors for the ports like supported protocols, TLS, SNI etc are described by the resource. 

Users can then use standard Istio RouteRules to control HTTP requests & TCP traffic entering a Gateway by binding the gateway to a route rule. The route rule will be applied at the gateway if and only if the rule’s destination (or one of its domain aliases) corresponds to one of the wildcard domains supported by the Gateway. 

Note that this is not the same as the recent proposals for multiple zones that advocate for a new access model (direct addressing of services across platforms). These proposals target users who fully buy into Istio and are willing to modify their applications, while the proposal here targets the more immediate need - deployments that already exist.

Detailed discussion can be found in https://goo.gl/96UTKQ

PR #267, opened 2017-12-08
Added ring hash configuration for load balancing

pacevedom:
Add ring hash configuration to LoadBalancing DestinationPolicy.

Ring hashing can be used to provide stickiness and/or predictability on the hosts serving a specific request in a cluster. Configuration includes the same parameters as Envoy, the ring size and also the hashing algorithm.


routing/v1alpha1/dest_policy.proto
```diff
@@ -114,11 +114,23 @@ message LoadBalancing {
...
+    uint32 ring_size = 1;
+
+    // Use Envoy standard hash in the ring. If false, Envoy will use xxHash.
+    bool std_hash = 2;
+  }
   oneof lb_policy {
     // Load balancing policy name (as defined in SimpleLBPolicy below)
     SimpleLBPolicy name = 1;
+    // Load balancing policy for ring hash (as defined in RingHash below)
+    RingHash ring_hash = 2;
```
rshriram: There was a reason I didn't add this.. if its pure ketama hash with load distribution, its fine.. However, envoy supports a much richer model (https://github.com/envoyproxy/data-plane-api/blob/master/api/rds.proto#L466), that allows you to use source IP, cookie, or HTTP header. I guess looking at this, its fine to have this feature standalone. But we would need improvements to the route as well, in order to take advantage of the other properties

rshriram: For what you are looking at, https://www.envoyproxy.io/docs/envoy/latest/api-v1/route_config/route.html#hash-policy, is also probably useful..

routing/v1alpha1/dest_policy.proto
```diff
@@ -114,11 +114,23 @@ message LoadBalancing {
...
   }
+  message RingHash {
+    // Minimum ring size (e.g. virtual nodes). A larger size will provide better
+    // request distribution since each host in the cluster will have more virtual
+    // nodes. If the number of hosts is greater than the ring size, each host
+    // will be allocated a single virtual node.
+    uint32 ring_size = 1;
+
+    // Use Envoy standard hash in the ring. If false, Envoy will use xxHash.
+    bool std_hash = 2;
```
rshriram: This will default to false. is that what you want?

routing/v1alpha1/dest_policy.proto
```diff
@@ -114,11 +114,23 @@ message LoadBalancing {
     // checking policy is configured.
     RANDOM = 2;
   }
+  message RingHash {
+    // Minimum ring size (e.g. virtual nodes). A larger size will provide better
```
rshriram: Please add REQUIRED if this is a mandatory param

routing/v1alpha1/dest_policy.proto
pacevedom: I was planning to do them separately, one PR for the ring hash, and another for the HashPolicy specified in the link you provided. One of them would map to pure destination policy (or upstream cluster), and the other one is per-route configuration. Would you prefer if I add it to this PR? I am halfway done with that part as well, so its no big hassle.

routing/v1alpha1/dest_policy.proto
pacevedom: Yes, but I am having some thoughts on this. Docs say that default is true, but it only exists because of migration purposes/different platforms, and xxHash should be used instead. Not sure on which value should be the default one. Maybe true would apply better, as it is what Envoy does.

https://www.envoyproxy.io/docs/envoy/latest/api-v1/cluster_manager/cluster_ring_hash_lb_config.html#config-cluster-manager-cluster-ring-hash-lb-config

routing/v1alpha1/dest_policy.proto
pacevedom: That one slipped, its changed now.

rshriram:
can you keep an eye on #254 to make sure these changes get applied to v1alpha2 as well? I have added the max_retries, but will wait for you to fix this up, so that I can add the ring hash as well.

routing/v1alpha1/dest_policy.proto
rshriram: Let it remain as is.. We can use xxhash

routing/v1alpha1/dest_policy.proto
rshriram: Ah then fine..

routing/v1alpha1/dest_policy.proto
rshriram: still missing. Start the sentence with REQUIRED.

routing/v1alpha1/dest_policy.proto
pacevedom: So, you prefer separate commits? I can do them both in one, if that suits you better (I am now thinking that I would ease things a little bit if I do them combined, you wont have to go through more PRs/changes).

routing/v1alpha1/dest_policy.proto
pacevedom: Fine then. My concern was having a different default value than Envoy, that might cause some confusion.

routing/v1alpha1/dest_policy.proto
pacevedom: Sorry, I meant done but still not submitted to my branch, it will be there in my next commit.

pacevedom:
Sure, I will keep an eye on #254. I am going to wait until you see if its better to do hash ring + hash policy combined, or separate PRs though, I dont mind doing them in the way you guys prefer.

routing/v1alpha1/dest_policy.proto
rshriram: One commit per functionality in istio/api.

rshriram: So from that point of view, you should do them in this commit, but its going to be painful, because the route rules would look weird having a LB field. If you wait for #254 , it should make it easier to add the header/cookie based hash, since route rules and destination policies are folded into one rule. WDYT?

pacevedom: I will wait to push the cookie/header/source ip hash policies until #254 is ready, looks like the best option for everyone, less conflicts, less redoing stuff, and less changes to look at. I will do the ring hash for now and leave the rest ready to submit. Thanks!

rshriram:
Can you apply this to destination_rule.proto in v1alpha2 as well ?

rshriram:
closing in favor of #282

PR #272, opened 2017-12-12
filter extensions for Envoy

rshriram:
Allow extending the proxy configuration with proxy-specific extensions through route rules.


rshriram:
cc @louiscryan

routing/v1alpha2/route_rule.proto
```diff
@@ -725,3 +734,84 @@ message PortSelector {
...
+    // Filter specific configuration which depends on the filter being
+    // instantiated. See the supported filters for further documentation.
+    google.protobuf.Struct config = 2;
+  };
+  
+  // Envoy HTTP filters to be added to every HTTP connection manager
+  // instance generated by Istio.
+  repeated HTTPFilter http_filters = 1;
+
+  // Envoy network filters to be added to every non-TCP listener instance
```
costinm: You mean non-HTTP ?

routing/v1alpha2/route_rule.proto
```diff
@@ -725,3 +734,84 @@ message PortSelector {
...
+//       hosts:
+//       - reviews
+//       rules:
+//       - http:
+//         - route:
+//           - destination:
+//               name: reviews
+//               labels:
+//                 version: v1
+//       extensions:
```
costinm: This is a bit confusing in this form, are you sure RouteRule is the right place ? How would the rules be merged,  do you set the filters on any of the rules to that destination ? For example you may have a rule for /foo and one for /bar - maybe with rules with different label selectors. Which one applies - and is it set on envoy server ( listener) side or on client ?

Maybe a couple of more explicit objects would be cleaner ? To avoid translations - it could simply be a 'bytes', defined as 'native proxy server config for the server listener, routing rule, cluster, etc'. 

Maybe even a config map ( untyped ), with annotation indicating the server type (envoy / etc ) and where it applies.
I know people are allergic to configmaps, but RouteRule is already complex enough ?

routing/v1alpha2/route_rule.proto
```diff
@@ -725,3 +734,84 @@ message PortSelector {
...
+//       - "@type": type.googleapis.com/istio.routing.v1alpha2.EnvoyFilters
+//         httpFilters:
+//         - name: envoy.lua
+//           config:
+//             inlineCode: "... lua code ... "
```
nmnellis: do you suggest any tools to help convert lua scripts to yaml/json compatible strings?

routing/v1alpha2/route_rule.proto
rshriram: The rule format has changed. We have only one rule for a service. The rule is an array of sub rules for both HTTP and TCP. Its essentially some equivalent to a listener if you will. The choice of struct and any was based on @htuch's article (https://medium.com/@htuch/dynamic-extensibility-and-protocol-buffers-dcd0bf0b8801). I don't know much about how the bytes thingy works. Will defer to the proto people (@louiscryan / @htuch ) to comment on this. 

With regard to config maps/etc., I want to avoid any tie in to kube concepts. So, config maps, annotations, etc. are all non portable. While configmaps can be thought of as files, a single global config file containing envoy filters might work, but its too coarse grained. We might need to inject lua filters/other filters on a per service basis. Since there is only one route rule for a service, sticking the lua filter into the route rule helps.

@nmnellis from the Weather Company, indicated that they wanted to be able to set filters per virtual host or listener (to call out to their internal auth systems). So this solution came about as a result of that discussion/use case.

routing/v1alpha2/route_rule.proto
rshriram: good catch.. fixed

routing/v1alpha2/route_rule.proto
ayj: Any thoughts on how the extension filters would be merged with existing Istio filters (e.g. mixerclient)?

routing/v1alpha2/route_rule.proto
```diff
@@ -725,3 +734,84 @@ message PortSelector {
...
   }
 }
+
+// Envoy-specific filters that can be used to enable customized processing
+// in addition to the features exposed by the routing rule. This feature
+// must be used with care, as incorrect configurations could potentially
+// destabilize the entire mesh. Common use cases for using the extensions
+// include use of Envoy's Lua filters, in addition to enabling any
+// proprietary filters developed inhouse (e.g, custom authorization filter
+// for HTTP).
```
frankbu: Aren't mixer adapters supposed to be the way that Istio users can plug in custom authorization? What use case(s) require this Envoy specific approach?

kyessenov:
We continue to have the tension here between what's Istio and what's Envoy.
I think with this change, we pretty much commit to Envoy proxy. The
use-cases are fairly standard - custom extensions to Envoy for things not
covered by Istio.

routing/v1alpha2/route_rule.proto
```diff
@@ -725,3 +734,84 @@ message PortSelector {
...
+//               labels:
+//                 version: v1
+//       extensions:
+//       - "@type": type.googleapis.com/istio.routing.v1alpha2.EnvoyFilters
+//         httpFilters:
+//         - name: envoy.lua
+//           config:
+//             inlineCode: "... lua code ... "
+//
+message EnvoyFilters {
```
frankbu: Should this go in a separate yaml file since it's Envoy specific and the core routing rule spec is supposed to be a general and independent of proxy implementation?

routing/v1alpha2/route_rule.proto
mandarjog: I think this makes it much better aligned with other proxies.
You always use directives inside vhost / route matches.
@ayj mixerclient config already has an awkward configuration where the filter config gets `map<string, ServiceConfig> service_configs`.  With this change those service configs can go into vhosts.
Similarly other API management items should also be folded in here.

routing/v1alpha2/route_rule.proto
```diff
@@ -131,6 +133,13 @@ message RouteRule {
 
   // A list of routes for TCP traffic.
   repeated TCPRoute tcp = 4;
+
+  // Extensions contain sidecar implementation specific configuration, that
+  // cannot be easily expressed through the routing rule feature
+  // set. Custom sidecar implementations could have enhancements such as
+  // caching, dynamic scripting, etc. The configuration for such features
+  // could be provided through the extensions mechanism.
+  repeated google.protobuf.Any extensions = 5;
```
ZackButcher: I think we'll wind up wanting a PATCH field mask here, so we can determine where in the document to place the config blob.

costinm:
'configmap' is equivalent with files, and would work very well with rawvm /
outside k8s.
Agree, single configmap is not flexible - I was thinking of one per
listener / cluster / etc.

Still think overloading RouteRule would result in confusion - in particular
if nginx or others implementing
the protocol start adding their own extensions this way. And I still don't
understand if the filter will apply
at source or destination, or how to customize the clusters or route rules.
I think the envoy config is
clearly documented and whatever we do should be as close as possible, with
minimal restrictions or
translations - so whatever fragment you would put in envoy.conf should just
be possible to add to istio.

Note there is a separate discussion on template-based injection - not sure
if relevant.

routing/v1alpha2/route_rule.proto
ZackButcher: So something like:

```proto
message Extension {
  FieldMask mask = 1;
  google.protobuf.Any body = 2;
}
repeated Extension extensions = 5;
```

kyessenov:
By the way, @rshriram, please demote fault and mixer filter to an
extension. We've got long standing issues with treating it natively (no
faults at ingress) and the choice of ordering between fault and mixer is
unclear. These decisions should ultimately be up to the use-case, so it
must surface somewhere in the ordering of extension filters.

routing/v1alpha2/route_rule.proto
nmnellis: we actually have a use case for exactly this. we do not want to use istio for authorization and instead have our own auth scheme currently. We want to intercept requests and auth them against our own service. its actually one of the core use cases in envoy. https://github.com/envoyproxy/envoy/blob/ad8e6a0eafbd6b71b79ae53c2cc5ede96b6c45d8/docs/configuration/http_filters/lua_filter.rst#currently-supported-high-level-features

routing/v1alpha2/route_rule.proto
ayj: Completely externalizing mixerclient config has merit. 

Including filter extensions in RouteRule combines traffic routing and policy management into the same unit of configuration for a service. The fact this single unit is called `RouteRule` is a bit confusing. For example, users need to create a `RouteRule` (or carefully modify an existing one with HTTP/TCP rules) to configure API management features. RouteRule becomes much more of a component-level config (i.e. ProxyConfig) than the previous intent-based routing config. 

FWIW, we've tried to move away from monolithic per-service config in the past in the context of service management. 

And as @costinm said, it isn't clear whether these filters are applied to client vs. server proxies in the mesh. For API management we'd need to differentiate.

routing/v1alpha2/route_rule.proto
frankbu: Wouldn't it be possible to plug in your auth policy as a custom mixer adapter instead? That way you could do it in a way that would work with any proxy implementation, instead of having an envoy-only solution?

nmnellis: from my understanding(which might be wrong) istio follows proper authentication protocols and standards, so if my apps don't follow these standards(and the teams that own them are unwilling to change) they wouldnt work with mixer

routing/v1alpha2/route_rule.proto
rshriram: Keep in mind that filters can do a lot of other things, such as custom transcoding, custom protocol decoding, and a whole lot of other things. Lua filter is just one example, and using Lua to do authz is a very point in time example. People can use this to manipulate headers, change URLs in conditional ways, etc. 

A routerule is always applied at the client. I am not sure if we should remove fault and mixer filter out of pilot and add it explicitly as a filter here. Pilot could use this approach internally, but thats an implementation detail.

For server side configurations, you probably need a separate resource that allows you to ship filters down to Envoy.

routing/v1alpha2/route_rule.proto
rshriram: @frankbu Lua is great for prototyping. There are a lot of other use cases, such as header manipulation, rewrites, etc. that simply don't fit inside the rigid configuration semantics we offer. I can remove the word "custom authorization" and use something else if it makes things clear.

routing/v1alpha2/route_rule.proto
rshriram: done

routing/v1alpha2/route_rule.proto
rshriram: what does that mean? We don't allow people to customize virtual hosts, listeners, etc. We just let people add custom filters. Thats all we offer with this.

rshriram:
@kyessenov that happens inside pilot implementation IMO. Asking users to configure envoy filters to inject faults seems too high touch. The filter extensions is strictly for advanced use cases only.

kyessenov:
Why is it high touch? We should allow extensibility and the best way to
demonstrate it is to modularize our config generation. Faults and mixer fit
perfectly into envoy extensibility model

routing/v1alpha2/route_rule.proto
```diff
@@ -131,6 +132,13 @@ message RouteRule {
 
   // A list of routes for TCP traffic.
   repeated TCPRoute tcp = 4;
+
+  // Extensions contain sidecar implementation specific configuration, that
+  // cannot be easily expressed through the routing rule feature
+  // set. Custom sidecar implementations could have enhancements such as
```
frankbu: maybe instead of `... expressed through the routing rule feature set.`, it should say something like `... expressed using proxy-independent routing rule or mixer policy configuration.`

kyessenov:
Envoy extension mechanism is also fairly proxy-independent. It's basically
a JSON and a name. The name has `envoy.` as prefix for out-of-the-box
filters.

frankbu:
Hmm, if Envoy extension mechanism is fairly proxy-independent, maybe we should downplay the Envoy-ness. Instead of naming the message `EnvoyFilters` and describing them as envoy extensions, we could just call them `ExtensionFilters`, or maybe even just `ProxyFilters`, and say that all filters may not be supported by all proxy implementations?

rshriram:
EnvoyFilters is not something nginx or traefik would use. Its the structure of the config that matters. If you notice, I used a `any` and then a struct inside the any. That would mean, anybody could come up with their custom FoobarProxyFilters class, provide a type URL to that foobarproxyfilter class, such that foobarproxy implements standard routing and allows users to extend the routing with more functions..

With respect to the high-touch ness of fault rules: people have to identify the upstream cluster that is associated with a set of conditions in order to restrict fault injection to a specific filter. (granted we don't allow it today, but I wanted to). That would mean they have to understand how Pilot is generating the clusters and use the exact same cluster name in the fault filter configuration. Also keep in mnd that faults can be applied to http/tcp/mongo.. And (unfortunately) in mongo, faults are part of core mongo configuration, not a separate filter. 

Pilot could use the same extension mechanism internally to plug in fault and mixer filters. But users don't have to know the low level config settings of mixer filter in Envoy. Only expert users would want to control the ordering, set of attributes being passed down, etc.

kyessenov:
IMHO, it would be better if the filter itself was not relying on any Envoy
terminology, could be generalized to non-envoy proxy with the same config.
That should be mostly true for mixerfilter I think.

routing/v1alpha2/route_rule.proto
ZackButcher: Why limit it to custom filters? We could support arbitrary update operations if we used a field mask.

kyessenov:
I wouldn't bet on PATCH support in k8s any time soon (hopefully in a year).
Perhaps, we should think how to structure API so that multiple rules can
provide extensions for the same route?

routing/v1alpha2/route_rule.proto
```diff
@@ -131,6 +132,13 @@ message RouteRule {
...
+  // caching, dynamic scripting, etc. The configuration for such features
+  // could be provided through the extensions mechanism.
+  repeated google.protobuf.Any extensions = 5;
```
louiscryan: What are we hoping to extend and where? At this level of the API I would expect the extension to operate at the same level as HTTPRoute & TCPRoute, i.e. to act as a routing rule for a protocol.

If we wanted to allow extensions to request-rewriting I would expect that at the same level inside a route-rule as header rewrites

If we wanted to allow extensions to the listener stack I would expect that in Gateway & DestinationPolicy etc.

routing/v1alpha2/route_rule.proto
louiscryan: Do we really need this for PATCH ? So far we've assumed that we either have overwrite or append for repeated field patching.

If Any type names were unique by use-case then we could do a name-merge (i.e treat repeated Any as th equivalent of a map<any-type-name-string, any-content> but I don't think that would fly because of the use of Struct for many use-cases.

routing/v1alpha2/route_rule.proto
rshriram: Per implementation, the route rule generates vhosts and http conn manager, fault filter as well. This extension would allow people to plug in custom filters. Destination policies correspond to clusters. 

You are right that this needs to reside at a higher level. Gateway is one listener. But internal services spec is derived from kube or consul. From that point of view, we could simply pull this out, pin it per ip:port, to allow configuring filter stacks per listener. Given that this is super advanced low level config, we don’t really have to worry about ux issues. 

Now, this might be a place that allows people to customize the generated config through the patch edit thingy that @ZackButcher is talking about. But I am not qualified there to make a decision. Leave that part to you guys. Or this might be an opportunity to simply do the Jsonnet thing that @kyessenov was doing.

routing/v1alpha2/route_rule.proto
ayj: Mixerclient filter is applied at server and client proxy. The proposed client-side only extension mechanism is not sufficient to replace the existing internal Pilot-managed mixerclient filter. 

It's probably worth documenting that these extensions are applied client-side. A service's custom authz filter could be bypassed if mTLS was disabled and the caller bypassed the client-side proxy.

routing/v1alpha2/route_rule.proto
rshriram: I am not sure we have a use case for patching whatever we generate. And even if we do, the fields in our route rule do not map one on one with the stuff in envoy (or people have to decode such mappings). Given the listeners, source based matches, etc. that we infer, patching might be hard.

I was trying to grab the low hanging fruit where most people want some additional scripting using Lua, or orgs are extending Envoy with some extra filters, and to that end, they end up forking/customizing Pilot because there is no way to add additional filters.

kyessenov:
cc @spikecurtis 
https://docs.google.com/document/d/1pGehJbW90r5U9bTlbzW4CTTupZQhaEebKW2TcjPfFk4/edit#heading=h.gryauiau89tc

I prefer the service oriented scoping as opposed to network-name oriented one.

PR #274, opened 2017-12-12
DestinationRules

rshriram:
Introduce concept of named subsets - i.e. names for service versions that route rules can refer to. One can write traffic policies for these named sets to customize Envoy cluster.


routing/v1alpha2/upstream_settings.proto
```diff
@@ -0,0 +1,236 @@
...
+//     metadata:
+//       name: bookinfo-upstream
+//     spec:
+//       destination:
+//         name: myredissrv
+//       lbPolicy: LEAST_CONN
+//
+// UpstreamSettings can be inherited based on scope. By omitting the
+// destination in an upstream setting (equivalent to a wildcard), the
+// configuration can be applied to all clients in the mesh. The following
```
costinm: To all clients in the namespace - we don't want namespace overriding another one.

Setting in istio-system may be used as default.

routing/v1alpha2/upstream_settings.proto
```diff
@@ -0,0 +1,236 @@
...
+// the sidecar, and outlier detection settings to detect and evict
+// unhealthy hosts from the load balancing pool. For example, a simple load
+// balancing policy for the ratings service would look as follows:
+//
+//     apiVersion: config.istio.io/v1alpha2
+//     kind: UpstreamSettings
+//     metadata:
+//       name: bookinfo-upstream
+//     spec:
+//       destination:
```
costinm: Shouldn't source be used ? For example a client making a large number of requests may need larger pool size.
Not sure if timeouts are included here - but they can be very specific to client.

routing/v1alpha2/upstream_settings.proto
```diff
@@ -0,0 +1,236 @@
...
+  message TCPSettings {
+    // Maximum number of HTTP/TCP connections to a destination host.
+    int32 max_connections = 1;
+
+    // TCP connection timeout.
+    google.protobuf.Duration connect_timeout = 2;
+  };
+
+  // Settings applicable to HTTP connections.
+  message HTTPSettings {
```
costinm: How about h2 and gRPC ?

routing/v1alpha2/upstream_settings.proto
rshriram: Yeah, but I am trying to avoid anything to do with namespace here as its not a native concept in other platforms. I don't have a good solution here unfortunately though. @nmittler since you did the file backed CRDs, what are your thoughts? If I put metadata.namespace, how will it play out, if I run this on a non kube environment ?

routing/v1alpha2/upstream_settings.proto
rshriram: Timeouts/retries etc. go into route rule. I think @louiscryan 's point here was that these settings should be set by the service owner and could be overridden by clients. May be source can be pulled in here, but I remember him saying that we shouldn't source match for destination policies. forget the reason

routing/v1alpha2/upstream_settings.proto
costinm: I don't think namespace is a k8s concept - but a security or Istio concept :-)

A user/group should not be able to override settings for a different group - settings for the entire mesh are a very dangerous concept, in particular for multi-cloud and very large meshes (as we hope Istio will support).

routing/v1alpha2/upstream_settings.proto
rshriram: fixed

routing/v1alpha2/route_rule.proto
```diff
@@ -80,18 +80,22 @@ package istio.routing.v1alpha2;
...
+//       - id: reviews-v1
+//         name: reviews
+//         labels:
+//           version: v1
+//         ..more settings like circuit breakers..
+//       - id: reviews-v2
+//         name: reviews
+//         labels:
+//           version: v2
+//         .. more settings like circuit breakers..
```
rshriram: @louiscryan / @kyessenov / @ZackButcher / @frankbu what do you think of this option? This will allow us to collapse destination policy and route rule into a single thing, solve the reference issue, allow us to extend route rules to support a global configuration, and version specific configuration, etc.

ZackButcher: How does the authorship model work with things merged like this? I guess we do expect the service owner to write both the routing and the CB/outlier/LB policy. However, destination policies tend to apply across versions of the service. E.g. it's not common that v2 requires a different retry policy than v1, but this model makes that required, or at least makes it prominent.

rshriram: My bad.. see the update. My thinking is we can have a service level setting and a destination level override

routing/v1alpha2/route_rule.proto
```diff
@@ -131,12 +139,25 @@ message RouteRule {
...
 // Destination indicates the network addressable service to which the
 // request/connection will be sent after processing a routing rule. See
 // RouteRule for usage examples.
 message Destination {
+  // The ID associated with this destination configuration. One or more
+  // routes in a single rule can refer to this configuration by its
+  // ID. These IDs are not visible outside the scope of this routing
+  // rule. Two rules with same destination IDs may refer to entirely
+  // different destinations.
+  string id = 1;
```
louiscryan: I'm not sure this is the right way to factor things. This ties named destination subsets to a single route rule as opposed to making them shared across route rules. 

By implication this means that 'id' is not guaranteed to be a unique name that can be used by itself for routing etc, e.g. cannot be passed in SNI.

A 'subset' is like a service in the service registry like another in many ways. I'd prefer to keep the label filters inline for the moment and solve how to make subsets be shared and the associated naming issues

rshriram:
On Wed, Dec 13, 2017 at 5:43 PM Louis Ryan <notifications@github.com> wrote:

> *@louiscryan* commented on this pull request.
>
> In routing/v1alpha2/route_rule.proto
> <https://github.com/istio/api/pull/274#discussion_r156807738>:
>
> >  }
>
>  // Destination indicates the network addressable service to which the
>  // request/connection will be sent after processing a routing rule. See
>  // RouteRule for usage examples.
>  message Destination {
> +  // The ID associated with this destination configuration. One or more
> +  // routes in a single rule can refer to this configuration by its
> +  // ID. These IDs are not visible outside the scope of this routing
> +  // rule. Two rules with same destination IDs may refer to entirely
> +  // different destinations.
> +  string id = 1;
>
> I'm not sure this is the right way to factor things. This ties named
> destination subsets to a single route rule as opposed to making them shared
> across route rules.
>

Yes but why would we need to share a destination across rules? All rules
for a service live under a single route rule. Destinations for another
service are likely to be different.

> By implication this means that 'id' is not guaranteed to be a unique name
> that can be used by itself for routing etc, e.g. cannot be passed in SNI.
>
>
This is a valid concern. Let me look at options
A simple option is rulename+id

> A 'subset' is like a service in the service registry like another in many
> ways. I'd prefer to keep the label filters inline for the moment and solve
> how to make subsets be shared and the associated naming issues
>
>
Yes but we don’t have to have the name persist in a service registry
(except for SNI). It could live in the route rule. There is reuse across
rules in a single route rule.

louiscryan:
On Wed, Dec 13, 2017 at 3:31 PM, Shriram Rajagopalan <

> On Wed, Dec 13, 2017 at 5:43 PM Louis Ryan <notifications@github.com>
> wrote:
>
> > *@louiscryan* commented on this pull request.
> > ------------------------------
> >
> > In routing/v1alpha2/route_rule.proto
> > <https://github.com/istio/api/pull/274#discussion_r156807738>:
> >
> > > }
> >
> > // Destination indicates the network addressable service to which the
> > // request/connection will be sent after processing a routing rule. See
> > // RouteRule for usage examples.
> > message Destination {
> > + // The ID associated with this destination configuration. One or more
> > + // routes in a single rule can refer to this configuration by its
> > + // ID. These IDs are not visible outside the scope of this routing
> > + // rule. Two rules with same destination IDs may refer to entirely
> > + // different destinations.
> > + string id = 1;
> >
> > I'm not sure this is the right way to factor things. This ties named
> > destination subsets to a single route rule as opposed to making them
> shared
> > across route rules.
> >
>
> Yes but why would we need to share a destination across rules? All rules
> for a service live under a single route rule. Destinations for another
> service are likely to be different.
>

I agree that generally speaking host->destination is 1:1 or 1:N and less
often M:1 or M:N but it can and does occur in more complex virtual hosting
setups

If we add subsets to RouteRule then we will definitely have to allow those
subsets to override the attributes of DestinationPolicy which would lead to
a pretty verbose resource over time.

If we allow subsets to be names that applications can reference (virtual
host, SNI,  header hint etc) then we will need to handle disambiguation
across rules when M:N occurs.

Clearly I think either can work, its more a matter of which one results in
a better compromise. Options are:

1. Subsets in RouteRule only

Favors immediacy of edit (single author) and punts semantic conflicts on
common names in the M:N case.

2. Subset in DestinationPolicy

Favors separation of concerns and disallows conflict resolution. Allows for
destination policy overrides to be done when no RouteRule is needed, common
enough in the 1 host -> 1 service case.

3. Both

Best/Worst of both worlds. Could for instance only allow subsets defined
using #2 to be application visible




>
> > By implication this means that 'id' is not guaranteed to be a unique name
> > that can be used by itself for routing etc, e.g. cannot be passed in SNI.
> >
> >
> This is a valid concern. Let me look at options
> A simple option is rulename+id
>

Not sure that would yield a good outcome. Can have a separate property
which is 'reserved_name' which would require referential integrity checking.


>
> > A 'subset' is like a service in the service registry like another in many
> > ways. I'd prefer to keep the label filters inline for the moment and
> solve
> > how to make subsets be shared and the associated naming issues
> >
> >
> Yes but we don’t have to have the name persist in a service registry
> (except for SNI). It could live in the route rule. There is reuse across
> rules in a single route rule.
>

Agree that it doesn't have to persist.

rshriram:
@louiscryan / @frankbu / @GregHanson / @ijsnellf / @ZackButcher / @kyessenov PTAL.

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
...
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+syntax = "proto3";
+
+import "google/protobuf/duration.proto";
+
+package istio.routing.v1alpha2;
+
+// DestinationRule defines client/caller-side policies for services. These
```
louiscryan: While some of the policies may be applied at the client I suspect we'll have some that apply on the server too in this context at some point. I would keep languages ambiguous. e.g

"DestinationRule defines policies that apply to traffic intended for a service after routing has occurred..."

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
...
+//           version: v3
+//         trafficPolicy:
+//           lbPolicy: ROUND_ROBIN
+//
+// Note that policies specified for subsets will not take effect until
+// a route rule explicitly sends traffic to this subset.
+message DestinationRule {
+  // The service associated with this configuration. The upstream settings
+  // can be applied to all versions of a service by omitting the labels
+  // field in the destination.
```
louiscryan: 'destination' isn't really defined here, not sure you need to say this sentence at all.

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
...
+  LBPolicy lb_policy = 1;
+
+  // Settings controlling the volume of connections to an upstream service
+  ConnectionPoolSettings connection_pool = 2;
+
+  // Settings controlling eviction of unhealthy hosts from the load balancing pool
+  OutlierDetection outlier_detection = 3;
+}
+
+// A subset of instances of a service. Typically indicates pods/VMs
```
louiscryan: A subset of endpoints...

Rather than saying 'Typically indicates' I would give examples of usage. E.g

Subsetting for A/B, versioning etc

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
...
+// implementing a particular binary version of a service.
+message Subset {
+  // REQUIRED. name of the subset. The service name, prefixed with the
+  // subset name can be used for traffic splitting in a route rule.
```
louiscryan: Provide an example of usage

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
...
+  string name = 1;
+
+  // REQUIRED. One or more labels that uniquely identify a specific version of the
```
louiscryan: Avoid 'version' here except when in the context of an example. The labels apply a filter over the endpoints in the service registry.

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
...
+  // service. See route rules for examples of usage.
+  map<string, string> source_labels = 2;
+
+  // Traffic policies that apply to this subset
```
louiscryan: Mention inheritance/override here

louiscryan:
Mostly LG. Some high level comments on doc.

@mccv  - Mark - this might be of interest to you. Would also be good to get your take on the utility of 'named subsets of services' as defined here.

routing/v1alpha2/route_rule.proto
```diff
@@ -508,9 +646,7 @@ message HTTPRedirect {
 //           uri: /v1/bookRatings
 //         route:
 //         - destination:
-//             name: ratings
-//             labels:
-//               version: v1
+//             name: v1.ratings
```
louiscryan: Nice :)

routing/v1alpha2/route_rule.proto
```diff
@@ -86,16 +86,26 @@ package istio.routing.v1alpha2;
 //           uri: "/api/v2"
 //         route:
 //         - destination:
-//             name: reviews
-//             labels:
-//               version: v2
-//           weight: 100
+//             name: v2.reviews
```
louiscryan: Actually I like the weighted routing here as the example rather than making it look like /api/v2 is correlated with v2.reviews, makes the example a little confusing.

routing/v1alpha2/route_rule.proto
```diff
@@ -143,16 +153,136 @@ message RouteRule {
 
 // Destination indicates the network addressable service to which the
 // request/connection will be sent after processing a routing rule. See
-// RouteRule for usage examples.
+// RouteRule for usage examples. The destination.name can be a short name
```
louiscryan: I would say that the name should 'unambiguously refer to a service in the service registry'

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
...
+  // The service associated with this configuration. The upstream settings
+  // can be applied to all versions of a service by omitting the labels
+  // field in the destination.
+  string name = 1;
+
+  // Traffic policies to apply (load balancing policy, connection pool
+  // sizes, outlier detection).
+  TrafficPolicy traffic_policy = 2;
+
+  // One or more named sets that represent individual versions of a service.
```
mccv: Is subsetting more general than a service version? Other use cases we've run into are subsets for on-prem vs. cloud, or EC2 vs. kubernetes.

louiscryan: Yes, its just a named 'subset' of the endpoints of a service composed out of the labels attached to the endpoints. Can be used for anything.

routing/v1alpha2/route_rule.proto
```diff
@@ -143,16 +153,136 @@ message RouteRule {
...
+//
+// 1. If the name is a single word, a short name, a lookup in the service
+// registry is performed to obtain instances of the service. If the lookup
+// fails, the rule will be ignored.
+//
+// The following examples are based on the reviews service running on
+// kubernetes, within prod namespace. They are applicable in other
+// environments as well. If the name consists of multiple words separated
+// by a dot (e.g., v1.reviews.prod, or reviews.prod), one of the following
+// types of inference occurs.
```
frankbu: Isn't overloading the "." character making this more complicated than necessary? Couldn't you use a different char to identify a subset? For example:
- `reviews:v1` instead of `v1.reviews`
- `reviews.prod:v2` instead of `v2.reviews.prod`

wora:
Please do not have DEFAULT load balancing algorithm. If the default is
random, just make RANDOM as 0. Or mark UNSPECIFIED = 0 and treat it as
error.

For config design, it is better to 0 and "" means nothing, and force users
to choose. Any default config is a source of bugs, as nobody can remember
the default behavior consistently, and people always disagree on the right
default.

For example, when someone reads config like "policy: DEFAULT", they has no
idea what it means. "policy: RANDOM" is obvious to almost everyone. So we
should not allow the former case at all.

routing/v1alpha2/route_rule.proto
kyessenov: +1
My feeling we overload "." in hostname with too much information which may lead to unnecessary ambiguity.

louiscryan: @costinm 

I don't necessarily disagree but there are a few things we should consider when making a decision here...

- Do we want to allow the subset names to be usable as virtual host names. If we don't they cannot be used without having a routing rule. Applications using them as virtual hostnames allows them to be used with SNI based routing etc. Allowing this is the logical equivalent of saying 'make this subset a first-class concept in the service registry'

- How do we want subset names to show up in UI/logs/graphs etc. Today those systems expect pretty flat label names and in the case of logging will likely accumulate downstream dependencies on the form that will be hard to change. Then again making the distinction concrete might facilitate analysis somewhat

We can of course give control over these options in the API. E.g give an option that declares as DNS name for the subset so that it can be used with virtual hosting / SNI. Im just not sure that the usability is better, if most subset names are simple then having the names that applications use look like the names that we use has its benefits.

We can also just allow the Destination specification in RouteRule to have two fields 'name' & 'subset' so we're not syntactically overloading while still allowing for interpreting virtual hosts specified by the applications.

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,235 @@
+// Copyright 2017 Istio Authors
```
louiscryan: Yay 2018

routing/v1alpha2/destination_rule.proto
louiscryan: Need to specify the name matching rules here. We support DNS prefix wildcarding with roughly the same semantics has the routing rule spec for host selection.

routing/v1alpha2/route_rule.proto
louiscryan: OK. I think we can defer this discussion by having a field called 'subset' that refers to a named subset within the destination. We can later decide whether we want subsets to have DNS names.

routing/v1alpha2/destination_rule.proto
rshriram: Do we want to be able to control settings for gateways as well ? I presume we do. Things like sticky sessions for example.

routing/v1alpha2/route_rule.proto
rshriram: Just realized. How will this work with IPs? For external services, I might use an IP with CIDR prefix. prefixing subset names does not work here.

routing/v1alpha2/route_rule.proto
frankbu: If the subset is identified using a separate field, I wonder if it would also be reasonable to also allow the subset to be identified by specifying the labels directly, like we used to (i.e., an optional alternative to formally defining and referencing a subset). Seems that for simple use cases where there aren't many rules and no destination policy is needed, that might be preferable.

rshriram:
Update: removed the subset.servicename. There is a separate field called subset under destination. Also removed the DEFAULT load balancing policy per @wora 's comment.. Its back to ROUNDROBIN.

rshriram:
@frankbu PTAL.

routing/v1alpha2/route_rule.proto
rshriram: So this might lead to inconsistencies. If someone wants to define subset specific load balancing policies , they would have to update the old format route rules using the labels in addition to defining subsets in destination rules. The other thing is that giving a name for the subset allows us to provide better logging in the dashboards. Instead of encoding all the labels in the cluster name, we can encode the subset name, adn the grafana dashboards would be cleaner.

frankbu: The more I think about it, I wonder if making subsets DNS addressable is really a requirement? Is there a known use case? It seems to me that subsets are really for the purpose of routing. If one wants a DNS addressable service for a given subset, they can always just define another service for the corresponding pods, can't they?

rshriram: That is true. I think @louiscryan was approaching it from a SNI (TLS) perspective, especially in the context of VMs, where labels may not exist. But in that case, VMs already have DNS names per subset... Either way, we no longer have the DNS requirement for subset. It is something that can be added later, without changing the route rule format, since we now have subset as a first class entity.

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,276 @@
...
+//       name: bookinfo-ratings
+//     spec:
+//       name: ratings
+//       trafficPolicy:
+//         lbPolicy: LEAST_CONN
+//
+// Version specific DestinationRule can be specified by defining a named
+// subset and overriding the settings specified at the service level. The
+// following rule uses a round robin load balancing policy for all traffic
+// going to a subset named testversion that is composed of pods with labels
```
louiscryan: ... composed of endpoints with the label ...

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,276 @@
...
+  // rule.  Could be a DNS name with wildcard prefix or a CIDR
+  // prefix. Depending on the platform, short-names can also be used
+  // instead of a FQDN (i.e. has no dots in the name). In such a scenario,
+  // the FQDN of the host would be derived based on the underlying
+  // platform.
+  //
+  // For example on Kubernetes, when hosts contains a short name, Istio
+  // will interpret the short name based on the namespace of the client
+  // where rules are being applied. Thus, when a client in the "default"
+  // namespace applies a rule containing a name "reviews, Istio will setup
```
louiscryan: "reviews"

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,276 @@
...
+//         labels:
+//           version: v3
+//         trafficPolicy:
+//           lbPolicy: ROUND_ROBIN
+//
+// Note that policies specified for subsets will not take effect until
+// a route rule explicitly sends traffic to this subset.
+message DestinationRule {
+  // REQUIRED. The destination address for traffic captured by this routing
+  // rule.  Could be a DNS name with wildcard prefix or a CIDR
```
louiscryan: Not a routing rule

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,276 @@
...
+  string name = 1;
+
+  // Traffic policies to apply (load balancing policy, connection pool
+  // sizes, outlier detection). Atleast one of trafficPolicy or subsets
+  // must be defined.
+  TrafficPolicy traffic_policy = 2;
+
+  // One or more named sets that represent individual versions of a
+  // service.  Traffic policies can be overridden at subset level. Atleast
+  // one of trafficPolicy or subsets must be defined.
```
louiscryan: Do we need the "at least one of" requirement, don't think we do

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,276 @@
...
+  //
+  // Note that the name field applies to both HTTP and TCP services. In
+  // case of TCP services, the plain service name or an IP address with
+  // CIDR prefix can be used.
+  string name = 1;
+
+  // Traffic policies to apply (load balancing policy, connection pool
+  // sizes, outlier detection). Atleast one of trafficPolicy or subsets
+  // must be defined.
+  TrafficPolicy traffic_policy = 2;
```
louiscryan: Should clarify what the inheritance policy for policies is. 

If there is a DestinationRule for 'name=*' then I think we say that a DestinationRule for 'name=blah' does not inherit from * whereas subsets do inherit

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,276 @@
...
+  LBPolicy lb_policy = 1;
+
+  // Settings controlling the volume of connections to an upstream service
+  ConnectionPoolSettings connection_pool = 2;
+
+  // Settings controlling eviction of unhealthy hosts from the load balancing pool
+  OutlierDetection outlier_detection = 3;
+}
+
+// A subset of endpoints of a service. Subsets can be used for scenarios
```
louiscryan: Primary use case is to facilitate routing rules selecting the subset, secondary use-case is to override policy

routing/v1alpha2/destination_rule.proto
rshriram: Without that, the destination rule would be empty.. It seems pointless to define such an empty rule. Though there is no harm in that..

routing/v1alpha2/destination_rule.proto
rshriram: Oh that, yes, I am explicitly parking that for the meantime. Its a simple matter of updating the comments once we implement it. Put another way, I want to take baby steps to ease the implementation. If we decide to generate documentation based on these APIs (based on monthly releases), I didn't want to end up in a situation where we have documented features unimplemented.

routing/v1alpha2/destination_rule.proto
rshriram: Clarified.. Didn't want to have routing examples in the destination rule page and vice versa.

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,275 @@
...
+  // For example on Kubernetes, when hosts contains a short name, Istio
+  // will interpret the short name based on the namespace of the client
+  // where rules are being applied. Thus, when a client in the "default"
+  // namespace applies a rule containing a name "reviews", Istio will setup
+  // routes to the "reviews.default.svc.cluster.local" service. However, if
+  // a different name such as "reviews.sales" is used, it would be treated
+  // as a FQDN during virtual host matching.  In Consul, a plain service
+  // name would be resolved to the FQDN "reviews.service.consul".
+  //
+  // Note that the name field applies to both HTTP and TCP services. In
```
louiscryan: Need to clarify the precedence for CIDR & DNS here as both can match at the same time. In general I think precedence is determined by which name-form the client uses to address a service with the caveat that we do try to reverse lookup names from IPs in some cases so we are treating names preferentially over IPs to a degree.

routing/v1alpha2/route_rule.proto
```diff
@@ -143,16 +156,123 @@ message RouteRule {
 
 // Destination indicates the network addressable service to which the
 // request/connection will be sent after processing a routing rule. See
-// RouteRule for usage examples.
+// RouteRule for usage examples. The destination.name should unambiguously
```
louiscryan: this is already in RouteRule :)

routing/v1alpha2/route_rule.proto
```diff
@@ -143,16 +156,123 @@ message RouteRule {
...
+//             name: wikipedia.org
+//
 message Destination {
-  // REQUIRED: Service name from the service registry. The name can be a
-  // short name, a fully qualified domain name or an IP address. If short
-  // names are used, the FQDN of the service will be resolved in a platform
-  // specific manner.
+  // REQUIRED. The name can be a short name or a fully qualified domain
+  // name from the service registry, a resolvable DNS name, a service name
+  // from the service registry prefixed with a subset name or an IP
```
louiscryan: we're not doing prefixing any more

routing/v1alpha2/destination_rule.proto
rshriram: what do you mean by CIDR and DNS can match at the same time? For a given field, it can either be an IP or a DNS name right? I am operating under the assumption that DNS name validation mandates presence of a alphabet in the name.

louiscryan: Client talks to 10.11.12.13 which matches 10.11.12/24 and reverse lookup of that IP resolves to an unabiguous service name in the registry. There are different rules for those two things, which one applies ?

rshriram: Good point..

rshriram: Actually, using IP addresses for internal services (those found in the service registry) is disallowed. For a given endpoint, we use the endpoint's service as the lookup key in the routing rule. such a lookup would fail.

It is possible to circumvent this by defining IP blocks in the gateway, but that is essentially wrecking the system.

routing/v1alpha2/route_rule.proto
```diff
@@ -155,12 +157,12 @@ message RouteRule {
...
-// refer to a service in the service registry. It can be a short name or a
-// fully qualified domain name from the service registry, a resolvable DNS
-// name, an IP address or a service name from the service registry and a
-// subset name. The order of inference is as follows:
+// request/connection will be sent after processing a routing rule. The
+// destination.name should unambiguously refer to a service in the service
+// registry. It can be a short name or a fully qualified domain name from
+// the service registry, a resolvable DNS name, an IP address or a service
+// name from the service registry and a subset name. The order of inference
+// is as follows:
```
louiscryan: unfinished ?

routing/v1alpha2/destination_rule.proto
louiscryan: An IP address is really just another name, there's nothing inherently wrong with using IP addresses (particularly Kubernetes VIPs) to refer to services. We just reverse look them up to names whenever we can. If they don't have a name then the IP is the name.

I think we should just state that resolved names always take precedence over CIDR ranges and leave it at that

PR #276, opened 2017-12-15
source and gateways in match condition

rshriram:
Allows users to select which gateways a rule should apply to.
@vadimeisenbergibm / @louiscryan FYI.


routing/v1alpha2/route_rule.proto
```diff
@@ -62,7 +62,9 @@ package istio.routing.v1alpha2;
 // For example, the following rule routes all traffic by default to pods of
 // reviews service with label "version: v1". In addition, HTTP requests
 // containing for /wpcatalog/, /consumercatalog/ url prefixes will be
-// rewritten to /api/v2 and send to pods with label version: v2.
+// rewritten to /api/v2 and send to pods with label version: v2. The rule
+// will be applied at the gateway named bookinfo as well as all the
```
vadimeisenbergibm: Wording: The rule will be applied at the gateway named bookinfo as well as **at** all the sidecars

routing/v1alpha2/route_rule.proto
```diff
@@ -71,6 +73,9 @@ package istio.routing.v1alpha2;
 //     spec:
 //       hosts:
 //       - reviews
+//       gateways: #if omitted, defaults to "sidecars"
```
vadimeisenbergibm: if omitted defaults to "mesh"

routing/v1alpha2/route_rule.proto
```diff
@@ -117,13 +122,16 @@ message RouteRule {
...
-  // The names of gateways that should apply these routes. A single route
-  // rule could be used for sources inside the mesh as well as one or more
-  // gateways. The selection condition imposed by this field is
-  // independant of the sourceLabels match condition in HTTP/TCP routes. It
-  // is possible to define a route rule that applies to both the gateway
-  // and a specific internal workload by specifying gateways as well as the
-  // source match condition in HTTP/TCP routes.
+  // The names of gateways and sidecars that should apply these routes. A
+  // single route rule could be used for sidecars inside the mesh as well
+  // as one or more gateways. The selection condition imposed by this field
```
vadimeisenbergibm: ... as well as **for** one or more gateways

routing/v1alpha2/route_rule.proto
```diff
@@ -117,13 +122,16 @@ message RouteRule {
...
+  // can be overridden using source field in match condition in HTTP/TCP
+  // routes. The reserved word "mesh" is used to imply all sidecars in
```
vadimeisenbergibm: all **the** sidecars in the mesh

routing/v1alpha2/gateway.proto
```diff
@@ -109,8 +110,10 @@ package istio.routing.v1alpha2;
 //             name: reviews.qa
 //           weight: 20
 //
-// The following routing rule forwards traffic arriving at (external)
-// port 2379 from 172.17.16.* subnet to internal Mongo server on port 5555.
+// The following routing rule forwards traffic arriving at (external) port
+// 2379 from 172.17.16.* subnet to internal Mongo server on port 5555. This
```
ZackButcher: nit: use a CIDR address to describe the subnet (`172.17.16.0/24`)?

routing/v1alpha2/route_rule.proto
```diff
@@ -117,13 +122,16 @@ message RouteRule {
...
+  // the mesh. When this field is omitted, the default gateway ("mesh")
+  // will be used, which would apply the rule to all sidecars in the
```
vadimeisenbergibm: to all **the** sidecars in the mesh

routing/v1alpha2/route_rule.proto
```diff
@@ -117,13 +122,16 @@ message RouteRule {
...
+  // mesh. If a list of gateway names are provided, the rule will apply
```
vadimeisenbergibm: If a list of gateway names **is** provided...

routing/v1alpha2/route_rule.proto
ZackButcher: s/bookinfo/"bookinfo"

routing/v1alpha2/route_rule.proto
```diff
@@ -117,13 +122,16 @@ message RouteRule {
...
 
-  // The names of gateways that should apply these routes. A single route
-  // rule could be used for sources inside the mesh as well as one or more
-  // gateways. The selection condition imposed by this field is
-  // independant of the sourceLabels match condition in HTTP/TCP routes. It
-  // is possible to define a route rule that applies to both the gateway
-  // and a specific internal workload by specifying gateways as well as the
-  // source match condition in HTTP/TCP routes.
+  // The names of gateways and sidecars that should apply these routes. A
+  // single route rule could be used for sidecars inside the mesh as well
```
ZackButcher: s/could/can

routing/v1alpha2/route_rule.proto
```diff
@@ -270,7 +278,7 @@ message TCPRoute {
 // route rule to be applied to the HTTP request. For example, the following
 // route rule restricts the rule to match only requests where the URL path
 // starts with /ratings/v2/ and the request contains a "cookie" with value
-// "user=jason",
+// "user=jason", coming either from pods
```
vadimeisenbergibm: it seems the sentence is not finished: coming either from pods... **or** ...

routing/v1alpha2/route_rule.proto
```diff
@@ -361,8 +369,15 @@ message HTTPMatchRequest {
...
 
   // One or more labels that constrain the applicability of a rule to
-  // sources with the given labels.
+  // sidecars with the given labels. If the route rule has a list of
+  // gateways specified at the top, it should include the reserved gateway
+  // "mesh" in order for this field to be applicable.
   map<string, string> source_labels = 7;
+
+  // Names of gateways where the rule should be applied to. Gateway names
+  // at top of the rule (if any) are overridden. The gateway match is
```
vadimeisenbergibm: at **the** top of the rule

routing/v1alpha2/route_rule.proto
```diff
@@ -117,13 +122,16 @@ message RouteRule {
...
-  // rule could be used for sources inside the mesh as well as one or more
-  // gateways. The selection condition imposed by this field is
-  // independant of the sourceLabels match condition in HTTP/TCP routes. It
-  // is possible to define a route rule that applies to both the gateway
-  // and a specific internal workload by specifying gateways as well as the
-  // source match condition in HTTP/TCP routes.
+  // The names of gateways and sidecars that should apply these routes. A
+  // single route rule could be used for sidecars inside the mesh as well
+  // as one or more gateways. The selection condition imposed by this field
+  // can be overridden using source field in match condition in HTTP/TCP
```
ZackButcher: => using the `source` field in the match conditions of HTTP/TCP routes.

routing/v1alpha2/route_rule.proto
```diff
@@ -428,8 +443,15 @@ message L4MatchAttributes {
...
 
   // One or more labels that constrain the applicability of a rule to
-  // sources with the given labels.
+  // sidecars with the given labels. If the route rule has a list of
+  // gateways specified at the top, it should include the reserved gateway
+  // "mesh" in order for this field to be applicable.
   map<string, string> source_labels = 4;
+
+  // Names of gateways where the rule should be applied to. Gateway names
+  // at top of the rule (if any) are overridden. The gateway match is
```
vadimeisenbergibm: at **the** top of the rule

routing/v1alpha2/route_rule.proto
ZackButcher: I assume you were going to add a gateway to this example and say

"coming either from pods or outside of the network" or similar?

routing/v1alpha2/gateway.proto
```diff
@@ -86,6 +86,7 @@ package istio.routing.v1alpha2;
 //       - eu.bookinfo.com
 //       gateways:
 //       - my-gateway
+//       - mesh #applies to sidecars in the mesh
```
ZackButcher: `s/#applies/# applies`

vadimeisenbergibm: to **the** sidecars in the mesh

routing/v1alpha2/route_rule.proto
louiscryan: s/can/is

routing/v1alpha2/route_rule.proto
```diff
@@ -361,8 +369,15 @@ message HTTPMatchRequest {
   PortSelector port = 6;
 
   // One or more labels that constrain the applicability of a rule to
-  // sources with the given labels.
+  // sidecars with the given labels. If the route rule has a list of
```
louiscryan: its not about "sidecards with the given labels" - they dont have labels, the workloads they are attached to do.

louiscryan:
Largely LG. Schema is good, just to doc cleanup

routing/v1alpha2/route_rule.proto
```diff
@@ -117,13 +122,16 @@ message RouteRule {
...
+  // The names of gateways and sidecars that should apply these routes. A
+  // single route rule is used for sidecars inside the mesh as well
+  // as for one or more gateways. The selection condition imposed by this field
+  // can be overridden using the source field in the match conditions of HTTP/TCP
+  // routes. The reserved word "mesh" is used to imply all the sidecars in
+  // the mesh. When this field is omitted, the default gateway ("mesh")
+  // will be used, which would apply the rule to all sidecars in the
+  // mesh. If a list of gateway names is provided, the rule will apply
+  // only to the gateways. To apply a rule to both gateways and sidecars,
+  // specify "mesh" as one of the gateway names.
```
frankbu: This "mesh" reserved name approach seems kind of inelegant. Why not just make rules apply to all internal sidecars by default and use `source_labels` to filter them out? That's the way it works when you restrict access to specific internal sources, so it would seem like a clean extension to just do the same to restrict the rule to only the gateway(s) source. Something like this would apply the rule only to the gateways:
```
     spec:
       hosts:
       - reviews
       gateways:
       - bookinfo
       http:
       - match:
         - sourceLabels:
           server: istio-gateway
         route:
         - destination:
             name: reviews
           ...
```

routing/v1alpha2/gateway.proto
```diff
@@ -86,6 +86,7 @@ package istio.routing.v1alpha2;
 //       - eu.bookinfo.com
 //       gateways:
 //       - my-gateway
+//       - mesh # applies to all the sidecars in the mesh
```
costinm: You mean for the internal service-to-service traffic ?

routing/v1alpha2/route_rule.proto
costinm: It also leaves the issue of finer granularity: how about per-namespace ? 
Could it be mesh://foonamespace ? 
In general - we need some form of namespace or URL for gateway as well, 
and everything else. Even if k8s is not used, some internal organization is needed.

routing/v1alpha2/route_rule.proto
```diff
@@ -428,8 +443,15 @@ message L4MatchAttributes {
...
   // One or more labels that constrain the applicability of a rule to
-  // sources with the given labels.
+  // workloads with the given labels. If the route rule has a list of
+  // gateways specified at the top, it should include the reserved gateway
+  // "mesh" in order for this field to be applicable.
   map<string, string> source_labels = 4;
+
+  // Names of gateways where the rule should be applied to. Gateway names
+  // at the top of the rule (if any) are overridden. The gateway match is
+  // independent of sourceLabels.
```
costinm: We have the source service and labels, destination service name and labels.

Would it make sense to use the same 'labels' model for gateways as well ? And do we expect
this to be used for the zero-vpn ( service -> egress -> ingress -> service ), or only regular ingress ?

routing/v1alpha2/route_rule.proto
frankbu: @costinm I think `sourceLabels` (and maybe other potential source identifiers) is the way to handle finer granularity. The problem with using this `gateway` special value approach is that it is not consistent with the fine-grain filtering approach we use for internal calls.

louiscryan: @frankbu this option was something of a compromise but it was better than the other options that were toyed with, e.g.

- gateways & boolean field to disableMesh' -- Want the default of a boolean field to be false so would need to use a negative 'disableMesh' - yuck

- gateways & source labels -- Want to promote the one-route-rule-per-host model which would be hard if we allowed source labels at the top-level. Given that we expect source-labels to be rarer we pushed them down to matches. However we do expect the segregration of hosts between mesh & gateways to frequently be per-host and to make more sense at the top-level

We didn't want to use labels to match gateways as we don't actually dictate that they have labels at all (e.g. they may be run outside of Kubernetes). Also gateways have identifiers that represent their intent and we didn't expect there to be too many of them, hence the desire to reference them by name.

routing/v1alpha2/route_rule.proto
louiscryan: We could move to such a model later but I'd expect binding to mostly be done by name because a named gateway defines a virtual host. We can always support labels later if needed.

routing/v1alpha2/route_rule.proto
frankbu: @louiscryan I like the boolean `disableMesh` option much better. The negative term doesn't bother me, but you could always name it something like `gatewayOnly` or `gatewayExclusive` instead.

What I find really undesirable about the reserved gateway name approach is that `mesh` is a default gateway until other gateways are added. This switching the default behavior, not to mention the fact that `mesh` isn't really a gateway anyway, seems pretty ugly.

I'm also wondering if you've considered making `sourceLabels` in match clauses override the visibility defined at the top level gateways? It would seem more natural to say that `sourceLabels` override applicability of the individual rule rather than the current approach which says they are ignored.

PR #277, opened 2017-12-20
Add Makefile for mixer/v1 proto generation

douglas-reid:
This PR removes the `mixer/v1` bazel code for protoc-based generation, replacing it that logic with a Makefile.

Notes:
- `generate-protos.sh` is updated to call `make` targets
- all BUILD files were rewritten with `gazelle` to supply default targets around the generated files
- each of the `.proto` files is touched, as it is regenerated as part of this PR

Reviewers: I am not fluent in Makefile. Please take a closer look there. I'm not aiming for perfect, but rather something functional. It can be improved iteratively by those with better fluency.


rkpagadala:
@douglas-reid I don't see any Makefile in the PR. Did you forget to add it?

douglas-reid:
@rkpagadala apparently, yes. this is why we don't create PRs when running out to lunch ;).

Makefile
```diff
@@ -0,0 +1,128 @@
+# GOOGLEAPIS 
+GOOGLEAPIS_SHA = c8c975543a134177cc41b64cbbf10b88fe66aa1d
```
ldemailly: use := everywhere where you don't want the variable to be re-evaluated

Makefile
```diff
@@ -0,0 +1,128 @@
...
+ERR_PROTO = protoc-tmp/$(RPC_PATH)/error_details.proto
+
+PROTOC = protoc-min-version -version=3.5.0
+
+GOGOSLICK_PLUGIN_PREFIX = --gogoslick_out=plugins=grpc=
+GOGO_PLUGIN_PREFIX = --gogo_out=plugins=grpc=
+PLUGIN_SUFFIX = :.
+
+# BASIC STANDARD MAPPINGS
+MAPPING := Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,
```
ldemailly: what's the M for ? maybe a comment ? also maybe consider a list and add the Mgoogle/protobuf part in a loop?

Makefile
```diff
@@ -0,0 +1,128 @@
...
+
+MIXER_TEMPLATE = mixer/v1/template
+MIXER_TEMPLATE_PROTOS = $(MIXER_TEMPLATE)/extensions.proto $(MIXER_TEMPLATE)/standard_types.proto
+MIXER_TEMPLATE_PB_GOS = $(MIXER_TEMPLATE_PROTOS:.proto=.pb.go)
+
+generate: generate-mixer-go
+
+install-deps:
+	# Installing generation deps
+	@dep ensure
```
ldemailly: I find that using @ doesn't help troubleshooting - I would only use @ in front of echo commands

kyessenov: @costinm @ldemailly @rshriram as you guys seem to be Makefile aficionados, do you mind coming up with some style guide for Makefiles? let's circumvent bash nit-picking with some general guide how to write these things.

Makefile
douglas-reid: done.

Makefile
douglas-reid: M controls mapping from protos to the go package.

Added foreach for control.

Makefile
douglas-reid: removed '@'s everywhere for now.

douglas-reid:
I don't want to alarm anyone, but this PR just passed the prow build.

Makefile
```diff
@@ -0,0 +1,187 @@
...
+comma := ,
+empty:=
+space := $(empty) $(empty)
+
+GOGO_PROTO_PKG := github.com/gogo/protobuf/gogoproto
+GOGO_TYPES := github.com/gogo/protobuf/types
+GOGO_DESCRIPTOR := github.com/gogo/protobuf/protoc-gen-gogo/descriptor
+GOGO_GOOGLEAPIS_RPC := istio.io/gogo-genproto/googleapis/google/rpc
+
+importmaps := \
```
ldemailly: :+1:

douglas-reid:
w00t! Thanks!

Now to work on fixing istio/istio to match this PR.

PR #281, opened 2018-01-04
Add ServiceRole and ServiceRoleBinding proto.

liminw:
This is the first part of the change for implementing Istio RBAC adapter ([#1233)](https://github.com/istio/istio/issues/1233). 

Istio RBAC design doc: https://docs.google.com/document/d/1dKXUEOxrj4TWZKrW7fx_A-nrOdVD4tYolpjgT8DYBTY.


liminw:
cc @wora @geeknoid @mandarjog.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+//       - properties:
+//           service: "reviews"
+//           namespace: "abc"
+//       roleRef:
+//         kind: ServiceRole
+//         name: "products-viewer"
+
+// ServiceRole specification contains a list of access rules (permissions).
+message ServiceRole {
+  // Required. The set of access rules (permissions) that the role has.
```
wora: So the role doesn't have a name?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+  repeated AccessRule rules = 1;
+}
+
+// AccessRule defines a permission to access a list of services.
+message AccessRule {
+  // Required. A list of service names.
```
wora: What is the service name? Give an example, and map it to an attribute.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+  // If set to ["*"], it refers all services in the namespace.
+  repeated string services = 1;
+
+  // Optional. A list of HTTP paths.
+  // If not specified, it applies to any path.
+  repeated string paths = 2;
+
+  // Required. A list of HTTP methods or gRPC methods.
+  // If set to ["*"], it applies to any method.
+  repeated string methods = 3;
```
wora: I don't understand why we need 3 repeated fields inside a repeated message? Please show some examples.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+message ServiceRole {
+  // Required. The set of access rules (permissions) that the role has.
+  repeated AccessRule rules = 1;
+}
+
+// AccessRule defines a permission to access a list of services.
+message AccessRule {
+  // Required. A list of service names.
+  // If set to ["*"], it refers all services in the namespace.
+  repeated string services = 1;
```
wora: Why there is no namespace?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+  // If set to ["*"], it applies to any method.
+  repeated string methods = 3;
+
+  // Definition of a custom constraint.
+  message Constraint {
+    // The name of the constraint.
+    string name = 1;
+
+    // The list of valid values for the constraint.
+    repeated string value = 2;
```
wora: plural

Do we only support exact match? What about prefix, suffix, regex?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+//     kind: servicerole
+//     metadata:
+//       name: products-viewer
+//       namespace: default
+//     spec:
+//       rules:
+//       - services: ["products.svc.cluster.local"]
+//         methods: ["GET", "HEAD"]
+//         constraints:
+//         - name: "version"
```
wora: What version refers to?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+  repeated Constraint constraints = 4;
+}
+
+// ServiceRoleBinding assigns a ServiceRole to a list of subjects.
+message ServiceRoleBinding {
+  // Required. List of subjects that are assigned the ServiceRole object.
+  repeated Subject subjects = 1;
+
+  // Required. Reference to the ServiceRole object.
+  RoleRef roleRef = 2;
```
wora: This should be the first field.

ServiceRoleRef

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+// RoleRef refers to a ServiceRole object.
+message RoleRef {
+  // Required. The type of the role being referenced.
+  // Currently, "ServiceRole" is the only supported value for "kind".
+  string kind = 1;
+
+  // Required. The name of the ServiceRole object being referenced.
+  // The ServiceRole object must be in the same namespace as the ServiceRoleBinding
+  // object.
+  string name = 2;
```
wora: Why this is not part of role?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+  //     service: "reviews"
+  //     namespace: "abc"
+  map<string, string> properties = 1;
+}
+
+// RoleRef refers to a ServiceRole object.
+message RoleRef {
+  // Required. The type of the role being referenced.
+  // Currently, "ServiceRole" is the only supported value for "kind".
+  string kind = 1;
```
wora: Why do we create this open ended definition?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.rbac.v1;
+
+// Istio RBAC (Role Based Access Control) defines ServiceRole and ServiceRoleBinding
+// objects. Both are provided as Kubernetes CRDs.
```
geeknoid: Let's not talk about CRDs or k8s in here, these definitions are independent of the implementation details of k8s.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+//     spec:
+//       rules:
+//       - services: ["products.svc.cluster.local"]
+//         methods: ["GET", "HEAD"]
+//         constraints:
+//         - name: "version"
+//           value: ["v1", "v2"]
+//
+// A ServiceRoleBinding specification includes two parts:
+// * "roleRef" refers to a ServiceRole object in the same namespace.
```
geeknoid: Why is there a constraint that they be in the same namespace? Wouldn't you want to be able to have different ACLs on the roles vs. the bindings?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.rbac.v1;
+
+// Istio RBAC (Role Based Access Control) defines ServiceRole and ServiceRoleBinding
+// objects. Both are provided as Kubernetes CRDs.
+//
+// A ServiceRole specification includes a list of rules (permissions). Each rule has
```
geeknoid: Why ServiceRole and not Role?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,139 @@
...
+//       namespace: default
+//     spec:
+//       subjects:
+//       - properties:
+//           user: "alice@yahoo.com"
+//       - properties:
+//           service: "reviews"
+//           namespace: "abc"
+//       roleRef:
+//         kind: ServiceRole
```
geeknoid: Why is the kind useful here? Are there other choices?

rbac/v1/rbac.proto
geeknoid: The name/namespace is part of the 'envelope' around this proto. This proto just defines the "spec" stanza of the above examples

rbac/v1/rbac.proto
liminw: version is a custom field. I added comments and examples to explain it. 

// In addition to the standard fields, operators can use custom fields in the "constraints"
// section. The name of a custom field must match one of the "properties" in the "action" part
// of the "authorization" template (https://github.com/istio/istio/blob/master/mixer/template/authorization/template.proto).
//
// For example, suppose we define an instance of the "authorization" template, named "requestcontext".
//
//     apiVersion: "config.istio.io/v1alpha2"
//     kind: authorization
//     metadata:
//       name: requestcontext
//       namespace: istio-system
//     spec:
//       subject:
//         user: source.user | ""
//         groups: source.user | ""
//         properties:
//           service: source.service | ""
//           namespace: source.namespace | ""
//       action:
//         namespace: destination.namespace | ""
//         service: destination.service | ""
//         method: request.method | ""
//         path: request.path | ""
//         properties:
//           version: request.headers["version"] | ""

rbac/v1/rbac.proto
liminw: Yes, as Martin pointed out, the ServiceRole and ServiceRoleBinding protos just defines the "spec" part of the objects. I added comments before ServiceRole and ServiceRoleBinding definitions to clarify it.

rbac/v1/rbac.proto
liminw: The ServiceRole/ServiceRoleBinding example and Instance example (mapping to attributes) are provided in the comments.

rbac/v1/rbac.proto
liminw: namespace is provided in metadata.

rbac/v1/rbac.proto
liminw: See the examples in the comment.

rbac/v1/rbac.proto
liminw: Changed "value" to "values". 

I plan to support exact match, prefix, suffix matches. I added comments and examples for that.

rbac/v1/rbac.proto
liminw: The main reason is to keep it consistent with K8S RBAC. Also, if in the future, we support additional types of roles, say, MeshServiceRole (e.g, for all services having "abc" prefix in the entire mesh), ServiceRoleBinding can bind to that type of role as well.

rbac/v1/rbac.proto
liminw: It is in the metadata of ServiceRole object.

rbac/v1/rbac.proto
liminw: I removed the places mentioning CRD.

rbac/v1/rbac.proto
liminw: "Role" is already used by K8S RBAC.

rbac/v1/rbac.proto
spikecurtis: Since they are different `Kind`s, access control can be different for objects in the same namespace.

Requiring the ServiceRole and ServiceRoleBinding be in the same namespace is for simplicity, so you can refer to ServiceRoles by name (rather than name and namespace). It is somewhat opinionated---that operators group these objects in the namespace where the services they apply to reside.

rbac/v1/rbac.proto
liminw: Hong asked the same question. Just copy&paste my answer here. 

The main reason is to keep it consistent with K8S RBAC. Also, if in the future, we support additional types of roles, say, MeshServiceRole (e.g, for all services having "abc" prefix in the entire mesh), ServiceRoleBinding can bind to that type of role as well.

liminw:
@wora Hong, thanks for the comments. Regarding the RBAC design and how it aligns with Cloud IAM. We had a lot of discussions and meetings earlier with Cloud IAM team, RpcSecurityPolicy team, K8S team to evaluate different alternatives. The consensus was that we should move forward with the current design.

liminw:
@wora @geeknoid Thanks for the comments! I think I have addressed all your comments. Please take a look at the updated files when you have time.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,185 @@
...
+
+// Istio RBAC (Role Based Access Control) defines ServiceRole and ServiceRoleBinding
+// objects.
+//
+// A ServiceRole specification includes a list of rules (permissions). Each rule has
+// the following standard fields:
+// * services: a list of services.
+// * methods: HTTP methods or gRPC methods. Note that gRPC methods should be
+//   presented in the form of "packageName.serviceName/methodName".
+// * paths: HTTP REST paths. It is ignored in gRPC case.
```
spikecurtis: nit: drop the reference to REST, which is a specific way of structuring APIs. Any HTTP API path can be selected with this.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,185 @@
...
+//
+// In addition to the standard fields, operators can use custom fields in the "constraints"
```
spikecurtis: I thought that per our discussion with Joe & Jordan, we were going to narrow this down to specific, named fields, rather than allow custom ones.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,185 @@
...
+//     metadata:
+//       name: test-binding-products
+//       namespace: default
+//     spec:
+//       subjects:
+//       - properties:
+//           user: "alice@yahoo.com"
+//       - properties:
+//           service: "reviews"
+//           namespace: "abc"
```
spikecurtis: Can we simplify this:
```
spec:
  subjects:
  - user: "alice@yahoo.com"
  - service: "reviews"
    namespace: "abc"
```

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,185 @@
...
+//       roleRef:
+//         kind: ServiceRole
+//         name: "products-viewer"
```
spikecurtis: should be
```
roleRef:
- kind: ServiceRole
  name "products-viewer"
```

rbac/v1/rbac.proto
liminw: Updated. Thanks.

rbac/v1/rbac.proto
liminw: I remember there were quite some confusion on the custom fields when we discussed with Joe and Jordan. I think it provide a lot of flexibility by providing custom fields. As long as we provide good documentation, it should be fine to support it. What do you think? 

One thing we should be fix is that (also mentioned in the discussion with Joe and Jordan), for "subject", we should only use authenticated attributes, instead of any attributes. We can add that restriction in the next step.

rbac/v1/rbac.proto
liminw: I tried to do that. The problem is, for that purpose, I need to define "subjects" as repeated map (as follows). But "repeated map" is not supported in proto.

  repeated map<string, string> subjects = 1;

So I had to define a "Subject" type, and define a map inside "Subject".

  repeated Subject subjects = 1;

message Subject {
  map<string, string> properties = 1;
}

Let me know if you have a better suggestion.

rbac/v1/rbac.proto
liminw: With current proto, ServiceRole and ServiceRoleBinding is 1 to n. That is, a ServiceRoleBinding can be bound to one ServiceRole. This is for simplicity. K8S RBAC has the same restriction. A binding can only refer to one role.

rbac/v1/rbac.proto
spikecurtis: In the template we always have a `user`, and `group`, so maybe
```
message Subject {
  string user = 1;
  string group = 2;
  map<string, string> properties = 3;
}

Comment (liminw) [reply rbac/v1/rbac.proto]: Sounds good. I made the change as you suggested.

File: rbac/v1/rbac.proto (hunk 17 above)
Comment (spikecurtis) [reply rbac/v1/rbac.proto]: kk

**Comment by liminw** [comment]
Gently ping for further comments or approval.

File: rbac/v1/rbac.proto [hunk 18]
```diff
@@ -0,0 +1,187 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.rbac.v1;
+
+// Istio RBAC (Role Based Access Control) defines ServiceRole and ServiceRoleBinding
+// objects.
+//
+// A ServiceRole specification includes a list of rules (permissions). Each rule has
```
geeknoid: Is there a reason we call these rules instead of actually just calling them permissions?

rbac/v1/rbac.proto
geeknoid: This constraint makes me uncomfortable, it doesn't seem like a robust story. Can we have it so there is a namespace field and it is empty than it means "current namespace"?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,187 @@
...
+// ServiceRole specification contains a list of access rules (permissions).
+// This represent the "Spec" part of the ServiceRole object. The name and namespace
+// of the ServiceRole is specified in "metadata" section of the ServiceRole object.
+message ServiceRole {
+  // Required. The set of access rules (permissions) that the role has.
+  repeated AccessRule rules = 1;
+}
+
+// AccessRule defines a permission to access a list of services.
+message AccessRule {
```
geeknoid: Why are all the fields in here repeated, given that then ServiceRole contains a repeated set of AccessRule messages? Isn't this redundant?

Can you clarify in the docs the OR vs. AND semantics that apply here. ServiceRole ORs all the rules together for example. What about all the fields in a single rule?

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,187 @@
...
+
+  // Optional. The set of properties that identify the subject.
+  // In the above ServiceRoleBinding example, the second subject has two properties:
+  //     service: "reviews"
+  //     namespace: "abc"
+  map<string, string> properties = 3;
+}
+
+// RoleRef refers to a ServiceRole object.
+message RoleRef {
```
geeknoid: Shouldn't this be ServiceRoleRef?

rbac/v1/rbac.proto
liminw: Just to be consistent with K8S RBAC terms.

rbac/v1/rbac.proto
liminw: Each AccessRule defines a permission. A role is a collection of permissions. Inside a single AccessRule (permission), all the fields are ANDs. This is the same way K8S RBAC was defined. 

The example in the doc defines a single permission in the role. 

//     apiVersion: "config.istio.io/v1alpha2"
//     kind: servicerole
//     metadata:
//       name: products-viewer
//       namespace: default
//     spec:
//       rules:
//       - services: ["products.svc.cluster.local"]
//         methods: ["GET", "HEAD"]
//         constraints:
//         - name: "version"
//           value: ["v1", "v2"]

The explanation in the doc says: 

Below is an example of ServiceRole object "product-viewer", which has "read" ("GET" and "HEAD") access to "products.svc.cluster.local" service at versions "v1" and "v2".

mandarjog:
@liminw can you put a link to the approved design document that this PR is based on?

rbac/v1/rbac.proto
liminw: Updated the comment.

liminw:
@mandarjog I added a link to the approved design doc.

rbac/v1/rbac.proto
geeknoid: I still don't understand what the semantic different would be between:

N AccessRules where each field in the AccessRule is a singleton
1 AccessRule where each field has N values.

Are those logically equivalent?

rbac/v1/rbac.proto
spikecurtis: I misspoke before about the reason for this, and am now remembering the design reasoning.

The ServiceRole establishes permissions, and the ServiceRoleBinding grants those permissions to subjects.  Both of these need to be in the same namespace as the service(s) we are granting permissions on, so that the namespace can be an effective security boundary.  Imagine a service foo in namespace A.  Alice operates the services in namespace A, and she creates an admin role that can access foo.  Bob operates namespace B and has no permissions in namespace A.  If Bob is allowed to create bindings that reference Alice's foo admin role, then Alice can't control who has access to foo anymore.  Bob could bind Alice's foo admin role to himself or anyone he wants.

Instead we say only bindings in A can reference roles in A.  This follows the [Kubernetes RBAC model](https://kubernetes.io/docs/admin/authorization/rbac/#rolebinding-and-clusterrolebinding).

geeknoid:
Ah, that makes sense.

rbac/v1/rbac.proto
liminw: I composed a simple example to illustrate the differences: 

Role A has the following rules (N AccessRules where each field in the AccessRule is a singleton):
- services: [products]
  methods: [get]
- services: [reviews]
  methods: [post]

Role B has the following rule (1 AccessRule where each field has N values):
- services: [products, reviews]
  methods: [get, post]

Role A and Role B are different because: 
* Role B allows you to do "post" on "products" service, but Role A does not. 
* Role B allows you to do "get" on "reviews" service, but Role A does not.

geeknoid: Thank you for explaining this so clearly. That really helped me get it.

rbac/v1/rbac.proto
```diff
@@ -0,0 +1,187 @@
...
+
+  // Optional. The set of properties that identify the subject.
+  // In the above ServiceRoleBinding example, the second subject has two properties:
+  //     service: "reviews"
+  //     namespace: "abc"
+  map<string, string> properties = 3;
+}
+
+// RoleRef refers to a role object.
+message RoleRef {
```
geeknoid: I thought I commented on this before, but I don't see the comment anywhere.

Any reason why this is not ServiceRoleRef?

liminw: This was my main reasoning on keeping an open-typed RoleRef: 

The main reason is to keep it consistent with K8S RBAC. Also, if in the future, we support additional types of roles, say, MeshServiceRole (e.g, for all services having "abc" prefix in the entire mesh), ServiceRoleBinding can bind to that type of role as well.

wora:
I don't think this should be approved as istio.rbac.v1. The comment says v1alpha2 in many places. RBAC is an extremely complicated topic. It is very unlikely to get it right without prototype. How does it integrate with other systems would be an even bigger challenge. I think we should have one version before v1 to test things out.

PR #285, opened 2018-01-05
Destination Rules

rshriram:


routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,352 @@
...
+    // balancer will use a random number as the hash, effectively making
+    // the load balancing policy random.
+    string http_header = 1;
+
+    // The minimum number of virtual nodes to use for the hash
+    // ring. Defaults to 1024. Larger ring sizes result in more granular
+    // load distributions. If the number of hosts in the load balancing
+    // pool is larger than the ring size, each host will be assigned a
+    // single virtual node.
+    uint32 minimum_ring_size = 2;
```
wora: All other fields are using int32, why this one is special?

routing/v1alpha2/destination_rule.proto
```diff
@@ -0,0 +1,352 @@
...
+//       destination:
+//         name: myredissrv
+//       connectionPool:
+//         tcp:
+//           maxConnections: 100
+//           connectTimeout: 30ms
+//
+message ConnectionPoolSettings {
+  // Settings common to both HTTP and TCP upstream connections.
+  message TCPSettings {
```
wora: Do we use acronyms inside symbols? Envoy and Google APIs don't use acronyms.

routing/v1alpha2/destination_rule.proto
rshriram: Copied from Envoy API, which is also using uint32.

routing/v1alpha2/destination_rule.proto
rshriram: We have `TCP<*>`, and `HTTP<*>` in several places in our APIs and Envoy APIs.. (https://github.com/envoyproxy/data-plane-api/blob/master/api/filter/http/fault.proto#L26)..

PR #286, opened 2018-01-05
circleci config

rshriram:


rshriram:
Merging as this has nothing to do with code.. its circleci config

PR #287, opened 2018-01-05
Circleci tweaks and remove unnecessary bazel build in presubmit

rshriram:
Fix circleci build. Remove unnecessary bazel build step from scripts/generate-protos.sh. Circleci and prow use the same underlying script.


routing/v1alpha2/destination_rule.proto
```diff
@@ -90,6 +90,8 @@ message DestinationRule {
   // One or more named sets that represent individual versions of a
   // service. Traffic policies can be overridden at subset level.
   repeated Subset subsets = 3;
+
+  uint32 foo = 4;
```
douglas-reid: did you mean to add this to this PR ?

rshriram: YEs, for testing. removed it in subsequent commit

PR #294, opened 2018-01-09
TLS settings in DestinationRule

rshriram:
Introduce TLS settings (equivalent to ssl_context in the envoy cluster), that controls how upstream connections are secured. this will allow users to establish mTLS connections to egress services or mesh internal services using custom certificates.

Along the way, clean up some wording in TLS related areas, and added missing private_key field to gateway's TLS.


routing/v1alpha2/gateway.proto
```diff
@@ -214,42 +215,45 @@ message Server {
...
       PASSTHROUGH = 0;
 
       // If set to "simple", the gateway will secure connections with
-      // standard TLS semantics (server certs only).
+      // standard TLS semantics.
       SIMPLE = 1;
 
-      // If set to "mutual", the gateway will use standard mTLS authentication.
+      // If set to "mutual", the gateway will secure connection with the
+      // client using mTLS authentication.
```
ZackButcher: I think client is a bit ambiguous here. What about:

> If set to "mutual", the gateway will secure connections to the upstream using mutual TLS by presenting client certificates for authentication.

routing/v1alpha2/gateway.proto
```diff
@@ -214,42 +215,45 @@ message Server {
...
+    // secured using TLS. The value of this field determines how TLS is
     // enforced.  
     TLSmode mode = 2;
 
-    // REQUIRED if mode == SIMPLE/MUTUAL. The name of the file holding the
-    // server-side TLS certificate to use.  It is the responsibility of the
-    // underlying platform to mount the certificate as a file under
-    // /etc/istio/ingress-certs with the same name as the specified in this
-    // field.
+    // REQUIRED if mode == SIMPLE/MUTUAL. The path to the file holding the
```
ZackButcher: I think "REQUIRED if mode is SIMPLE or MUTUAL" reads a bit better.

ZackButcher: And above on the enum we use `"mutual"` to describe values `mode` can take, and `MUTUAL` to refer to the same values in every other context. We should standardize.

routing/v1alpha2/gateway.proto
```diff
@@ -214,42 +215,45 @@ message Server {
...
+    string private_key = 4;
+
+    // REQUIRED if mode == MUTUAL. The path to a file containing
+    // certificate authority certificates to use in verifying a presented
+    // client side certificate.
+    string ca_certificates = 5;
+
+    // A list of alternate names to verify the subject identity in the
+    // certificate presented by the client.
+    repeated string subject_alt_names = 6;
```
ZackButcher: Can SANs encode wildcard names? Do we support wildcards in the names in this field? We should mention in the comment.

rshriram: Oh.. @myidpt ^^^^

routing/v1alpha2/destination_rule.proto
```diff
@@ -350,3 +355,78 @@ message OutlierDetection {
...
+  string client_certificate = 2;
+
+  // REQUIRED if mode is "mutual". The path to the file holding the
+  // client's private key.
+  string private_key = 3;
+
+  // OPTIONAL: The path to the file containing certificate authority
+  // certificates to use in verifying a presented server certificate. If
+  // omitted, the proxy will not verify the server's certificate.
+  string ca_certificates = 4;
```
wora: Why this one is plural?

ZackButcher: It represents a cert bundle, not a single cert, so the plural makes sense in that context.

PR #296, opened 2018-01-10
Change rbac version to "v1alpha1".

liminw:


liminw:
cc @wora

rbac/v1alpha1/rbac.proto
```diff
@@ -32,7 +32,7 @@ package istio.rbac.v1;
 //
 // For example, suppose we define an instance of the "authorization" template, named "requestcontext".
 //
-//     apiVersion: "config.istio.io/v1alpha2"
+//     apiVersion: "config.istio.io/v1alpha1"
 //     kind: authorization
```
wora: is kind title case or lower case?

kyessenov:
@geeknoid 
It seems that there are some comments, and the API might need a few iterations to be included in Istio. Do you think it would be a good idea to continue working on a branch before merging into master? It's difficult to quickly iterate on an API once istio has some code that uses it, since it leads to repos going out of sync, and blocking subsequent PRs to istio/api.

geeknoid:
Limin, what do you think? Can you address Hong's questions, or should we
move this to a branch for further bake time?

liminw:
@wora Sorry for replying late. 

gRPC has two types, some supports HTTP transcoding, some do not (i.e., native gRPC). You mentioned the case that gRPC supports HTTP transcoding, but we also need to support native gRPC calls. 

"constraints" limits the permission to a certain condition. For example, if you have a constraint "version: v1", it means that the permission only applies to services with "v1" version. If you think that is not a good name, I can rename it to "properties" or something else. 

Properties are not labels or tags, they are mapped from attributes. Calling them labels or tags will be confusing. 

I am not sure what you mean "k8s has standard way to refer resources". k8s RBAC also has "roleRef" referring to roles. Here is an example of k8s RBAC. 

kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: read-pods
  namespace: default
subjects:
- kind: User
  name: jane
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: Role
  name: pod-reader
  apiGroup: rbac.authorization.k8s.io

rbac/v1alpha1/rbac.proto
liminw: I am going to change kind to CamelCase "Authorization". It seems that
* Kubernetes uses CamelCase for kind. For example, "RoleBinding", "ClusterRole". 
* Pilot also uses CamelCase. For example, "RouteRule". 
* Mixer uses lower case. For example, "attributemanifest", "listchecker". 

We should fix the inconsistency. @geeknoid @mandarjog

kyessenov:
Yeah, we should use camel case. Just because k8s is lenient now, doesn't
mean it's going to stay that way.

wora:
I chatted with @liminw in person yesterday. I think we have reached an agreement. She will make some changes based on our discussion.

liminw:
@wora I have addressed your comments in the new commits.

@geeknoid @kyessenov I filed Issue #2633 to track CRD kind CamelCase vs. lowercase problem.

PR #297, opened 2018-01-10
rename gateway to exposure for clarity

rshriram:
The top level gateways field in the route rule has been causing confusion, with people asking how is the mesh a gateway, and whether it can be tinkered with, etc. In addition, there has been confusion about whether a route rule can have just gateways and no host declaration. To clarify all of these, I have renamed the top level gateways array to exposure.

```yaml
kind: RouteRule
spec:
 hosts:
 - foo.com
 exposure:
     sidecarVisible: True
     gateways:
     - bookinfo-lb
http:
 ...
```

Secondly, the domains field in the gateway has been renamed to hosts, to align with the usage in route rules, destination rules, foreign services, etc.

cc @frankbu @ZackButcher @louiscryan


kyessenov:
It's still confusing to have mesh and gateway on the same level. I don't know what "mesh" is, and my best guess it's some sort of special gateway.

rshriram:
@kyessenov take a look at the format now.. got rid of mesh

kyessenov:
How about you add a boolean field to the rule `gateway_only`. It's false by default, and that means applicable to sidecars. The non-default true value, restricts it to gateways.

rshriram:
how do I indicate that a rule is applicable to both gateways and sidecars? May be an enum with sidecars, gateways, sidecar_and_gateways ?

kyessenov:
Rules are applicable to sidecar when the flag is not set.
Rules are applicable to gateways explicitly listed in the field (this needs to be improved).

rshriram:
and that is the reason I added the exposure message.. See the PR description, not the comments as I have not updated them yet

louiscryan:
@kyessenov Shriram and I worked through using separate fields to segregate gateways from mesh and decided against it for conciseness. In addition at some point down the road we may use the gateway resource to actually model the sidecar.

routing/v1alpha2/gateway.proto
```diff
@@ -84,7 +84,7 @@ package istio.routing.v1alpha2;
 //       - reviews.prod
 //       - uk.bookinfo.com
 //       - eu.bookinfo.com
-//       gateways:
+//       exposure:
```
louiscryan: Why are we changing this term?

rshriram: Per this PR.. I forgot about the discussion we had. This PR was an offshoot of the confusion we had in Zack's ingress doc and some comments from @frankbu. Just wanted to see how it looked

routing/v1alpha2/route_rule.proto
```diff
@@ -137,17 +137,23 @@ message RouteRule {
...
-  // routes. The reserved word "mesh" is used to imply all the sidecars in
-  // the mesh. When this field is omitted, the default gateway ("mesh")
-  // will be used, which would apply the rule to all sidecars in the
-  // mesh. If a list of gateway names is provided, the rule will apply
-  // only to the gateways. To apply a rule to both gateways and sidecars,
-  // specify "mesh" as one of the gateway names.
-  repeated string gateways = 2;
+  // Exposure controls the locations in the mesh where the rule should be
+  // applied.
+  message Exposure {
```
louiscryan: We discussed this pattern in detail before and rejected it as it leads to an API that is hard to parse when the values are partially overridden. I'm not a fan of this and not sure why we're rehashing this discussion

kyessenov:
Shouldn't readability trump conciseness for this particular case? Don't you agree that `mesh` as a keyword is confusing to most people?

louiscryan:
Its not just about conciseness, the readability concern is what do overrides look like. If we have two fields folks have to remember to override both and will likely get it wrong.

We are also reserving the keyword to allow for the fact that we may the 'gateway' resource to configure sidecars, a need we'll likely need to meet.

I'm fine with reserving a different name if we can think of a better one but the obvious ones were evaluated and rejected 'sidecar' (implies a deployment model), "envoy" (vendor specific) etc.

routing/v1alpha2/route_rule.proto
kyessenov: If we pull the bit up to top-level, then there's no partial override problem, right?

kyessenov:
For future work, it would be good to work out scoping mechanisms for gateways (namespaces and wildcards). That would help with mapping ingress scopes to gateway scopes. cc @ZackButcher

frankbu:
I think `mesh` as a keyword is confusing but the fact that `mesh` is a default gateway unless other gateways are explicitly added (and then it's not anymore) is even more confusing. I think the boolean solution is better, but the name should be something like `gatewayOnly` (false default), and I don't think that the `exposure` wrapper is needed.
```
kind: RouteRule
spec:
  hosts:
  - foo.com
  gateways:
  - bookinfo-lb
  gatewaysOnly: True // Set to True for rules not applicable for internal calls
http:
 ...
```

kyessenov:
I also suggest a better gateway selector than a list of names. In a basic case, the rule should simply apply to all gateways, without having to list them all by name one by one.

louiscryan:
@kyessenov @frankbu 

I still don't think that works. What about situations like

- We have two different sidecars
- We don't have any sidecars

If and when we want to configure a sidecars listeners or extensions this will make it hard to use gateway to achieve that effect (gateways are still named after all)

Then there's the 'override to empty' problem. Extending your example how do I override in a sub-route to restrict the rule to just the sidecar?

```
kind: RouteRule
spec:
  hosts:
  - foo.com
  gateways:
  - bookinfo-lb
  gatewaysOnly: True 
http:
   match:
      ...
     gatewaysOnly: False  # Does user intend to limit this route to the sidecar while still allowing bookinfo-lb ?
    gateways: []  # ... so I'm forced to declare an empty array to achieve the effect, not good

louiscryan:
.. and no by default a rule should NOT apply to all gateways, that would not be expected behavior. Applying them to all sidecars by default makes sense which is what we have now.

frankbu:
@louiscryan 
I would say that any override of what @shriram was calling `exposure` attributes, wipes out the top level definitions, so both `gateways` and `gatewaysOnly`. In your example, the `match.gatewaysOnly: True`, will also default the rule back to `gateways: []`. Maybe this is the reason @shriram introduced the `exposure` wrapper in the first place? It's less concise, but very straightforward.

routing/v1alpha2/route_rule.proto
wora: This name is too hard to reason. Can we move this change to a separate PR and discuss there?

frankbu:
It's OK if the end result of this discussion is to leave the design as is, but I don't think closing the PR in the middle of the discussion makes sense. At least include a comment that summarizes why it was decided to end the discussion.

rshriram:
Fair enough @frankbu .. Sorry to close this prematurely. I have updated the PR slightly, but the crux of the issue remains the same. Whether to call the mesh as a virtual gateway or not, so that we can apply some gateway style transformations for the mesh as well.

Given that we are arguing about the user perception and feedback, lets actually ask some of the them..  @pacevedom / @christian-posta / @nmnellis / @rosenhouse / whoever else is interested..

PR #300, opened 2018-01-10
Fix fixed_cfg.pb.go generation

douglas-reid:
This PR wires-in generation of the fixed_cfg.pb.go file into `make generate`.


douglas-reid:
This should be abandoned once rolled into: https://github.com/istio/api/pull/298.

PR #304, opened 2018-01-11
Introduce Value for template-time equivalent of ValueType enum.

guptasu:
After this PR, I will replace all the existing templates to use standard_type:Value instead of ValueType enum. In the same PR I will also introduce pkg/adapter/ValueType enum which will be referenced by all the adapters instead of api/mixer/v1/config/descriptor/ValueType.


guptasu:
@rshriram can you push this change.. Circeci is not liking my change for some reason.

PR #306, opened 2018-01-12
Include hash policy header configuration in v1alpha2.route

pacevedom:
cc @rshriram
I changed the http_header configuration from DestinationRule to HTTPRoute. The way the configuration is fetched via xDS makes this quite difficult to do elegantly, plus another problem.
You can have multiple clusters on the same route, and each of these clusters could have a different load balancing strategy. If two (or more) state that they want hashing, each of those can also define a different header_name, and the HTTPRoute is only able to handle one.


rshriram:
@pacevedom this is a good catch. Thanks for following it up.

I agree that this is kind of very yucky. I do not like straddling routerule and destinationrule boundaries (but this is unavoidable as you point out)

routing/v1alpha2/destination_rule.proto
```diff
@@ -212,20 +212,14 @@ message LoadBalancerSettings {
...
-    // the load balancing policy random.
-    string http_header = 1;
-
     // The minimum number of virtual nodes to use for the hash
     // ring. Defaults to 1024. Larger ring sizes result in more granular
     // load distributions. If the number of hosts in the load balancing
     // pool is larger than the ring size, each host will be assigned a
     // single virtual node.
-    uint32 minimum_ring_size = 2;
+    uint32 minimum_ring_size = 1;
```
wora: We should use min_ring_size. min/max are much more widely used in API design than minimum/maximum.

routing/v1alpha2/destination_rule.proto
```diff
@@ -182,7 +182,7 @@ message Subset {
 //       trafficPolicy:
 //         loadBalancer:
 //           consistentHash:
-//             http_header: Cookie
+//             minimum_ring_size: 10
```
rshriram: can you complete this example, since the header has moved to a route rule?
i.e., add something like, 

And the associated routing rule to specify the header to hash on.

```yaml
apiVersion...
kind..
...
hashPolicy:...
```

routing/v1alpha2/route_rule.proto
```diff
@@ -862,6 +866,15 @@ message HTTPFaultInjection {
   }
 }
 
+// HashPolicy is used to specify the hashing policy configuration if the
+// upstream cluster is using a hash load balancer
```
rshriram: Also copy the example from the destination rule, to this section. Otherwise, the page will render without an example, and it won't make any sense for a viewer

routing/v1alpha2/destination_rule.proto
pacevedom: I left this like the envoy's parameter, same name, as it also had that before. I dont really have a strong opinion in favor/against it, so up to you guys. What do you think? @rshriram @ijsnellf 

Reference:
[v2 API](https://www.envoyproxy.io/docs/envoy/latest/api-v2/cds.proto.html?highlight=minimum_ring_size#cluster-ringhashlbconfig)
[v1 API](https://www.envoyproxy.io/docs/envoy/latest/api-v1/cluster_manager/cluster_ring_hash_lb_config.html?highlight=minimum_ring_size)

routing/v1alpha2/destination_rule.proto
pacevedom: Done

routing/v1alpha2/route_rule.proto
pacevedom: Done

geeknoid:
Do we still want this change?

rshriram:
closing in favor of #562

PR #308, opened 2018-01-15
Direct envoy config via crds

rshriram:
Allow uses to overwrite/add/update envoy objects by name (cluster name/ listener name). If we can figure out names per route, we might be able to allow folks to customize things at route level.

This is contingent on us moving to the new Envoy config format (protos) while still using the REST API to supply the configs. [This is not same as gRPC stuff, which requires ADS, etc. for proper use].


kyessenov:
cc @spikecurtis who started a design doc for this.

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,54 @@
...
+    OUTBOUND = 1;
+    INBOUND = 2;
+  }
+
+  message Listener {
+    map <string, string> endpoint_labels = 1;
+    Operation operation = 2;
+    Direction direction = 3;
+    uint32 port = 4;
+    string config = 5;
```
kyessenov: this should protobuf `Any` or `Struct`.
Where are route configs? Those are most complicated.

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  message Cluster {
```
kyessenov: Is this an operation on a cluster or an operation on a collection of clusters?

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,54 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.routing.v1alpha2;
+
+message UnsafeEnvoyConfig {
```
kyessenov: We need a scoping mechanism that is consistent with the rest of istio and amenable to RBAC.

rshriram: Yes, but those can go in with all the config RBAC work that is going on.. I am not going to put any form of RBAC inside routing.

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: single cluster.

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: This is listener config. People can choose to overwrite routes or update the listener with more info. I prefer people to use YAML end to end. with string, its easy
```
config: |
  .. more yaml ..
```

We can convert the YAML into proto or what not and add/merge/replace.

kyessenov:
I have two general comments:
1. Are we OK promoting Envoy as the de-factor Istio proxy? A non-Envoy proxy is unlikely to implement this API, yet this API is core Istio.
2. Are we interested in compositional properties for this API? Do we want combinations of overrides across xDS responses? For example, imagine mixer and fault were defined with this CRD, how do they interact with each other? How do overrides set opaque metadata?

spikecurtis:
Filter ordering matters; I don't see any mechanism for controlling the filter order when updating a listener.

mandarjog:
@rshriram this seems to break layering and istio abstractions.
The general question of arbitrary per route configuration is valuable, but why as `unsafe_PPP_config` ?
Is the plan to have `unsafe_Nginx_config` and so on ?
We could just as easily have opaque per route config that **is** proxy specific.

kyessenov:
@mandarjog @rshriram further on question 1: should we be designing xDS composition API in Envoy data-plane-api instead? What is there from Istio in this particular API?

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,119 @@
...
+//
+message UnsafeEnvoyConfig {
+  enum Operation {
+    INVALID_OP = 0;
+    // Add a listener or cluster.
+    ADD = 1;
+    // Replace an existing listener or cluster.
+    REPLACE = 2;
+    // Update an existing listener or cluster. Any missing fields in the
+    // configuration will be replaced with information computed by Istio.
```
ZackButcher: What about when the listener already has a value? Presumably this config wins and we override the original config with values from here?

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,119 @@
...
+  // Listener/cluster related configuration.
+  message EnvoyConfig {
+    // REQUIRED. Add/Replace/Update operation
+    Operation operation = 1;
+
+    // REQUIRED. Unique name associated with the listener. Istio generated
+    // listeners are named using the format
+    // (http|tcp)_(ip)_(port). Wildcards can be used when selecting
+    // listeners. For example, to select all listeners on port 8080, use
+    // *_0.0.0.0_8080.
```
ZackButcher: We should probably add more info about IP addresses: which IP address do we use in the name? What does the `0.0.0.0` listener mean? etc.

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,119 @@
...
+//           - socketAddress:
+//               protocol: tcp
+//               address: 1.1.1.1
+//               portValue: 5555
+//
+message UnsafeEnvoyConfig {
+  enum Operation {
+    INVALID_OP = 0;
+    // Add a listener or cluster.
+    ADD = 1;
```
ZackButcher: Add operations should have some notion of filter order so we know the correct place to insert the listener, right? (cc @spikecurtis)

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,119 @@
...
+    // out.(serviceName)|(portName/portNumber)|subsetName for outbound
+    // clusters (client side) and in.(portName/portNumber) for inbound
+    // clusters (server side). Wildcards can be used when selecting
+    // clusters. For example, to select all clusters for
+    // foo.default.svc.cluster.local service, use
+    // *.foo.default.svc.cluster.local|*|*
+    string name = 2;
+
+    // Restrict the changes to a subset of pods/VMs in the mesh
+    map <string, string> labels = 3;
```
ZackButcher: Do we need more than labels here? What about using `subset` here instead?

/cc @louiscryan IMO this is another use case for subsets being a top level API element we can refer to in other resources.

ZackButcher:
W.r.t. treating Envoy as the defacto Istio proxy: I don't see a problem with us adding stop-gaps like this for other proxies as well. It's just that Envoy is the proxy we use now, so of course its the first to get this kind of treatment.

@mandarjog: IMO embedding this style config in RouteRules breaks our abstractions in a far worse way (we did debate doing this via RouteRule before rejecting that model). The fact is that Istio will _always_ lag behind Envoy in features. This is especially true for people a Envoy filters for Istio, like [Squash](https://github.com/solo-io/squash). We need to provide an escape valve to allow for custom proxy configuration that Istio does not yet represent.

ZackButcher:
Specifically about RouteRules: the problem is that model does not cleanly map to the configuration we need to feed Envoys: RouteRules generally define routes we expose via RDS, and in a few cases they result in us creating clusters we expose via CDS too. But, RouteRules do not generally correspond to listeners. Modeling listener config in RouteRules means the logic in Pilot to compute Envoy config is a _lot_ more complex, or we wind up with a lot more listeners (too many to be feasible, I believe; @rshriram had some rough math IIRC).

kyessenov:
@ZackButcher it's up to us to worry about backwards- and future-compatibility of configuration like this proposed API. I don't see any design for what to do with this config when envoy changes its API, or when we hypothetically choose another proxy as the main proxy (e.g. using a non-sidecar model, for example). This all has to be taken into account before providing a hacky escape path. 

At the very least, we need RBAC to prevent arbitrary overrides into routing. For that, we need some scoping rules.

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: yep

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: Nope. This has nothing to do with subsets. Its not indicating a version. I could apply this config at client side or server side. To one pod or a whole set of pods across multiple services.

rshriram:
With regard to filter ordering: I knew I was missing something. But my thinking is that all filters we have today are sort of independent of the location (mixer filter can come in the end, so does fault filter). Any filters user specifies can go in the front. I have not come across any case where a filter has to occur after the router filter or after the mixer filter.

With regard to RBAC - lets not conflate RBAC with individual configuration elements. RBAC applies generically to all configuration elements based on type, etc.. It should not be embedded into the configuration (thats the model kubernetes follows).

With regard to portability/tracking envoy design/etc. - The contract with this config is that its unsupported escape hatch. Use at your own peril. And that this will change as we change underlying Envoy. So, no backward compatibility either (i.e. it will be in permanent alpha mode). This allows us to experiment, see how people use in long term, and not worry too much about elegance of implementation. In my guess, there is a handful of people modifying Envoy for special purposes. Such folks are smart enough to track changes in Istio and Envoy. This is not meant for the average operator.

Once you cross that barrier, you can relax lots of assumptions. Expecting normal istio users to learn and use Envoy config beats the purpose of most of the stuff Istio does.

routing/v1alpha2/unsafe_envoy_config.proto
louiscryan: Is 'UPDATE' effectively a protobuf merge where the merge source is the istio generated config.

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,119 @@
...
+//         config:
+//           "@type": type.googleapis.com/envoy.api.v2.Cluster
+//           type: static
+//           hosts:
+//           - socketAddress:
+//               protocol: tcp
+//               address: 1.1.1.1
+//               portValue: 5555
+//
+message UnsafeEnvoyConfig {
```
louiscryan: @ZackButcher 

If we're going to have a merge operation like 'UPDATE' or 'REPLACE' we probably need a 'match' mechanism. Match can help with ordering too if its specific enough. Something like

type :  cluster|listener|...
match : <regex against name of resource of above type to match against>
operation: <enum of:  before | after | replace | merge>
labels: <context match if needed>

routing/v1alpha2/unsafe_envoy_config.proto
louiscryan: See above note. Hopefully subsets make cluster names more predictable so that they can be stably overridden using a match

routing/v1alpha2/unsafe_envoy_config.proto
louiscryan: ... also I think we should match on pre-obfuscated names :) where we can. i.e generated clusters

ZackButcher: Yup, I like it: enum for `type`, then `name` needs to be a full regex (if we did something like support wildcarding parts of the name based on `type` then I'd probably want to split each `type` out to its own message, with a `oneof` here).

By pre-obfuscated, do you mean Istio resource names, or `Destination`and`subset` names, or something else? We need to think carefully about how to specify pre-obfuscated names: I suspect there're cases where it won't be possible to target some specific subset of workloads. I think we'll need a backdoor to target Envoy names directly, so IMO we might as well make that the primary way things work and spend more time making it ergonomic. For example, we could create a tool that ingests a set of Istio configs and spits out all of the listener/cluster/... names that would be generated (even better, it could ingest the `UnsafeEnvoyConfig` too, and mark which resources match and would be modified).

My general thought is that this config is already a break-glass measure, forcing the operator to infer names isn't such a burden.

routing/v1alpha2/unsafe_envoy_config.proto
ZackButcher: Yup; if we use that literal implementation, `repeated` fields are appended together. Is that the behavior we want for lists?

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: We don't have to worry about subsets. We are already generating non obfuscated cluster names (I fixed that a while ago). And with the subsets, our cluster names will be even more clear (serviceName|portName|subsetName). I have it documented in the comments.. I do like the before/after thing - which applies only to listeners. Does not matter for clusters.

I went through the regex motion mentally earlier. I discarded it as it seemed like a pain to manage in the implementation. May be its worth while to revisit.

ZackButcher: After thinking more, I prefer the wildcard name you have now, based on `type`, to a full regex.

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: May be I should make it a wildcard instead of 0.0.0.0. ?

ZackButcher: Yeah, I think wildcard would be better than `0.0.0.0`.

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,119 @@
...
+
+    // Restrict the changes to a subset of pods/VMs in the mesh
+    map <string, string> labels = 3;
+
+    // REQUIRED. Envoy listener/cluster configuration. Refer to
+    // https://github.com/envoyproxy/data-plane-api/ for details.
+    google.protobuf.Any config = 4;
+  }
+
+  repeated EnvoyConfig config = 1;
```
wora: plural please.

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -112,8 +112,8 @@ message UnsafeEnvoyConfig {
 
     // REQUIRED. Envoy listener/cluster configuration. Refer to
     // https://github.com/envoyproxy/data-plane-api/ for details.
-    google.protobuf.Any config = 4;
+    google.protobuf.Any envoy_config = 4;
```
ZackButcher: What about `spec` instead (again mirroring how k8s handles this kind of thing)? Otherwise, IMO, keep it named `config`.

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,121 @@
...
+  // Listener/cluster related configuration.
+  message EnvoyConfig {
+    // REQUIRED. Add/Replace/Update operation
+    Operation operation = 1;
+
+    // REQUIRED. Unique name associated with the listener or cluster. Istio
+    // generated listeners are named using the format
+    // (http|tcp)_(ip)_(port). Wildcards can be used when selecting
+    // listeners. For example, to select all listeners on port 8080, use
+    // *_*_8080.
```
spikecurtis: One of the things we want to do is be able to select all inbound HTTP listeners.  I'm not sure this wildcard match allows me to distinguish between inbound and outbound listeners (although Pilot makes this distinction)

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,121 @@
...
+    // foo.default.svc.cluster.local service, use
+    // *.foo.default.svc.cluster.local|*|*
+    string name = 2;
+
+    // Restrict the changes to a subset of pods/VMs in the mesh
+    map <string, string> labels = 3;
+
+    // REQUIRED. Envoy listener/cluster configuration. Refer to
+    // https://github.com/envoyproxy/data-plane-api/ for details.
+    google.protobuf.Any envoy_config = 4;
```
spikecurtis: This mechanism makes it awkward to insert HTTP filters, which are nested inside the HTTP Connection Manager. In this case I'm performing different operations at different layers of the nested config.  I want to UPDATE the listener, UPDATE the HTTP Connection Manager filter, and ADD additional HTTP filters.

routing/v1alpha2/unsafe_envoy_config.proto
ZackButcher: We make that distinction with clusters (in/out prefix), but not with listeners. However, when building listeners we do know whether its in or out, so we could include a prefix on the name and support wildcarding it here if there's a need. This is probably a decent reason to include it.

routing/v1alpha2/unsafe_envoy_config.proto
ZackButcher: FWIW this is why I originally proposed a more general model based on PATCH which uses a field mask to update arbitrary values in the structure.

louiscryan: Can that be modelled as three different operations even though they represent one unit of work ?

We can make our selection/match language a bit more like xpath if we need to accomodate greater precision than just type+regex. After all you can do proto -> JSON -- [json-path] --> JSON --> proto which is pretty much what we're doing here.  We don't need to implement all of that to have something useful and proto reflection makes it easy enough to implement.

Thinking about it some more maybe we can steal from
https://github.com/google/cel-spec

spikecurtis: It's really just one operation --- the UPDATE semantics just means find this existing thing and modify it, the nested operations are specifying *how* to modify.

A path expression would work if we want to allow arbitrary changes.

kyessenov: The _real_ complicated case is adding LISTENER filters for some ROUTES and adding some opaque metadata for per ROUTE that is dependent on ROUTE properties. That is what mixerfilter does.

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: I had the direction field in this earlier. But after looking at our listener configs, it made no sense. Because we have a single listener shared between inbound and outbound. We don't have two. Especially, when you consider something like listener on port 80, you are bound to see a bunch of outbound and inbound services under same listener.

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: fwiw, adding a filter is still an update operation as far as the listener is concerned.. So, its really "updating" the listener. That is one of the reasons, I didn't make a distinction of http connection manager vs tcp proxy. You are updating a listener, thats it. And I don't think its as complicated as it looks. Take a look at the example: https://github.com/istio/api/pull/308/files#diff-81b14d829a1cf05cfe53bf56f1710627R49 

Secondly, keep in mind the base assumptions about this API. No UX, no thought given to perfect api spec or backward compatibility, or any such thing. The moment the operator descends down into the data plane, we expose the entire data plane in all its glory/goriness. It keeps our implementation footprint small, malleable, and most important of all: "confined" to a single stage/file in the config generation.

routing/v1alpha2/unsafe_envoy_config.proto
kyessenov: I don't think we share listeners between inbound and outbound directions. There are three types of listeners:
1. inbound for local IP:PORT;
2. outbound for each remote TCP service;
3. outbound for each remote HTTP port, further dispatched by RDS.

kyessenov: I don't think we should use internal listener names here. That's too much implementation exposure. We can reference listeners/routes by service-ports, or inbound listeners by ports.

routing/v1alpha2/unsafe_envoy_config.proto
kyessenov: There's a difference between client-side and server-side that should not be conflated. Pods that are labelled by a _superset_ acquire the config for inbound, but only clusters that are _exactly_ matched by labels acquire the config for outbound.

I don't think revealing subsets as cluster names is beneficial. In the longer term, we want to avoid creating too many clusters, and the cluster names are not guaranteed to be stable.

routing/v1alpha2/unsafe_envoy_config.proto
kyessenov: The strategic patch over structured JSONs is very effective in `kubectl apply`, but we would need to annotate data-plane-api with the "merge strategy" bits (e.g. how to merge two lists or two maps).

routing/v1alpha2/unsafe_envoy_config.proto
rshriram: how do you differentiate service-port and inbound listener port? I think if we create names for listeners, we can refer to them from these configs.

kyessenov:
It's the same problem with listener names. If we expose them, then we
commit to using the same listener split (which is not the only way to
structure the config). If we later change names, then the config no longer
works, and it would be better to apply code changes to pilot

routing/v1alpha2/unsafe_envoy_config.proto
spikecurtis: @rshriram So an UPDATE on a listener always merges the given config with the existing config on the listener? What if you wanted to replace rather than merge?

How would we distinguish between adding an HTTP filter vs adding a Network filter? We can't just assume that HTTP listeners will always get HTTP filters, operators might want to add network filters.

ZackButcher: Updates _must_ be full replacements with the current API: we can't merge the body of a listener with something provided here because we have no indication of which parts of the document to modify. If we had a field mask then we could talk about merge/replace of individual fields in the listener.

rshriram: Will try to clarify more soon. Btw, you cannot add network filters to http listeners. Or vice versa.

My intention was to provide update, replace and add operations. Please take a look at the example in the pr (above the proto) where I add a lua filter without touching other aspects of http conn manager. That is an example of update.

rshriram:
I updated PR to add directionality for listeners. This should help differentiating inbound/outbound listeners.. (@spikecurtis )

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,135 @@
...
+// service, for all inbound calls to the reviews HTTP service on port 8080.
+//
+//     apiVersion: config.istio.io/v1alpha2
+//     kind: UnsafeEnvoyConfig
+//     metadata:
+//       name: my-lua-script
+//     spec:
+//       configs:
+//       - labels:
+//           name: reviews # selects the pods where this config will be applied
```
costinm: Can we avoid another label selection/clustering issue  ? 
For example, can we attach the custom rule to a gateway or route rule ?

costinm: Also, is this only for listener - and how to specify client side vs server side ?

routing/v1alpha2/unsafe_envoy_config.proto
```diff
@@ -0,0 +1,135 @@
...
+//         operation: update # update existing listener
```
costinm: What does it mean - replace I understand, but update is a bit confusing.

ZackButcher:
@costinm FYI we met and discussed this PR earlier today. Here are the notes (courtesy of @spikecurtis):

> We reached the following conclusions:
>  * A template-based solution is appealing in the long term, but we want to focus on what can work for a 1.0 time frame.
>  * We will immediately work on a webhook based solution, the minimum viable solution will look like this:
>     * The webhook will operate as a REST endpoint with JSON payload of the xDS config.
>     * Pilot will call the webhook with the full xDS config immediately before it sends it to Envoy. The webhook may modify as needed and return the config.
>     * Only one webhook is supported
>     * The webhook is unencrypted and unauthenticated. It is expected that operators will distribute their webhooks as a sidecar to Pilot, running in the same trust zone (i.e. pod)
>     * The webhook is configured via a command-line flag to Pilot
>  * Future increments to this solution may add
>     * Ability to chain multiple webhooks
>     * mTLS connection so the webhook can be securely reached over the network
>     * Ability to scope webhooks to be called for particular namespaces, services, etc.
>     * Calls to the webhook with finer granularity than the entire xDS config blob.
>  * Separately, we will consider adding "first-class" configuration for Lua filters, External Auth filter, or other widely used Envoy built-in filters.
> 
> Action Items:
>  * Shriram will author a new PR for the "minimum viable" webhook solution and ping the group.

costinm:
I'm still a bit confused of the webhook relation with this PR - it seems
this particular CRD is not used by the webhook, so I assume
this PR is obsolete now ?

rshriram:
Yes, the PR is obsolete. ITs here for posterity. I will close it. The end result of the discussion was that any other mechanism we come up with at this moment was going to involve a lot more work/design, while the webhook one could be a short term stop gap solution that people can use to customize pilot without having to fork it.

costinm:
I think webhook can also become a longer term solution - in some form.

For the record, using template-based generation for the config - as was
started with the injector webhook,
and I'm planning to do for bootstrap config - is another option. Less
flexible compared with webhook,
but may be easier to deploy and work without extra binaries.

kyessenov:
We discussed that. I have a prototype that replaces the bulk of pilot with
one template, but it's too disruptive is out of scope for now.

costinm:
I'm not talking about replacing anything in current pilot - but the new
code for LDSv2 REST can
use a template, and won't disrupt anything. Same for the new bootstrap.yaml
generation - new
code, no risk of breaking existing stuff and likely the cleanest/easiest
way to implement it.

PR #309, opened 2018-01-17
Destination Rules: ports, (no)inheritance, and more

rshriram:
cc @louiscryan / @ZackButcher / @kyessenov / @frankbu 

The PR currently has docs that describe how the destination rules would look. Once we converge on this, I can update the protos.

Some pending issues
================

TLS upgrade clunkiness
------------------------

Contrary to what I thought earlier, this does not solve the clunky declaration in Foreign services where the user has to indicate that traffic arriving on google.com:80 on the sidecar has to be sent to upstream google.com:443 over HTTPS, i.e. the current proposed foreign services syntax is

```yaml
hosts:
- *.google.com
portMaps: # this is the proposal
- inPort: 80
   outPort: 443
   tlsUpgrade: true
   discovery: none #indicates pass through
```

It is possible to achieve the behavior above using the following set of rules (for non wildcard services) using a simple foreign services declaration of the form

```yaml
kind: ForeignService
hosts:
- simpleapi.google.com #can't do wildcards if we want port remapping
ports:
- number: 80
   protocol: http
- number: 443
   protocol: https
```

and a route rule to reroute traffic from 80 to 443

```yaml
kind: RouteRule
hosts:
- simpleapi.google.com
http:
- match:
   - port:
         number: 80
   route:
   - destination:
          name: simpleapi.google.com
          port: 443
```

followed by a destination rule:

```yaml
kind: DestinationRule
name: simpleapi.google.com
http:
- port: 443
   tls:
      mode: simple
```

Ignoring some implementation issues for the moment, this experience seems awfully bad. 50% of folks are going to use HTTPS directly. 50% of folks will want to use HTTP, to get metrics and want us to originate TLS. The second half will have to create 3 rules for every external service that they access. Now, if this number was like 10% of people, we can live with this added verbosity as they fall in the minority. But given that this upgrade scenario is pretty common, I feel that the API should accommodate for this common use case and provide a simpler configuration.

This issue is present even internally, when we want to do incremental migration from port X to TLS enabled port Y. I would have to setup a route rule and a destination rule to handle this. Given that the migration is likely to happen over a long time (weeks), a route rule seems like a risky place, because I could be doing other app development stuff in the meantime. Such development would likely result in more route rules (new versions, etc.). It is easy for an end user to mess up the migration, if they create a new route rule that misses the from xxx to yyy transition that we ask users to do.

Egress gateway usage
----------------------

IF someone wants to use a dedicated Egress gateway, we have to augment the foreign service spec to specify the gateway where the service should be made visible. Imagine something like this:

```yaml
kind: ForeignService
hosts:
- someapi.google.com #wildcards wont work for now. Different discussion
gateways:
- foogateway # Like route rules, "mesh" is default gateway
discovery: dns
port:
- number: 80 # from sidecar to gateway, traffic arrives over TLS using istio-auth
   protocol: http
- number: 443 # what will this do? this is where I think the inPort/outPort helps
   protocol: https
```

And then a gateway spec:

```yaml
kind: Gateway
name: foogateway
services:
- port:
       number: 80
   hosts:
   - someapi.google.com
```

and a route rule to reroute traffic to the gateway as well as for gateway to send outside mesh

```yaml
kind: RouteRule
hosts:
- someapi.google.com
gateways:
- mesh
- foogateway
http:
- match: # from mesh to foogateway
   - port:
          number: 80
   - gateways:
      - mesh
   route:
   - destination:
          gateway: foogateway # this is new
 - match: # from foogateway to external world
    - port:
           number: 80
     - gateways:
        - foogateway
    route:
    - destination:
            name: someapi.google.com
            port:
                 number: 443
```

and finally a destination rule that does the TLS upgrade

```yaml
kind: DestinationRule
name: someapi.google.com
http:
- port: 443
   tls:
      mode: simple
```

I feel like there is something missing from this flow, and its not immediately intuitive from an end user perspective.

### DIFF
--- a/routing/v1alpha2/destination_rule.proto
+++ b/routing/v1alpha2/destination_rule.proto
@@ -91,7 +91,7 @@ package istio.routing.v1alpha2;
 //         trafficPolicy:
 //           loadBalancer:
 //             simple: LEAST_CONN
-//       - port: 7777
+//       - port: http-monitoring #can use names or numbers
 //         trafficPolicy:
 //           loadBalancer:
 //             simple: ROUND_ROBIN
--- a/routing/v1alpha2/destination_rule.proto
+++ b/routing/v1alpha2/destination_rule.proto
@@ -34,7 +34,7 @@ package istio.routing.v1alpha2;
 //         loadBalancer:
 //           simple: LEAST_CONN
 //
-// Version specific DestinationRule can be specified by defining a named
+// Version specific destination rule can be specified by defining a named
 // subset and overriding the settings specified at the service level. The
 // following rule uses a round robin load balancing policy for all traffic
 // going to a subset named testversion that is composed of endpoints (e.g.,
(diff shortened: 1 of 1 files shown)


nmnellis:
so the issue is that if you add an envoy sidecar to your app, all https traffic will need these 3 things right? because you wont be able to call the services directly over ssl with a sidecar proxy.
Right now in Istio i just have to create an Egress Rule and http://stuff:443 which isnt great but a lot easier than what you propose. Granted all my apps have to accommodate the upgrade by coding their apps to call the funky url rather than a configuration driven virtual gateway.

nmnellis:
After thinking about this logically it definitely makes sense to have the ForeignService declaration as it acts like a firewall in a sense. Secondly a RouteRule to upgrade the connection also makes logical sense if your application is no longer handling it. I do not however think DestinationRule is hard to conceptualize until you need it (example mutual auth). If it could somehow be assumed simple tls that seems reasonable to me.

louiscryan:
Bunch'o'notes...

I think we can support wildcards with port remapping, we just can't do it if 'resolution: dns' and there are no endpoints specified.

Re verbosity I think its somewhat overstated. We need ExternalService no matter what but if we have wildcarding then it doesn't have to be one-per-service. Given that the port is declared as HTTPS in the ExternalService we don't need the DestRule declaration to set any TLS properties unless we need to do something other than 'simple' which is likely to be quite rare. Yes, folks do have to write RouteRules but they can be wildcarded too which keeps verbosity down. The main thing Pilot needs to do is to be able to create two cluster representations for the HTTPS backend, one for TCP passthrough when client originates TLS and the other for when Envoy is originating. The reference in the route rule from the 'http' block lets us know to do this.

I also like RouteRule for this is as it allows us to control the scope of this translation. Its also worth pointing out that the rollout is more likely to be converting applications away from using HTTPS to HTTP so it's essentially the applications opting into the rule which means the rule can be pretty much global from the start as both protocols are supported at the same time. Between that and wildcarding I think that makes it relatively safe to express the migration. (Would be good to add this recipes)


Egress Gateway Usage ---------
I don't think we can have ExternalService refer to a gateway. We use a RouteRule to declare this binding just like we use RouteRule to declare a similar binding for Ingress. We do need two RouteRules one to send traffic to the gateway from the sidecar and the other to send traffic to the ExternalService from the gateway. One reason why we have to do this is that we need to support having two different Gateways for egress to the same ExternalService. If we declare one or more Gateways in foreign service we still wouldn't have a mechanism to choose which one to use.

I think your RouteRule example here is right but just as above I think the DestRule can be omitted as @nmnellis suggestes

PR #310, opened 2018-01-18
Simplify Makefile

douglas-reid:
This PR is an attempt to simplify and cleanup the Makefile in response to feedback re: functionality and readability. It sets about this in the following way:

- removes downloading of googleapis protos, preferring the copied source in istio/gogo-genproto
- uses `find` shell script to get dynamic list of protos per protoc invocation, avoiding listing of each new file
- moves to lowercase for vars
- reorgs sections to consolidate all related vars and rules
- adds a clean up target to remove generated files

Additionally, to simplify protoc output path, this PR adds explicit `go_package` options to each proto. This should be of no impact on istio/istio, as all references that I could find in a quick search seem to have explicit import aliasing.

Reviewers: as before, I want to emphasize that I'm make-illiterate, so if there are better ways to do all of this, please suggest.


Makefile
```diff
@@ -1,200 +1,82 @@
...
+ifeq ($(protoc_bin),)
+	protoc_bin = must-rebuild
 endif
 
 # Figure out which machine we're running on.
 UNAME := $(shell uname)
 
 # TODO add instructions for other operating systems here
-$(PROTOC_BIN):
+$(protoc_bin):
```
kyessenov: nit: I don't think we should be installing stuff here, but provide a docker image with everything pre-installed

PR #311, opened 2018-01-18
Bind gateway to a workload using labels

rshriram:
Currently, there is no way to bind a gateway to a workload. One option is to start envoy with the gateway name as a parameter. But thats too static, and does not allow multiple gateways to be bound to a single envoy. Instead, this PR proposes to add a set of workload specifiers to the gateway, that will let the user specify the set of pods/vms where this gateway config should be applied.

@louiscryan related to the discussion today, I feel uneasy about having yet another resource to bind the gateway and the workload. Teh gateway to workload binding is probably better off living in the same resource. Its the most common use case for people migrating from ingress to Gateways. Its easy for people to simply start up an envoy pod in nodeport or load balancer mode, and then apply the gateway to it.


routing/v1alpha2/gateway.proto
```diff
@@ -138,6 +141,10 @@ option go_package = "istio.io/api/routing/v1alpha2";
 message Gateway {
   // REQUIRED: A list of server specifications.
   repeated Server servers = 1;
+
+  // REQUIRED: One or more labels that indicate a specific set of pods/VMs
+  // on which this gateway configuration should be applied.
+  map<string, string> workloadLabels = 2;
```
ZackButcher: Can we call this `selector`, mirroring k8s service spec?

routing/v1alpha2/gateway.proto
```diff
@@ -144,7 +144,7 @@ message Gateway {
 
   // REQUIRED: One or more labels that indicate a specific set of pods/VMs
   // on which this gateway configuration should be applied.
-  map<string, string> workloadLabels = 2;
+  map<string, string> selectors = 2;
```
ZackButcher: In k8s it's singular, `selector`. I don't have a strong preference w.r.t. plurality.

routing/v1alpha2/gateway.proto
```diff
@@ -138,6 +141,10 @@ option go_package = "istio.io/api/routing/v1alpha2";
 message Gateway {
   // REQUIRED: A list of server specifications.
   repeated Server servers = 1;
+
+  // REQUIRED: One or more labels that indicate a specific set of pods/VMs
```
kyessenov: I thought we wanted flexibility in how gateway is materialized. It should be possible to run gateway as a managed service out of the mesh?

rshriram: Yeah, but there has to be something to bind the gateway to the workload right? Pilot has to know which pod/VM/container is eligible to receive the gateway config.

frankbu: Shouldn't this work more like ingress_controllers, i.e., a gateway controller/workload will implement all Gateway's that are not configured to specify some particular controller should be used? This way, by default, we could have one gateway controller (similar to istio-ingress-controller today), running in the control plane, that handles ingress by default (the common case). Then if we need an egress Gateway, for example, we can start another gateway controller service ourselves (in the data plane) and create a Gateway that specifies it should be used, instead of the default one.

I also wonder if instead of using pod labels to select a gateway controller, it would make more sense to specify a service. If we are going to create a route rule to send traffic to an egress gateway, the gateway controller needs to have a corresponding service that we use for the routerule destination, right?

rshriram: I am skeptical about the ingress controller model because it breaks uniformity. We don't have such a thing in Istio. We have traditionally stuck to the simple direct association model, instead of a more loose coupling where arbitrary number of controllers could implement the same gateways. This then leads to all sorts of trouble, like merging gateways, and such. Having a simple one-deployment, one-gateway model is easier to reason about, in exchange for flexibility. 

I am open to the idea of using a service. But keep in mind that we have almost moved away from the concept of a service name. its actually a host name. @ZackButcher thoughts?

rshriram:
@frankbu / @kyessenov in the interest of unblocking the gateway transition work, I am merging this PR and creating an issue to discuss this. Binding from the api or using another resource to bind - the behavior is going to be the same in the common case: gateway gets bound to a workload. So, lets merge this, while we iterate on the controller or no controller options in an issue.

routing/v1alpha2/gateway.proto
louiscryan: I think I agree with @frankbu here. In the simple case folks shouldn't have to 'select' the controller for the API to work. We've already assessed that gateway merging is manageable for a controller (per @ZackButcher)

Wouldn't a better example be 'app: envoy-controller' ?

PR #314, opened 2018-01-19
Fix default tls mode for destination and add 80-443 mapping for ext svc

rshriram:


routing/v1alpha2/destination_rule.proto
```diff
@@ -393,19 +393,29 @@ message OutlierDetection {
 message TLSSettings {
   // TLS connection mode
   enum TLSmode {
-    // If set to "disable", the proxy will use not setup a TLS connection to the
-    // upstream server.
-    DISABLE = 0;
-
+    // Placeholder
+    TLSMODE_INVALID = 0;
```
ZackButcher: I think we should keep `DISABLE` the default (or change it to `PASSTHROUGH`). Either way, IMO our default behavior should be to not touch the connection the app set up.

routing/v1alpha2/destination_rule.proto
```diff
@@ -393,19 +393,29 @@ message OutlierDetection {
...
-    MUTUAL = 2;
-  };
+    // authentication. The certificates should be specified using the
+    // configuration fields (clientCertificate, privateKey).
+    MUTUAL = 3;
+
+    // If set to "disable", the proxy will use not setup a TLS connection
+    // to the upstream server.
+    DISABLE = 4;
+};
```
ZackButcher: nit: wrong indentation here

routing/v1alpha2/destination_rule.proto
ZackButcher: There's also no reason to add an invalid value to our enum when there's a reasonable default value we can use instead.

routing/v1alpha2/external_service.proto
```diff
@@ -54,6 +54,33 @@ option go_package = "istio.io/api/routing/v1alpha2";
 //       tls:
 //         mode: simple # initiates HTTPS when talking to example.com
 //
+// To reduce the configuration burden for the common case where most
+// external services are accessed over port 443, there is an implicit
+// internal mapping from port 80 to port 443. Such that an external service
```
ZackButcher: ~~Such that~~ **A**n external service declared without any port is assumed...

routing/v1alpha2/destination_rule.proto
rshriram: Right, but that is covered right? I can omit TLSmode completely, in which case, the semantics is keep things as is.. if I specify a TLS mode, I have to pick something or other (certificates, modes, etc..). Disable implies do not setup a TLS context in the upstream cluster.. I felt that it was a bit dangerous to leave as default.

routing/v1alpha2/external_service.proto
```diff
@@ -54,6 +54,33 @@ option go_package = "istio.io/api/routing/v1alpha2";
...
+//       hosts:
+//       - foo.com
+//       - bar.com
+//       - baz.com
+//       - scooby.com
+//       discovery: dns
+//
+// The application is expected to access these services over port 80 using
+// plain text http (i.e. http://foo.com, http://scooby.com, etc.). The
+// sidecar would originate TLS connections to the external service over
```
ZackButcher: The sidecar **will** originate TLS connections to the external service **on** port 443

routing/v1alpha2/external_service.proto
```diff
@@ -54,6 +54,33 @@ option go_package = "istio.io/api/routing/v1alpha2";
...
+// port 443 (i.e. https://foo.com, https://bar.com). Note that currently,
```
ZackButcher: Nit: no comma after currently, I think it reads fine w/o it:

Note that currently this special case functionality is only available for non-wildcard domains.

routing/v1alpha2/external_service.proto
```diff
@@ -54,6 +54,33 @@ option go_package = "istio.io/api/routing/v1alpha2";
...
+// this special case functionality is available only for non-wildcard
```
ZackButcher: "only available" instead of "available only" reads better IMO.

routing/v1alpha2/destination_rule.proto
```diff
@@ -393,19 +393,29 @@ message OutlierDetection {
...
-    // upstream server.
-    DISABLE = 0;
-
+    // Placeholder
+    TLSMODE_INVALID = 0;
+
+    // If set to "istio", the proxy will set up mutual TLS authentication
+    // using Istio CA issued certificates. This is the default when Istio
+    // mTLS authentication is enabled mesh-wide.
+    ISTIO = 1;
```
frankbu: Do we need an explicit value for this? Could we instead use `MUTUAL` and say that this is what happens if `clientCertificate` and `privateKey` are not set?

routing/v1alpha2/external_service.proto
```diff
@@ -174,7 +201,11 @@ message ExternalService {
   // the specified destination endpoint IP/host.
   repeated string hosts = 1;
 
-  // REQUIRED. The ports associated with the external service.
+  // The ports associated with the external service. If omitted, service
```
ZackButcher: The ports associated with the external service. If no ports are provided, the external service is assumed to be reachable over HTTPS on port 443. Applications calling the external service over HTTP on port 80 will be automatically forwarded to the external destination over HTTPS  on port 443 by the sidecar.

rshriram:
@ijsnellf feels that this is going to lead to trouble. His point of view is that the following is not so terrible after all.

```yaml
kind: ExternalService
hosts:
- simpleapi.google.com #can't do wildcards if we want port remapping
ports:
- number: 80
   protocol: http
- number: 443
   protocol: https
---
kind: RouteRule
hosts:
- simpleapi.google.com
http:
- match:
   - port:
         number: 80
   route:
   - destination:
          name: simpleapi.google.com
          port: 443
---
kind: DestinationRule
name: simpleapi.google.com
http:
- port: 443
   tls:
      mode: simple
---
```

If we get the ability to rewrite ports for wildcards as well, we can avoid repetition of rules, by lumping lots of domains (e.g., *.watson.net) under a single combo of ext service, route rule, dest rule as shown above.

routing/v1alpha2/destination_rule.proto
ZackButcher: Actually, to that point: is `DISABLED` even needed in that case? If you're providing certs, you probably don't want TLS disabled. If you're not providing certs, there's no reason for a TLSOptions config at all, right?

rshriram: Yeah true. I was looking at it from a disable istio_mtls thing.. So that we have the ability to control the mTLS setting on a service by service basis. But may be this should not get into route rule in the first place. Should be kept strictly separate and let admins tackle that.. From that point, i think even specifying Istio as an mTLS option here seems like conflating auth and routing (I don't want to inherit RBAC related issues inside route rule semantics)..

frankbu: If Istio is turning on mTLS by default with it's own provided certificates, isn't `mode: DISABLED` needed so the user can turn it off for a specific destination? Just omitting the TLSOptions will leave it as is. I don't see, however, why `ISTIO` needs to be called out as a specific option. It seems to me that it's the same as `MUTUAL`, just without user provided certificate. Am I missing something?

routing/v1alpha2/destination_rule.proto
```diff
@@ -393,19 +393,29 @@ message OutlierDetection {
 message TLSSettings {
```
costinm: One small issue: if destination is *.foo.com, I'm not sure how SNI and SAN will be specified... Maybe indicate in the doc that if SNI or SAN is needed, you need individual hosts, no wildcard ?  

Also, once Piotr has the SNI sniffing - we may want to add mode=TRANSPARENT, in which case the client will initiate
a https connection ( all other modes assume client connects to http://foo.com:443 - I assume ? ).

routing/v1alpha2/external_service.proto
```diff
@@ -37,7 +37,7 @@ option go_package = "istio.io/api/routing/v1alpha2";
 //       hosts:
 //       - example.com
 //       ports:
-//       - number: 443
+//       - number: 9443
```
costinm: While 9443 is better then 443 - it may be better to avoid the 443 completely, still red flags... 9080 ?

routing/v1alpha2/external_service.proto
costinm: Unless we set it otherwise ? Not all https servers are on 443.

routing/v1alpha2/destination_rule.proto
```diff
@@ -393,19 +393,29 @@ message OutlierDetection {
...
     // If set to "mutual", the proxy will secure connections to the
     // upstream using mutual TLS by presenting client certificates for
-    // authentication.
-    MUTUAL = 2;
-  };
+    // authentication. The certificates should be specified using the
+    // configuration fields (clientCertificate, privateKey).
+    MUTUAL = 3;
+
+    // If set to "disable", the proxy will use not setup a TLS connection
```
louiscryan: doc is unclear. I tend to agree with Zack on PASSTHROUGH

routing/v1alpha2/destination_rule.proto
louiscryan: @wattli @wattli 

I like the idea of using the enum to distinguish between 'mutual using local mesh identity' and 'mutual using specific key'. At some point we'll likely end up having to support 'mutual with specific mesh identity' when our identity management system allows for identity delegation.

We can just have the enum be "MUTUAL" but then we have to push the other options down into a one-of.

oneof {
    ClientCert { client cert & provate key}
    Identity "self" | <other user identity that can be asserted>
}

routing/v1alpha2/destination_rule.proto
ZackButcher: This is covered by the SAN and SNI fields on this object: https://github.com/istio/api/blob/911b3fcf8a944ee30c470a155d743d924edb14f6/routing/v1alpha2/destination_rule.proto#L437-L443.

We should update the comment on the SNI field to make clear that it must be a a subset of the domains in the destination.

routing/v1alpha2/destination_rule.proto
rshriram: You are right in that ISTIO implies mTLS. But I was trying to avoid this implicit inference that if its mTLS it will be istio, unless you specify your own certs. (for what its worth, I could turn off istio auth completely). In that case, having a distinct ISTIO_MTLS is more clearer than simply mTLS. But that may be me reading too much into this. I am open to any option here.

routing/v1alpha2/destination_rule.proto
wattli: Sorry just saw this message, right, current setup is more extendable, but maybe we can give a more meaningful name. What about

MUTUAL_TLS_ISTIO_CERT
TLS
MUTUAL_TLS_CLIENT_CERT
...

PR #318, opened 2018-01-23
Add config field for stats update interval.

JimmyCYJ:
Add config field for stats update interval, so that we can control the frequency to update check, quota and report stats at proxy such as Envoy.

@qiwzhang


mixer/v1/config/client/client_config.proto
```diff
@@ -70,6 +70,10 @@ message TransportConfig {
   }
   // Specifies the policy when failed to connect to Mixer server.
   NetworkFailPolicy network_fail_policy = 4;
+
+  // Specifies interval in milliseconds at which stats maintained by proxy for
+  // check, quota and report calls get updated.
```
qiwzhang: Please add
// if not specified, or 0 protobuf default value, the interval is 10 seconds

mixer/v1/config/client/client_config.proto
```diff
@@ -70,6 +70,10 @@ message TransportConfig {
   }
   // Specifies the policy when failed to connect to Mixer server.
   NetworkFailPolicy network_fail_policy = 4;
+
+  // Specifies interval in milliseconds at which stats maintained by proxy for
```
qiwzhang: Specify fresh interval in milliseconds to write mixer client statistics to Envoy share memory.

JimmyCYJ:
Please take a look. Thanks!

mixer/v1/config/client/client_config.proto
```diff
@@ -70,6 +70,11 @@ message TransportConfig {
   }
   // Specifies the policy when failed to connect to Mixer server.
   NetworkFailPolicy network_fail_policy = 4;
+
+  // Specify fresh interval in milliseconds to write mixer client statistics to
```
ayj: s/fresh/refresh

PR #322, opened 2018-01-26
Update for docs

geeknoid:


geeknoid:
I don't understand the cause of the test failure. It's complaining about not finding goimports. I don't believe I changed anything that could affect whether or not goimports is found.

Any ideas?

PR #326, opened 2018-01-29
Doc fixes.

geeknoid:


mixer/v1/template/istio.mixer.v1.template.pb.html
```diff
@@ -4,6 +4,9 @@
 location: https://istio.io/docs/reference/config/mixer/istio.mixer.v1.template.html
 layout: protoc-gen-docs
 ---
+<p>This proto describes the types that can be used inside Mixer templates. These message types are used to specify
```
douglas-reid: > These types express the ValueType equivalent for each of these message types for expression mapping.

PR #328, opened 2018-01-29
Add configure field for tcp filter report interval.

JimmyCYJ:
This is for solving issue# [539](https://github.com/istio/proxy/issues/539)


mixer/v1/config/client/client_config.proto
```diff
@@ -124,4 +124,8 @@ message TcpClientConfig {
   // Quota specifications to generate quota requirements.
   // It applies on the new TCP connections.
   QuotaSpec connection_quota_spec = 5;
+
+  // Specify report interval in milliseconds to send report for TCP filter.
```
qiwzhang: to send periodical reports for long TCP connections.

JimmyCYJ: Comment is updated. Thanks!

mixer/v1/config/client/client_config.proto
```diff
@@ -124,4 +124,9 @@ message TcpClientConfig {
   // Quota specifications to generate quota requirements.
   // It applies on the new TCP connections.
   QuotaSpec connection_quota_spec = 5;
+
+  // Specify report interval in milliseconds to send periodical reports for
+  // long TCP connections. If not specified, or 0 protobuf default value,
+  // the interval is 10 seconds.
+  int32 report_interval_ms = 6;
```
wora: Please use google.protobuf.Duration. Please always use Timestamp and Duration unless it is performance critical code.

JimmyCYJ: Thanks for the advice. I have a PR that updates this field in https://github.com/istio/api/pull/330

PR #334, opened 2018-01-31
Update broker API package. 

xiaolanz:
Update into broker.dev and follow api style guidelines.


rshriram:
this is in @geeknoid 's territory

xiaolanz:
PTAL. I think all these message should be included in the doc.

geeknoid:
The point of $hide_docs is to hide them from our current doc set. These
features are not ready for anybody to use and showing them in our reference
docs on istio.io is misleading for our customers.

When the feature is ready for use, then $hide_from_docs can be removed at
that point.​

andraxylia:
@xiaolanz This change is not backward compatible and it breaks istio/istio when we try to pick up the latest version of the api. Please fix istio/istio or revert this PR.

xiaolanz:
yes. I'm aware of that.  i'm working on fixing istio/istio. What's the current process of picking up new api versions?

andraxylia:
@xiaolanz you have to change revision number for istio/api in Gopkg.toml and run dep ensure -update.
You have to replace the imported pkg for broker eveywhere in istio and make sure it builds.

PR #338, opened 2018-01-31
add TCP and MONGO to the list of protocols supported by Egress Rules

vadimeisenbergibm:


vadimeisenbergibm:
@ZackButcher yes, already starting from 0.3.0.
> Pilot respects the new list of protocols already, right?

PR #343, opened 2018-02-01
Update STYLE-GUIDE.md

wora:
Mention the CRD kind must match proto message.

### DIFF
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -41,7 +41,8 @@ When defining Kubernetes Custom Resource Definition (CRD) using
 `proto3`, follow the following guidelines:
 * The proto `package` name must match the Kubernetes `apiVersion`,
   excluding the `.io` DNS suffix and reversing the DNS segment
-  ordering.
+  ordering. The Kubernetes `apiVersion` has the format of
+  `group/version`.
 * The proto message type must match the CRD `kind` name.
 
 #### Example
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -38,9 +38,13 @@ naming convention:
 ## Versioning
 
 When defining Kubernetes Custom Resource Definition (CRD) using
-proto3, the proto versioning must match the Kubernetes versioning,
-and the proto message type must match the CRD kind name.
-see the following example.
+`proto3`, follow the following guidelines:
+* The proto `package` name must match the Kubernetes `apiVersion`,
+  excluding the `.io` DNS suffix and reversing the DNS segment
+  ordering.
+* The proto message type must match the CRD `kind` name.
+
+#### Example
 
 The Kubernetes CRD:
 
--- a/STYLE-GUIDE.md
+++ b/STYLE-GUIDE.md
@@ -38,7 +38,8 @@ naming convention:
 ## Versioning
 
 When defining Kubernetes Custom Resource Definition (CRD) using
-proto3, the proto versioning should match the Kubernetes versioning,
+proto3, the proto versioning must match the Kubernetes versioning,
+and the proto message type must match the CRD kind name.
 see the following example.
 
 The Kubernetes CRD:


STYLE-GUIDE.md
```diff
@@ -38,7 +38,8 @@ naming convention:
 ## Versioning
 
 When defining Kubernetes Custom Resource Definition (CRD) using
-proto3, the proto versioning should match the Kubernetes versioning,
+proto3, the proto versioning must match the Kubernetes versioning,
+and the proto message type must match the CRD kind name.
```
ayj: Should group names and version also be consistent with CRDs? We've defined versioned package in istio/api but these all map to the same CRD group and version (config.istio.io). 

For example, does the following mapping make sense?

```golang
group   = routing.api.istio.io
version = v1alpha2
kind    = RouteRule

group   = rbac.api.istio.io
version = v1alpha1
kind    = ServiceRoleBinding
```

xiaolanz: @ayj  this is exactly what I am proposing: https://groups.google.com/d/msg/istio-config/CzH6nRBT_o0/3lvIBao_AgAJ

I like this to be consistent with api group too.

wora: Yes. I should have made that more clear. The CRD apiVersion must match proto package name, minus the `.io` suffix, and the ordering. I will update the text.

xiaolanz: These are apiGroup. not apiVersion.

STYLE-GUIDE.md
```diff
@@ -38,8 +38,13 @@ naming convention:
 ## Versioning
 
 When defining Kubernetes Custom Resource Definition (CRD) using
-proto3, the proto versioning should match the Kubernetes versioning,
-see the following example.
+`proto3`, follow the following guidelines:
+* The proto `package` name must match the Kubernetes `apiVersion`,
+  excluding the `.io` DNS suffix and reversing the DNS segment
+  ordering.
+* The proto message type must match the CRD `kind` name.
```
ayj: I'm not sure if its worth highlighting here, but the `apiVersion` as shown in the YAML snippet is the  concatenation of the Kubernetes API _group_ and _version_. In the example below, `config.istio.io` is the group and `v1alpha` is the version of that group. Together they identify the versioned schema the object should have.

wora:
I addressed the concern of apiVersion vs group/version.

STYLE-GUIDE.md
xiaolanz: Maybe in the yaml example, listing out a full CRD definition. We probably also need to make rule that CRD resource name should be in format of plurals + api group. 

apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: serviceclasses.config.istio.io
spec:
  group: config.istio.io
  version: v1alpha1
  names:
    plural: serviceclasses
    singular: serviceclass
    kind: ServiceClass

wora: Can we address this question in another PR? Let's make some incremental improvement first.

wora:
@geeknoid Can you merge this PR?

ozevren:
With this guidance, is it possible to create meta-APIs such as LROs? It looks like requiring 1-1 mapping between proto package name and apigroup seems to be work against this.

STYLE-GUIDE.md
```diff
@@ -38,8 +38,14 @@ naming convention:
 ## Versioning
 
 When defining Kubernetes Custom Resource Definition (CRD) using
-proto3, the proto versioning should match the Kubernetes versioning,
-see the following example.
+`proto3`, follow the following guidelines:
```
geeknoid: , follow these guidelines:

geeknoid: , follow these guidelines:

geeknoid:
@ozevren Doesn't the sentence "When defining Kubernetes Custom Resource Definition" cover the case you're worried about? LRO is specifically not a CRD.

ozevren:
Meant to use LRO as an example. It is a meta-API. We're using CRDs to
create a public API surface (which happens to be backed by the API Server).
Adding this constraint implies you cannot create a meta set of protos to
use as top-level mix-in resources. And, it also means you cannot re-use
top-level CRD protos for another "API group" without violating this rule
(i.e. think of the Config Click Stop-0 we've been talking about).

wora:
K8s APIs support watch semantics, which is equivalent to LRO. I don't think watch semantics is implemented by any api group. It is more like a design pattern. We need to address it next, but I don't think it is blocked by this discussion.

geeknoid:
Oz was using LRO as an example of a pattern, not as a specific case.

As stated right now, the guide doesn't allow for common definitions that
can be shared between multiple APIs because we are forcing a 1:1 mapping
pattern. We just need to add more nuance to the text to allow for those
cases.

rshriram:
I have no idea about LROs, but as such, I am not against this proposal for routing section.

wora:
Can someone merge this PR or let me any further feedback? Thanks.

Issue #344, opened 2018-02-01
Proposal for refactoring mixerclient configuration


ayj:
The current mixerclient protos conflates user intent with implementation specific filter config. We ought to decouple them. User facing config is dimensioned by feature (auth, quota, api-spec) and should be organized according to the [proposed proto package scheme](https://docs.google.com/document/d/1C_NMYj9ZhdUWwZMWUoX_SLU8VoiV8U8V6yUJVYcdSS8/edit#heading=h.qex63c29z2to). Filter config is dimensioned by service (i.e. envoy route).

This will allow better independent evolution of the user-facing APIs from filter/proxy specific implementations, e.g. mixerclient, upstream envoy ([auth](https://github.com/envoyproxy/data-plane-api/blob/master/envoy/api/v2/auth/auth.proto), [ratelimit](https://github.com/envoyproxy/data-plane-api/blob/master/envoy/api/v2/ratelimit/ratelimit.proto)).

We should also refactor the mixerclient filter configuration to be defined per route (via RDS) instead of per listener (via LDS). This is both more performant, as service config can then be delivered via RDS instead of the more costly LDS, which requires draining listeners, and more consistent with upstream Envoy filters, e.g. [here](https://github.com/envoyproxy/data-plane-api/blob/a07a36853e698d5fee7b011d7abd1dcb87eda45e/envoy/api/v2/route/route.proto#L98).

- [x] Introduce a new user-facing AuthenticationPolicy. @diemtvu already started this work (see https://github.com/istio/api/pull/335). Pilot consumes AuthenticationPolicy and produce the corresponding mtls/auth filter configuration for envoy/mixerclient.

- [ ] Introduce a new user-facing API proto. OpenAPI (v2/v3), gRPC, and generic REST schemes should be considered. Options include a single umbrella type with subtypes per scheme or first-class types for each of the scheme. Pilot consumes these API spec(s) and produces the corresponding normalized [HTTPAPISpec](https://github.com/istio/api/blob/master/mixer/v1/config/client/api_spec.proto) for mixerclient filter.

- [ ] Introduce a new user-facing Quota proto. Schema is TBD. Config WG probably has useful input on what this should look like. Pilot consums this new proto and produce corresponding [QuotaSpec](https://github.com/istio/api/blob/master/mixer/v1/config/client/quota.proto) for mixerclient filter.

[...]

PR #345, opened 2018-02-02
Add no_jwt_patterns to auth.proto

qiwzhang:
If a request doesn't have JWT, it will be rejected.
But some requests are OK without JWT, such as OPTIONS for CORS,
or some paths for health checking.

This auth config change allow users to specify these pattern.


mixer/v1/config/client/auth.proto
```diff
@@ -146,6 +146,22 @@ message EndUserAuthenticationPolicySpec {
...
+
+     // Defines a path prefix to match
+     string path_prefix = 2;
+  }
+
+  // If a request doesn't have JWT, it will be rejected.
+  // But some requests are OK without JWT, such as OPTIONS for CORS,
+  // or some paths for health checking.
+  // If a request matches any of no_jwt_patterns, it is OK without JWT.
+  repeated Pattern no_jwt_patterns = 3;
```
wora: This becomes another specialized URL map, and you also put the business logic with the map. There are too much business logic connected together here.

Why EndUser is mentioned in the proto spec? Why the auth is not a service account, or a mobile device, or a IoT, or a delegated service, or impersonation? I think we should have an abstraction here. The URL map produce an operation id, and the operation id maps to authentication requirements. The concept of JWT or end user should not be the top level concept.

qiwzhang: We will address some of your concerns with this issue
https://github.com/istio/api/issues/344

ayj: EndUserAuthenticationPolicy can be renamed to auth filter config with #344. There was some previous discussion about treating this matching as authZ problem and handling in the mixer. But that can be inefficient as we end up unconditionally trying to validate all JWT for a host regardless of the path/method. Could an attribute match condition similar to the quotaspec make sense here, or should it be specialized for specific attributes (e.g. api.operation)?  

The top-level AuthenticationPolicy (see https://github.com/istio/api/pull/335) possibly ought to support this mapping of auth requirements to particular sets of operations (or entire service or maybe even namespace). @diemtvu

diemtvu: If I understand correctly, this is aligned with the SelectCriteria discussed [here](https://docs.google.com/document/d/1ezP4UuOn3JXEs_cXW4GyPGq-Ppq_XhS9-M-lN6ocOA4/edit?ts=5a67bda0#heading=h.84ucij65egwn), though I haven't fully fleshed out all details. However, our intention is to have that in the AuthN policy.

qiwzhang: We should be careful about which level of config we are talking here.  This is NOT user level config.  This is low-level component level config.  Specifically this is the config for Envoy auth filter defined here: https://github.com/istio/proxy/blob/master/src/envoy/auth/config.proto

Auth filter will be upstreamed to Envoy.  It should not depended on any Istio proto files.  That config proto is just copied from this proto file. So this proto file is the low level Envoy filter config for Mixer Filter and Auth filter.

User level auth config will be defined in @diemtvu PR https://github.com/istio/api/pull/335

kyessenov: We _can_ depend on core protos in data-plane-api. If there is no good match, then we can start a discussion to add another core proto. There's value in consolidating proto definitions, instead of copying them from place to place.

wora: I am a bit confused by the structure of this proto. Pattern matching is a common way for software, and it should looks like this way for most people:
```
match xxx:
  action
```
This proto puts the pattern matching inside the action, which is counter intuitive.

Pattern matching on request should be a common function across Envoy and Istio. I strongly prefer we define it in a central place. JWT rule essentially an action that produces the request auth attributes.

Issue #350, opened 2018-02-02
Add missing docs


geeknoid:
## Generate mixer/v1/config/client/*.pb.go + mixer/v1/config/client/istio.mixer.v1.config.client.pb.html
 no comment found for istio.mixer.v1.config.client
mixer/v1/config/client/client_config.proto:69:2: no comment found for NetworkFailPolicy
mixer/v1/config/client/auth.proto:150:0: no comment found for EndUserAuthenticationPolicySpecReference
mixer/v1/config/client/auth.proto:116:2: no comment found for locations
mixer/v1/config/client/quota.proto:125:2: no comment found for QuotaSpecReference
#

PR #353, opened 2018-02-02
add missing docs for mixerclient filter config

ayj:
fixes https://github.com/istio/api/issues/350


ayj:
cc @qiwzhang

mixer/v1/config/client/quota.proto
```diff
@@ -123,6 +123,7 @@ message QuotaSpecBinding {
   // REQUIRED. One or more services to map the listed QuotaSpec onto.
   repeated IstioService services = 1;
 
+  // QuotaSpecReference defines a reference to a QuotaSpec.
```
douglas-reid: Can we make this less of a regurgitation?

Maybe something like:
> QuotaSpecReference uniquely identifies the QuotaSpec used in the Binding

?

mixer/v1/config/client/auth.proto
```diff
@@ -148,6 +149,8 @@ message EndUserAuthenticationPolicySpec {
   repeated JWT jwts = 2;
 }
 
+// EndUserAuthenticationPolicySpecReference defines a reference to an
+// an EndUserAuthenticationPolicySpec.
```
douglas-reid: Can we make this more useful ?

> ...Ref uniquely identifies a policy being bound to a set of services.

?

PR #355, opened 2018-02-03
Correct a small makefile issue

geeknoid:
- Was passing a incorrect directory when building the protoc-gen-doc plugin. Not sure how come this worked before,
but the directory was definitely incrrect.

- Sync to the latest doc gen tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,7 @@ $(protoc_gen_gogoslick) : vendor
 
 $(protoc_gen_docs) : vendor
 	@echo "Building protoc-gen-docs..."
-	go build --pkgdir $(protoc_gen_docs_path)/tools/protoc-gen-docs -o $(protoc_gen_docs) ./$(protoc_gen_docs_path)/protoc-gen-docs
+	go build --pkgdir $(protoc_gen_docs_path)/protoc-gen-docs -o $(protoc_gen_docs) ./$(protoc_gen_docs_path)/protoc-gen-docs
 
 $(protoc_min_version) : vendor
 	@echo "Building protoc-min-version..."
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -50,7 +50,6 @@
     "ptypes",
     "ptypes/any",
     "ptypes/duration",
-    "ptypes/struct",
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -235,7 +235,11 @@ $(mixer_template_pb_gos) $(mixer_template_pb_doc) : $(mixer_template_protos) | d
 mixer/v1/config/fixed_cfg.pb.go mixer/v1/config/istio.mixer.v1.config.pb.html: mixer/v1/config/cfg.proto | depend $(protoc_gen_gogo) $(protoc_bin)
 	# Generate mixer/v1/config/fixed_cfg.pb.go (requires alternate plugin and sed scripting due to issues with google.protobuf.Struct)
 	@$(protoc) $(proto_path) $(gogo_plugin) $(protoc_gen_docs_plugin)mixer/v1/config $^
-	@sed -e 's/*google_protobuf.Struct/interface{}/g' -e 's/ValueType_VALUE_TYPE_UNSPECIFIED/VALUE_TYPE_UNSPECIFIED/g' mixer/v1/config/cfg.pb.go >mixer/v1/config/fixed_cfg.pb.go
+	@sed -e 's/*google_protobuf.Struct/interface{}/g' \
+	    -e 's/ValueType_VALUE_TYPE_UNSPECIFIED/VALUE_TYPE_UNSPECIFIED/g' \
+	    -e 's/package config/package istio_mixer_v1_config/g' \
```
douglas-reid: i don't think we want to do this. let's fix the package in istio/istio

geeknoid:
Can we deal with the naming in a separate PR? I'm trying to get a bunch of doc work done, and the fact this is currently broken makes it so I can't do "go generate" on the protos in istio/istio. With this fix, the end to end pipeline is working.

douglas-reid:
Let's use this: https://github.com/istio/istio/pull/3146.

The package in istio/istio is not needed. And we will simplify the codegen.

douglas-reid:
@geeknoid we'll also need something like: https://github.com/istio/api/pull/356.

geeknoid:
PTAL

PR #357, opened 2018-02-05
Extract package-level comments into a per-package README.proto, to slghtly increase discoverability.

geeknoid:


geeknoid:
Turns out this is breaking thing for Go code generation. Stuff doesn't like having a .proto which doesn't define anything :-(

Closing for now.

PR #358, opened 2018-02-05
Configurable mixer cluster names for mixer filter

rshriram:
The mixer cluster names is hardcoded to "mixer_server". This name is prone to collisions if user has one by the same name. Also, it is quite inconvenient to have a hardcoded cluster that one cannot control/customize.

This PR introduces configuration options for check cluster and report cluster. By default, they will continue to use the currently hardcoded "mixer_server" as the cluster names. Over time, we can expose global configs that will let users point to separate upstream clusters for mixer check and report calls (for load distribution/management/etc.).


rshriram:
fyi @elevran

douglas-reid:
In general, I don't have any problems with this change. It does impact some of the assumptions hardcoded in the mixer grafana dashboard. I'm not sure what the right approach is to making that more flexible is, but I'm not sure it should block this feature (if needed).

rshriram:
I think those dashboards won't break. We could document that if you change the default names of check/report cluster, it is the user's responsibility to update the out-of-box dashboards that Istio provides.

mixer/v1/config/client/client_config.proto
```diff
@@ -79,6 +79,17 @@ message TransportConfig {
...
+  // Name of the cluster that will forward check calls to a pool of mixer
+  // servers. Defaults to "mixer_server".
+  string check_cluster = 6;
+
+  // Name of the cluster that will forward report calls to a pool of mixer
+  // servers. Defaults to "mixer_server". By using different names for
+  // checkCluster and reportCluster, it is possible to have one set of
+  // mixer servers handle check calls, while another set of mixer servers
+  // handle report calls.
+  string report_cluster = 7;
```
geeknoid: Please add a comment about the impact of renaming these clusters on downstream things like the dashboards.

Please replicate the comments between check_cluster and report_cluster. That is, don't assume someone reads the comment of report_cluster when they're looking at the check_cluster field.

mandarjog:
This is great. I was going to add this for Mixer HA story anyway.

PR #359, opened 2018-02-06
introduce api/mixer/adapter/model/v1beta

guptasu:
introduce api/mixer/adapter/model/v1beta
As per the plan https://docs.google.com/document/d/1C_NMYj9ZhdUWwZMWUoX_SLU8VoiV8U8V6yUJVYcdSS8/edit?ts=5a7220f9#


mixer/adapter/model/v1beta/BUILD
```diff
@@ -0,0 +1,26 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
```
geeknoid: Nuke this file, no bazel no more!

mixer/adapter/model/v1beta/extensions.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package istio.mixer.adapter.model.v1beta;
+import "google/protobuf/descriptor.proto";
+
+// Specifies the variety of the the Template.
```
geeknoid: // The available varieties of templates, controlling the semantics of what an adapter does with each instance.

mixer/adapter/model/v1beta/extensions.proto
```diff
@@ -0,0 +1,21 @@
...
+enum TemplateVariety {
+    // TEMPLATE_VARIETY_CHECK makes the template applicable for Mixer's check calls.
```
geeknoid: No need to repeat the enum's name in the description text, it doesn't look good in the generated docs.

mixer/adapter/model/v1beta/value_type.proto
```diff
@@ -0,0 +1,59 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.mixer.adapter.model.v1beta;
```
geeknoid: One of the protos in this package needs to have a package-level comment. This will be used at the top of the page on istio.io.

Additionally, one of the protos needs to have a comment that contains $location, $title, and $overview which are used to drive code gen. See other protos in istio/api for examples.

mixer/adapter/model/v1beta/BUILD
guptasu: D

mixer/adapter/model/v1beta/value_type.proto
guptasu: Will fix it. Do you think we should fail the BUILD if the package is missing these necessary stuff. I remember you already  have warnings for that stuff in your tool.

guptasu: D

mixer/adapter/model/v1beta/standard_types.proto
```diff
@@ -0,0 +1,45 @@
...
+
+// This proto describes the types that can be used inside Mixer templates. These message types are used to specify
+// field datatype to express the equivalent ValueType for the expressions the field can be mapped to.
+package istio.mixer.adapter.model.v1beta;
+
+option go_package="istio.io/api/mixer/adapter/model/v1beta";
+
+// Value is used inside templates for fields that have dynamic types. The actual datatype
+// of the field depends on the datatype of the expression used in the operator configuration.
+message Value {}
```
wora: If you want to define standard types, you should define it under istio.type package and promise never making any breaking change. Don't call it standard if you cannot commit to it.

mixer/adapter/model/v1beta/value_type.proto
```diff
@@ -0,0 +1,61 @@
...
+
+option go_package="istio.io/api/mixer/adapter/model/v1beta";
+
+// ValueType describes the types that values in the Istio system can take. These
+// are used to describe the type of Attributes at run time, describe the type of
+// the result of evaluating an expression, and to describe the runtime type of
+// fields of other descriptors.
+enum ValueType {
```
wora: Why don't we use attribute type?

mixer/adapter/model/v1beta/standard_types.proto
guptasu: These are meant to be used only for templates, so putting inside istio.type might be too much. However, I agree standard is not necessary, I can fix it, will rename it to type.proto. WDYT ?

mixer/adapter/model/v1beta/standard_types.proto
```diff
@@ -0,0 +1,45 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+// This proto describes the types that can be used inside Mixer templates. These message types are used to specify
```
geeknoid: Only one package-level comment per package, not one per file.

Thanks

mixer/adapter/model/v1beta/value_type.proto
guptasu: @wora which attribute type. [AttributeManifest is actually ](https://github.com/istio/api/blob/master/mixer/v1/config/cfg.proto#L85)using ValueType. Am I missing something ?

mixer/adapter/model/v1beta/standard_types.proto
guptasu: Oops. This was already there, forgot to remove it.

PR #362, opened 2018-02-07
separate check and report clusters

rshriram:
Pursuant to PR #358 ..
TODO: Once this is implemented in Pilot such that policy_check_cluster and telemetry_cluster are always specified, we need to 

1. Change proxy such that a missing check_cluster implies disable policy checks, a missing telemetry_cluster implies disable telemetry, but enable policy.
2. After changing proxy, remove the disable_policy_checks option from global config, update docs, and remove dead code from pilot.


mesh/v1alpha1/config.proto
```diff
@@ -117,15 +117,25 @@ message ProxyConfig {
 // MeshConfig defines mesh-wide variables shared by all Envoy instances in the
 // Istio service mesh.
 message MeshConfig {
-  // Address of the egress Envoy service (e.g. _istio-egress:80_).
-  string egress_proxy_address = 1;
```
douglas-reid: why is `egress_proxy_address` being removed as part of the `mixer_address` separation?

rshriram: Ah shux.. It was an unused config item.. We got rid of egress proxy a long time ago. Thought I would remove it as well.

mesh/v1alpha1/config.proto
```diff
@@ -117,15 +117,25 @@ message ProxyConfig {
...
-
-  // Address of the mixer service (e.g. _istio-mixer:15004_).
-  // Empty value disables Mixer checks and telemetry.
-  string mixer_address = 2;
-
-  // Disable policy checks by the mixer service. Metrics will still be
-  // reported to the mixer for HTTP requests and TCP connections. Default
+  // Address of the server that will be used by the proxies for policy
+  // check calls (e.g. _istio-mixer:15004_). By using different names for
+  // policyCheckServer and telemetryServer, it is possible to have one set
```
ayj: Can we use `mixer_check_server` and `mixer_report_server` for consistency with the mixer [API service](https://github.com/istio/api/blob/ffa742d26a8884c4f144ab5ae41f3b8ae0117e59/mixer/v1/service.proto#L48)?

rshriram: That is a good point. What I am trying to do here is to keep the config at concept level (policies, telemetry, auth, etc.). There are other parts of the config that are essentially pointing to implementation specific things, which to a layman is going to be hard to understand. If I am starting off with Istio, I look at global config and turn off things I don't want/enable things I want. Make sense?

(I am waiting to get rid of those RDSrefresh delays, etc. but haven't found a good way to do it yet).

ayj: Somewhat. We seem to be conflating mesh administrator configuration and operator policy config into the same protobuf which makes this difficult. Configuring whether a feature (e.g. auth, telemetry) is enabled globally (or per namespace / service) is a different concept than how the services that enable those features are accessed by the proxies.

rshriram:
Lets not merge this until all tests pass in istio/istio for my PR. Dont want to upset stuff..

PR #367, opened 2018-02-09
Make fail policy configuration per service.

liminw:
This PR introduces per-service fail policy configuration when Mixer is unreachable. We currently only have mesh level fail policy, which is by default FAIL_OPEN. For services that use Mixer for critical checks like authorization, FAIL_OPEN is not acceptable.  We should provide finer grain control for fail policy. 

Issue [#2805](https://github.com/istio/istio/issues/2805) provides background of the PR.


mixer/v1/config/client/client_config.proto
```diff
@@ -54,6 +63,11 @@ message ServiceConfig {
 
   // End user authentication policy.
   EndUserAuthenticationPolicySpec end_user_authn_spec = 6;
+
+  // Specifies the policy when failed to connect to Mixer server.
+  // This is the service-level policy. It overrides mesh-level policy (defined
+  // in TransportConfig#network_fail_policy).
+  NetworkFailPolicy network_fail_policy = 7;
```
qiwzhang: It is better to embed it inside a message so that we can use has_ call to check if this field is set.

Our logic should be if service_config.has_network_fail_policy()   use it, otherwise use the one from global.

mixer/v1/config/client/client_config.proto
```diff
@@ -34,6 +34,21 @@ option (gogoproto.goproto_getters_all) = false;
 option (gogoproto.equal_all) = false;
 option (gogoproto.gostring_all) = false;
 
+// Specifies the policy when failed to connect to Mixer server.
+message NetworkFailPolicy {
```
qiwzhang: I would prefer to have service_config one to be embedded, but not the mash level one.

liminw: Once we create such a NetworkFailPolicy message, they should be used consistently. Are you worried about changing existing code?

liminw:
@geeknoid @douglas-reid Could you review the change? I need to code owner review in order to check it in.

mixer/v1/config/client/client_config.proto
```diff
@@ -34,6 +34,21 @@ option (gogoproto.goproto_getters_all) = false;
 option (gogoproto.equal_all) = false;
 option (gogoproto.gostring_all) = false;
 
+// Specifies the policy when failed to connect to Mixer server.
```
geeknoid: Specifies the behavior when the client is unable to connect to Mixer.

mixer/v1/config/client/client_config.proto
```diff
@@ -34,6 +34,21 @@ option (gogoproto.goproto_getters_all) = false;
...
+message NetworkFailPolicy {
+  // Describes the policy.
+  enum FailPolicy {
+    // If network connection fails, request is passed to the backend.
```
geeknoid: request is allowed and delivered to the service.

mixer/v1/config/client/client_config.proto
```diff
@@ -34,6 +34,21 @@ option (gogoproto.goproto_getters_all) = false;
...
+    FAIL_OPEN = 0;
+
+    // If network connection fails, request is rejected.
+    FAIL_CLOSE = 1;
+  }
+
+  // Specifies the policy when failed to connect to Mixer server.
```
geeknoid: Reuse comment from proto above.

mixer/v1/config/client/client_config.proto
```diff
@@ -54,6 +69,11 @@ message ServiceConfig {
 
   // End user authentication policy.
   EndUserAuthenticationPolicySpec end_user_authn_spec = 6;
+
+  // Specifies the policy when failed to connect to Mixer server.
+  // This is the service-level policy. It overrides mesh-level policy (defined
+  // in TransportConfig#network_fail_policy).
```
geeknoid: To reference a type, use [TextToDisplay][PackageName.TypeName]

mixer/v1/config/client/client_config.proto
```diff
@@ -54,6 +69,11 @@ message ServiceConfig {
 
   // End user authentication policy.
   EndUserAuthenticationPolicySpec end_user_authn_spec = 6;
+
+  // Specifies the policy when failed to connect to Mixer server.
```
geeknoid: Reuse description from proto above .

liminw:
@geeknoid Thanks for the review! I have addressed your comments. Please take a look again.

PR #370, opened 2018-02-09
Add a separate mTLS port for pilot and mixer

wattli:
App Envoy talks with pilot/mixer based on when auth is enabled during the app is brought up. However, if auth is disabled and then enabled at runtime, app envoy will not able to take to pilot. This allows pilot/mixer to expose two ports and fix the bug.


mesh/v1alpha1/config.proto
```diff
@@ -110,38 +110,40 @@ message ProxyConfig {
...
   int32 stat_name_length = 15;
 
-  // The number of worker threads to run. Default value is number of cores on the machine.
-  int32 concurrency = 16;
+  // Address of the discovery service exposing xDS without mTLS (e.g. _istio-pilot:8080_).
+  string discovery_address = 16;
+
+  // Address of the discovery service exposing xDS with mTLS (e.g.
+  // _istio-pilot:15005_).
+  string discovery_mtls_address = 17;
```
rshriram: Suggestion: create a common message called ServerAddress, which should have two strings: secure, insecure 
Then do ServerAddress pilot, ServerAddress mixer_check, ServerAddress mixer_report
(for backward compatibility, mark the current discovery_address as deprecated).

wattli: Done, thanks

mesh/v1alpha1/config.proto
```diff
@@ -41,6 +41,15 @@ enum AuthenticationPolicy {
   INHERIT = 1000;
 }
 
+// ServerAddress defines the address of servers like mixer, pilot, etc.
```
wora: defines => specifies

Please be more clear what address means. For example, is it an address of a specific server node? Is it a load balanced address like a DNS name? Is it an address of an HTTP server or gRPC server?

If we need to add deadline, retry hint, and other settings, is ServerAddress still the right term? Envoy uses `GrpcService` to specify equivalent concept.

mesh/v1alpha1/config.proto
```diff
@@ -41,6 +41,15 @@ enum AuthenticationPolicy {
...
+message ServerAddress {
+  // The address for mTLS server, e.g., (_istio-pilot:15003_)
+  string mutual_tls = 1;
+
+  // The address for plain text server, e.g., (_istio-pilot:15005_)
+  string plain_text = 2;
```
wora: Are the two address both required or only one is required?

mesh/v1alpha1/config.proto
```diff
@@ -74,7 +83,7 @@ message ProxyConfig {
   // MUST BE greater than _drain_duration_ parameter.
   google.protobuf.Duration parent_shutdown_duration = 5;
 
-  // Address of the discovery service exposing xDS (e.g. _istio-pilot:8080_).
+  // Deprecated, use server_address instread.
```
wora: If you mark a feature is deprecated, please try to give a deadline for this to be removed. Otherwise it would sit here forever.

mesh/v1alpha1/config.proto
wattli: Comment is updated. It's only for mixer and pilot and this name is suggested by @rshriram , i think it makes sense.

mesh/v1alpha1/config.proto
wattli: comment updated, at least one needs to be specified.

mesh/v1alpha1/config.proto
wattli: That's a good point, i think the idea to put a 'Deprecated' here is to prevent people adding new code to use it. And for clean up, i am not sure the general process of Istio api. (As you can see there are other deprecated fields in this file)

mesh/v1alpha1/config.proto
```diff
@@ -112,27 +122,18 @@ message ProxyConfig {
...
 message MeshConfig {
-  // Address of the server that will be used by the proxies for policy
-  // check calls (e.g. _istio-mixer:15004_). By using different names for
-  // mixerCheckServer and mixerReportServer, it is possible to have one set
-  // of mixer servers handle policy check calls, while another set of mixer
-  // servers handle telemetry calls.
-  //
-  // NOTE: Omitting mixerCheckServer while specifying mixerReportServer is
-  // equivalent to setting disablePolicyChecks to true.
+  // Deprecated, use mixer_check instread.
```
rkpagadala: typo: instead

mesh/v1alpha1/config.proto
```diff
@@ -112,27 +122,18 @@ message ProxyConfig {
...
   string mixer_check_server = 1;
 
-  // Address of the server that will be used by the proxies as an Istio
-  // telemetry sink (access logs, API/connection metrics)
-  // (e.g. _istio-mixer:15004_). By using different names for
-  // mixerCheckServer and mixerReportServer, it is possible to have one set
-  // of mixer servers handle policy check calls, while another set of mixer
-  // servers handle telemetry calls.
+  // Deprecated, use mixer_report instread.
```
rkpagadala: same here

PR #375, opened 2018-02-15
Rename RouteRule

frankbu:
When moving from `v1alpha1` to `v1alpha2` routing all the resource names have changed except `RouteRule`. This is a problem because of the ambiguity and also because a RouteRule resource in `v1alpha2` contains more than one route config (i.e., all routes for a destination are stored in a single resource now), so the name `RouteRule` is not a good one anymore.

I think `TrafficRule` is a good alternative for 2 reasons:

1. It now really is a rule that configures the "traffic routing" for a particular destination
2. It seems complimentary to `DestinationRule` which is the rule configuring a particular destination's policies/settings.

With this change, the migration from `v1alpha1` to `v1alpha2` is summarized as follows:

1. `RouteRule` -> `TrafficRule`
2. `DestinationPolicy` -> `DestinationRule`
3. `EgressRule` -> `ExternalService`
4. `Ingress` -> `Gateway` (recommended to use)


kyessenov:
fine with me, too. if you're willing we could consider replacing "rule"
with "policy" (a-la NetworkPolicy vs NetworkRule)

rshriram:
That would bring DestinationPolicy back wouldn't it? (s/rule/policy)

kyessenov:
oh shoot. can we reconsider changing that to ClientPolicy? i just think
policy is a more appropriate term since it's a collection of rules.

rshriram:
ClientPolicy or ClientRule was considered a while ago, and we dumped it coz the destinationrule is something the service owner defines and does not care where it gets enforced. (not that destinationrule is a perfect name, but it seems to be the better of the two).

louiscryan:
I'm not sold. TrafficRule is really ambiguous - it doesn't actually describe the intent. (NetworkPolicy isn't great either but at least people think it policy==acl so they get it)

As an example how do the names describe the difference between TrafficRule & DestinationRule ?

frankbu:
I kind of like Kuat's suggestion of `TrafficPolicy`. To address the problem with `DestinationRule` -> `DestinationPolicy` reintroducing a `v1alpha1` conflict, I'm wondering if we can solve it by get rid of DestinationRule/Policy entirely, and just put those properties into the TrafficPolicy resource as well. It seems kind of awkward anyway, now that we have subsets, to need to update two separate resources whenever we want to route to the new subset. I took a pass at this and updated this PR with the merge. PTAL

louiscryan:
@frankbu The choice to move subsets out of RouteRule was explicit done as a separation of concerns after careful review. The goal is not to optimize for how many resources need to be updated to cause a specific effect, its whether the resources properly model the roles and responsibilities of the folks who use them. In particular DestinationRule is unlikely to be used all that often and should be owned by the service producer. RouteRule is targeted at allowing customizations to routing that are more driven by a service consumer.

louiscryan:
... more specifically a RouteRule is constrained to a set of hosts that have a common behavior for routing (i.e a set of virtual hosts that are synonymous with each other) whereas DestinationRule is a set of services that have common policies applied to traffic post-routing. Trying to couple these things will cause there to be far more DestinationRule's than is usefully necessary.

frankbu:
Ah, I was thinking users can combine, for example, the subsets and the routes into a single `TrafficPolicy` resource where it makes sense, but that it isn't required. One could alternatively separate the them into two resources, if that makes more sense, only the two resources would both be type `TrafficPolicy`, instead of one `RouteRule` and one `DestinationRule`. Is there something fundamentally broken with such a flexible arrangement?

frankbu:
If we stick with 2 resources, to encourage proper separation of concerns, we still need to rename `RouteRule`, at least.

The other suggestion was `RoutingRule` (just to get a slight name change), but it still seems like there should be a better name for both `RouteRule` and `DestinationRule` to better describe what they are. Any other suggestions?

mandarjog:
For http ‘virtualHost’ correctly describes the contents. 
Even though it is not fully consistent with tcp, I contend that most people will get the analogy.

rshriram:
HostRules or VHostRule or RoutePolicy or Routelets?

Or PreRoutingRule (routing rule) and PostRoutingRule (for destination rule)

wora:
I think the challenge here is network traffic has at least 2 ends: source and destination. Whatever policy we have, it means different party depending on which direction we are looking at it. I find it is very difficult to understand these protos in general, regardless how we name them.

On the other hand, policies are generally easier to understand, because are often attached to a specific entity, such as ACL on a document. I think we need to clarify how network rules or policies work in general. The naming alone won't solve the problem.

xiaolanz:
A few suggestions from the perspective of API style and config. 
- Given the fact that v2 route rule are completely different and incompatible with v1, renaming RouteRule is recommended. Renaming the proto message implies that we will rename the CRD Kind and underlying a new storage. For the long run, we expect version evolution of the same CRD Kind to be both forward and backward compatible so Istio config can have a smooth upgrade path. Therefore, any incompatible breaking change would be better with a new Kind. 

- Please keep in mind that the new route rule should follow the new api group naming convention that using routing.istio.io as the api group name vs. config.istio.io. I'm working on getting Pilot to use allow different apiversions of istio configs. "config.istio.io" will be deprecated. DestinationRule CRD resource name will be like DestinationRule.routing.istio.io. So RoutingRule.routing.istio.io will not be a good choice.  All CRD api group should in general align the proto package path name. See https://github.com/istio/istio/pull/3214

frankbu:
My first reaction to the suggestion: `RouteRule` -> `VirtualHost` was that it seems to be missing the primary purpose that it's configuring the routes, but given what @xiaolanz said (above) about the api group rename: `config.istio.io` -> `routing.istio.io`, I think it could work:
```
apiVersion: routing.istio.io/v1alpha2
kind: VirtualHost
```
Basically that says it's an "Istio routing host" config.

Then, maybe we could align `DestinationRule` better with `VirtualHost` by just naming it `Destination`.

`RouteRule` -> `VirtualHost`
`DestinationRule` -> `Destination`

@louiscryan @rshriram what do you think?

frankbu:
@rshriram pointed out that `routing.istio.io` might be to specific and suggested `networking.istio.io` instead. With that in mind, here is another proposal:

`config.istio.io` -> `networking.istio.io`
`RouteRule` -> `RouteHost`
`DestinationRule` -> `RouteDestination`

rshriram:
I like the RouteHost and RouteDestination under networking.istio.io.

louiscryan:
RouteHost feels a little off when i try to put it into documentation. 

"To split traffic for a service to a canary create a RouteHost for sales.internal that ...'

I like the idea of getting 'host' into the name as that is what the resource semantics are coupled too.

Why not just use 'RoutingRule' ? It doesn't collide with RouteRule which get us out of the short-term hole while keeping a naming scheme that was mostly OK. It also side-steps the initial concern about RouteRule now containing more than one 'route'.  'HostRoutingRule' works too as does 'HostRouter'

"To split a % of traffic for a service to a canary create a RoutingRule for sales.internal that ...'

I considered the following ...

Host - Not descriptive of intent
VirtualHost - has a strict definition that doesn't match even if the component words are good
VirtualService - ok, though we're trying to avoid the term service
CompositeService - vague enough but same issues as above
RoutedHost - ok semantically but doesn't read well in sentences
ForwardingHost - passable but wrong connotations I think
ForwardingRule - same as a above
Router - has its attractions but don't think we can go there
HostRouter - not terrible, I could get behind this, also reads well
ServiceRouter - also ok but 'service' ...
HostRouting - doesn't read well 
HostOverride - doesn't account for the fact that we can use the feature to create synthetic hosts
SyntheticHost - Similar to VirtualHost

wora:
RoutingRule sounds great. It really reflects the concept well and is very
easy to understand in the documentation. Let's go for it.

frankbu:
I can live with `RoutingRule`, but it isn't ideal now that the resource doesn't contain a rule anymore, but rather a list of routes/rules.

I also agree with @louiscryan that `RouteHost` is a little awkward in documentation (I noticed that when thinking about updating the API doc itself) but also think that `HostRouter` solves the problem. If we go with `HostRouter` I think that it also makes more sense to go with other suggested rename `DestinationRule` -> `RouteDestination`.

mandarjog:
What is our position on multiple `RouteRule` resources for the same set of `host`? Do we allow it?

If we do allow it then it is a significant departure from the typical `VirtualHost` abstraction.

If we don't allow multiple `RouteRules` per `host` set,  then the statement @louiscryan used as an example should we reworded from
```To split traffic for a service to a canary create a RouteHost for sales.internal that ...'
```
to 
```To split traffic for a service to a canary add an HPPTRoute to the for sales.internal VirtualHost that ...'
```
The 2nd statement reads better and is also more accurate.
We need to have `Host` to better denote the meaning.

On a related note, If we separate `TCPRoutes` and `HTTPRoutes` then the http part is clearly a `VirtualHost`. What is the rationale behind making tcp and http routes peers?  The amount of sharing between then is minimal. (Just hosts and gateways)

kyessenov:
How about RouteRuleSet or HostRoutes? We should not call it route rule
since it's not backwards compatible, and that's awfully confusing to
preserve the name but change the semantics.

By the way, the exceptional case in the paragraph is a hell to parse and
understand TBH. Once you have thousands of rules, thinking through the
traffic patterns is going to be hard if one has to keep exceptions like
this in the head. Can we simplify the assumptions somehow? Disjoint hosts
per routes is a lot simpler to reason about.

louiscryan:
@mandarjog 

The defined meaning of VirtualHost is 'more than one domain server by the same IP'. This is not what RoutingRule does, it does something more useful that in my mind. Its a pity because I actually like the term VirtualHost when its unconstrained by that meaning.

In most cases RoutingRule would be created to managed an 'exceptional' case. i.e they only exist when the default behavior of routing by host/service name is insufficient or when folks are creating new hosts as composites of multiple backend services (typically via Gateway). The phrasing I used in my example was for the exception case which is why I did not pre-suppose that a RoutingRule already existed.

Separating out resources by protocol was discussed and rejected for a variety of reasons:
- Unlikely to be additional value in the per-protocol resources having distinct lifecycles. Would almost always have the same author, same ACL etc.
- To add additional or 'meta' protocols we'd have to have an API extension mechanism coupled to CRDs which would be onerous

@kyessenov The affordance Zack described is to effectively allow for mergeability with K8S Ingress and some other niche use-cases. It should not affect common usage and I suspect we'll work on removing this over time via a better mechanism.

louiscryan:
tldr; preference VirtualService and DestinationRule | ServiceTrafficPolicy, can live with most permutations of
(VirtualService, VirtualHosts) X (DestinationPolicy, DestinationRule, ServiceTrafficPolicy)

... I should say that if we want to be brave and try to co-opt VirtualHost to mean

'a way to take an existing concrete host and virtualize how traffic intended for it is routed **_regardless_** of protocol' 
...as well as..
'a way to create a new host composed out of parts of others services'

then by all means we should do that. It is after all what a service mesh lets you do. The same argument stands if we want to adopt 'VirtualService' which I could get behind. Similarly you could make an argument for 'ProxiedHost' given the generic meaning of 'proxy' to act in stead of. Alternate forms along that theme

EffectiveHost - really just a synonym for VirtualHost, connotation of effectiveness
SurrogateHost - doesn't cover the 'create a new host' scenario well
SubstituteHost - same as above
DelegatingHost - ok-ish but similar to RoutingRule as its the verb form.
DelegatedHost - inversion of intent, should be the delegator not the delegatee
SyntheticHost - okish
PseudoHost - Sounds like a fake
HostRelay - implies intermediate hop which is not intent

I should also say that I prefer DestinationPolicy over DestinationRule given that we are as @wora mentioned attaching policies to existing services and that its common parlance to say things like 'load balancing policy'. Its just that CRDs are hosing us, if we can dig out of that hole easily then yay! Absent that Rule is a reasonable replacement for Policy. The only other option I can see would be:

ServicePolicy - feels way too generic, likely to collide with K8S 
ServiceTrafficPolicy - not too bad
HostTrafficPolicy - also not bad

@frankbu I don't like HostDestination as it implies that we're defining the destination which we are not, we are attaching policy to an already extant destination.

frankbu:
I kind of like `ProxiedHost`. Along the same lines, maybe `RoutableHost` would tie it better to its contents, route rules. Alternatively `RoutingHost`? 

@louiscryan I still think that if we go with some kind of `<thing>Host` name for the route rules/policies, we could/should do something similar for the destination rules. I think my previous suggestion, `RouteDestination`, is a good one because, in effect, it is sort of defining, i.e. modeling, the route destination, especially when you add subsets along with the policies. Note that `<something>Host` has the same issue since it's essentially describing the routing policies for a host that exists.

I think we're stuck in an endless loop with this issue and it sounds more and more like @louiscryan and @rshriram have already gone though all of this before this latest round even started :-( Bottom line is that the best choices seem to be blocked by restrictions that we can't use `RouteRule`, `DestinationPolicy`, plural name, and we have an large number of networking terminology collisions/ambiguities.

Unless we can agree to one of these latest proposals, I think we should just admit defeat, move on (for now) and just go with `RouteRule` -> `RoutingRule`.

WDYT

louiscryan:
I did a broad survey of the Google team and the consensus was to go with VirtualService and DestinationRule. The debate here rehashed much of what has been discussed on this thread but the outcome was strongly in favor of this pairing over others suggested above and we went through a lot of them.

kyessenov:
We also should use a fresh version, not v1alpha2.
Due to some history, our CRDs are actually already v1alpha2, and have been that way.
This new group should at least be v1alpha3 or v1beta1

linsun:
My preference is not to rename e.g. continue use RouteRule... I have been having a hard time to keep up with all the changes, I think not renaming is one less thing for me to keep up and be able to understand it quickly.

frankbu:
@linsun we have to rename because of the CRD conflict with v1.

I can live with `VirtualService` and I agree with @kuat that we should change the version as well because the CRDs are already `v1alpha2`. I think `v1alpha3` is best, since I don't think we want to call this `beta` yet.

I'll close this now and open new PRs for the 2 changes.

linsun:
@frankbu that is pretty sad to give up a good name because of the potential CRD conflicts when user has both v1alpha1 route rule and v1alpha2 or 3 route rules.  I'm okay with the new name... agreed we are not ready to call beta before we put it in alpha for testing first.

frankbu:
@linsun: `RouteRule` is not a very good name anymore because in the new version it is not just a single rule, it is now a list of rules for a given destination. I think `VirtualService` is a reasonable name, and better than RouteRule.

PR #379, opened 2018-02-17
create fixed_cfg.pb.go

guptasu:
do the fix_cfg stuff (replace struct with interface{}) in the new policy/v1beta1 package.


guptasu:
@geeknoid @rshriram  is it possible to force push this ? The prow-presubmit is complaining that the file `cfg.pb.go`, which is deleted in this PR, is not actually deleted and I should do `git add` on it. circleci builds fine which does the same validation. I need to investigate why prow seems to think cfg.pb.go is not delete. But to unblock my other work, I would request to force push this. 

Worst case if I am wrong, then I will roll back immediate.

PR #382, opened 2018-02-21
add name convention for global and namespaced configs.

xiaolanz:


STYLE-GUIDE.md
```diff
@@ -51,17 +51,24 @@ When defining Kubernetes Custom Resource Definition (CRD) using
 
 * The proto message type must match the CRD `kind` name.
 
+## Scoping
+
+Istio config supports global cluster and namespaced in similar way to Kubernetes resources. 
+`Cluster` a reserved word for Istio config Kind. Kind with a `Cluster` prefix, such as `ClusterPolicy` indicates a global scoped `Policy`. `ClusterPolicy` and `Policy` can map to the same proto message `Policy`, or one can chose to have a separate proto message for `ClusterPolicy`. 
+
+`ClusterPolicy` and `Policy` must be defined in the same `apiGroup`. The only difference is the scope.
```
rshriram: Does everyone have to define a Policy and ClusterPolicy ?

wora:
Let's not to mix the concept global and cluster. I don't think a cluster is
global. For comparison, YouTube is a global service, and it is likely to
have dozens of very large clusters. For AWS, almost everything is
regionalized, and the largest AWS region is probably bigger than most
companies.

Following GCP IAM and AWS IAM, a policy is a policy. You can apply the same
policy to different entities. There was never a concept of cluster policy.
Why do we need to do things differently if AWS is by far the most popular
cloud provider.

If we design a new system, we need to make sure the design is at least as
good as the existing one. We need to do some A/B study before we decide a
solution.

STYLE-GUIDE.md
xiaolanz: No. It's case by case. Some configs can just use one of it.

xiaolanz:
@wora I agree that cluster scope doesn't equal to global. I will fix that. Istio config currently only supports two scopes: cluster/namespace. The naming followings k8s convention, such as k8s RBAC: clusterrole and role: https://kubernetes.io/docs/admin/authorization/rbac/

wora:
In AWS, a role can be applied to either users or resources. It is quite
flexible, but it does come with some complexity. Having different types of
roles is a serious risk in the long term. Once we have more than one type
of things, we very likely will have N types in the future, then it becomes
unsustainable.

Kubernetes is a low level system. It is nowhere close to run a business
using Kubernetes alone. One the other hand, AWS or GCP is more closer to
meet business needs, but there are still serious gaps.

I recommend Istio to go after higher level business needs, not to simulate
Kubernetes. I don't think it is worthwhile to design two similar systems.
If we have to design a new system, I would ask can I run a very large
organization using this system.

xiaolanz:
A scoping model for Istio will capture this instead of extending k8s clusterRole model.

PR #383, opened 2018-02-23
Add two fields for discoveryService

wattli:
We need to convert yaml to ProxyConfig proto, and embedding proto makes the code less readable.


rkpagadala:
why is mixer and pilot different?

wattli:
I am working on pilot now, and will handle mixer later, and i think mixer is migrating to two different ports in the meantime.

wattli:
@rkpagadala , not sure how to reply your comment inline, please see my previous reply on top, thanks.

mesh/v1alpha1/config.proto
```diff
@@ -123,8 +123,14 @@ message ProxyConfig {
   // The number of worker threads to run. Default value is number of cores on the machine.
   int32 concurrency = 16;
 
-  // Address of the discovery service exposing xDS.
+  // Deprecated, use discovery_mtls_address/discovery_plain_address instead.
```
rkpagadala: i thought we wanted to use ServerAddress

I am ok with the change, please fix mixer also before .7

wattli: We need to convert yaml to this proto, and embedded proto makes the code less readable. And as mentioned above, i will update mixer after pilot work is done.

PR #385, opened 2018-02-26
adding correct kind name to v1alpha2 protos so users are not confused

GregHanson:
since there is discussion of having a release with both v1alpha2 and v1alpha1 rules present, we need to update the inline examples with the correct kind for route rules so users have valid v1alpha2 examples to play with


GregHanson:
resolves https://github.com/istio/istio/issues/3580

ZackButcher:
@GregHanson I think this PR isn't needed anymore in light of the renames in https://github.com/istio/api/pull/386. Is that correct?

wora:
Do we really want to include version into Kind? That is against k8s API and Istio API Style Guide.

GregHanson:
no longer needed after routerule->virtualservice rename

PR #386, opened 2018-02-26
Rename RouteRule -> VirtualService in v2 routing api

frankbu:
See #375 for discussion leading to this decision.

@rshriram @louiscryan @ZackButcher I reworded some of the doc, so please review to make sure it's correct.


routing/v1alpha2/gateway.proto
```diff
@@ -84,7 +84,7 @@ option go_package = "istio.io/api/routing/v1alpha2";
 // (i.e. 80 redirects to 443).
 //
 //     apiVersion: config.istio.io/v1alpha2
```
xiaolanz: FYI, move this onto the write api group: routing.istio.io/v1alpha3 when https://github.com/istio/istio/issues/3586 is done.

louiscryan: As discussed on old PR, lets not use routing.istio.io and use networking.istio.io or network.istio.io

routing/v1alpha2/destination_rule.proto
```diff
@@ -115,7 +115,7 @@ message TrafficPolicy {
 
 // A subset of endpoints of a service. Subsets can be used for scenarios
 // like A/B testing, or routing to a specific version of a service. Refer
-// to Route Rules documentation for examples on using subsets in these
+// to VirtualSerive documentation for examples of using subsets in these
```
louiscryan: typo

routing/v1alpha2/gateway.proto
```diff
@@ -69,10 +69,10 @@ option go_package = "istio.io/api/routing/v1alpha2";
 //           protocol: MONGO
 //
 // The gateway specification above describes the L4-L6 properties of a load
-// balancer. Routing rules can then be bound to a gateway to control 
+// balancer. A VirtualService can then be bound to a gateway to control
```
louiscryan: capitalize 'gateway'

routing/v1alpha2/gateway.proto
```diff
@@ -209,10 +209,10 @@ message Server {
   // services using TLS with SNI. Standard DNS wildcard prefix syntax
   // is permitted.
   //
-  // RouteRules that are bound to a gateway must having a matching host
-  // in their default destination. Specifically one of the route rule
+  // A VirtualService that is bound to a gateway must having a matching host
+  // in its default destination. Specifically one of the virtual service
```
louiscryan: keep VirtualService as proper noun

routing/v1alpha2/virtual_service.proto
```diff
@@ -45,35 +45,35 @@ import "google/protobuf/wrappers.proto";
...
-// The rule applies only to the ports on the destination service for which
-// it provides protocol-specific match condition, e.g. if the rule does not
-// specify a TCP condition, the rule does not apply to TCP traffic.
-//
-// For example, the following rule routes all traffic by default to pods of
-// reviews service with label "version: v1". In addition, HTTP requests
-// containing for /wpcatalog/, /consumercatalog/ url prefixes will be
-// rewritten to /newcatalog and send to pods with label "version: v2". The
-// rule will be applied at the gateway named "bookinfo" as well as at all
+// VirtualService provides the traffic routing policy for a destination service.
```
louiscryan: I would say something like.

"A VirtualService defines a set of traffic routing rules to apply when a host is addressed. Each routing rule defines matching criteria for traffic of a specific protocol, if the traffic is matched then it is sent to a named destination service defined in the registry. 

The source of traffic can also be matched in a routing rule, this allows routing to be customized for specific client contexts"

routing/v1alpha2/gateway.proto
kyessenov: For this, we need to move protos to the corresponding package, e.g. `networking/v1alpha3/gateway.proto`.

frankbu: I was originally planning to do this refactoring in a followup PR. @kyessenov any problem if I do it now, in this PR before merging?

kyessenov: I have no problem. Heads up to @ijsnellf since he's maintaining the implementation in istio.

PR #387, opened 2018-02-26
enable custom path to proxy template file

kidiyoor:
wrt https://github.com/istio/istio/issues/3745

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -132,8 +132,8 @@ message ProxyConfig {
   // Address of the discovery service exposing xDS with plain text connection.
   string discovery_plain_address = 19;
 
-  // Path to the proxy template file
-  string proxy_template_path = 20;
+  // Path to the proxy bootstrap template file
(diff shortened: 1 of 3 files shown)


mesh/v1alpha1/config.proto
```diff
@@ -131,6 +131,9 @@ message ProxyConfig {
 
   // Address of the discovery service exposing xDS with plain text connection.
   string discovery_plain_address = 19;
+
+  // Path to the proxy template file
+  string proxy_template_path = 20;
```
ZackButcher: Lets indicate that this is the bootstrap config template - maybe `proxy_bootstrap_template_path`?

kidiyoor: yes. fixed. using proxy_bootstrap_template_path now

ZackButcher:
No technical problem with this, just want to verify that `--custom-config-file` doesn't work for your use case first (see the comments in the issue).

kidiyoor:
There should be a way for users to provide 'customized template' in a 'not hardcoded path'.

Issue #377, opened 2018-02-17
Gateway selector not being parsed correctly


nmnellis:
I am seeing the following error in pilot after creating the below gateway config.
```
2018-02-17T02:46:37.724804Z	warn	LDS listing gateways: 1 error occurred:
 * YAML decoding error: selector:
- istio: gateway
servers:
- hosts:
  - example.com
  port:
    name: http
    number: 80
 json: cannot unmarshal array into Go value of type map[string]json.RawMessage
```

Gateway.yaml
```yaml
apiVersion: config.istio.io/v1alpha2
kind: Gateway
metadata:
  name: istio-ingress
  namespace: istio-system
spec:
  selector:
  - istio: gateway
  servers:
  - port:
      number: 80
      name: http
    hosts:
    - example.com

```

nmnellis:
@GregHanson should issues like these be created in api or istio/istio?

rshriram:
yuo might want to try the code from master. I fixed these issues.

rshriram:
Also, please move to istio/istio

nmnellis:
@rshriram what do you mean you fixed them? we updated the documentation to just reflect how it works without arrays, using maps

rshriram:
The gateway is actually implemented now :).

PR #392, opened 2018-02-28
Gateway selector doc

nmnellis:
@frankbu the documentation for gateway selectors was incorrect. See here https://github.com/istio/api/issues/377


frankbu:
Thanks @nmnellis !

PR #394, opened 2018-02-28
fix proxyBootstrapTemplatePath variable mismatch

kidiyoor:
mismatch was introduced by this PR : https://github.com/istio/api/pull/387


mesh/v1alpha1/config.pb.go
```diff
@@ -238,9 +238,9 @@ func (m *ProxyConfig) GetBinaryPath() string {
 	return ""
 }
 
-func (m *ProxyConfig) GetProxyTemplatePath() string {
+func (m *ProxyConfig) GetProxyBootstrapTemplatePath() string {
         if m != nil {
-                return m.proxyTemplatePath
+                return m.proxyBootstrapTemplatePath
```
GregHanson: `                return m.ProxyBootstrapTemplatePath`

ldemailly:
merged per #oncall discussion

ZackButcher:
Is this the result of running https://github.com/istio/api/blob/master/scripts/generate-protos.sh? The [file descriptor](https://github.com/kidiyoor/api/blob/222befc4a9f4009b4c1eee3da4c06d13b2aa8ed9/mesh/v1alpha1/config.pb.go#L575-L645) should change with the field name change too.

kidiyoor:
The generate-protos.sh was not run during https://github.com/istio/api/pull/387. proto.pb was manually updated.

ldemailly:
damn, sorry, zack is correct, is this hand crafted ? the hash/fd at the bottom should be changing each time - can we get a follow up PR asap with properly regen'ed files?

ldemailly:
should have been ran now, not manually tweaked

kidiyoor:
https://github.com/istio/api/pull/395

PR #395, opened 2018-02-28
run scripts/generate-protos.sh, cover up for PR387

kidiyoor:


geeknoid:
I don't think we need this anymore.

PR #397, opened 2018-03-01
Update .pb.go files

geeknoid:


ZackButcher:
Looks like the diff is dirty after running ./scripts/generate-protos.sh

douglas-reid:
/do-not-merge

It appears these were generated with go1.9.  Please use go1.10 for generation.

PR #398, opened 2018-03-02
Make the numbering of message fields consecutive

lei-tang:
Make the numbering of the message fields in the authentication policy consecutive.


wora:
Are you sure these fields are not being used? No breaking change should be allowed for Istio APIs.

lei-tang:
These fields are not being used (policy.proto is for the Istio authentication policy, which is a work in progress).

diemtvu:
lgtm, though I don't think this clean up is necessary.

douglas-reid:
Does this PR will need python generation updates now?

lei-tang:
Just did a search for "*.py" files under this repo and have not found any "*.py" files generated for the proto files. If python code is needed, please let me know.

douglas-reid:
https://github.com/istio/api/blob/master/python/istio_api/authentication/v1alpha1/policy_pb2.py ?

lei-tang:
Thanks! Python update has been added.

wora:
I recommend we never renumber a field unless it is marked as not ready, do not use. There is very little value to do it. Sooner or later a mistake would cause a major outage somewhere else.

geeknoid:
Should we close this PR?

@diemtvu

PR #405, opened 2018-03-13
unhide faults from virtual service

rshriram:
Fault filter can be implemented with the current API spec, with a small loss of perf (repeated header matching). It is independent of the work involving route-local filter stacks. 

We need to enable it in the LDSv2 and v1alpha3, so as not to cause a regression in current feature set.

The semantics are as follows:
1. fault filter must be configured with the same set of HTTP header match conditions (and URL, authority, etc.) as the route match
2. The fault filter must be configured with the same upstream cluster that is used by the route.
3. If the route uses a weighted destination, we need to add N fault filters to envoy listener (one per destination), where the only difference between each filter is the upstream cluster name. This has minimal perf implication because the upstream cluster is matched first in the fault filter code.
4. IF multiple routes have different fault filters, we need to add a fault filter for each. Since each route has a different match condition, the correct fault filter will automatically get selected.
5. If a route with a source match condition precedes a route without a source match condition, but has the same HTTP match conditions (header/url/authority, etc.), the latter must be ignored when supplying RDS and (LDS /faults) to the source.

cc @kyessenov @ostromart @ijsnellf @louiscryan


kyessenov:
The rules are too complicated IMHO. I think we all agree that faults must be per route in v1alpha3 APIs. We can have a window during which faults are only supported by v1alpha1, until we update envoy to bind fault filter config to RDS, and then we unhide faults from v1alpha3.

PR #406, opened 2018-03-13
Service definition for OOP adapter that will be on session plan.

guptasu:
As per the API design https://docs.google.com/document/d/1Fic1Mgp0W5Sxhd0BxAFOc82DbSYWTsYpi5P-pv8ZjSQ/edit


mixer/adapter/model/v1beta1/service.proto
```diff
@@ -0,0 +1,110 @@
...
+// This package also defines the types that are used to create Mixer templates.
+package istio.mixer.adapter.model.v1beta1;
+
+option go_package="istio.io/api/mixer/adapter/model/v1beta1";
+option cc_generic_services = true;
+
+import "google/protobuf/any.proto";
+import "google/rpc/status.proto";
+
+// `AdapterService` is implemented by backends that wants to provide telemetry and policy functionality to Mixer as an
```
geeknoid: Can we call this InfrastructureService instead?

These things are not adapters, they're just services.

mixer/adapter/model/v1beta1/service.proto
```diff
@@ -0,0 +1,110 @@
...
+// out of process adapter.
+//
+// `AdapterService` allows Mixer and the backends to have a session based model. In a session based model, Mixer passes
+// the relevant configuration state to the backend only once and estabilshes a session using a session identifier.
+// All future communications between Mixer and the backend uses the session identifier which refers to the previously
+// passed in configuration data.
+//
+// For a given `AdapterService`, Mixer calls the `Validate` function, followed by `CreateSession`. The `session_id`
+// returned from `CreateSession` is used to make subsequent request-time Handle calls and the eventual `CloseSession` function.
+// Mixer assumes that, given the `session_id`, the backend can retrive the necessary context to serve the
```
geeknoid: retrieve

mixer/adapter/model/v1beta1/service.proto
```diff
@@ -0,0 +1,110 @@
...
+    // `CreateSessionRequest` contains the adapter specific handler configuration and the inferred type information about
+    // the instances the handler would receive during request processing.
+    //
+    // `CreateSession` must return a `session_id` which Mixer uses to invoke template-specific Handle functions during
+    // request processing. If the backend couldn't create a session for a specific handler configuration and
+    // returns non S_OK status, Mixer will not make request-time Handle calls associated with that handler configuration.
+    // The `session_id` provides the Handle functions a way to retrive the necessary configuration associated with the session.
+    // Upon Mixer configuration change, Mixer will re-invoke `CreateSession` for all handler configurations whose existing
+    // sessions are invalidated or didn't existed.
+    rpc  CreateSession(CreateSessionRequest) returns (CreateSessionResponse);
```
geeknoid: We should mention that a backend is allowed to return the same session id if given the same config block. This would happen when multiple instances of Mixer in a deployment all create sessions with the same config.

mixer/adapter/model/v1beta1/service.proto
```diff
@@ -0,0 +1,110 @@
...
+
+    // Closes the session associated with the `session_id`. Mixer closes a session when its associated handler
+    // configuration or the instance configuration changes. Backend is suppose to cleanup all the resources associated
```
geeknoid: supposed

mixer/adapter/model/v1beta1/service.proto
geeknoid: Note that given CloseSession, reusing session_id assumes the backend is doing refcounting.

mixer/adapter/model/v1beta1/service.proto
guptasu: I agree they are not adapters. However, I feel Infrastructure is also not super helpful; but don't have any other suggestions. I think we can drop the suffix `Service`.

How about few options that intent do imply Istio integration backends.

1. InfrastructureIntegration ?
2. InfrastructureIntegrationBackend ?
3. IstioIntegration ?
4. IstioIntegrationBackend ?

I like #1 because it is uses the terms we commonly use around Istio and is yet generic enough to make back-ends feel they are not completely locked into Istio.

@mandarjog @douglas-reid for their opinions

geeknoid:
We've generally referred to these things as "infrastructure backends", so I think that would make a fine name. So I'll put my vote down for #5:

  InfrastructureBackend

mixer/adapter/model/v1beta1/service.proto
```diff
@@ -0,0 +1,117 @@
...
+    // handler configurations whose existing sessions are invalidated or didn't existed.
+    //
+    // Backend is allowed to return the same session id if given the same configuration block.
+    // This would happen when multiple instances of Mixer in a deployment all create sessions with the same configuration.
+    // Note that given individial instances of Mixer can call `CloseSession`, reusing `session_id` by the backend
+    // assumes that the backend is doing reference counting.
+    //
+    // If the backend couldn't create a session for a specific handler configuration and
+    // returns non S_OK status, Mixer will not make request-time Handle calls associated with that handler configuration.
+
```
geeknoid: Remove empty line.

mixer/adapter/model/v1beta1/service.proto
```diff
@@ -0,0 +1,117 @@
+// Copyright 2018 Istio Authors
```
geeknoid: Maybe rename the file?

guptasu:
@mandarjog over to you now for second lgtm. PTAL.

guptasu:
@mandarjog I am getting this in, since it is needed by my next PRs. Feel free to leave comment and I can address them in a separate PR. Thanks.

guptasu:
@douglas-reid I sneaked this in while you were out. Feel free to suggest improvements if you see anything that is not right. I can definitely use as many reviews as possible on this external API definition. @mandarjog you too :-). It is a lot easy to change stuff now than later.

PR #414, opened 2018-03-19
Gogo optimizations

kyessenov:


networking/v1alpha3/destination_rule.proto
```diff
@@ -15,9 +15,9 @@ syntax = "proto3";
 
 import "google/protobuf/duration.proto";
 
-package istio.networking.v1alpha3;
+import "gogoproto/gogo.proto";
 
-option go_package = "istio.io/api/networking/v1alpha3";
```
douglas-reid: removing this line will cause issues with the build

PR #417, opened 2018-03-20
Add placeholder for connection data inside handler configuration.

guptasu:
WDYT  ? There is a proposal in this PR to deprecate the existing way of specifying handler configuration. We can discuss over the PR if folks have other alternatives. I just took a stab at this problem.

NOTE: We will need some auth experts to guide us on what are the highly used auth mechanism that we will need to support.


guptasu:
@mandarjog @douglas-reid @geeknoid friendly ping!!

guptasu:
@mandarjog @geeknoid friendly ping

mixer/adapter/model/v1beta1/info.proto
```diff
@@ -37,4 +37,19 @@ message Info {
 
     // Base64 encoded proto descriptor of the adapter configuration.
     string config = 4;
+
+    // True if backend has implemented the
+    // [InfrastructureBackend](https://github.com/istio/api/blob/master/mixer/adapter/model/v1beta1/infrastructure_backend.proto)
+    // service; false otherwise.
+    //
+    // If true, during configuration time, Mixer would call the `InfrastructureBackend`' rpcs
```
geeknoid: Throughout this paragraph, remove all 'would' and use the present tense instead.

policy/v1beta1/cfg.proto
```diff
@@ -192,13 +192,53 @@ message Instance {
 // Handler allows the operator to configure a specific adapter implementation.
 // Each adapter implementation defines its own `params` proto.
 //
-// In the following example we define a `metrics` handler using the Mixer's prepackaged
-// prometheus adapter. This handler doesn't require any parameters.
+// In the following example we define a `metrics` handler for `prometheus` adapter.
```
geeknoid: for the prometheus

policy/v1beta1/cfg.proto
```diff
@@ -192,13 +192,53 @@ message Instance {
...
+// The example is in the form of kubernetes resource:
```
geeknoid: a Kubernetes

policy/v1beta1/cfg.proto
```diff
@@ -192,13 +192,53 @@ message Instance {
...
+// * The `metadata.name` is the name of the handler
+// * The `kind` refers to the adapter name
+// * The `spec` block represents adapter specific configuration as well as the connection information
```
geeknoid: adapter-specific

policy/v1beta1/cfg.proto
```diff
@@ -192,13 +192,53 @@ message Instance {
...
+//     metrics:
+//     - name: request_count
+//       instance_name: requestcount.metric.istio-system
+//       kind: COUNTER
+//       label_names:
+//       - source_service
+//       - source_version
+//       - destination_service
+//       - destination_version
+//   connection:
```
geeknoid: Can we make this more generic?

We talked before about needing some way to have  generate per-adapter config state. This would be the place to put such state. For example, we'd like to controlling logging parameters for each adapter. And we want to control fail open/close semantics for check adapters, etc.

policy/v1beta1/cfg.proto
```diff
@@ -210,4 +250,16 @@ message Handler {
   // Optional. Depends on adapter implementation. Struct representation of a
   // proto defined by the adapter implementation; this varies depending on the value of field `adapter`.
   google.protobuf.Struct params = 3;
+  // Optional. Information on how to connect to the out-of-process adapter.
+  // This is used if the adapter is not compiled into Mixer binary and is running as a separate process.
+  google.protobuf.Struct connection = 4;
```
geeknoid: Why is this a struct instead of a Connection message?

policy/v1beta1/cfg.proto
guptasu: D

policy/v1beta1/cfg.proto
guptasu: Chatted offline. Leaving connection parallel is fine. The additional information can be parallel to `connection` and `param` field, may be called `common` where logging, fail open/close configuration can be specified.

mandarjog:
Have we started adding `python` code too now ?

PR #420, opened 2018-03-20
Authentication policy v1alpha2.

diemtvu:
Changes compare to v1alpha1:
- "Origin" authentication is defined (and process) in similar way as peers.
- Credential rule match is removed. Logic to select credential to use is simplified as the first validated one in each peer/origin authentication methods list. 
- Add boolean flags to allow peer and/or origin authentication can fail (to delay rejection decision to the next layer, e.g authorization).
- Add flag for mTLS params to relax client certificate check.
- Define target selector message locally (instead of reuse networking/destination) to avoid cross API dependency.


authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,320 @@
...
+  // This flag is ignored if no authentication defined for origin (origins field is empty).
+  bool origin_is_optional = 5;
+
+  // Define whether peer or origin identity should be use for principal. Default
+  // value is USE_PEER.
+  // If peer (or orgin) identity is not available, either because of peer/origin
+  // authentication is not defined, or failed, principal will be left unset.
+  // In other words, binding rule does not affect the decision to accept or
+  // reject request.
+  PrincipalBinding principal_binding = 6;
```
liminw: I know we had this discussion back and forth. Do you think it is worthwhile to add a simple condition based on peer identity? For example, if peer is A, USE_ORIGIN; otherwise, USE_PEER.

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// TargetSelector defines a matching rule to a service/destination.
+message TargetSelector {
```
liminw: Can we apply target to a broader scope, like namespace?

diemtvu: If there is no selector, it matches to all service in the namespace the policy defined. 
As we don't have global scope (at least with the CRD based implementation), there is no other way (without violate ACL)

authentication/v1alpha2/policy.proto
diemtvu: Still need to and name to the 'method', then a rule for matching and few other concepts. IMHO, such feature can be somewhat supported with current authZ, so we don't lose functionality. Let's have this simple version done first and gather feedback to see if that option is needed.

liminw: You don't need to select authorization mechanisms (e.g., FB jwt, google JWT), just specify the peer identities in the condition. CredentialRule is still just setting principal, and nothing else. 

That's more close to RpcSecurityPolicy authentication policy. You are right that Authz can handle it. But since authentication policy already has principalBinding, let's handle principal binding more properly. 

If you want to go with simple version first, that's fine with me too.

PR #426, opened 2018-03-21
Proto definition for adapter info.

guptasu:
This proto should work when authored as a custom resource or when returned by backend via GetInfo (if we go that route in the future) rpc.

Idea is to use this proto to load the inventory yaml which has all the adapter specific data mixer needs.


guptasu:
@geeknoid @mandarjog @douglas-reid friendly ping

PR #428, opened 2018-03-22
Allow configuration of unknown outbound traffic handling in `MeshConfig`

loewenstein:
to be used in https://github.com/istio/istio/pull/4456

see also https://github.com/istio/istio/issues/3876

cc @rshriram @costinm


loewenstein:
/hold
/cc @rshriram @costinm @kyessenov 

Given https://github.com/envoyproxy/data-plane-api/issues/208#issuecomment-375081375, should we change this from boolean to list of ports envoy should fallback to passthrough?

loewenstein:
I was more concerned that with 
```
// [V2-API-DIFF] This is deprecated in v2, all Listeners will bind to their
// port. An additional filter chain must be created for every original
// destination port this listener may redirect to in v2, with the original
// port specified in the FilterChainMatch destination_port field.
google.protobuf.BoolValue bind_to_port = 1;
```
https://github.com/envoyproxy/data-plane-api/blob/b085af2c8485ba8525a128f466b1ec9ccf14574e/api/lds.proto#L146-L150 

we will no longer be able to provide the simple boolean switch with v2 in the long run. But I might be mistaken.

mesh/v1alpha1/config.proto
```diff
@@ -221,4 +221,9 @@ message MeshConfig {
   // DEPRECATED. Mixer address. This option will be removed soon. Please
   // use mixer_check and mixer_report.
   string mixer_address = 16;
+
+  // Flag to control pass through of traffic to unknown destinations.
+  // If set to true sidecar proxies will be configured to route traffic to unknown
+  // destinations to the original destination. Default is false.
```
costinm: I'm not sure about the name and description. What is falling back ? Why would a user care about filters ? What is this feature really doing (the name is more of an implementation description) ?

My understanding is that this will simply allow traffic to any destination that doesn't have a specific destination policy
or service. The opposite is to block any traffic that doesn't have a specific destination policy/service.

costinm: Also I doubt a user reading this will know what 'original destination' means, it's pretty internal term for envoy.

loewenstein: @costinm fair point. Will change naming and comment along @rshriram s proposal.

But first I would like to sort out whether and how to support a boolean vs. a port list property for v1/v2 (see my above comment).

loewenstein:
@kyessenov Given your comment https://github.com/envoyproxy/data-plane-api/issues/208#issuecomment-375081375 and what I have understood following the docs from there, in the foreseeable future there will be no way to simply handover all traffic to another listener. On the other hand you comment https://github.com/istio/api/pull/428#issuecomment-375396192 that either boolean or list of ports or even both would be ok.

I would be fine with either (or all three) given we can support that even when `FilterChainMatch` is implemented and the deprecated `Filter.use_original_dst` gets removed.

rshriram:
```
message DefaultOutboundRoutingPolicy {
enum OutboundRoutingPolicy {
 ALLOW_ANY // outbound traffic to unknown destinations will be allowed
 REGISTRY_ONLY // outbound traffic to services defined in the service registry or pre-defined external services will be allowed
 VIRTUAL_SERVICE_ONLY // outbound traffic will be restricted to destinations defined in VirtualServices only.
}

message PortList {
 repeated uint32 port;
}

oneof allow_policy {
  OutboundRoutingPolicy outbound_routing_policy;
}
} 
```

mesh/v1alpha1/config.proto
```diff
@@ -221,4 +221,23 @@ message MeshConfig {
   // DEPRECATED. Mixer address. This option will be removed soon. Please
   // use mixer_check and mixer_report.
   string mixer_address = 16;
+
+  message OutboundTrafficPolicy {
+    enum Mode {
+      // outbound traffic to services defined in the service registry or pre-defined external services will be allowed
```
rshriram: outbound traffic will be restricted to services defined in the service registry as well as those defined through ExternalServices

PR #432, opened 2018-03-24
Proto for local RBAC.

liminw:
This API is added for implementing Istio RBAC on Envoy.

Design doc is at:
https://docs.google.com/document/d/1XkN9ls-ZDISz2uXGN20RaIre63DkJJOJNUOzcKCxbWA


rbac/v1alpha1/rbac.proto
```diff
@@ -41,8 +41,8 @@ syntax = "proto3";
 //       namespace: istio-system
 //     spec:
 //       subject:
-//         user: request.auth.principal | ""
-//         groups: request.auth.principal | ""
+//         user: source.user | ""
```
yangminzhu: Why do we need to change the user to a different attribute, Is this part of the local RBAC change? Do we also need to change the corresponding setup yaml file?

rbac/v1alpha1/local_rbac.proto
```diff
@@ -0,0 +1,82 @@
...
+//    },
+//  ]
+message LocalRBAC {
+  // A list of ServicePolicy. One per service.
+  repeated ServicePolicy service_policies = 1;
+}
+
+// ServicePolicy specifies a service and the Istio RBAC policies that apply to the service.
+message ServicePolicy {
+  // The name of the service.
```
yangminzhu: Could you make it clear this would be the full qualified name?

rbac/v1alpha1/local_rbac.proto
```diff
@@ -0,0 +1,82 @@
...
+//         rules:
+//         - services: [“products.default.svc.cluster.local”]
+//           paths: [“/products/*”, “*/reviews”]
+//           methods: [“GET”, “HEAD”]
+//       bindings:
+//       - subjects:
+//         - user: “alice@yahoo.com”
+//         roleRef:
+//           kind: ServiceRole
+//           name: “service-admin”
```
yangminzhu: This should be "product-viewer"?

rbac/v1alpha1/local_rbac.proto
liminw: Updated comments.

rbac/v1alpha1/local_rbac.proto
liminw: Good catch! fixed.

rbac/v1alpha1/rbac.proto
liminw: This is not really related to local RBAC change. I just updated the example here to be consistent with our sample.

PR #434, opened 2018-03-27
Move a couple doc files to their correct location in istio.io.

geeknoid:


geeknoid:
Let's go with this for now, I'm working on a proposal to create a new site/location for developer-focused material.

PR #435, opened 2018-03-27
Comments update for Istio RBAC proto

liminw:


rbac/v1alpha1/rbac.proto
```diff
@@ -120,18 +120,19 @@ message AccessRule {
...
   repeated string paths = 2;
 
-  // Required. A list of HTTP methods (e.g., "GET", "POST") or gRPC methods.
-  // gRPC methods must be presented as fully-qualified name in the form of
-  // packageName.serviceName/methodName.
-  // If set to ["*"], it applies to any method.
+  // Optional. A list of HTTP methods (e.g., "GET", "POST").
+  // It is ignored in gRPC case because the value is always "POST".
+  // If set to ["*"] or not specified, it applies to any method.
   repeated string methods = 3;
```
yangminzhu: I think the 'Optional' only applies in gRPC case? In other cases this field is still required, maybe rephrase a little bit. Also I think if not specified, it defaults to deny the request, see [here](https://github.com/istio/istio/blob/692e481c0b1ae6c58ef8df4a8fb7f7a9b4a67bdf/mixer/adapter/rbac/rbacStore.go#L169).

liminw: I will relax the code to allow "method" to be optional. It will be confusing to say "optional" only applies to gRPC, not HTTP.

PR #436, opened 2018-03-27
Add RbacConfig to mesh/v1alpha1/config.proto.

yangminzhu:
See the design doc [here](https://docs.google.com/document/d/1XkN9ls-ZDISz2uXGN20RaIre63DkJJOJNUOzcKCxbWA/edit#).


mesh/v1alpha1/config.proto
```diff
@@ -243,4 +243,35 @@ message MeshConfig {
   // Users are strongly encouraged to use ExternalServices to explicitly declare any external dependencies,
   // instead of using allow_any.
   OutboundTrafficPolicy outbound_traffic_policy = 17;
+
+  // Istio RBAC config defines the global switch to control the Istio RBAC behavior.
```
liminw: "Istio RBAC config" -> IstioRBACConfig

mesh/v1alpha1/config.proto
```diff
@@ -243,4 +243,35 @@ message MeshConfig {
...
+    IstioRBACMode mode = 1;
+
+    // The list of services that are excluded from the mesh global Istio RBAC setting.
+    repeated string exception_services = 2;
+
+    // The list of namespaces that are excluded from the mesh global Istio RBAC setting.
+    repeated string exception_namespaces = 3;
+  }
+
+  // Istio RBAC config defines the global switch to control the Istio RBAC behavior.
```
liminw: Remove "Istio RBAC config". i.e., // Defines ...

mesh/v1alpha1/config.proto
```diff
@@ -243,4 +243,35 @@ message MeshConfig {
...
+      // Disable Istio RBAC for the mesh, except the services/namespaces listed
+      // in exception_services and exception_namespaces (whitelist).
+      OFF_WITH_EXCEPTION = 2;
+
+      // Enable Istio RBAC for the mesh, except the services/namespaces listed
+      // in exception_services and exception_namespaces (blacklist).
+      ON_WITH_EXCEPTION = 3;
+    }
+
+    // Istio RBAC configuration.
```
liminw: Istio RBAC configuration mode.

yangminzhu:
@geeknoid @xiaolanz Could you take a look if this is good to merge? Thank you!

mesh/v1alpha1/config.proto
```diff
@@ -243,4 +243,35 @@ message MeshConfig {
   // Users are strongly encouraged to use ExternalServices to explicitly declare any external dependencies,
   // instead of using allow_any.
   OutboundTrafficPolicy outbound_traffic_policy = 17;
+
+  // IstioRBACConfig defines the global switch to control the Istio RBAC behavior.
+  message IstioRBACConfig {
+    enum IstioRBACMode {
```
geeknoid: I don't think we should have the name "Istio" on all these types.

I know you're trying to make a distinction with k8s RBAC, but we have package names and other qualifiers already.

xiaolanz: Please following the naming style guide to use "RbacConfig" for proto messages. 
https://github.com/istio/api/blob/master/STYLE-GUIDE.md#naming

mesh/v1alpha1/config.proto
```diff
@@ -243,4 +243,35 @@ message MeshConfig {
...
+      // Enable Istio RBAC for the mesh, except the services/namespaces listed
+      // in exception_services and exception_namespaces (blacklist).
+      ON_WITH_EXCEPTION = 3;
+    }
+
+    // Istio RBAC configuration mode.
+    IstioRBACMode mode = 1;
+
+    // The list of services that are excluded from the mesh global Istio RBAC setting.
+    repeated string exception_services = 2;
```
xiaolanz: What does the "exclude" means? if this setting is really hierarchical, it should be in a separate CRD instead of mesh config.

mesh/v1alpha1/config.proto
xiaolanz: IstioRBACMode -> Mode since it's inside RbacConfig

yangminzhu: Done.

yangminzhu: Done.

mesh/v1alpha1/config.proto
yangminzhu: The "exclude" means that the RBAC policy won't enforce on the specified services/namespaces. I personally feel it's OK to be in the mesh config as it's part of the whole config, could you explain more about the rational that it should be a separate CRD?
+@liminw to see if it makes more sense to extract this to a separate CRD.

mesh/v1alpha1/config.proto
```diff
@@ -243,4 +243,35 @@ message MeshConfig {
...
+      // Enable Istio RBAC for the mesh.
+      ON = 1;
+
+      // Disable Istio RBAC for the mesh, except the services/namespaces listed
+      // in exception_services and exception_namespaces (whitelist).
+      OFF_WITH_EXCEPTION = 2;
+
+      // Enable Istio RBAC for the mesh, except the services/namespaces listed
+      // in exception_services and exception_namespaces (blacklist).
+      ON_WITH_EXCEPTION = 3;
```
wora: I recommend we don't use one field to change the semantics of another field.

repeated string enabled_services;
repeated string disabled_services;

is much cleaner, or you can use oneof {x, y}.

xiaolanz: For the same reason, this config should more look like a "service config" instead of "mesh config". It also aligns with how authentication policy is scoped right now. 

I suggest to create a CRD in istio.rbac.v1alpha1
message Mode {
      string service_name;
      bool enabled;
}

If a service needs RBAC on, it just sets this CRD with service X. If it's empty, it applies to all services in the namespace.

liminw: @xiaolanz, I don't want to introduce another type of CRD just for enabling RBAC. Also my concern with the approach is that there will be too many CRD objects. Authentication policy is different. It is essentially enabled on the entire mesh. If an authentication policy is defined on a scope (namespace/service/port), it is always honored.

I am ok with @wora's suggestion of introducing 

oneof {
repeated string enabled_services;
repeated string disabled_services;
repeated string enabled_namespaces;
repeated string disabled_namespaces;
}

xiaolanz: The above oneof basically says that you cannot "enable a service" and "disable a namespace" at the same time. Is it what you need? 

By looking at the original proto, I thought the intention is to allow some hierarchy: like enabling a namespace, while disabling for a service within the namespace. or vise versa. 

Creating a new CRD type should not be a problem. Also for the common usecase, I don't think there will be too many CRD resources: like one CRD resource per namespace?  This design also aligns well with the design of service config, which I plan to include the rbac mode.

yangminzhu: @wora  I updated to use whitelist/blacklist to make this more clear.
@xiaolanz I think we won't want to support the "enable"/"disable" case at the same time, also, I updated to make it more clear that a whitelist and a blacklist couldn't be used at the same time as we found it's confusing and not very useful in reality.

liminw: @xiaolanz For the CRD approach, I am just worried about user experience. Suppose I need to enable RBAC on ten services, then I need to create ten CRDs for that. That is certainly not user friendly. Since we will anyway use some client tool to process user facing configurations, I prefer to have an aggregated configuration.

yangminzhu: @xiaolanz 
I combined the Whitelist and Blacklist but kept the naming, I feel the mode and enabled_with_exclusion/disabled_with_exclusion is a little confusing.

xiaolanz:
Suggest to bring the Mode enum back so it's clear when the exclusion/inclusion list should be used.

PR #440, opened 2018-03-29
Add the allow_missing_or_failed option in jwt authn

lei-tang:
Add the allow_missing_or_failed option from https://github.com/envoyproxy/data-plane-api/blob/11c750628bcaeeeca3f72f89fe5c99921aa727b1/envoy/config/filter/http/jwt_authn/v2alpha/config.proto.


mixer/v1/config/client/auth.proto
```diff
@@ -157,6 +157,12 @@ message EndUserAuthenticationPolicySpec {
   // JWT validation is skipped if the user's traffic request does not
   // include a JWT.
   repeated JWT jwts = 2;
+  // From
+  // https://github.com/envoyproxy/data-plane-api/blob/11c750628bcaeeeca3f72f89fe5c99921aa727b1/envoy/config/filter/http/jwt_authn/v2alpha/config.proto
+  // If true, the request is allowed if JWT is missing or JWT verification
+  // fails. Default is false, a request without JWT or failed JWT verification
+  // is not allowed.
```
wora: If the request is allowed, how does backend detect the reason? Does this allow show up in monitoring system in a special way?

If a request is denied, what should be the behavior, 401 error? What about error message and other information?

diemtvu: In Istio, we will place another filter (authn) behind this, which will check the policy and reject request accordingly (401 error). At the moment, we don't report the exact failure reason, but it should be possible to add.

Btw, this PR is not needed as we already switch to the upstream API, which already has such flag.

PR #442, opened 2018-04-02
Add `name` back to destination proto.

guptasu:


kyessenov:
Is this due to the deprecation policy?

PR #444, opened 2018-04-03
Define transportProxy field into MeshConfig

rlenglet:
Define boolean `transportProxy` field in `MeshConfig` to enable transparent proxying using TPROXY in all sidecar proxies, mesh-wide.

Implemented in and required by https://github.com/istio/istio/pull/4654.

Signed-off-by: Romain Lenglet <romain@covalent.io>


andraxylia:
@rlenglet please sign the CLA.

rlenglet:
@andraxylia I got the CLA signed.

andraxylia:
@sebastienvas how can we re-ran the bot that checks for the CLA, or otherwise check manually?

ldemailly:
@rlenglet can you type just
I signed it!
and nothing else in a comment
also you got some conflicts

ldemailly:
andra, him not you :-)

rlenglet:
I signed it!

mesh/v1alpha1/config.proto
```diff
@@ -283,4 +283,7 @@ message MeshConfig {
 
   // The global config to control Istio RBAC behavior.
   RbacConfig rbac_config = 18;
+
+  // If set to true, run Envoy as a transparent proxy and with the `CAP_NET_ADMIN` capability in sidecars.
+  bool transparent_proxy = 19;
```
rshriram: you might want to add more info here (about use of TPROXY vs REDIRECT and the benefits). This stuff goes in the docs and is visible. So its helpful to provide context.
I would also change this to a oneOf.. Something like this:
```
enum TrafficInterceptionMode {
 // Use destination nat, and IP tables redirect target to route all inbound/outbound traffic to the sidecar
 // Source address will be lost during redirection.
 REDIRECT = 0;
// Use IPtables TPROXY mode to route all inbound/outbound traffic to the sidecar. Source address will
// be preserved in the sidecar and allows for advanced filtering and manipulation. Note: use of TPROXY
// will require the sidecars to have CAP_NET_ADMIN capability.
 TPROXY = 1;
}

TrafficInterceptionMode interception_mode = 19;
```

rlenglet: I'll take your suggestion, I like it. I agree this needed more documentation.

costinm:
See my comments in 4654 - I think it's better to drop the API change and just enable it by default in proxyv2, with an option for user to switch to a different image if they can't do NET_ADMIN.

rshriram:
Don’t you think that asking people to increase their proxy container privileges is an issue ? A proxy that gets compromised will have more privileges in this case than the default one that simply does redirect

rlenglet:
@costinm @rshriram Would you be OK to move this field into `ProxyConfig`?

mesh/v1alpha1/config.proto
```diff
@@ -115,6 +115,17 @@ message ProxyConfig {
...
+  // The mode used to redirect inbound traffic to Envoy.
+  // This setting has no effect on outbound traffic: iptables REDIRECT is always used for
+  // outbound connections.
+  // If "redirect", use iptables REDIRECT to NAT and redirect to Envoy. The "redirect" mode
+  // loses source addresses during redirection.
+  // If "tproxy", use iptables TPROXY to redirect to Envoy. The "tproxy" mode preserves both
+  // the source and destination IP addresses and ports, so that they can be used for advanced
+  // filtering and manipulation. The "tproxy" mode also configures the sidecar to run with the
+  // CAP_NET_ADMIN capability, which is required to use TPROXY.
+  string interception_mode = 18;
```
wora: Is this really free text instead of enum?

rlenglet: @wora Yes, as I commented earlier:

> I changed this design to add an interception_mode field into ProxyConfig instead.
> I decided to use a string instead of an enum, because I believe it would have required users to set integer values in config maps.
> If we wanted to be strict, I could use Lyft's protoc-gen-validate to ensure that only "tproxy" and "redirect" are accepted as valid values.

rlenglet: OK, seems like I was wrong. We could use the enum values instead of integers. I'll change to a proper enum.

rshriram:
@rlenglet you can specify the enum names. We parse the config as a JSON input, and there is some proto/json magic that parses ENUMs properly.

rshriram:
@rlenglet proxy config is fine, as long as the outbound listeners we generate in Pilot remains the same. If you need to know what mode is being used by the proxy, you might either need to encode it in the sidecar ID (or pass it as node metadata).

rshriram:
@costinm we could still do the two image thing you suggest, and I prefer that as well.
But its better to have a config option present so that we can easily use it in testing, without having to switch between two images or have two test variants using two different proxy images.

PR #445, opened 2018-04-09
networking: clarify semantics of ports, hostnames, and general cleanup

rshriram:
cc @ZackButcher @louiscryan

Few assorted changes:
1. Due to some implementation difficulties, and the ambiguity caused by
VirtualServices with multiple services in the hosts, lets stick to using
just the port number for matching destination port. Its more universal and
easily understood.
2. Use a consistent FQDN throughout docs, and have a separate section under Destination, explaining how shortnames will expand to FQDNs in Kubernetes.
3. Clean up references to use of IPs in destination.hosts since we now require people to declare
ExternalServices (i.e. Destination should always be a service in the registry). Added some text to clarify that as well.


networking/v1alpha3/virtual_service.proto
```diff
@@ -119,25 +113,25 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
   //
-  // For example on Kubernetes, when hosts contains a short name, Istio will
-  // interpret the short name based on the namespace of the rule. Thus, when a
-  // client namespace applies a rule in the "default" namespace containing a name
-  // "reviews, Istio will setup routes to the "reviews.default.svc.cluster.local"
-  // service. However, if a different name such as "reviews.sales.svc.cluster.local"
-  // is used, it would be treated as a FQDN during virtual host matching.
-  // In Consul, a plain service name would be resolved to the FQDN
-  // "reviews.service.consul".
+  // *Note for Kubernetes users*: When short names are used (i.e. single
```
ZackButcher: What about:
> When short names are used (e.g. "reviews" instead of "reviews.default.svc.cluster.local"), Istio will interpret...

I think the example is more illuminating in this case (what does "single word" mean in the context of a domain name?).

networking/v1alpha3/virtual_service.proto
```diff
@@ -119,25 +113,25 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
+  // word), Istio will interpret the short name based on the namespace of
+  // the rule, not the service. A rule in the "default" namespace
+  // containing a name "reviews will be interpreted as
```
ZackButcher: Here and in a few other places, should this read:

> containing a host  "reviews"

rather than "name", given we've renamed the field from `names` to `hosts`?

networking/v1alpha3/virtual_service.proto
```diff
@@ -231,26 +220,46 @@ message VirtualService {
 //           version: v2
 //
 // The following VirtualService sets a timeout of 5s for all calls to
-// productpage.prod service. Notice that there are no subsets defined in
-// this rule. Istio will fetch all instances of productpage.prod service
-// from the service registry and populate the sidecar's load balancing
-// pool.
+// productpage.prod.svc.cluster.local service in Kubernetes. Notice that
+// there are no subsets defined in this rule. Istio will fetch all
+// instances of productpage.com service from the service registry and
```
ZackButcher: productpage.prod

networking/v1alpha3/virtual_service.proto
```diff
@@ -231,26 +220,46 @@ message VirtualService {
...
+// populate the sidecar's load balancing pool.  Also, notice that this rule
+// is set in istio-system namespace, while using the FQDN of the
+// productpage service. Hence, the rule's namespace does not have an
+// impact.
```
ZackButcher: does not have an impact _? on name resolution ?

What about

> Also, notice that this rule is set in the istio-system namespace but uses the fully qualified domain name of the productpage service, productpage.prod.svc.cluster.local. Therefore the rule's namespace does not have an impact in resolving the name of the productpage service.

networking/v1alpha3/virtual_service.proto
```diff
@@ -267,39 +276,37 @@ message VirtualService {
 //
 message Destination {
   // REQUIRED. The host can be a short name or a fully qualified domain
-  // name from the service registry, a resolvable DNS name, or an IP
-  // address.
+  // name of a service or a CIDR prefix from the service registry. CIDR
+  // prefixes can be used only if there is a corresponding ExternalService
+  // declaration that uses the same CIDR prefix in the
+  // ExternalService.hosts. Traffic forwarded to destinations that are not
```
ZackButcher: CIDR addresses should be usable anywhere in the mesh, not just in ExternalService, right? E.g. I can give a CIDR as a hostname for a VirtualService, right? Put another way, just because the service is addressed by IP range doesn't mean its external to the mesh (e.g. thinking about VM use case).

ZackButcher:
> Clean up references to use of IPs in destination.hosts since we now require people to declare
ExternalServices (i.e. Destination should always be a service in the registry). Added some text to clarify that as well.

Oh, when did we decide this? I'm not sure I buy it, given non-k8s use cases where IP blocks are more common.

networking/v1alpha3/virtual_service.proto
rshriram: ack

networking/v1alpha3/virtual_service.proto
rshriram: Per Costin, even VMs are being planned to be modeled as ExternalServices, though I didn't fully grok why.

That said, what I wanted to convey here was that the name MUST be an entry in the service registry. Since by default, service registries only have alpha-numeric names, the only other way for you to add a non-alphanumeric (i.e. CIDR) name is through externalService - which ends up in the service registry as well.

By imposing this restriction, we prevent people from creating arbitrary clusters via virtual services. The Destination spec here is insufficient to describe the critical cluster properties (cluster resolution - eds/dns/original_dst IP).

rshriram:
>>Clean up references to use of IPs in destination.hosts since we now require people to declare
ExternalServices (i.e. Destination should always be a service in the registry). Added some text to clarify that as well.

>Oh, when did we decide this? I'm not sure I buy it, given non-k8s use cases where IP blocks are more common

Repeating what I said above: the thought here is that clusters can only be formed from services in the registry and from subsets declared through destination rules. You can add additional services to Istio's registry using External Service specs - whose hosts field can contain both names and CIDRs. But any name given here has to match either an entry in the registry or a host from external services.

ldemailly:
can you either revert this or make the corresponding istio/istio changes ?

going to sync up vendor api to previous commit for now/in the meantime

ldemailly:
it's actually strange I don't see the compilation breaking change scanning the diff

networking/v1alpha3/virtual_service.proto
```diff
@@ -472,29 +481,24 @@ message HTTPMatchRequest {
...
   // only expose a single port or label ports with the protocols they support,
-  // in these cases it is not required to explicitly select the port. Note that
-  // selection priority is to first match by name and then match by number.
-  //
-  // Names must comply with DNS label syntax (rfc1035) and therefore cannot
-  // collide with numbers. If there are multiple ports on a service with the
-  // same protocol the names should be of the form <protocol-name>-<DNS label>.
-  PortSelector port = 6;
+  // in these cases it is not required to explicitly select the port.
+  uint32 port = 6;
```
ldemailly: maybe just revert this 1 change for now ?

PR #448, opened 2018-04-12
Add a note to mesh config to stop it from expanding

kyessenov:
Mesh config currently serves as a global dump of random features. We should stop it from growing any further. The plan forward is to define CRDs for individual aspects of the mesh, such as RBAC, tracing, authn/z, instead of a single CR that is globally scoped. In particular, this config type needs an owner since it does not completely fit into any WG.


kyessenov:
Ping.

xiaolanz:
Mesh config continues to exist as a global singleton, complimentary to service config (see the overall configure architecture plan [link](https://docs.google.com/document/d/1rMb7Qr2_QzRXdLXNB_R9fRyVUJgIfWZDCzQm8Bu0ID4/edit)  We will be careful of what we plan to put in mesh (basically a small set of runtime config that is mesh global and non-hierarchical).

mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
```diff
@@ -51,6 +51,10 @@ <h2 id="MeshConfig">MeshConfig</h2>
 <p>MeshConfig defines mesh-wide variables shared by all Envoy instances in the
 Istio service mesh.</p>
 
+<p>NOTE: This configuration type is going to be deprecated in the near future and
```
xiaolanz: Please update the wording. Instead of deprecating, we will only allow putting mesh global singletons here.

kyessenov:
Updated wording.

kyessenov:
ping the approvers.

PR #452, opened 2018-04-16
Add ClientSetings to authentication policy

diemtvu:
Issue: https://github.com/istio/istio/issues/4861


diemtvu:
We decided to use destination rule to configure clients. This is no longer needed.

PR #456, opened 2018-04-19
Remove EUC policy API

diemtvu:
Issue: https://github.com/istio/istio/issues/4744


ayj:
Should the remaining references in istio/istio be deleted first?

```bash
$ grep -rn EndUserAuthenticationPolicySpec pilot/ |wc -l 
38 
```

qiwzhang:
It is still used here https://github.com/istio/proxy/blob/master/src/envoy/http/mixer/filter_factory.cc#L36

diemtvu:
Thanks for checking. These refs have been removed in recent PRs https://github.com/istio/istio/pull/5080 and https://github.com/istio/proxy/pull/1477. We should be good now. PTAL

PR #457, opened 2018-04-23
Add RbacConfig to rbac.proto as a CRD to control Istio RBAC behavior.

yangminzhu:
The second step of istio/istio#5033.


rbac/v1alpha1/rbac.proto
```diff
@@ -198,6 +198,21 @@ message RoleRef {
...
 // RbacConfig defines the global config to control Istio RBAC behavior.
+// This CRD is a singleton where only one CR should be created globally in the mesh and the
+// namespace should be the same to other Istio components, which usually is istio-system.
+// Below is an example of RbacConfig object "istio-rbac-config" which enables Istio RBAC for all
+// services in the default namespace.
+//
+//     apiVersion: "config.istio.io/v1alpha1"
+//     kind: RbacConfig
+//     metadata:
+//       name: istio-rbac-config
```
xiaolanz: what do we do if there is more than one CR defined for this CRD?

rbac/v1alpha1/rbac.proto
```diff
@@ -196,3 +196,40 @@ message RoleRef {
   // object.
   string name = 2;
 }
+
+// RbacConfig defines the global config to control Istio RBAC behavior.
```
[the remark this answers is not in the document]
yangminzhu: Thanks for the remind, Added more comments. We'll also provide a sample yaml for this and in most cases the user only needs to make changes to the sample file.

rbac/v1alpha1/rbac.proto
yangminzhu: I'm not sure if we can enforce the singleton restriction here, any ideas? In the code side, we can  prompt the user that there is already an existing instance of this CRD and ask the user to either change or delete the old one if the user is using istioctl (which is the recommended way to interact with Istio RBAC.

xiaolanz: We prefer server validation vs client side validation.

yangminzhu: Yeah, we'll sure also have the server side validation, we'll just reject the new config if we found any existing one and log this clearly.

xiaolanz: would you please state that in the comment?

yangminzhu: Done.

yangminzhu:
@xiaolanz Would you mind to take a second look at this? Thank you.

rbac/v1alpha1/rbac.proto
```diff
@@ -196,3 +196,58 @@ message RoleRef {
...
+// This CRD is a singleton where only one CR should be created globally in the mesh and the
```
liminw: What does CR stand for? CRD?

rbac/v1alpha1/rbac.proto
liminw: I can see that enforcing a global singleton CRD is difficult. Shall we use a fixed name? 

@mandarjog @kyessenov, any suggestion on that?

rbac/v1alpha1/rbac.proto
yangminzhu: Updated comments.

yangminzhu:
@kyessenov Kuat, It seems @xiaolanz's approval doesn't grant merge permission to this PR, could you take a look and give your approval? Thank you!

rbac/v1alpha1/rbac.proto
xiaolanz: Mesh config is intended to hold these singleton today so that's why this is originally created as part of mesh config.  It's ok to move it into CRD because I plan to convert mesh config into CRD as well. It's ok for having it namespaced in 'istio-system' since mixer uses this scope for mesh global.

rbac/v1alpha1/rbac.proto
```diff
@@ -196,3 +196,58 @@ message RoleRef {
   // object.
   string name = 2;
 }
+
+// RbacConfig defines the global config to control Istio RBAC behavior.
+// This Custom Resource Definition is a singleton where only one Custom Resource should be created globally in
```
xiaolanz: s/Custom Resource Definition/Custom Resource

They are different concept. CRD is type. CR is resource (objects). Type is always a cluster scope unique. What we want is that the resource is a singleton.

yangminzhu: Done.

PR #458, opened 2018-04-24
Change MutualParams to support more mode.

incfly:
For background see https://github.com/istio/istio/issues/5059


authentication/v1alpha1/policy.proto
```diff
@@ -22,14 +22,33 @@ package istio.authentication.v1alpha1;
 
 option go_package = "istio.io/api/authentication/v1alpha1";
 
-// mTLS authentication params.
+// TLS authentication params.
+// TODO: consider to give it a better name, maybe TlsParams?
```
ayj: Any plans (issue#?) to reconcile authentication TLS options with [networking's gateway](https://github.com/istio/api/blob/98683419985a2d75509cbc9f67aec73573515236/networking/v1alpha3/gateway.proto#L225)? See

incfly: Gateway's config does seem have each one corresponding mode. But I don't have much context to tell if they should reconcile. If we do need to share the proto message, should that live in `policy.proto` or somewhere else?

@diemtvu wdty?

diemtvu: Click the merge button by accident :). Anyway, to answer Jason question, we start gathering ideas https://docs.google.com/document/d/1I_kAd68hcmRuKm-PGQganprc3elM5nDXZNFT_CZEQc4/edit (the scope of this work is also getting deeper)

PR #459, opened 2018-04-24
Rename ExternalService to ServiceEntry

rshriram:
Renaming the resource so that it can be used for Mesh Expansion use cases as well.


networking/v1alpha3/service_entry.proto
```diff
@@ -198,10 +211,10 @@ message ExternalService {
 
   // Service discovery mode for the hosts. If not set, Istio will attempt
   // to infer the discovery mode based on the value of hosts and endpoints.
-  Discovery discovery = 3;
```
xiaolanz: we should never renumber the fields. but it's a break changing anyway. I assume you know that the existing CRs will stop to work.

rshriram: Yes.. Trying to do this before the 0.8 LTS so that we don't have to live with ExternalService for MeshExpansion. Technically, this entire folder is not even released yet. So, capitalizing on that

louiscryan:
@ZackButcher or @costinm you should weigh in but I think this is fine

costinm:
Absolutely fine.

My choice was "Destination" - to avoid confusion with Service and VirtualService. 
I'm also a bit concerned with 'Entry' - since it can be multiple endpoints or DNS names. 

But almost anything is better than ExternalService.

costinm:
Also tactical: can you cut&paste ServiceEntry and call it "ExternalService"  - to avoid breaking any code when this get merged.

After code in istio/istio is updated we can remove ExternalService

linsun:
I agree external service is confusing in the vm expansion scenario or services running in consul or eureka.  I'm however concerned with serviceEntry naming too, because I'm having trouble to conceptually map this to my service in that mesh (e.g. guestbook) that needs to connect to a cloud service like AWS or Google or IBM Watson services.  In this case, it is not really a service entry but going out to use/consume a cloud service.  

Maybe destination like @costin suggested or DestinationService?

costinm:
DestinationService may be confusing with VirtualService. Already have to deal with the confusion with k8s Service.  

Destination works well with DestintionRule (which customizes the Destination with subsets, etc), and the attribute 'destination' in the VirtualService.

The main problem is we'll need to rename the (non-user-visible) Destination message inside VirtualService, for example to DestinationRef - it points to an ExternalService or DestinationRule.

mandarjog:
This is a much better name than before, but why not `ServiceRegistryEntry` ?
That way the name of the message conveys the correct info without reading the description.

rshriram:
Some folks felt it was one too many keystrokes 😀 . I have no issues for that though. In fact, the docs state that service entry is a short form for service registry entry

rshriram:
```
shriramr@shriramr-a01:~/go/src/istio.io/api $ echo "ServiceRegistryEntry" |wc -c
      21
shriramr@shriramr-a01:~/go/src/istio.io/api $ echo "VirtualService" |wc -c
      15
shriramr@shriramr-a01:~/go/src/istio.io/api $ echo "DestinationRule" |wc -c
      16
shriramr@shriramr-a01:~/go/src/istio.io/api $ echo "ExternalService" |wc -c
      16
```

mandarjog:
keystrokes .... :-)

I think 
"For mesh expansion create a `ServiceRegistryEntry` for the vm node"   reads clearer than 
"For mesh expansion create a `ServiceEntry` for the vm node"

mandarjog:
@louiscryan I think we should not reject `ServiceRegistryEntry` on account of number of keystrokes. The name is clearer and needs to explanation when used in a sentence.

networking/v1alpha3/service_entry.proto
```diff
@@ -158,24 +136,50 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
   // REQUIRED. The ports associated with the external service.
-  repeated Port ports = 2;
+  repeated Port ports = 3;
+
+  // Set this field to true to indicate that this service is external to
+  // the mesh. When communicating with services outside the mesh, Istio's
+  // mTLS authentication is disabled when communicating with services
+  // outside the mesh. Policy enforcements are performed on the client-side
+  // as opposed to server-side.
+  bool mesh_external = 4;
```
frankbu: Isn't external service the most common use case? If so, shouldn't this field default be `true` (or rename the field to `mesh_internal`)?

frankbu:
@mandarjog I agree that `ServiceRegistryEntry` may be slightly clearer, but the name really seems klunky to me, so I don't think the tradeoff is worth it. The way @rshriram has it currently documented, makes `ServiceEntry` a fine name IMO, but I can live with either name, so this is just my 2 cents.

networking/v1alpha3/service_entry.proto
```diff
@@ -0,0 +1,244 @@
...
+
+    // Signifies that the service is part of the mesh. Typically used to
+    // indicate services added explicitly as part of expanding the service
+    // mesh to include unmanaged infrastructure (e.g., VMs added to a
+    // Kubernetes based service mesh).
+    MESH_INTERNAL = 1;
+  };
+
+  // Specify whether the service should be considered external to the mesh
+  // or part of the mesh.
```
frankbu: If not REQUIRED, should specify the default is MESH_EXTERNAL.

costinm:
+1 on ServiceEntry

Shriram - do you want to make the change the hard way or the easy way ? I.e. keep the old proto so
we don't break istio/istio, and remove it after migration ? Or just do the update with the vendor update ?

rshriram:
The thought is to do it in one shot.. This is literally a search and replace. Given that the vendor update will involve changing all the other references (and that code cannot be functioning sensibly with both old and new names in place), I vote for the hard way. We already have a PR out there for this, thanks to Greg.

GregHanson:
istio/istio PR related to api change: https://github.com/istio/istio/pull/5195

currently blocked until PR https://github.com/istio/istio/pull/5194 goes through

PR #460, opened 2018-04-25
Update global dictionary for new attributes.

JimmyCYJ:


qiwzhang:
global dictionary changes can only be appended at the end.

JimmyCYJ:
I have moved all new attributes to the bottom of this file to make it backward compatible. Please take a look. Thanks.

PR #461, opened 2018-04-26
Add request.auth.claims and source.principal

lei-tang:
Add the attributes: request.auth.claims and source.principal


mixer/v1/global_dictionary.yaml
```diff
@@ -44,7 +44,9 @@
 - source.uid
 - source.namespace
 - source.labels
+# source.user is to be deprecated, please refer to the attribute source.principal
 - source.user
+- source.principal
```
qiwzhang: All new attributes have to be appended at the end.

PR #465, opened 2018-04-27
Add context.internal attribute

kyessenov:
`context.internal` attribute should be set only for policy & telemetry calls initiated within istio control plane proper. It is useful to separate application and istio internal telemetry, as existing adapters are tailored for the application-level metrics.

edit: most -> existing
cc @theganyo


douglas-reid:
I'm not sure that I would go so far as to make a claim like "most adapters are not interested in internal telemetry". This was a big request from some interested parties.

That being said, having an easy way to distinguish `istio-system` telemetry from other telemetry seems useful.

Can we use `*.namespace == istio-system` instead?  If not, is `context.internal` the right way to model this concept?  Will `internal` be meaningful for operators?

kyessenov:
There is a need to taint all internal calls to prevent recursion. The taint
needs to be carried through hops. This is just a reflection of that idea.
We can treat the system namespace as the taint, that is fine with me.

kyessenov:
I wonder if this should be also reflected in the tracing context. We cannot assume that mixer back-ends carry this bit through, unless we require some form of distributed tracing. Since we do not yet consume the tracing data in mixer, it might be pre-mature to add this bit just yet.

kyessenov:
PTAL. Given the config extension https://github.com/istio/api/pull/502, we can inject attributes that mixer sidecar sends to mixer (which are not the same as payload attributes that mixer sidecar forwards to mixer). One of the attributes can be `context.internal`, in addition to `source.uid`.

geeknoid:
I understand the concept, but I think context.internal is not descriptive. Can we come up with a name that is evocative of what the attribute actually represents?

kyessenov:
Here are a bunch of ideas for boolean attributes:

- `context.reporter.internal`
- `context.reporter.self`
- `context.self`
- `context.meta.internal`
- `context.self_monitoring`

I'm not good at naming, so suggestions are very much welcome.

kyessenov:
@qiwzhang The idea is to mark Check and Report calls that are initiated from mixer/pilot sidecars, so that we stop recursive calls that may happen if keep Reporting on Report.

qiwzhang:
When the proxy in front of mixer sending a report,  it should directly to the mixer, doesn't it?   Are you saying that it will go to the proxy again?

kyessenov:
Yes, it does. However, mixer can have downstream systems that may themselves call Report. You can see how a loop can be formed.

douglas-reid:
Is this just for Mixer-to-Mixer calls right? Are we also proposing this as a way of filtering out Pilot telemetry?

For the more general case, maybe `context.mesh_component` with a range of values `[pilot, mixer, citadel, proxy, application]` ?

For loop prevention: do we want something like `origin.service == 'mixer'` ?

douglas-reid:
Is this still a need?

kyessenov:
Depends on how we plan to provide telemetry for OOP adapters. Not an issue if we don't use istio-on-istio approach.

PR #466, opened 2018-04-30
Define policy-driven route directive

kyessenov:
Specifies the routing effects template, to be filled by mixer adapters.

Sending it to release-0.8 branch, since master has diverged and the diff is very large.


kyessenov:
cc @qiwzhang : this is the sort of things we want to be able to perform form within mixerfilter.

mixer/v1/check.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Istio Authors
+// Copyright 2018 Istio Authors
```
geeknoid: Copyright dates should reflect date of creation, not last mod.

mixer/v1/check.proto
```diff
@@ -82,6 +83,10 @@ message CheckResponse {
     // The total set of attributes that were used in producing the result
     // along with matching conditions.
     ReferencedAttributes referenced_attributes = 5 [(gogoproto.nullable) = false];
+
+    // An optional routing action, used to override the default route action
+    // when all pre-conditions are satisfied.
+    RouteAction action = 6;
```
geeknoid: Given this, I think we should delete the "attributes" field from this proto, we'll never use it. Attributes are strictly inputs to Mixer, not an output.

mixer/v1/check.proto
```diff
@@ -149,3 +154,45 @@ message ReferencedAttributes {
...
+  bool remove = 3;
+
+  // If true and remove is not set, the value is appended instead of replacing it.
+  // This operation is not applicable to the special headers, such as "scheme".
+  bool append = 4;
+}
+
+// Expresses the actions to be performed by the proxies in response
+// to a successful pre-condition check for an HTTP request.
+message RouteAction {
```
geeknoid: I'm not happy with the name RouteAction:

- We use "action" in Mixer's input config model, so it's unwise to use the same term in a completely different context for the output.

- This is more than just routing control, since we allow general-purpose header mutation.

How about instead of RouteAction we call this ClientDirective? Tells the client what to do.

mixer/v1/check.proto
```diff
@@ -149,3 +154,45 @@ message ReferencedAttributes {
...
+  // This operation is not applicagble to the special headers, such as "scheme".
+  bool remove = 3;
+
+  // If true and remove is not set, the value is appended instead of replacing it.
+  // This operation is not applicable to the special headers, such as "scheme".
+  bool append = 4;
+}
+
+// Expresses the actions to be performed by the proxies in response
+// to a successful pre-condition check for an HTTP request.
```
geeknoid: Should this be constrained to HTTP? Seems like this could be used also when establishing a TCP connection, right?

mixer/v1/check.proto
```diff
@@ -149,3 +154,45 @@ message ReferencedAttributes {
...
+message RouteAction {
+  // Operations on the request headers.
+  repeated HeaderOperation request_headers = 1 [(gogoproto.nullable) = false];
+
+  // Operations on the response headers.
+  repeated HeaderOperation response_headers = 2 [(gogoproto.nullable) = false];
+
+  // If set, enables a direct response without proxying the request to the routing
+  // destination.
+  uint32 response_code = 3;
```
geeknoid: In the case of an error condition, setting this is logically no different than failing the precondition check, right? That's a little confusing then since that gives two ways to do the same thing.

mixer/v1/check.proto
kyessenov: Fixed.

mixer/v1/check.proto
kyessenov: Removed and reserved.

mixer/v1/check.proto
kyessenov: Renamed to `Directive`. I hesitate to use client in the term, since the directive can be executed by a middle-proxy.

mixer/v1/check.proto
kyessenov: The spec can be expanded to TCP. So far it's restricted to HTTP, since that's where most of the value is. I don't think there is compatibility concerns to add more TCP-specific fields.

mixer/v1/check.proto
kyessenov: Yes, but we also want to support 3xx codes (redirects). That's not an error condition. I understand that it allows check to pass, and then return an error. Although, we could invalidate this code for 4xx or 5xx error codes.

mixer/v1/check.proto
```diff
@@ -71,17 +72,15 @@ message CheckResponse {
...
-    // ferry new attributes that Mixer derived based on the input set of
-    // attributes and its configuration.
-    CompressedAttributes attributes = 4 [(gogoproto.nullable) = false];
+    reserved 4;
 
     // The total set of attributes that were used in producing the result
     // along with matching conditions.
     ReferencedAttributes referenced_attributes = 5 [(gogoproto.nullable) = false];
+
+    // An optional directive, used to override the default route action
```
geeknoid: not just route action, right?

mixer/v1/check.proto
```diff
@@ -149,3 +148,45 @@ message ReferencedAttributes {
...
+message HeaderOperation {
+  // Header name.
+  string key = 1;
+
+  // Header value.
+  string value = 2;
+
+  // If true, the header is removed.
+  // This operation is not applicagble to the special headers, such as "scheme".
+  bool remove = 3;
```
geeknoid: It'd probably be cleaner and more extensible if "remove" and "append" were replaced by an enum instead?

enum {Remove, Add, Append}

mixer/v1/check.proto
geeknoid: Should we then say in the comment that only values in the 200 or 300 range are allowed?

mixer/v1/check.proto
```diff
@@ -149,3 +148,45 @@ message ReferencedAttributes {
...
+
+  // If true and remove is not set, the value is appended instead of replacing it.
+  // This operation is not applicable to the special headers, such as "scheme".
+  bool append = 4;
+}
+
+// Expresses the actions to be performed by the proxies in response
+// to a successful pre-condition check for an HTTP request.
+message Directive {
```
geeknoid: Hmmm. How about ProxyDirective then?

The slight concern I have is that we've now decided to use this API from services as well as from the proxy. So services can cell Mixer.Check directly. But these directives are only really meaningful to the proxy case.

WDYT?

mixer/v1/check.proto
kyessenov: @guptasu is there any concern with using enums in templates?

mixer/v1/check.proto
kyessenov: For a basic policy check, I'm not sure what the direct API would do with this directive. For a rich library substituting the proxy, the library should exact the same way as the proxy. The library should have some default routing behavior (route to destination), that can be augmented by the directive. So it's not just a proxy directive. Maybe `RouteDirective`?

mixer/v1/check.proto
kyessenov: That would be OK. Do we have a way to override the negative check response code? There is a large number of 4xx and 5xx codes.

mixer/v1/check.proto
```diff
@@ -149,3 +148,44 @@ message ReferencedAttributes {
...
+  // If set, enables a direct response without proxying the request to the routing
+  // destination.
+  uint32 response_code = 3;
+
+  // Supplies the response body for the direct response.
+  // If this setting is omitted, no body is included in the generated response.
+  string response_body = 4;
+
+  // If set, overrides the routing destination to the specified destination service FQDN.
+  string destination_service = 5;
```
rshriram: and how will it be resolved?

kyessenov: Reverse lookup to a cluster, and overriding the route.

mixer/v1/check.proto
kyessenov: Rephrased as "manipulate the traffic metadata". Let me know if there is a better way to put it in words.

mixer/v1/check.proto
kyessenov: Clarified that the range must be 2xx and 3xx.

mixer/v1/check.proto
```diff
@@ -149,3 +148,46 @@ message ReferencedAttributes {
...
+  // destination. Required to be a value in the 2xx or 3xx range.
+  uint32 direct_response_code = 3;
+
+  // Supplies the response body for the direct response.
+  // If this setting is omitted, no body is included in the generated response.
+  string direct_response_body = 4;
+
+  // If set, overrides the routing destination to the specified destination service fully
+  // qualified name.
+  string override_destination_service = 5;
```
rshriram: isn't this the target cluster? and how do you hope to accomplish this in envoy? Allow filters to choose the target cluster?

kyessenov: Correct. Mixerfilter will override the cluster by mapping this string to a cluster name. The idea being that mixer should not be using envoy cluster names directly since that is an abstraction leak.

rshriram: Assuming that this destination service has a cluster in Envoy right? in other words, every reachable destination should be part of the service registry (either in kube or via service entry). You might want to mention that here. Also missing is the port. A destination could have multiple ports. And multiple subsets. Finally, this filter unless run on client side, will work in odd ways (esp with tracing, etc.).

Given that this is literally filter configuration, I would not mind exposing the cluster name directly here and generating the cluster name in mixer adapter implementation.

hklai:
We are about the lock down the release-0.8 branch. Is this meant for 0.8? If so, when it this expected to be merged? If not, please update the target branch.

kyessenov:
It's not needed for 0.8. the target branch is selected to make the diff
readable, and only for that.

PR #467, opened 2018-05-02
add ISTIO_MUTUAL mode to indicate the key cert provisioned by istio.

incfly:
This is for [incremental TLS adoption](https://docs.google.com/document/d/1fubdQ-HD_7IknE71HdaglgTkid1ScMq6LlIDvxKQDc4/edit#heading=h.x76uxhjmjzkj). Customer will write a `DestinationRule.TLSettings.Mode = ISTIO_MUTUAL` and authn policy together to adopt TLS.

https://github.com/istio/istio/issues/5059

@costinm @diemtvu PTAL


costinm:
Confused - what is the difference between MTLS and ISTIO_MTLS ? In both cases we use the certificates from the defined location. Also not sure what it has to do with adoption, using MTLS should be sufficient.

costinm:
I see, the other fields are now required. 

It may make sense to have a special mode with explicit MTLS certs and one with managed (since in future they may be from SDS)

incfly:
@costinm `ISTIO_MUTUAL` only means istio components will take care of the key certs provisioning. It could be well defined paths (for now), could be SDS (future). In either case, customer does not need to specify anything other than this mode.

incfly:
@costinm Please let me know if any rename or clarification are needed to avoid confusion, thx!

incfly:
@louiscryan @rshriram Could you approve this PR? Thanks.

networking/v1alpha3/destination_rule.proto
```diff
@@ -451,6 +462,11 @@ message TLSSettings {
     // Secure connections to the upstream using mutual TLS by presenting
     // client certificates for authentication.
     MUTUAL = 2;
+
+    // Originate a mutual TLS connection by presenting Istio provided
+    // key certs. All other fields in TLSSettings should be empty if using
+    // this mode.
```
rshriram: Copy explanation from mutual. Then add, “compared to Mutual mode, this mode uses certificates generated automatically by Istio for mTLS authentication. When istio mutual is used, all other fields in ...”

incfly: Done.

PR #468, opened 2018-05-03
Allow Unix Domain Socket endpoints for ServiceEntry

spikecurtis:
Adds the ability to specify Unix Domain Socket endpoints for networking v1alpha3 ServiceEntries. This update is back-compatible.

Our specific use case is to convince Pilot to add a cluster with this endpoint to sidecar proxies, so that the cluster can be used for RPCs from a custom filter. However, it can also be used as a regular Istio service accessible from workloads (via the host and port specified).

Signed-off-by: Spike Curtis <spike@tigera.io>


spikecurtis:
@costinm your review please
cc @rshriram

spikecurtis:
@rshriram are you suggesting something like
```
kind: ServiceEntry
metadata:
  name: unix-domain-socket-example
spec:
  hosts:
  - "pipe:///var/path/to/pipe"
  location: MESH_EXTERNAL
  ports:
  - number: 80
    name: http
    protocol: HTTP
```
or something else?

That could possibly work, but it means the service we create won't have a DNS compatible name.  I don't care about that for our particular use case, but it makes it awkward to expose the service to workloads if people wanted to do that.  I also think it is conflating the service name clients use with the actual address  the proxy uses to open a socket; and keeping those things separate is kinda what Istio is all about.

rshriram:
Not hosts. Addresses field.

rshriram:
Addresses and endpoints.address

spikecurtis:
@rshriram sorry; brain wasn't working yesterday afternoon.

I don't think we should allow this on the top level addresses field. That field is about how Envoy listens for connections to the service.  We don't intercept Unix domain socket traffic like we do with iptables, and I don't think we want to build that right now (I certainly don't have an interest in doing so).

In terms of allowing `pipe://` addresses in endpoints, that could work. I personally prefer the way I've proposed with an optional `family`. proto/YAML are already structured, so cramming additional structure into a string field smells funny. It means rolling string parsing functions on the implementation. It introduces an odd asymmetry where IP addresses & domain names are entered without a scheme, but Unix sockets need one.

If we do go that way, it should be `unix://`. I'm not sure why the Envoy core API calls it a pipe...

rshriram:
We can support connecting to services over uds. This has been a long time ask by folks that didn’t want to use loopback for app to sidecar and instead wanted a more direct mechanism like uds. We can always mark it as not implemented. I won’t hold you responsible for implementing it. :)

As for endpoints, I don’t have a string pref. So I am okay with extra field. 

Given that this is not even out, I wouldn’t mind changing address into a message that takes ip fqdn or uds. This would fit in nicely with top level address as well. 

The sooner it’s done the better. The resulting change in pilot is mostly mechanical.

Thoughts @costinm / @frankbu / @ZackButcher? I know this will be a breaking change (what I am proposing), but the alternative is not easily extensible and makes for a clunky api when we enable Unix domain socket based access mode for outbound calls.

spikecurtis:
@rshriram that sounds good to me

rshriram:
@ZackButcher suggested literally using Envoy's address proto (https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/address.proto#envoy-api-msg-core-address) in all address fields in ServiceEntry..

spikecurtis:
@rshriram @ZackButcher 
Envoy's address is slightly different---in the case of IP it also includes the port.  That makes it awkwardly redundant with the `ports` field where we want to gather some additional port metadata like name and application layer protocol. There is just this fundamental mismatch where we bundle up multiple ports into a "service", but Envoy doesn't really care about that. It also includes things like `resolver_name` which is Envoy-specific.

Lastly, do we have any current imports of Envoy APIs? I'm not enthusiastic about adding it as a new dependency.

All that is to say, it could work, but I don't think it's my first choice.

rshriram:
Fair enough. If we add an address struct, it needs to be done now and changes made to 0.8 before RC.

spikecurtis:
@rshriram I've added #471 which creates a new `Address` message.  We'll keep this PR around for comparison, until one (and only one) gets merged, then close the other.

networking/v1alpha3/istio.networking.v1alpha3.pb.html
```diff
@@ -2169,6 +2169,28 @@ <h2 id="ServiceEntry">ServiceEntry</h2>
...
+  hosts:
+  - &quot;example.unix.local&quot;
+  location: MESH_EXTERNAL
+  ports:
+  - number: 80
+    name: http
+    protocol: HTTP
+  resolution: STATIC
+  endpoints:
+  - address: /var/run/example/socket
```
costinm: I would also suggest what Shriram proposed: address: uds:/var/run/socket
(not pipe - I think unix domain socket is the common use, pipe is one way)

networking/v1alpha3/service_entry.proto
```diff
@@ -223,17 +246,30 @@ message ServiceEntry {
...
+    // service.  Do not use for `UNIX` endpoints.
     map<string, uint32> ports = 2;
     
     // One or more labels associated with the endpoint.
     map<string, string> labels = 3;
+
+    // AddressFamily identifies the kind of address the endpoint represents.
+    enum AddressFamily {
+      // The address is an IP address or domain name.
+      IP = 0;
```
costinm: I would guess TCP is a better name - but I would rather use a url syntax, with tcp as default

spikecurtis:
We'll use the `unix://` syntax. Will open a new docs-only PR.

PR #477, opened 2018-05-10
Regenerate docs given latest doc extraction tool.

geeknoid:
Also, fix python generation logic to stop passing directory names as proto files.
This eliminates a bunch of warnings when building the repo, and eliminates a bunch
of broken dud checked in files.


geeknoid:
Updated gogo I believe.

ldemailly:
Should this go in the 0.8 branch of API and propagate from there ?

geeknoid:
Er, I didn't realize there was a 0.8 branch in this repo. Yeah, this change might need to be moved into the branch.

PR #479, opened 2018-05-11
WIP FilterAugment

spikecurtis:
Work in progress; do not merge.

Adds a new object `FilterAugment` to the v1alpha3 APIs. This object allows operators to "augment" the filter chains of listeners in the service mesh, adding existing Envoy filters, like `envoy.lua` or `envoy.ext_authz`, or entirely custom filters (accompanied by a custom build of the Istio Proxy).

Based loosely on https://github.com/istio/api/pull/447 for inspiration

Some specific questions for reviewers:

1. I don't particularly like the name "FilterAugment".  We could go simple and just call it "Filter" or "CustomFilter". I ended up not calling it "CustomFilter" for the prototype because that name suggested it wouldn't apply to existing envoy built-ins. What do people think?

2. The host and gateway match fields select by name. This follows existing Istio conventions, but a more powerful alternative is Kubernetes-style label selectors. Is that a good idea? Do we currently track service or gateway labels?  Would we want to label sidecars?


spikecurtis:
CC @rshriram @nmnellis @mandarjog

networking/v1alpha3/filter_augment.proto
```diff
@@ -0,0 +1,249 @@
...
+//       name: reviews-route
+//     spec:
+//       matches:
+//       - hosts: [reviews.prod.svc.cluster.local]
+//         directions: [INBOUND]
+//       order:
+//         position: BEFORE
+//         relativeTo: router
+//       httpFilter:
+//         name: envoy.lua
```
nmnellis: I think there might have to be a httpFilter `type` to differentiate which filter the user wants to run. Otherwise how does Istio/Envoy know how to translate the config into valid envoy configuration?

networking/v1alpha3/filter_augment.proto
```diff
@@ -0,0 +1,249 @@
...
+    // repeated, the listener can match any of the listed values. For example,
+    // a match
+    //
+    //     match:
+    //     - gateways: [my-gateway, mesh]
+    //       listenerTypes: [TCP]
+    //
+    // will TCP listeners on the gateway "my-gateway", and also TCP listeners on
+    // sidecars in the service mesh.
+    message Match {
```
nmnellis: Would it be easier to define a filter via a FilterAugment but leave the matching to the VirtualService? This would be more clear to the user when the filter would be applied.

Example of what the filter configuration could look like.

```yaml
kind: VirtualService
metadata:
  name: example
spec:
  gateways:
  - istio-ingress
  - mesh
  hosts:
  - example.weather.com
  http:
  - route:
    - destination:
        name: example
  - filter:
      name: reviews-route
      direction: INBOUND
      order: FIRST
  - match:
    - uri:
        exact: /v1/example
```

spikecurtis: I don't think we should require the use of VirtualServices. There will be plenty of cases where the insertion of the filter is desired for a "normal" service.

In our motivating use case, we want to insert the filter on all inbound listeners on all sidecars.  We don't want to have to create a new config object for every service to make this work.

networking/v1alpha3/filter_augment.proto
nmnellis: thanks, missed that

networking/v1alpha3/filter_augment.proto
nmnellis: i think that is valid as i have the same use-case, though you could also allow gateway level filters... Im not advocating for either, just offering another solution to think about that is a little more clear to the user when it is being applied
```
kind: Gateway
metadata:
  name: istio-ingressgateway
spec:
  selector:
    istio: ingressgateway
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - example.weather.com
  filters:
    - filter:
      name: reviews-route
      direction: INBOUND
      order: FIRST
```

spikecurtis: I actually want the filters on sidecars, which don't have an explicit gateway object

nmnellis: which could be solved with a simple label and new gateway definition?

frankbu: Having the filters reference the gateway(s) that they should be bound to, like VirtualService's do it, seems like a natural extension.

frankbu:
This looks pretty cool to me. The only thing I'm not crazy about is the name: `FilterAugment`. Maybe something like `ProxyFilter` or `ListenerFilter` would be better?

kyessenov:
What's the relationship of this to pilot plugins? Is it worth having two distinct extensibility mechanisms for envoy xDS output? I assume both this and the plugins operate in a similar fashion as a processor in the config pipeline.

For mixer specifically, [here](https://github.com/kyessenov/envoymesh/blob/master/envoy.jsonnet#L124) is the desired augmentation that we need. I am not sure how to translate that into the proposed augmentation API, but it might be worthwhile to cover all the cases that we need for mixer.

spikecurtis:
@kyessenov this mechanism is much more narrow than what pilot plugins allow. Pilot plugins allow arbitrary modification of the listener and cluster objects. This mechanism is used only to insert new filters into listeners. This is a really common use case, so having a config API makes sense. If people need more flexibility, the pilot plugin mechanism is open to them.

The pilot implementation for this API will almost certainly be a pilot plugin, BTW.

*Regarding Mixer*
The proposal as it stands only allows filter config to be statically defined. I believe that the mixer filter config depends on the node and other dynamic parameters that Pilot computes for each instance, so it would not be possible to configure the Mixer filter using this mechanism.

The intention is to allow insertion of filters beyond the "base set" of filters that Istio handles (which includes the mixer filter).  We could explore enhancing the API to include the mixer use case, but my preference would be to work incrementally. I think we would need, at minimum, some sort of template capability to configure mixer. We would also want to think carefully about how this mechanism interacts with the current configuration options for mixer, and have a plan to transition and deprecate.

kyessenov:
Thanks for responding. It would be great to align this API with the internal plugins if the plan is to implement it as a plugin. In particular, I see a lot of similarity in how matching is performed with how pilot plugins are invoked using callbacks. The argument between callbacks and what the API allows should at least more or less be the same.

I agree that we need templating for mixerfilter.

spikecurtis:
@louiscryan I'm reworking this proposal to better reflect the reality of how we use outbound listeners, so you might want to hold off on your review. I'll ping you again when it is done; expect it later today.

spikecurtis:
I've done some reworking to reflect the fact that outbound listeners are handled significantly differently than inbound, and to a certain extent, Gateways also act differently than sidecars.

Direction   | Gateway    | Sidecar
------------|------------|--------------------------------------
**Inbound**     | None       | per local service & port
**Outbound**    | per Server | per HTTP port, per TCP service & port

Following the suggestion by @nmnellis and @frankbu , I have split the filter config definition and the way they are bound to listeners.  For `Gateway`s, this is a straightforward extension of the `Server` message, since there is one listener per server (i.e. per port). I think this fits the gateway usage model quite well.

For sidecars, I think we want the ability to bind both narrowly (e.g. to a single service port) and widely (e.g. to every inbound listener), so I have introduced a `SidecarFilterBinding` that includes match criteria, ordering, and of course, the name of the filter config.

@kyessenov I think this also moves us materially closer to covering the mixer use case, by splitting the filter config from the binding. We could use the binding mechanism to control where mixer is included in the filter chain. "mixer" could be a reserved filter config name until we build the capability to have filter config templates.

@louiscryan please let me know what you think.

kyessenov:
I have a couple of points to add.

1. Do we want port-level filter or service-level filter for HTTP? Currently, we bunch multiple services on the same HTTP route configuration (to be dispatched by virtual host). This is rather fragile when you have a VIP for the same HTTP service (e.g. gRPC does not respect virtual host). That means we should allow have both options. However, we cannot simply combine HTTP filter configs across many services onto the same HTTP route, unless the filter is compliant with `per_filter_config` semantics.

2. mixerfilter also wants EDS access to supply metadata per host in the upstream.

kyessenov:
By the way, we are actively trying to get rid of the notion of inbound service. There is just an instance that has an IP and a port. We will not have service info attached to that listener, since the service and its address is generally known by the client only. I suggest that anything inbound only takes into account pod IP, pod port, and maybe port protocol.

spikecurtis:
Reviewers: I want to call attention to the fact that this new design has the disadvantage of coupling the API fairly closely with the way Pilot currently sets up Envoy listeners.

Firstly, I think that to a certain extent, it is unavoidable for developers or operators looking to extend Istio functionality to have to reason about what listeners get created and what they do.

Secondly, I think that this may be a case of perfect being the enemy of good. It might be possible to implement an interface that didn't invoke listeners and instead allowed the user to say "please make sure traffic that matches these criteria traverses this filter, I don't care how." Implementing such an interface, I believe, would be very disruptive to how Pilot allocates listeners, since filters must be instantiated at that scope in Envoy. I don't think we want to bite off that kind of disruption before 1.0, and there are many of us who do not want to wait post 1.0 for this functionality to appear at all.

spikecurtis:
#447 meets this need, so closing.

PR #480, opened 2018-05-14
Add Script modeling to VirtualService

mandarjog:
1. Add Script as a top level message
2. Associate scripts with a `VirtualService`, `HTTPRoute` and `TCPRoute`


networking/v1alpha3/virtual_service.proto
```diff
@@ -158,6 +158,9 @@ message VirtualService {
   // An ordered list of route rules for TCP traffic.
   // The first rule matching an incoming request is used.
   repeated TCPRoute tcp = 4;
+
+  // An ordered list of scripts associated with the VirtualService.
+  repeated Script scripts = 5;
```
rshriram: why repeated?

mandarjog: You can have multiple `Scripts` elements one for each proxy flavor.

networking/v1alpha3/virtual_service.proto
```diff
@@ -928,3 +937,32 @@ message PortSelector {
...
+
+  // proxy is the type of proxy that can excute this script.
+  // examples: envoy, nginx
+  // default: envoy
+  string proxy = 2;
+
+  // phase is the phase of proxy processing.
+  // This is proxy dependant. Envoy has a request and a response phase.
+  // nginx has by_lua variations of `rewrite`, `access`, `header_filter` and other phases.
+  string phase = 3;
```
rshriram: we don't need this do we? We let the user define an opaque string that has the script. ITs upto the user to use the envoy APIs to make sure it acts on request/response (i.e. envoy_on_request, envoy_on_response, etc.)

rshriram:
should go into release 0.8 branch?

I had envisioned something much more simpler like adding the script to every http route for starters. We can always add this to entire virtual service later if we need to (notice that we uniformly normalized all settings to per route. So this means more repetition but very easy to script and generate the config).
Also for starters, we could simply have the script be a string and document that the interpretation is specific to the implementation or proxy platform being used..
Something like
```HTTPRoute:
- ...
   string customScript
```

so that the user could stick in a Lua script or a ngin script or anything. Given that a large section of this config is already attuned to envoy XDS protos, for purpose of simplicity and ease of implementation, lets just stick with a single string var where we could dump lua/js or any script (its lua for the forseable future)

geeknoid:
@mandarjog This is many months old, is this still relevant?

PR #481, opened 2018-05-14
Move types to policy/v1beta1

guptasu:
These types are interesting for the operator. There are two issues if we keep types.proto inside mixer/adapter/model/v1beta1
1. Operator does not care about adapter/model, so it can be confusing for them.
2. With the current structuring we either end up documenting entire `mixer/adapter/model/v1beta1` which also contains very adapter specific types, or we end up in-lining the relevant types inside each template documentation. Both the options are not ideal.
If we move type.proto into policy/v1beta1, it nicely goes together with Value_Type enum (relevant for operator and already in policy/v1beta1) and we already create one separate documentation for policy/v1beta1 which the individual template documentation can reference. Therefore no in-lining or showing extra adapter-specific stuff to operator.

**Note** the types inside type.proto are used by operator, adapter developer as well as template developer. But our documentation for those types must be more operator focused to make things less confusing for them (others will figure it out :-)). I have made the necessary documentation text change in this PR.


douglas-reid:
will there be any impact on existing config, etc., when this merges into istio/istio ?

I'm starting to wonder whether the `policy` namespace is the right place to hold this kind of information. The use of `policy` implies, I think, some special relationship between the contents and Istio policies. But, `IpAddress` isn't restricted to policies really, and is a more abstract concept.  Should we have named this something more like `attributes` or `rules` ?

guptasu:
@douglas-reid this should not affect any existing configuration. These types are currently just used in Mixer to transform and hold data and eventually pass them to adapters. So, existing adapters do break since they need to reference these new types. But since adapters are now just local inside Mixer, we should be fine. If not this is will cause compile time error and it is just a simple find and replace fix.

Regarding, `policy` is right or not: Are you suggesting the entire `policy` package is probably a wrong name because it is confusing as it contains attribs and types and nothing about istio's policy? We can rename `policy` to  `rules`. Or you are suggesting `type.proto` and may be `value_type.proto` move to a sibling of `policy` package into something something like `rules`. I think  you are referring to the former, and I mostly agree for the rename. @geeknoid WDYT ?

policy/v1beta1/type.proto
```diff
@@ -0,0 +1,127 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.policy.v1beta1;
+
+option go_package="istio.io/api/policy/v1beta1";
```
mandarjog: is this identical to deleted `mixer/adapter/model/v1beta1/type.proto`?
If so can you you `git mv a b` so as to preserve where this came from?

douglas-reid:
@guptasu yeah, i was concerned more about the naming. I don't know if we want to fix it or not -- and it is probably re-opening a closed topic. It just seems like a potential source of confusion.

geeknoid:
Can we go ahead with this change? I need it in order to generate nice-looking docs for 0.8. We can do further renaming/refactoring later.

douglas-reid:
@geeknoid yeah, i was just wondering on the side about the proper place for this info. please go ahead with the change.

geeknoid:
Sunny, can you refresh the PR? Once that's done, I think this can go in.

PR #483, opened 2018-05-14
Update global dictionary for new attributes.

JimmyCYJ:
Add connection.mtls into global dictionary.


JimmyCYJ:
I am not going to cherry pick this PR into release-0.8. This change is not urgent.

PR #484, opened 2018-05-14
add auth.claims and auth.raw_claims

mandarjog:


hklai:
I think @qiwzhang had merged something like this already. Is this still needed?

We are about the lock down the release-0.8 branch.

qiwzhang:
https://github.com/istio/api/pull/488
already has this change.

PR #487, opened 2018-05-15
Release-0.8: Add source.principal and request.auth.claims attributes to global dictionary

qiwzhang:


mixer/v1/global_dictionary.yaml
```diff
@@ -246,3 +247,7 @@
 # whether mixer check and quota calls hit cache.
 - check.cache_hit
 - quota.cache_hit
+
+- source.principal
+- request.auth.claims
```
douglas-reid: @theganyo - might as well add `request.auth.raw_claims` right?

qiwzhang: This file global_dictioinary.yaml should be only source of truth.  Istio/mixer and proxy should generate their global dictionary from this file.

qiwzhang: mixer list.go should be generated from this file.

PR #490, opened 2018-05-16
clarify wildcard usage in gateway server hosts

vadimeisenbergibm:
based on the documentation of Envoy sni_domains - https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener/listener.proto.html?highlight=sni_domains 

> May contain a wildcard prefix for the bottom-level domain of a domain name, e.g. *.example.com. Note that foo.example.com will be matched by foo.example.com and *.example.com SNI domain names, but not by *foo.example.com, *oo.example.com, *example.com, *.com or *.

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -64,13 +64,13 @@ option go_package = "istio.io/api/networking/v1alpha3";
 //           name: http-wildcard
 //           protocol: HTTP
 //         hosts:
-//         - "*"
+//         - "*.bookinfo.com"
 //       - port:
 //           number: 2379 # to expose internal service via external port 2379
 //           name: mongo
 //           protocol: MONGO
 //         hosts:
-//         - "*"
+//         - "*.bookinfo.com"
 //
 // The Gateway specification above describes the L4-L6 properties of a load
 // balancer. A `VirtualService` can then be bound to a gateway to control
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -221,7 +221,11 @@ message Server {
   // REQUIRED. A list of hosts exposed by this gateway. At least one
   // host is required. While typically applicable to
   // HTTP services, it can also be used for TCP services using TLS with
-  // SNI. Standard DNS wildcard prefix syntax is permitted.
+  // SNI. May contain a wildcard prefix for the bottom-level domain of
+  // a domain name, e.g.`*.example.com`. Note that `prod.example.com`
+  // will be matched by the `prod.example.com` and `*.example.com` host
+  // names, but **not** by `*prod.example.com`, `*rod.example.com`,
+  // `*example.com`, `*.com` or `*`.
```
frankbu: If this is true, why is `bookinfo-gateway.yaml` working? It uses `hosts: "*"`

vadimeisenbergibm: Good point. `bookinfo-gateway.yaml` with `hosts: "*"` would not work with HTTPS (TLS). I propose to limit the host names for HTTP as well, for consistency and to prevent confusion.

rshriram:
this is a very good find.. Thanks for the doc fix.. We might need to disallow wildcard `*` for HTTPs

hklai:
We are about the lock down the release-0.8 branch. Is this meant for 0.8? If so, when it this expected to be merged? If not, please update the target branch.

rshriram:
This is for this branch..

PR #492, opened 2018-05-16
mixerfilter: add forward attributes to service config

kyessenov:
Extend the config for mixerfilter to allow route-level overrides for forwarded attributes.


mandarjog:
@kyessenov is this needed for 0.8?

mixer/v1/config/client/client_config.proto
```diff
@@ -73,6 +73,10 @@ message ServiceConfig {
   // [mesh-level
   // policy][istio.mixer.v1.config.client.TransportConfig.network_fail_policy].
   NetworkFailPolicy network_fail_policy = 7;
+
+  // Default attributes to forward to upstream. This typically
```
bianpengyuan: To my understanding this set of forward_attributes is going to be combined with the one in http client config and forwarded? If so, should we update the comment to reflect that?

qiwzhang: Please update comment to reflect the relationship between this one and the one from filter_config

kyessenov: Thanks. I've clarified both the new and the old `mixer_attributes` override. Internally, we use `MergeFrom` on the client context attributes first, service config second.

kyessenov:
@mandarjog No, it's not required for 0.8. I can mark it un-implemented in the docs if that is necessary. Diff between this and master is unreadable at this point.

mixer/v1/config/client/client_config.proto
kyessenov: By the way, can you confirm the following full ordering of overrides:
1. mixer_attributes from destination filter config
2. mixer_attributes from destination route config
3. forward_attributes from source filter config
4. forward_attributes from source route config (new)
5. request headers (host, referer, etc)

Here, it's strict override from 1 to 5, with no way to delete attributes. Once you confirm, I can put it into the spec, so that we don't have to read the code again :)

kyessenov:
@hklai is prow dead?

hklai:
@kyessenov https://k8s-gubernator.appspot.com/build/istio-prow/pull/istio_api/492/api-presubmit/687/

It passed. Maybe github is missing some notification?

mixer/v1/config/client/client_config.proto
qiwzhang: Yes.

kyessenov:
@hklai Yeah, I do not intend to merge it into the release. I'll wait until master is on par and rebase then.

PR #493, opened 2018-05-16
Move a couple Mixer docs

geeknoid:


mixer/v1/service.proto
```diff
@@ -14,12 +14,6 @@
 
 syntax = "proto3";
 
-// $title: Mixer
-// $description: API definitions to interact with Mixer
-// $location: https://istio.io/docs/reference/api/istio.mixer.v1.html
-// $front_matter: redirect_from: /docs/reference/api/mixer/mixer.html
-
-
```
guptasu: We don't want to document Mixer API ?

geeknoid:
Nope, this doesn't belong on istio.io. istio.io is for operators and not
developers.

guptasu:
Agree, just wanted to make sure this is not by accident.

PR #499, opened 2018-05-22
Document v1alpha1 routing is deprecated

frankbu:


routing/v1alpha1/route_rule.proto
```diff
@@ -20,11 +20,14 @@ import "google/protobuf/wrappers.proto";
...
 import "routing/v1alpha1/l4_fault.proto";
 
-// $title: Route Rules v1alpha1
+// $title: Route Rules v1alpha1 (deprecated)
 // $overview: Configuration affecting traffic routing
 // $location: https://istio.io/docs/reference/config/istio.routing.v1alpha1.html
 // $front_matter: redirect_from: /docs/reference/config/traffic-rules/routing-rules.html
 
+// **Note: This version of the traffic management API has been deprecated and will
+// not be supported from Istio 0.9 onwards.**
```
geeknoid: Just say "after 0.8" since we don't know if we're doing 0.9 at this point...

PR #501, opened 2018-05-22
attributes: add context reporter attributes

kyessenov:
`context.reporter.proxy` has values `client` or `server` for the two sidecars on the data path.
`context.reporter.uid` carries the UID value of the pod.

### DIFF
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -277,5 +277,5 @@
 - destination.container.name
 - destination.container.image
 
-- context.reporter.proxy
+- context.reporter.local
 - context.reporter.uid
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -219,10 +219,10 @@
 - destination.namespace
 - destination.labels
 # destination.user is deprecated, please refer to destination.principal
-- destination.user 
+- destination.user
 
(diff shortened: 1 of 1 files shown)


kyessenov:
It's the same format as `source.uid` and `destination.uid`. I think we've been using something like `kubernetes://pod1.ns2`, which may need to be extended to support multiple clusters.

douglas-reid:
ok. yeah, i think the latest attribute proposal updated the proposed string for those UIDs. we'll need to figure out how to manage that transition.

kyessenov:
PTAL @geeknoid

geeknoid:
Is .proxy the right choice here? I wonder if it would be better as a boolean:

context.reporter.local

kyessenov:
I like the simplicity, but not sure if a boolean is sufficient. For a middle proxy (e.g. ingress), what would the value be?

geeknoid:
I guess it depends on what you think the use case for the attribute is. How
do you expect it to be used?

kyessenov:
I'd like to have a simple condition for telemetry templates, that selects metrics from a proxy. For example, I'd like to aggregate traffic to a destination from the perspective of the destination sidecar, or from the perspective of the every other client in the mesh. The ingress/middle proxy in that case is just another client of the destination. So I guess, we should have "local" boolean that indicates whether the reporter is the server sidecar or not. I'll update the PR.

PR #502, opened 2018-05-22
mixerfilter: control upstream forwarding

kyessenov:
For mixer self-monitoring, we want to forward attributes in gRPC metadata to avoid parsing the payload at the proxy. This adds an option to encode forwarded attributes (e.g. `source.uid`) as an HTTP header that mixer sidecar can decode.


mixer/v1/config/client/client_config.proto
```diff
@@ -113,6 +113,10 @@ message TransportConfig {
   // NOTE: Any value other than the default "mixer_server" will require the
   // Istio Grafana dashboards to be reconfigured to use the new name.
   string report_cluster = 7;
+
+  // Default attributes to forward to mixer upstream. This typically
+  // includes the "source.ip" and "source.uid" attributes.
+  Attributes control_forward_attributes = 8;
```
qiwzhang: can we use the field in line 143?

kyessenov: No, this is distinct from the payload attributes. For example, the server `Report` should be using `source.uid` set to pod ID, but the payload attributes using `source.uid` set to the request source pod ID.

qiwzhang: per offline discussion,  I understand its purpose now.  how about call it:

// These attributes are specific for the istio/proxy sitting in front of mixer 
attributes_for_mixer_proxy

kyessenov: Done

qiwzhang: Could you add this comment too.  I think it will help

// These attributes are send to the istio/proxy sitting in front of mixer

kyessenov: Done

PR #475, opened 2018-05-10
Update global dictionary with new attributes

douglas-reid:
This PR adds the new attributes proposed in [Updated Source & Destination Attributes](https://docs.google.com/document/d/1IcR6C7j5Vl-4Jx8ervVT4lCTBDrBgsD7Uac77K5C5h4/edit#) to the
global dictionary used for communication with Mixer.

PR #503, opened 2018-05-22
Merging release-0.8 to master

hklai:
We have the supposedly last PR in the release-0.8 branch so this merges all 0.8 PRs to master.

The release-0.8 branch of istio/api is now locked.


mixer/v1/global_dictionary.yaml
```diff
@@ -255,6 +255,8 @@
 - request.auth.claims
 - request.auth.raw_claims
 
+# whether request is sent over mTLS enabled downstream connection.
+- connection.mtls
```
kyessenov: cc @douglas-reid  It's very dangerous to insert attributes in the middle of this file. Are we sure this is OK?

hklai: It was a conflict that I manually resolved. 

Stuff were added to both release-0.8 and master.

https://github.com/istio/api/pull/494/commits/c009438faf23c99cf3a714bb792e8f9d39fa7a5f
https://github.com/istio/api/pull/475/commits/601338fba3a1f3b1fbdab8ebaae46edb2e8c68ad

Either one will have to be inserted. Please advise if the current order is not desired..

douglas-reid: yeah, maintaining order here is paramount.  Since the other stuff came later -- and hasn't been used anywhere, this _should_ be ok.

cc: @mandarjog

PR #504, opened 2018-05-23
hide unused fields in networking protos from docs

rshriram:
Especially hide portselector.name, which is not implemented.


kyessenov:
@hklai -- stale PR?

rshriram:
It’s not stale. It’s fixing docs

PR #506, opened 2018-05-23
fix wildcard doc

frankbu:


rshriram:
force merging as its just doc change..

PR #510, opened 2018-05-25
mixer: route directive

kyessenov:
Re-creating the PR on master.

Removed destination service override for now, since it's contentious. I suspect to properly implement routing override in the proxy we need to have an incremental xDS implementation in pilot.


mixer/v1/check.proto
```diff
@@ -149,3 +148,42 @@ message ReferencedAttributes {
...
+  // Header value.
+  string value = 2;
+
+  // Header operation.
+  Operation operation = 3;
+}
+
+// Expresses the routing manipulation actions to be performed on behalf of
+// Mixer in response to a successful pre-condition check.
+message RouteDirective {
```
qiwzhang: Envoy ext_authz is trying to define something similar

https://github.com/envoyproxy/envoy/pull/3162/files

Could we use its?

-Wayne

kyessenov: I'd like to avoid importing alpha-level APIs yet. I agree that they are similar in spirit, but we want some level of abstraction over envoy here.

mixer/v1/check.proto
```diff
@@ -71,17 +72,15 @@ message CheckResponse {
...
-    CompressedAttributes attributes = 4 [(gogoproto.nullable) = false];
+    reserved 4;
 
     // The total set of attributes that were used in producing the result
     // along with matching conditions.
     ReferencedAttributes referenced_attributes = 5 [(gogoproto.nullable) = false];
+
+    // An optional directive, used to manipulate the traffic metadata
+    // whenever all pre-conditions are satisfied.
+    RouteDirective directive = 6;
```
rshriram: route_directive

kyessenov:
Ping? Implementation is in progress.

mixer/v1/check.proto
```diff
@@ -71,17 +72,15 @@ message CheckResponse {
...
-    // attributes and its configuration.
-    CompressedAttributes attributes = 4 [(gogoproto.nullable) = false];
+    reserved 4;
 
     // The total set of attributes that were used in producing the result
     // along with matching conditions.
     ReferencedAttributes referenced_attributes = 5 [(gogoproto.nullable) = false];
+
+    // An optional routing directive, used to manipulate the traffic metadata
+    // whenever all pre-conditions are satisfied.
```
geeknoid: pre-conditions -> preconditions

mixer/v1/check.proto
```diff
@@ -149,3 +148,42 @@ message ReferencedAttributes {
   // Describes a set of attributes.
   repeated AttributeMatch attribute_matches = 2 [(gogoproto.nullable) = false];
 }
+
+// Operation on HTTP headers to replace, append, or remove a header. Header
+// names are normalized to lower-case with dashes, e.g.  "x-request-id".
+// Special headers "uri", "scheme", "method", and "authority" are permitted,
+// and should be used to effect the general request metadata.
```
geeknoid: effect -> affect

mixer/v1/check.proto
```diff
@@ -149,3 +148,42 @@ message ReferencedAttributes {
...
+
+  // Header value.
+  string value = 2;
+
+  // Header operation.
+  Operation operation = 3;
+}
+
+// Expresses the routing manipulation actions to be performed on behalf of
+// Mixer in response to a successful pre-condition check.
```
geeknoid: pre-condition -> precondition

kyessenov:
Thank you @geeknoid . Corrected my spelling.

kyessenov:
@geeknoid @douglas-reid @mandarjog Are we happy with this shape of the API? I've implemented it in envoy, but it's getting increasingly difficult to automate testing, since code is spread all over the repos.

mixer/v1/check.proto
```diff
@@ -149,3 +148,42 @@ message ReferencedAttributes {
...
 }
+
+// Operation on HTTP headers to replace, append, or remove a header. Header
+// names are normalized to lower-case with dashes, e.g.  "x-request-id".
+// Pseudo-headers ":path", ":authority", and ":method" are supported to modify
+// the request headers.
+message HeaderOperation {
+  // Operation type.
+  enum Operation {
+    REPLACE = 0;  // replaces the header with the given name
```
douglas-reid: This implies that the default action is REPLACE.  Is that what we want?  Would it make more sense to have APPEND be the default case?

mixer/v1/check.proto
```diff
@@ -149,3 +148,42 @@ message ReferencedAttributes {
...
+
+  // Header operation.
+  Operation operation = 3;
+}
+
+// Expresses the routing manipulation actions to be performed on behalf of
+// Mixer in response to a successful precondition check.
+message RouteDirective {
+  // Operations on the request headers.
+  repeated HeaderOperation update_request_headers = 1 [(gogoproto.nullable) = false];
```
douglas-reid: nit: i'd name these `request_header_operations` and `response_header_operations`.

mixer/v1/check.proto
```diff
@@ -71,17 +72,15 @@ message CheckResponse {
...
     int32 valid_use_count = 3;
 
-    // The attributes returned by Mixer.
-    //
-    // The exact set of attributes returned is determined by the set of
-    // adapters Mixer is configured with. These attributes are used to
-    // ferry new attributes that Mixer derived based on the input set of
-    // attributes and its configuration.
-    CompressedAttributes attributes = 4 [(gogoproto.nullable) = false];
+    reserved 4;
```
douglas-reid: super nit: can we move this reservation up to the top of the message ?

mixer/v1/check.proto
kyessenov: I think so. Headers are usually assumed to a key-value map, so replacing is consistent with a map semantics. Appending means the value becomes a space separated list, which can be surprising.

mixer/v1/check.proto
kyessenov: Done

mixer/v1/check.proto
kyessenov: Done

PR #511, opened 2018-06-05
Remove mtls_excluded_services from mesh config.

diemtvu:


mesh/v1alpha1/config.proto
```diff
@@ -234,13 +234,15 @@ message MeshConfig {
...
+  // DEPRECATED. See https://istio.io/docs/setup/kubernetes/upgrading-istio/#migrating-mtls-excluded-services-config-to-destination-rules
+  // for alternative.
   // List of remote services for which mTLS authentication should not be expected by Istio .
   // Typically, these are control services (e.g kubernetes API server) that don't have Istio sidecar
   // to transparently terminate mTLS authentication.
   // It has no effect if the authentication policy is already 'NONE'.
   // DO NOT use this setting for services that are managed by Istio (i.e. using Istio sidecar).
   // Instead, use service-level annotations to overwrite the authentication policy.
-  repeated string mtls_excluded_services = 15;
+  repeated string mtls_excluded_services = 15 [deprecated = true];
```
rshriram: why can't we remove this completely ? set the field to reserved or something

diemtvu: I would love to, but I'm not sure if it cause any trouble for upgrading (i.e new pilot has no knowledge of this field but the existing config map has it)?

diemtvu:
Given this flag is removed in https://github.com/istio/istio/pull/6056, I remove this flag completely.

PR #512, opened 2018-06-06
Move back jwt_auth/config.proto from Envoy

qiwzhang:
Envoy jwt_auth config is making breaking change.  Istio/proxy jwt_auth fitler could not directly use the one in Envoy.
* Add the config file in istio/api,  this PR
* Change istio/proxy and istio/istio to use this one.


qiwzhang:
Envoy jwt_authn config (PR: https://github.com/envoyproxy/envoy/pull/3381) is not (wire) compatible with the one used by Istio anymore.   This is OK,  since the filer name is different, Istio Pilot needs to generate new Envoy config when Istio switches to use the Envoy jwt_authn.  So it is not required to support (wire) compatible between Envoy jwt_authn config, and Istio/proxy jwt_auth config.

Istio/proxy jwt_auth  filter will continue to use the one added in this PR.  When Envoy jwt_authn is ready, and Istio/proxy can switch to use it directly.  The change will be in Pilot: to use different filter with new config format.  After that jwt_auth filter in Istio/proxy can be removed.  

@diemtvu any comment on this plan?

diemtvu:
Sounds good, except please make sure the filter names are different (between envoy impl and our proxy). May be we should change our filter name to have `istio_` prefix.

kyessenov:
The correct naming should be "istio.jwt_auth".

qiwzhang:
Envoy filter names prefixes with "envoy." .  For sure, it will be different.

PR #513, opened 2018-06-06
Add EnforcementMode to RBAC ServiceRole/ServiceBinding config

quanjielin:
https://github.com/istio/istio/issues/5345

I think it's OK to add enforcement_mode on both ServiceRole and ServiceBinding configs, pilot is going handle the conversion to envoy config.


rbac/v1alpha1/rbac.proto
```diff
@@ -108,12 +108,26 @@ package istio.rbac.v1alpha1;
 
 option go_package="istio.io/api/rbac/v1alpha1";
 
+// RBAC config enforcement mode, used to verify new ServiceRole/ServiceBinding configs
+// work as expected before rolling to prod. RBAC engine only logs results from
+// ServiceRole/ServiceBinding configs in dark launch mode, and discards the result
+// before return to the user.
+enum EnforcementMode {
+  ENFORCEMENT_ENFORCED = 0;
```
diemtvu: I prefer ACTIVE/INACTIVE, or ENFORCED (or STRICT)/PERMISSIVE (or LOG_ONLY) which explain better the direct impact of the `mode`. Darklaunch is the process that based on this (mode) feature.

rbac/v1alpha1/rbac.proto
```diff
@@ -171,6 +185,9 @@ message ServiceRoleBinding {
 
   // Required. Reference to the ServiceRole object.
   RoleRef roleRef = 2;
+
+  // Indicates enforcement mode of the ServiceRoleBinding config, by default mode is ENFORCED.
```
diemtvu: You need examples how these mode can be used. Especially what does it mean for Role is in mode=DL, but rolebinding is not (and vice versa)

yangminzhu: Maybe explain the following cases in the comment:
- ServiceRole doesn't have mode but associated ServiceRoleBinding has mode.
- ServiceRole has mode but associated ServiceRoleBinding doesn't have mode.

rbac/v1alpha1/rbac.proto
quanjielin: renamed

rbac/v1alpha1/rbac.proto
quanjielin: added more comments, I think the comment is enough to explain(without yaml sample), we may need to add examples about how to map ServiceRole/Binding to envoy config, but not in this file(here is for user-facing).

rbac/v1alpha1/rbac.proto
```diff
@@ -108,12 +108,28 @@ package istio.rbac.v1alpha1;
 
 option go_package="istio.io/api/rbac/v1alpha1";
 
+// RBAC config enforcement mode, used to verify new ServiceRole/ServiceBinding configs
+// work as expected before rolling to prod. RBAC engine only logs results from
+// ServiceRole/ServiceBinding configs in dark launch mode, and discards the result
+// before return to the user.
+enum EnforcementMode {
```
diemtvu: Please add // $hide_from_docs, and/or note that they are work-in-progress so users won't confuse.

You might also want to add examples for these (even without implementation) to make sure the concept is clear.

rbac/v1alpha1/rbac.proto
```diff
@@ -108,12 +108,28 @@ package istio.rbac.v1alpha1;
...
+  ENFORCED = 0;
```
diemtvu: Add comment for each mode.

rbac/v1alpha1/rbac.proto
```diff
@@ -108,12 +108,28 @@ package istio.rbac.v1alpha1;
...
+ 
 // ServiceRole specification contains a list of access rules (permissions).
 // This represent the "Spec" part of the ServiceRole object. The name and namespace
 // of the ServiceRole is specified in "metadata" section of the ServiceRole object.
 message ServiceRole {
   // Required. The set of access rules (permissions) that the role has.
   repeated AccessRule rules = 1;
+
+  // Indicates enforcement mode of the ServiceRole config, ENFORCED by default.
+  // If ServiceRole is in PERMISSIVE mode, all ServiceRoleBindings that
```
diemtvu: You can move this to the enum comments instead.

rbac/v1alpha1/rbac.proto
quanjielin: added hide_from_docs and more comment, will add examples after feature code is in(and remove the hide at same time)

rbac/v1alpha1/rbac.proto
quanjielin: added

PR #522, opened 2018-06-11
Fix VirtualService in Gateway example for v1alpha3

x15zhang:
Correct destination field from "name" to "host" in VirtualService for Gateway example.


frankbu:
This is generated documentation. These changes need to be made in the source file https://github.com/istio/api/blob/master/networking/v1alpha3/gateway.proto

PR #525, opened 2018-06-11
Add field to keep human-readable information of authentication policy.

diemtvu:
- Add 'description' to authentication policy for human readable note.
- Add policy_name to authentication filter config to show the corresponding policy that pilot used to config filter (for debugging purpose).


envoy/config/filter/http/authn/v2alpha1/config.proto
```diff
@@ -33,4 +33,7 @@ message FilterConfig {
   // This information is added by pilot when construct and add Jwt and
   // authN filters.
   map<string, string> jwt_output_payload_locations = 2;
+
+  // Name of the policy above in the istio config storage. This is for debugging purpose.
```
incfly: Why we need this? I think policy itself already has the name?

diemtvu: CRD has name, but the config we send from pilot to envoy contain the proto only.

incfly: (up to you) but do we need this? We add `description` above and that will be propagated from pilot to sidecar for free. This field I assume you need to change pilot code to copy from CRD name?

diemtvu:
Yeah, I plan to populate policy name in pilot so it's easier to check. However, it might make the code a bit ugly.

Adding extra text field to the user facing API may not be a good approach neither.

 Close this PR for now until we needed.

PR #526, opened 2018-06-11
Create a separate field compiled in template and compiled in adapters.

guptasu:


policy/v1beta1/cfg.proto
```diff
@@ -180,7 +180,15 @@ message Instance {
   // to an instance produced by this instance.
   string name = 1;
 
-  // Required. The name of the template this instance creates instances for.
+  // Required. The name of the compiled in template this instance creates instances for. For referencing non compiled-in
+  // templates, use the `template` field instead.
+  //
+  // The value must match the name of the available template Mixer is built with.
+  string template_compiled = 67794676;
```
douglas-reid: i think our API guidelines suggest making this `compiled_template` instead.

PR #528, opened 2018-06-12
mesh: client side check option

kyessenov:
This will enable client-side checking. It is off by default.


mesh/v1alpha1/config.proto
```diff
@@ -269,4 +269,7 @@ message MeshConfig {
   OutboundTrafficPolicy outbound_traffic_policy = 17;
 
   reserved 18;
+
+  // Enables clide side policy checks.
+  bool enable_client_side_policy_check = 19;
```
frankbu: Will there also be a way to enable for specific services, or is the global switch the only way?

kyessenov: This is workload-bound. Gateways always have this enabled, so this only affects sidecars.

kyessenov:
need lgtm

PR #527, opened 2018-06-11
[WIP]: Add a gRPC protocol envelope generator

ozevren:

PR #531, opened 2018-06-12
Add core MCP protocol protos.

ozevren:
These are the bare-minimum core protos for the MCP protocol.


ozevren:
@andraxylia @ayj @geeknoid

config/mcp/v1alpha1/metadata.proto
```diff
@@ -0,0 +1,28 @@
...
+syntax = "proto3";
+
+package istio.config.mcp.v1alpha1;
+
+option go_package="istio.io/api/config/mcp/v1alpha1";
+
+// Metadata is the metadata information that all resources within the Mesh Configuration Protocol must have.
+message Metadata {
+
+	// The name of the resource. It is unique within the context of a resource type and the origin server of
```
geeknoid: Funy formatting here (tabs vs. spaces?)

config/mcp/v1alpha1/resource.proto
```diff
@@ -0,0 +1,34 @@
...
+
+// This package defines the common, core types used by the Mesh Configuration Protocol.
+package istio.config.mcp.v1alpha1;
+
+import "google/protobuf/any.proto";
+import "config/mcp/v1alpha1/metadata.proto";
+
+option go_package="istio.io/api/config/mcp/v1alpha1";
+
+// Resource is an envelope message that encapsulates the contents of a resource, and the associated
```
geeknoid: This reads a bit strange, and so does the comment on contents below.

Is this message a resource, or does it carry a resource?

ozevren: Updated. Does this look better?

config/mcp/v1alpha1/resource.proto
```diff
@@ -0,0 +1,34 @@
...
+
+// This package defines the common, core types used by the Mesh Configuration Protocol.
+package istio.config.mcp.v1alpha1;
+
+import "google/protobuf/any.proto";
+import "config/mcp/v1alpha1/metadata.proto";
+
+option go_package="istio.io/api/config/mcp/v1alpha1";
+
+// Envelope message for transferring configuration resources via Mesh Configuration Protocol. It contains
```
geeknoid: A configuration resources as transferred via the Mesh Configuration Protocol.

Each resource is made up of common metadata, and a type-specific payload.

config/mcp/v1alpha1/resource.proto
```diff
@@ -0,0 +1,34 @@
...
+// both the configuration resource, as well as the associated metadata.
+message Resource {
+
+    // Metadata about the config resource.
```
geeknoid: Common metadata describing this resource.

config/mcp/v1alpha1/resource.proto
```diff
@@ -0,0 +1,34 @@
...
+    istio.config.mcp.v1alpha1.Metadata metadata = 1;
+
+    // The contents of the config resource.
```
geeknoid: The type-specific payload for this resource.

ozevren:
One more /lgtm & /approve?

config/mcp/v1alpha1/resource.proto
```diff
@@ -0,0 +1,34 @@
...
+import "config/mcp/v1alpha1/metadata.proto";
+
+option go_package="istio.io/api/config/mcp/v1alpha1";
+
+// A configuration resource as transferred via the Mesh Configuration Protocol.
+// Each resource is made up of common metadata, and a type-specific payload.
+message Resource {
+
+    // Common metadata describing this resource.
+    istio.config.mcp.v1alpha1.Metadata metadata = 1;
```
costinm: I'm still not convinced this is the right way - we have the ADS discovery request containing an Any containing another Any. 

Did you discuss the issue with Envoy, and they refused to add it to ADS ?

ozevren: I thought we decided that the protocol would be based on ADS, not necessarily a strict adherence to ADS? Adding to upstream ADS seems like a very arduous endeavor, and waiting on it to be implemented and percolated down would be an unnecessary delay on our work. We can still ask Envoy folks to incorporate some metadata model, and switch over to that when it percolates down, but we still need to make forward progress.

Of all the choices (add metadata field to every config proto, add an envelope proto for every config proto, do any-in-any encoding), this is the most straightforward thing to implement. It is probably the easiest one to pivot from as well. (I've actually experimented with envelope generation here: https://github.com/istio/api/pull/527/files).

config/mcp/v1alpha1/metadata.proto
```diff
@@ -0,0 +1,28 @@
...
+package istio.config.mcp.v1alpha1;
+
+option go_package="istio.io/api/config/mcp/v1alpha1";
+
+// Metadata information that all resources within the Mesh Configuration Protocol must have.
+message Metadata {
+
+    // The name of the resource. It is unique within the context of a resource type and the origin server of
+    // the resource. The resource type is identified by the TypeUrl of the contents field of the Resource
+    // message.
```
andraxylia: Is this the FQDN? Can you give an example in the comments?

config/mcp/v1alpha1/metadata.proto
```diff
@@ -0,0 +1,28 @@
...
+	string name = 1;
+}
```
andraxylia: We need to add the source of the config in the metadata.

config/mcp/v1alpha1/resource.proto
```diff
@@ -0,0 +1,34 @@
...
+
+    // The type-specific payload for this resource.
+    google.protobuf.Any contents = 2;
```
andraxylia: Contents is too generic. Can we rename Resource to MCPResource and contents to resource?

config/mcp/v1alpha1/resource.proto
andraxylia: Great idea. let's follow up with Envoy @htuch and see if it possible to extend.

Since we need support for multiple sources, we need to add at a minimum the source of the config, and having this plus other non-configurable fields as a separate "metadata" is preferable to adding it to every resource, since we want to maintain the separation between what is user configurable/configured and what is added by the system.

config/mcp/v1alpha1/resource.proto
ayj: The package name includes MCP, i.e. `istio.io.api.config.mcp.v1alpha1.Resource`. Prefixing the message will lead to stutter in usage. 

```golang
import (
    mcppb "istio.io/api/config/mcp/v1alpha1"
)

var resource = &mcppb.Resource{ /* ... */ }
// vs. 
var resource = &mcppb.MCPResource{ /* ... */ }
```

config/mcp/v1alpha1/metadata.proto
ayj: Isn't source implied by the client opening different streams to different servers (i.e.g sources)? What additional source information do we need to include here?

config/mcp/v1alpha1/resource.proto
ozevren: +1. Resource is unambiguous in this context. We can rename it to ProtocolResource or something similar if this is a major concern.

config/mcp/v1alpha1/metadata.proto
ozevren: I've explicitly avoided giving an example to avoid giving the impression to that there is a well-supported structure to the name. (Yes the code today may need tread this is FQDN and parse , but I am hoping that that is a point-in-time issue). Config stored in non-K8s environments may not be FQDN.

config/mcp/v1alpha1/metadata.proto
ozevren: Let's check-in this as is. I was about to add the source field and send an updated PR, but I don't think we have a common understanding of what this field is about (e.g. is it preserved transitively? Is it just an in-memory field for Pilot et.al. to use ephemerally? Does it indicate Galley instance, origin cluster, or origin source control system? etc.)

ozevren: I'll follow up and do a follow-up PR after this.

andraxylia: Ideally it should indicate al, that is Galley, that is a certain cluster (local/remote) etc. Do not worry yet about what Pilot does with it, it will of course be in memory with the rest of the info.

config/mcp/v1alpha1/resource.proto
andraxylia: Why not call the inner thing resource and the outer Envelope? I really do not like content, it gets mixed with http content, which means the media type of resources (text, video, image).

andraxylia:
Please consider renaming the inner object to Resource and the outer to Envelope. I really do not like content, it gets mixed with http content, which means the media type of resources (text, video, image).

config/mcp/v1alpha1/metadata.proto
ozevren: Of course I am worried about what Pilot does with it. :)

if this is only a convenience field that Pilot will use to store its own ephemeral data, then it is fine. I can add a comment indicating as much.

However, if the Galley is on the hook for populating / preserving the contents of this field, then I'd like to understand what this is and how it is going to be used.

andraxylia: The question was about if it stored in memory, which of course it is. We would need the source info for the merging logic in Pilot, and likely for the multi-cluster/multi-cloud implementation.

Not sure what is origin source control system.

ozevren: ```if this is only a convenience field that Pilot will use to store its own ephemeral data, then it is fine.```

The question is not about whether the data is stored in memory or not. It is about the use case. One valid use-case is Pilot associating its own internal metadata here that doesn't escape protocol boundaries. Then the data in this field is ephemeral, and it is a convenience field for Pilot's use.

But based on your answer, it looks like Pilot is planning to obtain this data via the protocol when receiving the metadata and plans to perform merging based on that. Great. So, what information should go in the field? Simply: what the comment on the field should read?

PR #532, opened 2018-06-14
add match_sni boolean field to TCPRoute

vadimeisenbergibm:
For a TCP route, Pilot should know if it is TLS with SNI, or some other TCP protocol.
For TLS with SNI, the VirtualService can verify that the sent SNI matches the value
in the hosts field. For TCP which is not TLS with SNI, no such match can be performed.


rshriram:
Thanks. I was thinking of adding this actually.

networking/v1alpha3/virtual_service.proto
```diff
@@ -404,17 +405,21 @@ message HTTPRoute {
 //           number: 5555
 // ```
 message TCPRoute {
+  // Match the SNI of the TLS connection with the hosts value of the VirtualService.
+  // The semantics of match_sni and match variable is AND.
+  bool match_sni = 1;
+
```
rshriram: this should go into the L4MatchAttributes as check_tls_sni.

frankbu: I agree that it's better to put this inside the `match`. I'd prefer just simple `tls_sni` vs. `check_tls_sni` though, because "check" is kind of redundant, since it's already implied by "match".

I also wonder if there's a use case for making it more than just a true/false value, that is, match if the sni matches the value of something other than the host?

rshriram:
so I realized that this PR is actually not necessary. Within a mesh, we are doing istio auth. So SNI is not useful. Traffic to external services needs to be configured using Service entries. and service entries have a way of specifying whether the port is a TCP or TCP-TLS port (latter not implemented). And if there is a service entry, you do not need to add a corresponding virtual service (its redundant. Every service (incl service entry) gets default routes).

So, we don't need this extra check at all, right?

louiscryan:
Want to make sure I understand how we expect someone to use the API in this form. I assume the primary use-case is ingress based SNI routing.

If so then we expect a Gateway to be declared that
- declares ports for TCP that do not terminate TLS, i.e have no cert binding
- declares that those ports use this SNI-routed-TCP variant (do we call this a protocol?)

We also expect a VirtualService to be declared for services names that we want to be routed using this mechanism that:
- define a binding for this SNI-routed-TCP protocol so the Gateway can use it

We need the ability to control which protocols are available on a Gateway, especially tunneling protocols like this or PROXY. We do that today by declaring protocols on ports in the Gateway spec.

I like the idea of using VirtualService to bind service+protocol pairs to Gateways explicitly though there may be some usabilty gains in allowing the publication of all services using the protocol.

vadimeisenbergibm:
@rshriram @louiscryan The current use case is directing TLS traffic thru Egress Gateway for external services. You define a VirtualService to direct the traffic to the egress gateway, see the HTTP equivalent here - https://github.com/istio/istio.github.io/pull/1496.

networking/v1alpha3/virtual_service.proto
vadimeisenbergibm: @rshriram @frankbu OK, let me add `tls_sni` flag to `L4MatchAttributes`.

rshriram:
closing in favor of #549

PR #535, opened 2018-06-14
Fix envoy_filter Struct generation issue.

ozevren:
Added a golang => gogo mapping for the struct as well. This helps with the generation of the struct related code.


ozevren:
@rshriram

ozevren:
@geeknoid

PR #537, opened 2018-06-15
[Proposal] Generate README.md files for all packages

douglas-reid:
This PR is a proposal for generating a `README.md` file for each package in the api repo. The idea behind this PR is to make it easier to browse through the proto files from the github web UI and from preview windows in IDEs. These docs are intended only for quick reference when browsing the repo.

A few notes about the tool used:
- The tool used is a fork (in my personal space) of an existing `proto-gen-doc` tool. I had to fork it because it did not properly account for new lines in field descriptions, leading to broken tables.
- It uses html templates as the basis for the markdown conversion. This leads to some problematic conversions in some places, especially related to character escaping. This might be acceptable -- especially as a first pass.
- If this is generally useful, we can add markdown capabilities to the existing istio docgen tool.

### DIFF
--- a/authentication/v1alpha1/README.md
+++ b/authentication/v1alpha1/README.md
@@ -1,0 +1,322 @@
+# Protocol Documentation
+<a name="top"/>
+
+## Table of Contents
+
+- [authentication/v1alpha1/policy.proto](#authentication/v1alpha1/policy.proto)
+    - [Jwt](#istio.authentication.v1alpha1.Jwt)
+    - [MutualTls](#istio.authentication.v1alpha1.MutualTls)
+    - [OriginAuthenticationMethod](#istio.authentication.v1alpha1.OriginAuthenticationMethod)
+    - [PeerAuthenticationMethod](#istio.authentication.v1alpha1.PeerAuthenticationMethod)
+    - [Policy](#istio.authentication.v1alpha1.Policy)
+    - [PortSelector](#istio.authentication.v1alpha1.PortSelector)
+    - [TargetSelector](#istio.authentication.v1alpha1.TargetSelector)
+  
+    - [MutualTls.Mode](#istio.authentication.v1alpha1.MutualTls.Mode)
+    - [PrincipalBinding](#istio.authentication.v1alpha1.PrincipalBinding)
+  
+  
+  
+
+- [Scalar Value Types](#scalar-value-types)
+
+
+
+<a name="authentication/v1alpha1/policy.proto"/>
+<p align="right"><a href="#top">Top</a></p>
+
+## authentication/v1alpha1/policy.proto
+
+
+
+<a name="istio.authentication.v1alpha1.Jwt"/>
+
+### Jwt
+JSON Web Token (JWT) token format for authentication as defined by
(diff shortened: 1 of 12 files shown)


broker/dev/README.md
```diff
@@ -0,0 +1,195 @@
+# Protocol Documentation
+<a name="top"/>
+
+## Table of Contents
+
+- [broker/dev/service_class.proto](#broker/dev/service_class.proto)
+    - [CatalogEntry](#istio.broker.dev.CatalogEntry)
```
geeknoid: You might need to canonicalize these IDs, . are not allowed.

broker/dev/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+- [Scalar Value Types](#scalar-value-types)
+
+
+
+<a name="broker/dev/service_class.proto"/>
+<p align="right"><a href="#top">Top</a></p>
+
+## broker/dev/service_class.proto
+Copyright 2017 Istio Authors
```
geeknoid: Is outputting the copyright banner intentional?

broker/dev/README.md
```diff
@@ -0,0 +1,195 @@
...
+  distributed under the License is distributed on an &#34;AS IS&#34; BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+
+<a name="istio.broker.dev.CatalogEntry"/>
+
+### CatalogEntry
+$hide_from_docs
```
geeknoid: Should either respect this, or strip it.

authentication/v1alpha1/README.md
```diff
@@ -0,0 +1,322 @@
+# Protocol Documentation
+<a name="top"/>
```
geeknoid: What's this for?

broker/dev/README.md
douglas-reid: Probably not. I'm guessing the source project wasn't putting copyright in their protos ;)

authentication/v1alpha1/README.md
douglas-reid: beats me.  Maybe something else inserted a link to `#top` somewhere?

broker/dev/README.md
douglas-reid: ouch. ok.

Looks like the source code has too many minor issues to be to usable without modification.

geeknoid: proto-gen-docs has a canonicalizeID function that I use for this.

broker/dev/README.md
geeknoid: We also have $class which is used to control CSS classes on generated objects which would need to be handled.

I think it'd be consistent to skip the element and not emit it in the docs.

authentication/v1alpha1/README.md
geeknoid: There are many such dummy links throughout. Weird.

geeknoid:
Do we still want this PR?

douglas-reid:
I think so. It just fell down the Priority ladder.

geeknoid:
Old in the tooth, closing.

PR #541, opened 2018-06-15
Make MCP protos to use gogo.

ozevren:


ozevren:
@ayj

ozevren:
@geeknoid , a little push?

PR #546, opened 2018-06-19
SNI routing primitives

rshriram:
Treating SNI routing as an explicit act rather than implicitly enabling it
by port names.


networking/v1alpha3/virtual_service.proto
```diff
@@ -151,13 +151,27 @@ message VirtualService {
   // specify `mesh` as one of the gateway names.
   repeated string gateways = 2;
 
-  // An ordered list of route rules for HTTP traffic.
-  // The first rule matching an incoming request is used.
+  // An ordered list of route rules for HTTP traffic. HTTP routes will be
+  // applied to platform service ports named http-*/http2-*/grpc-*, gateway
+  // ports with protocol HTTP/HTTP2/GRPC/HTTPS and service entry ports
```
louiscryan: Note conflict with HTTPS here and mention below.

Suggest differentiating as "TLS terminated HTTPS" vs "unterminated TLS"

networking/v1alpha3/virtual_service.proto
```diff
@@ -151,13 +151,27 @@ message VirtualService {
...
+  // using HTTP/HTTP2/GRPC protocols.  The first rule matching an incoming
+  // request is used.
   repeated HTTPRoute http = 3;
 
-  // An ordered list of route rules for TCP traffic.
-  // The first rule matching an incoming request is used.
-  repeated TCPRoute tcp = 4;
+  // An ordered list of route rules for TCP+TLS or HTTPS traffic. Routing
```
louiscryan: Alternate phrasing suggestion

"An ordered list of route rule for non-terminated TLS & HTTPS traffic..."

networking/v1alpha3/virtual_service.proto
```diff
@@ -151,13 +151,27 @@ message VirtualService {
...
+  // is typically performed using the SNI value presented by the
+  // ClientHello message. TLS routes will be applied to platform service
```
louiscryan: I wonder if we should just call the protocol tls instead of TCP_TLS. I think it reads better and doesn't necessarily imply the underlying protocol. Also the tunneling protocol should be listed first so for instance we can have 'tls-mongo' as a port label and have it work with tls-* matching rules

"TLS routes can be applies to service ports labelled tls-* & https-*, ...

networking/v1alpha3/virtual_service.proto
```diff
@@ -380,10 +394,55 @@ message HTTPRoute {
...
+// the reviews.prod.svc.cluster.local service.
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: VirtualService
+// metadata:
+//   name: bookinfo-sni
+// spec:
+//   hosts:
+//   - reviews.prod.svc.cluster.local
```
louiscryan: SNI host should be in agreement with outer host right ??

The more useful example would be outer host is wildcard and SNI host is a sub-domain for routing

networking/v1alpha3/virtual_service.proto
```diff
@@ -380,10 +394,55 @@ message HTTPRoute {
...
+//   - match:
+//     - port: 443
+//       sni: reviews
+//     route:
+//     - destination:
+//         host: reviews.prod.svc.cluster.local
+//         port:
+//           number: 443
+// ```
+message TLSRoute {
```
louiscryan: @PiotrSikora should we include ALPN in here at some point ?

networking/v1alpha3/virtual_service.proto
```diff
@@ -628,6 +687,36 @@ message L4MatchAttributes {
...
+
+  // Specifies the port on the host that is being addressed. Many services
+  // only expose a single port or label ports with the protocols they support,
+  // in these cases it is not required to explicitly select the port.
+  uint32 port = 2;
+
+  // SNI (server name indicator) to match on. Wildcard prefixes can be used
+  // in the SNI value. E.g., *.com will match foo.example.com as well as
+  // example.com.
+  string sni = 3;
```
louiscryan: repeated ? Also sni_hosts to be consistent with terms.

networking/v1alpha3/virtual_service.proto
```diff
@@ -628,6 +687,36 @@ message L4MatchAttributes {
   repeated string gateways = 5;
 }
 
+// TLS connection match attributes.
+message TLSMatchAttributes {
+  // IPv4 or IPv6 ip address of destination with optional subnet.  E.g.,
+  // a.b.c.d/xx form or just a.b.c.d. This is only valid when the
+  // destination service has several IPs and the application explicitly
+  // specifies a particular IP.
+  string destination_subnet = 1;
```
louiscryan: Do we think this will be needed ? I guess eventually it might but I suspect we can leave it out for the moment. Same for port as the primary use case is Gateway binding.

louiscryan: I guess it does no harm and we can see if it gains traction.

Aside - We need a refined way to mark individual API features as alpha/beta

networking/v1alpha3/virtual_service.proto
```diff
@@ -703,18 +709,22 @@ message TLSMatchAttributes {
   // SNI (server name indicator) to match on. Wildcard prefixes can be used
   // in the SNI value. E.g., *.com will match foo.example.com as well as
   // example.com.
-  string sni = 3;
+  repeated string sni_hosts = 3;
+
+  // ALPN protocols to match on. Multiple ALPN protocols can be specified
```
louiscryan: @PiotrSikora 

lets leave this for a follow on PR. Needs some thought

PR #552, opened 2018-06-21
Add the Mesh Configuration Protocol (MCP) service.

ayj:
The MCP service is to be used as a environment neutral downward API
for some Istio runtime components. Motivation for the MCP service
along with high-level design proposals are documented in the following
two docs:

* Pilot Decomposition (see https://goo.gl/qAsEQx)
* Galley component config distirbution (see https://goo.gl/EJu5hg)

MCP is conceptually aligned with envoy project's xDS protocol and ADS
service (see
https://github.com/envoyproxy/data-plane-api/blob/master/XDS_PROTOCOL.md). The
proto definitions are derived from the envoy API definitions with
unnecessary or envoy specific messages and fields removed. Names have
also been changed to make it unambiguous with xDS protocol.

Fixes istio/istio#6474


ayj:
cc @ZackButcher @andraxylia

ozevren:
It seems some stuff needs to be generated. Otherwise lgtm.

config/mcp/v1alpha1/mcp.proto
```diff
@@ -0,0 +1,126 @@
...
+import "config/mcp/v1alpha1/envelope.proto";
+
+option (gogoproto.equal_all) = true;
+
+// Identifies a specific MCP client instance. The client identifier is
+// presented to the management server, which may use this identifier
+// to distinguish per client configuration for serving.
+message Client {
+  // An opaque identifier for the MCP client.
+  string id = 1;
```
geeknoid: How is this enforced? Shouldn't the client identity be determined by the transport (certs) rather than in the payload?

config/mcp/v1alpha1/mcp.proto
```diff
@@ -0,0 +1,126 @@
...
+
+  // Opaque metadata extending the client identifier. The client will
```
geeknoid: I don't understand the comment about "the client will pass this directly". Isn't this the client?

config/mcp/v1alpha1/mcp.proto
```diff
@@ -0,0 +1,126 @@
...
+  // pass this directly to the management server.
+  google.protobuf.Struct metadata = 2;
+
+  // This is motivated by informing a management server during canary
+  // which version of the client is being tested in a heterogeneous
+  // fleet. This will be set by the client in management server RPCs.
+  string build_version = 3;
```
geeknoid: Shouldn't this be labels instead? Labels to identity the client. Build version is only one possible criteria for selection.

config/mcp/v1alpha1/mcp.proto
ayj: Yes. These definitions and comments are derived from envoy's API [here](https://github.com/envoyproxy/data-plane-api/blob/4b04ff76110299a1210c93ededfbcfd0fa9db616/envoy/api/v2/core/base.proto#L40). `Client` (renamed from Node) was used elsewhere (e.g. bootstrap config) in which case the comment made more sense. I'll remove the 2nd sentence for now since we're only using `Client` over-the-wire.

config/mcp/v1alpha1/mcp.proto
ayj: Sounds reasonable. This was carried over from xDS but we can start with a pure label approach for now. We can add an explicit field later if we need a more formal contract between client/server.

config/mcp/v1alpha1/mcp.proto
```diff
@@ -0,0 +1,126 @@
...
+  // instance is ready to ACK/NACK the new configuration. ACK/NACK
+  // takes place by returning the new API config version as applied or
+  // the previous API config version respectively. Each type_url (see
+  // below) has an independent version associated with it.
+  string version_info = 1;
+
+  // The client making the request.
+  Client client = 2;
+
+  // $hide_from_docs
```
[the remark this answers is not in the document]
ayj: Removed

config/mcp/v1alpha1/mcp.proto
```diff
@@ -0,0 +1,126 @@
...
+  // List of resources to subscribe to. If this is empty, all
+  // resources for the API are returned. Reserved for future use to
+  // support incremental and on-demand (see
+  // https://github.com/envoyproxy/envoy/issues/3229)
+  // repeated string resource_names = 3;
+  reserved 3;
```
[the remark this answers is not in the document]
ayj: Yup, removed.

config/mcp/v1alpha1/mcp.proto
```diff
@@ -0,0 +1,126 @@
...
+
+  // Type of the resource that is being requested, e.g.
+  // "type.googleapis.com/istio.io.networking.v1alpha3.VirtualService".
+  string type_url = 4;
```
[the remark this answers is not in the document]
ayj: Clients can _request_ any type. Galley is free to reject/filter/ignore requests based on ACL policy (TBD). We may, for example, need to add an `google.rpc.Status error_details` to the `MeshConfigResponse` to indicate unauthorized requests.

config/mcp/v1alpha1/mcp.proto
ayj: Added a comment to make it clear that this entire client message is not authoritative. The underlying transport credentials (e.g. [TLS / Token](https://grpc.io/docs/guides/auth.html#supported-auth-mechanisms)) should be used instead.

ayj:
@googlebot

PR #556, opened 2018-06-24
add server type to gateway server 

vadimeisenbergibm:
to distinguish between ingress and egress

For egress, Citadel mTLS should be enabled for incoming connections, since
the incoming connections are from the sidecar proxies in the mesh.


rshriram:
You don’t need additional server types. You declare the server to use istio mTLS certs (ie specify the cert path where istio certs are stored). Keep in mind that ingress and egress are not the only gateways possible. I could have a purely internal gateway per namespace or gateway between different platforms in same mesh etc.

costinm:
+1 on Shriram comments.  

You can also add an option when starting a gateway - this is not an injection case so you can mount or set flags easily.

vadimeisenbergibm:
closing, agree with the comments

PR #557, opened 2018-06-27
We don't publish API proto docs to istio.io

geeknoid:


mixer/adapter/model/v1beta1/infrastructure_backend.proto
```diff
@@ -16,7 +16,6 @@ syntax = "proto3";
 
 // $title: Mixer Adapter Model
 // $description: Definitions used to create adapters and templates
-// $location: https://istio.io/docs/reference/api/mixer/istio.mixer.adapter.model.v1beta1.html
```
frankbu: Has something changed? I thought this is the way to populate the reference section of the docs. What is special about this proto file? Do we need to remove the $location field from all the protos now? For example, https://github.com/geeknoid/api/blob/73ca97549e17e50a18f837659a28cc40543d813b/networking/v1alpha3/virtual_service.proto#L22

frankbu: Oh, never mind. I see, you mean reference/api.

PR #562, opened 2018-07-02
Fix consistent hash LB and TLS docs

rshriram:
Mark certain fields as required. Add Cookie based consistent hash.


networking/v1alpha3/virtual_service.proto
```diff
@@ -427,10 +427,10 @@ message HTTPRoute {
 //         host: reviews.prod.svc.cluster.local
 // ```
 message TLSRoute {
-  // Match conditions to be satisfied for the rule to be activated. All
-  // conditions inside a single match block have AND semantics, while the
-  // list of match blocks have OR semantics. The rule is matched if any one
-  // of the match blocks succeed.
+  // REQUIRED. Match conditions to be satisfied for the rule to be
```
ZackButcher: `Match` is not required since empty match = true, right?

networking/v1alpha3/virtual_service.proto
```diff
@@ -689,33 +689,38 @@ message L4MatchAttributes {
...
-  uint32 port = 2;
+  uint32 port = 3;
 
-  // SNI (server name indicator) to match on. Wildcard prefixes can be used
-  // in the SNI value. E.g., *.com will match foo.example.com as well as
-  // example.com.
-  repeated string sni_hosts = 3;
+  // IPv4 or IPv6 ip address of source with optional subnet. E.g., a.b.c.d/xx 
+  // form or just a.b.c.d
+  // $hide_from_docs
```
ijsnellf: why hide?

rshriram: coz we don't implement it?

networking/v1alpha3/virtual_service.proto
rshriram: YEs, except for TLS.. coz empty match would make it same as TCP without match.

networking/v1alpha3/virtual_service.proto
ijsnellf: `destination_subnet` is also unimplemented at the moment

PR #564, opened 2018-07-02
update minimum_ring_size to uint64

utako:
to be consistent with envoy v2 api minimum_ring_size:
https://github.com/envoyproxy/envoy/blob/79bce5fe1cd8d1ab03dc6085497fcda653320a67/api/envoy/api/v2/cds.proto#L351-L358


kyessenov:
It's not OK to update the type of proto fields, especially right at the time of release of 1.0.
Proto fields are forever.
/cc @costinm

ZackButcher:
That's only true if the protos are actually persisted somewhere that depends on that encoding. Given that this is end-user facing config typically stored as text, int32->int64 is totally safe and as far as we know the only code depending on this field being an int32 is in Pilot itself.

PR #565, opened 2018-07-03
Add a Readme.md file for the MCP protocol.

ozevren:


config/mcp/Readme.md
```diff
@@ -0,0 +1,12 @@
+# Mesh Configuration Protocol Protos
```
ayj: This is a good start. We should also document the protocol semantics and/or reference https://github.com/envoyproxy/data-plane-api/blob/master/XDS_PROTOCOL.md#streaming-grpc-subscriptions with notes where the protocol diverges.

ozevren:
Great idea. I'll leave that part upto you. ;)

I'll add a link to the XDS protocol page.

ayj:
Opened https://github.com/istio/api/issues/567 to track follow-up documentation improvements.

PR #566, opened 2018-07-03
Add two new attributes: request.url_path and request.query_params

bianpengyuan:
The two attributes will be populated by https://github.com/istio/proxy/pull/1837

cc @douglas-reid @kyessenov @qiwzhang

### DIFF
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -68,6 +68,8 @@
 - request.size
 - request.time
 - request.useragent
+- request.queries
+- request.url_path
 - response.headers
 - response.size
 - response.time
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -291,7 +291,7 @@
 - connection.requested_server_name
 
 # url_path and query_params are derived from http path header.
-# url_path is the path part of a url, with query params being stripped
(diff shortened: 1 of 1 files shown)


mixer/v1/global_dictionary.yaml
```diff
@@ -68,6 +68,8 @@
 - request.size
 - request.time
 - request.useragent
+- request.queries
```
douglas-reid: This file is append-only, unfortunately.

kyessenov:
These are traditionally called query parameters, not queries. There is one
query string per request, several components in it, first one being path
and the remainder are parameters

mixer/v1/global_dictionary.yaml
bianpengyuan: Moved them to the end of file.

bianpengyuan:
@kyessenov @douglas-reid I've updated `request.query` to `request.query_params`. Let me know how you think about this.

kyessenov:
Since this PR is parked, do you mind adding two keywords: "inbound" and "outbound". They would help compress reporter kind strings.

wora:
Please document whether the string is normalized based on spec. For
example, + vs %20.

bianpengyuan:
@wora comment updated, thanks!. @kyessenov @douglas-reid Please take another look, thanks!

PR #568, opened 2018-07-03
source_ip field is bool

zachgersh:
- on the envoy side the source_ip field is
used as an indicator to determine whether or not
to use it as the hash value.

https://github.com/envoyproxy/envoy/blob/2c3c3e7546451a03cf4b7e9036ee48dda26fe49c/api/envoy/api/v2/route/route.proto#L535


zachgersh:
This should be safe due to the fact that this field was just added on 07/02 and isn't a part of an official release yet.

networking/v1alpha3/destination_rule.proto
```diff
@@ -302,7 +302,7 @@ message LoadBalancerSettings {
       HTTPCookie http_cookie = 2;
 
       // Hash based on the source IP address.
-      string source_ip = 3;
+      bool source_ip = 3;
```
rshriram: change to use_source_ip

PR #572, opened 2018-07-05
Bump release-1.0

zachgersh:
When we cut this branch we missed the `source_ip` field change that is actually required.


networking/v1alpha3/destination_rule.proto
```diff
@@ -302,7 +302,7 @@ message LoadBalancerSettings {
       HTTPCookie http_cookie = 2;
 
       // Hash based on the source IP address.
-      string source_ip = 3;
+      bool use_source_ip = 3;
```
andraxylia: The title does not seem to match the changes.

Why is this field changed?

Thanks for fixing all the typos.

zachgersh: Hey andra - this was a bump to grab everything from master that was missed when we cut release 1.0. That `use_source_ip` is part of another PR I am making. I don't believe anyone was using this field in the wild since @rshriram introduced it in another PR. This is all in service of finishing out session affinity `ring_hash`

rshriram: Envoy treats this as bool.. We set it as string by mistake.

PR #578, opened 2018-07-11
Fix improper example in documentation

metacosm:


metacosm:
I signed it!

metacosm:
Actually, this is wrong, just a problem parsing YML on my side… :(

PR #579, opened 2018-07-11
Adding creation_timestamp to ServiceEntry proto

nmittler:
This is to support an on-going PR: https://github.com/istio/istio/pull/6993


nmittler:
FYI @rshriram @costinm

nmittler:
/hold

I'm actually not certain we need this.  Will follow up shortly.

kyessenov:
Why can't you use CRD timestamp? It's going to be super confusing to have creation timestamps within creation timestamps.

rshriram:
Yes. CRD timestamps should be sufficient. Closing for the moment

PR #580, opened 2018-07-12
Update RBAC comments for Istio 1.0.

yangminzhu:
istio/istio#6377

### DIFF
--- a/rbac/v1alpha1/rbac.proto
+++ b/rbac/v1alpha1/rbac.proto
@@ -172,8 +172,8 @@ message Subject {
   // Optional. The user name/ID that the subject represents.
   string user = 1;
 
+  // $hide_from_docs
   // Optional. The group that the subject belongs to.
-  // [#not-implemented-hide:]
   string group = 2;
 
   // Optional. The set of properties that identify the subject.
--- a/rbac/v1alpha1/rbac.proto
+++ b/rbac/v1alpha1/rbac.proto
@@ -16,52 +16,28 @@ syntax = "proto3";
 
 // $title: RBAC
 // $description: Configuration for Role Based Access Control
-// $location: https://istio.io/docs/reference/config/istio.rbac.v1alpha1.html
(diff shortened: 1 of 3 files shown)


rbac/v1alpha1/rbac.proto
```diff
@@ -192,21 +166,18 @@ message ServiceRoleBinding {
...
-// the "properties" in the "subject" part of the "authorization" template
-// (https://github.com/istio/istio/blob/master/mixer/template/authorization/template.proto).
+// Subject defines an identity. The identity is either a user or identified by a set of `properties`.
+// The supported keys in `properties` are listed in "constraint and properties" page.
 message Subject {
   // Optional. The user name/ID that the subject represents.
   string user = 1;
 
   // Optional. The group that the subject belongs to.
+  // [#not-implemented-hide:]
```
geeknoid: To hide this from the generated docs, you need to use $hide_from_docs instead. You can see what is generated in the pb.html file.

yangminzhu: Done. (It seems `[#not-implemented-hide:]` also works)

rbac/v1alpha1/rbac.proto
```diff
@@ -232,10 +203,10 @@ message RoleRef {
...
 // ```yaml
-// apiVersion: "config.istio.io/v1alpha1"
+// apiVersion: "config.istio.io/v1alpha2"
 // kind: RbacConfig
 // metadata:
-//   name: istio-rbac-config
+//   name: default
 //   namespace: istio-system
 // spec:
 //   mode: ON_WITH_INCLUSION
```
liminw: Missing "inclusion" parameter here?

yangminzhu: I think it's just hided from the github page, you can find it here: https://github.com/istio/api/blob/f39acdf001b3a425a862a5c23f9e591daddccdab/rbac/v1alpha1/rbac.proto#L213

PR #581, opened 2018-07-13
remove legacy routing apis

ayj:


geeknoid:
@rshriram We should also simplify the title of the new rules so they look nicer on the web site.

PR #582, opened 2018-07-13
Bring back array of destination subnets in L4/TLS Match

rshriram:
This aligns well with the filter chain match in Envoy that
takes multiple destination subnets.


networking/v1alpha3/gateway.proto
```diff
@@ -143,7 +143,6 @@ option go_package = "istio.io/api/networking/v1alpha3";
 //   tcp:
 //   - match:
 //     - port: 27017
-//       sourceSubnet: "172.17.16.0/24"
```
frankbu: I assume you are taking sourceSubnet out of the example just because it's not supported yet?

networking/v1alpha3/virtual_service.proto
```diff
@@ -659,11 +659,9 @@ message DestinationWeight {
...
 // is incomplete.
 message L4MatchAttributes {
-  // IPv4 or IPv6 ip address of destination with optional subnet.  E.g.,
-  // a.b.c.d/xx form or just a.b.c.d. This is only valid when the
-  // destination service has several IPs and the application explicitly
-  // specifies a particular IP.
-  string destination_subnet = 1;
+  // IPv4 or IPv6 ip addresses of destination with optional subnet.  E.g.,
+  // a.b.c.d/xx form or just a.b.c.d.
+  repeated string destination_subnet = 1;
```
frankbu: should `source_subnet` be changed to repeated as well?

networking/v1alpha3/gateway.proto
rshriram: Yep

networking/v1alpha3/virtual_service.proto
rshriram: I thought of it. but envoy has non repeated source subnet. Thought I will keep these two in sync

PR #584, opened 2018-07-16
destination_subnet to destination_subnets

rshriram:


frankbu:
There is some inconsistency in the pluralizing of repeated fields (e.g., `repeated route:`)

PR #591, opened 2018-07-24
Update Protoc Dockerfile with more protobuf dependencies

gargnupur:
Update Protoc Dockerfile with more protobuf dependencies needed for go generate in istio repository

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,6 +36,15 @@ RUN go get -u -v -ldflags '-w -s' \
         github.com/gogo/protobuf/protoc-gen-gogoslick \
         github.com/istio/tools/protoc-gen-docs \
         && install -c ${GOPATH}/bin/protoc-gen* ${OUTDIR}/usr/bin/
+RUN go get -u -v -ldflags '-w -s' \
+        github.com/gogo/protobuf/proto \
+        && install -c ${GOPATH}/bin/proto ${OUTDIR}/usr/bin/
+RUN go get -u -v -ldflags '-w -s' \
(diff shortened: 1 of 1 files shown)


ozevren:
/lgtm
@geeknoid , any objections?

docker/Dockerfile
```diff
@@ -36,6 +36,7 @@ RUN go get -u -v -ldflags '-w -s' \
         github.com/gogo/protobuf/protoc-gen-gogoslick \
         github.com/istio/tools/protoc-gen-docs \
         && install -c ${GOPATH}/bin/protoc-gen* ${OUTDIR}/usr/bin/
+RUN go get -u -v github.com/gogo/protobuf/jsonpb
```
kyessenov: I think we should use released versions of gogo. Master is not stable and does not always cooperate with grpc/standard proto (e.g. the table driven-serialization took awhile to sort out across all three).

gargnupur:
@douglas-reid : I added jsonpb because it was suggested here https://github.com/gogo/protobuf#most-speed-and-most-customization. It didn't really have any jsonpb binary, hence step like this "install -c ${GOPATH}/bin/protoc-gen* ${OUTDIR}/usr/bin/" was missing.
I ran go generate using this protoc in istio directory without any problem... so will skip this PR for now.

docker/Dockerfile
gargnupur: Using docker file should help cut down a lot of this pain I think as versions will get locked when we build this file.

@kyessenov: Please let me know what you think an I can probably try it in a separate PR... taking back this PR for now..

PR #594, opened 2018-07-25
jwt: add a excluded_path to jwt policy for health check support.

yangminzhu:
istio/istio#7264


authentication/v1alpha1/policy.proto
```diff
@@ -23,6 +23,25 @@ package istio.authentication.v1alpha1;
 
 option go_package = "istio.io/api/authentication/v1alpha1";
 
+// Describes how to match a given string in HTTP headers. Match is case-sensitive.
+message StringMatch {
```
diemtvu: Do we have smt similar in RBAC? Can we make it re-usable?

yangminzhu: No, currently the rbac.proto only uses native `string` with `*` character to indicate prefix/suffix match.

How about let me move this to a dedicated directory so it's easier to reuse later?

yangminzhu: Moved to `common/v1alpha1/string.proto`, PTAL.

PR #596, opened 2018-07-27
Remove auth_policy from MeshConfig

diemtvu:
Completely remove auth_policy from mesh config. Data plane mTLS enablement should be done using authentication policy.

Cleanup on pilot side is in PR https://github.com/istio/istio/pull/7456

Issue: https://github.com/istio/istio/issues/5900


rshriram:
Fix conflict.

costinm:
/hold

This will break upgrade - needs to be gradually, first make sure there is a way to ignore, deprecate, etc.

diemtvu:
@costinm, I have https://github.com/istio/istio/pull/7456 that remove usage of this flag ((I do plan to submit this one after that PR as well). In 1.0, we already not use this flag in installation. Do you suggest to wait for one more release cycle, or just after the above PR is ok?

diemtvu:
@constinm, #7456 (remove all refs of this flag in istio) has been submitted. Can I proceed with this one?

rshriram:
Yeah. it was marked as deprecated long ago. The sooner we get rid of rhis rhe better. Can you confirm that setting this field doesn't cause issues? i.e. old config map with these fields set - the unknown fields should be ignored..

diemtvu:
It will fail if the config map has that field. Though in 1.0 (even 0.8), we do not use that flag any more.

diemtvu:
Withdrawn after discussion with @costinm. The field is already marked deprecated, so it's good enough for now.

PR #597, opened 2018-07-30
Set default resolution mode to invalid

rshriram:
This is a last minute change. The rationale for this is that
people are creating service entries without setting a resolution
(which defaults to NONE -- original dst cluster). This ends up
opening a huge hole that lets all traffic through.
E.g., 
```yaml
host: google.com
port:
 443
```
The above will result in a listener on 0.0.0.0:443 with
a original dst cluster that allows traffic to any IP:443.


networking/v1alpha3/istio.networking.v1alpha3.pb.html
```diff
@@ -2817,8 +2817,9 @@ <h2 id="ServiceEntry">ServiceEntry</h2>
 <td><code>resolution</code></td>
 <td><code><a href="#ServiceEntry-Resolution">ServiceEntry.Resolution</a></code></td>
 <td>
-<p>Service discovery mode for the hosts. If not set, Istio will attempt
-to infer the discovery mode based on the value of hosts and endpoints.</p>
+<p>REQUIRED. Service discovery mode for the hosts. If not set, Istio will
```
frankbu: `If not set, ...` isn't possible if the field is `REQUIRED`

PR #601, opened 2018-07-31
Fix description of `allow_headers`

MrSaints:


MrSaints:
Done :) Thanks @frankbu!

PR #603, opened 2018-08-02
use gogoproto for all generated *.proto code

ayj:


ayj:
cc @liminw (rbac)
cc @costinm (mesh config)

PR #605, opened 2018-08-07
Add proxy_error_code

mandarjog:
Proxy returns error codes generated by itself to Mixer.


mandarjog:
Do I need to run generation scripts when updating `mixer/v1/global_dictionary.yaml`, other PRs that touch this file did not update anything else.

mixer/v1/global_dictionary.yaml
```diff
@@ -295,3 +295,25 @@
...
+- NR
+- LH
+- UT
+- LR
+- UR
+- UC
+- DI
+- FI
+- RL
+- "-"
```
yangminzhu: @mandarjog 
Line 316 `- "-"` break the proxy build as [create_global_dictionary.py](https://github.com/istio/proxy/blob/4865583db000db91be0469a7f1756cc98fea9992/src/istio/mixerclient/create_global_dictionary.py#L65) parse it to: `    ""-"",` which is invalid in .cc.

Could you suggest the proper way to fix this? Should we just remove the `"` or make it `\"-\"` or even update the script?

PR #607, opened 2018-08-09
Update JWT to support per-path validation.

yangminzhu:


authentication/v1alpha1/policy.proto
```diff
@@ -127,10 +127,9 @@ message Jwt {
   // For example, `query=jwt_token`.
   repeated string jwt_params = 7;
 
-  // URL paths that should be excluded from the JWT validation. If the request path is matched,
-  // the JWT validation will be skipped and the request will proceed regardless.
+  // URL paths that should be excluded from the JWT validation. If the request path is
+  // matched, the JWT validation will be skipped and the request will proceed regardless.
   // This is useful to keep a couple of URLs public for external health checks.
-  // Example: "/health_check", "/status/cpu_usage".
   repeated istio.common.v1alpha1.StringMatch excluded_paths = 8;
```
diemtvu: May be we should have both `included` and `excluded` rules, and the final trigger is ORs of `included_path` AND NOT (ORs of `excluded_path`)  with exception included_path.empty() it's simply NOT (ORs of `excluded path`)  (i.e include_path.empty() is treated as match all)

authentication/v1alpha1/policy.proto
```diff
@@ -127,10 +127,9 @@ message Jwt {
   // For example, `query=jwt_token`.
   repeated string jwt_params = 7;
 
-  // URL paths that should be excluded from the JWT validation. If the request path is matched,
-  // the JWT validation will be skipped and the request will proceed regardless.
+  // URL paths that should be excluded from the JWT validation. If the request path is
```
diemtvu: More accurate comment could be ` If trigger condition doesn't meet, authentication will skip the underlying JWT validation'

And in the origin block, add comment to mention JWT is used only when the trigger condition match. If none of them are trigger, request will be accepted as if there is no JWT defined.

authentication/v1alpha1/policy.proto
diemtvu: We may put them inside another message, say `TriggerCondition`

yangminzhu:
@diemtvu Please take a look, I have updated the design based on our discussions.

yangminzhu:
@diemtvu Updated all comments, also renamed to TriggerRule, PTAL.

PR #609, opened 2018-08-14
Deprecate int percent in favor of double percentage

venilnoronha:
This PR deprecates the integer `percent` field in `Delay` and `Abort` types in favor of the new double `percentage` field which allows finer control.

This PR serves as a prereq for the changes necessary for istio/istio#7903.

Signed-off-by: Venil Noronha <veniln@vmware.com>

/cc @rshriram


networking/v1alpha3/virtual_service.proto
```diff
@@ -1001,6 +1011,9 @@ message HTTPFaultInjection {
       // $hide_from_docs
       string http2_error = 4;
     }
+
+    // Percentage of requests to be aborted with the error code provided.
+    FractionalPercent percentage = 5;
```
rshriram: we shouldn't use the envoy model as it is here. Please use double or float and then derive the envoy level stuff automatically from the user specified value.

venilnoronha: Addressed in fe10f32.

networking/v1alpha3/virtual_service.proto
```diff
@@ -938,17 +938,18 @@ message HTTPFaultInjection {
   //         subset: v1
   //     fault:
   //       delay:
-  //         percent: 10
+  //         percentage: 0.000001
```
rshriram: Update the explanation above this example that says delay 10% of requests. It should say delay 1 out of every 1000 requests (and change the percentage to 0.001 everywhere). Do the same for all example texts that say 10%.

kyessenov:
Floats make me a bit uncomfortable. Note that these protos go through JSON translation that loses all number precision (since Javascript has only one number datatype), so that's some loss of precision during internal processing. I'd hate to find out that some small number like 0.000001 gets approximated to 0 internally in istio.

networking/v1alpha3/virtual_service.proto
venilnoronha: Addressed in 9628157.

wora:
Please consider using standard type defined by `envoy.type.Percent`
package. It was created for this use case. While it may not be perfect, it
is still better for Istio to align with Envoy.

venilnoronha:
I've addressed the changes via 35280a4 and aee3bd2.

wora:
Is there any reason we cannot use Envoy definition directly? I assume Istio
has to trust Envoy anyway.

venilnoronha:
As per the readme, this project should be independent of others.

Makefile
```diff
@@ -310,26 +310,39 @@ clean-authn:
...
-envoy_pb_gos := $(envoy_protos:.proto=.pb.go)
-envoy_pb_pythons := $(envoy_protos:.proto=_pb2.py)
-
-generate-envoy-go: $(envoy_pb_gos) $(envoy_pb_doc)
+envoy_types_protos := $(shell find $(envoy_path) -type f -path '*/types/*' -name '*.proto' | sort)
+envoy_types_pb_gos := $(envoy_types_protos:.proto=.pb.go)
+envoy_types_pb_pythons := $(envoy_types_protos:.proto=_pb2.py)
+envoy_v2alpha1_protos := $(shell find $(envoy_path) -type f -path '*/v2alpha1/*' -name '*.proto' | sort)
+envoy_v2alpha1_pb_gos := $(envoy_v2alpha1_protos:.proto=.pb.go)
+envoy_v2alpha1_pb_pythons := $(envoy_v2alpha1_protos:.proto=_pb2.py)
```
rshriram: ???

venilnoronha: See golang/protobuf#39.

rshriram: Revert all of this please

envoy/types/percent.proto
```diff
@@ -0,0 +1,31 @@
...
+
+// Using "types" for the package name as "type" is a keyword in Golang.
+option go_package = "istio.io/api/envoy/types";
+option (gogoproto.equal_all) = true;
+
+// Copied from @envoy/api/envoy/type/percent.proto
+// Identifies a percentage, in the range [0.0, 100.0].
+message Percent {
+  double value = 1;
+}
```
rshriram: revert

networking/v1alpha3/virtual_service.proto
```diff
@@ -958,11 +961,14 @@ message HTTPFaultInjection {
       // $hide_from_docs
       google.protobuf.Duration exponential_delay = 3 ;
     }
+
+    // Percentage of requests on which the delay will be injected.
+    istio.envoy.types.Percent percentage = 5;
```
rshriram: define a Percent message in this API and use it. Don't want to import from another packages. As such proto generated docs are not that friendly.

venilnoronha:
@rshriram all comments have been addressed via 81320ce and 6418b55.

PR #611, opened 2018-08-15
Revert "add rbac attributes to mixer global dict (#608)"

mandarjog:
Reverts istio/api#608

re-ordering of dictionary elements in not allowed.
Please append to the end.

@geeknoid


geeknoid:
Ouch, that didn't cross my mind. Sorry about that.

mandarjog:
np, That's why we need a build gate here!

PR #612, opened 2018-08-16
mixer: add header operation templates

kyessenov:
This is a minimal addition to expose routing features from Mixer.
I've put the combinatorial expression problem out of scope for now, and only added the header template and action name. 
With these two, we have a way to propagate adapter attributes out to the client. Header operations are easy to combine, so that is also orthogonal to the check rule combination discussion.

/assign @geeknoid 
/assign @mandarjog 


policy/v1beta1/cfg.proto
```diff
@@ -124,6 +124,34 @@ message Rule {
...
+    string name = 1;
+
+    // Optional. Header value.
+    string value = 2;
+
+    // Header operation type.
+    enum Operation {
+      REPLACE = 0;  // replaces the header with the given name
+      REMOVE = 1;   // removes the header with the given name (the value is ignored)
+      APPEND = 2;   // appends the value to the header value, or sets it if not present
```
geeknoid: In this context, I don't quite understand the difference in semantics between REPLACE and APPEND. Can you elaborate? Also, the command says "appeands the value to the header value", and I'm not sure what "appends" means here. String concatenation?

policy/v1beta1/cfg.proto
```diff
@@ -124,6 +124,34 @@ message Rule {
...
+    }
+
+    // Optional. Header operation type.
```
geeknoid: Optional? What's the behavior when not present?

policy/v1beta1/cfg.proto
```diff
@@ -146,6 +174,9 @@ message Action {
   // Referenced instances are evaluated by resolving the attributes/literals for all the fields.
   // The constructed objects are then passed to the `handler` referenced within this action.
   repeated string instances = 3;
+
+  // Optional. A handle to refer to the results of the action application.
```
geeknoid: Remove "application"

policy/v1beta1/cfg.proto
kyessenov: HTTP header pairs can be repeated. For example, the following is correct in the protocol:
```
x-forwarded-for: 10.0.0.1
x-forwarded-for: 10.0.0.2
...
```

It is usually up to the client to decide what to do with multiple values for a header. Typically, the values are space-concatenated, e.g. the client would receive `x-forwarded-for: 10.0.0.1 10.0.0.2`. Append operation corresponds to inserting a new pair into the request headers, without regard whether a value is present or not for the same header.

policy/v1beta1/cfg.proto
kyessenov: Default value is 0, which is `REPLACE`.

policy/v1beta1/cfg.proto
kyessenov: Done

policy/v1beta1/cfg.proto
geeknoid: OK, so then what is the semantic of REPLACE and REMOVE given that there can be multiple headers with the same name? Do they remove all headers with the given name? What if you want to remove only the second instance of a particular named header?

policy/v1beta1/cfg.proto
geeknoid: Then let's add a comment about the default behavior when not specified

policy/v1beta1/cfg.proto
kyessenov: Operations are executed one by one. Semantically, there is just one header with the given name, but it can be represented on the wire as several "key: value" pairs. That means removing is really removing all pairs on the wire, yet appending is fast, since you can just add one more pair. 

Removing a second value in a list is simply not supported (not in envoy, nor any other proxy I know of). Usually you want to append a header value, or to replace it entirely. Replacing is more intuitive, so I made it the default. 

There are a few examples here: https://github.com/istio/istio/blob/master/mixer/test/client/route_directive/route_directive_test.go#L95

geeknoid: Sorry to keep harping on this...

What you say here makes sense. So in this context, each header is thus a name with an array of values. In that context, I think the value field should be a []string instead of just a single string. Thus:

REMOVE. Remove a header with all its value
REPLACE. Replace a header and all its values with the new array of values
APPEND: Appends an array of values to a header's values.

kyessenov: My intent was to optimize for the common case, which is just a single header key-value pair. If you insist we can change the value to be an array type. I am just worried this might confused the user, since we would need to explain the semantics of HTTP headers again.

In terms of implementation, the two are interchangeable:
```
REMOVE_ARRAY(key, _) = REMOVE(key, _)
REPLACE_ARRAY(key, {"value1", "value2"}) = REPLACE(key, "value1, value2")
APPEND_ARRAY(key, {"value1", "value2"}) = APPEND(key, value1); APPEND(key, value2)
```

policy/v1beta1/cfg.proto
kyessenov: Done

policy/v1beta1/cfg.proto
geeknoid: I agree the patterns are isomorphic. I just find it confusing that the model is one header to multiple values, but the API just deals in terms of a single value like that's all there is.

Ultimately, it boils down to the syntax operators will use to manipulate this stuff. Do you have an idea how this would look in config?

kyessenov: Here is an example from the doc:
```yaml
 actions:
    - name: authorizer
      handler: authorizer
      instances: [“cookie”]
requestHeaderOperations:
    - name: user-info
      value: $denier.user
```

geeknoid:
OK, so if you want to add two values, you need to repeat a whole bunch of
stuff. If the value field is an array, it's a trivial thing instead. So to
me, it seems like a win to make the value field an array. Am I missing
something?

kyessenov:
Fair enough. Here is an updated example:
```
actions:
    - name: authorizer
      handler: authorizer
      instances: [“cookie”]
requestHeaderOperations:
    - name: user-info
      values: 
      - $authorizer.user
```

geeknoid:
Thanks.

/lgtm
/approve

PR #615, opened 2018-08-17
fix proxy build broken by global_dictionary.yaml

yangminzhu:
I made a comment in https://github.com/istio/api/pull/605/files but think it's faster to send out a PR directly.

@mandarjog 
Line 316 `- "-"` break the proxy build as [create_global_dictionary.py](https://github.com/istio/proxy/blob/4865583db000db91be0469a7f1756cc98fea9992/src/istio/mixerclient/create_global_dictionary.py#L65) parse it to: `    ""-"",` which is invalid in `.cc` file.

Could you suggest the proper way to fix this? Should we just remove the `"` or make it `\"-\"` or even update the script?

### DIFF
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -313,7 +313,7 @@
 - DI
 - FI
 - RL
-- "-"
+- -
 
 - inbound
 - outbound


mandarjog:
@yangminzhu can you verify that the resulting file is still a valid yaml file?
There is a go generate command in `istio/istio` where mixer server generates go code from this.

@bianpengyuan You had faced an issue with `-`, and I thought we got rid of it.

bianpengyuan:
@mandarjog instead of removing it, I added some logic to escape `"` in list generation code in https://github.com/istio/istio/pull/8087.

yangminzhu:
@bianpengyuan  so I think the right way to fix is to also update [create_global_dictionary.py](https://github.com/istio/proxy/blob/4865583db000db91be0469a7f1756cc98fea9992/src/istio/mixerclient/create_global_dictionary.py#L65) to do the same escape? If so I'll update this PR as well. Thanks.

bianpengyuan:
@yangminzhu Yes I think so. Let's do that.

PR #617, opened 2018-08-21
Add a configuration option for consecutive gateway errors

arnecls:
Introduce a new option `ConsecutiveGatewayFailure` to the outlier detection.
This change is required to enable outlier detection based on 502,503 and 504 errors as defined by envoy here: https://www.envoyproxy.io/docs/envoy/latest/api-v1/cluster_manager/cluster_outlier_detection.html

Required for https://github.com/istio/istio/issues/8082

An example config should look like this
```yaml
kind: DestinationRule
metadata:
  name: reviews-cb-policy
spec:
  host: reviews.prod.svc.cluster.local
  trafficPolicy:
    connectionPool:
      tcp:
        maxConnections: 100
      http:
        http2MaxRequests: 1000
        maxRequestsPerConnection: 10
    outlierDetection:
      consecutiveGatewayErrors: 5
      interval: 5m
      baseEjectionTime: 15m
```


arnecls:
I signed it!

networking/v1alpha3/destination_rule.proto
```diff
@@ -436,6 +436,11 @@ message OutlierDetection {
   // Maximum % of hosts in the load balancing pool for the upstream
   // service that can be ejected. Defaults to 10%.
   int32 max_ejection_percent = 4;
+
+  // Number of gatway errors before a host is ejected from the connection
+  // pool. Disabled by default. When the upstream host is accessed over 
+  // HTTP, a 502, 503 or 504 return code qualifies as an error.
+  int32 consecutive_gateway_failure = 5;
```
rshriram: I am not sure we need both. I would be happy to treat the existing consecutive_error as gateway errors instead of any 5xx error code, as thats what people mostly expect. When someone wants to treat misused 500 error for outlier, we can add a special option.

arnecls: I think it depends wether you have a proxy in between or not and wether you use an external service (that you might not be able to change) or your own.
If you target service errors you will want to react on all 5xx codes.
If you only want to target proxy errors 502-504 is probably fine.
So IMO it makes sense to have a choice here.

arnecls: Also I would consider modifying the setting this a breaking change as existing installations might change their behaviour.

arnecls: @rshriram Any additional thoughts on this?
Otherwise I will continue as requested.

rshriram:
We try not to expose the plethora of sometimes confusing Envoy options to end users, in the routing api. 

Within a mesh, gateway errors will be more common (502/503/504) while most sensible external services will return a 503 to shed load. 

Secondly we just made the outlier detection generic to both tcp and http. The consecutive gateway error applies only to http and will make no sense in tcp context. 

I also feel that 500 error code is not something indicative of overload. The whole idea behind outliers is to remove overloaded servers from the lb pool. 

We don’t have very many users relying on this behavior I think. We kept it intentionally generic so that we can switch to a more specific error code in future (which happens to be now).

@ceposta thoughts on interpreting consecutive_errors as 502/503/504 only instead of any 5XX ?

arnecls:
Sounds reasonable to me.

PR #618, opened 2018-08-21
mixer: add output attributes to check adapters

kyessenov:
Allow gRPC check adapters to return attributes.
/assign @geeknoid 
/assign @mandarjog 


mixer/adapter/model/v1beta1/check.proto
```diff
@@ -35,4 +37,6 @@ message CheckResult {
   google.protobuf.Duration valid_duration = 2 [(gogoproto.nullable) = false, (gogoproto.stdduration) = true];
   // The number of uses for which this result can be considered valid.
   int32 valid_use_count = 3;
+  // Output attributes
+  istio.mixer.v1.Attributes output_attributes = 4;
```
geeknoid: I don't think this is right. Adapters don't deal with attributes. Look at how APAs work for example.

kyessenov: APAs have not been converted to use gRPC adapters. Mappers in APA access the output attribute bag directly. How do we smuggle the state out of adapters? If we send an instance, then  that instance would need to be used as an attribute value?

kyessenov: cc @douglas-reid for thoughts on gRPC APAs.

kyessenov: Right, we don't need reverse operation here (converting values to attributes). We can define attribute manifest from the output template just like in APAs, and use those produced values in expressions (e.g. treating `$action_name.field_name` as attributes in the expressions). The problem then is by using custom outputs for each check adapter, we need a different type signature for every adapter, unless we are OK using any or struct.

douglas-reid: I've lost the thread a bit on some of this design work, but have we considered possibly adding a new template type (`CHECK_WITH_RETURN_VAL` or similar with better name)?  It could still be used to generate new method signatures, a la:

```go
func (h *handler) HandleListEntryWithOutput(context.Context, *listentry.Instance) (adapter.CheckResult, *ktmpl.Output, error) {
```

And then we could auto-gen the `attribute_bindings` bit of the spec like we do for APAs?

Just thinking out loud.

douglas-reid: Our comments passed each other in the ether, it appears. We would need a different type signature for every *template* right (not per every *adapter*)?

kyessenov: It's per-adapter I think. Something like this:
```proto
// HandleListEntryService is implemented by backends that wants to handle request-time 'listentry' instances.
service HandleMapEntryService {
    // HandleMapEntry is called by Mixer at request-time to deliver 'mapentry' instances to the backend.
    rpc HandleMapEntry(HandleMapEntryRequest) returns (HandleMapEntryResponse);
}

message HandleMapEntryResponse {
  istio.mixer.adapter.model.v1beta1.CheckResult result = 1;
  ValueMsg value = 2;
}
message ValueMsg {
  string map_value = 1;
}

```

douglas-reid: `ValueMsg` should be encoded in the *template*, I would think. Each `MapEntry` adapter would be bound to produce only `ValueMsg` values in addition to `CheckResult`, based on the contract of the template.

However, a different template, say `IAM Policy`, might have a `AlternateValueMessage` or no additional message, and the adapter would be bound by that contract for `HandleIAMPolicyService` impls, right?

kyessenov: We're probably talking about same things, and I misunderstand the meaning of template. When I say per-adapter, I mean that there is custom code written for each particular variation of ValueMsg. IAM Policy and Map Entry would be two different implementations, each producing two distinct protos.

douglas-reid: Ah, yes. So, there is probably some refactoring that could enable reuse across templates. But, a naive way would mean different message per template. That's why it would be good to have *something* that these could be mapped back into (like the `attribute_bindings` for APAs) for standard processing.

kyessenov: Followed up with PR #619 .

geeknoid:
Even though this was closed, let me add a few things.

What we tried to do with the adapter model is to deal with strongly-typed
objects. Adapters received strongly-typed data derived from the templates
they implement. Adapters have been specifically isolated from the details
of attributes, attributes are purely a concern for Mixer proper and its
configuration model.

Similarly, for data coming out of adapters, we should be on a fully
structured plan. Adapters should return a well-known proto, and the
operator is then responsible for consuming this data as it arrives.

PR #619, opened 2018-08-21
mixer: add new check template variety

kyessenov:
This is an alternative to https://github.com/istio/api/pull/618.

We introduce a new template variety for check adapters that also produce values.

These are hybrid adapters between APAs and Check in that they produce both `CheckResult` and an instance of `OutputTemplate`. The generated instance is then bound to `$action_name` from the rule, similar to how `$out` is bound in the APA mapper.

This PR has two benefits over #618:
- avoids the discussion how to put polymorphic proto into check result
- less disruptive to existing gRPC adapters


kyessenov:
@mandarjog @douglas-reid 

I have trouble wiring the output proto through the dynamic handlers. The concrete go struct for the output message is not loaded into mixer for dynamically loaded adapters. That means we cannot easily unmarshal the custom output for the new template handlers. Is this the reason why APAs are not yet converted to gRPC?

I am thinking to use proto.Struct trick to recover the output message. As long as we can read attributes from a struct, I think that might just work.

geeknoid:
This looks like a winner

/lgtm
/approve

PR #620, opened 2018-08-23
add incremental mcp support

ayj:
Extend the MCP to include support for incremental configuration pushes. This is *heavily* derived from the new incremental xDS protocol. It has advantages over the non-incremental variant:

- **efficiency** - MCP servers don't need to deliver the full state of the world for minor config updates. This is particularly important for high churn config (e.g. service entries) and session resumption.
- **per-resource versions** - The orig resource version (e.g. CRD) can be encoded in the MCP resource version. 

xref: https://github.com/envoyproxy/data-plane-api/blob/master/XDS_PROTOCOL.md#incremental-xds
xref: https://github.com/istio/istio/issues/8148


mcp/v1alpha1/mcp.proto
```diff
@@ -98,6 +98,68 @@ message MeshConfigResponse {
   string nonce = 4;
 }
 
+// IncrementalMeshConfigRequest are be sent in 2 situations:
+//
+//   1. Initial message in a MCP bidirectional gRPC stream.
+//   2. As a ACK or NACK response to a previous IncrementalMeshConfigResponse.
+//      In this case the response_nonce is set to the nonce value in the Response.
+//      ACK or NACK is determined by the absence or presence of error_detail.
+message IncrementalMeshConfigRequest {
```
rshriram: This works in Envoy because envoy tries to resolve missing references to clusters and routes (the initial_resource_versions field below). Even in Envoy, its unclear how you would use this to send new listeners.

In Pilot, its harder. There is nothing referring to virtual services/service entries/destination rules/Envoy filters by names. The only resource that is referred to by name is the gateway.

We would have to adapt the model/protocol such that Galley can send intelligent diff messages like add/remove/update virtual service (not all services), add/remove/update destination rules, etc. The update message should come in, only if the contents of the CRD has changed. Not when resource time stamps change or someone adds spaces/comments/etc.

ayj: I updated the wording to make it clear the `initial_resource_versions` is only included in the initial request. Subsequent requests must omit this field.

PR #623, opened 2018-08-28
add EnforcementMode to RbacConfig

quanjielin:
https://github.com/istio/istio/issues/5345


rbac/v1alpha1/rbac.proto
```diff
@@ -247,4 +247,8 @@ message RbacConfig {
   // A list of services or namespaces that should not be enforced by Istio RBAC policies. Note: This field have
   // effect only when mode is ON_WITH_EXCLUSION and will be ignored for any other modes.
   Target exclusion = 3;
+
+  // $hide_from_docs
+  // Indicates enforcement mode of the RbacConfig.
+  EnforcementMode enforcement_mode = 4;
```
yangminzhu: What's the behavior when it is set to `Permissive`?

quanjielin: rbac is transparent to users; when mapping to rbac filter config, there is only shadow_rules(no normal rules).

yangminzhu: This `enforcement_mode` is supposed to be used by user, right? What's the expected behavior from user's perspective when setting this to `Permissive`?

diemtvu: +1 Please add some comment to explain the behavior.

quanjielin: added more comments

rbac/v1alpha1/rbac.proto
```diff
@@ -247,4 +247,11 @@ message RbacConfig {
   // A list of services or namespaces that should not be enforced by Istio RBAC policies. Note: This field have
   // effect only when mode is ON_WITH_EXCLUSION and will be ignored for any other modes.
   Target exclusion = 3;
+
+  // $hide_from_docs
+  // Indicates enforcement mode of the RbacConfig, in ENFORCED mode by default.
+  // It's used to verify new RbacConfig work as expected before rolling to production. 
+  // When setting as PERMISSIVE, RBAC isn't enforced and has no impact on users.
+  // RBAC engine run RbacConfig in PERMISSIVE mode and logs stats.
```
diemtvu: Can this be PERMISSIVE and the one under ServiceRoleBinding is STRICT, and what does it mean in that case?

quanjielin: shouldn't allow user input like that, added comment to indicate it's invalid case

PR #628, opened 2018-08-31
[WIP] Add MeshService and MeshEndpoint for Pilot Decomposition

andraxylia:
Introduce 2 new APIs: MeshSevice and MeshEndpoint to be use for [Pilot Decomposition ](https://goo.gl/qAsEQx), as a way to ingest Service and Endpoint  definitions from the various platforms and from other Pilots.

The ServiceMesh APIs are very similar to the existing ServiceEntry, but not re-using it to keep a separation from what the user configures explicitly vs what may be available in the service registry. We also needed a separate API for endpoints, more aligned with what envoy has, for two reasons:
- to allow Endpoints to be added/deleted incrementally, without sending the full list of Endpoints. Incremental MCP (based on incremental xDS) is already supported.
- to support load assignment etc

MeshService is similar to ServiceEntry, but without the list of Endpoints - some re-use may be possible.

MeshEndpoint is based on Envoy endpoint.proto with some extra fields: service_name and labels.


andraxylia:
cc @louiscryan

costinm:
We have 4 representations for endpoints in istio - I am 100% sure a 5th is not needed. 

Please document any features that are missing in Envoy's endpoint representation of ServiceEndpoint - since both can be improved. 

Strong -1 on yet another interface, we have >50 APIs to deal with.

andraxylia:
Can you paste links to the 4 APIs you refer to so we can see if any is suitable or can be used?

andraxylia:
If you refer to ServiceEntry, I agree this is a good start, and we will use it to unblock the implementation, but at some point we will need this new APIs. The features that are missing in ServiceEntry are documented in this PRs description: 

"The ServiceMesh APIs are very similar to the existing ServiceEntry, but not re-using it to keep a separation from what the user configures explicitly vs what may be available in the service registry. We also needed a separate API for endpoints, more aligned with what envoy has, for two reasons:
- to allow Endpoints to be added/deleted incrementally, without sending the full list of Endpoints. Incremental MCP (based on incremental xDS) is already supported.
- to support load assignment etc"

rshriram:
Endpoints without a service association is of very limited use. So, we might as well relax the service entry semantics (or may be part of the implementation) such that if a service entry has service definition but no endpoints, it can be merged with another service entry for same service, that has endpoints but no service definition.

andraxylia:
This MeshEndpoint API I am proposing wraps an Envoy endpoint and a service association.

This is to support in the most optimal way a new Endpoint added to a service. The same functionality cannot be achieved with ServiceEntry, which requires sending the full list of Endpoints. In fact, with the current API and non-incremental MCP, an endpoint change requires sending all ServiceEntries with all endpoints.

There are multiple optimization we can do:
1) Incremental MCP: this will allow to send a single ServiceEntry with full list of Endpoints. Best return on investment for a scale of up to 100 Endpoints.

2) Incremental API for an Endpoint - to optimize even further

andraxylia:
@rshriram I just realized you mean this is proto3 and Endpoint is not required, so it is valid to use a ServiceEntry with zero endpoints. This means I can remove the MeshService.

Still no solution for the Endpoint itself. Since it is optional also, can we deprecate it, move it's definition out of the ServiceEntry, and create a proper Endpoint? ( we may need to call it differently than Endpoint, hence MeshEndpoint). We also need to add support for multiple networks, ServiceEntry definition restricts endpoints to a single network.

andraxylia:
@rshriram @ayj Related to having a cluster wide ServiceEntry as proposed in istio/istio#8462, this has to work for services in a namespace, and support 2 services with the same name in 2 different namespaces. This means we remove the namespace and we require the FQDN to be specified in the name.

Can we call this new cluster wide service entry CRD "MeshService"? This will allow us to make faster progress in the implementation without impacting the existing ServiceEntry.

rshriram:
> Related to having a cluster wide ServiceEntry as proposed in istio/istio#8462, this has to work for services in a namespace, and support 2 services with the same name in 2 different namespaces. This means we remove the namespace and we require the FQDN to be specified in the name.

WHY? Service entries are cluster wide. There is no point in directing google.com to one IP in one namespace and to another IP in another namespace. The associated implementation complexity in Pilot is a lot. 99% of the use cases simply want a service entry that allows external traffic to other services.

costinm:
Joining services across namespaces has many security risks and is hard to
understand. The opposite ( a service in same namespace in 2 clusters being
kept separated) may be reasonable, and we can handle using distinct domain.
Cluster.local can continue to mean local or 'equivalent'.

andraxylia:
@rshriram An update on this one: based on the latest discussions we had in the networking WG, ServiceEntry is not a cluster-wide resource, it's a namespaces resource.

And based on the incremental EDS work done in Pilot for performance optimization, we do need an Endpoint API that works with incremental MCP. Since ServiceEntries are mergeable, it's Ok to send an incremental Endpoint inside a ServiceEntry. But we are still unable to separate the user config from the synthetic ServiceEntries - those converted from the service registry. The options I see are:
- introduce a new resource (this PR, MeshService, a bit redundant)
- introduce an enum field in the ServiceEntry "synthetic/user-config", or an origin field :"k8s, cf, user-configured". I kind of like this one, it will also give some indication as to who the producer is.

@ayj @louiscryan wdyt ?

costinm:
This is discussed in https://github.com/istio/istio/issues/10220 and we talked about in Env WG.

There is no need to add an enum field or origin - user should not be exposed or have to worry about how the implementation works. 

This PR introduces a new protocol - similar with ServiceEntry but hard to keep in sync. 

The first option - to have MCP indicate a ServiceEntry is a synthetic one, either via Envelope or 
by using a different resource type identifier:
- is no exposed to user - no confusion
- doesn't introduce another API to express almost the same thing with ServiceEntry
- works for any other future 'synthetic' resource
- is extremely simple to implement 
- does not require protocol changes - ADS and MCP can already handle this.  

All we need is MCP server to send user-edited configs when ServiceEntry resource type is requested,
and send 'synthetic' ones when "synthetic/ServiceEntry" resource type is used. Consul and multi-cluster 
MCP servers will support only the synthetic/ServiceEntry - that's all we need for EDS MC.

andraxylia:
@costinm you are describing in fact a 4th which I have not thought about, but would work very well. This is to use a different [type_url ](https://github.com/istio/api/blob/master/mcp/v1alpha1/mcp.proto#L58) for the synthetic resources in general, which would contain "synthetic" in the path.

@ayj @ozevren do you foresee any problem?

ozevren:
Changing the type URL is really abusing what it is meant for. But that is mainly a problem with the current state of the protocol (which we inherited from xDS). These are, in reality, collection names. Regardless of this discussion, I think we should pick a new nomenclature that indicates that these are collections, rather than intimately tying them to Type URLs with no obvious benefit. 

We can have separate collections for resources (i.e. different TypeURLs), but that is awkward. Just from a protocol/cleanliness standpoint, having a single collection of things that are semantically the same is cleaner.

We can have a "synthetic" flag at the MCP level, but I am worried that it will end up being a vestige that is not very useful for the general case. It is probably much more useful to have a synthetic field on the ServiceEntry itself, where it would have a much better understood meaning. We don't have to expose this to the user through the CRD, this can be filtered out at the CRD level.

andraxylia:
@ozevren This is interesting, how can we filter it at the CRD level? I thought the CRD spec is automatically generated based on the .proto definition.

ozevren:
Spec is manual written. We have a PR going on in tools that somewhat automates it, but it is not baked: https://github.com/istio/tools/pull/30

In either case, we can filter out the field value in Galley and reject any setting of such a field in validation. We can also hide it from docs.

costinm:
I think this is once again picking complicated solutions when a simple
solution is already possible.

If it is ok to make changes to the protocol - we can simply change the
documentation and/or name of TypeUrl,
 there is no particular reason to use the low-level protocol type in
ADS/MCP requests.

Even K8S API allows parameters and filters to be specified in the request -
and they use a structured URL.
My proposal was to just follow the K8S pattern - and I would like to hear
arguments why this pattern
is bad instead of 'hacky' or 'abusing' or non-technical

I hope we'll not get into another bike shedding/5 proposal situation -

ozevren:
I am not sure I understand what you're objecting to. It looks like we're in
agreement that these are really collections, and it should be immaterial if
the collection name is a bona-fide Type URL or some other name. The main
point is about simply changing the field/concept name in the protocol so
that the intent is clear, instead of trying to work around the quirks of
our *own* protocol which we haven't really shipped yet.

costinm:
To be clear: my proposal is to change the semantic of TypeUrl  to Type +
Url, i.e. to include both the type
and optional filters.  It is consistent with the K8S protocol use - where
the 'ApiVersion' is combined with
query parameters and resource names in the over-the-wire protocol. It is
consistent with how all REST
APIs are implemented today, and with how pubsub systems work.

It is fully backward compatible with ADS and current MCP, and can be
implemented and understood by anyone familiar
with how HTTP works. I'm not aware of any protocol in use today that is
using the type as primary identifier and doesn't allow any
filtering or customization, and I don't think it is something we must
preserve at all costs.

ResourceNames has an important role in on-demand and incremental - and I
certainly hope we'll not look
for reasons to arbitrarily diverge from ADS. AFAIK the plan remains to try
to keep the protocol in sync and
propose any changes back to Envoy community.

costinm:
I am objecting to bike-shedding and backward incompatible protocol changes
or adding extra complexity
to our user-visible APIs.

We have short time - and if we start another 'network scope' bike-shedding
session we'll ship 1.1 with the
current complex diff and inefficient implementation. I can't believe using
TypeUrl as a URL is typically
used -- to indicate both the type and parameters, like K8S protocol is
doing with their ApiVersion would
be controversial.

ozevren:
What part of this is backward incompatible? It is changing the name of a proto field and having better meaning around it. Otherwise, what you're arguing for is effectively what I am saying.

andraxylia:
There are 2 requirements:
- do not expose a field the user does not need to configure
- allow an MCP client to request only **user-configured** ServiceEntries, as opposed to synthetic or all of them. 

Starting with the current convention about [type_url](https://github.com/istio/api/blob/master/mcp/v1alpha1/mcp.proto#L57) , @costinm 's proposal is to change it from
"type.googleapis.com/istio.io.networking.v1alpha3.VirtualService"

to something like:

"type.googleapis.com/**synthetic**/istio.io.networking.v1alpha3.VirtualService"

If we are in agreement, let's ship it, otherwise please spell out the concrete place where this will go.

costinm:
It's ok to rename the field in MCP.

But please keep the same proto number. I think keeping the MCP and ADS
protocol aligned is very important, improvements in one should
be reflected in the other.

IMO type_url is literally correct - it identifies a type and a url. And
perfectly consistent with apiVersion in k8s - that includes both api and
version. It is not called 'type'.

Ultimately what we need is to allow pilot to customize what it is watching
- we will have other cases where we don't want everything of
type ServiceEntry or VirtualService, and using a URL is the most  common
mechanism for this ( on the web and in k8s )

andraxylia:
I also think type_url is a good name, and do not see why collections is better or warrants the change.

Do we have volunteers for making this change? It needs to go in Galley @qufel , Pilot, and co-pilot @nino-k .

I will close this PR.

ozevren:
Type URL has a very specific meaning and a particular format. It doesn't simply mean "a type & a URL". https://developers.google.com/protocol-buffers/docs/proto3. There is also some Google specific history behind it, AFAIK.

andraxylia:
@ozevren are you referring to the type_url and value generated inside the Any?
https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Any

I agree, the one we have is an explicit different field, which happens to be named the same. @ayj was this on purpose so we can write some generic code?

andraxylia:
We had a quick meeting and agreed we will start by having Pilot requesting ServiceEntries, and Galley returning an annotation indicating the type of ServiceEntry (exact values to be debating during that PR). The annotation will be added in the MCP response.

In the longer term, we will add the ability to specify in the MCP request a filter on specific fields of a resource. This will be used to optimize notifications to include only Endpoints for instance, or only relevant fields of a Service.

I will close this PR for now since we decided not to go with a separate API for Endpoint.

Issue #610, opened 2018-08-15
Documentation of TargetSelector needs to be 'workload oriented'


louiscryan:
The documentation here is misleading

https://github.com/istio/api/blob/33feaf42f8982948e9b065f744d22a7645c8e62b/authentication/v1alpha1/policy.proto#L345

We need to make it clearer that the target selector is primarily designed to select workloads and not services. 

A 'service' is the name used to address a workload, there can be many 'service' names for a single workload. It would be very unusual for policies to change simply because there were two different names a client could use to address a service. It is far more common and reliable to attach policies to the workload so that they are enforced for all the services that address that workload.

@diemtvu @liminw @wenchenglu

ayj:
What does the `targetSelector` map onto for non-k8s workloads?

louiscryan:
There still needs to be some workload declaration thats associated with the non-k8s workload when we Istio-ify it

wora:
A workload and a service are often referring to the same entity. Workload
is internal facing concept, while service is external facing concept. Many
people will use both concepts at the same time.

In this case, we need to make a clear distinction between a workload and a
service, and design the policy accordingly.

diemtvu:
I agree with Hong. At the moment, the policy is applied per-service, with the assumption service-to-workload is one-to-many mapping. We didn't change it to 'workload' because: (1) we don't have a way to identify workload and (2), it also has similar issue where a service involve multiple workloads (imo, this is even more common than the other case).

PR #629, opened 2018-09-06
Change TargetSelector documentation to be 'workload oriented'

diemtvu:
Issue: https://github.com/istio/api/issues/610


authentication/v1alpha1/policy.proto
```diff
@@ -357,14 +357,28 @@ message Policy {
   PrincipalBinding principal_binding = 6;
 }
 
-// TargetSelector defines a matching rule to a service/destination.
+// TargetSelector defines a matching rule to an workload. A workload is selected if it associated
```
liminw: Same nits here.

PR #631, opened 2018-09-10
Remove special case conversion of "Struct" => interface{} for Mixer config protos.

ozevren:
We had some legacy conversion code that converted Struct fields in Mixer's config protos to interface{}. While the reasons for doing this is lost in Istio history, experimentation with removing this showed that we can make things work: https://github.com/istio/istio/pull/8561.

This PR simply removes the custom sed/grep processing of the proto files and allows them to be exposed in vanilla form. A sibling checkin on the Istio repo will update the code to use the new form of code.


mandarjog:
woohoo.
/lgtm

PR #633, opened 2018-09-13
RBAC proto changes.

liminw:
* Promoted some commonly used properties, like ports, IPs to be first class
entities.
* Added "exclusion" and "or" semantics.
* Added "conditions" that allows any arbitrary AND/OR/NOT combinations.

Signed-off-by: Limin Wang <liminwang@google.com>


rbac/v1alpha1/rbac.proto
```diff
@@ -163,10 +281,11 @@ message ServiceRoleBinding {
 
   // $hide_from_docs
   // Indicates enforcement mode of the ServiceRoleBinding.
-  EnforcementMode mode = 3;
+  EnforcementMode enforcement_mode = 3;
```
costinm: Backward incompatible change ?

(and please add a test in the code with all the possible properties set, so we can detect next time a backward incompatible change happens )

rbac/v1alpha1/rbac.proto
```diff
@@ -133,6 +133,124 @@ message AccessRule {
...
+    map<string, string> orProperties = 8;
+  }
+
+  // $hide_from_docs
+  // Optional. Specifies the conditions that the permission does not apply.
+  Exclusions exclusions = 10;
+
+  // $hide_from_docs
+  // Optional. Additional conditions that apply to the permission.
+  repeated Condition conditions = 11;
```
yangminzhu: Is this replacing the `properties`?

rbac/v1alpha1/rbac.proto
liminw: This is a hidden field for an unreleased feature. @quanjielin

rbac/v1alpha1/rbac.proto
```diff
@@ -133,6 +133,124 @@ message AccessRule {
...
+    repeated string values = 2;
+
+    // List of values that should be excluded. Exact match, prefix match,
+    // and suffix match are supported for the values.
+    repeated string notValues = 3;
+  }
+
+  oneof RuleType {
+    // A single rule entry.
+    Rule rule = 1;
```
yangminzhu: how is it different from a single rule in `andRules` or `orRules`?

rbac/v1alpha1/rbac.proto
yangminzhu: Also why this not included in the exclusion?

liminw:
@costinm This is completely control plane RBAC proto changes. We have a separate data plane RBAC proto in Envoy repo. There will not be Envoy code change. So it does not impact runtime performance.

rbac/v1alpha1/rbac.proto
```diff
@@ -133,6 +133,124 @@ message AccessRule {
...
+  // $hide_from_docs
+  // Optional. Definition of custom properties, which are key-value pairs.
+  // A key refers to an Istio attribute. The supported keys are listed in the
+  // "constraint and properties" page. Exact match, prefix match, and suffix
+  // match are supported for property values. The properties are `OR`ed together.
+  map<string, string> orProperties = 9;
+
+  // $hide_from_docs
+  // Specifies the conditions that the permission does not apply.
+  message Exclusions {
```
yangminzhu: I think we shouldn't repeat all these fields two times, it's not only making the definition duplicate, but also double the code to parse it as they are different types.

I think we should wrap them in a standalone message like (the name could be changed if necessary):
```
message AccessRuleContent {

repeated string services = 1;
... ...
map<string, string> orProperties = 9;
}
```
And reuse it in `AccessRule`.

rbac/v1alpha1/rbac.proto
```diff
@@ -176,9 +295,84 @@ message Subject {
...
+  // $hide_from_docs
+  // Optional. Definition of custom properties, which are key-value pairs.
+  // A key refers to an Istio attribute. The supported keys are listed in the
+  // "constraint and properties" page. Exact match, prefix match, and suffix
+  // match are supported for property values. The properties are `OR`ed together.
+  map<string, string> orProperties = 8;
+
+  // $hide_from_docs
+  // Specifies the conditions that the subject should be excluded.
+  message Exclusions {
```
yangminzhu: Same to the comment on `AccessRule`, we should reuse these fields.

rbac/v1alpha1/rbac.proto
```diff
@@ -176,9 +295,84 @@ message Subject {
...
+    map<string, string> orProperties = 6;
+  }
+
+  // $hide_from_docs
+  // Optional. Specifies the conditions that the subject should be excluded.
+  Exclusions exclusions = 9;
+
+  // $hide_from_docs
+  // Optional. Additional conditions that identifies the subject.
+  repeated Condition conditions = 10;
```
yangminzhu: Same question as above, is this replacing the `properties`?

rbac/v1alpha1/rbac.proto
yangminzhu: This will break the existing code, what's the benefit to rename?

rbac/v1alpha1/rbac.proto
```diff
@@ -133,6 +133,124 @@ message AccessRule {
...
+
+  // $hide_from_docs
+  // Optional. Specifies CIDR IP addresses for the destination workload.
+  repeated string ips = 6;
+
+  // $hide_from_docs
+  // Optional. Selects the destination workloads with the specified labels
+  // that the rule applies to. Exact match, prefix match, and suffix match
+  // are supported for label values.
+  repeated string labels = 7;
```
yangminzhu: Same concern as the ports.

rbac/v1alpha1/rbac.proto
```diff
@@ -133,6 +133,124 @@ message AccessRule {
   // Optional. Extra constraints in the ServiceRole specification.
   // The above ServiceRole example shows an example of constraint "version".
   repeated Constraint constraints = 4;
+
+  // $hide_from_docs
+  // Optional. Specifies the ports of the destination service.
+  repeated string ports = 5;
```
yangminzhu: Is this the ports for workload (i.e. pod in k8s) but not for the service?
Current the rbac pilot plugin is based on service which is converted from the k8s Service CRD, not sure if there will have any conflicts here.

@diemtvu Do you have any insights about this?

rbac/v1alpha1/rbac.proto
```diff
@@ -133,6 +133,124 @@ message AccessRule {
...
+
+  // $hide_from_docs
+  // Optional. Specifies CIDR IP addresses for the destination workload.
+  repeated string ips = 6;
```
yangminzhu: Same concern as the ports.

rbac/v1alpha1/rbac.proto
```diff
@@ -133,6 +133,124 @@ message AccessRule {
...
+
+  // $hide_from_docs
+  // Specifies the conditions that the permission does not apply.
+  message Exclusions {
+    // Optional. A list of services that should be excluded. Exact match, prefix match,
+    // and suffix match are supported for service names. This is similar to `services`
+    // field, but specifies exclusion.
+    repeated string services = 1;
+
+    // Optional. A list of HTTP paths or gRPC methods that should be excluded.
```
rshriram: Categorize this please:
services
http:
 paths,
 methods,..
tcp:
 ip, ports
mysql:
 database
  user
  table,
  operation
mongo:
  database
  user, etc..
Lets not pile all of this into one top level object.

rbac/v1alpha1/rbac.proto
liminw: I want to provide two levels of condition support, (1) The simple conditions can be specified with properties/orProperties. (2) "conditions" here allows any arbitrary combinations of AND/OR/NOT logic, which is more flexible, but can make the policy look complicated.

rbac/v1alpha1/rbac.proto
liminw: It's a recursive semantics. andRules/orRules are composed with the single rules here.

rbac/v1alpha1/rbac.proto
liminw: ditto.

rbac/v1alpha1/rbac.proto
yangminzhu: I know `andRules` and `orRules` are recursive, I mean it seems the `rule` is not necessary in the oneof as you can always put a single rule in `andRules` or `orRules` to achieve the same effect?

rbac/v1alpha1/rbac.proto
liminw: We discussed it in @quanjielin's PR. The goal is to make the terminology consistent for per-policy and global permissive mode settings.

rbac/v1alpha1/rbac.proto
diemtvu: If this is similar to authN, it should be service port (inbound listener is always on service port, I think).

Btw, why this is string? For consitency, you may want to use the same type we have in authN for port selector.

rbac/v1alpha1/rbac.proto
yangminzhu: If so I think it needs better documentation. Also why this not appear in the `exclusion` part as seems fields exist in both places.

Another question, what's the relationship for different condition in this list?

liminw:
Replaced with [Authorization policy V2 design](https://docs.google.com/document/d/1diwa9oYVwmLtarXb-kfWPq-Ul3O2Ic3b8Y21Z3R8DzQ).

PR #638, opened 2018-09-25
Add option for LightStep tracing to ProxyConfig

JulianGriggs:
This PR adds configuration for a LightStep tracing backend to the ProxyConfig proto.

A WIP/POC implementation can be found [here](https://github.com/lightstep/istio/pull/3). When this PR gets finalized I will update and submit the implementation for review.


JulianGriggs:
I signed it!

mesh/v1alpha1/config.proto
```diff
@@ -133,6 +148,10 @@ message ProxyConfig {
 
   // The mode used to redirect inbound traffic to Envoy.
   InboundInterceptionMode interception_mode = 18;
+
+  // Configuration for a LightStep tracing backend (optional). If set, it will
+  // take precedence over [zipkinAddress][istio.mesh.v1alpha1.ProxyConfig.zipkin_address].
+  LightstepCollector lightstep_collector = 19;
```
rshriram: Mark the zipkin address as deprecated. And create a tracing message that is a one-of between Zipkin/lightstep. something like
```
message Tracing {
 message Zipkin {
   address string
 }
 message LightStep {
  ...
 }
  one_of {
    Zipkin zipkin;
    Lightstep lightstep;
 }
}
Tracing tracing
```

The proxyconfig is internal data structure. We formalized it here for some documentation purposes. So, there is no backward compatibility issues here. It merely translates to the set of CLI args that we pass to the pilot agent [which you will have to update anyway to enable lightstep].

JulianGriggs:
@rshriram Thanks for the review! I created a consolidated Tracing message as you described. Question: do you think it should be a top-level message or should it be nested within ProxyConfig? I currently have it as a top level message but am happy to change it if you think it would be better off nested.

JulianGriggs:
pinging @frankbu @hklai @rshriram for review :)

PR #639, opened 2018-09-26
remove out of date fields from config

GregHanson:


GregHanson:
removed reference to fields in istio/istio here:
https://github.com/istio/istio/pull/8956

PR #643, opened 2018-10-01
audit: add AuditDirective to check response

yangminzhu:
istio/istio#8535


audit/v1alpha1/audit.proto
```diff
@@ -0,0 +1,37 @@
...
+// Expresses the audit actions to be performed after a check call.
+message AuditDirective {
+  enum Level {
+    // No log is generated.
+    NONE = 0;
+
+    // Log request metadata (timestamp, caller identity, request path, method, etc) only.
+    METADATA = 1;
+
+    // Audit content is provided by the backend application.
```
liminw: Log request metadata and annotated content from request or response.

audit/v1alpha1/audit.proto
```diff
@@ -0,0 +1,37 @@
...
+syntax = "proto3";
+
+// This package defines audit related protos.
+package istio.audit.v1alpha1;
+
+option go_package = "istio.io/api/audit/v1alpha1";
+
+// $hide_from_docs
+// Expresses the audit actions to be performed after a check call.
+message AuditDirective {
```
liminw: Why do we need to create AuditDirective message? Can we directly use AuditLevel in CheckResponse?

yangminzhu: Just give us more flexibility to extend the audit function later, it's easier to add another field to this message instead of putting it directly under the CheckResponse.

liminw: Can we move the enum AuditLevel to outside AuditDirective? We will need to use it in AuditPolicy.

yangminzhu: Good point, done.

mixer/v1/check.proto
```diff
@@ -82,6 +83,9 @@ message CheckResponse {
     // An optional routing directive, used to manipulate the traffic metadata
     // whenever all preconditions are satisfied.
     RouteDirective route_directive = 6;
+
+    // An optional audit directive, used to decide audit actions.
+    audit.v1alpha1.AuditDirective audit_directive = 7;
```
geeknoid: The audit directive stuff should not be in a separate package, it is just part of Mixer's API and should just be with the rest of the Mixer stuff.

geeknoid:
We're putting in place a new process to review API changes. @ozevren, are we ready to engage the process?

geeknoid:
This PR seems premature. We don't have a design describing where the audit data can come from within Mixer and how operators control it. Before we start changing the surface API, we need a sketch of the end-to-end story.

mixer/v1/check.proto
yangminzhu: Moved the `AuditDirective` to the check.proto.

yangminzhu:
@geeknoid This PR doesn't include any detail audit policy, I'm planning to add them later. Could you take a look of this PR to see if there are more changes needed? Thanks

yangminzhu:
@geeknoid Can we move forward on this if no more changes needed at this stage? Thank you.

PR #646, opened 2018-10-03
Updates to service entries

rshriram:
Some updates to service entries

Network name to service entry endpoints: This allows differentiating
services that are spread across multiple network enclosures (e.g. multiple  
kubernetes services, multiple different RFC1918 networks, etc.).

Spiffee ID to service entries:
So that we can validate services defined through service entries.
Note: I have explicitly refrained from adding service accounts as    
it allows folks to embed spiffee compatible VSID in the identity
and still get mTLS, without being unduly bound to the semantics 
of service accounts that may not be applicable to external services
or legacy services.    

Second: I have also intentionally refrained from adding the spiffee
ID per endpoint because neither Pilot logic nor Envoy logic allows for
validating the SAN of individual endpoints. Its at the Envoy cluster level
which in Istio terms, is either a service or a subset of a service.
Curiously, the auth logic in Pilot is written to assume that if one endpoint
in a service has a service account, all endpoints must have the same service
account.


networking/v1alpha3/service_entry.proto
```diff
@@ -390,8 +390,21 @@ message ServiceEntry {
...
   };
 
   // One or more endpoints associated with the service.
   repeated Endpoint endpoints = 6;
+
+  // One or more spiffe IDs associated with the endpoints of this
+  // service. The spiffee IDs will be used to verify the identity of the
+  // endpoints when the destination rule for the service uses ISTIO_MUTUAL
+  // as the TLS mode.
+  repeated string spiffee_identities = 7;
```
baodongli: Do you expect that this is a complete spiffe string such as:
_spiffe://cluster.local/ns/default/sa/default_?

networking/v1alpha3/service_entry.proto
```diff
@@ -390,8 +390,21 @@ message ServiceEntry {
 
     // One or more labels associated with the endpoint.
     map<string, string> labels = 3;
+
+    // Network names allow to group endpoints resident in the same
+    // failure domain/administrative boundry/rfc1918 network (e.g., a
+    // kubernetes cluster, availability zone, etc.). The network name
+    // can be used for a variety of purposes, one of which is zone-aware
+    // routing.
+    string network_name = 4;
```
frankbu: Can these be any made up simple name or some special format? What is the scope of uniqueness?

networking/v1alpha3/service_entry.proto
yangminzhu: `spiffee` -> `spiffe` in the comment and field name?

networking/v1alpha3/service_entry.proto
baodongli: in that case, network_name doesn't sound like an appropriate name here.

networking/v1alpha3/service_entry.proto
louiscryan: We already have this feature in DestinationRule by virtue of being able to declare SANs there. 

The only reason to know the set of allowed identities that the endpoints can run as is so that the client can validate them. We also simply cant restrict ourselves to SPIFFE ids being the only allowable identity type, common names could also be allowed.

Don't think we need this

networking/v1alpha3/service_entry.proto
louiscryan: I don't think this is a fleshed out model and we need a design doc on the structure here.

We can't merge the concepts of RFC 1918 namespace and region into a single field in a sensible manner. Many folks have single IP spaces that span global networks.

when we have a merged service discovery mechanism this information will need to reside on the endpoints too

rshriram: This is already in the endpoint. At the end of the day, its some opaque ID that describes the network reachability boundary. I could split this up into zone, region, country, if that suffices.. The networking topology should either be part of this hierarchy or it shouldn't figure in this at all. Because from pilot's point of view, it simply sees an endpoint thats in some zone/region/country and it will try to setup envoy to route to local zone/local region/country before going somewhere else. How do you envision a flat or not flat network adding another orthogonal dimension to this?

The only sane way I can think of this is to establish a hierarchy somewhere else and specify what part of the hierarchy this endpoint falls in (which is the network_name here. We could call this group_name if it helps.)

networking/v1alpha3/destination_rule.proto
```diff
@@ -539,10 +539,8 @@ message TLSSettings {
   // A list of alternate names to verify the subject identity in the
   // certificate. If specified, the proxy will verify that the server
   // certificate's subject alt name matches one of the specified values.
-  // Should be empty if mode is `ISTIO_MUTUAL`.
```
costinm: +1 - but can you add an extra line documenting what should happen if this field is not empty ? 
"will  be combined with the spiffe names guessed from workloads"  
or 
"If not empty, only the values in this list will be allowed"

I prefer the second option - it's more deterministic and efficient (if user sets it we don't have to scan).

networking/v1alpha3/istio.networking.v1alpha3.pb.html
```diff
@@ -3042,6 +3042,17 @@ <h2 id="ServiceEntry">ServiceEntry</h2>
 <td>
 <p>One or more endpoints associated with the service.</p>
 
+</td>
+</tr>
+<tr id="ServiceEntry-spiffee_identities">
+<td><code>spiffeeIdentities</code></td>
+<td><code>string[]</code></td>
```
costinm: I'm not sure hardcoding 'spiffee' in the networking API is the right option. We do use spiffee, but also support non-spiffee - and in future other kind of identities may be supported. 

'identities' seems a better name - and the doc should not tie this to ISTIO_MUTUAL. A spiffee identity can also be used by an external service and in any other TLS mode.

costinm: Also - when was this field added ? It's not in 1.0 - so we can still remove/change.

networking/v1alpha3/service_entry.proto
```diff
@@ -390,8 +390,21 @@ message ServiceEntry {
 
     // One or more labels associated with the endpoint.
     map<string, string> labels = 3;
+
+    // Network names allow to group endpoints resident in the same
+    // failure domain/administrative boundry/rfc1918 network (e.g., a
```
costinm: Please remove the 'failure domain / administrative boundry/rfc1918' . It's confusing - you can have a single admin boundry and rfc1918 and failure domain, but still group endpoints in different networks.

networking/v1alpha3/service_entry.proto
costinm: Not the cluster name. Large cluster spanning multiple zones exist.

The 'flat/not flat' is orthogonal - network creates a grouping of endpoints, if the network is flat you can still chose to use gateways to communicate, and in a mixed environment some networks may use direct and some may use gateways.

networking/v1alpha3/service_entry.proto
costinm: See my comment in the doc. Please don't restrict ourselves to only SPIFFE. 
Next year a new spec will be the hot thing. We already have the 'san' entry - and 
san can be a URL or domain or IP, according to the spec. Spifffe is just one type of 
URL SAN - with spiffee scheme. 

We already support this - would be very confusing for users to have some SAN singled out
here, and other in the subject_alternative_name.

networking/v1alpha3/service_entry.proto
rshriram: Yes, but this network name here is purely for zone aware routing. If someone wants to communicate via gateways in a flat network, then they should use all the routing rules. the implementation for that feature should not be baked into Pilot, as the same thing can be accomplished through the API.

networking/v1alpha3/service_entry.proto
rshriram: Failure domain is the zone.. And this is simply providing an example. Merely stating a network name doesn't ring a bell in anyone's mind. We need to map this to something they already know.  Otherwise, this ends up becoming zones, regions, countries, etc. etc.. and there is no end to that hierarchy.

networking/v1alpha3/service_entry.proto
```diff
@@ -391,20 +402,29 @@ message ServiceEntry {
...
+    // used to establish connectivity (usually using the
+    // ISTIO_REGISTRY_PASSTHROUGH mode in a Gateway Server).
     string network_name = 4;
+
+    // The failure domain associated with the endpoint.  All endpoints
+    // within the same failure domain and the same network should be
+    // directly routable. Endpoints in same failure domain across different
+    // networks need to be routable via the Gateway attached to each
+    // network.
+    FailureDomain failure_domain = 5;
```
costinm: I'm not sure I like the name "FailureDomain". The primary use I think is 'zone/region aware' load balancing. 

Yes, typically the availability zone is also a failure domain - but the concept is very subtle and more relevant when 
you deploy workloads. From networking perspective we don't do anything special about failure domains - scheduler may replicate workloads so they are in multiple regions and AZ, but I can't think of anything that networking will do differently about failure domains.

networking/v1alpha3/service_entry.proto
```diff
@@ -391,20 +402,29 @@ message ServiceEntry {
...
-  // service. The spiffee IDs will be used to verify the identity of the
-  // endpoints when the destination rule for the service uses ISTIO_MUTUAL
-  // as the TLS mode.
-  repeated string spiffee_identities = 7;
+  // One or more identities associated with the endpoints of this
+  // service. These identities will be used to validate the endpoint
+  // identity when clients connect to this service using Istio mTLS.  To
+  // enable Istio mTLS, use destination rule and specify TLS mode as
+  // ISTIO_MUTUAL.
+  repeated string subject_alternate_names = 7;
```
costinm: Except in destination_rule it's called subject_alt_names.

And I'm not sure we want to have it in both places - not sure how they'll interact either. 
I think destination rule is a better place for this.

rshriram: It’s 2x the configuration. And there is conflict with global destination rule that sets mode to istio mutual.

rshriram: This is actually in our service model today. The workload identity or service account or whatever.

networking/v1alpha3/gateway.proto
```diff
@@ -260,6 +261,18 @@ message Server {
       // Secure connections to the upstream using mutual TLS by presenting
       // client certificates for authentication.
       MUTUAL = 2;
+
+      // Similar to the passthrough mode, except servers with this TLS mode
```
costinm: So in this PASSTHROUGH we treat the SNI as a string, matching a hostname ( port/subset not included).
This mode SNI is the istio-mangled cluster name string - and is mapped directly to a cluster.

It's probably ok - I still think it would be cleaner if the cluster names were proper DNS HostNames ( as I interpret the SNI spec to say - it allows opaque but has quite a bit to indicate it's expected to be hostname).
If we had it as hostname - we could use PASSTHROUGH.

networking/v1alpha3/service_entry.proto
```diff
@@ -390,8 +390,42 @@ message ServiceEntry {
...
+    // endpoints in the locality are available, endpoints parent locality
+    // (but within the same network ID) will be chosen. For example, if
+    // there are two endpoints in same network (networkID "n1"), say e1
+    // with locality us/us-east-1/az-1/r11 and e2 with locality
+    // us/us-east-1/az-2/r12, a sidecar from us/us-east-1/az-1/r11 locality
+    // will prefer e1 from the same locality over e2 from a different
+    // locality. Endpoint e2 could be the IP associated with a gateway
+    // (that bridges networks n1 and n2), or the IP associated with a
+    // standard service endpoint.
+    string locality_id = 5;
```
costinm: I don't think this is an 'id' - it's the actual locality.

networking/v1alpha3/service_entry.proto
```diff
@@ -390,8 +390,42 @@ message ServiceEntry {
...
 
   // One or more endpoints associated with the service.
   repeated Endpoint endpoints = 6;
+
+  // One or more identities associated with the endpoints of this
+  // service. These identities will be used to validate the endpoint
+  // identity when clients connect to this service using Istio mTLS.  To
+  // enable Istio mTLS, use destination rule and specify TLS mode as
+  // ISTIO_MUTUAL.
+  repeated string subject_alt_names = 7;
```
costinm: Some comments on how this interacts with the subject_alt_name in DestionationRule. 

Still not clear why we need them in both places.

networking/v1alpha3/gateway.proto
rshriram: Yes, we are going to move to dns name for the multicluster gateway clusters alone,. to start with. We still need a special mode though to activate the new filter sni_cluster

networking/v1alpha3/service_entry.proto
rshriram: This is mostly for automation.. When Galley is streaming service entries for rmeote clusters, its easier to package all info in service entry (network, locality, sans, etc.). Otherwise, it has to create a destination rule for each service entry from remote cluster. Not only that, it also has to merge multiple destination rules (user specified and galley generated). Too much complexity and makes automation hard

ZackButcher:
(Commit already removed it, but to record our discussion:) I think we should remove the SAN from DestinationRule. It definitely doesn't make sense to have it in both ServiceEntry and DestinationRule, and I don't think there are very compelling use cases for SAN-per-subset (maybe identity per cluster, but we're using ServiceEntries to model that instead). Otherwise LGTM.

networking/v1alpha3/destination_rule.proto
```diff
@@ -536,13 +536,9 @@ message TLSSettings {
   // Should be empty if mode is `ISTIO_MUTUAL`.
   string ca_certificates = 4;
 
-  // A list of alternate names to verify the subject identity in the
-  // certificate. If specified, the proxy will verify that the server
-  // certificate's subject alt name matches one of the specified values.
-  // Should be empty if mode is `ISTIO_MUTUAL`.
-  repeated string subject_alt_names = 5;
+  // Deprecated. Use the Service entry to specify subject alt names.
+  repeated string subject_alt_names = 5 [deprecated=true];
```
costinm: Well - that's one way to do it. However, not everything has service entry. I think DestinationRule is still a good place to put the SAN - it also includes the TLS info ( expected CA certificate ) for example. 

And deprecating stuff is not nice. 

I agree there is some overlap between DestinationRule and ServiceEntry. 

@louiscryan  - what do you think ?

rshriram: Everything that doesn't have a service entry is already part of the k8s registry, i.e. already has a SAN. Otherwise, it will not figure in the service registry in Pilot (i.e. destination rule's hosts need to be in service registry to start with). 

I added to service entry because its hard to automate destination rule for large scale multi cluster scenarios where we import service entries in bulk. Just like the way pods come with service accounts by default and we can use destination rules to override, service entries come with service accounts by default.

But @ZackButcher pointed out that it then makes no sense to specify the SAN in destination rule as people don't use it in k8s context. They use it only when doing multicluster/or referring to service entry hosts.

I have no preference to leaving it in destination rule, as a way to override SAN obtained through service entry. But we really need to have SANs in service entry (san is just a term, but the implied meaning is the service account.). Pilot's internal registry has the concept of service account for each service. This is merely bringing the service entry spec in line with Pilot's internal registry model.

costinm: Why ? Consul doesn't have SAN, nor ServiceEntry. 

For multi-cluster - whatever we do is likely to end up automated and not exposed to normal users.
That means it can create whatever internal representation we need, including DestinationRule.

For users who do want to have explicit control or write their own automation - I think DestinationRule is more intuitive.

Having SAN in a different place than the cert and SNI is very strange. 

Also: the SAN obtained from 'workload' is a very debatable concept. SAN is used as a mean of authorizing, 
i.e. who do you expect to talk with.  In k8s we trust the environment (i.e. if you can deploy a workload - you control the service accounts ), but in general it is dangerous to say 'if any workload of service S happens to run as user A,  then A becomes an authorized SAN for service S'. Even in k8s where we do that - it is restricted to SA and services within a namespace (because it's the unit of isolation and trusted). It may be secure and easier for users in this specific environment - but can't be generalized.

rshriram: The problem is this: If I import 500 service entries and 500 destination rules through tooling, and the user also ends up defining her own destination rule for one of these services, then we need a way to merge. We also need a way to support global mTLS policies defined using *.svc.cluster.local, tlsMode: ISTIO_MUTUAL [can't specify every workload's san here].

Hence the hack..

PR #648, opened 2018-10-05
Add sampling parameters to Rule

douglas-reid:
This PR is a proposal for adding sampling support to  `policy.v1beta1.Rule`s. Two basic strategies are included (and proposed to be supported).

Reviewers:
- Do we believe that this is the right place to add this config? Or would inside of an action or even at the Handler level feel more natural?
- Do the options for sampling feel complete?  Are there alternatives we should support instead?


policy/v1beta1/cfg.proto
```diff
@@ -309,3 +318,34 @@ message Connection {
...
+
+// RateLimitSampling provides the ability to limit the number of Rule action executions that
+// occur over a period of time.
+message RateLimitSampling {
+  // Window in which to enforce the sampling rate.
+  google.protobuf.Duration sampling_duration = 1 [(gogoproto.nullable)=false,(gogoproto.stdduration) = true];
+  // Number of entries to allow during the `sampling_duration` before sampling is enforced.
+  int64 max_unsampled_entries = 2;
+  // The rate at which to sample entries once the unsampled limit has been reached. Sampling will be enforced
+  // as 1 per every `sampling_rate` entries allowed.
```
kyessenov: I feel like these are two distinct strategies. Maybe not use `oneof` but instead allow both strategies to be specified? One strategy constraints max bursts per specified duration, and another one samples. Both are optional, and can be used in isolation.

policy/v1beta1/cfg.proto
```diff
@@ -309,3 +318,34 @@ message Connection {
...
+message SamplingConfig {
+  oneof strategy {
+    RandomSampling random_sampling = 1;
+    RateLimitSampling rate_limit_sampling = 2;
+  }
+}
+
+// RandomSampling will filter based on the comparison of a randomly-generated value
+// against the threshold provided. When available, the `request.headers[x-request-id]`
+// attribute will be used as the source of the value. Otherwise, a PRNG will be used.
```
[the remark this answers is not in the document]
kyessenov: Theoretically, sampling needs to be pushed all the way down to the proxy. Therefore, producing random values at the proxy has merits. As long as correlation is acceptable...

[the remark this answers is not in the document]
douglas-reid: I guess this raises the issue of whether or not Rules should be truly independent. Is it better to have all `Report` rules fire for the same request or only have partial (say `logs` but not `tracespan` or `metrics` or vice versa).

As for referencing the attribute, I was thinking this information could be conveyed through some sort of side-channel. The idea would be, I think, to match the trace sampling behavior in the proxy (which, iirc, is based on `x-request-id`). But generating a random value in the proxy sounds fine too.

Pushing sampling down to the proxy sounds good, but that does prevent sampling logs but always generating metrics, etc. Perhaps in future iterations of Mixer API impls we will get that for free?

policy/v1beta1/cfg.proto
```diff
@@ -309,3 +318,80 @@ message Connection {
...
+// `sampling_rate` would be set `12.5`.
+//
+// This sampling configuration is meant to closely match the access log RuntimeFilter configuration
+// supported by Envoy:
+// https://github.com/envoyproxy/data-plane-api/blob/master/envoy/config/filter/accesslog/v2/accesslog.proto#L113
+message RandomSampling {
+  // Optional. Specify an attribute to use to extract numerator for `percent_sampled` field.
+  // If this value is set, but no value is found OR if that value is not a numeric value, then
+  // the derived sampling rate will be 0 (meaning no `Action`s are executed for a `Rule`).
+  string attribute = 1;
```
geeknoid: Should this be an expression?

douglas-reid: Makes sense. Updated.

douglas-reid:
@geeknoid Do I need to hide these params until we provide an implementation?

geeknoid:
This will be published with the 1.1 doc set. If you think we'll have an implementation for 1.1, then you can leave it there. Otherwise, you can hide the fields & types from the docs using $hide

douglas-reid:
Ok. Added `$hide_from_docs` and regenerated. Will un-hide them when impl is complete. This should allow flexibility with respect to 1.1.

policy/v1beta1/cfg.proto
```diff
@@ -309,3 +319,84 @@ message Connection {
...
+  // the derived sampling rate will be 0 (meaning no `Action`s are executed for a `Rule`).
+  string attribute_expression = 1;
+  // The default sampling rate, expressed as a percentage. Defaults to 0% with a denominator
+  // of 100.
+  FractionalPercent percent_sampled = 2;
+  // By default sampling will be based on the value of the request header `x-request-id`.
+  // This behavior will cause consistent sampling across `Rule`s and for the full trace of a 
+  // request through a mesh (across hosts). If that value is not present and/or 
+  // `use_independent_randomness` is set to true, the sampling will be done based on the value of
+  // attribue specified in `attribute`. If that attribute does not exist, the system will behave as
```
kyessenov: nit: `attribute_expression`

policy/v1beta1/cfg.proto
kyessenov: IMHO, I'd expect the items that exceed the throttle to be entirely dropped, not randomly sampled.

douglas-reid: This was inspired by the sampling config in the underlying Istio logging library (zap). You can set the `sampling_rate` to 0 to match your expectation. This allows you, additionally, to sample overflow. If that flexibility has operator value or not, I'm not entirely clear.

kyessenov:
I don't know if I have the authority to approve. cc @geeknoid @ozevren 
I have no problem merging this since it's hidden from public docs and not gonna be published anyways until 1.1.

kyessenov:
/lgtm

(and it's a revert away if it's premature)

PR #649, opened 2018-10-07
Add selectors for workload oriented authn policy

lei-tang:
Add selectors for workload oriented authn policy


authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,12 @@ message PortSelector {
     string name = 2;
   }
 }
+
+// WorkloadSelector specifies the labels and ports to be used for
+// selecting targets for authentication policy.
```
wora: This comment mentions both workload and target. Can you explain the difference? Please give a concrete example that people are supposed to use this feature for?

lei-tang: Revised this comment in the proto file (the "targets" in this comment should be "workloads").  The examples and details for workload oriented policy can be found at [this link](https://docs.google.com/document/d/1eInPqAjOfsyJWtMIOeRRza0b0aErKqVzO9mbxNKcucA/edit#heading=h.7zgnj8bwqfld).

wora:
The proto file should be the spec, not the design doc. A user must be able
to read the proto spec and know how to use it, at least for people working
in this area.

authentication/v1alpha1/policy.proto
lei-tang: An concrete example has been added.

lei-tang:
An example has been added.

authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,18 @@ message PortSelector {
...
+// WorkloadSelector specifies the labels and ports to be used for
+// selecting workloads for authentication policy.
+// The following is an example of using WorkloadSelector, which selects
+// the workload that has the label {"env": "testing} and is on the port 1234.
+// WorkloadSelector {
+//   labels: {"env", "testing"}
+//   ports: [1234]
+// }
+message WorkloadSelector {
+  // Labels used for selecting workloads
```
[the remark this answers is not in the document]
lei-tang: Done.

authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,18 @@ message PortSelector {
...
+  map<string, string> labels = 1;
```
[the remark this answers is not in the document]
lei-tang: Done.

authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,18 @@ message PortSelector {
...
+  // Ports used for selecting workloads
```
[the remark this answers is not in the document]
lei-tang: Done.

authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,20 @@ message PortSelector {
...
   }
 }
+
+// WorkloadSelector specifies the labels and ports to be used for
+// selecting workloads for authentication policy.
+// The following is an example of using WorkloadSelector, which selects
+// the workload that has the label {"env": "testing} and is on the port 1234.
+// WorkloadSelector {
+//   labels: {"env", "testing"}
+//   ports: [1234]
```
wora: What syntax you are using here? yaml or prototext or json?

lei-tang: Changed to use the yaml format.

authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,29 @@ message PortSelector {
...
+//     ports:
+//     - 1234
+// ```
+message WorkloadSelector {
+  // One or more label matching conditions to select subset of pods/VM on which this authentication
+  // policy should be applied. When there are multiple label requirements, all must be satisfied.
+  map<string, string> match_labels = 1;
+  // Limit the policy's scope to workloads only on these ports. Leave empty to apply to all ports
+  // (provided by the selected pods/VM)
+  repeated uint32 ports = 2;
```
costinm: That's not possible - VMs can't be forced to use specific ports, and we don't want to track/index on this.

Please use the service port, as everything else in networking.

authentication/v1alpha1/policy.proto
```diff
@@ -313,6 +313,12 @@ message Policy {
   // If empty, policy will be used on all workloads in the same namespace.
   repeated TargetSelector targets = 1;
 
+  // List rules to select workloads that the policy should be applied on.
+  // Different from TargetSelector, WorkloadSelector is based on the labels
+  // and ports.
+  // If empty, policy will be used on all workloads in the same namespace.
```
costinm: Very confusing. So if 'targets' is not empty and this is -  the policy still applied to all workloads ? 
And I still don't understand from the docs what is different from the target selector. 

I would just add the labels to the 'TargetSelector' - which also selects a workload. It's very confusing to have both TargetSelector and WorkloadSelector select workloads - but be different just because one select by 
workload belonging to a service ( which happens to be defined as a set of labels ), the other select by labels directly. In the end both select workloads, both use labels - and the difference is extremely subtle.

costinm:
Please add this to the ToC review list.

authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,29 @@ message PortSelector {
...
+//   selectors:
+//   - match_labels:
+//       env: 'testing'
+//     ports:
+//     - 1234
+// ```
+message WorkloadSelector {
+  // One or more label matching conditions to select subset of pods/VM on which this authentication
+  // policy should be applied. When there are multiple label requirements, all must be satisfied.
+  map<string, string> match_labels = 1;
```
yangminzhu: The comment should be clear that this only selects pods in the same namespace as the policy.

authentication/v1alpha1/policy.proto
```diff
@@ -389,3 +395,29 @@ message PortSelector {
     string name = 2;
   }
 }
+
+// WorkloadSelector specifies the labels and ports to be used for
+// selecting workloads for authentication policy.
+// The following is an example of using WorkloadSelector, which selects
+// the workload that has the label {"env": "testing"} and is on the port 1234.
```
yangminzhu: Same to below, the comment should be clear that this only selects pods in namespace `ns`.

authentication/v1alpha1/policy.proto
yangminzhu: `targets` will be deprecated by this `selectors`, It's better to make this clear in the comment. Also if both `targets` and `selectors` are set, it should be clear which one will be used.

lei-tang: Moved the labels to TargetSelector.

authentication/v1alpha1/policy.proto
lei-tang: Moved the labels to TargetSelector. WorkloadSelector is removed.

authentication/v1alpha1/policy.proto
lei-tang: WorkloadSelector is removed. Moved the match_labels to TargetSelector.

authentication/v1alpha1/policy.proto
lei-tang: WorkloadSelector is removed. Moved the match_labels to TargetSelector.

lei-tang:
WorkloadSelector is removed. Moved the match_labels to TargetSelector.

authentication/v1alpha1/policy.proto
yangminzhu: Then the comment of `match_labels` in `TargetSelector` should be clear it only selects workloads in the same namespace as the policy.

lei-tang: The comment is added.

lei-tang:
@costinm Where is the ToC review list? How to add this PR to the ToC review list?

authentication/v1alpha1/policy.proto
diemtvu: Indeed, we want to get rid of TargetSelector, i.e the way using service name. Adding option using labels under TargetSelector may be ok, except for the confusion of having port specification, as it is **service** ports. I saw the other comments that VM's port is not available, so we may need to step back how to address that.

authentication/v1alpha1/policy.proto
diemtvu: We can ignore port in that case, can we? Not ideal, but I feel mixing label (i.e workload/pod centric) with service port for the selector is kinda confusing in different way.

authentication/v1alpha1/policy.proto
```diff
@@ -376,6 +378,16 @@ message TargetSelector {
...
+
+  // One or more label matching conditions to select subset of pods/VM on which this authentication
+  // policy should be applied. When there are multiple label requirements, all must be satisfied.
+  // It only matches workloads in the same namespace as the policy.
+  // For example, the following match_labels matches the workloads that have the label {"env": "testing"}.
+  // ```yaml
+  //   match_labels:
+  //     env: 'testing'
+  // ```
+  map<string, string> match_labels = 3;
```
rshriram: Can you please use the same terminology as the Gateway for workload selector?

lei-tang: Done.

authentication/v1alpha1/policy.proto
```diff
@@ -355,6 +355,8 @@ message Policy {
 
 // TargetSelector defines a matching rule to a workload. A workload is selected
 // if it is associated with the service name and service port(s) specified in the selector rule.
+// If match_labels are specified, the policy will only be applied to the workloads
+// for which all label requirements are satisfied.
 message TargetSelector {
   // REQUIRED. The name must be a short name from the service registry. The
   // fully qualified domain name will be resolved in a platform specific manner.
```
costinm: I assume REQUIRED will be replaced with DEPRECTED ? This was the service name ? And some doc that to migrate to the new labels, user should copy the service labels into the target selector.

authentication/v1alpha1/policy.proto
```diff
@@ -376,6 +378,16 @@ message TargetSelector {
   // ```
   //Leave empty to match all ports that are exposed.
```
costinm: I don't think 'targetPort' makes sense - since a service may select multiple target ports. 
Also I would want to reduce the 'port name' - Istio is now using service port number in cluster and internally. 

Since different policy would apply to different ports - it is probably important to have the service port.

authentication/v1alpha1/policy.proto
```diff
@@ -376,6 +378,16 @@ message TargetSelector {
...
   repeated PortSelector ports = 2;
+
+  // One or more label matching conditions to select subset of pods/VM on which this authentication
+  // policy should be applied. When there are multiple label requirements, all must be satisfied.
+  // It only matches workloads in the same namespace as the policy.
+  // For example, the following selector matches the workloads that have the label {"env": "testing"}.
+  // ```yaml
+  //   selector:
+  //     env: 'testing'
```
costinm: How about using a more realistic example - I doubt anyone will use env:testing to select workloads.

app: "productpage"
version: "1"

authentication/v1alpha1/policy.proto
lei-tang: From the explanation in the proto file on lines 365~366, targetPort is not used in PortSelector and the service port is used in PortSelector. As to "Also I would want to reduce the 'port name'", do you mean removing the port name from PortSelector?

authentication/v1alpha1/policy.proto
lei-tang: Done.

authentication/v1alpha1/policy.proto
lei-tang: For the deprecation planning, I suggest: 
Phase 1: if no matching labels are defined for an Istio authn policy, it will have the same behavior as current Istio authn policy. Thus no breaking for existing users. If matching labels are defined, service name and matching labels are used to select workloads and the selected workloads must be in the same namespace as that of the authn policy.
Phase 2: deprecate service name and only use matching labels

authentication/v1alpha1/policy.proto
costinm: Yes, internally we're using port numbers. Having both may be confusing to users.

geeknoid:
@lei-tang Do we still need this PR?

lei-tang:
@geeknoid The PR is closed.

PR #650, opened 2018-10-08
TLS mode for multicluster passthrough in Gateway

rshriram:


networking/v1alpha3/gateway.proto
```diff
@@ -260,6 +260,12 @@ message Server {
       // Secure connections to the upstream using mutual TLS by presenting
       // client certificates for authentication.
       MUTUAL = 2;
+
+      // Forward the connection to the upstream envoy cluster specified by
```
costinm: I would call it 'ZVPN' or 'ZERO_VPN' - since that's what we used in the past. 

Using Gateway and this mode is IMO valid for single cluster as well - for example a single cluster with mesh expansion ( where the isolated machines are using SNI). Or simply a gateway running on a VM, with no cluster anywhere. 

Given that envoy is also using the term 'cluster' in a different context, and we have multicluster with flat network - I would be happy with any name that doesn't have cluster. Including 'federated'.

costinm: Also - what is the difference between this an 'PASSTHROUGH' - which is documented as 'upstream server selected on the SNI string' ?

rshriram: Pass through is standard SNI matching and picking an internal cluster. This is encoding the cluster name in the SNI itself. 
 I don’t want to call it zvpn since it doesn’t seem fitting. Probably mesh_dnat or l6_dnat or simply dnat

PR #653, opened 2018-10-09
Cleanup and Implement Basic API Guidelines

ozevren:
The Guidelines is going to be a living document. This PR puts some basics in place.


GUIDELINES.md
```diff
@@ -0,0 +1,192 @@
...
+message HttpRequest {/*...*/}  // Not HTTPRequest!
+
+rpc DoHttpRequest(/*...*/)     // Not DoHTTPRequest!
+
+enum HttpStatusCodes {/*...*/} // Not HTTPStatusCodes!
+```
+
+
+#### Messages
+* **Do** use ```CamelCase``` for message names.
```
geeknoid: Are some of these redundant with the google guidelines referenced above?

GUIDELINES.md
```diff
@@ -0,0 +1,192 @@
...
+    ```
+* **Do not** use postpositive adjectives in names.
+  * ```proto
+    repeated Items collected_items = 3; // Not items_collected!
+    ```
+ 
+#### Enums
+* **Do** use ```CamelCase``` for  types names.
+  * ```proto
+    enum Types {/*...*/}
```
geeknoid: Should we ask for enum names to be plural?

GUIDELINES.md
```diff
@@ -0,0 +1,192 @@
...
+* **Do** have an enum entry for value ```0```.
+  * When a new field with an enum type is introduced to a proto, when reading 
+  the older versions of the data, it will be defaulted to ```0``` value.
+* **Do** name the ```0``` value either as ```<ENUMTYPE>_UNDEFINED``` or use a sane,
+well-known value that will be considered as default.
+  * ```proto
+    enum Types { TYPE_UNDEFINED = 0;}
+    ```
+
+### Basic Proto Versioning
```
geeknoid: I think this needs an intro paragraph that describes when these rules apply, since they are temporal in nature.

GUIDELINES.md
```diff
@@ -0,0 +1,192 @@
...
+* **Do not** change field numbers.
+  * Proto depends on field numbers to find fields.
+  * ```proto
+        // Field number has changed from 1 to 2!
+        
+        // string field = 1; // Deleted
+        string field = 2;         
+    ```
+* **Do not** rename fields.
+  * Our tooling automatically maps Yaml fields to proto fields.
```
geeknoid: Should that be YAML?

GUIDELINES.md
```diff
@@ -0,0 +1,192 @@
...
+  * If the field types changes, the new type **must** be structurally equivalent to the old.
+  ```proto
+      // Field type changed from Boo to Zoo!
+      // Boo and Zoo must be structurally equivalent!
+      message Foo {
+         // Boo boo = 1;  
+         Zoo boo = 2;
+      }
+    ```
+* **Do not** rename enums.
```
geeknoid: How about enum values?

GUIDELINES.md
```diff
@@ -0,0 +1,192 @@
...
+* **Do** use the top-level proto Message name as the ```kind``` of the CRD.
+
+```yaml
+# A custom resource that describes a Gateway resource
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway 
+# ...
+```
+
+Matches to:
```
geeknoid: Perhaps show the proto first?

GUIDELINES.md
ozevren: Somewhat. It is hard to fish through those documents though. Besides, we already had some of these in the existing guideline doc.

ozevren:
? /lgtm?

PR #655, opened 2018-10-16
policy: add missing APA attribute bindings

kyessenov:
OOP adapters use a fixed instance schema. In-process attribute-producing adapters relied on code generation to extend the instance schema with an `attribute_bindings` field. Due to the fixed schema, to support OOP attribute-producing adapters, we need to extend the fixed schema for instance config.


policy/v1beta1/cfg.proto
```diff
@@ -231,6 +231,10 @@ message Instance {
   // Required. Depends on referenced template. Struct representation of a
   // proto defined by the template; this varies depending on the value of field `template`.
   google.protobuf.Struct params = 3;
+
+  // Optional. Defines attribute bindings to map the output of attribute-producing adapters back into
+  // the attribute space as part of pre-processing.
```
mandarjog: 1. I don't think we want to restrict it to `pre-processing` 
2. Please add doc showing a couple of attribute bindings.

kyessenov:
PTAL

PR #656, opened 2018-10-18
Multicluster/partial-registry view network configuration API

rshriram:
Trying to consolidate various discussions. The updates here provide a way to specify the multiple service registries from which Pilot should source its info, the config stores and also group endpoints into networks depending on the topology. This should formalize the split-horizon stuff we have been doing.


networking/v1alpha3/mesh.proto
```diff
@@ -0,0 +1,66 @@
...
+
+option go_package = "istio.io/api/networking/v1alpha3";
+
+// `Mesh` describes various global properties of a single Istio mesh. A
+// single mesh can be composed of one or more registries. The endpoints
+// managed by the registries can be in one or more L3 networks creating
+// islands of visibility across the mesh.
+message Mesh {
+  // REQUIRED: A list of registry specifications.
+  repeated Registry registries = 1;
```
costinm: I would keep each Registry and Network as a top level object. They can be ACLed and edited independently, and consistent with what we do for Gateways and Services.

networking/v1alpha3/mesh.proto
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Provides information about the service registry - type, tls settings, etc.
+message Registry {
+  string name = 1;
+  enum RegistryType {
+    KUBERNETES = 0;
+    CLOUD_FOUNDRY = 1;
+    CONSUL = 2;
+    GALLEY = 3;
```
costinm: MCP - galley is one implementation of MCP, but not the only one.

networking/v1alpha3/mesh.proto
```diff
@@ -0,0 +1,66 @@
...
+  }
+  RegistryType registry_type = 2;
+  // ip or host
+  string address = 3;
+  // port where the remote server is providing registry data
+  uint32 port = 4;
```
costinm: Maybe we can use 'url' - and not have port + address.

networking/v1alpha3/mesh.proto
```diff
@@ -0,0 +1,66 @@
...
+  // TLS settings (certs, sans, sni, etc.) required to talk to the remote server.
+  // User is responsible for mounting these certs into Pilot (e.g. using Kubernetes secrets/files).
+  TLSSettings tls_settings = 5;
```
costinm: Hmm. Maybe a more elegant solution is to just define a DestinationRule for the registry ? We have that already. 

And maybe we can also use a ServiceEntry instead of address/port. A 'registry' is really just a regular service - may be external, nothing special about it to require to duplicate configs.

networking/v1alpha3/mesh.proto
```diff
@@ -0,0 +1,66 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "networking/v1alpha3/destination_rule.proto";
+
+package istio.networking.v1alpha3;
```
costinm: Maybe this could be a separate package ( and alpha1 ) ? Keep in mind we said networking is 'beta' (despite the name).

costinm: Or maybe we can add it to the old mesh config ?

networking/v1alpha3/mesh.proto
```diff
@@ -0,0 +1,66 @@
...
+
+// Provides information about the endpoints in this network.  At this point
+// in time, an endpoint can be in only one network.  When Pilot ingests
+// service registry information from various registries, it maps the
+// endpoints to the networks defined here based on the CIDR range that
+// matches the endpoint IP.
+message Network {
+  string name = 1;
+  // e.g., 192.158.10.0/24
+  string endpoint_cidr = 2;
```
costinm: Optional ( or remove). Why do we need it ? Also needs to be repeated ( so it includes IPv6 and multiple ranges ). But not sure what is the use case - networks may have duplicated endpoint CIDRs, and I don't think we use them in any form.

networking/v1alpha3/mesh.proto
```diff
@@ -0,0 +1,66 @@
...
+  // A DNS name or IP address associated with the gateway that mediates access from remote networks.
+  // The address must be reachable from sidecars in other networks.
+  repeated string gateways = 3;
```
costinm: Now missing the port - maybe url ? 

But I think better to use a reference to a ServiceEntry - after all a Gateway is not different enough from an external service to require special/duplicated config.

networking/v1alpha3/mesh.proto
rshriram: This is logically part of networking. The whole thing is to establish L4/l7 connectivity to endpoints across multiple registries. So creating a new api domain doesn't make sense. 
Second, the mesh config is a static one time config map. We don;t have watchers, don't have the ability to react to changes in the mesh config, etc. May be we could do that now

networking/v1alpha3/mesh.proto
rshriram: we need this so that when we import information from multiple registries, we can automatically tag the network to which each endpoint belongs. A service registry doesn't explicitly tag the network name to every endpoint.

networking/v1alpha3/mesh.proto
costinm: Zack: I believe we mentioned they are 'beta' ( still called alpha3 to confuse people - and because k8s has some bugs preventing rename). 

I don't think configuring istio-system needs to be in the same package as user-settable configs.
We believe we have a watcher - but we don't use it, half done - we can certainly fix this. I'm not sure mesh config is the right place - but but I'm pretty sure this is not a CRD that should be exposed to regular users. It will only work on istio-system anyways ( since secrets to mount must be in same package with pilot). And will require updating restarting pilot - to add new mounts, like we have to do with Ingress, if we use mounted files.

networking/v1alpha3/mesh.proto
costinm: Ok with keeping them together. 

But I am not sure in most cases users will want to configure the Gateway IP manually. It's very hard - add a remote cluster, wait for the cluster IP to stabilize, update the configs in all clusters that need to view that. If you happen to reinstall - or something happens with the load balancer IP - go and modify each cluster.

I don't mind having a way to explicitly set a Gateway - but it should be the exception. For 99% of users who use the default names - the Gateway IP is available once you load the remote services.

I also didn't like mixing them because networks and registries are orthogonal - you can have a single cluster ( for example consul ) that span multiple networks.  It also confuses people who are already mixing zvpn with multi-cluster, and clusters with registries, and registries with networks.

networking/v1alpha3/mesh.proto
costinm: url can be uds:// or foo://.

networking/v1alpha3/mesh.proto
costinm: You mean a registry spanning multiple networks, and the registry code auto-assigning networks ? 
In this case - you also need to add the registry name to the network. Same cidr may be picked by multiple registries. 

BTW - a Network may span multiple registries, and a registry may have multiple networks.

costinm:
I think the idea of using TLSConfig from DestinationRule - or DestinationRule directly - may be the solution we were looking for wrt authentication. 

One extra detail: one of the reasons we used Secret to configure remote K8S registries is that it's far easier to take a .kubeconfig file and shove it in a Secret than to extract the different parts ( there are many auth schemes supported by k8s besides mtls). But if we could extend DestinationRule or define a mechanism to fetch credentials from a Secret, without mandating PEM files - we could solve other problems as well, and preserve the simplicity of the current approach.

costinm:
How about focusing on Network configuration - which is what we really need for zvpn, and leave registires for 1.2 ? A network may span multiple registries, a registry may span multiple networks - and no need to tie them together. With isolated-pilot-per-cluster you don't even need to worry about multiple  registries and credentials. 

It is clear we can't require all registries use mtls ( it's one of the many auth mechanisms ), and we shouldn't be in the business of re-defining kubeconfig and consul and other registries config and auth mechanisms. 

If we focus only on networks: we need to define the IP address(es) of the remote gateway. And I think using a ServiceEntry for that is the natural thing to do - it's an external service. The only thing 'special' is that ServiceEntry happens to be a gateway for the network. We can express this in the ServiceEntry itself ( with some restrictions that a SE defining gateaway for net must be in istio-system, or maybe not). It is ok for a network to have multiple gateways. 

If extending ServiceEntry with a 'repeated 

gateway-for: 
- net1
- net2

is too confusing (since most users will not be authorized to define gateways and don't need to confused about this section) we can define a config section in mesh config (and use the watcher), with 

   - network: foo
     gateway: service-entry-reference

Technically the 'gateway' for reaching a remote network may be a service running in the local cluster, so it may happen to be configured using local endpoint discovery ( like the egress - or it might be egress gateway itself ). So extending ServiceEntry is not my top choice, since SE is mostly for external stuff. 

Extetnding DestinationRule remains my preferred solution. We can indicate that a destination service must be reached via a gateway - we can also indicate that a destination is a gateway for a set of networks. And it keeps 'how to reach stuff' in DestinationRule, independent of discovery/ServiceEntry.

linsun:
@rshriram thank you much for proposing this!   

Scope q:  is this mesh config designed for both multi cluster istio scenarios we talked about at the WG mtg?  e.g.  https://preliminary.istio.io/docs/setup/kubernetes/multicluster-install/gateways/ & https://preliminary.istio.io/docs/setup/kubernetes/multicluster-install/vpn/

Agreed on just focus on what we need for 1.1, which is network name and gateway info... not sure how CIDR is planned to be used, what is the scenario to have it exposed?

networking/v1alpha3/mesh.proto
rshriram: I should have clarified. Gateway need not be IP. Its DNS/IP. So, if you use DNS name, then envoy will resolve, won't it?

mesh/v1alpha1/data_source.proto
```diff
@@ -0,0 +1,67 @@
...
+// DataSource describes information about a remote service registry or a
+// configuration store inside a mesh. A single control plane instance can
+// interact with one or more data sources.
+message DataSource {
+  // A unique name assigned to this data source.
+  string name = 1;
+
+  // The type of the data source determines the protocol used to interact with
+  // it and obtain information.
+  enum SourceType {
```
ZackButcher: Should we add filesystem based SourceType given all our servers support it (AFAIK)?

mesh/v1alpha1/config.proto
```diff
@@ -128,5 +129,113 @@ message MeshConfig {
   reserved 21;
 
   // Address of the galley service exposing the Mesh Control Protocol (MCP).
+  // TODO: remove
```
costinm: I don't think we shipped it in 1.0 - so we can remove it before 1.1 ships. 

It should be a list of MCP address anyways - galley is one implementation, but we don't want to 
name each implementation and add a field.

mesh/v1alpha1/config.proto
```diff
@@ -128,5 +129,113 @@ message MeshConfig {
...
+  // access to the mater nodes of each cluster (to talk to the API
+  // servers), the following configuration describes a mesh spanning all
+  // the clusters and using cluster-local istio-ingressgateway to route
+  // traffic to endpoints within the cluster.
+  //
+  // serviceRegistries:
+  // - name: us-east-c1
+  //   type: KUBERNETES
+  //   credentials:
+  //     kubernetesSecret: us-east-c1-secret
```
costinm: 1. Please don't hardcode kubernetes. Maybe "secret:..." Or better - "configref:  ....". 
2. Indicate that the config ref is a kubeconfig including private key or credentials in the case of k8s, for backward compat with 1.0 implementation. 
3. no need for secrets to be named something-secret :-)
4. Also for backward compat - and simplicity - I think it's reasonable to allow current auto-loading of k8s registires to continue. It's simpler and happen immediately ( and tested )

rshriram: 1. Secrets are there only in kubernetes :). Other platforms (cloud foundry for example) doesn't have secrets.
2. I have already added that info in the comment above the field kubernetes_secret. It has to be a kubeconfig. So the effort required is to simply reuse the existing code as is.
4. Its also specified in the comment that if no registries are provided, then Pilot will load service registry info from the ambient kubernetes cluster. [Not sure if we still have to maintain the multicluster secret registry and all the other stuff, given that the implementation is still very much alpha as per feature status, and other things].

mesh/v1alpha1/config.proto
```diff
@@ -128,5 +129,113 @@ message MeshConfig {
...
+  //   credentials:
+  //     kubernetesSecret: us-west-c1-secret
+  // configStores: #all configuration from us-east-c1
+  // - name: us-east-c1
+  //   type: KUBERNETES
+  //   credentials:
+  //     kubernetesSecret: us-east-c1-secret
+  // networks:
+  // - name: us-east # some arbitrary name to group endpoints
+  //   registries:
```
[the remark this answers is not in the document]
rshriram: This is the GKE use case. one VPN, spanning multiple AZs/regions.. They are all directly accessible to one another. I don't think GKE puts anything specific on each node saying here is the name of the network. 

Also, this name is a very arbitrary string. Its just a simple grouping construct. Nothing to do with anything in the layer below. 

If you apply this on say an IBM cluster that has a separate network for each k8s, then you will see that there is no way for the k8s registry to know the network as all k8s clusters have same pod cidr/service cidr. The explicit config here helps. 

The reason I put network on top of the registry is because an orchestration platform with a service registry can be on only one network. You can't have two different sets of pod cidrs on a k8s cluster, such that one set of pods can't access the other set of pods [when Istio is not installed]. On the other hand, you can have multiple k8s clusters on the same routable L3 network [gke style, or on prem where you use some calico etc. to establish pod-pod connectivity across clusters].

mesh/v1alpha1/config.proto
```diff
@@ -128,5 +129,113 @@ message MeshConfig {
...
+  //   - us-east-c1
+  //   - us-east-c2
+  //   gateways:
+  //   - registryServiceName: istio-ingressgateway.istio-system.svc.cluster.local
```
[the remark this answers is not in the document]
rshriram: ServiceEntry is also a registry. We don't differentiate it. But I can add it in comments

mesh/v1alpha1/config.proto
```diff
@@ -128,5 +129,113 @@ message MeshConfig {
...
+// relation to the locality of the endpoint. The endpoint locality will be
+// obtained from the service registry.
+message Network {
+  // A unique name assigned to the network.
+  string name = 1;
+
+  // One or more service registries that are in this network. Endpoints
+  // from these registries will be directly reachable to one another.
+  // The names of the registries should correspond to the data sources.
+  repeated string registries = 2;
```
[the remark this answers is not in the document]
rshriram: Yeah it can't. I don't know how the consul thing operates. Logically, every endpoint in a service registry must be able to reach one another. If you have network policies that do firewalling, its not part of this design - because we route traffic to endpoints in the registry. Firewalls/policies decide if the endpoint should be allowed to reach the other or not.


All said, our consul support is super brittle and broken. Thats the reason I omitted consul from the list because its likely going to be removed (we tell people to use the Consul K8S sync thing from hashicorp).

mesh/v1alpha1/config.proto
```diff
@@ -128,5 +129,113 @@ message MeshConfig {
...
+      // lookup the service from the service registries in the network and
+      // obtain the endpoint IPs of the gateway from the service
+      // registry. Note that while the service name is a fully qualified
+      // domain name, it need not be resolvable outside the orchestration
+      // platform for the registry. e.g., this could be
+      // istio-ingressgateway.istio-system.svc.cluster.local.
+      string registry_service_name = 1;
+
+      // IP associated with the gateway.
+      string ip = 2;
```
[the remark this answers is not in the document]
rshriram: Sure. Just remember that if this is a DNS address, there is no EDS. ITs all through CDS with strict_dns clusters.

mesh/v1alpha1/data_source.proto
```diff
@@ -0,0 +1,67 @@
...
+// Credentials provide the requisite information for Istio Pilot/Galley to
+// interface with a remote data source (registry/config).
+message Credentials {
+  oneof creds {
+    // When interfacing with a Kubernetes data source, the
+    // kubernetes_secret specifies the name of the secret that contains the
+    // kubeconfig file associated with the remote kubernetes cluster.
+    string kubernetes_secret = 1;
+    
+    // When interfacing with a remote data source that supports Istio Mesh
```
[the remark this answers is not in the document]
rshriram: the TLS settings has a mode called ISTIO_MUTUAL . You can specify that and not have to specify any cert info. But it still leaves the door open in case you want to use custom certs.

mesh/v1alpha1/data_source.proto
```diff
@@ -0,0 +1,67 @@
...
+  SourceType type = 2;
+
+  // URI where this data source can be reached.
+  // Can be tcp://, https://, unix://, etc.
+  // Note: URI will be ignored if kubernetes_secret is specified
+  // as the kubeconfig file in the secret provides the URI.
+  string uri = 3;
+
+  // Credentials to access the remote data source
+  Credentials credentials = 4;
```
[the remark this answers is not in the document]
rshriram: ITs mTLS but there is no restriction (per taxonomy) that the CA has to be shared. And as I said above, you can simply specify ISTIO_MUTUAL and not worry about specifying anything else

rshriram:
@costinm you cannot define a registry without telling how to get the endpoints from the registry. It makes no sense to tell the user that "just specify some name, and make sure to use the same name when you follow the Istio task to define cluster registry using secrets". Thats not how our API design works. ITs brittle and given the state of our documentation, it will break again and again, just like this ad-hoc design

```
 kubectl label secret ${CLUSTER_NAME} istio/multiCluster=true -n ${NAMESPACE}
```

whose semantics are embedded in the code.

mesh/v1alpha1/data_source.proto
```diff
@@ -0,0 +1,67 @@
...
+    // When interfacing with a Kubernetes data source, the
+    // kubernetes_secret specifies the name of the secret that contains the
+    // kubeconfig file associated with the remote kubernetes cluster.
+    string kubernetes_secret = 1;
+    
+    // When interfacing with a remote data source that supports Istio Mesh
+    // Configuration Protocol (MCP), use tls_settings to specify the tls
+    // mode to use and the certificate information required to establish a
+    // TLS or mutual TLS connection to the MCP server.
+    istio.networking.v1alpha3.TLSSettings tls_settings = 2;
```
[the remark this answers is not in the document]
rshriram: Why?

mesh/v1alpha1/config.proto
```diff
@@ -128,5 +129,113 @@ message MeshConfig {
...
+  // The names of the registries should correspond to the data sources.
+  repeated string registries = 2;
+
+  // The gateway associated with this network. Traffic from remote networks
+  // will arrive at the specified gateway:port. All incoming traffic must
+  // use mTLS.
+  message IstioNetworkGateway {
+    oneof gw {
+      // A fully qualified domain name of the gateway service.  Pilot will
+      // lookup the service from the service registries in the network and
```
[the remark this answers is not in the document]
costinm: Registries may not be always accessible - and it's perfectly possible to access a registry via a gateway (with Istio MTLS). My concern was use of IPs - most people use DNS, which can resolve to multiple IPs that may change.

linsun:
The new API is clean and reasonable imho.  I can define mesh config, and inside mesh to define network (including gateway address & port) and datasources (including source type & creds).   With this, someone who is interested in leveraging split horizon can easily configure these configuration on the primary cluster that has the mesh control plane.

mesh/v1alpha1/mesh.proto
```diff
@@ -0,0 +1,184 @@
...
+    // The locality associated with an explicitly specified gateway (i.e. ip)
+    string locality = 4;
+  }
+
+  repeated IstioNetworkGateway gateways = 3;
+}
+
+// Credentials provide the requisite information for Istio Pilot/Galley to
+// interface with a remote data source (registry/config).
+message Credentials {
```
costinm: Can we leave credentials out ? K8S doesn't use mtls in all cases, and we may want to get away from using Secret (node agent, etc), and we don't want to have k8s-specific names in the API.  I don't think we have any other 'kubernetes-' attribute.

mesh/v1alpha1/mesh.proto
```diff
@@ -0,0 +1,184 @@
...
+// interact with one or more data sources.
+message DataSource {
+  // A unique name assigned to this data source.
+  string name = 1;
+
+  // The type of the data source determines the protocol used to interact with
+  // it and obtain information.
+  enum SourceType {
+    // Kubernetes
+    KUBERNETES = 0;
```
costinm: "For backward compatibility with 1.0, K8S data sources will also be created automatically from Secrets labeled .... , using the ObjectName of the secret "

I assume you don't want to break existing launched feature. We can deprecate the use of secrets once we have feedback from users - on which they prefer to use, the install/upgrade time changes in mesh config or the dynamic method from 1.0

mesh/v1alpha1/mesh.proto
```diff
@@ -0,0 +1,184 @@
...
+    // mode to use. If the MCP server uses Istio MTLS and shares
+    // the root CA with Pilot, specify the TLS mode as ISTIO_MUTUAL.
+    istio.networking.v1alpha3.TLSSettings tls_settings = 2;
+  }
+}
+
+// DataSource describes information about a remote service registry or a
+// configuration store inside a mesh. A single control plane instance can
+// interact with one or more data sources.
+message DataSource {
```
costinm: Is this something that will be implemented in 1.1 ? Anyone working on it ? Don't forget testing - and the fact mesh config is replaced on upgrade...

costinm:
The network part is reasonable.

Strongly disagree with the Credentials part - and I really don't believe
changing k8s registries from a dynamic, one-step process to a
2-step plus upgrade is a good idea. And I doubt a proper implementation can
be available in 1.1 time frame for this new registry.

How about merging the network part - Gateway, etc - and leaving registries
for 1.2 ? Having the API without implementation is not ideal,
and it took a lot of testing to get the Secrets stuff working well.

mesh/v1alpha1/mesh.proto
```diff
@@ -0,0 +1,61 @@
...
+package istio.mesh.v1alpha1;
+
+option go_package="istio.io/api/mesh/v1alpha1";
+
+// Network provides information about the endpoints in a routable L3
+// network. A single routable L3 network can have one or more service
+// registries. Note that the network has no relation to the locality of the
+// endpoint. The endpoint locality will be obtained from the service
+// registry.
+message Network {
```
costinm: Well - maybe we can rename the file to 'network.proto' ?

PR #657, opened 2018-10-18
Cleanup Mesh config

rshriram:
I have removed all unused options


rshriram:
??
this is the mesh config map. the Auth stuff is not used in our code. I retained the field and made it reserved.

PR #658, opened 2018-10-18
add flag to meshconfig to indicate whether token volume mount it enabled 

quanjielin:

This flag is used by secret discovery service(SDS). 
If set to true(prerequisite: https://kubernetes.io/docs/concepts/storage/volumes/#projected), K8s API server mounts k8s service account JWT to envoy container, which will be used to generate key/cert eventual

pilot generates corresponding sdsconfig when token volume mount is enabled.
sidecar injector inject token volume mount when this flag is enabled.


linsun:
Could you pls explain in what scenarios and why user needs to enable this flag in mesh config?

quanjielin:
This flag is to make secret discovery service(SDS) work in master branch, if set, k8s api server mount k8s sa jwt to envoy, which will be passed to sds server (to generate ke/cert eventually). 

https://kubernetes.io/docs/concepts/storage/volumes/#projected for more information about token volume mount.

mesh/v1alpha1/config.proto
```diff
@@ -130,6 +130,9 @@ message MeshConfig {
   // Address of the galley service exposing the Mesh Control Protocol (MCP).
   string galley_address = 22;
 
-  // If set to true, identity is generated using k8s sa jwt token.
-  bool enable_sds_token = 23;
+  // $hide_from_docs
+  // This flag is used by secret discovery service(SDS). 
+  // If set to true(prerequisite: https://kubernetes.io/docs/concepts/storage/volumes/#projected), K8s API server 
+  // mounts k8s service account JWT to envoy container, which will be used to generate key/cert eventually.
```
diemtvu: May be more accurate to say `If set to true, Istio will inject volumes mount for k8s service account JWT to ...`.
Also, you may want to explain how it is used for non-k8s.

quanjielin: thanks, updated.

PR #661, opened 2018-10-19
undo breaking change

rshriram:
Apparently its not enough to convert unused fields as reserved.
Golang proto parser breaks if there are unknown fields in the JSON/yaml.


costinm:
Thanks !

PR #662, opened 2018-10-19
allow file as a config source

rshriram:


mesh/v1alpha1/config.proto
```diff
@@ -149,6 +151,43 @@ message MeshConfig {
...
+  // URI where this data source can be reached.  Can be tcp://, https://,
+  // unix://, file:/// etc.  Note: URI will be ignored if source is of type
+  // kubernetes.  If the datasource is a file, uri should contain the
+  // absolute path to the directory where the configuration is available.
+  string uri = 3;
+
+  // Credentials to access the remote data source. If the MCP server uses
+  // Istio MTLS and shares the root CA with Pilot, specify the TLS mode as
+  // ISTIO_MUTUAL.
+  istio.networking.v1alpha3.TLSSettings tls_settings = 4;
```
costinm: +1 on this - except the tls_settings.

mesh/v1alpha1/config.proto
```diff
@@ -149,6 +151,43 @@ message MeshConfig {
...
+    FILES = 2;
+  }
+
+  SourceType type = 2;
+
+  // URI where this data source can be reached.  Can be tcp://, https://,
+  // unix://, file:/// etc.  Note: URI will be ignored if source is of type
+  // kubernetes.  If the datasource is a file, uri should contain the
+  // absolute path to the directory where the configuration is available.
+  string uri = 3;
```
costinm: How about 'in-cluster' ? Or if registry is an accessible service, can it use the short name (https://istio-galley) ?

rshriram:
It seems none of this is necessary as file based config source is part of Galley configuration. And its unclear how Galley does mTLS, with san validation.

costinm:
Well - the use case was Pilot loading some initial config from files.
Including mesh config and other associated config needed to connect to
Galley.

On the second part - I did a bit of research, Galley uses 'native' Istio
MTLS, very similar with Pilot.

PR #663, opened 2018-10-19
Bring back removed Mesh config entries.

ozevren:
Based on diff of master against 76349c53b87f03f1e610b3aa3843dba3c38138d7 in release-1.0 branch.


mesh/v1alpha1/config.proto
```diff
@@ -104,7 +104,9 @@ message MeshConfig {
   // $hide_from_docs
   AuthPolicy auth_policy = 10 [deprecated=true];
 
-  reserved 11;
+  // $hide_from_docs
```
geeknoid: I don't think we want $hide_from_docs. Just mark the field as deprecated and it will be rendered appropriately by the doc infra.

ozevren: Done

rshriram: We don't want to render this.. as this is not implemented at all. Unless you mean marking something as deprecated will also hide it from the docs.

mesh/v1alpha1/config.proto
```diff
@@ -126,8 +127,37 @@ message MeshConfig {
...
+  }
+
+  // DEPRECATED. UNUSED.
+  // Set the default behavior of the sidecar for handling outbound traffic from the application.
+  // While the default mode should work out of the box, if your application uses one or more external services that
+  // are not known apriori, setting the policy to ALLOW_ANY will cause the sidecars to route traffic to the any
+  // requested destination.
+  // Users are strongly encouraged to use ServiceEntries to explicitly declare any external dependencies,
+  // instead of using allow_any.
+  OutboundTrafficPolicy outbound_traffic_policy = 17 [deprecated=true];
```
rshriram: This was never implemented from the beginning. Do we have to keep it/bring it back?

ozevren: Did we actually create any mesh config files with a value for this field? If so, yes.

mesh/v1alpha1/config.proto
ozevren: Marking deprecated will indicate that is deprecated, but otherwise it is visible.

What do you mean "not implemented at all"? Was the field available in 1.0 (I thought it was).

The higher-order decision bit is, whether there are config maps that have a value for this field or not. If we can say we won't have for sure, then we can simply remove it. If not (i.e. there was a default value put when config was being generated), then we can't remove it.

(rendering is a separate thing, I'll let you guys come to an agreement and do the what is agreed on ).

geeknoid:
We decided at TOC that deprecated fields should be visible in the docs. This tells the user "this field is going away soon" which is clearly different than "this is a new field that hasn't yet been documented". To avoid all confusion, the field is documented and gets removed from the docs when it gets removed from the implementation.

We haven't used the deprecation feature in the docs yet. Once this change goes in and I update the reference docs on istio.io, we can look at the rendering and decide whether we want something different there. If I recall correctly, deprecated fields just get rendered in a different color and it says something about being deprecated if you hover over the field. We could also use striketrhough or other approaches...

mesh/v1alpha1/config.proto
rshriram: Nope. As I said not even in docs or helm.

rshriram:
closing as another PR that does the same has merged

PR #664, opened 2018-10-19
Updating description of network definitions

rshriram:


mesh/v1alpha1/network.proto
```diff
@@ -27,10 +27,23 @@ message Network {
   // A unique name assigned to the network.
   string name = 1;
 
-  // One or more service registries that are in this network. Endpoints
-  // from these registries will be directly reachable to one another.
-  // The names of the registries should correspond to the data sources.
-  repeated string registries = 2;
+  message NetworkEndpoints {
+    oneof ne {
+    // A CIDR range for the set of endpoints in this network
```
costinm: Few more words here ? It assumes the endpoints in the mesh have non-overlapping CIDR ranges ? 

An endpoint will be assigned to a network:
1. directly by the registry, if it can infer the network from its backends
2. by matching the registry name with one of the "from_registries" in the mesh config. A "from_registry" can only be assinged to a single network.
3. by matching the IP against one of the CIDR ranges in a mesh config network. The CIDR ranges must not overlap and be assigned to a single network.

hzxuzhonghu: >by matching the registry name with one of the "from_registries" in the mesh config. A "from_registry" can only be assinged to a single network.

@costinm What do you mean by `from_registries`? I can not find it.

mesh/v1alpha1/network.proto
```diff
@@ -27,10 +27,44 @@ message Network {
...
+    oneof ne {
+    // A CIDR range for the set of endpoints in this network. The CIDR
+    // ranges for endpoints from different networks must not overlap.
+    string from_cidr = 1;
+
+    // Add all endpoints from the specified registry into this network.
+    // The names of the registries should correspond to the secret name
+    // that was used to configure the registry (kubernetes multicluster) or
+    // supplied by MCP server.
+    string from_registry = 2;
```
hzxuzhonghu: can you explain how this can be used? I am still confused after through the istio code.

PR #669, opened 2018-10-22
MCP server configuration

rshriram:


mesh/v1alpha1/config.proto
```diff
@@ -149,6 +150,25 @@ message MeshConfig {
...
+
+// ConfigSource describes information about a configuration store inside a
+// mesh. A single control plane instance can interact with one or more data
+// sources.
+message ConfigSource {
+  // Address of the server implementing the Istio Mesh Configuration
+  // protocol (MCP). Can be IP address or a fully qualified DNS name.
+  string address = 1;
+
+  // Use the tls_settings to specify the tls mode to use. If the MCP server
```
rshriram: Please note: we need to specify TLS settings here because ultil we talk to a MCP server, we can't get destination rules etc. Also, these settings are for Pilot internal implementation (go/mtls, etc.) and not meant for Envoy. So, asking people to specify a destination rule will create bootstrap issues within Pilot code.

mesh/v1alpha1/config.proto
```diff
@@ -152,3 +155,18 @@ message MeshConfig {
...
   // Next available field number: 25
 }
+
+// ConfigSource describes information about a configuration store inside a
+// mesh. A single control plane instance can interact with one or more data
+// sources.
+message ConfigSource {
+  // Address of the server implementing the Istio Mesh Configuration
+  // protocol (MCP). Can be IP address or a fully qualified DNS name.
+  // Use file:/// to specify a file-based backend with absolute path to the directory.
```
ZackButcher:
```suggestion
  // Use file:// to specify a file-based backend with absolute path to the directory.
```

PR #670, opened 2018-10-23
Add mesh config to allow setting network policy fail open

yangminzhu:
https://github.com/istio/istio/issues/9483


yangminzhu:
@rshriram Could you help to merge this? Do we need this in both 1.1 and the potential 1.0.4?

PR #675, opened 2018-10-24
Add Protolock.

jeffmendoza:
Move the protoc docker files to tools/protoc. Add docker files for
protolock image. Update Makefile precommit hook and prow presubmit
check to run protolock.

The contents of `tools/protolock` will need to be built and pushed to `gcr.io/istio-testing/protolock:2018-10-23`  before merging for this PR to work.


jeffmendoza:
Also, we will need to get `protolock` on CircleCI.

ozevren:
@hklai @geeknoid PTAL. Jeff is introducing protolock (https://github.com/nilslice/protolock) into istio/api.

geeknoid:
Looks promising.

ozevren:
> Also, we will need to get `protolock` on CircleCI.

Is the reason for the call to protolock in each Makefile entry? Can we remove it to expedite things?

kyessenov:
With circle, you might want to put all the tools into the base image for circle jobs. That would also work for prow, by the way.

jeffmendoza:
> Is the reason for the call to protolock in each Makefile entry? Can we remove it to expedite things?

This is the recommended usage:
```
 $ protolock status && protoc -I ...
```
This way if a developer runs `make generate-routing-go`, for example, they will know right away that they broke the proto.

The status command is taking ~20ms on my computer.

jeffmendoza:
> With circle, you might want to put all the tools into the base image for circle jobs. That would also work for prow, by the way.

Ah, hmm, it looks like the Circle config is using "gcr.io/istio-testing/protoc:2018-06-12" as the image. So, we will need an image with both protoc and protolock it seems.

ozevren:
Perhaps we can use multi-stage builds to create a top-level image?
https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

I suspect we really want to version each tool separately.

jeffmendoza:
> Perhaps we can use multi-stage builds to create a top-level image?
Sounds good. I'll add that

jeffmendoza:
Added `tools/all` which we can build and use for CircleCI, and potentially Prow in the future.

ozevren:
Ok. Do you have permissions to upload the generated image to GCR? If not, @hklai can help there.
Once the image is there, let's get the presubmit passing and shoot for getting this in.

jeffmendoza:
No, I don't.

hklai:
@jeffmendoza I can add you to istio-testing so you can push, or I can clone this PR and do it myself (later today) if this PR is ready.

Let me know.

jeffmendoza:
I'm happy to do it myself to lighten your load.

jeffmendoza:
Looks like prow is working.

Circle CI config needs to be update to use the `gcr.io/istio-testing/api-build-tools:2018-10-24` image instead of the current one:

ap-circle

@hklai Can you do this?

ozevren:
That should be a config file .circleci folder. You should be able to update
this in your PR.

jeffmendoza:
@ozevren All CI config updated and passing.

PR #678, opened 2018-10-25
Minor correction to avoid confusion in ServiceEntry reference doc

ymesika:
This is a minor correction to prevent user confusion in the istio.io reference doc for ServiceEntry.


rshriram:
You have to fix the pb.go and html files

PR #679, opened 2018-10-25
Add HTTPRetry config to allow setting retryOn policy

hzxuzhonghu:
https://github.com/istio/istio/issues/8081


networking/v1alpha3/virtual_service.proto
```diff
@@ -867,6 +868,13 @@ message HTTPRetry {
 
   // Timeout per retry attempt for a given request. format: 1h/1m/1s/1ms. MUST BE >=1ms.
   google.protobuf.Duration per_try_timeout = 2;
+
+  // Specifies the conditions under which retry takes place.
+  // One or more policies can be specified using a ‘,’ delimited list.
+  // The supported policies can be found in
+  // "https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/router_filter#x-envoy-retry-on"
+  // and "https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/router_filter#x-envoy-retry-grpc-on"
+  string retryOn = 3;
```
rshriram: Note: when you implement this in istio/istio, make sure to set the retry plugin as wel
```retry_policy:
  retry_host_predicate:
  - name: envoy.retry_host_predicates.previous_hosts
  host_selection_retry_max_attempts: 3
```

along with other options from here. The above settings will prevent Envoy from selecting the same failed host (which is the default).

hzxuzhonghu: Thanks for reminder, will do it. And learn more about envoy configuration.

hzxuzhonghu:
@rshriram Is this need for master branch?  so that I can update vendor in istio.io/istio. And implement this in it.

PR #690, opened 2018-11-01
Fix style for TLS enum in networking.

jeffmendoza:
I don't think this requires any change in Istio code. This is mapped to:
https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/cert.proto

The code here: https://github.com/istio/istio/blob/release-1.1/pilot/pkg/networking/core/v1alpha3/gateway.go#L435 is copying it over, but only referencing the envoy proto.

Let's get this into 1.1 before it ships.


costinm:
Was this added after 1.0.x ? If not - need to test if upgrade works. I suspect this is safe, but not sure.

rshriram:
verified that it doesn't exist in 1.0.x

PR #692, opened 2018-11-01
rbac: rename field roleRef to role_ref

### LINKED ISSUE #686: RBAC proto uses improper style for roleRef
```
rbac/v1alpha1/rbac.proto:162:3:Field name "roleRef" must be lower_snake_case.
```

Please fix if possible and remove lint exception below. Thanks.

https://github.com/istio/api/blob/master/prototool.yaml#L9

yangminzhu:
Fixes #686 

/cc @jeffmendoza


jeffmendoza:
I think you will want to keep it as ID 2. Use `protolock commit --force` to override the backwards compatibility warning.

costinm:
Unless it was added recently in master - and never used in 1.0 (or 1.1)

ozevren:
The Yaml representation of both should names (before/after) should be the
same.

yangminzhu:
The `roleRef` field is added long time ago before 1.0. I tested this change with Istio and yaml config using `roleRef` and nothing breaks.
@costinm could you clarify more about why you think the rename here (`roleRef` -> `role_ref`) is not backward compatible?

costinm:
In general renames are not backward compatible when protos are converted to
json and yaml.

It is possible fooBar and foo_bar are ok for yaml - and might be ok for
json too, but this needs to be tested
for upgrade. Just testing istio builds and new resource works is not
sufficient.

ozevren:
Thanks. Is there a particular test suite that can be used to verify this?

yangminzhu:
@costinm There is no new resource and the user yaml file doesn't change here. What else do we need to test?

costinm:
No, the API upgrade has zero coverage. The new proto is not really covering
this - may make it easier to notice changes, but doesn't
verify that a change is backward compatible. We rely on unit tests to cover
implementations - but obviously they can't catch if previous
implementation or API will work on upgrade, only that the current version
is in sync with the golden and implementation.

Both API changes - and installer change - have broken upgrade countless
times.

I don't know what impact this particular change can have - it may be safe
or it may not be. Resource is represented as JSON in
apiserver, yaml in user configs. I guess trying to create a config with
1.0.3, upgrade to 1.1 with this PR applied, and check
 everything still works ? I don't know that the fact the yaml looks the
same is enough. If it does - then what's the point of the
change if user yaml is exactly the same ? We don't expose protos to the
users yet.

It's not a problem for APIs or values.yaml changes that didn't ship yet in
1.0.x - but for anything that has shipped I think
it's reasonable to request upgrade tests, even manual is better than
nothing.

ozevren:
I've started working on a golden-files based integration test for Galley. That should cover textual Yaml => MCP Proto path. That should be good enough for testing and ensuring that this change is safe.

Let's hold on to this change until that lands. After that, we can use it to prove that this is correct.

(Alternatively, we can manually test this)

PR #696, opened 2018-11-06
Protoc gen validate

jeffmendoza:
Add https://github.com/lyft/protoc-gen-validate to generate validation code from protos.

@hklai I need `gcr.io/istio-testing` push permissions again to push `gcr.io/istio-testing/api-build-tools:2018-11-06` and `gcr.io/istio-testing/protoc-gen-validate:2018-11-06` which are added/updated in this PR. Thanks!


kyessenov:
Is the plan here to annotate all APIs with protoc-gen-validate annotations? The validation functions are no-ops without the annotations.

kyessenov:
@geeknoid I think if we are to maintain the template model for adapters, then it would be great to switch to using declarative validation for adapter configs. The authors would not need to write validation code in adapter language.

jeffmendoza:
After chatting with Oz, we'll target the Mixer protos first. And look at removing some of the validation code for those in Galley.

PR #697, opened 2018-11-07
add trust domain to meshconfig

quanjielin:

later pilot needs to read meshconfig.trustdomain to construct secure naming


mesh/v1alpha1/config.proto
```diff
@@ -157,8 +157,12 @@ message MeshConfig {
   // will be used to generate key/cert eventually. This isn't supported for non-k8s case.
   bool enable_sds_token_mount = 23;
 
+  // The trust domain corresponds to the trust root of a system.
+  // Refer to https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#21-trust-domain
```
diemtvu: What's the expected behavior if this is not set?

quanjielin: add comment ``` // Fallback to old identity format(without trust domain) if not set.```

PR #701, opened 2018-11-09
Minor fix in istio.networking.v1alpha3.pb.html

AdamDang:
Delete the duplicated "the the"


geeknoid:
The .html file is actually a generated file. Could you fix this in the original .proto file, and then run make to generate the .html file?

geeknoid:
Hi,

This change needs to be applied to the .proto file. Once the change is done in the .proto file, then you do "make" at the top of the tree, and check in all changed files. This will include the .html file.

PR #704, opened 2018-11-12
Adding weights to service entry endpoints

rshriram:
The endpoint weights correspond to Envoy's LB endpoint weights. This will allow service entries for remote cluster endpoints to carry weights.


networking/v1alpha3/service_entry.proto
```diff
@@ -419,6 +419,10 @@ message ServiceEntry {
     // (that bridges networks n1 and n2), or the IP associated with a
     // standard service endpoint.
     string locality = 5;
+
+    // The load balancing weight associated with the endpoint. Endpoints
+    // with higher weights will receive proportionally higher traffic.
+    uint32 weight = 6;
```
frankbu: How does this work in conjunction with route destination weights? When do you use which?

rshriram: Route destination weights are for a set of endpoints.. a service version etc. This is within an envoy cluster (i.e. a route destination if you will), where you can have one single service instance receive more traffic than other service instances in the same version.

The weights here are used in the multicluster scenario where you use the ingress gateway address as the endpoint for a host (foo.ns1.global) in the remote cluster. If the service is present in many clusters, then you would have multiple ingrss gateway endpoints here. In such a scenario, we need a way to specify the weight associated with each gateway (proportional to the number of pods of that service in that cluster).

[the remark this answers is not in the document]
ZackButcher: And to answer my own question: not yet. [They still use a uint32 today, and must be in the range 1-128.](https://github.com/envoyproxy/envoy/blob/fad993e5aed40fba95897e9017afd19bdf170ed0/api/envoy/api/v2/endpoint/endpoint.proto#L84-L85) We should probably add a comment about the allowed range here.

costinm: Doesn't make sense - why would the weight be associated with the ServiceEntry ( the full set of endpoints) ? I think it needs to be part of the endpoint.

rshriram: Its part of endppint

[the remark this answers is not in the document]
rshriram: we can normalize in code across all endpoints in a cluster. No point in asking users to normalize

andraxylia:
I see this has been merged already, but isn't the weight here actually capacity? If this is meant to be used to show how many Endpoints are behind a Gateway.

rshriram:
Yep. The weight here is capacity. But the term weight is more well understood - service has N endpoints each with a different weight.

networking/v1alpha3/service_entry.proto
costinm: Sorry, missed the indentation. 
/lgtm

PR #707, opened 2018-11-16
Setup CI to test for incompatible changes against 1.0 and 1.1

jeffmendoza:
New make target `release-lock-status` uses protolock against lock files from 1.0 and 1.1. CI will block on any incompatible changes.


jeffmendoza:
cc @ozevren

jeffmendoza:
@ozevren Rebased. No changes needed to the status files, looks like we haven't broken backwards compatibility =)

scripts/release-locks.sh
```diff
@@ -0,0 +1,37 @@
...
+
+set -eu
+
+locks=$(find ./scripts/locks -type f -name '*proto.lock' | sort)
+fail=none
+
+for lock in $locks; do
+    echo "Testing $lock"
+    mv ./proto.lock ./proto.lock.bak
+    cp $lock ./proto.lock
```
ozevren: Sorry, just coming back to this.
This doesn't look safe.
Can you instead use the command-line flags to proto-lock?
--lockdir [.]		directory of proto.lock file
--protoroot [.]		root of directory tree containing proto files

The rest of the PR looks good to me.

jeffmendoza: Updated to use `--lockdir`.

scripts/release-locks.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/ash
```
ozevren: Change the name of the script to check-release-locks.sh?

Release being a verb itself makes it harder to make sense of what the script does, from its name.

scripts/locks/release-1.0/proto.lock.status
```diff
@@ -0,0 +1,118 @@
+CONFLICT: "AuthenticationPolicy" field: "INHERIT" has been removed, but is not reserved [mesh/v1alpha1/config.proto]
```
ozevren: Should we put these lock files to a separate, top-level folder?
releaselocks?
scripts doesn't seem to be the right place.

scripts/locks/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+1. Edit `scripts/release-locks.sh` and comment out the line `rm status`.
+
+1. Run `make release-lock-status`.
+
+1. Copy the file `status` over the file
+    `scripts/locks/release-<ver>/proto.lock.status`, corresponding to
+   the release version.
+
+1. `scripts/relese-locks.sh` can be reverted.
```
ozevren: relese -> release.

scripts/locks/README.md
```diff
@@ -0,0 +1,21 @@
+## How to update
+
+1. First ensure the changes you are making are not breaking backwards
```
ozevren: It is not clear exactly when this update should be done. Is this for breaking backwards compatibility, or for simply making a change to the API.

scripts/release-locks.sh
jeffmendoza: Done

scripts/locks/release-1.0/proto.lock.status
jeffmendoza: Done

scripts/locks/README.md
jeffmendoza: Done

scripts/locks/README.md
jeffmendoza: Done

Issue #708, opened 2018-11-17
Overwrite headers


itaysk:
Related to https://github.com/istio/istio/issues/9595 which I'm working on.

In Envoy we can overwrite a header by setting `append: false` in [HeaderValueOption](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/base.proto#envoy-api-msg-core-headervalueoption) but this is not exposed in `appendRequestHeaders` which is `map[string]string`.

This choice of append vs set should be reflected in the API. I propose adding `set_request_headers` (and response).

PR #709, opened 2018-11-17
allow setting headers

itaysk:
for https://github.com/istio/api/issues/708


networking/v1alpha3/virtual_service.proto
```diff
@@ -403,6 +403,14 @@ message HTTPRoute {
...
   // destination service.
   map<string, string> append_request_headers = 15;
+  
+  // HTTP headers to set (overwrite) before returning a response to the
+  // caller.
+  map<string, string> set_response_headers = 16;
+
+  // HTTP headers to set (overwrite) before forwarding a request to the
+  // destination service.
+  map<string, string> set_request_headers = 17;
```
rshriram: Instead of this, I suggest creating a new message like this:
```
 headers:
   Request:
     Set
     Remove
   Response:
     Set
     Remove
```
Do you think it’s sufficient to always set, for the most common use case? Append simply buys you the ability to have duplicate headers while common expectation is to set or unset

rshriram:
..

networking/v1alpha3/virtual_service.proto
itaysk: your suggestion makes alot of sense. both the reorg into one spec level field, and to always set by default. But can we eliminate the case for headers with multiple values (which is allowable by HTTP: https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

rshriram: We could always have Set, Add, Remove. Something like this:

```
message HeaderOperations {
  map<string, string> set
  map<string, string> add
  map<string, string> remove
}

message Headers {
  HeaderOperations request
  HeaderOperations response
}

Headers headers
```

itaysk: Yes this is what I meant. I'll make the necessary changes.

itaysk: @rshriram I assume that now `append_{request,reqponse}_headers` should be marked as deprecated?

itaysk: @rshriram additionally, doesn't it make sense to nest the messages like:

```
message Headers {
  HeaderOperations request
  HeaderOperations response

  message HeaderOperations {
    ...
  }
}

WDYT?

Comment (rshriram) [reply networking/v1alpha3/virtual_service.proto]: Yep.. go for it.. and please mark the others as deprecated

Comment (itaysk) [reply networking/v1alpha3/virtual_service.proto]: @rshriram updated based on our discussion

**Comment by itaysk** [comment]
Tests are failing, I looked at the logs and I'm not sure why. Something with the protolock which I'm not entirely familiar with. Can someone help clarify?

**Comment by itaysk** [comment]
OK please disregard my prev message, I did some reading on the build scripts and protolock and now understand I needed to commit the `proto.lock` file. Pushed the commit and now the build is passing.

**Comment by rshriram** [comment]
Do you have the implementation ready for this change? If so, I would push this to release-1.1 branch.. If not, then stick with master

**Comment by itaysk** [comment]
Yes I already wrote the core implementation, just need to finalize the auxiliary code like tests. I can have it ready for tomorrow.
(BTW I wanted to implement the validation for this this PR and submitted https://github.com/istio/istio/pull/10119 along the way, should I wait for it to be merged? it's a small fix)

**Comment by itaysk** [comment]
changes base to `release-1.1`

File: networking/v1alpha3/virtual_service.proto [hunk 2]
```diff
@@ -385,24 +385,49 @@ message HTTPRoute {
...
+  HeaderOperations response = 2;
+
+  // HeaderOperations Describes the header manipulations to apply
+  message HeaderOperations {
+    // Overwrite the headers specified by key with the given values
+    map<string, string> set = 1;
+    // Apppend the given values to the the headers specified by keys 
+    // (will create a comma seperated list of values)
+    map<string, string> add = 2;
+    // Remove a the specified headers
```
rshriram: typo: a .. "Remove the specified headers"

PR #710, opened 2018-11-19
fix naming convention between access log format vs encoding

GregHanson:


GregHanson:
related to https://github.com/istio/istio/issues/10056

mesh/v1alpha1/config.proto
```diff
@@ -121,10 +121,14 @@ message MeshConfig {
...
   
-  // Format for the proxy access log (text or json).
-  // Default value is text.
+  // Format for the proxy access log 
+  // Empty value results in proxy's default access log format 
   string access_log_format = 24;
 
+  // Encoding for the proxy access log (text or json).
+  // Default value is text.
+  string access_log_encoding = 27;
```
ozevren: Shouldn't this be an enum?

GregHanson: @mandarjog @rshriram care to weigh in?

mandarjog: Makes sense for this to be `enum`. 
The other PR probably is merged by now, so needs the same fix ...

GregHanson: done

PR #713, opened 2018-11-19
add access log encoding

GregHanson:


GregHanson:
related to istio/istio#10056

mandarjog:
Note: Even though this looks like changing the meaning of a field, the previous change had never gone out in a release.

PR #715, opened 2018-11-20
API changes to enable TCP keepalive configuration.

shakti-das:
API changes PR for https://github.com/istio/istio/issues/9378

Signed-off-by: Shakti <shaktiprakash.das@salesforce.com>


networking/v1alpha3/destination_rule.proto
```diff
@@ -351,15 +351,30 @@ message LoadBalancerSettings {
...
+//         keepalive:
+//           time: 7200s
+//           interval: 75s
 // ```
 message ConnectionPoolSettings {
   // Settings common to both HTTP and TCP upstream connections.
   message TCPSettings {
+    // TCP keepalive.
+    message TcpKeepalive {
+      // TCP keepalive time.
```
hzxuzhonghu: Expect a more detailed description, like envoy doc ` The number of seconds a connection needs to be idle before keep-alive probes start being sent. Default is to use the OS level configuration (unless overridden, Linux defaults to 7200s (ie 2 hours.)`

networking/v1alpha3/destination_rule.proto
```diff
@@ -351,15 +351,30 @@ message LoadBalancerSettings {
...
+      google.protobuf.Duration time = 1;
```
hzxuzhonghu: I doubt this should not be Duration type. 

https://github.com/envoyproxy/data-plane-api/blob/4619dc3f6246465a924cd3a19e3089ff0942e9be/envoy/api/v2/core/address.proto#L65-L77

networking/v1alpha3/destination_rule.proto
```diff
@@ -351,15 +351,30 @@ message LoadBalancerSettings {
...
+
+      // TCP keepalive interval.
+      google.protobuf.Duration interval = 2;
```
hzxuzhonghu: same here type mismatch

hzxuzhonghu:
I think we also need an additional field to set `keepalive_probes`

rshriram:
1. to release-1.1 branch
2. The issue asks for Mesh level configuration while what you are doing here is at DestinationRule level
 Granted that you can define a Global DestinationRule, but it will get overwritten when someone declares something local. Does it make sense to move this to mesh/config.proto (its a bit of a kitchen sink right now). I don't mind having this in the DestinationRule (in fact we should have it there in addition to mesh config I guess).

networking/v1alpha3/destination_rule.proto
shakti-das: Will add more details.

networking/v1alpha3/destination_rule.proto
shakti-das: I doubt the goal is to keep the same proto definition. I read in one of the PR review comments that the set of configuration options need to be added as required. Looking at the discussion in the linked issue, it seems we want to expose keepalive time and interval at the moment.

Regarding the Duration fields, considering it's a time related field, instead of specifying it as uint and mentioning in comments that the unit is seconds, I felt it's better to keep it as duration in Istio proto. In the impl, we can easily convert duration to seconds while populating Envoy proto fields.

shakti-das:
1. Will send PR to 1.1 branch. 
2. Will add it to MeshConfig + DestinationRule. Implementation will then have following defaulting behavior:
DestinationRule (if set, else next) > MeshConfig (if set, else next) > OS default (skip the envoy config)

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: https://github.com/istio/istio/issues/9378#issuecomment-436101687 
Cloud Foundry operators have requested support for control of TCP keepalive probe count and interval.

shakti-das: thanks for the reference, will add probes to the PR.

PR #717, opened 2018-11-20
default network scope

rshriram:


mesh/v1alpha1/config.proto
```diff
@@ -196,8 +196,36 @@ message MeshConfig {
...
+      // in addition to namespaces specified in sharedNamespaces.
+      CURRENT_NAMESPACE = 2;
+    }
+
+    // REQUIRED: The default scope associated with every sidecar in the mesh.
+    Mode default_scope = 1;
+
+    // Specifies one or more namespaces that should be accessible to all sidecars
+    // in the mesh (such as istio-system).
+    repeated string shared_namespaces = 2;
```
hzxuzhonghu: I can not understand the relationship with `default_scope `?

**What does below mean?**
eg. default_scope = ALL_NAMESPACES  and shared_namespaces =nsA
      default_scope = CURRENT_NAMESPACE and shared_namespaces =nsA

rshriram: Sorry.. Will improve docs.. Btw, read this https://docs.google.com/document/d/1x8LI3T7SHW-yDrrt3ryEr6iSs9ewMCt4kXtJpIaSSlI for more context

costinm: Also confused - it would read better if we don't break it into a separate message.
Just add the 2 fields.

I'm also not convinced we need to commit to 'shared_namespaces' - the discussion on 'defaults'
vs. 'overrides' is not completel resolved. MeshConfig is already 1/2 filled with deprecated or bad configs.

How about keeping 'shared' (whatever it means) experimental until we get more feedback.
We need both to go into values.yaml ( where we have a very different convention, and that's what people 
are really using - mesh config has not been directly editable or exposed to users since 1.0).
In values.yaml I would use:

networkScopes:
   enabled: true

  defaults: 
      imports....
  
and for default I would rather use the NetworkScopes resource format

rshriram: I would much rather have a CRD for the mesh global config with exact same semantics as network scope and stating that there can be only one such resource in the entire mesh. 

I don't want to stick the entire network scope config as a flag as you suggest. The shared mesh global namespace is not always known in advance, that we can't simply stick it as some arbitrary flag and tell people that to change it, you need to delete and restart an operational Pilot.

mesh/v1alpha1/config.proto
```diff
@@ -196,8 +196,36 @@ message MeshConfig {
...
+      // Configure routes to services in all namespaces.
+      ALL_NAMESPACES = 1;
+
+      // Configure routes to services in the same namespace as the sidecar
+      // in addition to namespaces specified in sharedNamespaces.
+      CURRENT_NAMESPACE = 2;
+    }
+
+    // REQUIRED: The default scope associated with every sidecar in the mesh.
+    Mode default_scope = 1;
```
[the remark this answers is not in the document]
rshriram: yeah but how do you specify the default namespace only network scope?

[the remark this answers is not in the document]
costinm: How about first implement the normal network scope - and we worry about default namespaces and other cases after we get more feedback ? This can be handled in Galley or MCP implementations, it is not something that needs to be a global mesh config that we'll be stuck for a long time.

Don't forget we'll need to test this as well. 

One istio-defaults namespace should be enough for 1.1

mesh/v1alpha1/config.proto
costinm: I don't even want to think of what will happen when this CRD is updated and how we'll test that.

Again - the MeshConfig is a big mess, but it's alpha. It is not exposed to users - so whatever goes in doesn't really matter that much, so I won't fight too much over is. Just please don't add the same craziness to values.yaml, that's the one actually exposed to users.

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
rshriram: Yes but thats basically what this thing is doing. We will also be stuck with backward compatibility if we hardcode some namespace and that happens to conflict with something else. We can always not implement this in the API and hide it in the docs but its much better than choosing some adhoc implementation and then trying to make everyting else work with it.

costinm:
One problem: default should be the current behavior, and we must add the
field commented out to mesh config.
We had problems before with roll-back - basically new pilot with old config
map should work.
So no setting == current 'all namespaces'.

costinm:
We should do a diff and make sure all 1.1 changes in mesh config are
compatible and not set by default, we already have a problem
in 1.0.4 with a field added to mesh config (the fail open).

rshriram:
> ```One problem: default should be the current behavior, and we must add the
field commented out to mesh config.
We had problems before with roll-back - basically new pilot with old config
map should work.
So no setting == current 'all namespaces'.```

Thats the reason for enclosing the whole thing in a NetworkScope object. If not specified, its nil and so we assume all namespaces. If specified (i.e. newer config map, etc.), then we force user to explicitly choose the mode.

costinm:
If that's the reason - then we don't need the object, just the enum with
default value as 'all namespaces' will work fine.
Or just a bool, as I suggested. We're not trying to force user to do
anything ( and as I mentioned - user doesn't interact
directly with this file anyways, it's just the template that'll be a bit
more complicated).

mesh/v1alpha1/config.proto
```diff
@@ -199,8 +199,45 @@ message MeshConfig {
   // Fallback to old identity format(without trust domain) if not set.
   string trust_domain = 26;
 
+  // Network scope settings for the mesh. Network scope controls the reachability of
+  // sidecars to other services in the mesh.
+  message NetworkScope {
+    enum Mode {
+      // Invalid mode
+      INVALID = 0;
```
frankbu: why is this one needed? ALL_NAMESPACES is the default right?

rshriram: True. Just building safeguards for future when we need to distinguish between an enum that’s unset vs enum set to a default value. We had a couple of cases like this. Notably the service entry resolution one.

mesh/v1alpha1/config.proto
```diff
@@ -199,8 +199,45 @@ message MeshConfig {
...
+
+    // REQUIRED: The default scope associated with every sidecar in the mesh.
+    Mode default_scope = 1;
+
+    // Specifies one or more namespaces that should be accessible to all
+    // sidecars in the mesh (such as istio-system). This field has no
+    // effect when the default scope is set to ALL_NAMESPACES and a sidecar
+    // has no explicit NetworkScope resource attached to it. However when
+    // the sidecar has a NetworkScope resource attached to it (and the
+    // default scope is set to ALL_NAMESPACES or CURRENT_NAMESPACE),
```
linsun: Sorry but the description here is a little confusing... It is stated the field has no effect with default is ALL_NAMESPACES, however,  this line explains how ALL_NAMESPACES and the field can work together.

PR #718, opened 2018-11-21
API changes to enable TCP keepalive configuration.

shakti-das:
API changes PR for istio/istio#9378 for release 1.1

Signed-off-by: Shakti <shaktiprakash.das@salesforce.com>


shakti-das:
Comments from previous PR for master:
> to release-1.1 branch
> The issue asks for Mesh level configuration while what you are doing here is at DestinationRule level
> Granted that you can define a Global DestinationRule, but it will get overwritten when someone declares something local. Does it make sense to move this to mesh/config.proto (its a bit of a kitchen sink right now). I don't mind having this in the DestinationRule (in fact we should have it there in addition to mesh config I guess).

This PR adds tcp_keepalive settings in DestinationRule and MeshConfig. The assignment logic would be as follows:
DestinationRule (if set, else next) > MeshConfig (if set, else next) > OS default (skip the envoy config)

Skipped keepalive probes and only kept keepalive time and interval for now. OS default probes will be sent.

shakti-das:
@hzxuzhonghu pointed out in the other thread that probes is also part of requirement. So, adding probes as per istio/istio#9378 (comment). 

I was thinking with these 3 items added, the message is similar to the one used in envoy proto. Does istio ever use imported protos from envoy? Code search doesn't show any prior art where envoy messages are reused in istio though, maybe to avoid dependency issues?

networking/v1alpha3/destination_rule.proto
```diff
@@ -351,15 +351,39 @@ message LoadBalancerSettings {
...
   // Settings common to both HTTP and TCP upstream connections.
   message TCPSettings {
+    // TCP keepalive.
+    message TcpKeepalive {
+      // Maximum number of keepalive probes to send without response before
+      // deciding the connection is dead. Default is to use the OS level configuration
+      // (unless overridden, Linux defaults to 9.)
+      uint32 probes = 1;
+
+      // The number of seconds a connection needs to be idle before keep-alive
```
hzxuzhonghu: >The number of seconds

not accurate, actually we can set `100s` `10m` `1h`

shakti-das: will update it to time duration

shakti-das:
@rshriram The review comments from PR #715 are implemented here. Please review.

PR #719, opened 2018-11-27
NetworkScopes

rshriram:


networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+
+option go_package = "istio.io/api/networking/v1alpha3";
+
+// `NetworkScope` describes the set of services that a workload depends on
+// for its operation. In other words, it describes the properties of egress
+// traffic from a given workload. By default, the service mesh established
+// by Istio will have a full mesh connectivity - i.e. every workload will
+// have proxy configuration required to reach every other workload in the
+// mesh. However most connectivity graphs are sparse in practice. The
+// NetworkScope provides a way to prune the connectivity graph
```
costinm: s/to prune the graph/to declare dependencies/ ? I'm not sure 'prune' is a very clear term, we used it in the internal discussions.  

NetworkScope is only controlling the immediate dependency - it is true that a combination of NetworkScopes 
affects the 'graph'.

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.networking.v1alpha3;
+
+option go_package = "istio.io/api/networking/v1alpha3";
+
+// `NetworkScope` describes the set of services that a workload depends on
+// for its operation. In other words, it describes the properties of egress
```
frankbu: Since egress usually mean external (outside the mesh) calls in our docs, maybe it would be better to say `outgoing traffic` instead of `egress traffic` for NetworkScope

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// traffic from a given workload. By default, the service mesh established
+// by Istio will have a full mesh connectivity - i.e. every workload will
+// have proxy configuration required to reach every other workload in the
+// mesh. However most connectivity graphs are sparse in practice. The
+// NetworkScope provides a way to prune the connectivity graph
+// (i.e. dependencies) associated with each workload.
+//
+// Services and configuration in a mesh are organized into one or more
+// namespaces (e.g., a Kubernetes namespace or a CF org/space).  Workloads
```
costinm: I would remove this sentence - or rewrite it. "reach other workloads" might be a bit confusing - there are traffic and other policies that prevent reaching. How about 'by default they have an implicit dependency on workloads in same namespace' ?

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// NetworkScope resource exists in a given namespace. The set of
+// dependencies specified in a NetworkScope resource will be used to
+// compute the [outbound] connectivity graph for every workload in the
+// namespace.
+//
+// NOTE: If workloads in the mesh depend only on other workloads in the
+// same namespace, use the mesh global setting to specify the default
+// network scope as CURRENT_NAMESPACE. To facilitate incremental pruning of
+// the connectivity graph, the default network scope for the mesh is set to
+// ALL_NAMESPACES. In otherwords, every workload will be able to reach
```
frankbu: other words

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// mesh. However most connectivity graphs are sparse in practice. The
+// NetworkScope provides a way to prune the connectivity graph
+// (i.e. dependencies) associated with each workload.
+//
+// Services and configuration in a mesh are organized into one or more
+// namespaces (e.g., a Kubernetes namespace or a CF org/space).  Workloads
+// in a namespace will be able to reach other workloads in the same
+// namespace. To declare dependencies on workloads in other namespaces, a
+// NetworkScope resource has to be specified in the current
+// namespace. *_Each namespace can have only one NetworkScope
```
costinm: How about defining a name ("default") - to make it easier to write RBAC rules - instead of leaving it 'arbitrary name'?

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// resource_*. The behavior of the system is undefined if more than one
+// NetworkScope resource exists in a given namespace. The set of
+// dependencies specified in a NetworkScope resource will be used to
+// compute the [outbound] connectivity graph for every workload in the
```
costinm: I would also remove graph from here, it may be confusing.

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// namespace.
+//
+// NOTE: If workloads in the mesh depend only on other workloads in the
+// same namespace, use the mesh global setting to specify the default
```
costinm: Please refer to values.yaml configs - we currently don't support user directly editing mesh config ( in default installer)

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// The following examples illustrate a few specific use cases of NetworkScope.
+//
+// The example below delcares a NetworkScope resource in the prod-us1
+// namespace. The workload with the `version: newlandingpage` label will be
+// able to reach all services declared in the egress namespace as well as
+// the landing.qa.foo.com service in the qa namespace. All other workloads
+// in the prod-us1 namespace will be able to reach the services in the
+// egress namespace only.
+//
+//
```
frankbu: delete extra line

networking/v1alpha3/network_scope.proto
costinm: If workloads in the mesh depend only ...., create an empty NetworkScope resource.

If all namespaces have NetworkScope declared or only depend on same workloads - upgrade
istio settings by setting helm "networkScope.enabled=true" or adding

networkScope:
   enabled: true 

to values.yaml.

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: NetworkScope
+// metadata:
+//   name: default
+//   namespace: prod-us1
+// spec:
+//   dependencies:
+//   - sourceWorkloadLabels:
```
costinm: Are we going to implement sourceWorkloadLabels in 1.1 ? If not ( and probably we can leave it for 1.2, and just do namespace ) - may be good to hide or comment out this section.

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+// The configuration above will allow workloads in the metrics-collection
+// namespace with the labels `app: metricsScraper` to access service in any
+// namespace while other workloads in the same namespace will be configured
+// for namespace local access as per the global default network scope
+// (CURRENT_NAMESPACE).
+//
+message NetworkScope {
+  // REQUIRED. The set of services that workloads in this namespace are
+  // expected to talk to, in addition to other workloads in the same
+  // namespace. Dependencies describe the properties of egress traffic from
```
frankbu: outgoing traffic

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,160 @@
...
+//       version: newlandingpage
+//     imports:
+//     - namespace: qa
+//       host: landing.qa.foo.com
+//     - namespace: egress
+//   - imports:
+//     - namespace: egress
+// ```
+//
+// In a mesh where the default network scope is set to CURRENT_NAMESPACE
```
costinm: I'm not sure - if NetworkScope is specified in a namespace, it should be respected regardless of the default. 
So it doesn't matter what the default is.

networking/v1alpha3/network_scope.proto
costinm: +1 - I would also avoid 'egress' in the example.

networking/v1alpha3/network_scope.proto
rshriram: good point. removing it..

mesh/v1alpha1/config.proto
```diff
@@ -199,8 +199,45 @@ message MeshConfig {
   // Fallback to old identity format(without trust domain) if not set.
   string trust_domain = 26;
 
+  // Default network scope settings for the mesh. Network scope controls the reachability of
+  // sidecars to other services in the mesh.
+  message DefaultNetworkScope {
+    enum Mode {
+      // Invalid mode
+      INVALID = 0;
```
frankbu: why is this one needed? ALL_NAMESPACES is the default right?

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,151 @@
...
+  // Dependency describes a workload and the set of service dependencies
+  // for the workload.
+  message Dependency {
+    // REQUIRED: Import describes the set of namespaces whose exported
+    // services will be accessed by the workloads selected by the
+    // sourceWorkloadLabels. The sidecars attached to the workloads will be
+    // configured with information required to reach other services in the
+    // same namespace and the imported services. In addition to the
+    // explicitly specified namespaces, namespaces specified in the global
+    // mesh config (through NetworkScope.sharedNamespaces) will also be
```
frankbu: DefaultNetworkScope.sharedNamespaces

mesh/v1alpha1/config.proto
```diff
@@ -199,8 +199,45 @@ message MeshConfig {
...
+
+      // Configure routes to services in all namespaces.
+      ALL_NAMESPACES = 1;
+
+      // Configure routes to services in the same namespace as the sidecar
+      // in addition to namespaces specified in sharedNamespaces.
+      CURRENT_NAMESPACE = 2;
```
mbanikazemi: Won't OWN_NAMESPACE be better? If yes, change throughout the files.

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,153 @@
...
+    // mesh config (through defaultNetworkScope.importNamespaces) will also be
+    // imported.
+    repeated Import imports = 2;
+  };
+
+  // REQUIRED. The set of services that workloads in this namespace are
+  // expected to talk to, in addition to other workloads in the same
+  // namespace. Dependencies describe the properties of outbound traffic from
+  // a given workload.
+  repeated Dependency dependencies = 1;
```
hzxuzhonghu: why `dependencies` and `imports` an array  at the same time?

networking/v1alpha3/network_scope.proto
hzxuzhonghu: Have the same doubt: is cluster scoped `NetworkScope` enough?

mesh/v1alpha1/config.proto
frankbu: OWN_NAMESPACE is ambiguous because own has two meanings: "my own" or "owned by". Maybe SAME_NAMESPACE is a better choice?

networking/v1alpha3/network_scope.proto
frankbu: Dependency has other things, like source pod selectors, but I guess that's going to be added in a future release.

rshriram: Because workload labels will be added to dependencies in future..

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,153 @@
...
+//   - imports:
+//     - namespace: '*'
+// ```
+//
+// The configuration above will allow workloads in the metrics-collection
+// namespace to access service in any namespace while workloads in other
+// namespaces will be configured for namespace local access as per the
+// global default network scope (CURRENT_NAMESPACE).
+//
+message NetworkScope {
```
mbanikazemi: I feel the use of the word Network in NetworkScope is not really desirable. I find it confusing and there is no need for someone using Istio to define anything "network" related. I am not sure what a better name would be. Would IstioServiceScope be better? I hesitate to suggest using ServiceScope.

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,153 @@
...
+  repeated Dependency dependencies = 1;
+}
+
+// ConfigScope defines the visibility of an Istio configuration artifact in
+// a namespace when the namespace is imported.  By default all
+// configuration artifacts are public. Configurations with private scope
+// will not be imported when the namespace containing the configuration is
+// imported in a NetworkScope.
+enum ConfigScope {
+  PUBLIC = 0;
```
mbanikazemi: This is essentially MESH rather than PUBLIC. Right? Does it make sense to use MESH and leave PUBLIC or something along that line possibly for multi-mesh scenarios in the future?

networking/v1alpha3/network_scope.proto
```diff
@@ -0,0 +1,153 @@
...
+  message Dependency {
+    // REQUIRED: Import describes the set of namespaces whose exported
+    // services will be accessed by the workloads selected by the
+    // sourceWorkloadLabels. The sidecars attached to the workloads will be
+    // configured with information required to reach other services in the
+    // same namespace and the imported services. In addition to the
+    // explicitly specified namespaces, namespaces specified in the global
+    // mesh config (through defaultNetworkScope.importNamespaces) will also be
+    // imported.
+    repeated Import imports = 2;
```
mbanikazemi: I know that using 2 as the field number is correct but it seems we start from 1 in the rest of the code. We may want to change to 1.

networking/v1alpha3/network_scope.proto
rshriram: The config exposure is either local namespace or throughout the service registry. Embodying the concept of multi-mesh into Istio is a much bigger discussion and we already went through this in the doc (on exact same topic). I am simply translating the doc into proto. Including the use of NetworkScopes

mesh/v1alpha1/config.proto
```diff
@@ -199,8 +199,41 @@ message MeshConfig {
...
+      ALL_NAMESPACES = 0;
+
+      // Only configure routes to services that are in the same namespace
+      // as the workload as well as services in namespaces specified in
+      // importNamespaces.
+      SAME_NAMESPACE = 1;
+    }
+
+    // REQUIRED: The default import setting for every workload in the mesh.
+    Mode import_mode = 1;
```
linsun: Curious why the default isn't 0, ALL_NAMESPACES?

It stated in other place:   "By default, the service mesh
established by Istio will have a full mesh connectivity - i.e. every
workload will have proxy configuration required to reach every other
workload in the mesh."

frankbu: The 1 is a proto field number, not a value.

linsun: Ah, thanks Frank!

PR #720, opened 2018-11-29
rename NetworkScope to ServiceDependency

rshriram:


costinm:
Why ?

hzxuzhonghu:
Yes.  I talked with @rshriram on slack. He meant not to make user confused with k8s network policy.

linsun:
How about ServiceScope?  @rshriram @costinm @hzxuzhonghu @louiscryan 

service dependency feels a little odd to me.

costinm:
I think we agreed on network scope.

rshriram:
@costinm this was after talking to @louiscryan . you might want to sync with him on all the happenings in this area over the last couple of days

costinm:
So far we have clear consensus on moving forward with ns isolation - not on
actual syntax or naming for deps, or how it relates with the existing use
of ServiceEntry for defining  dependencies.

My proposal is to hold on this new CRD - and for 1.1 just reuse the
ServiceEntry mechanism to declare deps on external hosts. We may expand
Gateway as well, but at least not confuse things with a new resource that
overlaps with both until we have a good plan.

PR #725, opened 2018-12-05
Fix typo, 'target' is not known 

pilhuhn:
target -> targets

Supersedes bad pr https://github.com/istio/istio.io/pull/2946


pilhuhn:
Thank you @frankbu for guiding me. The `make proto-commit` had no (visible) effect. I guess because the definitions stayed the same and only the comment changed (?)

frankbu:
@pilhuhn That's right, you only need to run `make proto-comit` if you change the actual message structure.

hklai:
@pilhuhn/@frankbu master is actually locked. Please merge to release-1.1 before master is unlocked.

Also, perhaps it is worthwhile to cherry-pick this to release-1.1 as well?

PR #726, opened 2018-12-06
Add per service locality weight setting

hzxuzhonghu:
This is to support envoy [locality-weighted-load-balancing](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing#locality-weighted-load-balancing)


rshriram:
@hzxuzhonghu sorry I missed this. This looks good.

hzxuzhonghu:
@costinm  https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing/locality_weight

It will generate [Cluster.CommonLbConfig](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#cluster-commonlbconfig) for Envoy.

costinm:
That seems to be an empty proto, that indicates to envoy to use the load
info from EDS.

What I'm trying to figure out is how does this DestinationRule translate
into an envoy config that can
provide what the API is claiming. Is it going to affect the EDS results ?
But how ?

And secondary - what is the use case for user to specify "80% to local
zone, 20% to remote zone".
Never seen this use case - all configs I've seen want as much as possible
local, then fallback to
zone with extra capacity in order of latency.

costinm:
To be clear: we really want to implement locality based LB, and many thanks
for starting - I just
want to understand what is the plan, and if this is the right API.

costinm:
How is it different than existing split, with labels for zone/region ?

The envoy feature seems to be intended for real load balancing, not exact
split.

And how will this api interact with the real lb ?

Design doc or discussions before api change would help, in particular for
things with broad implications...

hzxuzhonghu:
>Locality weighted load balancing is configured by setting locality_weighted_lb_config in the cluster configuration and providing weights in LocalityLbEndpoints via load_balancing_weight.

The implementation is simple, 

>How is it different than existing split, with labels for zone/region ?

Actually, the existing split does not take effect I think.  We dont set config as envoy expected.

costinm:
On Tue, Jan 1, 2019 at 5:40 PM Zhonghu Xu <notifications@github.com> wrote:

> Locality weighted load balancing is configured by setting
> locality_weighted_lb_config in the cluster configuration and providing
> weights in LocalityLbEndpoints via load_balancing_weight.
>
> The implementation is simple,
>

Some details ? I don't see any simple option based on the envoy docs on
locality_weighted_lb_config that could achieve what the API seems to
promise.



> How is it different than existing split, with labels for zone/region ?
>
> Actually, the existing split does not take effect I think. We dont set
> config as envoy expected.
>

I meant: the current traffic split we do for subsets ( based on labels,
etc).  If the workloads are labeled to reflect zone/etc - why wouldn't
the destination rule split satisfy the requirement ? I assume it's because
you want different splits by source - is this something we
want for traffic split/DestinationRule in general ?

But my primary concern is to not confuse users - we still want real
locality-based load balancing, taking into account load, endpoints, etc.
And the envoy option doesn't seem to match the percent based split in the
API you are proposing.

hzxuzhonghu:
> If the workloads are labeled to reflect zone/etc - why wouldn't
the destination rule split satisfy the requirement ? I assume it's because
you want different splits by source - is this something we
want for traffic split/DestinationRule in general ?

That's right. As you said DestinationRule can not achieve sourced routing. We may have workloads reside in multi region/zones, and access workloads in other multi region/zones. We need to control traffic based on both region/zone and the workloads number within it.   

>Some details ?

Envoy docs requires setting `Cluster.CommonLbConfig.LocalityWeightedLbConfig`  and combined `endpoint.LocalityLbEndpoints.load_balancing_weight` . currently LocalityWeightedLbConfig is more like a flag(https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#cluster-commonlbconfig-localityweightedlbconfig), and `load_balancing_weight` has already been set in previous prs.

costinm:
On Tue, Jan 1, 2019 at 10:20 PM Zhonghu Xu <notifications@github.com> wrote:

> If the workloads are labeled to reflect zone/etc - why wouldn't
> the destination rule split satisfy the requirement ? I assume it's because
> you want different splits by source - is this something we
> want for traffic split/DestinationRule in general ?
>
> That's right. As you said DestinationRule can not achieve sourced routing.
> We may have workloads reside in multi region/zones, and access workloads in
> other multi region/zones. We need to control traffic based on both
> region/zone and the workloads number within it.
>
Is there a doc describing the use cases, requirements - and how the
implementation will work ?

I'm as confused as before - and I suspect other devs are also not familiar
with this, and I don't remember any discussion.

Maybe a cleaner solution is to extend DestinationRule to allow source-based
config. What you described so far doesn't sound like
load balancing, but traffic split. In general "load balancing" implies
"load" will be a factor.



> Some details ?
>
> Envoy docs requires setting
> Cluster.CommonLbConfig.LocalityWeightedLbConfig and combined
> endpoint.LocalityLbEndpoints.load_balancing_weight . currently
> LocalityWeightedLbConfig is more like a flag(
> https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#cluster-commonlbconfig-localityweightedlbconfig),
> and load_balancing_weight has already been set in previous prs.
>

Without a document or discussion on the list or in WG - I don't think many
people know about them. We need to improve the review process to make sure
all PRs have clear indication if it is a bug fix or adds a new feature -
and in the later case link to some doc. We can't require all developers to
review all PRs.
Even if I saw the PRs - without comments in the PR it's hard to understand
what the final picture will be and that they were related to this feature.
I've
seen a bunch of PRs related to zero vpn, using load weight to adjust the
gateway - but they can't be used in the general case, and we can't have a
feature
like this require zvpn and gateways. As I mentioned, 'load balancing' is a
very complicated subject and the API in this PR doesn't fit any pattern or
use I've
seen - so docs please...

linsun:
how does envoy side car know the region and zone info within istio env?   I don't recall setting up instruction for this.

also, does this new traffic policy allow user to configure this most common user case @costinm outlined earlier?  I can set to 100% to local region/zone but how do I configure the fallback?

`all configs I've seen want as much as possible local, then fallback to zone with extra capacity in order of latency`

rshriram:
Service entry has locality field. And you can add region/az annotation to k8s nodes. We have been parsing these values for two years. Not using them though. Basically no user config. Just make sure cluster nodes are annotated per standard kubernetes docs. 

In terms of the fallback, the functionality exists in a convoluted way in Envoy and works only if active health checking or outlier detection is enabled or retry with priority levels are enabled. This needs more work in Envoy. Alternatively you can write your own automation to change the values in the destination rule during such incidents.

linsun:
thank you @rshriram for the clarification!!

PR #727, opened 2018-12-06
Add fields into ingress gateway.

JimmyCYJ:
Adds two fields **source** and **name** into gateway tls config, to allow provisioning k8s secrets to ingress gateway via SDS.

Design doc:
https://docs.google.com/document/d/1agtVQADCk32mLwot8iygVGhHcP1Lna7G1Cne4Zwllgg/edit#


JimmyCYJ:
cc @myidpt

rshriram:
this is a crude hack to enable SDs in the gateway without conveying the intent to end user.

costinm:
+1 on Shriram's comment.

Does it even need config? Envoy has the name of the host from SNI - SDS has
all certs and the names in the san - just need to match them

myidpt:
@rshriram 
https://docs.google.com/document/d/1agtVQADCk32mLwot8iygVGhHcP1Lna7G1Cne4Zwllgg/edit?ts=5c09dbc8#heading=h.qex63c29z2to

myidpt:
We should keep the "source", and have an enum of "SourceOptions" including "K8S_SECRET and FILE".
We can remove "name" if it can be automatically generated through the host.

rshriram:
Source is not indicative of using a Cert fetching agent in the background. And definitely not one that is amenable to people adding their own SDS agents either.

if you want an API driven way of doing this, then 
1. The API should explicitly specify that the certs are fetched dynamically from another server
2. Provide a way for configuring the address of this server
3. Provide a way to specify an opaque resource name by which the certs for a downstream cluster can be obtained.

something like this:
 ```
message TLSCredStore {
 string address; // ip/uds
 string credentialName;
}

// If credentialStore is specified, serverCertificates, privateKey and CACertificates will be obtained via an API call to the credential store.
TLSCredStore credential_store;
```

This would allow the end user to specify the address of the SDS server. If we ship one by default in the IngressGateway, we can document it in the API (say uds://istio-local-cred-store).

myidpt:
I think we can adopt the schema suggested by @rshriram, except that the "address" is optional, with default value "/var/run/sds/uds_path".

JimmyCYJ:
> I think we can adopt the schema suggested by @rshriram, except that the "address" is optional, with default value "/var/run/sds/uds_path".

@rshriram and @myidpt 
I have updated the PR, would you please take a look? Thanks.

costinm:
If the cert can be determined and provisioned automatically - I don't know
why user has to specify it in config. We have enough configs.

The name in SNI identifies the cert - all you need is parse the cert in sds
and match.

costinm:
In general - we want to simplify the setup, we can automate provisioning
and user should not have to worry or care if he is running with SDS or not.
It's a deployment/install detail, not something users should hardcode in
config.

costinm:
For 1.1 and Ingress Gateway (which is really the only place this option is needed ) I suggest either following the cluster install option to detect if SDS is enabled, or using the ADS metadata to detect if 
a workload has SDS. No need for user explicit config in Gateway.

The name of the credential can either be the SNI hostname (best - since the SDS server can match against certs), or use the filename as a key, which will also help migration. Right now users manage TLS certs in Secrets that get mounted in the ingress gateway - when SDS is deployed, the SDS server 
could handle the same Secrets without expensive migration effort.

rshriram:
The problem here is that Pilot needs to know whether to generate SDS config or non SDS config for a gateway or sidecar. And in a heterogeneous deployment, some may use SDS while others may use standard file mounts. So, we need some level of configurability in the API.

While we can send a use_sds or dont_use_sds flag via teh metadata, we also need to send the URI of the SDS server so that Pilot can configure this properly. WE cannot infer it or hard code it. Secondly, the cluster level thing still needs to come in as a runtime option that pilot code can check so that it can enable SDS or not. And this has to be portable across platforms. Having an API driven way of configuring the SDS usage helps in this regard. Ideally, any place TLS is being specified should have a one of.. but we are too late for that..

This doesn't have to apply to the sidecar resource though.

costinm:
Also: we can (and should) automate the Ingress cert provisioning
completely. User should not have to specify even the file paths in the
ideal case.
We have all infrastructure in place for this - with either Certmanager or
to plug in other tools.

If the user has its own certificate - we need to support the case where
they run on ingress on a VM without any k8s, only files.

For SDS, the SDS UDS ( or maybe HTTPS) address is an install option for
each cluster, not a per-gateway config that the user
needs to worry about. And the 'key' used to retrieve the secret can be
either derived from the SNI/host, or simply use the current filename.
An SDS server could watch the Secret (in running on k8s) or watch the files
(on raw VM), and avoid the need to restart envoy, which
is the problem we need to solve. No need for user to make config changes or
learn about SDS paths just to avoid restarting envoy.
Config is already too complicated.

We can both infer and hardcode the SDS default path.

rshriram:
I agree that specifying the UDS path is clunky. But you still need a way to specify if that gateway server should be using using SDS or not. We can then infer the SDS settings from some global mesh config (the SDS uds path) and infer the resource name from the hosts. 

A cluster level gateway option to specify SDS or no-SDS is not a good idea. One or more servers in a gateway could be fetching their certs dynamically while others may be using pre-mounted cert files (and okay with the reload issue as their certs may not be changing). We cannot force every gateway to use a particular thing based on some global option. What we can do is have a standard setting for the SDS stuff (path/https) in some global config.

costinm:
Sds - like CNI, autoinjection - are install options.
User should not have to write configs that are dependent of
install/environment - an app and it's configs need to be portable.

I would suggest relaxing the tls cert file to use relative paths, and allow
the relative path to be used as a key in SDS.

Note that the SDS server may be remote - we certainly need values.yaml and
install options.
What I'm strongly against is leaking environment options into the user
configs.

costinm:
To make a java analogy - Gateway and VirtualService, etc are similar to
web.xml, while values.yaml matches the implementation specific settings for
each servlet container.

PR #729, opened 2018-12-11
add enable_sds_k8s_sa_jwt flag to meshconfig

quanjielin:

This flag is used by secret discovery service(SDS).
 If set to true, envoy will fetch normal k8s service account JWT(this is fallback from k8s trustworthy jwt, which is only available for k8s 1.12 or higher) from '/var/run/secrets/kubernetes.io/serviceaccount/token'
(https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod)and pass to sds server, which will be used to generate key/cert eventually.

This isn't supported for non-k8s case.


mesh/v1alpha1/config.proto
```diff
@@ -190,10 +190,19 @@ message MeshConfig {
   // $hide_from_docs
   // This flag is used by secret discovery service(SDS). 
   // If set to true(prerequisite: https://kubernetes.io/docs/concepts/storage/volumes/#projected), Istio will inject volumes mount 
-  // for k8s service account JWT, so that K8s API server mounts k8s service account JWT to envoy container, which 
-  // will be used to generate key/cert eventually. This isn't supported for non-k8s case.
+  // for k8s service account trustworthy JWT(which is avaialbe for k8s 1.12 or higher), so that K8s API server 
+  // mounts k8s service account JWT to envoy container, which will be used to generate key/cert eventually. 
+  // This isn't supported for non-k8s case.
   bool enable_sds_token_mount = 23;
```
diemtvu: Are these 2 flags independent?

quanjielin: rename and update comment based on offline chat, PTAL, thanks!

mesh/v1alpha1/config.proto
```diff
@@ -190,10 +190,20 @@ message MeshConfig {
...
+  // mounts k8s service account trustworthy JWT to envoy container, which will be used to request key/cert eventually. 
+  // This isn't supported for non-k8s case.
   bool enable_sds_token_mount = 23;
 
+  // $hide_from_docs
+  // This flag is used by secret discovery service(SDS). 
+  // If set to true, envoy will fetch normal k8s service account JWT from '/var/run/secrets/kubernetes.io/serviceaccount/token' 
+  // (https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod) 
+  // and pass to sds server, which will be used to request key/cert eventually. 
+  // If both enable_sds_token_mount and sds_use_k8s_sa_jwt are set to true, enable_sds_token_mount(trustworthy jwt) takes precedence.
```
diemtvu: Or simply say `this flag is ignored if enable_sds_token_mount is set.`

PR #738, opened 2018-12-13
Add global flag to enable ingress gateway agent

JimmyCYJ:
Add a new flag enable_ingress_gateway_agent. If this flag is set to true, an ingress gateway agent will be deployed in the same pod as ingress gateway proxy.

User could define TLS ingress gateway config as below. The new gateway agent watches kubernetes secrets httpbin.example.com in namespace istio-ingress, and provisions this credential to gateway via SDS.

cat <<EOF | kubectl apply -f -
kind: Gateway
...
    tls:
      mode: SIMPLE
    hosts:
    - "httpbin.example.com"
EOF

Design doc: https://docs.google.com/document/d/1agtVQADCk32mLwot8iygVGhHcP1Lna7G1Cne4Zwllgg/edit#


mesh/v1alpha1/config.proto
```diff
@@ -204,6 +204,10 @@ message MeshConfig {
   // This isn't supported for non-k8s case.
   bool sds_use_k8s_sa_jwt = 30;
 
+  // Deploy ingress gateway agent, which runs in the same pod as ingress gateway proxy and provisions credentials to ingress
+  // gateway proxy via SDS. The ingress gateway agent watches kubernetes secrets in namespace istio-ingress.
+  bool enable_ingress_gateway_agent = 31;
```
costinm: Confused - is this something that will dynamically change ? It's an install option - values.yaml and a flag or environment on pilot ( plus whatever helm is needed )

JimmyCYJ:
This is not the correct place to change. Just discussed with @quanjielin , and I am going to add environment variable and change values.yaml.

PR #741, opened 2018-12-21
Introduce forward looking MCP enhancements

ayj:
The initial version of Mesh Configuration Protocol (MCP) was
introduced to decouple Pilot/Mixer from the k8s kube-apiserver. These
enhancements address additional forward looking requirements as we
bring Galley and the MCP API to beta quality.

* Enable alternative control topologies where the source of
  configuration is not publicly accessible.

* Provide a feedback mechanism to report the observed config state to
  the user (e.g via CRD status).

* Improve performance at scale (e.g. Enterprise use case)

* Rationalize the resource model

The intent is to introduce these backwards incompatible API changes
now before Galley and MCP ship as beta quality and on-by-default.

Rationalization of the resource model and incremental improvements can
be implemented immediately as they effect the contract between Pilot
and Galley. Alternative control topologies and feedback/status are
inherently new features and can be implemented later with the same
APIs without concerns of breaking compatibility.

Design proposal: https://goo.gl/RTKMwF


ayj:
cc @jeffmendoza @Nino-K

mcp/v1alpha1/resource.proto
```diff
@@ -24,12 +24,12 @@ import "mcp/v1alpha1/metadata.proto";
...
-// Envelope for a configuration resource as transferred via the Mesh Configuration Protocol.
-// Each envelope is made up of common metadata, and a type-specific resource payload.
-message Envelope {
+// Resource for a resource as transferred via the Mesh Configuration Protocol. Each
+// resource is made up of common metadata, and a type-specific resource payload.
+message Resource {
     // Common metadata describing the resource.
     istio.mcp.v1alpha1.Metadata metadata = 1;
 
     // The resource itself.
```
geeknoid: How about instead saying "The primary payload for the resource".

mcp/v1alpha1/resource.proto
```diff
@@ -24,12 +24,12 @@ import "mcp/v1alpha1/metadata.proto";
 option go_package="istio.io/api/mcp/v1alpha1";
 option (gogoproto.equal_all) = true;
 
-// Envelope for a configuration resource as transferred via the Mesh Configuration Protocol.
-// Each envelope is made up of common metadata, and a type-specific resource payload.
-message Envelope {
+// Resource for a resource as transferred via the Mesh Configuration Protocol. Each
```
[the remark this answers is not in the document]
ayj: Thanks - I thought I caught all of these search/replace errors.

mcp/v1alpha1/mcp.proto
```diff
@@ -184,3 +184,109 @@ service AggregatedMeshConfigService {
...
+// in the Resources. ACK/NACK is determined by the presence
+// of error_detail.
+message RequestResources {
+  // An opaque identifier and generic set of labels to identify the MCP sink node.
+  Node node = 1;
+
+  // Type of resource collection that is being requested, e.g.
+  //
+  // istio/networking/v1alpha3/VirtualService
+  // k8s/<apiVersion>/<kind>
```
rshriram: ?? what is the k8s/apiVersion stuff for? Arent we converting k8s services to ServiceEntries and streaming them instead?

rshriram:
Left some top level comments in the doc.

mcp/v1alpha1/metadata.proto
```diff
@@ -18,22 +18,43 @@ package istio.mcp.v1alpha1;
...
+  // resource updates. It should be treated as opaque by consumers/sinks.
   string version = 3;
+
+  // Map of string keys and values that can be used to organize and categorize
+  // resources within a collection.
+  map<string,string> labels = 4;
+
+  // Structured data that can be used by source and sink to communicate arbitrary
+  // metadata about this resource.
+  google.protobuf.Struct annotations = 5;
```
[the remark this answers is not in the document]
costinm: Please don't use Structs. Super expensive and verbose and wire-inefficient. 
There may be a use case ( like representing json or xml ) - but this is not it.

mcp/v1alpha1/mcp.proto
```diff
@@ -184,3 +184,109 @@ service AggregatedMeshConfigService {
...
+///////////////////////////////////////////////////////////////////////////
+
+// A RequestResource can be sent in two situations:
+//
+// Initial message in an MCP bidirectional change stream
+// As an ACK or NACK response to a previous Resources. In
+// this case the response_nonce is set to the nonce value
+// in the Resources. ACK/NACK is determined by the presence
+// of error_detail.
+message RequestResources {
```
costinm: How about ResourceRequest ? That seems to be a more common pattern for naming a request.

mcp/v1alpha1/mcp.proto
costinm: I think the protocol can be used to watch/stream any k8s ( or consul, etc ) object.

mcp/v1alpha1/mcp.proto
```diff
@@ -184,3 +184,109 @@ service AggregatedMeshConfigService {
...
+  // When the RequestResources is the first in a stream, the initial_resource_versions must
+  // be populated. Otherwise, initial_resource_versions must be omitted. The keys are the
+  // resources names of the MCP resources known to the MCP client. The values in the map
+  // are the associated resource level version info.
+  map<string, string> initial_resource_versions = 3;
+
+  // When the RequestResources is an ACK or NACK message in response to a previous RequestResources,
+  // the response_nonce must be the nonce in the RequestResources. Otherwise response_nonce must
+  // be omitted.
+  string response_nonce = 4;
```
costinm: One of the biggest pain points in ADS is this 'nonce' and error_detail.  It is never clear when a Request is a real request, and when it is just an ack/nack.

Keep in mind rpc.Status can also be 'OK' - so the field should probably be named status.

I would just make things clear - the request should include either an ack (status=OK), nack ( status != ok, response_nonce set) or an update/new request.

mcp/v1alpha1/mcp.proto
ayj: I would normally agree if this was only ever sent by the client. However, in the reverse service direction `ResourceRequest` is sent by the server.

mcp/v1alpha1/mcp.proto
ayj: The protocol is more general than the current implementation. Pilot would continue to watch collections of Istio resources directly. e.g.

```
istio/networking/v1alpha3/VirtualService
istio/networking/v1alpha3/ServiceEntry
// etc ...
istio/authentication/v1alpha1/Policy
// etc ...
```

The k8s/apiVersion stuff could be used to move k8s resources between networks/clusters where direct access to the k8s apiserver is not possible.

mcp/v1alpha1/metadata.proto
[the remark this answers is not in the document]
ayj: ... or we stick with `map<string,string>` which covers the majority of use cases we want to encourage (e.g. simple annotations). Arbitrary complex annotations would still be possible (via JSON), but they would be discouraged in favor of adding well defined field(s) in the common metadata or resource specific proto.

mcp/v1alpha1/mcp.proto
ayj: The presence of `response_nonce` should be sufficient to differentiate ACK/NACK from updated watch, no? 

- ACK (nonce!=""; error_details==nil)
- NACK (nonce!=""; error_details!=nil) 
- New/Update request (nonce=="", error_details ignored)

Note that this is slightly different than the non-incremental spec where request/response are always paired in per-version nonce _and_ version.

mcp/v1alpha1/metadata.proto
```diff
@@ -18,22 +18,43 @@ package istio.mcp.v1alpha1;
...
-  // type is identified by the TypeUrl of the resource field of the
-  // Envelope message.
+  // Fully qualified name of the resource. Unique in context of a collection.
+  //
+  // The FQ name consists of path segments which includes the resources
+  // location in a resource hierarchy. On k8s, this hierarchy is
+  // two-levels: cluster and namespace, e.g.
+  //
+  //   namespaced     => <k8s-namespace>/<k8s-name>
+  //   cluster-scoped =>                /<k8s-name>  # alt: reserved namespace (e.g. :cluster, :global)
```
geeknoid: What does the # alt: comment mean?

mcp/v1alpha1/metadata.proto
```diff
@@ -18,22 +18,43 @@ package istio.mcp.v1alpha1;
...
+  //
+  // Deeper hierarchy could also be represented, e.g.
+  //
+  //        <org>/<team>/<name>
+  //
+  // “/” should be used as the standard delimiter. The rightmost segment is
```
geeknoid: should be used as the standard delimiter -> are used as the segment delimiter.

mcp/v1alpha1/metadata.proto
```diff
@@ -18,22 +18,43 @@ package istio.mcp.v1alpha1;
...
+  // the most specific identifier (e.g. leaf). Segments moving towards the left
+  // represent higher positions in the resource hierarchy.
   string name = 1;
 
   // The creation timestamp of the resource.
   google.protobuf.Timestamp create_time = 2;
```
geeknoid: This really represents the creation time of the specific version of the resource, correct? Perhaps change this to version_create_time and update the comment accordingly?

mcp/v1alpha1/mcp.proto
```diff
@@ -19,21 +19,21 @@ package istio.mcp.v1alpha1;
...
-// presented to the management server, which may use this identifier
-// to distinguish per client configuration for serving. This
-// information is not authoriative. Authoritative identity should come
+// Identifies a specific MCP sink node instance. The node identifier is
+// presented to the resource source, which may use this identifier
+// to distinguish per sink configuration for serving. This
+// information is not authoritative. Authoritative identity should come
 // from the underlying transport layer (e.g. rpc credentials).
-message Client {
+message Node {
```
geeknoid: I don't like the choice of "Node" in this context as it is unrelated to Kubernetes Nodes. Maybe SinkNode?

mcp/v1alpha1/mcp.proto
```diff
@@ -19,21 +19,21 @@ package istio.mcp.v1alpha1;
...
   // An opaque identifier for the MCP client.
   string id = 1;
 
-  // Opaque metadata extending the client identifier.
+  // Opaque metadata extending the node identifier.
   google.protobuf.Struct metadata = 2;
```
geeknoid: This is unrelated to what's defined in metadata.proto, right? Then perhaps this should use a different name than metadata.

How about annotations, and use a map<string, string> just like metadata annotations?

mcp/v1alpha1/mcp.proto
```diff
@@ -184,3 +184,109 @@ service AggregatedMeshConfigService {
       returns (stream IncrementalMeshConfigResponse) {
   }
 }
+
+///////////////////////////////////////////////////////////////////////////
```
geeknoid: I think we should nuke this comment...

mcp/v1alpha1/mcp.proto
geeknoid: Perhaps add this cheat-sheet to the comments?

mcp/v1alpha1/mcp.proto
```diff
@@ -184,3 +184,109 @@ service AggregatedMeshConfigService {
...
+  // are the associated resource level version info.
+  map<string, string> initial_resource_versions = 3;
+
+  // When the RequestResources is an ACK or NACK message in response to a previous RequestResources,
+  // the response_nonce must be the nonce in the RequestResources. Otherwise response_nonce must
+  // be omitted.
+  string response_nonce = 4;
+
+  // This is populated when the previously received resources could not be applied
+  // The *message* field in *error_details* provides the source internal exception
```
geeknoid: "exception" -> error

mcp/v1alpha1/metadata.proto
ayj: This was an alternative way to represent cluster-scoped resources using a special reserved namespace.

mcp/v1alpha1/metadata.proto
ayj: This represents the initial creation time of the resource. I think you're asking for `update_time` per https://cloud.google.com/apis/design/standard_fields. Unfortunately, k8s resources don't provide this information, only create/delete time.

mcp/v1alpha1/mcp.proto
ayj: This is carried over from Envoy's xDS protocol which also uses node. In their case node=envoy instance.

mcp/v1alpha1/mcp.proto
ayj: Metadata is a common concept (e.g. https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md). We end up with metadata at different layers, e.g.

- connection/stream metadata (grpc)
- node metadata (e.g. pod labels)
- resource metadata (e.g. k8s resource metadata)

Note that this was also carried over from the envoy/xDS protocol.

geeknoid: I'm looking at this in the context of the MCP API proper, and reusing the same term for two different meanings within the same API feel confusing to me. I realize that the notion of metadata is generally known in the industry, but this specific API has a type called Metadata and it has nothing to do with this field, and so it would be preferable to chose a different name.

mcp/v1alpha1/mcp.proto
geeknoid: At this point, MCP is pretty different than xDS so I think we should make decisions that make sense for MCP.

The primary place where MCP is used will be in Kubernetes environments that use Nodes as a very core concept. And when I need this name here, I immediately wondered whether this is the name of the k8s node that is requesting config. It's clearly not, but the name 'node' suggests that.

mcp/v1alpha1/metadata.proto
geeknoid: Too bad.

mcp/v1alpha1/mcp.proto
ayj: The name is context sensitive, i.e. node metadata vs. resource metadata. It isn't ideal but I'm lacking a good alternative. I'm not sure annotation is any better since annotation appears under metadata elsewhere.

costinm:
On Wed, Jan 2, 2019 at 12:19 PM Jason Young <notifications@github.com>
wrote:

> *@ayj* commented on this pull request.
>
> In mcp/v1alpha1/mcp.proto
> <https://github.com/istio/api/pull/741#discussion_r244842967>:
>
> > +  // Type of resource collection that is being requested, e.g.
> +  //
> +  // istio/networking/v1alpha3/VirtualService
> +  // k8s/<apiVersion>/<kind>
> +  string collection = 2;
> +
> +  // When the RequestResources is the first in a stream, the initial_resource_versions must
> +  // be populated. Otherwise, initial_resource_versions must be omitted. The keys are the
> +  // resources names of the MCP resources known to the MCP client. The values in the map
> +  // are the associated resource level version info.
> +  map<string, string> initial_resource_versions = 3;
> +
> +  // When the RequestResources is an ACK or NACK message in response to a previous RequestResources,
> +  // the response_nonce must be the nonce in the RequestResources. Otherwise response_nonce must
> +  // be omitted.
> +  string response_nonce = 4;
>
> The presence of response_nonce should be sufficient to differentiate
> ACK/NACK from updated watch, no?
>
>    - ACK (nonce!=""; error_details==nil)
>    - NACK (nonce!=""; error_details!=nil)
>    - New/Update request (nonce=="", error_details ignored)
>
>
It's not that clear - for example on reconnect you may have an ACK/NACK
from a previous message ( which may
have went to a different server ).

And you can never know if the message has additional resources to watch -
we need to keep doing diffs instead of
having a clear separating between updates and ACK/NACK.

So far I haven't seen any use case or benefit of mixing ACK/NACK with an
update of the watched resources ( and
the forced diff and confusion).

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.pb.go linguist-generated=true
```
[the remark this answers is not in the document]
ayj: The generated python protobuf code is already hidden by default.

https://github.com/github/linguist/blob/e761f9b013e5b61161481fcb898b59721ee40e3d/lib/linguist/generated.rb#L302
https://github.com/github/linguist/blob/e761f9b013e5b61161481fcb898b59721ee40e3d/lib/linguist/generated.rb#L302

PR #742, opened 2018-12-27
Sidecar config resource

rshriram:
Based on discussions and past networking meetings around 
https://docs.google.com/document/d/1jmH8N0OdyJb1W6hB6xdLZXzdGUKs6GknI5Ayb6wD430

Renamed and revamped ServiceDependency and made some gateway edits as well.


mesh/v1alpha1/config.proto
```diff
@@ -209,38 +209,7 @@ message MeshConfig {
...
-    repeated string import_namespaces = 2;
-  }
-
-  // The default service dependency setting associated with every workload
-  // in the mesh.  Pilot will program the routes in the sidecars and
-  // gateways accordingly. If omitted, sidecars will be configured to reach
-  // every service in the mesh. The default scope can be overriden by
-  // supplying a ServiceDependency resource per namespace.
-  DefaultServiceDependency default_service_dependency = 29;
+  reserved 29;
```
costinm: Since it never shipped - we can skip this.

networking/v1alpha3/gateway.proto
```diff
@@ -350,3 +359,23 @@ message Port {
   // Label assigned to the port.
   string name = 3;
 }
+
+// WorkloadSelector specifies the criteria used to determine if the Gateway
+// or Sidecar resource can be applied to a given workload. If multiple
+// conditions are specified, all conditions need to match in order for the
+// workload to be selected.
+message WorkloadSelector {
```
costinm: I don't remember this discussion - and using the IP is clearly dangerous in many ways. Please keep the old selector.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+  // configuration will be applied to all workloads in the current config
+  // namespace.
+  WorkloadSelector workload_selector = 1;
+
+  // Ingress specifies the configuration of the sidecar for processing
+  // inbound traffic to the attached workload. If omitted, Istio will
+  // autoconfigure the sidecar based on the information about the workload
+  // obtained from the service registry (e.g., exposed ports, services,
+  // etc.).
+  repeated IstioListener ingress = 2;
```
costinm: Should we include some indication of what is implemented and what is not ?

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+  // mesh. If omitted, Istio will autoconfigure the sidecar to be able to
+  // reach every service in the mesh.  etc.).
+  repeated IstioListener egress = 3;
+}
+
+// IstioListener specifies the properties of a single listener on the
+// sidecar proxy attached to a workload.
+message IstioListener {
+  // An arbitrary name associated with the listener used for emitting metrics.
+  string name = 1;
```
costinm: Required ? Optional with default value ? ( I think later is better )

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+
+  // The ip:port or the unix domain socket to which the listener should be
+  // bound to. Format: tcp://x.x.x.x:yyyy or unix:///path/to/uds or
+  // unix://@foobar (Linux abstract namespace).  To bind to any IP with
+  // specific port, use tcp://0.0.0.0:<port>
+  string bind_address = 2;
```
costinm: This is technically correct - but far too verbose ? I understand in advanced cases we'll have unix:// - but 99% of users just want a simple port. 

We detect full URI syntax - can we allow the simple IP:port and :port syntax for the common case ?

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+  // When the bind address is an IP:port, the bindToPort option dictates
+  // whether or not the sidecar should bind its listener socket to the
+  // specified port. Set bindToPort to false (default) if application
+  // traffic entering/leaving a pod/VM is captured automatically through
+  // iptables redirection and forwarded to the sidecar on a specific port
+  // (see proxyListenPort in the global MeshConfig). When not using
+  // iptables for traffic capture, set bindToPort to true to force the
+  // sidecar to bind to the specified port. Note that the binding might
+  // fail if the application workload is already bound to the same port.
+  bool bind_to_port = 3;
```
costinm: Can you remove or hide this ? It is not very clear how it interacts with environment setting ( i.e. different workloads may use different interception modes ). We can add it in 1.2 - we just started to implement iptables-less configs. 
I'm afraid this will commit us to an API that is both hard to understand and will not work well.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+
+  // The protocol associated with this listener.
+  // MUST BE one of HTTP|TCP|TLS|MONGO.
```
costinm: User may have envoy extensions to support more than MONGO - without having to rev Istio API.
How about allowing other names.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+  // DestinationRule, ServiceEntry, etc. for details.
+  repeated string hosts = 5;
+
+  // The IP endpoint or unix domain socket to which traffic should be
+  // forwarded to by default. In the context of an ingress server, this
+  // configuration can be used to redirect traffic arriving at the bind
+  // point on the sidecar to a port or unix domain socket where the
+  // application workload is listening for connections. Format should be
+  // tcp://127.0.0.1:PORT or unix:///path/to/socket
+  string default_endpoint = 6;
```
costinm: Same - would be easier to just also allow :port

networking/v1alpha3/gateway.proto
rshriram: See the doc. The selector should not be just labels. We need to be able to use stuff from X509 certs, or other stuff that proxies provide as a way of identification. So putting them in a container leaves room for future expansion (like adding subject name or some jwt stuff in the workloadSelector)

networking/v1alpha3/sidecar.proto
rshriram: ack

networking/v1alpha3/sidecar.proto
rshriram: This is simply determining if we will add this listener to the existing 15000 listener or create a standalone listener, irrespective of the interception mode (ip tables/tproxy). Its in fact very easy to implement and super useful as you can now define one or more listeners on a sidecar in pod/VM that are not directly related to a service port. Its especially useful in VM world because you can now choose to install listeners for a specific set of ports and allow the application to explicitly communicate with the sidecar over the specified port.

By default its false, i.e. it gets added to the 15000 listener. Whether or not we use the listener-level filter or the deprecated bind_to_port in envoy, the net effect is the same. 

Remember that this is an array of servers/listeners. And once we implement workload selector, users can apply this to a specific workload (a VM  in a namespace).

costinm:
In any case - the IP should not be in the selector. It is a clear major
problem in the current implementation.


Using stuff from the cert can't be used at selection time - since this info
is not known except at connect.

Let's remove this until we have a clear need for an expansion - and then we
can figure out if it needs to replace selector
or where to add the fields.

costinm:
It would be atrocious to ask users to configure individual VMs by IP or
know ahead of time how the configs will be deployed.

And the naming 'bind to port' is very confusing for a user - the doc
doesn't say anything that the port where it'll be bound is 15000 or
whatever else we'll do at implementation level to capture traffic.

I'm not saying it can't be implemented - just that it's a horrible model to
introduce the option to not bind. It should be part of the
node metadata - i.e. environment based - not something a user should
configure in Sidecar.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+// in the mesh, as well as accept traffic on all the ports associated with
+// the workload. The Sidecar resource provides a way to fine tune the set
+// of ports, protocols that the proxy will accept when forwarding traffic
+// to and from the workload. In addition, it is possible to restrict the
+// set of services that the proxy can reach when forwarding outbound
+// traffic from the workload.
+//
+// Services and configuration in a mesh are organized into one or more
+// namespaces (e.g., a Kubernetes namespace or a CF org/space). A Sidecar
+// resource in a namespace will apply to all workloads in the namespace.
```
hzxuzhonghu: Then this is not consistent with `WorkloadSelector `.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,180 @@
...
+// *_Currently, each namespace should have only one Sidecar resource named
+// "default"_*. The behavior of the system is undefined if more than one
```
hzxuzhonghu: This kind of fixed name (default) limitation is not user friendly like ` Policy MeshPolicy`. We'd better prevent such stuff.

networking/v1alpha3/sidecar.pb.go
```diff
@@ -0,0 +1,949 @@
...
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: Sidecar
+// metadata:
+//   name: default
+//   namespace: prod-us1
+// spec:
+//   ingress:
+//   - bind: tcp://10.10.10.10:9080
```
hzxuzhonghu: Does `tcp` here really necessary since have `protocol` below?

networking/v1alpha3/sidecar.proto
hzxuzhonghu: >This is simply determining if we will add this listener to the existing 15000 listener or create a standalone listener

Is there any doc about this, I am a little confused. And Iunderstand that all traffic will be intercepted by iptables and redirected to 15001?

networking/v1alpha3/sidecar.proto
rshriram: I will in next iteration, depending on what I can get done in the first cut for the PR in istio/istio

rshriram:
workloadSelector in the doc is in first example
```
 selector:
    labels:
      app: bookstore
```

in terms of the workloadSelector, right now we only have labels, but I have created room for this to grow once we have the ability to identify the workload by other information (from certs/etc.).

networking/v1alpha3/sidecar.proto
rshriram: I have hidden this for now.

rshriram:
cc @louiscryan

costinm:
Yes, this 'room to grow' is what I'm saying is not agreed on, we can't add
APIs without a clear use case and knowing where we'll go.

Label-based selection is what we had and I'm not aware of any plan to
deprecate and replace this.

We can't identify from certs except at connect time, it is not known from
discovery  - and even there the cert is just used to verify the info passed
in labels and the namespace. IP is clearly a dead end.

costinm:
Please revert the selector -> workloadSelector to avoid confusions. The
rest is fine.

We can't just rename fields and change structure without very strong
reasons, the API is supposed to be stable
and backward compatible. I understand you 'deprecated' selector - but there
is absolutely no reason to do this.

hzxuzhonghu:
Looks very good after this round. 

>We can't just rename fields and change structure without very strong
reasons, the API is supposed to be stable
and backward compatible

Agree,  unless one day selector can not satisfy users.

costinm:
I don't think deprecating the selector is 'the only way' for anything - we
can fix the 'all namespaces' with a global flag, or by detecting if they
use any Sidecar resource in the namespace,
or by detecting if the VirtualServices use qualified names for the gateway
( istio-system:ingress-gateway ). Last is probably the best option - since
it avoids most breakages.

Matching by x509 certs may be an interesting issue - but completely
orthogonal to what this PR is supposed to do and I don't remember any
discussion or
documents about this feature.

Deprecation is a major, siginificant change - and I don't think it should
be done without careful consideration, as a side-effect of another change.
If there is a use case for the x509 matching - we can discuss what are the
exact requirements and behavior and what's the cleanest way to
support it - and if it ends up requiring to deprecated selector, or to add
a separate field without deprecating selector - we can do it.

costinm:
You mean everything else will use 'selector' - and only Sidecar will use
'workloadSelector' ? Just to add some x509 attributes - which could as well
be added in a backward compatible way, by simply adding a x509 section ? I
don't think either is a good idea. I am not convinced x509 cert should
be used this way - it should certainly be used to validate namespace and
other labels, but certs are not the best way to pass information and
we also need to handle JWTs, which have different structure.

Most users will have a default Sidecar for the entire namespace -  and so
far the cert infra we support doesn't have any attributes.
When we have a design and requirements for other uses of certs elements -
we can discuss the API, I would be happy to see
more info in the certs - but with a clear design ( how are the extra
attributes in cert validated, how they fall-back to JWT, how it
impacts cert rotation, etc ).

Another related aspect is matching Envoy metadata ( which includes other
things besides labels ). But we can add this as
either a separate element, or by using a micro-format in the selector keys.

The reason I'm very strongly against this change is the deprecation and
inconsistency of a core API, selector is one of the most
used concepts in k8s and istio.

rshriram:
deprecating selector in favor of workloadSelector also provides an easy upgrade path for people using Gateways (if you use WorkloadSelector.labels, then it applies only to the gateways in same namespace). Also, remember we have non k8s use cases as well. 

As far as sidecar is concerned, whether we use selector or workload selector doesnt matter as both are new. On the gateway side, the workloadSelector is new but allows us to clearly say we are referring to labels..

Syntactically, 
```
workloadSelector:
  labels:
    foo: bar
````

is clearer than simply 

```
selector:
  foo: bar
```

Even if we decide to not introduce anything related to x509 or jwt, I fail to see why workloadSelector.labels is worse than simply using selector. The default case (sidecar resource per namespace) is not going to be affected either way. The specialized case (one that is yet to be implemented) is the one that requires *literally* just another level of indentation to add the word labels. 

Also in terms of backward compatibility, global flags are less useful. It does not make it known to the end user that the API's intent has changed. Its also not obvious. Also its a black and white change. You can either have namespace local gateway or not. Thats unlikely to be helpful in a large installation. People have declared a gateway and a virtual service in the same namespace, with the gateway referring to the workload in istio-system namespace. Take a look at the mailing lists to get an idea. Any hacky move to break this is going to create some very irate users. The flag also means that the default we will have is going to be to allow the current behavior. We will have no way of informing existing users that the behavior is going to change soon as the API fields are the same.

How about this: if you can spin up a PR that manages this transition in 100% backward compatible way: a default istio install that allows gateway resources only in same namespace as the gateway yet allowing all existing users to use existing gateway behavior (refer to workload in different namespace) -- I will be happy to get rid of this workloadSelector and go back to selector. I can guarantee this behavior with the API change above. But if you can guarantee the exact same behavior purely with code changes for all scenarios, then we are good to go with the old selector thing.

**Comment by costinm** [comment]
We use selector in the other apis, let's keep consistency.

And it is not even selecting a workload - but multiple workloads.

**Comment by costinm** [comment]
Can you explain again how is calling it workloadselector instead of
selector impact the migration ?

**Comment by costinm** [comment]
I mean: to get to the new behavior, users will need to explicitly change
VirtualService to include the namespace of the gateway.
Without this - their config would be broken.

**Comment by costinm** [comment]
On Tue, Jan 1, 2019 at 8:26 AM Shriram Rajagopalan <notifications@github.com>
wrote:

> As I said, this is the litmus test:
>
> a default istio install that allows gateway resources only in same namespace as the gateway yet allowing all existing users to use existing gateway behavior (refer to workload in different namespace)
>
> With your admin-flag-in-pilot model, the above is not possible. I can
> either run Istio in old mode or new mode. So my painful transition path
> will be, run default Istio install (which WILL HAVE TO SUPPORT OLD MODE
> else things will break), then change all gateway resources to same
> namespace as gateway workload, and fix all virtual services and then
> finally *reinstall* istio with the flag so that I get new behavior only
> -- all the while praying that I actually changed all the virtual services &
> gateways correctly. Any small mistake and everything is affected.
>
The problem with this transition: in the new world, there will be no
Gateway resource in namespaces that don't run gateway workloads.
They would have no purpose. You can only place Gateway resource in the
namespaces where Gateway workloads (and certs) are running.

So in istio-system (or namespaces using a-la-carte gateways) - no change is
really needed either way. In your proposal we still make a gratuitous
change,
but still same effect - the selection and behavior will be unchanged in
those namespaces.

In the 'app' namespace, where we run a VirtualService - in the new world
you can't have a Gateway, no matter how you rename the fields inside.
So the Gateway resource should be removed to keep things clean - but even
if it's left around, it'll select nothing and will have no effect once
'secure'
mode is enabled. And in transition mode - it can still preserve old
behavior.

There is no opportunity for 'mistake' in my proposal - we can easily report
any namespace still using the old-style Gateway, and after transition
is off - old Gateways can be removed ( but don't hurt in any way, since
they're not used and don't select anything).



> With the dual-selector model, I can provide the above behavior (the old
> selector field that has old behavior but its deprecated - we emit lots of
> warnings, and the new workloadSelector field that has the new behavior.).
> After one release cycle, we can remove the old selector field from the API
> or stop supporting it giving people time to transition, and TIME TO PLAY
> WITH NEW TYPE OF SELECTOR.
>
Not really - workloadSelector in a regular namespace can't select anything.
That is what we're trying to fix - Gateways in a non istio-system ( or
a-la-carte gateway ) namespace
selecting or touching anything in istio-system/gateway-namespace.

And in istio-system (or gateway-namespace) - it doesn't make any difference
if we keep the old selector, since what changes is restricting selection to
current
namespace.



> The security aspect of things is true only for multi-tenant k8s clusters
> with a team per namespace. There are also a lot of single tenant users (*including
> the team inside VMware, and juspay.in <http://juspay.in> folks*) -- one
> team, one cluster, bunch of services in different namespaces, one gateway,
> one cert, all managed by 5-6 people. Its not fair to break all their
> production installations by simply changing the semantics of the selector
> field in a backward incompatible way and not providing a smooth upgrade
> path [the flags as stated above will not provide the smooth upgrade path,
> but the dual selectors will provide smooth upgrade path].
>

In this case - there is no problem either way. If they run a gateway and
workload using gateway in same namespace - there is no bug and nothing to
fix.
If they run the gateway/cert in istio-system - the transition I'm proposing
for multi-tenant works as well, and whatever Gateways they keep in the
app namespace should still be deleted/useless after transition - with
workloadSelector or selector, it makes no difference.


Costin

>

**Comment by rosenhouse** [comment]
I'm not sure I'm following everything here, but for what it's worth, in the Cloud Foundry use-case, we will probably want to scope a `Sidecar` resource per "CF App", due to concerns about scaling.

We don't currently have the concept of labels in CF, and for our integration today our MCP server just makes up labels and puts them on the ServiceEntry before shipping to Pilot.

We do currently mint a unique cert per App container, which includes metadata about the CF App.  These certs are automatically rotated every N hours.  We plan to have the sidecar Envoy use this cert when connecting to Pilot.

So, we could use our pseudo-labels for Sidecar selection.  We could also take advantage of metadata in the x509 cert used by Envoy when connecting to the Pilot gRPC server.

I think we'd slightly prefer the latter, since it doesn't depend on these fake labels.  But no strong opinions.

**Comment by ZackButcher** [comment]
+1 on selectors other than labels - cert fields (or SPIFFE identities) are a clear use case for non-label selection. CIDR ranges are also a valuable selector (and supporting a single IP address falls out of supporting CIDR ranges).

**Comment by costinm** [comment]
Please start a doc on selectors other than labels - and how they'll work.
No problem with that - I would love to see cert info used, but
this is not the right way to introduce this change, sneaking it in an
unrelated PR. We agreed to add Sidecar, and we agreed to deprecate
Gateway selecting across namespaces.

CIDR ranges are also interesting - but again it's a complex subject that
needs to be properly reviewed.

And we can introduce additional ways to identify workloads without
deprecating 'selector' - they can be added in a backward compatible way.

The doc on adding Sidecar has nothing on deprecating 'selector' - and the
plan to use the rename as a convoluted way to drive the migration
makes no sense, even if someone finds an obscure use case for Gateways
still existing in namespaces that don't run Gateway workloads. It
can be done by adding an extra field to indicate the migration ( since we
still can't rev the API version ).

For 1.1 - the story on selecting by IP on non-flat namespace is still not
clear, and using the certs is even less clear ( fine for using the cert
on connect - but selector is also used to identify workloads that are not
directly connected ). My proposal was to start with using the
cert to validate the labels, and also to use it for the data plane
connections.

I can see a case for stopping the use of selectors in Gateway - for example
if the 1.2 operator takes care of starting the envoy gateway workloads
based on the Gateway config itself.

**Comment by rshriram** [comment]
https://docs.google.com/document/d/15p1PcYIUE18q5dbWJoYY5teyNgnnjBg73597-Tw6dzY is the doc restating what I noted earlier on the extensibility

File: mesh/v1alpha1/config.proto [hunk 12]
```diff
@@ -151,23 +151,27 @@ message MeshConfig {
...
-  // Set the default behavior of the sidecar for handling outbound traffic from the application.
-  // While the default mode should work out of the box, if your application uses one or more external services that
-  // are not known apriori, setting the policy to ALLOW_ANY will cause the sidecars to route traffic to the any
-  // requested destination.
-  // Users are strongly encouraged to use ServiceEntries to explicitly declare any external dependencies,
-  // instead of using allow_any.
+  // Set the default behavior of the sidecar for handling outbound traffic
+  // from the application.  If your application uses one or more external
+  // services that are not known apriori, setting the policy to ALLOW_ANY
+  // will cause the sidecars to route traffic to the any unknown traffic
```
louiscryan: typo

networking/v1alpha3/gateway.proto
louiscryan: Over time yes but do we have a strong use-case for doing this change in 1.1 ?

networking/v1alpha3/gateway.proto
```diff
@@ -227,10 +236,16 @@ message Gateway {
...
 message Server {
+  // An arbitrary name associated with this server. Used for emitting metrics.
+  string name = 6;
+
   // REQUIRED: The Port on which the proxy should listen for incoming
   // connections
   Port port = 1;
 
+  // The IP address to which the listener should be bound to.
+  string bind_address = 5;
```
louiscryan: Field should either be called 'bind' or 'address' but not both. My vote would be 'bind'

Doc :  "The IP address to which the listener should be bound"

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,199 @@
...
+// the workload. The Sidecar resource provides a way to fine tune the set
+// of ports, protocols that the proxy will accept when forwarding traffic
+// to and from the workload. In addition, it is possible to restrict the
+// set of services that the proxy can reach when forwarding outbound
+// traffic from the workload.
+//
+// Services and configuration in a mesh are organized into one or more
+// namespaces (e.g., a Kubernetes namespace or a CF org/space). A Sidecar
+// resource in a namespace will apply to all workloads in the namespace in
+// the absence of a workload selector.  *_Currently, each namespace should
```
louiscryan: Temporary limitation of this PR ?

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,199 @@
...
+// have only one Sidecar resource without any workload selector_*. The
+// behavior of the system is undefined if more than one Sidecar resource
+// exists in a given namespace.
+//
+// The example below delcares a Sidecar resource in the prod-us1 namespace
+// that configures the sidecar to proxy egress traffic for services in the
```
louiscryan: that configures the sidecar to allow egress traffic to services in the

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,199 @@
...
+  // The ip or the unix domain socket to which the listener should be bound
+  // to. Port MUST be specified if bindAddress is not empty. Format:
+  // x.x.x.x or unix:///path/to/uds or unix://@foobar (Linux abstract
+  // namespace). If omitted, Istio will autoconfigure the defaults based on
+  // imported services and the workload to which this configuration is
+  // applied to.
+  string bind_address = 3;
+
+  // $hide_from_docs
+  // When the bind address is an IP:port, the bindToPort option dictates
```
louiscryan: Description here conflicts with doc of bind_address which states that it's just IP and not IP:port.

Also not clear why this field is necessary ? This is for egress only no?

If so then most of the time the user would just not configure the port and be happy, or configure it but only to declare a protocol. That presence or absence is sufficient to indicate whether the user wants the sidecar to explicitly bind to the port, i.e. this boolean is redundant

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,199 @@
...
+  // Egress specifies the configuration of the sidecar for processing
+  // outbound traffic from the attached workload to other services in the
+  // mesh. If omitted, Istio will autoconfigure the sidecar to be able to
+  // reach every service in the mesh.  etc.).
+  repeated IstioListener egress = 3;
+}
+
+// IstioListener specifies the properties of a single listener on the
+// sidecar proxy attached to a workload.
+message IstioListener {
```
louiscryan: Likely should split into two types given that default_endpoint is not necessary for egress and 'hosts' are also somewhat redundant for ingress. Would also make the field documentation a little clearer.

networking/v1alpha3/sidecar.proto
louiscryan: I think the syntax here is fine although we can support :port too if need be.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,199 @@
...
+  // The IP endpoint or unix domain socket to which traffic should be
+  // forwarded to by default. In the context of an ingress server, this
+  // configuration can be used to redirect traffic arriving at the bind
+  // point on the sidecar to a port or unix domain socket where the
+  // application workload is listening for connections. Format should be
+  // 127.0.0.1:PORT or unix:///path/to/socket
+  string default_endpoint = 6;
+}
+
+// ConfigScope defines the visibility of an Istio configuration artifact in
```
louiscryan: Unused ? If so remove.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,199 @@
...
+  // inbound traffic to the attached workload. If omitted, Istio will
+  // autoconfigure the sidecar based on the information about the workload
+  // obtained from the service registry (e.g., exposed ports, services,
+  // etc.).
+  repeated IstioListener ingress = 2;
+
+  // Egress specifies the configuration of the sidecar for processing
+  // outbound traffic from the attached workload to other services in the
+  // mesh. If omitted, Istio will autoconfigure the sidecar to be able to
+  // reach every service in the mesh.  etc.).
```
louiscryan: every service visible to this namespace.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,199 @@
...
+  // sidecar configuration should be applied. If omitted, the sidecar
+  // configuration will be applied to all workloads in the current config
+  // namespace.
+  WorkloadSelector workload_selector = 1;
+
+  // $hide_from_docs
+  // Ingress specifies the configuration of the sidecar for processing
+  // inbound traffic to the attached workload. If omitted, Istio will
+  // autoconfigure the sidecar based on the information about the workload
+  // obtained from the service registry (e.g., exposed ports, services,
```
louiscryan: We don't really use the 'service registry' but rather what we know about the workload from the environment and it's orchestrator. Want to keep the implied joins to a minimum

networking/v1alpha3/sidecar.proto
rshriram: Oh this is used. carry over from the old service dependency.. public/private for services/virtual services/destination rules/service entries

networking/v1alpha3/gateway.proto
```diff
@@ -158,10 +158,10 @@ message Gateway {
...
 
   // REQUIRED: One or more labels that indicate a specific set of pods/VMs
-  // on which this gateway configuration should be applied.
-  // The scope of label search is platform dependent.
-  // On Kubernetes, for example, the scope includes pods running in
-  // all reachable namespaces.
+  // on which this gateway configuration should be applied. The scope of
+  // label search is restricted to the configuration namespace in which the
+  // the resource is present. In other words, the Gateway resource must
+  // reside in the same namespace as the gateway workload.
```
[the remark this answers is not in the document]
costinm: We'll need a 'migration' section - explaining that the behavior in 1.0 had a bug, and how to move away. 
But it can be in the release notes or separate doc - the text in this PR is correct.

networking/v1alpha3/gateway.proto
```diff
@@ -231,6 +231,9 @@ message Server {
   // connections
   Port port = 1;
 
+  // The IP address to which the listener should be bound.
```
costinm: Few more words / example / optional with 0.0.0.0:port as default ? 

Do we really want to restrict this to IP ? what about UDS ? I assume UDS is tricky due to the ports.

I wonder if 'bind' wouldn't be better inside port - since it allows more flexibility ( bind 443 on the external interface, and 
8443 on the internal interface ). Having a single bind means all ports are in the same boat.

networking/v1alpha3/gateway.proto
```diff
@@ -334,6 +337,14 @@ message Server {
...
   // https, and the TLS modes to use.
   TLSOptions tls = 3;
+
+  // The IP endpoint or unix domain socket to which traffic should be
+  // forwarded to by default. In the context of an ingress server, this
+  // configuration can be used to redirect traffic arriving at the bind
+  // point on the sidecar to a port or unix domain socket where the
+  // application workload is listening for connections. Format should be
+  // 127.0.0.1:PORT or unix:///path/to/socket
+  string default_endpoint = 5;
```
costinm: Should we use the same format for the 'bind' ? 

Do we allow default_endpoint to be a non-local IP or a DNS entry ? If not, would be good to be explicit ( and why ). 
If we do - an example.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,235 @@
...
+//     - "istio-system/*"
+// ```
+//
+message Sidecar {
+  // $hide_from_docs
+  // Criteria used to select the specific set of pods/VMs on which this
+  // sidecar configuration should be applied. If omitted, the sidecar
+  // configuration will be applied to all workloads in the current config
+  // namespace.
+  WorkloadSelector workload_selector = 1;
```
costinm: Let's introduce workload_selector consistently, for all configs - it's confusing to have Sidecar use one style 
and the rest use a different one, and the discussion on how to use non-label selection is still not resolved.

costinm: I see it is hidden - so maybe it's ok to leave it there if in 1.1 we just implement the default Sidecar.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,235 @@
...
+  // reach every service in the mesh that is visible to this namespace.
+  repeated IstioListener egress = 3;
+}
+
+// IstioListener specifies the properties of a single listener on the
+// sidecar proxy attached to a workload.
+message IstioListener {
+  // An optional arbitrary name associated with the listener used for
+  // emitting metrics.
+  string name = 1;
```
costinm: Required ? If not - doc the default value ? ( and I think it should be optional with default value, to reduce user confusion)

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,235 @@
...
+  Port port = 2;
+
+  // $hide_from_docs
+  // The ip or the unix domain socket to which the listener should be bound
+  // to. Port MUST be specified if bind is not empty. Format:
+  // x.x.x.x or unix:///path/to/uds or unix://@foobar (Linux abstract
+  // namespace). If omitted, Istio will autoconfigure the defaults based on
+  // imported services and the workload to which this configuration is
+  // applied to.
+  string bind = 3;
```
costinm: Same comments as in Gateway - do we want bind to apply to all ports, or each port to be able to bind differently ?

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,235 @@
...
+    // expected to explicitly communicate with the listener port/unix
+    // domain socket. When used in ingress listener, care needs to be taken
+    // to ensure that the listener port is not in use by other processes on
+    // the host.
+    NONE = 2;
+  }
+
+  // When the bind address is an IP, the captureMode option dictates
+  // how traffic to the listener is expected to be captured (or not).
+  CaptureMode capture_mode = 4;
```
costinm: For inbound traffic with 'none' -  we need to indicate which port to forward to, can't be just the enum.
And like bind, it is one app port per listening port. 

How about extending the Port proto to indicate the local app port - for example 
- port: 443
  localPort: 8443

and if the localPort is specified, it implies NONE capture mode ( for that port )

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,235 @@
...
+  // DestinationRule, ServiceEntry, etc. for details.
+  repeated string hosts = 5;
+
+  // The IP endpoint or unix domain socket to which traffic should be
+  // forwarded to by default. In the context of an ingress server, this
+  // configuration can be used to redirect traffic arriving at the bind
+  // point on the sidecar to a port or unix domain socket where the
+  // application workload is listening for connections. Format should be
+  // 127.0.0.1:PORT or unix:///path/to/socket
+  string default_endpoint = 6;
```
costinm: Same comment as in gateway - but likely more important for Sidecar. Each incoming port needs an endpoint.

networking/v1alpha3/sidecar.proto
rshriram: there is a one to one mapping from bind to port. Its a repeated field (port+bind in same message)

networking/v1alpha3/sidecar.proto
rshriram: defaultEndpoint allows you to specify 127.0.0.1:PORT - which achieves the forwarding behavior. And as I stated above, its already one bind per port and one defaultEndpoint per port. So, we have a 1-1-1 mapping already

networking/v1alpha3/sidecar.proto
louiscryan: Agreed, should be latter.

networking/v1alpha3/sidecar.proto
```diff
@@ -0,0 +1,226 @@
...
+  // (see proxyListenPort in the global MeshConfig). When not using
+  // iptables for traffic capture, set bindToPort to true to force the
+  // sidecar to bind to the specified port. Note that the binding might
+  // fail if the application workload is already bound to the same port.
+  bool bind_to_port = 4;
+
+  // One or more services/virtualServices exposed by the listener in
+  // namespace/dnsName format.  _*Hosts will be ignored for ingress
+  // servers*_. For egress servers, the hosts field results in importing
+  // one or more publicly scoped services and VirtualServices from remote
```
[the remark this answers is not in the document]
louiscryan: Then the VirtualService would have to be defined in the same namespace as the private service

networking/v1alpha3/sidecar.proto
louiscryan: Should be a top-level resource then. Where is this used in this PR or just adding now for uture use ?

rshriram: This was the only leftover from the old service dependency proto. Its already being used by all configs (Service entries, virtual servifes, dest rules, etc.). and implemented in Pilot.

PR #744, opened 2018-12-31
remove deprecated from OutboundTrafficPolicy. REGISTRY_ONLY and ALLOW_ALL

vadimeisenbergibm:
related to https://github.com/istio/istio/issues/10330


mesh/v1alpha1/config.proto
```diff
@@ -242,8 +242,14 @@ message MeshConfig {
   // supplying a ServiceDependency resource per namespace.
   DefaultServiceDependency default_service_dependency = 29;
 
+  // The flag instructs Pilot to allow traffic to ports for which no
+  // Services/Service Entries exist. In particular, it will allow traffic
+  // to any external service for which no internal/external Service/
+  // Service Entry exist.
+  bool allow_traffic_to_unknown_ports = 31;
```
rshriram: I am fine with this option (I know I suggested this :) ). That said, I also realized we have something similar and unused currently (https://github.com/istio/api/blob/d1d3a0275a29d1a0e6bac076d67d3cc59a7d0a57/mesh/v1alpha1/config.proto#L152). Should we reuse that (keep the virtual_service_only as deprecated, but remove the deprecated for other two enum values, and the deprecated for the outbound_traffic_policy field). One less option to add to global mesh config.

vadimeisenbergibm: Sounds good, let me do it.

vadimeisenbergibm:
@rshriram How about this change, note the changed comment. Could you please approve it again?

rshriram:
Per discussion with @louiscryan , we decided to not do any global flag for the sidecar resource. People have to specify the resource for each namespace.

mesh/v1alpha1/config.proto
```diff
@@ -145,17 +145,17 @@ message MeshConfig {
...
   string mixer_address = 16 [deprecated=true];
-  
+
   message OutboundTrafficPolicy {
     enum Mode {
       // outbound traffic will be restricted to services defined in the service registry as well as those defined
       // through ServiceEntries
-      REGISTRY_ONLY = 0 [deprecated=true];
-      // outbound traffic to unknown destinations will be allowed
-      ALLOW_ANY = 1 [deprecated=true];
```
rshriram: remove deprecated from the mode field below and the outbound_traffic_policy field after that.

vadimeisenbergibm: Done.

PR #748, opened 2019-01-04
Update RBAC for Authorization v2 API.

yangminzhu:

### DIFF
--- a/rbac/v1alpha1/rbac.proto
+++ b/rbac/v1alpha1/rbac.proto
@@ -1,4 +1,4 @@
-// Copyright 2018 Istio Authors
+// Copyright 2019 Istio Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -84,9 +84,42 @@ package istio.rbac.v1alpha1;
 
 option go_package="istio.io/api/rbac/v1alpha1";
 
+// $hide_from_docs
+// This is forked from the networking/v1alpha3/sidecar.proto to avoid a direct
+// dependency from the rbac API on networking API.
+// TODO: Move the WorkloadSelector to a common place to be shared by other packages.
+// WorkloadSelector specifies the criteria used to determine if the Gateway
+// or Sidecar resource can be applied to a proxy. The matching criteria
+// includes the metadata associated with a proxy, workload info such as
+// labels attached to the pod/VM, or any other info that the proxy provides
+// to Istio during the initial handshake. If multiple conditions are
+// specified, all conditions need to match in order for the workload to be
+// selected. Currently, only label based selection mechanism is supported.
+message WorkloadSelector {
+  // One or more labels that indicate a specific set of pods/VMs on which
+  // this sidecar configuration should be applied. The scope of label
+  // search is restricted to the configuration namespace in which the the
+  // resource is present.
+  map<string, string> labels = 1;
+}
+
+// $hide_from_docs
+// AuthorizationPolicy to enforce access control on a selected workload.
+message AuthorizationPolicy {
+  // $hide_from_docs
+  // Optional. One or more labels that indicate a specific set of pods/VMs
+  // on which this authorization policy should be applied. Leave this empty to
+  // select all pods/VMs.
+  // The scope of label search is platform dependent. On Kubernetes, for example,
+  // the scope includes pods running in the same namespace as the authorization policy itself.
+  WorkloadSelector workload_selector = 1;
+
+  // $hide_from_docs
+  // A list of bindings that specify the subjects and permissions to the selected workload.
+  repeated ServiceRoleBinding allow = 2;
+}
+
 // ServiceRole specification contains a list of access rules (permissions).
-// This represent the "Spec" part of the ServiceRole object. The name and namespace
-// of the ServiceRole is specified in "metadata" section of the ServiceRole object.
 message ServiceRole {
   // Required. The set of access rules (permissions) that the role has.
   repeated AccessRule rules = 1;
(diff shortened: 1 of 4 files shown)


rbac/v1alpha1/rbac.proto
```diff
@@ -18,89 +18,138 @@ syntax = "proto3";
...
+}
+
+// AuthorizationPolicy to enforce access control on a selected workload.
+message AuthorizationPolicy {
+  // Optional. One or more labels that indicate a specific set of pods/VMs
+  // on which this authorization policy should be applied. Leave this empty to
+  // select all pods/VMs.
+  // The scope of label search is platform dependent. On Kubernetes, for example,
+  // the scope includes pods running in the same namespace as the authorization policy itself.
+  Selector selector = 1;
```
rshriram: Since you guys are defining this freshly, can you please use the sidecar.WorkloadSelector for this? it has room for expansion and will benefit all APIs.

rbac/v1alpha1/rbac.proto
```diff
@@ -18,89 +18,138 @@ syntax = "proto3";
...
   // Exact match, prefix match, and suffix match are supported for service names.
   // For example, the service name "bookstore.mtv.cluster.local" matches
   // "bookstore.mtv.cluster.local" (exact match), or "bookstore*" (prefix match),
   // or "*.mtv.cluster.local" (suffix match).
   // If set to ["*"], it refers to all services in the namespace.
-  repeated string services = 1;
+  // Deprecated. Use selector in authorization policy instead.
+  repeated string services = 1 [deprecated = true];
+
+  // Optional. A list of HTTP hosts. If not specified, it applies to any host.
```
rshriram: why HTTP ? Just state that its a list of hosts corresponding to services in the registry.
Also, do you allow wildcard hosts like Pilot does? (*.foo.com)?

Secondly, do you think its useful to adopt the same convention as Pilot's sidecar format which has configNamespace/Hosts ? This allows people to import stuff in bulk, such as ns1/*, ns2/foo.com, etc. Note that the config namespace has nothing to do with k8s namespace.

rbac/v1alpha1/rbac.proto
```diff
@@ -18,89 +18,138 @@ syntax = "proto3";
...
+  repeated string hosts = 5;
+
+  // Optional. A list of HTTP hosts that must not be matched.
+  repeated string not_hosts = 6;
```
rshriram: Do you need this not? isn't it sufficient to specify hosts alone?

rshriram: hmm I can see some use for this. But you should change the comments to state that you can either have hosts or notHosts, but not both.

rbac/v1alpha1/rbac.proto
```diff
@@ -112,11 +161,24 @@ message AccessRule {
   // If not specified, it applies to any path.
   repeated string paths = 2;
 
+  // Optional. A list of HTTP paths or gRPC methods that must not be matched.
+  repeated string not_paths = 7;
+
```
rshriram: same as above. oneOf semantics

rbac/v1alpha1/rbac.proto
```diff
@@ -112,11 +161,24 @@ message AccessRule {
...
 
+  // Optional. A list of HTTP methods that must not be matched.
+  repeated string not_methods = 8;
+
+  // Optional. A list of port numbers of the request. If not specified, it applies
+  // to any port number.
+  repeated int32 ports = 9;
+
+  // Optional.  A list of port numbers that must not be matched.
+  repeated int32 not_ports = 10;
```
rshriram: one of semantics must be specified in the comment.

rbac/v1alpha1/rbac.proto
```diff
@@ -170,15 +232,48 @@ message ServiceRoleBinding {
 // The supported keys in `properties` are listed in "constraint and properties" page.
 message Subject {
   // Optional. The user name/ID that the subject represents.
-  string user = 1;
+  // Deprecated. Use users and not_users instead.
+  string user = 1 [deprecated = true];
+
+  // Optional. A list of users that the subject represents. This is matched to the
+  // `source.principal` attribute. If not specified, it applies to any users.
+  repeated string users = 4;
```
rshriram: since you are adding a new field, why do you even want to call this as User (its confusing given that most people tend to think of users as people, e.g., liminw@google.com, and not as machines/source principals). I suggest renaming this as sourcePrincipals and notSourcePrincipals.

rbac/v1alpha1/rbac.proto
```diff
@@ -170,15 +232,48 @@ message ServiceRoleBinding {
...
+
+  // Optional. A list of groups that must not be matched.
+  repeated string not_groups = 7;
+
+  // Optional. A list of namespaces that the subject represents. This is matched to
+  // the `source.namespace` attribute. If not specified, it applies to any namespaces.
+  repeated string namespaces = 8;
+
+  // Optional. A list of namespaces that must not be matched.
+  repeated string not_namespaces = 9;
```
rshriram: You dont need namespace here because the hosts section would specify it (ns1/*)

rbac/v1alpha1/rbac.proto
yangminzhu: By HTTP we're trying to make it clear it's the HOST header in a HTTP request. The deprecated `services` field refer to services in registry but we're deprecating it in favor of the workload label selectors. cc @liminw 

Under this condition (it's the HOST header in a HTTP request) we could support the wildcard hosts easily.

Secondly, for now I think we're not going to support the configNamespace/Hosts syntax but would revisit this later.

rbac/v1alpha1/rbac.proto
yangminzhu: We're treating the `host` field as the HOST field in a HTTP request, so we still need the `namespace` here. Also note the `namespace` field is in the `Subject` message which is for the identity while the `host` field is in the `AccessRule` message which is for the permission.

rbac/v1alpha1/rbac.proto
yangminzhu: We're not using `one_of` for this (and other fields) because we allow to use them at then same time. Like the following one (Allows all sub domains of `abc.com` except `admin.abc.com`):
```yaml
hosts: ["*.abc.com"]
notHosts: ["admin.abc.com"]
```

rbac/v1alpha1/rbac.proto
yangminzhu: The problem is that we cannot use `repeated` inside `one_of`, so we have to wrap this in a message if we use `one_of`. Instead, I want to avoid wrapping this in a message but update the comments, and also check this in the validation stage.

rbac/v1alpha1/rbac.proto
yangminzhu: Same reply with the `paths`  field.

yangminzhu:
@rshriram
> Also, the widespread use of use and notUse is going to make the API and the spec a bit clumsy.

It may looks a bit clumsy in the API, but in reality, I think it's fine as most people won't use all these fields at the same time. cc @liminw 

For example, the following yaml actually doesn't look too bad.
```yaml
rules:
- notMethods: ["POST"]
  paths: ["/info"]
- methods: ["GET"]
  constraints:
  - key: "destination.labels[version]"
    value: ["test"]
```

For more background about the design, also see the [design doc](https://docs.google.com/document/d/1diwa9oYVwmLtarXb-kfWPq-Ul3O2Ic3b8Y21Z3R8DzQ/edit?usp=sharing_eip&ts=5bfda1cd&summarizationDetails=%5Bnull,false,%5B0,false,null,null,null,null,null,null,null,null,213,0%5D,null,null,6,1%5D)

yangminzhu:
@rshriram 
I updated the PR to include the v2 changes but hide them from the documentation. I will not do the deprecation in this PR so that we can merge this into the 1.1 branch.

rbac/v1alpha1/rbac.proto
```diff
@@ -84,9 +86,23 @@ package istio.rbac.v1alpha1;
...
+// $hide_from_docs
+// AuthorizationPolicy to enforce access control on a selected workload.
+message AuthorizationPolicy {
+  // $hide_from_docs
+  // Optional. One or more labels that indicate a specific set of pods/VMs
+  // on which this authorization policy should be applied. Leave this empty to
+  // select all pods/VMs.
+  // The scope of label search is platform dependent. On Kubernetes, for example,
+  // the scope includes pods running in the same namespace as the authorization policy itself.
+  istio.networking.v1alpha3.WorkloadSelector workload_selector = 1;
```
liminw: Can we duplicate WorkloadSelector definition here to avoid creating dependency on networking API? IMO, if workload selector is a common concept used by all Istio APIs, it should eventually be moved to a common package so that all APIs can import.

yangminzhu: Yes, I agree with you. @diemtvu also mentioned the same thing. Updated to fork the the `WorkloadSelector`

cc @rshriram

PR #755, opened 2019-01-11
Add option to select worload using lables for authn policy.

diemtvu:
This is a prep work to use label selector for authentication. Selector using service name will be converted to labels internally.


yangminzhu:
Should this go to 1.1 branch or are you planning to merge this after master is unblocked?

authentication/v1alpha1/policy.proto
```diff
@@ -360,6 +360,12 @@ message TargetSelector {
   // fully qualified domain name will be resolved in a platform specific manner.
   string name = 1;
 
+  // $hide_from_docs
+  // $not-implemented-hide
+  // Select workload by labels.
+  // Once implemented, this is a prefer way over using service name.
+  map<string, string> labels = 3;
```
rshriram: can we use the workloadSelector from networking please? RBAC has that as well. Since you are adding a new field, we might as well align on it. If changes are needed, its easy to do it in one place.

diemtvu: I plan to do it with v2 authn. Currently, with the current api, this is already behind a target selector so it's a bit weird.

Also, I feel we should refactor to move the WorkloadSelector to a 'common' package before spread the use of it. It doesn't make sense, and also causing dependency issues, to include networking APIs just for WorkloadSelector.

rshriram: I agree..

diemtvu:
@hklai, can we have this merge to master to have some implementation started. Thanks.

hklai:
@diemtvu perhaps not until it is unlocked. we don't want significant divergence between 1.1 and master.

PR #756, opened 2019-01-11
mixer: allow setting empty header if necessary

kyessenov:
Adding an operation to set a header to an empty value.
See https://github.com/istio/istio/issues/10801


linsun:
what would be the scenario to set empty header?

geeknoid:
@kyessenov Do we still need this PR?

geeknoid:
Please reopen if this is still needed.

PR #758, opened 2019-01-15
Replace 'scope' with 'export_to' namespace

louiscryan:
Currently 'scope' is only implemented on service and virtual service but not destination rule. This change also aligns with the revised naming conventions for referring to namespaces that have been adopted in the Sidecar API.


networking/v1alpha3/service_entry.proto
```diff
@@ -429,10 +428,13 @@ message ServiceEntry {
...
-  // where the service entry was added. The default scope is public,
-  // i.e. the service added by the service entry will be visible to
-  // workloads in the entire mesh.
-  ConfigScope config_scope = 7;
+  // The namespaces to which this service is exported. By default
+  // services are exported to all namespaces. This behavior can
+  // be overridden by explicitly listing the namespaces that are allowed
+  // to refer to this declaration.
+  //
+  // The value "." is reserved and defines an export to the namespace where
```
costinm: Let's clearly define the defaults - we have the same problem for Sidecar.

- if field is missing ( no entries ) - we clearly have to support the backward compat mode ( export to all )
- if it is present - it must have at least one entry  ( we can't have an empty list represented in proto). I assume "." will be equivalent with the former "private".
- not clear what is the use case for '*' - I assume it makes sense only if we define a mesh config indicating 'default export_to', allowing transition of the default from '*' to '.' (or some other explicit list).

networking/v1alpha3/service_entry.proto
```diff
@@ -429,10 +428,13 @@ message ServiceEntry {
...
+  // the service is declared, similarly the value "*" is reserved and
+  // defines an export to all namespaces.
+  repeated string export_to = 7;
```
costinm: export

costinm: Also worth noting that export_to still require an explicit 'egress' in the destination namespace.

networking/v1alpha3/virtual_service.proto
```diff
@@ -177,11 +176,15 @@ message VirtualService {
...
-  ConfigScope config_scope = 6;
+  // The namespaces to which this virtual service is exported. By default
+  // virtual services are exported to all namespaces. This behavior can
+  // be overridden by explicitly listing the namespaces that are allowed
+  // to refer to this declaration.
+  //
+  // The value "." is reserved and defines an export to the same namespace where
+  // this virtual service is declared, similarly the value "*" is reserved and
+  // declares an export to all namespaces.
+  repeated string export_to = 6;
```
costinm: We also need some annotation for Service - which is treated as a ServiceEntry but doesn't require one.

networking/v1alpha3/service_entry.proto
frankbu: Why? I thought no egress mean all public (i.e., with export_to my namespace) services in any other namespace.

costinm:
That behavior is not scalable - users can explicitly enable it with */*,
but it's not something we want to encourage.
If a Sidecar is used - we want to default to explicit imports ( and if
they are ok with slow performance or have low scale - can use */* ).

We need to stop all the 'all/random namespaces' behaviors - and Sidecar is
the way to opt in, at some point we'll have a global flag and
figure out how to completely deprecate the old behavior. For 1.1 - no
Sidecar means the old way for backward compat, Sidecar means
user understands they need explicit import and isolation.

networking/v1alpha3/service_entry.proto
louiscryan: Tidied up the language a bit to match that. I kept asterisk as a reserved value even if the use case is equivalent to the empty list case.

networking/v1alpha3/service_entry.proto
louiscryan: If no Sidecar is defined and we have the old behavior the export_to can still restrict it. Its an additive feature of the API so doing that does not impose any backward compatability risks

networking/v1alpha3/virtual_service.proto
louiscryan: Added doc for the

networking.istio.io/configScope

annotation on K8S Service

costinm: exportTo ? (for consistency)

louiscryan: Sorry, thats what I did. See definition in ServiceEntry. Don't need to do anything for VirtService

linsun:
what is the rational of choosing export_to over just export?

mesh/v1alpha1/config.proto
```diff
@@ -217,8 +217,50 @@ message MeshConfig {
...
+  // make the default 'Current namespace' so that services are only visible
+  // within their own namespaces by default. Operators can then expand the
+  // visibility of services to other namespaces as needed. Use of 'No Namespace'
+  // is expected to be rare but can have utility for deployments where
+  // dependency management needs to be precise even within the scope of a single
+  // namespace.
+  //
+  // For further discussion see the reference documentation for ServiceEntry,
+  // Sidecar, and Gateway.
+  string default_service_export_to = 31;
```
rshriram: since this is a config map, can we have the same repeated string here as well? or can we make others (in service entry/vs/dr) a comma separated list, just like this?

rshriram:
Also missing is the root namespace flag.

louiscryan:
> what is the rational of choosing export_to over just export?

If we just said 'export' it would read more like a boolean, 'export to' implies a specification of the receiver.

louiscryan:
> Also missing is the root namespace flag.

This can't be in config.proto as we'll read config.proto from that namespace. It has to be an install time flag for the control plane.

louiscryan:
> > Also missing is the root namespace flag.
> 
> This can't be in config.proto as we'll read config.proto from that namespace. It has to be an install time flag for the control plane.

I should say that there are two choices here. Either the control plane reads config.proto from it's own namespace or from the root namespace. I'm fine with either and it might be that reading config.proto from the control-plane namespace makes the most sense

louiscryan:
@rshriram PTAL

mesh/v1alpha1/config.proto
```diff
@@ -238,70 +295,3 @@ message ConfigSource {
   // mode as ISTIO_MUTUAL.
   istio.networking.v1alpha3.TLSSettings tls_settings = 2;
 }
-
-
```
rshriram: ? I was asking if you would be able to clean up the language in this locality weighted routing stuff. This is the code moved from the destination rule and shoved into the global config. I can clean it up later as well.

louiscryan: ah, ok. Updating

louiscryan:
PTAL

mesh/v1alpha1/config.proto
```diff
@@ -295,3 +298,98 @@ message ConfigSource {
...
+// {region}/{zone}/{sub-zone} form. For additional detail refer to
+// https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing/locality_weight
+// The following example shows how to setup locality weights mesh-wide.
+//
+// Given a mesh with workloads and their service deployed to "us-west/zone1/*"
+// and "us-west/zone2/*". This example specifies that when traffic accessing a
+// service originates from workloads in "us-west/zone1/*", 80% of the traffic
+// will be sent to endpoints in "us-west/zone1/*", i.e the same zone, and the
+// remaining 20% will go to endpoints in "us-west/zone2/*". This setup is
+// intended to favor routing traffic to endpoint in the same locality. A similar
```
frankbu: endpoints

mesh/v1alpha1/config.proto
hzxuzhonghu: Thanks

mesh/v1alpha1/config.proto
```diff
@@ -221,7 +218,70 @@ message MeshConfig {
...
+  //
+  // If not set the system will use "*" as the default value which implies that
+  // destination rules are exported to all namespaces
+  repeated string default_destination_rule_export_to = 33;
+
+  // $hide_from_docs
+  // The namespace to treat as the administrative root namespace for
+  // istio configuration. When processing a namespace Istio will treat
+  // declarations in the root namespace as if they belonged to each leaf
+  // namespace.
```
andraxylia: Unless there is overriding declaration in the leaf namespace?

networking/v1alpha3/service_entry.proto
```diff
@@ -135,15 +134,23 @@ option go_package = "istio.io/api/networking/v1alpha3";
 //
 // The following example demonstrates the use of a dedicated egress gateway
 // through which all external service traffic is forwarded.
+// The 'exportTo' field allows for control over the visibility of a service
```
andraxylia: export_to for consistency

andraxylia: `	// The 'export_to' field allows for control over the visibility of a service`

networking/v1alpha3/destination_rule.proto
```diff
@@ -119,11 +118,11 @@ message DestinationRule {
...
-  // The visibility setting associated with this DestinationRule. Set to
-  // PRIVATE if this destination rule should not be exported, i.e. restrict
-  // the applicability of this destination rule to only workloads in the same
-  // namespace as the destination rule.
-  ConfigScope config_scope = 4;
+  // The resolution of a DestinationRule to apply to a service occurs in the
+  // context of a hierarchy of namespaces. This rule control whether those
+  // namespaces are allowed to select this rule.
+  //
+  repeated string exportTo = 4;
```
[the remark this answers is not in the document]
andraxylia:
```suggestion
  repeated string export_to = 4;
```

networking/v1alpha3/virtual_service.proto
```diff
@@ -177,11 +176,20 @@ message VirtualService {
...
+  // and mesh administrators to control the visibility of virtual services
+  // across namespace boundaries.
+  //
+  // If no namespaces are specified then the virtual service is exported to all
+  // namespaces by default.
+  //
+  // The value "." is reserved and defines an export to the same namespace that
+  // the virtual service is declared in, similarly the value "*" is reserved and
+  // defines an export to all namespaces.
+  //
```
andraxylia: Add ~ for none.

proto.lock
```diff
@@ -2655,8 +2678,9 @@
               },
               {
                 "id": 4,
-                "name": "config_scope",
-                "type": "ConfigScope"
+                "name": "exportTo",
```
andraxylia:
```suggestion
                "name": "export_to",
```

mesh/v1alpha1/config.proto
louiscryan: done

mesh/v1alpha1/config.proto
louiscryan: Done

networking/v1alpha3/virtual_service.proto
louiscryan: Its not clear that 'none' has any utility on VirtualService or ServiceEntry beyond what 'this namespace only' provides. 

This is different from the use of 'none' on the referring side

PR #760, opened 2019-01-15
Locality LB failover api

hzxuzhonghu:
the failover can be implemented in Envoy by adding local locality endpoints in priotity 0, and failover locality endpoints in priority 1.

actually, whenever we specify locality routing with failover, have to specify outlier detection. because outlier detection will detect when local endpoints have failed and  proxy traffic to priority 1.

By default, if no failover or distribute specified, The default failover policy is:
if endpoints in same subZone fails, we failover to all endpoints in same zone. if zone fails, we failover to all endpoints in same region. if region fails, we look at the destination rule to see if there is a specific failover region. If no failover region is specified, then we failover to all endpoints in all regions.


networking/v1alpha3/destination_rule.proto
```diff
@@ -352,34 +373,54 @@ message LoadBalancerSettings {
...
+    };
+
+    // Specify the traffic failover policy.
+    // As zone and sub_zone failover is supported by default, only region can be specified here.
+    message Failover{
+        // Originating region.
+        string from = 1;
+
+        // Destination region the traffic will fail over to when endpoints in local region becomes unhealthy.
+        // Support wildcard '*' matching.
```
rshriram: what wildcard? you dont need a wildcard here right?

hzxuzhonghu: That means all other regions. Equals to not set.

networking/v1alpha3/destination_rule.proto
```diff
@@ -352,34 +373,53 @@ message LoadBalancerSettings {
...
     ConsistentHashLB consistent_hash = 2;
   }
 
-  // Explicitly assign loadbalancing weight across different zones and geographical locations.
-  // Refer to [Locality weighted load balancing](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing.html?highlight=load_balancing_weight#locality-weighted-load-balancing)
-  // If empty, the locality weight is set according to the endpoints number within it.
-  // If duplicated settings are present, then the first one will take effect.
-  repeated LocalityWeightSetting locality_weight_settings = 3;
+  // Locality load balancing settings.
+  message LocalitySetting{
```
rshriram: can we move this entire thing into mesh/v1alpha1/config.proto

hzxuzhonghu: You mean a mesh-wide config instead of per service?

rshriram: Yep. mesh wide.. Its because this API is very new and we dont know what customers want. So lets keep it in mesh/v1alpha1/config.proto -- so that we can change it if need be

networking/v1alpha3/destination_rule.proto
```diff
@@ -279,15 +279,36 @@ message Subset {
...
+//  metadata:
+//    name: bookinfo-ratings
+//  spec:
+//    host: ratings.prod.svc.cluster.local
+//    trafficPolicy:
+//      loadBalancer:
+//        localitySettings:
+//          failover:
+//            - from: region1
+//              to: region2
```
linsun: is it possible to configure failover after X number of retries?

networking/v1alpha3/destination_rule.proto
louiscryan: +1 lets move this to mesh config

hzxuzhonghu: done

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: No, this is lb stuff, not related to retry. On converse, retry may base on lb policy to choose which new host to try. 
(https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing/priority), plan to implement this by setting different priorities.

louiscryan:
Updated

PR #761, opened 2019-01-15
[WIP] Move WorkloadSelector to a common place and reuse it in both sidecar and rbac.

yangminzhu:
@rshriram The `WorkloadSelector` is already used in the Istio repo but I suppose it should be okay to refactor it to a common directory in this PR?

@geeknoid Previously you have concerns (https://github.com/istio/istio/issues/7997) about creating a common directory for common proto shared by multiple Istio API, could you let us know if you still think so for this common directory for `WorkloadSelector`? Thanks


costinm:
Also I think we should still have a review for 1.1, if using WorkloadSelector is the best option ( since it's inconsistent with the existing APIs, and few people are not comfortable with the naming) .

liminw:
Let's hold on this refactor until we have a broader consensus on the naming and mechanism. @smawson is driving the effort of having a common base selector for Istio policies.

PR #764, opened 2019-01-17
istio/istio#8772 add kube types, informers, clients, listers for istio crd types

rcernich:
Tool to generate kube types from istio types for istio/istio#8772

Need to add documentation and integrate it into the build.

Signed-off-by: rcernich <rcernich@redhat.com>


ozevren:
Hey rob,

I don't think the API repo is the right place for this. tools is probably the better place to check this in. 
FWIW, I've started a separate pack for this problem, quite some time ago: https://github.com/istio/tools/pull/30.

ozevren:
Also, is there a sample of the generated code?

rcernich:
This should be ready to go.  Please review with any comments.  It's mostly automated, with the exception of having to generate some custom DeepCopyInto() functions for the Istio types (simple enough and saves having to generate deepcopy for all the istio types).

ozevren:
@rcernich Please target the tools repo, not api.

rcernich:
Okay.  I've moved the code generation tool to istio/tools and created [this PR](istio/tools#37).  This PR now only contains updates to the build (to run code generators), tags added to istio types (to drive the generators), and the generated code.  These are in three separate commits to aid in reviewing the changes.  Please let me know if there's anything else you'd like to see changed.

rcernich:
I'm working on using k8s go-to-protobuf tool to generate the pb.go files for the k8s api types.  Unfortunately, the effort has devolved into updating the import paths used by the existing .proto files.  That said, I'm just about finished with that bit of refactoring.

I'm probably going to need help testing the updates to the build, to ensure I didn't break anything.

rcernich:
Okay, I think this is ready to go.  The kube/apis package will need to be regenerated if any new CRD types are added, removed, or if there are changes to any of the `+kubetype-gen` tags in the source types.  I've added targets to the `Makefile`, which are invoked via [generate-kube-apis.sh](./scripts/generate-kube-apis.sh).

In order to make this work, I had to modify the protobuf imports, which should require a new build of the `protoc` image.  I believe this is why the builds are failing.

rcernich:
I think, with a new build image, the build will now succeed.

rcernich:
@ozevren, what is the process for updating the builder images?  A new protoc image needs to be used, as the import paths for the *.proto imports have changed (they are now full package paths, as opposed to relative, with mappings back to the packages).

geeknoid:
To regenerate, you go to tools/protoc and run the build-and-push script.
This will regenerate a new docker image and upload it. You then need to
change references in the Makefile to point to the new image. The image is
named with the date it was generated on.

I'm not sure what permissions are necessary to run the script. If it's
failing for you when it tries to push up the image, let me know and I can
run it for you.

rcernich:
This should be ready to go now.  I'll need somebody to rebuild and push the protoc and all images, using the Dockerfiles in this commit.  After that, I can update the PR to point to the new all image, and the tests should pass.

The protoc image now includes all the code generation binaries.  For simplicity, I've refactored the build to just use the all image.  Theoretically, all the images could be reduced to a single Dockerfile that includes all the tools (which is what the all image is).  This might make it a little simpler to maintain going forward.

I've also added an env flag to allow you to run the k8s code generation without using the image.  Just define `LOCAL_KUBE_BUILD` before running `scripts/generate-kube-apis.sh`, and everything will execute locally.  (Note, you'll need to have `protoc` in your path.)  If you run locally, it will `go build` the generated code to ensure it compiles.  (Note, you'll get failures in some of the googleapis packages, because the versions are not consistent between it and gogo protobuf.  Specifically, we need to use an older version to preserve the current format of the generated pb.go files.  At some point, the versions should be updated to point to release versions instead of the commits that they are currently pinned to.)

rcernich:
There seems to be a problem with the serialization of `oneof` fields.  I created a simple test case that attempts to load a `VirtualService` that might have been created through the helm installer.  The `Port` fields on the route destinations (`obj.Spec.Tcp.Route.Destination.Port`) fail to deserialize.

rcernich:
Below is an example of an encoded object.  Notice the contents of `spec.tcp.route.destination.port`.

```yaml
        kind: VirtualService
        metadata:
          creationTimestamp: null
          name: meshexpansion-ilb-vs-pilot
          namespace: somenamespace
        spec:
          hosts:
          - istio-pilot.somenamespace.svc.cluster.local
          tcp:
          - route:
            - destination:
                host: istio-pilot.somenamespace.svc.cluster.local
                port:
                  Port:
                    Number: 15011
```

ozevren:
This is precisely the serialization issue I was alluding to in the other thread. :)
I don't think you can safely use simple json serialization for protos. jsonpb is the right thing to use.

rcernich:
Wouldn't that mean we'd need a specialized controller for the CR types then?  I don't believe the built-in k8s api controller uses pbjson.  Maybe somebody else can confirm/deny this.  If it is supported, that's good.  If it's not, then it seems there's a fundamental issue using `oneof` types with/in k8s applications.

rcernich:
It looks like it should be easy enough to shim custom json un/marshalers in for the oneof types.  I'll see if I can find some time to write a protoc plugin to generate the appropriate funcs for oneof types.  I've tested with some manually generated funcs and it seems to work okay with the kubernetes serialization scheme.

rcernich:
I may try updating gogo/protobuf to see if that fixes the problem.

Beside the extra layer, the json names aren't correct for the oneof contents either (no default lowerCamelCase, e.g. `Number` vs `number`).

rcernich:
I've integrated the `protoc-gen-jsonshim` extension into the build and generated _shim_ files where necessary.  De/serialization should now work properly with the kubernetes types.  I'll need to wait for istio/tools#49 to be merged so I can finalize the `Gopkg` files.

costinm:
Confused - what is the purpose of this PR ? It pulls a lot of dependencies - and makes api even more dependent of k8s. 

AFAIK we agreed to use standard protos and grpc - why would we need all this complexity ?

README.md
```diff
@@ -31,3 +31,47 @@ blocked by protolock. To force these changes in, install
 
 You must include a note in your PR that you had to force the
 protolock and why.
+
+## Maintaining Generated Kubernetes Types, Clientsets, Informers, Listers
```
costinm: We are in process to remove all direct uses of Informers/Listeners/etc from everything except Galley.

This may move to a separate repo, or if Galley moves to a separate repo as part of modularizing the repo it can go with Galley - but it is certainly not something that should be part of the API.

ozevren: This PR is mainly for client libraries for external use. If we can make use of it in Galley, I see that as an additional bonus. 

We can probably separate this into its own repo, the generator is in the tools at the moment.

rcernich: The only code that would be required in this repo would be the `deepcopy` functions, as these are defined on the objects.  I'm not sure about the registry.go files.  I'd need to verify whether or not the k8s code generators require those to be collocated with the api definitions.  The clientsets, listers, informers could be moved to a separate repository.  I don't believe there are any k8s runtime dependencies in the deepcopy functions..

rcernich: As @ozevren mentioned, the purpose is to provide typed k8s clients for external users (e.g. operators, management utilities, etc.).

costinm: That can be done in a separate repository - the istio API is not dependent on K8S, and all other components of Istio will not have direct deps. We expect everyone except Galley to use MCP, we do not recommend any client of Istio API to directly operate on k8s.

costinm: I would also not add the dep to tools either - we can create a istio-k8s for all code integrating istio with k8s.

rcernich: As mentioned previously, deepcopy-gen is required, as it generates the `DeepCopy()` functions for the objects.

To clarify, these k8s apis would be used by higher level tools that operate on Istio within k8s, e.g. kiali.  If I understand you correctly, you're saying that these other tools should use MCP instead?

costinm:
Yes, at least for any tool that needs to work in multicluster or mixed
environments it is better to use MCP.

And for k8s - specific tools - or any other environments - different repro
is better.

rcernich:
@costinm, does that mean you're looking to deprecate the CRDs for k8s?  These apis and other helpers (clientsets, etc.) are used to interact with the resources defined through CRDs.

That said, given the desire to move this to a separate repository, how would you like me to proceed?  would you like me to create a new repository in my account, to be used as the seed for something in the istio organization?  If so, I can look at putting something together in the next couple weeks.

costinm:
On Tue, Mar 5, 2019, 07:14 Rob Cernich <notifications@github.com> wrote:

> @costinm <https://github.com/costinm>, does that mean you're looking to
> deprecate the CRDs for k8s? These apis and other helpers (clientsets, etc.)
> are used to interact with the resources defined through CRDs.
>
Of course not - just to better organize the code and the dependencies, and
continue to support other environments for users who need them.





That said, given the desire to move this to a separate repository, how
> would you like me to proceed? would you like me to create a new repository
> in my account, to be used as the seed for something in the istio
> organization? If so, I can look at putting something together in the next
> couple weeks.
>
Let me ask around - istio/istio might be an option, or istio-ecosystem/...
We're trying to get 1.1 out this week, and for future there are few other
refactorings and code org changes.

We're also working on an operator, I think Martin is currently using a
private repo.

Costin

rcernich:
@geeknoid, FYI, I'm also working on an operator.  Here's a link to the repo:

https://github.com/rcernich/istio-operator/tree/native-helm

At this point, it uses a custom resource whose `spec` is a `values.yaml`.  I've gone through the effort of making the `spec` section typesafe and am in the process of getting the rough install working.  After that, I'll be working on the customizations required to get a "clean" install (i.e. I'm just trying to get resources created at this point).

costinm:
@Martin Ostrowski <mostrowski@google.com>

ostromart:
Hey @rcernich , as Costin mentioned I've been working on something similar:
https://github.com/ostromart/istio-installer
there's also a design doc in review I can point you to that's probably a better starting point. I had a quick look at your repo and I think our approaches are very similar in many ways (kubebuilder, helm). 
afaics both operators have a ways to go to be production ready so maybe we can collaborate on this going forward?

rcernich:
Hey @ostromart, that works for me.  I'm going with the helm based approach to minimize deviation from the helm solution (i.e. so we don't have to maintain two different installers).

That said, I just about have a working install (not pushed up, yet).  I have a few details to work out, but I should have a fully functional prototype by EOD tomorrow.

costinm:
Please take a look at istio-ecosystem/istio-installer repo - we are
attempting to modularise the installer and use different namespaces
for components that are managed by different roles ( security, gateways,
telemetry, policy, control plane, etc ).

ozevren:
Had some discussion with @geeknoid. Can we get this into istio/istio, instead of istio/api?

ozevren:
Specifically, does this end up pulling any additional vendor dependencies?

ekarlso:
Any idea on this? We would really like to use a "official" version of a client intsead of rolling our own :)

rcernich:
See my comments [here](https://github.com/istio/api/pull/764#discussion_r262142499), along with the rest of that thread.  I just haven't had time to work on migrating these changes.  If somebody else wants to split up this PR, I'd be happy to advise.

rcernich:
What needs to be done, more specifically, is to leave the `deepcopy` code here (as these functions need to be part of the api), as well as the code generator annotations in the .proto files.  The actual kubernetes code generation can then be moved to whatever project you like.

sdake:
SE support for a fork of Aspenmesh's client API.  There are 15 forks all with varying degrees of implementation.  Can people work together to consolidate on an API, or can we get agreement on what is needed to push this implementation of the finish line from folks in the `OWNERS` file?

https://github.com/aspenmesh/istio-client-go/issues/18

sdake:
And the PR to the most complete API I can find out of the various forks:

https://github.com/michaelkipper/istio-client-go/pull/14

sdake:
I think the most pressing question is whether adding the deepcopy code to this API implementation is acceptable?

The next most pressing question is are annotations acceptable in the API?

The alternatives are not good (forking istio.io/api to annotate and creating types.go, creating a specific client as Aspenmesh has done).

Cheers
-steve

rcernich:
Looking at what's there for deepcopy, I wonder if it's worth generating them at all.  It looks like most of them delegate to the simple wrapper that does a proto marshal/unmarshal.  If that's the case, that could all be managed by hand downstream fairly easily.

As for annotations, I think one could argue that those are part of the API, defining how those types are should be exposed within kubernetes.

If this is acceptable, maybe we can just accept the annotation changes for now, which should allow folks to easily generate k8s apis, client sets, etc. fairly easily.  deepcopy may be a bit of a hassle, but I think it's a once and done sort of thing.

nrjpoddar:
Our version of https://github.com/aspenmesh/istio-client-go was only supposed to be a stop gap until istio provided the K8s like client set functionality.  @sdake we don't copy any Istio generated Go definitions but just add wrappers and Marshalers to facilitate correctly populating the structs.

I have added a bunch of other API wrappers to https://github.com/aspenmesh/istio-client-go, feel free to submit PR if it's missing something.

`deepcopy` is another matter as it requires changing `istio/api` as this PR has tried to do.

sdake:
@nrjpoddar temporary becomes permanent. 😁

It would be nice to push this particular work forward, but without additional feedback on what is necessary on this PR, I am not clear an autogenerated API will ever magically appear in the istio github namespace.

sdake:
Implementation of EnvoyFilter: https://github.com/aspenmesh/istio-client-go/pull/24

paulangton:
Are there any updates on this or [istio/istio#8772](https://github.com/istio/istio/issues/8772)? There is a lot of value and interest in these APIs.

geeknoid:
Closing for now, please reopen if still useful.

PR #767, opened 2019-01-21
 Add Outlier detection to mesh config

hzxuzhonghu:


mesh/v1alpha1/config.proto
```diff
@@ -198,6 +198,9 @@ message MeshConfig {
   // Locality based load balancing distribution or failover settings.
   LocalityLoadBalancerSetting locality_lb_setting = 31;
 
+  // Settings controlling eviction of unhealthy hosts from the load balancing pool.
+  istio.networking.v1alpha3.OutlierDetection outlier_detection = 32;
```
andraxylia: This and above fields should be moved towards the end, in line with their field number.

hzxuzhonghu:
Will rebase right now.

hzxuzhonghu:
rebased

PR #772, opened 2019-01-24
Add sds_name into gateway config.

JimmyCYJ:
Add sds_name into gateway config. In order to obtain credentials, gateway sends SDS requests with this sds_name to gateway agent.


JimmyCYJ:
cc @myidpt @costinm

networking/v1alpha3/gateway.proto
```diff
@@ -335,6 +335,11 @@ message Server {
     // Optional: If specified, only support the specified cipher list.
     // Otherwise default to the default cipher list supported by Envoy.
     repeated string cipher_suites = 9;
+
+    // Optional: Name of credential to be queried by gateway via secret discovery
+    // service(SDS). If specified, gateway obtains credentials via SDS API calls
```
myidpt: s/gateway/the gateway controller

networking/v1alpha3/gateway.proto
```diff
@@ -335,6 +335,11 @@ message Server {
     // Optional: If specified, only support the specified cipher list.
     // Otherwise default to the default cipher list supported by Envoy.
     repeated string cipher_suites = 9;
+
+    // Optional: Name of credential to be queried by gateway via secret discovery
```
myidpt: s/gateway/the gateway controller

networking/v1alpha3/gateway.proto
```diff
@@ -335,6 +335,11 @@ message Server {
...
+    // service(SDS). If specified, gateway obtains credentials via SDS API calls
+    // to gateway agent.
```
myidpt: s/gateway agent/the gateway agent.

networking/v1alpha3/gateway.proto
JimmyCYJ: Done. Thanks!

networking/v1alpha3/gateway.proto
```diff
@@ -336,9 +336,11 @@ message Server {
     // Otherwise default to the default cipher list supported by Envoy.
     repeated string cipher_suites = 9;
 
-    // Optional: Name of credential to be queried by gateway via secret discovery
-    // service(SDS). If specified, gateway obtains credentials via SDS API calls
-    // to gateway agent.
+    // Optional: If specified, the gateway controllers (with SDS enabled)
+    // use the specified name as the SDS secret config name to call the SDS
+    // server, to retrieve the key and certificates. Otherwise, the gateway
+    // controllers (with SDS enabled) uses the first value in the hosts as
```
myidpt: Sorry, nit: uses --> use

networking/v1alpha3/gateway.proto
```diff
@@ -335,6 +335,13 @@ message Server {
     // Optional: If specified, only support the specified cipher list.
     // Otherwise default to the default cipher list supported by Envoy.
     repeated string cipher_suites = 9;
+
+    // Optional: If specified, the gateway controllers (with SDS enabled)
+    // use the specified name as the SDS secret config name to call the SDS
+    // server, to retrieve the key and certificates. Otherwise, the gateway
+    // controllers (with SDS enabled) use the first value in the hosts as
+    // the SDS secret config name to call the SDS server.
```
rshriram: I dont think you should use the first value in hosts. Thats very arbitrary and could end up fetching someone else's secret. There is no mandate for the hosts in a gateway to be unique.

Second, this API is very adhoc. How does this work with the certificates that the end user specifies? Do they work together or is there some other magic behind the scenes ? The agreement earlier was to make this a oneOf such that the user either specifies the certs or the secret name. Not both. And certainly not this gratuitous default of picking the first host.

Please fix this before implementing it.

costinm:
For most users the cert management will be automated, so user doesn't have to manually handle secrets.
 
The 'sdsSecret' is used if user has existing certs in place.

Having both 'file' and 'sdsSecret' is actually intended behavior if user runs Ingresses on both k8s and non-k8s environments. In non-k8s the file will be used. We should make it clear that sdsSecret will take precedence if SDS is available, falling back to file and default.

Naming certs with the first hostname happens to be the convention that few popular ACME tools use.
( and we need to make sure we match). Kubelego for example.

PR #773, opened 2019-01-24
Cherry pick #772 from release-1.1 into master

JimmyCYJ:
* Add sds_name into gateway config

* Update comment

* Update comment


JimmyCYJ:
> please resolve the comments from #772 before merging this to master.

For single host defined in gateway, we can support implicit naming. Customer can either specify sdsName explicitly or let Pilot pick the first Host name as sds resource name. For multiple hosts, we suggest customer to use the sdsName. We can enforce that in config validation. cc @myidpt @costinm

PR #777, opened 2019-01-29
Revert "Add sds_name into gateway config."

rshriram:
Reverts istio/api#772

This change had several implications that were not fully considered. The servers on a gateway are merged. Server hosts can be of ns/* format or just *. 

On an API level, asking for both the certificates and some SDS name does not provide an indication of which one will be used. One could specify arbitrary cert paths and soem valid SDS name or vice versa, making it hard to reason about the system behavior.

The right design should be to use the port name.gatewayname.namespace as the sds name in code, thus requiring no user input in the API. Or atleast a oneOf sdsName and certs so that the user can be explicit about the use of SDS, along with a constraint that SDS names must be unique across all servers in the system.


JimmyCYJ:
cc @myidpt

PR #778, opened 2019-01-29
Enabling SDS in the gateway

rshriram:


wenchenglu:
@myidpt please help review. I cannot add you to reviewer list, not sure why

networking/v1alpha3/gateway.proto
```diff
@@ -291,13 +291,55 @@ message Server {
     // enforced.
     TLSmode mode = 2;
 
+    // The file backend specifies the file system paths where the
+    // server's public and private keys can be obtained.
+    message CredentialStoreFileBackend {
+      // REQUIRED if mode is `SIMPLE` or `MUTUAL`. The path to the file
+      // holding the server-side TLS certificate to use.
+      string server_certificate = 3;
```
JimmyCYJ: Change the field ID to begin with 1?
string server_certificate = 1;
string private_key = 2;

myidpt: I think that's to facilitate the proto field orders when doing the encoding.

myidpt:
I'm generally OK with this approach. Wait for Costin's comments.
@costinm

costinm:
+1, it's a good compromise, we can refine it in 1.2 based on user feedback. 

I like the idea of using Gateway resource name as key in the SDS storage, it seems to solve all automated cases.

rshriram:
FYI, the server certificates is still there. 

```
servers:
- hosts:
   - foo.com
   tls:
    credentialStore:
      files:
        serverCertificate: ...
        privateKey...
```

Or 

```
servers:
- hosts:
   - foo.com
  tls:
    credentialStore:
     remote:
       resourceName: mysecret
```

Only the secret name is mandatory. the server path is additional customization.

And systems that are auto provisioning the gateways - they are already generating the Gateway configs and the k8s secrets based on the hostnames. So all they have to do is something like this:

```
servers:
- hosts:
   - foo.com
tls:
 credentialStore:
   remote:
     resourceName: foo.com
```

Explicit spec avoids confusions caused by auto inference of that one single field

PR #780, opened 2019-01-30
Fixing SDS field/semantics in the gateway

rshriram:
What this API allows is to always have SDS agent in the gateway pod 
while still using old style file based configs. Then users can incrementally 
migrate to the SDS server by changing their existing gateway configs to use 
the CredentialStoreRemoteBackend with the resourceName as secretName. In other 
words, it allows some servers in a gateway to have file based paths and others to have SDS.


rshriram:
repeating what I said earlier 

The server certificates is still there.

```
servers:
- hosts:
   - foo.com
   tls:
    credentialStore:
      files:
        serverCertificate: ...
        privateKey...
```

Or

```
servers:
- hosts:
   - foo.com
  tls:
    credentialStore:
     remote:
       resourceName: mysecret
```

Only the secret name is mandatory. the server path is additional customization.

And systems that are auto provisioning the gateways - they are already generating the Gateway configs and the k8s secrets based on the hostnames. So all they have to do is something like this:

```
servers:
- hosts:
   - foo.com
tls:
 credentialStore:
   remote:
     resourceName: foo.com
```

Explicit spec avoids confusions caused by auto inference of that one single field

rshriram:
ping

networking/v1alpha3/gateway.proto
```diff
@@ -299,6 +299,19 @@ message Server {
...
 
+    // The credentialName stands for a unique identifier that can be used
+    // to identify the serverCertificate and the privateKey (not the
+    // CaCertificates) associated with this server. Gateway workloads
+    // capable of fetching credentials from a remote credential store will
+    // be configured to retrive the server certificate and the private key
+    // using this name, instead of using the file system paths specified
+    // above. The semantics of the name are platform dependent. In
+    // Kubernetes, the default Istio supplied credentail server expects the
+    // credentialName to be of the form secretName.namespace, where the
```
myidpt: namespace is not required. The controller must always retrieve the secret from its own namespace.
@louiscryan

networking/v1alpha3/gateway.proto
```diff
@@ -299,6 +299,18 @@ message Server {
     // holding the server's private key.
     string private_key = 4;
 
+    // The credentialName stands for a unique identifier that can be used
+    // to identify the serverCertificate and the privateKey (not the
+    // CaCertificates) associated with this server. Gateway workloads
```
costinm: Curious - why not CaCertificates ? For MTLS we need that too. Not a big deal...

rshriram: I dunno.. @myidpt said only the public and private key is shipped

myidpt: Clarified. We support delivering the cacerts in SDS as well.

networking/v1alpha3/gateway.proto
```diff
@@ -304,6 +304,18 @@ message Server {
...
     string ca_certificates = 5;
 
+    // The credentialName stands for a unique identifier that can be used
+    // to identify the serverCertificate, the privateKey and the
+    // CaCertificates associated with this server. Gateway workloads
+    // capable of fetching credentials from a remote credential store will
+    // be configured to retrive the credentials using this name, instead of
+    // using the file system paths specified above. The semantics of the
+    // name are platform dependent. In Kubernetes, the default Istio
+    // supplied credentail server expects the credentialName to match the
```
louiscryan: @rshriram doc needs a little work :)

PR #781, opened 2019-01-30
revert sds name addition to gateway

rshriram:
While we continue to discuss the right way of introducing SDS config
into the gateway, revert the existing one which is not the right
way of introducing the feature as it goes against our current
API design, forcing users to specify all options (files, sds names, etc.)
without any idea of which one would be picked by the underlying system.
This might be useful for automation code but definitely not for end users
authoring configs.


wenchenglu:
@costinm 
@myidpt

is this the same issue you guys have discussed recently? have we reached agreement?

myidpt:
@rshriram 
Do you agree on Costin's comment:
"Let's try to use the gateway name in the implementation, so we can solve the immediate problem in ingress for 1.1 - we can add more APIs and customization in 1.2 after user feedback"
This means in 1.1, when the sds_name is not specified, we will use the "portname.gatewayname.namespace" to infer the the secret.

We need to make an agreement for 1.1 so that we can make corresponding changes on Pilot early. Otherwise, we will have risk on delivering the feature in 1.1.

myidpt:
@wenchenglu Please wait until Shriram agrees on the plan here:
https://github.com/istio/api/pull/781#issuecomment-459153585
My concern is if we don't have consensus on 1.1, we will have trouble moving forward to deliver the feature in 1.1.

rshriram:
I would love to use the portName.gatwayName.namespace .. But how do you infer the name of the secret from this? You would have to read the gateway spec from k8s, and extract the specific gateway from the specific namespace, get to the specific server in the gateway using the portName. And from there, you look at the cert paths in the server (/etc/istio/certs/<secretName>/..), extract the secret and then fetch the secret.

If this is okay by you, lets go for it. There is no need for any additional field.

PR #785, opened 2019-02-04
Add service_accounts field in ServiceEntry

andraxylia:
As discussed in istio/istio#11293 (comment)

The service_accounts field in ServiceEntry contains the reunion of all SAs from all workloads that are backends for that Service. Each MCP source computes the reunion and sends it to Pilot. Pilot merges the service_accounts corresponding to that service from all MCP sources, and maps the result to subject_alt_names field in the Envoy CDS for that ServiceEntry.

cc @nmittler


networking/v1alpha3/service_entry.proto
```diff
@@ -461,4 +461,8 @@ message ServiceEntry {
   // the annotation "networking.istio.io/export_to" to a comma-separated list
   // of namespace names.
   repeated string export_to = 7;
+
+  // The reunion of service accounts associated with workloads
```
nmittler: It might be helpful to describe what exact form this should take.  It came up in the meeting today and I suspect there will be confusion for folks going forward.  Would be good to try to make it clear now.

louiscryan: @smawson  - For TOC FYI even though this is not a user-visible API change, it affects contract between Pilot and Galley or other service registries

This should be marked $hidden_from_docs for the moment.

For documentation

"The set of service accounts identities allowed for workloads that implement this service. This information is used to enforce secure-naming (link https://istio.io/docs/concepts/security/#secure-naming)"

andraxylia: Done, added also an example with SA in SPIFEE format, ptal.

networking/v1alpha3/service_entry.proto
```diff
@@ -461,4 +461,36 @@ message ServiceEntry {
...
+  //     name: http
+  //     protocol: HTTP
+  //   resolution: STATIC
+  //   endpoints:
+  //   - address: 2.2.2.2
+  //   - address: 3.3.3.3
+  //   serviceAccounts:
+  //   - "spiffe://cluster.local/ns/httpbin-ns/sa/httpbin-service-account"
+  // ```
+  repeated string service_accounts = 8;
```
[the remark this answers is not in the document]
andraxylia: Makes sense to use the same nomenclature with existing, let me change it.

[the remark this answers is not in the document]
andraxylia: I changed the name.

We need to clarify what is the behavior if the user also configures a DestinationRule for the same service, with a different list of subject_alt_names. Do we want to override or merge? Override may cause auth failures, so merge seems to make more sense. Thoughts?

costinm:
Needs to override the synthetic service entry for sure - DR indicates what
user explicitly expects the service to authenticate as.

Or better - Galley could validate the endpoints, and if they don't run as
one of the identities declared in the DR, don't add it to the set/warn.

Also extra care on the namespace - the current scheme of using whatever the
deployment is running as secure identity only works because
in practice all deployments and SA in a namespace are mostly equivalent
from security perspective ( with current K8S implementation of SA).
If the SA in ServiceEntry is not constrained to the same NS - we make a
huge hole. This is super security-sensitive, I'm not even sure having
full spiffe URL is acceptable since it bypasses the namespace boundary.

networking/v1alpha3/service_entry.proto
```diff
@@ -461,4 +461,38 @@ message ServiceEntry {
...
+  //   - number: 80
+  //     name: http
+  //     protocol: HTTP
+  //   resolution: STATIC
+  //   endpoints:
+  //   - address: 2.2.2.2
+  //   - address: 3.3.3.3
+  //   subjectAltNames:
+  //   - "spiffe://cluster.local/ns/httpbin-ns/sa/httpbin-service-account"
+  // ```
```
rshriram: You cannot add yamls and other stuff in the field docs. It will mess up the UI. Move this example and text to the area above `message ServiceEntry`. And keep the documentation here to a minimum, describing the intent of the field.

rshriram:
while in general this is okay, its instructive to understand why the exact same change 6 months ago (https://github.com/istio/api/pull/646/commits/2ba6be935959ad9bcb0ec7827a5079bb3b9ae77c) with the same intent (of syncing service model with service entry) was objected to, by messers @costinm and @louiscryan :).

costinm:
See above - the secure naming scheme is not very secure if anyone who can
run a workload or create a ServiceEntry can modify the list of 'trusted
identities' of a service.

This is a very security-sensitive change - for the synthetic entries
generated by galley it might be fine if Galley is making sure they are
restricted to the namespace.
Allowing them in user-specified SE without equivalent validation - I hope
not.

This should not be a user-visible config.

networking/v1alpha3/service_entry.proto
andraxylia: Done, ptal.

andraxylia:
Galley can create the subject_all_names by using the namespace of the service : "spiffe://cluster.local/ns/httpbin-ns/sa/httpbin-service-account".

The list of endpoints for that service can only belong in the same namespace as the service in k8s, otherwise the pods would not be selected.

Since ServiceEntries can come from other MCP sources than Galley, same restriction applies.

networking/v1alpha3/service_entry.proto
```diff
@@ -461,4 +488,12 @@ message ServiceEntry {
   // the annotation "networking.istio.io/export_to" to a comma-separated list
   // of namespace names.
   repeated string export_to = 7;
+
+  // $hide_from_docs
```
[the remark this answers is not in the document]
andraxylia: done

ozevren:
@wenchenglu This is needed for https://github.com/istio/istio/issues/10589

PR #786, opened 2019-02-06
Update comment for credentialName

JimmyCYJ:
Update comment for credentialName.
When customer specifies credentialName (e.g. "foo.credential"), and use MUTUAL TLS.
Gateway workloads will send SDS requests with resource name "foo.credential" to retrieve serverCertificate and privateKey, and send SDS requests with resource name "foo.credential-cacert" to retrieve CaCertificates. In Kubernetes, customer needs to create a Kubernetes secret named "foo.credential", which contains server certificate, the private key, and the CA certificate.
If customer specifies credentialName (e.g. "foo.credential"), and use SIMPLE TLS. Gateway workload will only send SDS requests with resource name "foo.credential" to retrieve serverCertificate and privateKey. And the Kubernetes secret named "foo.credential" only contains server certificate and private key.

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -305,15 +305,18 @@ message Server {
     string ca_certificates = 5;
 
     // The credentialName stands for a unique identifier that can be used
-    // to identify the serverCertificate, the privateKey and the
-    // CaCertificates associated with this server. Gateway workloads
-    // capable of fetching credentials from a remote credential store will
-    // be configured to retrive the credentials using this name, instead of
-    // using the file system paths specified above. The semantics of the
-    // name are platform dependent. In Kubernetes, the default Istio
-    // supplied credentail server expects the credentialName to match the
-    // name of the Kubernetes secret that holds the server certificate, the
-    // private key, and the CA certificate (if using mutual TLS).
+    // to identify the serverCertificate and the privateKey. The credentialName
+    // appended with suffix "-cacert" is used to identify the CaCertificates
+    // associated with this server. Gateway workloads capable of fetching
+    // credentials from a remote credential store will be configured to retrieve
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -305,15 +305,18 @@ message Server {
...
-    // private key, and the CA certificate (if using mutual TLS).
+    // to identify the serverCertificate and the privateKey. The credentialName
+    // appended with suffix "-cacert" is used to identify the CaCertificates
+    // associated with this server. Gateway workloads capable of fetching
+    // credentials from a remote credential store will be configured to retrieve
+    // the serverCertificate and the privateKey using credentialName, instead of
+    // using the file system paths specified above. If using mutual TLS,
+    // gateway workloads will retrieve the CaCertificates using
+    // credentialName-cacert. The semantics of the name are platform dependent.
+    // In Kubernetes, the default Istio supplied credentail server expects the
```
rshriram: typo: credential

networking/v1alpha3/gateway.proto
```diff
@@ -305,15 +305,18 @@ message Server {
...
+    // credentialName to match the name of the Kubernetes secret that holds the
+    // server certificate, the private key, and the CA certificate
```
rshriram: and if using mTLS, "credentialName-cacert" to match the name of the Kubernetes secret that holds the CA certificate.

JimmyCYJ: If using mTLS, customer only creates one k8s secret with credentialName (e.g. "foo.credential"), Pilot creates two SDS configs for gateway listener, one is named "foo.credential", and is used to fetch key and cert. The other is named "foo.credential-cacert", and is used to fetch certificate validation context that has CA cert. So I think the comment is okay.

PR #787, opened 2019-02-06
Add attributes for destination rules and virtual services

douglas-reid:
Pending attributes review.

See: https://discuss.istio.io/t/rfc-new-attributes-for-virtual-service-and-destination-rule/734

### DIFF
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -324,3 +324,5 @@
 - networking.virtual_service.uid
 - networking.virtual_service.name
 - networking.virtual_service.namespace
+
+- istio-system
--- a/mixer/v1/global_dictionary.yaml
+++ b/mixer/v1/global_dictionary.yaml
@@ -317,3 +317,10 @@
 
 - inbound
 - outbound
+
+- networking.destination_rule.uid
+- networking.destination_rule.name
+- networking.destination_rule.namespace
+- networking.virtual_service.uid
+- networking.virtual_service.name
+- networking.virtual_service.namespace


douglas-reid:
Added "istio-system"

geeknoid:
I've got three bits of feedback:

- Why networking.*? Many of our attributes are about networking concepts, so using networking here doesn't really fit IMO. How about route.* instead?

- I presume initially, these attributes will only be present when context.reporter.kind==outbound, correct? I think this is leaking implementation details of the service to its client, it's breaking encapsulation. Are we OK with that? I realize the routing behavior is physically handled by the client, but this is logically a server-side concern.

- Presumably, to get this working for context.reporter.kind==inbound, we would need to use a tunnel to send the info from the client to the server, right?

kyessenov:
1. I tend to agree that `config.networking` or just `config` is more appropriate. But these attribute names are internal most of the time, so not directly user visible. I do want to make sure we keep proto-compatible schema.

2. It is our goal to hide minute details about which reporter is reporting as much as possible. Config is a property of the connection, so I think it's OK to forward it to the server from the client, just like we do with source. It's a question of perf overhead.

douglas-reid:
Are virtual services and destination rules really strictly server-side concerns?  They seem to be about how clients should contact a service. The information, btw, that is exposed is just the uid of the config, and not the config itself -- though I agree that is a new set of info exposed.

These (currently) are only available in outbound cases and at report time (client-side telemetry). I don't know what is required within the proxy implementation to forward that information across for server-side actions (check or reports).

As for naming, I don't have strong feelings in any direction. If we want `config` or `route` or `???` I'm fine.

geeknoid:
@louiscryan how are you guys thinking of the VS and DR resources? I seem to recall you're messing around with this stuff in order to put constraints on who can modify what resources, based on namespace boundaries. So is a VS and DR generally a client concern or a server concern?

geeknoid:
@kyessenov Why do you say that these attributes are mostly internal?

kyessenov:
@geeknoid Internal in the sense that they don't show up in dashboards, logs, tracing, etc visible to application developers. Operators are expected to be more familiar with internals of the system.

andraxylia:
@douglas-reid do we move forward with this one?

douglas-reid:
@andraxylia probably not? this was something we want to expose in 1.1. But disagreements about the visibility of these attributes led to a postponement. Because this info is available in the proxy metadata now, and because that's where Extensibility v2 will be looking for them, we may be able to avoid creating attributes for this information directly.

geeknoid:
@douglas-reid Do we still need this PR?

geeknoid:
Closing for now, please reopen if this is still useful.

PR #790, opened 2019-02-09
Rename "principals" to "names".

liminw:
Since this is defined under "subjects", we are basically referring to
the "name" of a subject.


liminw:
cc @pitlv2109

PR #791, opened 2019-02-09
Rename "principals" to "names".

liminw:
Since this is defined under "subjects", we are basically referring to
the "name" of a subject.


rbac/v1alpha1/rbac.proto
```diff
@@ -235,13 +235,14 @@ message Subject {
   string user = 1;
 
   // $hide_from_docs
-  // Optional. A list of principals that the subject represents. This is matched to the
-  // `source.principal` attribute. If not specified, it applies to any principals.
-  repeated string principals = 4;
+  // Optional. A list of subject names. This is matched to the
+  // `source.principal` attribute. If one of subject names is "*", it matches any subject.
```
yangminzhu: `If one of subject names is "*", it matches any subject.`
I find this a little confusing, does the second `subject` refer to the whole `subject` message (including other fields)? Or does it mean only the `names` field in the `subject`?

liminw: It only applies to the "names" field in the subject. Other conditions for the subject still needs to be evaluated.

PR #795, opened 2019-02-13
Fix typos in sidecar.proto

howardjohn:


geeknoid:
Could you also make this change in the 1.1 branch (assuming the same text is there)?

Thanks.

geeknoid:
You need to run "make" before submitting the PR so that the (unforutnately) checked in generated files are correctly updated. Doing so will fix the CI failures.

howardjohn:
Sure, here is the PR for 1.1 https://github.com/istio/api/pull/796

PR #799, opened 2019-02-14
Add idle timeout

devincd:
add idle timeout to meshConfig
link istio/istio PR https://github.com/istio/istio/pull/11756


rshriram:
We dont set HTTP keep alive. So, the Envoy issue related to this PR doesn't affect us right?

devincd:
Yes, but when the envoy serves as the client connecting to the banckend, the default envoy is turn on keep-alive.At this time, the server side will actively disconnect due to the arrival of keep-alive timeout on the server side.

costinm:
Does it even make sense to have this as a global setting ? I don't mind adding it to Sidecar or DestinationRule - but mesh config (in a normal install) can only be changed when doing an upgrade/restart, and it applies to the entire mesh. I can't imaging all servers in the mesh having same value.

rshriram:
@devincd , if idle timeout is not set, its not enabled. So there is no keep alive timeout at all.

geeknoid:
Closing due to inactivity.

Issue #567, opened 2018-07-03
Document semantics of MCP


ayj:
Document MCP protocol semantics. MCP is conceptually aligned with [xDS](https://github.com/envoyproxy/data-plane-api/blob/master/XDS_PROTOCOL.md), but behavior may different slightly as the protocols evolve.

ayj:
fixed by https://github.com/istio/api/pull/805

PR #805, opened 2019-02-20
Add high-level overview and diagrams for MCP

ayj:
fixes https://github.com/istio/api/issues/567


mcp/Readme.md
```diff
@@ -1,13 +1,141 @@
...
+pushed `Resources` will always have incremental set to false,
+regardles of whether the sink requested an incremental update. At any
+time, the source can decide to push a full-state update, ignoring the
+sink's request. Both sides must negotiate (i.e. agree) to use
+incremental on a per request/response basis for an update to be sent
+incrementally.
+
+![Resource full-state incremental success](v1alpha1/diagrams/collection-incremental-update-success.svg)
+
+### Error example
```
ozevren: Can you add more details here, specifically about the behavior of server in case of nack in both incremental and non-incremental cases?

ayj: done

ozevren: How about incremental behavior in case of nack? Does the server try to do keep doing incremental, or should it switch to full-state transfer?

mcp/Readme.md
```diff
@@ -1,13 +1,148 @@
...
+resources was invalid, malformed, or could not be decoded. NACK'd updates
+should raise an alarm for subsequent investigation by a human. The source 
+should not resend the same set of resources that were previously NACK'd. 
+Canary pushes to dedicated sinks may also be used to verify correctness 
+(non-NACK) before pushing to a larger fleet of resource sinks.
+
+The nonce in MCP is used to match RequestResources and Resources. On
+reconnect, the sinks may attempt to resume a session with the same
+source by specifying the known resources version with
+initial_resource_version for each collection.
```
ozevren: Can you describe the behavior of what happens when there is an unknown nonce? (i.e. server is restarted).

mcp/Readme.md
ayj: There currently isn't any difference in handling NACK between incremental and non-incremental. NACK is really an exceptional case which requires deeper human investigation and corrective measure, e.g. deploy a new version of Pilot. Switching to full-state shouldn't make a meaningful difference if, for example, schema versions are out of sync.

mcp/Readme.md
ayj: Updated - tl;dr requests with unknown/stale nonces are ignored.

ayj:
/assign @geeknoid 

@duderino @geeknoid @nino-K - this is a developer documentation change that would be nice to have in 1.1 release. However, we can also push it into the master branch of istio/api and point platform integrators (e.g. CF) there with the understanding that it applies to the 1.1 release.

PR #803, opened 2019-02-20
Add envoy metrics service configuration to ProxyConfig

joeyb:
Similar to the statsd sink config, this change is in preparation for adding support for the envoy metrics service API.

PR #809, opened 2019-02-22
Rename metrics_service_address to clarify that it is for Envoy's API

joeyb:
Follow-up to #803, Related to https://github.com/istio/istio/pull/11897

Since "metrics service" is so ambiguous, this change specifically denotes that the config is for Envoy's metrics service API. The comments also now link to detailed info about the metrics service API.

cc: @rshriram


rshriram:
thanks for the clarification

duderino:
Hi @joeyb and @rshriram I just noticed this when reviewing the changes for 1.1 RC1.  Could you hide these changes from docs until the feature is ready?

PR #811, opened 2019-02-22
Generate separate HTML page per proto under networking

rshriram:
This makes the networking reference docs a bit more readable instead
of one large compendium of all protos.


rshriram:
an example of how the docs are rendered: https://deploy-preview-3358--preliminary-istio.netlify.com/docs/reference/config/networking/

frankbu:
This is great. Makes "the docs a bit more readable ..." is an understatement. They are WAY MORE readable.

You need to fix the broken inter-doc and external links, though, or you won't be able to push a PR istio.io

geeknoid:
You need to add $aliases directives to point old links to new places.

rshriram:
we dont have any links to these fields directly. I added the alias in istio.io to redirect to the index page for this section..

geeknoid:
I just reverted this change since it's breaking the ability to generate working reference docs in istio.io. We need to work out linking issues before redoing this change.

geeknoid:
https://github.com/istio/api/pull/812 updates the tools. Once that PR is merged, you can retry this PR and hopefully this will fix some of the link issues with the per-file mode.

PR #813, opened 2019-02-24
Update to latest tools.

geeknoid:


geeknoid:
I'm quite puzzled by those CI failures. Everything builds fine locally for me, and since the tools are coming out of a docker image I don't understand how it could work for me and not the CI systems.

All I want to do is update the protoc-gen-docs plugin, but alas given our Dockerfile rebuilding the docker image ends up picking up the latest of gogo too, which seems to be problematic.

ozevren:
Stupid Q: have you tried running at the tip of the master? One explanation is, you may not have the PR that is causing the issue locally, in your branch. Fetching/rebasing can pull it and allow you to repro the problem locally.

rshriram:
I can reproduce this error locally if I use this docker image..

PR #824, opened 2019-02-28
Proxy config: Add tls_certs_to_watch field

drichelson:
See comments [here](https://github.com/istio/istio/issues/11984) for context


mesh/v1alpha1/proxy.proto
```diff
@@ -169,4 +169,7 @@ message ProxyConfig {
 
   // Tracing configuration to be used by the proxy.
   Tracing tracing = 19;
+
+  // Absolute paths to cert/key files to watch for changes (e.g. /etc/certs/key.pem,/etc/certs/cert.pem)
```
rshriram: Absolute path to certs/key files used for istio mTLS. If omitted, defaults to /etc/certs/...

mesh/v1alpha1/proxy.proto
```diff
@@ -169,4 +169,8 @@ message ProxyConfig {
 
   // Tracing configuration to be used by the proxy.
   Tracing tracing = 19;
+
+  // Absolute path to certs/key files used for istio mTLS. If omitted, defaults to
+  // /etc/certs/cert-chain.pem,/etc/certs/key.pem,/etc/certs/root-cert.pem
+  repeated string tls_certs_to_watch = 21;
```
hzxuzhonghu: I have a question : how can you distinguish which is the key/cert file?

rshriram: we dont. pilot agent simply watches files in that path and if any of them change, it forces envoy to restart

PR #826, opened 2019-02-28
Update gateway and sidecar hosts doc

frankbu:


networking/v1alpha3/gateway.proto
```diff
@@ -242,20 +242,29 @@ message Server {
...
-  // SNI. May contain a wildcard prefix for the bottom-level component of
-  // a domain name. For example `*.foo.com` matches `bar.foo.com`
-  // and `*.com` matches `bar.foo.com`, `example.com`, and so on.
+  // REQUIRED. One or more hosts exposed by this gateway.
+  // While typically applicable to
+  // HTTP services, it can also be used for TCP services using TLS with SNI.
+  // A host is specified as a `dnsName` with an optional `namespace/` prefix.
+  // The `dnsName` should be specified using FQDN format, opionally including
+  // a wildcard character in the left-most component (e.g., `prod/*.example.com`).
+  // Set the `dnsName` to `*` to select all services from the specified
```
rshriram: to select all virtual service hosts from .. and else where, might need to nuance that the service is actually the host in the virtual service

networking/v1alpha3/gateway.proto
```diff
@@ -242,20 +242,29 @@ message Server {
...
-  // example, if the server's hosts specifies "*.example.com",
-  // VirtualServices with hosts dev.example.com, prod.example.com will
-  // match. However, VirtualServices with hosts example.com or
-  // newexample.com will not match.
+  // example, if the server's hosts specifies `*.example.com`, a
+  // `VirtualService` with hosts `dev.example.com` or `prod.example.com` will
+  // match. However, a `VirtualService` with host `example.com` or
+  // `newexample.com` will not match.
+  //
+  // NOTE: Only exported virtual services from a
```
rshriram: only virtual services with exportTo set to * .. Private (i.e. exportTo set to .)...

networking/v1alpha3/sidecar.proto
```diff
@@ -178,24 +178,24 @@ message IstioEgressListener {
...
-  // the service registry (e.g., a Kubernetes or cloud foundry service) or
-  // a service specified via ServiceEntry configuration. In addition, any
-  // publicly scoped DestinationRule associated with the imported services
-  // will also be imported.
+  // REQUIRED: One or more service hosts exposed by the listener
+  // in `namespace/dnsName` format. Services in the specified namespace
+  // matching `dnsName` will be exposed.
+  // The corresponding service can be a service in the service registry
+  // (e.g., a Kubernetes or cloud foundry service) or a service specified
+  // using a `ServiceEntry` or `VirtualService` configuration. Any
```
rshriram: not virtual service.. just service entry or the registry

rshriram: sorry. ignore the comment

PR #828, opened 2019-03-01
Add config options for mesh server certs and watched certs in Pilot-agent

drichelson:
https://github.com/istio/istio/issues/11984
This covers [TLSSettings_MUTUAL](https://github.com/istio/istio/blob/96d0abec5c367e796a2e493af91bfed16bd66fef/vendor/istio.io/api/networking/v1alpha3/destination_rule.pb.go#L128) mode only.


mesh/v1alpha1/config.proto
```diff
@@ -279,9 +279,16 @@ message MeshConfig {
   // Locality based load balancing distribution or failover settings.
   LocalityLoadBalancerSetting locality_lb_setting = 35;
 
+  // The cert files to use when configuring Envoy listeners for mesh traffic.
+  // Used only when sds_uds_path is not set.
+  // If not set these defaults will be used:
+  //  cert_chain = '/etc/certs/cert-chain.pem'
+  //  key = '/etc/certs/key.pem'
+  //  root_cert = '/etc/certs/root-cert.pem'
+  ServerCertPaths mesh_traffic_server_cert_paths = 36;
```
rshriram: mtls_cert_paths

Also, why just server? it applies at client side too..

costinm:
/hold

AFAIK we already have a way to configure the certificate location - using one of the metadata. 

Even if we didn't - a global setting that applies to the entire mesh is the worst way to fix this problem. 
Keep in mind mesh config is a global setting. Just because some VMs use a different path doesn't mean
the entire mesh will need to use the same pat. 

Also mesh config is not really user editable and has huge impact on how mounts happen, etc. This is not an API we can even support - think of all the yaml files where we mount.  

Strong -1 on this change. 

Not sure if the using metadata (env var) is merged yet - but if this is for VM support it's the only viable option. SDS is also going to move to a metadata - so in a mesh we don't force everything at once to use the same setting with no way to gracefully migrate. 

In general - we must stop using global settings, they  were ok for the bookinfo demo but not for any complex  production environment.

mesh/v1alpha1/config.proto
drichelson: @rshriram for `TLSSettings_MUTUAL` mode the client certs are set by DestinationRules. This covers our use case. I'm open to ideas on how to cover the `TLSSettings_ISTIO_MUTUAL` mode with an example scenario.

rshriram: if its TLS_ISTIO_MUTUAL, then the changes you have here should be useful there as well.. When setting up the destination rule, if tls mode is istio_mutual, then use the cert paths specified here instead of using the default ones in model.Istio... 
(only for ISTIO_MUTUAL).

drichelson:
@costinm Thanks for taking a look.

> AFAIK we already have a way to configure the certificate location - using one of the metadata.
This sounds awesome. I'm working on master branch. Any idea where/how this will work?

> Even if we didn't - a global setting that applies to the entire mesh is the worst way to fix this problem.
I was following a suggestion based on [this comment](https://github.com/istio/istio/issues/11984#issuecomment-468316915). I can rework this if needed

> Keep in mind mesh config is a global setting. Just because some VMs use a different path doesn't mean
> the entire mesh will need to use the same pat.
Agreed. Our internal control-plane + mesh configures cert paths for each envoy, then passes the paths to the control plane as node metadata and the clusters/listeners are built using those paths. In Istio this metadata could serve to override the `/etc/certs/...` defaults.

> Also mesh config is not really user editable and has huge impact on how mounts happen, etc. This is not an API we can even support - think of all the yaml files where we mount.
> 
> Strong -1 on this change.
> 
> Not sure if the using metadata (env var) is merged yet - but if this is for VM support it's the only viable option. SDS is also going to move to a metadata - so in a mesh we don't force everything at once to use the same setting with no way to gracefully migrate.
Yes this is for VM/baremetal support where we don't control cert file locations.

> In general - we must stop using global settings, they were ok for the bookinfo demo but not for any complex production environment.
+1

rshriram:
@costinm this is not for VMs. this is a mesh wide setting as they have k8s clusters with their inhouse equivalent of citadel, mounting certs in a different path, instead of /etc/certs as we do. They (salesforce) also dont have SDS as of now.

We also dont have a metadata variable for per proxy cert paths. It would be nice to have it, to override as you said, but the only thing we have is the ability to specify the UDS path for SDS.

rshriram:
> Yes this is for VM/baremetal support where we don't control cert file locations.

?? I had assumed based on the original issue that this is for k8s clusters where you have certs mounted in different paths than the standard one. If this is for one-off VMs or bare metal machines, then node metadata is the way to go. In fact, setting a global config will force everything to use this path and break stock istio deployment on k8s.

drichelson:
@rshriram Our mesh is a mixed environment with K8s + VMs/baremetal hosts. Setting global paths for certs will work because we can always configure K8s volume mounts to result in the expected paths.

rshriram:
True, but it seems easier to simply leave k8s as it is and add the special metadata to proxies that need to have non standard (per istio code) paths.. 
In such a scenario, you dont need any API change PR. simply add a metadata field to the proxy in istio/istio, and update the code references to consider this metadata (if present) as the base path for certs.

drichelson:
@rshriram This approach works fine for us. I'll remove the changes to MeshConfig and add specific fields in the proxy config for each cert.

drichelson:
@rshriram @costinm I am closing this PR. Please see https://github.com/istio/istio/pull/12189

costinm:
For the record - ../features/pilot/pilot.go defines "BASE" metadata, which
is used in authentication plugin.

It was added for local tests - for same reasons, we can't require root for
local tests.
If we want to expose it to users - we may pick a better name. The intent
was to base all other absolute paths on this meta.

PR #829, opened 2019-03-01
Add jwks to JWT policy to allow using public key for token validation.

yangminzhu:
https://github.com/istio/istio/issues/11439

Moving a field to oneof should have no changes required for yamls. But we need to change some pilot code when syncing this API change. If prefer to avoid the change, we can add the `jwks` field as a new field without using oneof.

See [design doc](https://docs.google.com/document/d/1bMyVHHqbHqx0jsDGmHw_PuXmE-X1TM7Rfcff8T-izZE/edit#heading=h.7zgnj8bwqfld) (Require joining Istio community)


ozevren:
I think we should avoid using oneof, if possible. It is a constant source of pain when dealing with serialization/transformation.

yangminzhu:
@ozevren Thanks for the suggestion.
@diemtvu @geeknoid If there're no other opinions, I will add it as a separate field without using oneof.

PR #831, opened 2019-03-05
Update API docs (service entry, gateway, sidecar)

rshriram:
cc @frankbu

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -24,12 +24,12 @@ syntax = "proto3";
 // use, SNI configuration for the load balancer, etc.
 //
 // For example, the following Gateway configuration sets up a proxy to act
-// as a load balancer exposing port 80 and 9080 (http), 443 (https), and
-// port 2379 (TCP) for ingress.  The gateway will be applied to the proxy
-// running on a pod with labels `app: my-gateway-controller`. While Istio
-// will configure the proxy to listen on these ports, it is the
-// responsibility of the user to ensure that external traffic to these
-// ports are allowed into the mesh.
+// as a load balancer exposing port 80 and 9080 (http), 443 (https),
+// 9443(https) and port 2379 (TCP) for ingress.  The gateway will be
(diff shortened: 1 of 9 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -62,6 +62,15 @@ syntax = "proto3";
 //       serverCertificate: /etc/certs/servercert.pem
 //       privateKey: /etc/certs/privatekey.pem
 //   - port:
+//       number: 9443
```
frankbu: text above should mention this

networking/v1alpha3/gateway.proto
```diff
@@ -153,6 +163,32 @@ syntax = "proto3";
 //         port:
 //           number: 5555
 // ```
+//
+// It is possible to restrict the set of virtual services that can bind to
+// a gateway server using the namespace/hostname syntax in the hosts field.
+// For example, the following Gateway allows any virtual service in the ns1
+// namespace to bind to it, while restricts only the virtual service with
```
frankbu: restricting

networking/v1alpha3/gateway.proto
```diff
@@ -153,6 +163,32 @@ syntax = "proto3";
...
+// foo.bar.com host in ns2 namespace to bind to it.
```
frankbu: in the

networking/v1alpha3/service_entry.proto
```diff
@@ -72,27 +94,24 @@ import "networking/v1alpha3/gateway.proto";
 // ```
 //
 // The following example uses a combination of service entry and TLS
-// routing in virtual service to demonstrate the use of SNI routing to
-// forward unterminated TLS traffic from the application to external
-// services via the sidecar. The sidecar inspects the SNI value in the
-// ClientHello message to route to the appropriate external service.
+// routing in virtual service to steer traffic based on the SNI value to an
```
frankbu: a virtual service

PR #832, opened 2019-03-05
add configurable dns refresh rate to mesh for strict dns cluster, 

GregHanson:
https://github.com/istio/istio/issues/12181


rshriram:
fix for https://github.com/istio/istio/issues/12181
@dwradcliffe and @michaelkipper this should allow you to set the dns refresh rate globally for all proxies (and thus cut down the load on your DNS servers).

cc @duderino

PR #833, opened 2019-03-05
Hosts are not ignored in ServiceEntry for HTTPs

andraxylia:


networking/v1alpha3/service_entry.proto
```diff
@@ -339,18 +339,22 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
-  // name with wildcard prefix (external services only). DNS names in hosts
-  // will be ignored if the application accesses the service over non-HTTP
-  // protocols such as mongo/opaque TCP/HTTPS. In such scenarios, the
-  // IP addresses specified in the Addresses field or the port will be used
-  // to uniquely identify the destination.
+  // name with wildcard prefix (external services only). For HTTP traffic
+  // the HTTP Host/Authority header will be matched against the hosts field.
+  // For HTTPs or TLS traffic containing Server Name Indication (SNI), the SNI value
+  // will be matched against the hosts field. For non-HTTP(s) protocols
+  // such as TCP/mongo the hosts will be ignored.
```
rshriram: For all other protocols, the port and addresses field if present will be used. Note that when resolution is set to type DNS and no endpoints are specified, the host field will be used as the DNS name of the endpoint to route traffic to.

rshriram: cc @frankbu in case you have a more succinct way of  explaining this

networking/v1alpha3/service_entry.proto
```diff
@@ -339,18 +339,22 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
+  // In such scenarios, the IP addresses specified in
+  // the Addresses field or the port will be used to uniquely identify the destination.
   repeated string hosts = 1;
 
   // The virtual IP addresses associated with the service. Could be CIDR
-  // prefix. For HTTP services, the addresses field will be ignored and
+  // prefix. For HTTP(s) and TLS services, the addresses field will be ignored and
```
rshriram: only for HTTP. For all others, we do a CIDR match if the user provides a CIDR here..

networking/v1alpha3/service_entry.proto
```diff
@@ -339,18 +339,22 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
   // the destination will be identified based on the HTTP Host/Authority
-  // header. For non-HTTP protocols such as mongo/opaque TCP/HTTPS,
-  // the hosts will be ignored. If one or more IP addresses are specified,
+  // header or the Server Name Indication (SNI).
```
rshriram: I would just remove the for non-HTTP protocols such as mongo/opaque.. .. that line makes no sense here.
But keep the `If one or more IP addresses are

networking/v1alpha3/service_entry.proto
andraxylia: Updated, ptal.

PR #835, opened 2019-03-07
Changing "gateway" to "gateways"

tkishore1192:
Correcting gateway as gateways. If not changed, the yaml would not apply successfully.


tkishore1192:
Need to make changes to my cli account. For some strange reason it doesn't pick it up. Will do the necessary steps and create a PR. Apologies for the inconvenience.

PR #841, opened 2019-03-09
fix MeshNetworks demo

hzxuzhonghu:


mesh/v1alpha1/network.proto
```diff
@@ -35,7 +35,7 @@ message Network {
   //
   // 2. Explicitly:
   //
-  //    a. By matching the registry name with one of the "from_registries"
+  //    a. By matching the registry name with one of the "from_registry"
```
rshriram: fromRegistry (we refer to the YAML form in docs)

hzxuzhonghu: ACK

PR #844, opened 2019-03-10
Adding DestinationRule

tkishore1192:
Adding DestinationRule to the section of the page, which describes the use of "dedicated egress gateway through which all external service traffic is forwarded". Without the DestinationRule the example results in Status Code 503.


tkishore1192:
Need to resolve Google CLA issue to point to the right author name. Will re-open the PR once resolved. Thank you and apologies.

PR #845, opened 2019-03-10
adding DestinationRule

tkishore1192:
Adding DestinationRule to the section of the page, which describes the use of "dedicated egress gateway through which all external service traffic is forwarded". Without the DestinationRule the example results in Status Code 503.

### DIFF
--- a/networking/v1alpha3/service_entry.proto
+++ b/networking/v1alpha3/service_entry.proto
@@ -215,6 +215,7 @@ import "networking/v1alpha3/gateway.proto";
 //     route:
 //     - destination:
 //         host: istio-egressgateway.istio-system.svc.cluster.local
+//         subset: httpbin
 //   - match:
 //     - port: 80
 //       gateways:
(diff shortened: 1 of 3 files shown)


tkishore1192:
@andraxylia - Hello, could you please spare some time to review this PR? This PR comes under the same category of the Istio docs page where PR #835 changes have been made. And, 835 was reviewed and approved by you. Thank you!

networking/v1alpha3/service_entry.proto
```diff
@@ -215,6 +215,7 @@ import "networking/v1alpha3/gateway.proto";
 //     route:
 //     - destination:
 //         host: istio-egressgateway.istio-system.svc.cluster.local
+//         subset: httpbin
```
rshriram: why do you need the subset?

tkishore1192:
> I see no reason to add a subset and then add a dest rule for that subset.. it does not have any bearing on the 503s

The subset is there to route specifically to the egress pod, if we do proceed without the subset a "service unavailable status code occurs". The routing never gets fulfilled. When I was investigating this issue, found this document on the Istio examples page 
"https://istio.io/docs/examples/advanced-gateways/egress-gateway/".
Similar steps are mentioned on those docs.

geeknoid:
@rshriram Do we want to keep this?

rshriram:
nope. this can be closed

Issue #847, opened 2019-03-11
Invalid example in networking/v1alpha3/sidecar.proto


esnible:
An example Sidecar CRD in _networking/v1alpha3/sidecar.proto_ is invalid.

It's the example that starts at line 77.  It is generated into https://preliminary.istio.io/docs/reference/config/networking/v1alpha3/sidecar/ .  If I take that example and use _istioctl validate -f_ the error response is

```
* error validating resource (networking.istio.io/v1alpha3, Kind=Sidecar Name=default Namespace=): sidecar: the egress listener with empty port should be the last listener in the list
```

The same error appears in 1.1 RC 3 if I do
```
kubectl create ns prod-us1
kubectl apply -f /tmp/sidecar.yaml
```

Perhaps there is an extra `-` dash in the YaML?

esnible:
@geeknoid Can you give me permission to assign myself to Issues in this repo?

PR #849, opened 2019-03-11
Fix example Sidecar structure

esnible:
Resolves https://github.com/istio/api/issues/847

An example Sidecar CRD in networking/v1alpha3/sidecar.proto is invalid.

It's the example that starts at line 77. It is generated into https://preliminary.istio.io/docs/reference/config/networking/v1alpha3/sidecar/ . If I take that example and use istioctl validate -f the error response is

```
* error validating resource (networking.istio.io/v1alpha3, Kind=Sidecar Name=default Namespace=): sidecar: the egress listener with empty port should be the last listener in the list
```

### DIFF
--- a/networking/v1alpha3/sidecar.pb.go
+++ b/networking/v1alpha3/sidecar.pb.go
@@ -68,14 +68,14 @@
 //       name: somename
 //     defaultEndpoint: unix:///var/run/someuds.sock
 //   egress:
-//   - hosts:
-//     - "istio-system/*"
-//     port:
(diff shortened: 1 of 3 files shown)


mbanikazemi:
lgtm.

esnible:
@geeknoid This is a companion to the other fixes that were merged.  Should this target `master` or `release-1.1`?

geeknoid:
/lgtm
/approve

Please also copy this to the release-1.1 branch.

PR #860, opened 2019-03-14
Add support for inline role definition

pitlv2109:
Support inline role definition, essentially replace “roleRef” section in binding with the role definition itself.

[Context](https://docs.google.com/document/d/1diwa9oYVwmLtarXb-kfWPq-Ul3O2Ic3b8Y21Z3R8DzQ/#heading=h.wtcqkror4ke9)


pitlv2109:
RBAC v2 tracker: https://github.com/istio/istio/issues/12394

rbac/v1alpha1/rbac.proto
```diff
@@ -226,6 +226,30 @@ message ServiceRoleBinding {
   // $hide_from_docs
   // Indicates enforcement mode of the ServiceRoleBinding.
   EnforcementMode mode = 3;
+
+  // $hide_from_docs
+  // Support inline role definition, essentially replace “roleRef” section in binding
```
liminw: Let's call out that one of "roleRef" and "actions" should be used. 

Do not say "replace roleRef". Both "roleRef" (explicit role) and "actions" (inline role) are supported. Users can use inline roles if the role definition cannot be reused in other scenarios.

yangminzhu: Yes, roleRef is not to be replaced, both of them will exist at the same time, but shouldn't be used at the same time.

pitlv2109: It was taken out from the [RBAC v2 doc](https://docs.google.com/document/d/1diwa9oYVwmLtarXb-kfWPq-Ul3O2Ic3b8Y21Z3R8DzQ/edit#heading=h.wtcqkror4ke9), but let me change the wording a little bit. Thanks :)

pitlv2109: I updated the comment. Please take a look if it's good now. Thanks!

rbac/v1alpha1/rbac.proto
```diff
@@ -226,6 +226,32 @@ message ServiceRoleBinding {
   // $hide_from_docs
   // Indicates enforcement mode of the ServiceRoleBinding.
   EnforcementMode mode = 3;
+
+  // $hide_from_docs
+  // Support inline role definition, essentially give the user a choice to replace
+  // “roleRef” section in binding with the role definition itself. This does not
+  // mean that "roleRef" is deprecated. Both "actions" and "roleRef" are supported.
+  // However, note that only one of "actions" or "roleRef" can be used in a single binding.
```
liminw: Let me try to rephrase it :)

Inline role definition. An inline role is a role that is defined inside an authorization policy, instead of explicitly defined in a ServiceRole object. Inline roles can be used for the role definitions that are not intended to be reused in other bindings, while explicit roles are reusable. Both inline roles (defined in "actions" field) and explicit roles (defined in ServiceRole) are supported. Users should use only one of them in a single binding.

rbac/v1alpha1/rbac.proto
```diff
@@ -226,6 +226,32 @@ message ServiceRoleBinding {
...
+  //  namespace: ns1
+  // spec:
+  //  selector:
+  //    labels:
+  //      app: product
+  //  allow:
+  //  - subjects:
+  //    - names: ["cluster.local/ns/default/sa/frontend"]
+  //    actions:
+  //    - paths: [“/info/*”]
```
liminw: nit. Prefer to use exact match "/info".

rbac/v1alpha1/rbac.proto
pitlv2109: Updated. Thanks!

PR #861, opened 2019-03-15
ServiceEntry's Hosts shouldn't be required

ymesika:
Validation updated in https://github.com/istio/istio/pull/12497

/cc @hzxuzhonghu

### DIFF
--- a/networking/v1alpha3/service_entry.pb.go
+++ b/networking/v1alpha3/service_entry.pb.go
@@ -425,7 +425,7 @@ func (ServiceEntry_Resolution) EnumDescriptor() ([]byte, []int) {
 }
 
 type ServiceEntry struct {
-	// REQUIRED. The hosts associated with the ServiceEntry. Could be a DNS
+	// The hosts associated with the ServiceEntry. Could be a DNS
 	// name with wildcard prefix (external services only). DNS names in hosts
 	// will be ignored if the application accesses the service over non-HTTP
 	// protocols such as mongo/opaque TCP/HTTPS. In such scenarios, the
--- a/networking/v1alpha3/service_entry.pb.go
+++ b/networking/v1alpha3/service_entry.pb.go
@@ -42,8 +42,6 @@
 // metadata:
 //   name: external-svc-mongocluster
 // spec:
-//   hosts:
-//   - mymongodb.somedomain # not used
 //   addresses:
 //   - 192.192.192.192/24 # VIPs
 //   ports:
@@ -431,6 +429,8 @@ type ServiceEntry struct {
 	// protocols such as mongo/opaque TCP/HTTPS. In such scenarios, the
 	// IP addresses specified in the Addresses field or the port will be used
 	// to uniquely identify the destination.
+	// Note that the hosts and/or addresses field must be specified for
+	// a ServiceEntry to be valid.
 	Hosts []string `protobuf:"bytes,1,rep,name=hosts,proto3" json:"hosts,omitempty"`
 	// The virtual IP addresses associated with the service. Could be CIDR
 	// prefix. For HTTP services, the addresses field will be ignored and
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/service_entry.pb.go
```diff
@@ -425,7 +425,7 @@ func (ServiceEntry_Resolution) EnumDescriptor() ([]byte, []int) {
 }
 
 type ServiceEntry struct {
-	// REQUIRED. The hosts associated with the ServiceEntry. Could be a DNS
+	// The hosts associated with the ServiceEntry. Could be a DNS
```
nrjpoddar: @ymesika I would suggest adding a line at the end of both hosts & addresses fields:

`Note that the hosts and/or addresses field must be specified for a ServiceEntry to be valid.`

networking/v1alpha3/service_entry.proto
```diff
@@ -338,7 +338,7 @@ package istio.networking.v1alpha3;
 option go_package = "istio.io/api/networking/v1alpha3";
 
 message ServiceEntry {
-  // REQUIRED. The hosts associated with the ServiceEntry. Could be a DNS
```
nrjpoddar: @ymesika I would suggest adding a line at the end of both hosts & addresses fields:

Note that the hosts and/or addresses field must be specified for a ServiceEntry to be valid.

nrjpoddar: Additionally, one of the examples above should be updated:

The following configuration adds a set of MongoDB instances running on
// unmanaged VMs to Istio's registry, so that these services can be treated
// as any other service in the mesh. The associated DestinationRule is used
// to initiate mTLS connections to the database instances.
//
// ```yaml
// kind: ServiceEntry
// metadata:
//   name: external-svc-mongocluster
// spec:
//   hosts:
//   - mymongodb.somedomain # not used
//   addresses:
//   - 192.192.192.192/24 # VIPs
//   ports:
//   - number: 27018
//     name: mongodb
//     protocol: MONGO
//   location: MESH_INTERNAL
//   resolution: STATIC
//   endpoints:
//   - address: 2.2.2.2
//   - address: 3.3.3.3
// ```

Update this example to not mention `mymongodb.somedomain` and specify that only address is sufficient in this case.

ymesika: Thanks. Will update.

networking/v1alpha3/service_entry.proto
```diff
@@ -61,8 +61,6 @@ import "networking/v1alpha3/gateway.proto";
 // metadata:
 //   name: external-svc-mongocluster
 // spec:
-//   hosts:
-//   - mymongodb.somedomain # not used
```
nrjpoddar: Thanks @ymesika for addressing this. I added a comment earlier regarding:
```yaml
  endpoints:
   - address: 2.2.2.2
   - address: 3.3.3.3
```
 Do you think keeping endpoints in the example still makes sense?
Additionally, if you're removing endpoints then `resolution: STATIC` should be changed to `resolution: NONE`

ymesika: It creates a MongoDB service accessible to the mesh services using the virtual IP CIDR and routing the traffic to the VM's real IP addresses (2.2.2.2 and 3.3.3.3).
Why do you think it doesn't make sense?

nrjpoddar: I see. Your explanation makes sense. Thanks for clarifying!

hzxuzhonghu:
@rshriram https://github.com/istio/istio/pull/12497

>In addition to changes I went through all references to the ServiceEntry's Hosts and couldn't find a reason that it should be required.

ymesika:
@rshriram @hzxuzhonghu Do you mind showing me where in the code we require hosts? I couldn't find such.
If you are sure it's require you may want to revert istio/istio#12497

hzxuzhonghu:
look pilot/pkg/serviceregistry/external/conversion.go `convertServices`. if no hosts specified, the serviceEntry is ignored.

hzxuzhonghu:
Hosts is used to generate Service.Hostname.

rshriram:
Host is used to generate the cluster name as well.

nrjpoddar:
@rshriram all of this work was done for addressing https://github.com/istio/istio/issues/12493.
I'm advocating for re-opening this issue as it is not clearly stated anywhere why we need a bogus/unused host field in ServiceEntry if we are using the address directly. Additionally, I think the user experience is not great with the current API choice.

PR #862, opened 2019-03-15
Fix sample code in HTML reference docs

esimons:
Updating cookie matcher example to align with testdata: https://github.com/istio/istio/blob/master/tests/testdata/config/virtual-service-all.yaml#L18

Pasting here since diff is too big for preview:

```diff
  http:
  - match:
    - headers:
        cookie:
-         user: dev-123
+         regex: "^(.*?;)?(user=dev-123)(;.*)?"
```

### DIFF
--- a/networking/v1alpha3/istio.networking.v1alpha3.pb.html
+++ b/networking/v1alpha3/istio.networking.v1alpha3.pb.html
@@ -1166,7 +1166,7 @@ spec:
   - match:
     - headers:
         cookie:
-          user: dev-123
+          regex: "^(.*?;)?(user=dev-123)(;.*)?"
     route:
     - destination:
         port:


esimons:
@venilnoronha it doesn't appear that this is a generated file. As a sanity check, I just did a project search for what's being replaced and came up empty.

venilnoronha:
It's probably being generated from the following line.

https://github.com/istio/api/blob/f5ccf50ad5a13485bd13b3c0e7a4ee0b6462842d/networking/v1alpha3/gateway.proto#L120

esimons:
After doing a bit of digging, it seems like this file should actually be deleted entirely following https://github.com/istio/api/pull/811, since it's no longer generated via makefile and appears to have been broken down into multiple pages.
Perhaps @geeknoid @rshriram can confirm?

geeknoid:
Replaced by #866 which deletes the file.

PR #869, opened 2019-03-20
ns of egress gateway should be istio-system in the context

idouba:
ns of egress gateway should be istio-system in the context


linsun:
This is not strictly necessary. you can put ingress and egress onto its own namespace.

idouba:
@linsun yes, Gateway can be in their own ns .
In the example context
```
  http:
  - match:
    - port: 80
      gateways:
      - mesh
    route:
    - destination:
        host: istio-egressgateway.istio-system.svc.cluster.local
```
vs using the egress-gateway in **istio-system**.

PR #870, opened 2019-03-20
update host desc that a host can be defined using more than one vs

idouba:
the traffic properties of a host can be defined using more than one VirtualService
https://github.com/istio/istio/issues/12590


linsun:
Thanks, LG to me.  Defer to @frankbu @rshriram to approve this.

frankbu:
Thanks @idouba, but this needs to be sent to the `release-1.1` branch, so we can get it into the 1.1 docs.

idouba:
@frankbu  release 1.1 pls refer to #873

PR #871, opened 2019-03-20
Gateway selector doc correction

frankbu:


andraxylia:
Are we saying the selector scope is no longer limited to the namespace, but system-wide? Is this what is currently implemented or desired change? There are similar selectors across the API that are limited to the namespace. There is also a roadmap item to bring consistency in the namespace handling across all APIs.

cc @louiscryan @costinm

frankbu:
The documentation is currently incorrect. This is just fixing the document to describe what is currently implemented in 1.1. It is the same as 1.0.

rshriram:
this is doc fix. no behavior change since 0.8

louiscryan:
I understand that this is current behavior but what was previously documented is closer to desired behavior. We don't really want Gateways to define behavior for namespaces in which they are not defined as this at worst presents a security risk and at best violates administrator expectations

andraxylia:
We also should consider the api as the source of truth and fix the code when it does not match the API.

If the desired behavior is what was documented, let's revert this fix which has already been merged and open an issue against the code.

frankbu:
In 1.0 the documentation said:
```
  // on which this gateway configuration should be applied.
  // The scope of label search is platform dependent.
  // On Kubernetes, for example, the scope includes pods running in
  // all reachable namespaces.
  map<string, string> selector = 2;
```
So changing it in 1.1 to say that the Gateway and gateway workload MUST be in the same namespace would be a breaking change that will definitely break all current samples and any users that are using the built-in ingress or egress gateways in `istio-system`.

I think that for now the best we can so is add one more sentence, something like this:
```
  // It is recommended that the Gateway resource reside in the same namespace as the gateway workload.
```
and then we can figure out how to make the breaking change in the future.

linsun:
Clarification q - Is product page considered gateway workload or istio-ingressgateway considered a gateway workload?

BTW @frankbu what you suggested is exactly what we had in the upgrade notice:

https://istio.io/docs/setup/kubernetes/upgrade/notice/#traffic-management

```
We recommend storing gateway resources in the same namespace as the gateway workload (e.g., istio-system in case of istio-ingressgateway). When referring to gateway resources in virtual services, use the namespace/name format instead of using name.namespace.svc.cluster.local.
```
And you are right, our samples aren't using the recommended approach.

louiscryan:
I think Lin's verbiage is appropriate perhaps making the recommendation a little stronger signalling that we may move to making it a requirement in the future.

PR #872, opened 2019-03-20
Merge release-1.1 into master

howardjohn:
When approving:
> don't do regular lgtm/merge. an admin needs to do merge commit, which we don't see as an option


louiscryan:
This is a merge PR from commits to repo that require CLA. Setting CLA to yes

PR #874, opened 2019-03-21
update host desc that more than one vs share same host

idouba:
update host desc that more than one vs share same host to permit the ingress style gateway, that is two service share the same domain name and delegated by path.
#870 #873
istio/istio#12590


frankbu:
I don't think this one is needed. The fix in #873 should get cherry-picked into master at some point. At least that's the process for istio/istio.io. @rshriram can you confirm?

PR #876, opened 2019-03-22
Be precise about 'workload' vs 'workload instance'

esnible:
Resolves https://github.com/istio/istio.io/issues/3686

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -201,7 +201,7 @@ message Gateway {
   // on which this gateway configuration should be applied. The scope of
   // label search is restricted to the configuration namespace in which the
   // the resource is present. In other words, the Gateway resource must
-  // reside in the same namespace as the gateway workload.
+  // reside in the same namespace as the gateway workload instance.
   map<string, string> selector = 2;
 }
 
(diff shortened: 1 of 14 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -201,7 +201,7 @@ message Gateway {
   // on which this gateway configuration should be applied. The scope of
   // label search is restricted to the configuration namespace in which the
   // the resource is present. In other words, the Gateway resource must
-  // reside in the same namespace as the gateway workload.
+  // reside in the same namespace as the gateway workload instance.
```
frankbu: This should be `instances`. There may be more than one instance of the selected gateway deployment.

That said, why was this change merged into `master`, instead of `release-1.1`?
This particular change, in fact, is already going to have a merge conflict with a change I made at this same place  in `release-1.1`.

@geeknoid What's the process for doc changes for reference docs in istio/api? I was thinking that we were going to do the same as in istio/istio.io, i.e., for the first 3 weeks we should make doc fixes only in the `release-1.1` branch, and at the end of that period, they will get merged back into master, after which all changes needed in release 1.1 will need a PR in both branches.

PR #879, opened 2019-03-25
add http_idle_timeout to meshConfig

vadimi:
Based on https://discuss.istio.io/t/istio-ingress-and-aws-alb-idle-timeout/1455/5 and https://github.com/istio/istio/issues/12533 `http_idle_timeout` will allow to better integrate istio ingress gateway with other load balancers/proxies.


vadimi:
I signed it!

mesh/v1alpha1/config.proto
```diff
@@ -282,8 +282,11 @@ message MeshConfig {
   // Configures DNS refresh rate for Envoy clusters of type STRICT_DNS
   google.protobuf.Duration dns_refresh_rate = 36;
 
+  // Configures mesh proxy idle timeout
+  google.protobuf.Duration http_idle_timeout = 37;
```
hzxuzhonghu: Add more comments, what does it mean?

vadimi: I updated comments

rshriram:
compare with https://github.com/istio/api/pull/891 .. and please draw out the differences in the documentation so that end users clearly understand the value of the global config vs dest rule level config

vadimi:
@rshriram I updated docs. Also I was under assumption that DestinationRules handle upstream level connections only. Current value will be used by [HttpConnectionManager](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto.html?highlight=idle_timeout) that will handle downstream connections - this is what I'm experiencing issues with. If my assumption is correct I will add it to docs too.

mesh/v1alpha1/config.proto
```diff
@@ -282,8 +282,16 @@ message MeshConfig {
   // Configures DNS refresh rate for Envoy clusters of type STRICT_DNS
   google.protobuf.Duration dns_refresh_rate = 36;
 
+  // Configures idle timeout for edge and internal Envoy proxies
+  // by using Envoy HTTP connection manager `idle_timeout` property 
+  // to gracefully close idle connections. Comparing to DestinationRule level 
+  // `idle_timeout` the one in meshConfig allows to configure only one value for
+  // entire cluster without configuring individual rules.
+  // There is no default value set for this timeout.
+  google.protobuf.Duration http_idle_timeout = 37;
```
rshriram: this is more about the timeout imposed by the server on the inbound path. Dest rule is by client isnt it?

hzxuzhonghu: I think this is to set the  listener `HttpConnectionManager.IdleTimeout`, both inbound and outbound

That in DR is to set cluster.CommonHttpProtocolOptions.IdleTimeout   , function both on inbound and outbound I think.

vadimi: I just retested `common_http_protocol_options` on vanilla envoy using Wireshark and it looks like it doesn't handle downstream connections, only upstream ones. Idle timeout on `HttpConnectionManager` handles the downstream connections.

costinm:
Once again, it makes zero sense to gave a setting like this as a global.
Add it to destination rule or sidecar if needed, but please stop adding
mesh globals. Meshconfig is not user editable - so all options are really
installer options, and we have enough of them

vadimi:
I would really appreciate some more guidance here. I need to be able to configure `HttpConnectionManager.IdleTimeout` at least on the ingressgateway as this is what I'm having issues with (see https://github.com/istio/istio/issues/12533). `HttpConnectionManager.IdleTimeout` is a global setting, but DestinationRules and Sidecars control individual clusters. What would be the right and acceptable solution here?

costinm:
If you want to control an option based on the workload - use Sidecar. For
example to specify that all calls from a gateway or from a specific
deployment will have a set timeout.

If you want by destination - all calls to a specific service -
DestinationRule

Once this is added, defaults for the entire mesh use the same proto and the
new istio config dir.

rshriram:
If you want this only for the gateway, wait for #897 that adds the x-forwarded-for num trusted hops under http options. You can set the idle timeout under the same thing for the gateway.

vadimi:
yes, I only need it for the gateway, I'll wait then, thanks.

rshriram:
I think given https://github.com/istio/istio/issues/13355 we definitely need this. And it turns out that this is solely an install time option, setting the TCP connection idle timeout. For reference, we have similar stuff in mesh config like TCP connection connect timeout, dns refresh rate, etc. These settings are applicable for all workloads in different parts of envoy config (http connection manager, tcp proxy, and other places). Sidecar is not the right place for this as it would have to be defined for every namespace. Secondly, the user would also have to use some hack to get this into the gateway. The envoy filter merge PR I have might help, but it makes no sense when such common settings have to be applied to every proxy.

In the implementation, we can always override such things using node metadata if people want more customizability, just like the TLS cert path metadata that overrides the mesh config, or the SDS stuff that has both globals and per proxy overrides if need be.

costinm:
We do have ability to set a default value for Sidecar.

Things like timeout, etc can't be global settings - it doesn't make any
sense, each app has different needs.
Global settings have immediate impact on the entire mesh and as such are
rarely safe to change in
a production environment, and lack flexibility.

vadimi:
I'm closing this PR in favor of https://github.com/istio/istio/pull/13515 which works perfectly fine for my use case.

PR #882, opened 2019-03-26
Add extensible AuthnMethod for Istio Authn policy

lei-tang:
Add extensible AuthnMethod for Istio Authn policy, which can be extended by adding new authentication methods and reused by multiple Istio authn policies.


geeknoid:
@lei-tang Do we still need this PR?

lei-tang:
@geeknoid The PR is closed.

PR #883, opened 2019-03-26
Add configuration for matching on query params

venilnoronha:
This adds the necessary configuration for matching over query parameters. See istio/istio#12782 for more information.


venilnoronha:
@rshriram I've fixed the conflict.

PR #884, opened 2019-03-27
add sds related param to proxyconfig proto

quanjielin:
https://github.com/istio/istio/issues/5891

It will be used for control plane sds(asked from https://github.com/istio/istio/issues/11434) , the params will be passed param to sidecar/pilot/mixer and used in their bootstrap files.


costinm:
Was this reviewed by TOC or anyone working on operator / installer / etc ? 

Please revert - it goes against all the plans we have for modular config, upgrade, etc. Mesh config has only settings that are applied global.

costinm:
@duderino

quanjielin:
costin, this PR isn't on meshconfig, but proxyconfig; the two params are used in pilot-agent to construct bootstrap file for sidecar/pilot/mixer for controlplane sds, also later in per-pod rollout(sds migration). 

/cc @wenchenglu

PR #885, opened 2019-03-29
Add role to ServiceRoleBinding API to essentially replace roleRef later

pitlv2109:
Add `role` to `ServiceRoleBinding` API to essentially replace `roleRef` later.


rbac/v1alpha1/rbac.proto
```diff
@@ -254,6 +254,16 @@ message ServiceRoleBinding {
...
+
+  // $hide_from_docs
+  // A `role` inside a ServiceRoleBinding refers to the ServiceRole that this
+  // ServiceRoleBinding binds to. The value of `role` is the name of the ServiceRole,
+  // and it can start with or without a forward slash ("/").
+  // When a `role` starts with "/", e.g. "/service-viewer", it means that this ServiceRoleBinding
+  // refers to the ServiceRole in the configurable Istio root namespace.
+  // When a `role` starts without "/", this ServiceRoleBinding refers to the ServiceRole in the
+  // same namespace as the AuthorizationPolicy's, which contains said ServiceRoleBinding.
+  string role = 5;
```
yangminzhu: Why not just reuse the `roleRef` field?

pitlv2109: I think it's verbose, because the kind is always ServiceRole? On the other hand, using `roleRef` would not or require less migration changes. @liminw Thoughts?

yangminzhu: Deprecating a field requires a lot of migration work. For customer already using `roleRef`, we need to upgrade it to the new field which is very painful  and easy to make mistake. So if it's possible to reuse, I would prefer to reuse.

liminw: We are trying to align with Sven's Istio config model proposal, and use the form "namespace/name" to refer to an object. For root namespace, it is configurable and we can have multiple of them in case of mesh federation. So we chose to use a leading '/' to represent the ServiceRole under root namespace. 

@smawson @louiscryan

rbac/v1alpha1/rbac.proto
```diff
@@ -254,6 +254,16 @@ message ServiceRoleBinding {
   //      methods: ["GET"]
   // Required. The set of access rules (permissions) that the role has.
   repeated AccessRule actions = 4;
+
+  // $hide_from_docs
+  // A `role` inside a ServiceRoleBinding refers to the ServiceRole that this
+  // ServiceRoleBinding binds to. The value of `role` is the name of the ServiceRole,
```
liminw: Add: A ServiceRoleBinding can bind to a ServiceRole in the same namespace or root namespace. A ServiceRole in the root namespace represents a mesh global ServiceRole.

pitlv2109: Done.

rbac/v1alpha1/rbac.proto
yangminzhu: I know, my point is we can reuse `roleRef` to achieve this goal. it's just a string field and we can support the new semantics without problem. All the new stuff you mentioned can be supported right in the `roleRef` field, then all we need to do is stop using the `kind` field which is much easier to do.

liminw: @yangminzhu I am totally with you on the point that if we can reuse a field, do not introduce a new one. However, if we use `roleRef` here, the field will be quite verbose since you will write config like following:

```
roleRef:
   name: roleA
```

while if we have a new `role` field, we do not need the redundant `name` field. 
```
role: roleA
```

pitlv2109: Have we reached an agreement yet?

yangminzhu: I'm fine with adding a new field, but please make sure the old field is covered during the API migration.

liminw:
/lgtm
/approve
We may need further discussion on how to refer to root namespace (global role), but the semantics of referring to a role in the same namespace should be fine.

pitlv2109:
This is for https://github.com/istio/istio/issues/12394

PR #886, opened 2019-03-29
Correct percentage value to match 1 out of 1000 docs value

andor-pierdelacabeza:
Both for HTTPFaultInjection.Abort and HTTPFaultInjection.Delay, the docs give an example for a "1 out of every 1000 requests".

The value given in the example to the 'percentage' field is '0.001', but the field 'percentage' is a type 'Percent' which goes from 0 to 100, so '0.001' would mean "1 out of 100000"

I've corrected the value from '0.001' to '0.1' to match "1 out of 1000", was either that or changing the docs to "1 out of 100000", but seemed like a too big number to me ;)


andor-pierdelacabeza:
:thinking:

PR #887, opened 2019-03-29
Fixes istio/istio#12873. Add field Sidecar.OutboundTrafficPolicy to o…

robertpanzer:
…verride outbound traffic policy per sidecar

This PR tries to solve https://github.com/istio/istio/12873.
For that it adds a field Sidecar.OutboundTrafficPolicy that is currently a copy of the same field in the Mesh configuration.

I actually wanted to reuse that type and move it to network/v1alpha3, but the build has another opinion on this and fails.
Would it work to move the type mesh/v1alpha1.MeshConfig_OutboundTrafficPolicy to network/v1alpha3/OutboundTrafficPolicy?
In my understanding this should result in the same Wire format, so that it should still work to run a newer Galley with an older Pilot or vice versa.
Are there any opinions on this?
And if it would be ok to move this, how can I do it? Currently make fails when moving this type.


networking/v1alpha3/sidecar.proto
```diff
@@ -320,6 +327,31 @@ message WorkloadSelector {
...
+  enum Mode {
+    // outbound traffic will be restricted to services defined in the
+    // service registry as well as those defined through ServiceEntries
+    REGISTRY_ONLY = 0;
+    // outbound traffic to unknown destinations will be allowed, in case
+    // there are no services or ServiceEntries for the destination port
+    ALLOW_ANY = 1;
+
+    reserved 2;
+    reserved "VIRTUAL_SERVICE_ONLY";
```
rshriram: you dont need these two here.. since you are copying the struct

rshriram: I dont think you can remove the struct from the mesh/v1alpha1 without causing some form of circular dependency thing IIRC.. though @ozevren would have to attest to that. For the purpose of this PR though, I would suggest that you remove the reserved 2 and the reserved VIRTUAL.. line as this is an entirely new message introduced in this file..

robertpanzer: Ok, I'll do that. Thanks for the clarification.

My original intent was to have this struct only in networking.v1alpha3 and remove the one in mesh.v1alpha1.
There is already one dependency in that direction; adding a reference to mesh.v1alpha1.OutboundTrafficPolicy from networking.v1alpha3 would create a cyclic dependency.

robertpanzer: @rshriram I removed the reserved field and updated the PR by amending the commit (To avoid the breaking change in proto.lock in the commit history).

costinm:
Please revert and move to master - AFAIK we are not supposed to make API changes in 1.1 ( or it needs to go trough release manager, review, etc ) 

release-1.1 branch is released every week - so the TOC review can be too late.

costinm:
@duderino

costinm:
On the content of the API change: I don't think this is the best way, it has multiple problems:
- like public/private, it is too narrow
- not aligned with 'egress' and 'exportTo' - where for internal traffic we provide a list
- not aligned with ServiceEntry - which is normally used to control egress, and has different syntax

It is also quite confusing - what is 'outbound traffic' ? 
- Egress is used in most places in the docs. What is different between egress and outbound traffic and why a new term used here ?
- traffic to other Istio services is also outbond traffic. Is it affected ? Why ? Why not ? 
- what happens when we want to allow specific ports, or http hosts ?

robertpanzer:
Sorry for creating this.
OutboundTrafficPolicy is already part of Istio's Mesh configuration and I agree that it is confusing.
But all that configuration does is to allow application owners to refine this configuration for their workloads.

I think though that the term "external" refers to MESH_EXTERNAL.

PR #888, opened 2019-03-29
Update RBAC API to align with doc on the Concept page.

pitlv2109:
The [Concept page](https://istio.io/docs/concepts/security/#authorization) introduces `PERMISSIVE` mode. We should remove `$hide_from_docs` to update the [Reference page](https://istio.io/docs/reference/config/authorization/istio.rbac.v1alpha1/).

### DIFF
--- a/rbac/v1alpha1/rbac.proto
+++ b/rbac/v1alpha1/rbac.proto
@@ -223,7 +223,6 @@ message ServiceRoleBinding {
   // Required. Reference to the ServiceRole object.
   RoleRef roleRef = 2;
 
-  // $hide_from_docs
   // Indicates enforcement mode of the ServiceRoleBinding.
   EnforcementMode mode = 3;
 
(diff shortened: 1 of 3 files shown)


pitlv2109:
I think we should also change `mode` in `ServiceRoleBinding` to `enforcement_mode` to align with `ClusterConfig` (or `ClusterRbacConfig`).

rbac/v1alpha1/rbac.proto
```diff
@@ -223,7 +223,6 @@ message ServiceRoleBinding {
   // Required. Reference to the ServiceRole object.
   RoleRef roleRef = 2;
 
-  // $hide_from_docs
```
liminw: My concern is, we will likely change this field in the near future. I actually prefer hiding it for now, or at least clearly call out that do not use it since it is likely to change it soon.

rbac/v1alpha1/rbac.proto
```diff
@@ -382,7 +381,6 @@ message RbacConfig {
   // effect only when mode is ON_WITH_EXCLUSION and will be ignored for any other modes.
   Target exclusion = 3;
 
-  // $hide_from_docs
```
liminw: Same here.

rbac/v1alpha1/rbac.proto
pitlv2109: Sounds good.

pitlv2109:
/close 
Will update once the API is stable.

PR #891, opened 2019-04-04
Adding idle_timeout field to DestinationRule

arianmotamedi:
Per https://github.com/istio/istio/issues/9113, updating DestinationRule to support an idle_timeout value for upstream http connections. This corresponds to Envoy's HttpProtocolOptions in the cluster definition: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto#envoy-api-msg-core-httpprotocoloptions.


linsun:
DR seems the right place for this.

Any objection @frankbu @rshriram and others?

Also, can we get this in release-1.1 since multiple users have reported this and it is preventing them from using istio in prod:  https://github.com/istio/istio/issues/9113

PR #893, opened 2019-04-05
Adding not_ready_endpoints to ServiceEntry

nmittler:
This is needed to allow Pilot to generate inbound listeners for the sidecars, which are for readiness checks.

In support of #10589


nmittler:
cc @costinm @andraxylia @nino-k

networking/v1alpha3/service_entry.proto
```diff
@@ -509,4 +510,8 @@ message ServiceEntry {
   // If specified, the proxy will verify that the server
   // certificate's subject alternate name matches one of the specified values.
   repeated string subject_alt_names = 8;
+
+  // One or more endpoints associated with the service, which are not yet
+  // ready to handle traffic.
+  repeated Endpoint not_ready_endpoints = 9;
```
andraxylia: Do we have a design for handling readiness? I thought we would pre-download a listener that sniffs.

I am not convinced adding not ready endpoints is the right approach. Galley should only send endpoints that are ready. Readiness is a k8s notion that should be moved from Pilot to Galley.

nmittler: From our offline discussion...

The "sniffing" listener will require an Envoy change and we have no idea when that will land. The k8s API already exposes this information and it's reasonable that we just do the same. Once the Envoy changes land Pilot may eventually ignore this information, which is fine. Some non-Pilot-based systems, however, still may find this information useful.

andraxylia: Ack, this is a practical transition to moving the readiness to Galley.

costinm:
/hold
This doesn't make sense - it's an internal detail of MCP/Pilot interaction, should not be exposed to all users of the public API, only creates confusion.

Besides this seems k8s specific, and there are other ways to address this without API changes - the MCP protocol can send the 'not ready' in separate message, or with annotations, etc.

costinm:
Besides the fundamental problem is that 'unready' state is extremely time sensitive - I would like to see e2e tests that confirm pod startup is not affected by the increased latency. 

And long term using HDS is the clearly better solution - we'll be stuck for a very long time with the Istio API, it doesn't make sense to add user visible APIs when we know this will be deprecated as soon as HDS is on.

nmittler:
@costinm 

> there are other ways to address this without API changes - the MCP protocol can send the 'not ready' in separate message, or with annotations, etc.

A separate message would require yet another proto, and this would effectively become part of the Galley API, even if it isn't part of our configuration API. The same would be true for annotations. Are you saying we care less about the Galley API in this case? If so, then that's fine.

nmittler:
Discussed offline. We're in agreement to use an annotation on the synthetic service entry for this purpose for now.

PR #897, opened 2019-04-09
Enable HTTP options in the Gateway.

julianvmodesto:
Allow configuring HTTP options including use_remote_address and
xff_num_trusted_hops for Gateways with different internal or external
positioning.

Addresses https://github.com/istio/istio/issues/7679.
Follow up from https://github.com/istio/istio/pull/13157.

Envoy defaults use_remote_address=false and xff_num_trusted_hops=0,
while Istio ingress gateways are configured with
defaults use_remote_address=true and xff_num_trusted_hops=0.

See the following Envoy documentation on X-Forwarded-For for reference.
https://www.envoyproxy.io/docs/envoy/latest/configuration/http_conn_man/headers#x-forwarded-for


julianvmodesto:
cc @rshriram

networking/v1alpha3/gateway.proto
```diff
@@ -414,6 +414,20 @@ message Server {
   // be forwarded to by default. Format should be `127.0.0.1:PORT` or
   // `unix:///path/to/socket` or `unix://@foobar` (Linux abstract namespace).
   string default_endpoint = 5;
+
+  message HTTPOptions {
+    // If set to true, the gateway will use the real remote address of the
+    // client connection when determining internal versus external origin
+    // and manipulating various headers.
```
rshriram: You would have to massage this a bit instead of using envoy docs as is. is this a proxy for setting X-Forwarded-For ? if so, why not name it 'enable_x_forwarded_for' ?

rshriram: Also state the defaults.. can I specify xff_num_trusted_hops without specifying use_remote_address? I dont think so. May be this is better framed as
```
httpOptions:
  xForwardedFor:
    enabled: true
    trustedHops: 4
```

julianvmodesto: Hmm after investigating `use_remote_address`, it not only affects the header X-Forwarded-For, but it also affects the trusted client address.

For now, I'm going to remove this from httpOptions because the existing ingress gateway behavior to set `use_remote_address=true` actually makes sense and I cannot reason about an instance where setting `use_remote_address=false` for an ingress gateway is actually useful.

julianvmodesto: Done -- I've gone w/ the following:

```
httpOptions:
  xForwardedFor:
    trustedHops: 4
```

julianvmodesto:
@rshriram could we proceed w/ this api change?

The merge semantics in EnvoyFilter https://github.com/istio/api/pull/899 are helpful, but this gateway change seems like it would fit because for ingress gateways run on clouds, it's often the case that there's at least 1 external cloud load balancer (GCP LB or AWS ALB), and possibly internal load balancers in front of the ingress gateway.

This api change helps w/ setting the trusted client remote IP address, as well as the x-forwarded-for header (and the x-envoy-external-address and x-envoy-internal headers), so that the trusted client remote IP address and/or x-forwarded-for addresses can be used for Istio traffic management (consistent hash load balancing based on IP address or XFF header, or other types of routing based on IP ranges) or Istio policies (blacklist/whitelist IP ranges).

rshriram:
Here is the problem with this api change: it is applicable at http connection manager level. For gateway servers using http ports, a single http connection manager (on port 80) represents multiple servers. Sh how do you merge conflicting settings from different http servers? 

This is not a problem with https because we create a separate http connection manager for each server in the gateway.

If you state in the comments that this setting applies only for https servers, then we are good.

Also in the comments, please add the nice explanation you provided in your comment above. It would be helpful if you even add an example configuration in the comments on top of the server.

In the implementation pr, you have to add validation logic that will reject xff settings for http servers.

julianvmodesto:
Ah thanks, this limitation is good to know! 

This setting for https would be helpful since SSL termination at the load balancer is a common pattern. But ideally this setting is available for http, too.

This is interesting! Is there an issue where this problem is tracked yet? I'm interested in following it.

Since this limitation likely affects other features, what's the future strategy to address this?

Is the plan to change the architecture such that multiple servers are used instead of one, like for https?

Given that this limitation affects other http options, is it sufficient to introduce the `httpOptions` anyway and is it possible to have a gateway validation hook to prevent setting httpOptions targeting the same gateway proxy deployment?

rshriram:
Its complicated. In the beginning, we allowed only one gateway spec for every gateway deployment, where you had to put all the http hosts under one server, and https hosts in different servers. Then all users complained that that single gateway was becoming the bottleneck. So we then started allowing multiple gateway specs for ingress gateway. In the code, we merge these gateways [https://github.com/istio/istio/blob/master/pilot/pkg/model/gateway.go#L49] such that the merging code emits all servers bound to the same gateway workload (identified by the selector label).

And then there is Envoy limitation. The way envoy is designed, you can have only one HTTP connection manager inside a plain text listener for a given IP:Port. Within this single HTTP connection manager, yu can have any number of virtual hosts, but XFF options are per Http connection manager and not per vhost. So, for HTTP servers, the existing gateway definition does not allow customizability on a Gateway workload basis. It only allows customizability per virtual host basis.

Hence my suggestion for doing this at envoyfilter level, which while clunky will work for both HTTP and HTTPS

networking/v1alpha3/gateway.proto
```diff
@@ -414,6 +414,19 @@ message Server {
...
+
+  message HTTPOptions {
+    message XForwardedFor {
+      // The number of additional ingress proxy hops from the right side of the
+      // X-Xorwarded-For HTTP header to trust when determining the origin client’s
+      // IP address.
+      uint32 trusted_hops = 1;
+    }
+
+    XForwardedFor x_forwarded_for = 1;
```
costinm: Comments, examples.

networking/v1alpha3/gateway.proto
```diff
@@ -414,6 +414,19 @@ message Server {
   // be forwarded to by default. Format should be `127.0.0.1:PORT` or
   // `unix:///path/to/socket` or `unix://@foobar` (Linux abstract namespace).
   string default_endpoint = 5;
+
+  message HTTPOptions {
+    message XForwardedFor {
+      // The number of additional ingress proxy hops from the right side of the
+      // X-Xorwarded-For HTTP header to trust when determining the origin client’s
+      // IP address.
```
costinm: Please add more info - there are security implications, topology ( LB ), etc.

julianvmodesto:
> So we then started allowing multiple gateway specs for ingress gateway. In the code, we merge these gateways [https://github.com/istio/istio/blob/master/pilot/pkg/model/gateway.go#L49] such that the merging code emits all servers bound to the same gateway workload (identified by the selector label).

The effects of merging multiple Gateway specs for the same ingress gateway Deployment seems acceptable for users who do this, so it seems like this type of option is also acceptable w/ documentation + warnings that this type of option has effects for ingress gateway Deployments w/ multiple Gateway specs.

Similarly, for example, overlapping hosts (e.g. `*` wildcard hosts) in multiple Gateway definitions for the same Deployment can have messy merge behavior, and users should accept this type of limitation when defining multiple Gateway specs for a single Deployment.

> So if you have one HTTP server with http options but the other server on same port does not have those options or has conflicting options, then when we build the HTTP connection manager, we have no clean way of merging both.

It seems like for these http options, we could have them under `.spec` instead of `.spec.servers[]` while we have this Envoy limitation:

```yaml
kind: Gateway
metadata:
  name: my-gateway
spec:
  selector:
    app: my-gateway-controller
  httpOptions:
    ...
    xff_num_trusted_hops: 2
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - '*.app.com'
```

networking/v1alpha3/gateway.proto
julianvmodesto: Added more info, as well as a link to the Envoy reference doc

rshriram:
@julianvmodesto please checkout the new envoy filter in master branch or inn one of our daily releases. I believe it should satisfy your use case. Closing this until then..

fcantournet:
@rshriram Should we take closing this issue as "use an envoyfilter" ?
Because  right now EnvoyFilters are accompanied by a big warning like this :
```Since this is break glass configuration, there will not be any backward compatibility across different Istio releases. In other words, this configuration is subject to change based on internal implementation of Istio networking subsystem.```

I'm not sure I agree that setting the istio-ingressgateway to trust the http cloud load-balancer in front of it should require a "break glass situation".

fcantournet:
So regardless of the situation of envoy filters in terms of stability / support, I don't think it is possible to change `use_remote_address` with envoy filters.
I don't know where in the merging process it fails, but I cannot override it to `false` on my ingress-gateway using an envoy filter.

PR #899, opened 2019-04-12
Enhancing EnvoyFilter with listener/cluster/route mods

rshriram:
Rather than constantly keep adding more fields to the API in HCM or other areas, this PR allows users to override all aspects of the generated config using JSON xpath matches. See the proto file for examples.


mandarjog:
@rshriram This is great.
We need a similar one for upstream configuration.  `destinationRule` and upstream have a similar relationship.

The other question is regarding the future `EnvoyFilter` resource itself. The `Sidecar` resource is also bound to a sidecar, and this could be subsumed in it.

rshriram:
Envoy filter applies to Gateways as well. In fact, its mostly used with Gateways

julianvmodesto:
Neat! This type of merge EnvoyFilter would fulfill my use case

julianvmodesto:
> In fact, its mostly used with Gateways

That is interesting! From an istio user experience perspective, [the disclaimers at the top of the EnvoyFilter doc](https://istio.io/docs/reference/config/networking/v1alpha3/envoy-filter/) say EnvoyFilters are break-glass configs, so i'm curious how the Gateway + EnvoyFilter APIs will eventually align e.g. will `.tls.httpsRedirect` be deprecated or will certain EnvoyFilters become part of Gateway options

costinm:
I think having Gateway + Sidecar and EnvoyFilter is too much complexity. I
would rather see this added to
Gateway and Sidecar - and possibly replacing the EnvoyFilter with a similar
'patch' mechanism.

I agree that DestinationRule will also need such a 'break glass' - for
clusters.

There are few other inconsistencies - "workloadLabels" vs
"workloadSelector/labels" in Sidecar and
"selector" in Gateway - I would rather deprecate EnvoyFilter and move it to
Gateway and
Sidecar, which would simplify other things.

But I love the idea - for Installer we are also exploring a 'patch' style
using Kustomize, it
would be great to explore if for consistency we can use the same syntax.
After all the problem
is the same.

rshriram:
I am not against folding the envoy filter crd into gateway/sidecar but thats a separate issue from this. Will start one for that.

This one deals solely with path/merge semantics in addition to the existing insert filter semantics.
@julianvmodesto as you stated, frequently used settings will be folded into the Gateway. I could not fold the xforwarded for or idle timeout setting into Gateway because they are specific to HTTP connection manager and not to every virtual host. Every server in a gateway (where such settings are added) applies to a virtual host roughly. Hence this alternative.

@ozevren any backward compatibility or consistency concerns? otherwise, I would like to merge this and unblock our end users @julianvmodesto and @vadimi who have two independent PRs for functionality that would otherwise be solved by this.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -163,8 +189,21 @@ message EnvoyFilter {
...
 
-    // Insert position in the filter chain. Defaults to FIRST
-    InsertPosition insert_position = 2;
+    oneof op {
+      // Insert position in the filter chain. Defaults to FIRST.
+      InsertPosition insert_position = 2;
+
+      // If a filter by the same name exists in the filter chain, it
+      // will be replaced.
+      bool overwrite = 6;
```
ozevren: While I don't think there is a backward compat issue here, I'd very much prefer if we can avoid oneofs. They tend to create problems with serialization. Also, I think this is prone to create bizarre combinations. (i.e. I can write overwrite = false, or merge = false as valid values).

It looks like modeling this as a separate field as enums might be a cleaner approach?

```
  InsertPosition insert_position = 2;
  MergeMode merge_mode = 6;

  enum MergeMode {
     // The existing behvior applies
     NONE = 0;
     MERGE = 1;
     OVERWRITE = 2;
  } 
}

```

rshriram: we cannot specify merge and insert position as they are mutually exclusive semantically. or overwrite and insert position. In case of merge/overwrite, we update the existing filter whose position is already known. So the enum example you have will create even more strange combinations IMO like (insertPosition: last, mergeMode: MERGE) ..

mandarjog:
Can we move this along quicker? @lryan @constinm 
We either need this or model some hcm options separately.

louiscryan:
I would prefer to use something like a combination of the new structpath (JSONPath-like)[https://goessner.net/articles/JsonPath/index.html#e2] syntax and a limited set of mutation operators.

This is effectively equivalent to https://tools.ietf.org/html/rfc6902 but not using XPath like expressions

See the examples from structpath used in tests today

rshriram:
@louiscryan I looked at JSON path. Few issues here: if we allow arbitrary customizability, we have to convert the entire Envoy config into JSON (or in this case every cluster, every listener, every route), do the add/replace/delete operation specified, and then convert it all back to proto. 

This will take us back to what mixer did in Pilot (bunch of struct conversions I think, that was causing a huge perf issue). Doing the same in Envoy filter will effectively slow all of Pilot down dramatically especially when someone does very simple stuff like adding a IAM filter for all envoys in the system or adding an idle timeout field for all http connection managers in the system.

rshriram:
Right now, its a simple proto merge https://github.com/istio/istio/pull/13276/files#diff-e25fff23f70aeb715787b971f7773aa4R321

we convert the config provided here into the proto representation (json to proto), and merge with existing proto coming in from Pilot's main networking code. 

From that point of view, the current PR offers a more scalable albeit a bit more restrictive solution

costinm:
I guess a simple proto merge is a good idea, and we can extend it with more
advanced solutions.  And we can generalize it to all generated protos (CDS,
LDS, filters). Still need some selector to indicate which proto to merge
with.

louiscryan:
@rshriram Actually JSONPath will evaluate over protobuf objects just fine. You just have to fixup the field names to match the generated field names. Its based on generic object reflection in Go. i.e. there's no requirement to roundtrip to JSON to make this work at all.

rshriram:
>you just have to fixup the field names to match the generated field names. 

So users have to use the right field names (fixed up versions) based on whatever protobuf for go generates? That doesn't scale well because we are now asking people to look at the pb.go files, and use the right fields (not to mention understand go).

We could try intelligently translating user specified paths to the proper ones if and only if we know the names of the paths the user is going to provide apriori. Since we allow users to provide arbitrary paths, we wont be able to do code level fixups that easily. Or am I missing something here?

louiscryan:
> > you just have to fixup the field names to match the generated field names.
> 
> So users have to use the right field names (fixed up versions) based on whatever protobuf for go generates? That doesn't scale well because we are now asking people to look at the pb.go files, and use the right fields (not to mention understand go).
> 

No, the fixup is trivial to implement in code. Users only need to reference the protobuf schema

> We could try intelligently translating user specified paths to the proper ones if and only if we know the names of the paths the user is going to provide apriori. Since we allow users to provide arbitrary paths, we wont be able to do code level fixups that easily. Or am I missing something here?

The protobuf schema is known completely at runtime so we can fixup whatever we want. 

Are you worried about arbitrary paths mapping into Struct types in protobuf or something equally esoteric ?

costinm:
I'm ok with both simple proto merging and more advanced merging using jsonpath. 
We can document and provide examples for the paths that will typically be used, and for 'break the glass' it's likely ok. 

But I would really like whatever we chose to be usable across all Istio APIs and generated code. 
We are already planning to adopt Kustomize ( JsonPath) for the installer overrides, and we had the problem of "Sidecar" or DestinationRule "admin overrides" for a long time. This can solve both the 'admin' and 'defaults' problem, so we should get it right.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,142 @@ message EnvoyFilter {
...
+
+  // Changes to be made to Envoy listeners generated by Pilot.
+  message Listener {
+    // Criteria to select a specific listener. If omitted, the patch
+    // will be applied to all listeners.
+    ListenerMatch match = 1;
+    // Patches to apply on the matched listener.
+    repeated Patch patches = 2;
+    // Listeners to be added. They will be shipped as is to the workload.
+    repeated google.protobuf.Struct additions = 3;
```
louiscryan: Why do we need 'additions'. Can just use a simple patch of type 'ADD', no path & the cluster

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,142 @@ message EnvoyFilter {
...
-  // REQUIRED: Envoy network filters/http filters to be added to matching
-  // listeners.  When adding network filters to http connections, care
-  // should be taken to ensure that the filter is added before
-  // envoy.http_connection_manager.
+  // Conditions specified in ClusterMatch must be met for the patch
+  // to be applied to a cluster.
+  message ClusterMatch {
+    // A regular expression to match on the Envoy cluster name or
+    // route name. The regular expression pattern should follow
+    // Golang's regular expression (RE2) format as specified here
```
louiscryan: Why would we want to do this instead of using a match in the Patch expression?

The synthetic cluster names are also a terrible thing to suggest people to match on. They are unstable and we'd rather people match on a more reliable attribute

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,142 @@ message EnvoyFilter {
...
+    // (inbound|outbound)|<portNumber>|<subset>|<FQDNofService>.  For
+    // example, outbound|80|v1|productpage.default.svc.cluster.local.
+    string name = 1;
+  };
+
+  // Conditions specified in RouteMatch must be met for the patch
+  // to be applied to a route.
+  message RouteMatch {
+    // A regular expression to match on the Envoy cluster name or
+    // route name. The regular expression pattern should follow
```
louiscryan: Same comment as above for RouteMatch

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,142 @@ message EnvoyFilter {
...
+    // plain text gateway or
+    // https.<server.portNumber>.<server.portName>.<gateway.Name>.<gateway.NameSpace>
+    string name = 1;
+  };
+
+  // Patch specifies the JSON path in the generated proto and the
+  // content to merge/remove on the specific path.
+  message Patch {
+    // A valid JSON Path expression against the type declared
+    // above. The path expression specifies the point in the object
```
louiscryan: 'declared above' is probably not going to generate into doc well

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,142 @@ message EnvoyFilter {
...
+    // graph where the mutation operation should occur. The patch is
+    // ignored if there is no object on the specified path.
+    string path = 1;
+
+    enum Operator {
+
+      // Overlay the value onto the element selected by the path using
+      // protobuf merge semantics. For primitive fields this is
```
louiscryan: No need to make the user facing documentation refer to 'protobuf merge semantics'. Just elide protobuf

networking/v1alpha3/envoy_filter.proto
rshriram: There is not much of a path for clusters. its literally clusterManager/[clusterNames]. So, the regex to select all inbound clusters, or all outbound clusters, or all clusters for port 80, etc. has to go somewhere. The other alternative is to devise listenerMatch style conditions that allow people to explicitly specify the match properties (port=80, hostname=*.local, direction=inbound, context=sidecar).

networking/v1alpha3/envoy_filter.proto
rshriram: So routes are tricky. While we can do 1-1 match for clusters, a single route is shared by many services (e.g., http.80). The reason for regex here is because of the gateways. We generate a route per https server in the Gateway, of the form (https.<port>.<portName>.crdName.crdNamespace) to clearly disambiguate the filter chain for each server. So, in this case, you could have 10s of servers on same port, differing only by their portNames/crd names. 

It becomes unwieldy to then introduce structured match conditions for these kinds of things.

Also note that unlike listeners, the clusters and routes are shipped via CDS where they are literally on the root of the path.. i.e.

```
 cds:
   resources: list of clusters
 rds:
   resource: route
```

So any path based expression that goes from the listener all the way to a route, or from clusterManager to a cluster is not going to work.

costinm:
Regex creates far too many problems -  efficiency, complexity, etc.

costinm:
Better alternative - use a json path expression.

rshriram:
> Regex creates far too many problems -  efficiency, complexity, etc.

Have you read the contents of the PR? The only place that uses regex is when matching the names of the objects (clusters, routes). The rest is JSON path.

costinm:
Yes - my comment was about not using regex in any place. Louis also has a
comment earlier about not matching on the name of the object at all,
and I agree with that comment since the naming is currently quite horrible,
but if we do change the naming to something more intuitive - we should
avoid regexes on matching the name as well.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,143 @@ message EnvoyFilter {
...
+  message ClusterMatch {
+    // A regular expression to match on the Envoy cluster name or
+    // route name. The regular expression pattern should follow
+    // Golang's regular expression (RE2) format as specified here
+    // (https://golang.org/pkg/regexp/). See
+    // https://github.com/google/re2/wiki/Syntax for examples.
+    //
+    // The names for Envoy clusters generated by Pilot follow the
+    // pattern
+    // (inbound|outbound)|<portNumber>|<subset>|<FQDNofService>.  For
```
costinm: How about instead add explicit matchers for each component. ( 
 host=productpage.bookinfo, [port=8080], [subset=v1]
with 'outbound' as default since it's the most common use case.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,143 @@ message EnvoyFilter {
...
+    // https://github.com/google/re2/wiki/Syntax for examples.
+    //
+    // The RDS routes for Envoy are named as follows:
+    // For sidecars,
+    // route name is the portNumber. Note that multiple HTTP services
+    // on the same port share the same RDS route.
+    //
+    // For gateways, route name is either http.<server.portNumber> for
+    // plain text gateway or
+    // https.<server.portNumber>.<server.portName>.<gateway.Name>.<gateway.NameSpace>
```
costinm: Same as clusters - better to have explicit match on each attribute instead of regex, and certainly not turn the 
current naming scheme into supported API.

louiscryan:
So there's a general issue around 'name' matching that we need to make a judgement about.

If we intend to use a generic expression engine it will be hard to have it match against synthetic information, e.g. inbound vs outbound, compound cluster names etc. We can look to extend the expression engine with custom functions that can mitigate some of the above

FWIW - JSON path is one expression engine that's available. The other option here is CEL which is specifically designed to work with protobuf.

networking/v1alpha3/envoy_filter.proto
rshriram: I can certainly do that for clusters. but not for routes.. because we would then have to add matchers like gatewayName, gateway namespace, etc.. At this point, the user would have no idea why these match fields like serverPortName, server portNumber, gatewayName, gatewayNamespace matter when selecting a RDS route. Alteast (host, subset, port) are clear for clusters.

I am afraid for RDS, we have to stick with the route Name as is for the simplest way of matching with any route the user wants

rshriram:
As I noted above, I can definitely add more structured match for clusters. It just wont be helpful for routes, due to the way we construct the route names in gateways.

rshriram:
We generate routes like "80", "http.80", "http_proxy", and "unix://blah" [for unix domain socket based routes].

Also for clusters, we generate "PassthroughCluster" that just wont match any conditional thing we add. And @dwradcliffe 's PR to tweak settings for passthrough cluster shows that people want to tweak such clusters as well. Standard match conditions wont capture those. So we need a regex in this case in addition to standard match conditions

costinm:
Or we need to express whatever attributes are needed to identify a route,
and not depend on the name.

The names are a mangling of several pieces of information.

rshriram:
Yes, but since this is a break glass configuration, meant for expert users, I am not that worried about the UX at all. The number of people using EnvoyFilter is much low, compared to other APIs. And these people evidently understand the envoy config and how pilot generates them. So, they can certainly write a regex to match on the route. Its not helpful to add 8 different match components to match on a route (right now, this list includes, portNumber, http|https, sidecar|gateway, gatewayServerPortName, gatewayObjectName, gatewayObjectNamespace, then this arbitrary "http_proxy" on sidecars, and then another arbitrary string to match on the unix domain socket based routes in Sidecar crd.

 The compute cost is same with 8 different match conditions or with a single regex. Also, with 8 match conditions, we now introduce more complexities like adding the concept of a namespace, and dealing with missing conditions and how they apply to multiple objects. 

Route names and cluster names are opaque entities with a name. If someone is going to the trouble of patching Pilot generated Envoy configs, then I feel its much less overhead to provide a regex that matches on a set of routes/clusters instead of umpteen match conditions. The only reason listener match exists is because it uses IPs for naming which cannot be matched easily.

networking/v1alpha3/envoy_filter.proto
rshriram: done, in addition to a regex match as a fallback

costinm:
On Fri, May 31, 2019 at 6:00 AM Shriram Rajagopalan <

> Yes, but since this is a break glass configuration, meant for expert
> users, I am not that worried about the UX at all. The number of people
> using EnvoyFilter is much low, compared to other APIs. And these people
> evidently understand the envoy config and how pilot generates them. So,
> they can certainly write a regex to match on the route. Its not helpful to
> add 8 different match components to match on a route (right now, this list
> includes, portNumber, http|https, sidecar|gateway, gatewayServerPortName,
> gatewayObjectName, gatewayObjectNamespace, then this arbitrary "http_proxy"
> on sidecars, and then another arbitrary string to match on the unix domain
> socket based routes in Sidecar crd.
>

I am not worried so much about UX for filters - but about the stability and
format of our (half baked and hardly API reviewed or documented) naming
 scheme.  And even if I am not worried about UX for filters - it is
valuable to still try to keep it clean so we may use it in other parts of
the
system, and keep it consistent with other configs.

Most attribute names we are discussing are well defined in other APIs -
either networking CRDs, or in config API ( values or mesh).
So they are stable names that users already need to be familiar with.  I
think your list has the variable names used internally -
for example the selector label would be "namespace: foo" not
gatewayObjectNamespace.
But even with the ugly names - it is still easier and cleaner to say
portNumber: 80 instead of some regex to match the port between '|'
in the middle of a mangled name.




The compute cost is same with 8 different match conditions or with a single
> regex. Also, with 8 match conditions, we now introduce more complexities
> like adding the concept of a namespace, and dealing with missing conditions
> and how they apply to multiple objects.
>
Matching labels is what we do at all layers in Istio. The semantics should
be clear to everyone.

Each object has a set of labels, be it generated config or source object or
synthetic (locality, network, etc in endpoints) - we match just like we
match everything else.



> Route names and cluster names are opaque entities with a name. If someone
> is going to the trouble of patching Pilot generated Envoy configs, then I
> feel its much less overhead to provide a regex that matches on a set of
> routes/clusters instead of umpteen match conditions. The only reason
> listener match exists is because it uses IPs for naming which cannot be
> matched easily.
>
The 'umpteen match conditions' are the same in both cases - you either
express them in a regex or in a list, but if you want to match on
something you still need to include it in both cases.

port: 80
subset: v1
hostname: productpage.bookinfo

or

 outbound\|v1\|80\|productpage.bookinfo

I agree regex is more compact - but the information is the same.

costinm:
Also - if we decide to document and commit to a naming scheme for generated
objects, I think it is important
to have it reviewed and fixed.

Using "|" is not very nice ( in particular in regex ), the "outbound" term
is not commonly used in our APIs.

We typically use DNS names or URLs for almost all named entities in our
config.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -61,6 +61,37 @@ import "google/protobuf/struct.proto";
...
+//   namespace: istio-system
+// spec:
+//   workloadLabels:
+//     istio: ingress-gateway
+//   listeners:
+//   - match:
+//       listenerType: GATEWAY # will match all gateway listeners
+//     patches:
+//     - path: '{.filter_chains[*].filters[?(@.name=="envoy.http_connection_manager")]}'
+//       operation: ADD
```
douglas-reid: nit: the proto field below is `operator` not `operation`.

nrjpoddar:
Thanks @rshriram for this PR around the new extension APIs.

I feel like Istio's usability story is not that great currently and this API as you have said is for experts, break glass API with lots of failure modes. Adding this API now without having a clear definition of what's core in Istio APIs and fixing current API's stability and usability seems to me like taking on a whole new set of user found errors and long list of issues showing up on Github.

Additionally, going forward for every new API proposed in Istio there will be debates on what goes in core and what's done via this extension mechanism.

rshriram:
@ramaraochavali / @shakti-das / other salesforce folks, could you guys comment on this PR ? It allows for some more tweaks to simple things like listener settings, cluster settings etc.

rshriram:
@nrjpoddar this has nothing to do with generic extension apis that @louiscryan talked about. This is purely for the envoy filter crd that we currently have today. It simply allows people to set specific fields in listeners/http connection managers, etc. instead of piling more and more fields into virtual service/destination rule. The latter is what causes more trouble as we have to fully support those. Not the envoy filters.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -90,23 +135,9 @@ message EnvoyFilter {
     // insensitive.
     string port_name_prefix = 2;
```
louiscryan: Likely want to deprecate this over time in favor a protocol match

networking/v1alpha3/envoy_filter.proto
```diff
@@ -90,23 +135,9 @@ message EnvoyFilter {
...
-      // Gateway listener
-      GATEWAY = 3;
-    };
-
     // Inbound vs outbound sidecar listener or gateway listener. If not specified,
     // matches all listeners.
-    ListenerType listener_type = 3;
+    Context listener_type = 3;
 
     enum ListenerProtocol {
```
louiscryan: The filter chain matching rules should include protocol. Essentially the concrete matching fields for Filter should come from Envoy's FilterChainMatch specification.

https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener/listener.proto#envoy-api-msg-listener-filterchainmatch

networking/v1alpha3/envoy_filter.proto
```diff
@@ -90,23 +135,9 @@ message EnvoyFilter {
...
-      SIDECAR_OUTBOUND = 2;
-
-      // Gateway listener
-      GATEWAY = 3;
-    };
-
     // Inbound vs outbound sidecar listener or gateway listener. If not specified,
     // matches all listeners.
-    ListenerType listener_type = 3;
+    Context listener_type = 3;
```
louiscryan: I know we can't rename the field here but we can in the other usages to just be 'context'

networking/v1alpha3/envoy_filter.proto
```diff
@@ -163,7 +194,7 @@ message EnvoyFilter {
     // listener match criteria for all filter insertions.
     ListenerMatch listener_match = 1;
 
-    // Insert position in the filter chain. Defaults to FIRST
+    // Insert position in the filter chain. Defaults to FIRST.
```
louiscryan: Need to fix the positional insert specification to be more type neutral

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
     // REQUIRED: The type of filter to instantiate.
     FilterType filter_type = 3;
 
-    // REQUIRED: The name of the filter to instantiate. The name must match a supported
-    // filter _compiled into_ Envoy.
+    // REQUIRED: The name of the filter to instantiate or replace. The
+    // name must match a supported filter _compiled into_ Envoy.
```
louiscryan: Link to Envoy reference here

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+    // direction.
+    uint32 port_number = 1;
+
+    // The FQDN of the service name for this cluster.
+    string hostname = 2;
+
+    // The subset associated with the service.
+    string subset = 3;
+
+    // Direction of traffic (inbound|outbound)
```
louiscryan: Don't need this, comes from 'Context'

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+    // (https://golang.org/pkg/regexp/). See
+    // https://github.com/google/re2/wiki/Syntax for examples.
+    //
+    // The names for Envoy clusters generated by Pilot typically
+    // follow the pattern
+    // (inbound|outbound)|<portNumber>|<subset>|<FQDNofService>.  For
+    // example, outbound|80|v1|productpage.default.svc.cluster.local.
+    // In addition, Pilot generates a PassthroughCluster to handle
+    // traffic bound for unknown destinations.
+    string regex = 5;
```
louiscryan: Don't think we need this with the expanded fields above + Context

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+  // Conditions specified in ClusterMatch must be met for the patch
+  // to be applied to a cluster.
+  message ClusterMatch {
+    // The service port for which this cluster was generated.  If
+    // omitted, applies to all clusters matching hostname, subset, and
+    // direction.
+    uint32 port_number = 1;
+
+    // The FQDN of the service name for this cluster.
+    string hostname = 2;
```
louiscryan: Don't think we want to use 'host' as the terminology here. We're careful to segregate 'host' and 'service' in docs elsewhere so we delinate addressing from service registry.

Also need to consider that we may have reserved 'destinations' like original_dest which are services but are not hosts.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
   };
 
-  // REQUIRED: Envoy network filters/http filters to be added to matching
-  // listeners.  When adding network filters to http connections, care
-  // should be taken to ensure that the filter is added before
-  // envoy.http_connection_manager.
+  // Conditions specified in ClusterMatch must be met for the patch
+  // to be applied to a cluster.
+  message ClusterMatch {
+    // The service port for which this cluster was generated.  If
```
louiscryan: Add Context

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+  // Conditions specified in RouteMatch must be met for the patch
+  // to be applied to a route.
+  message RouteMatch {
+    // The service port or gateway server port for which this route
+    // was generated.  If omitted, applies to all routes.
+    uint32 port_number = 1;
+
+    // Selects either inbound routes, outbound routes in sidecars, or
+    // routes on a gateway.
+    Context route_type = 2;
```
louiscryan: use 'context' as field name for consistency

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+    // (inbound|outbound)|<portNumber>|<subset>|<FQDNofService>.  For
+    // example, outbound|80|v1|productpage.default.svc.cluster.local.
+    // In addition, Pilot generates a PassthroughCluster to handle
+    // traffic bound for unknown destinations.
+    string regex = 5;
+  };
+
+  // Conditions specified in RouteMatch must be met for the patch
+  // to be applied to a route.
+  message RouteMatch {
```
louiscryan: Is this for HTTP or TCP or both. Need to doc.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+    // The service port or gateway server port for which this route
+    // was generated.  If omitted, applies to all routes.
+    uint32 port_number = 1;
+
+    // Selects either inbound routes, outbound routes in sidecars, or
+    // routes on a gateway.
+    Context route_type = 2;
+
+    // A regular expression to match on the Envoy route name. The
```
louiscryan: Are we matching on 'names' ? I'd rather be more structured like what you've done in Cluster so we don't expose the route naming variations in different serving contexts

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+    // For sidecars, the route name is the portNumber. Note that
+    // multiple HTTP services on the same port share the same RDS
+    // route. In addition, sidecars have a route called "http_proxy"
+    // on the HTTP Proxy listener.
+    //
+    // For gateways, route name is either http.<server.portNumber> for
+    // plain text servers on a gateway or
+    // https.<server.portNumber>.<server.portName>.<gateway.Name>.<gateway.Namespace>
+    // for HTTPS servers (with TLS termination) on a gateway.
+    string regex = 3;
```
louiscryan: Lets avoid this like the plague and have structured fields :P

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+  };
+
+  // Patch specifies the JSON path in the generated proto and the
+  // content to merge/remove on the specific path.
+  message Patch {
+    // A valid JSON Path expression specifying the point in the object
```
louiscryan: Note clear that we need all the complexity of JSON-PATH. I think CEL would be a better fit.

If we wanted to be simple we could skip the 'path' part in the API for now and always merge at the root of the selected type. It would make the 'value' have some more nesting but I doubt that would be totally blocking. We can always add later if it becomes an issue. Less is more.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -163,7 +194,7 @@ message EnvoyFilter {
     // listener match criteria for all filter insertions.
     ListenerMatch listener_match = 1;
 
-    // Insert position in the filter chain. Defaults to FIRST
+    // Insert position in the filter chain. Defaults to FIRST.
     InsertPosition insert_position = 2;
 
     enum FilterType {
```
louiscryan: Does this matter any more with the switch to FilterChains ?

networking/v1alpha3/envoy_filter.proto
utako: I don't mind what path syntax we use as long as there's a way to filter arrays without using the index. I can keep exploring CEL in lieu of JSONPath. I've been finding JSONPath merging to be limited, anyway.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +211,171 @@ message EnvoyFilter {
...
+    // Default behavior is MERGE.
+    Operation operation = 2;
+
+    // The proto equivalent of JSON Value. This will be merged using
+    // json merge semantics with the existing proto in the path.
+    google.protobuf.Value value = 3; 
+  };
+
+  // Changes to be made to Envoy listeners generated by Pilot.
+  message Listener {
```
louiscryan: Consider renaming these types to XXXXPatch just so its easier to disambiguate them from the things they are actually patching.

networking/v1alpha3/envoy_filter.proto
rshriram: So this is not just for a filter chain in a listener. Its for the entire listener. We could add another condition for the filter chain, but how do you reason about things like destination cidrs, arrays etc. ? if user specifies only 2 elements in the array and the actual one has 10 elements, do we say match succeeded if the 2 elements matched? I find that a bit strange. I would much rather go for exact match.

On a side note, I think the need for a filter chain match thing is going to be unlikely. The only place we use this is in permissive mTLS. In the gateway, we have SNI matches - which I think we should capture in the match conditions - with exact match

networking/v1alpha3/envoy_filter.proto
rshriram: ack

networking/v1alpha3/envoy_filter.proto
rshriram: So envoy doesn't have one. I actually looked for it. Its in the protobuf spec. If thats okay..

networking/v1alpha3/envoy_filter.proto
rshriram: may be I could convert this to clusterName. the two use cases here are the generic clusters we generate like "PassthroughCluster", "BlackholeCluster". I was mostly looking out for the passthrough.. so clusterName might capture those.

networking/v1alpha3/envoy_filter.proto
rshriram: see comment above. We also dont have an original_dest cluster. I nuked it lonng time ago. We actually have a cluster with the hostname, and the lb type is original dst. 
If I say the service name for this cluster, how will the service name be represented? Its got to be qualified (name.namespace). But then it wont work for service entries which could result in multiple clusters [one per hostname]

networking/v1alpha3/envoy_filter.proto
rshriram: ack

networking/v1alpha3/envoy_filter.proto
rshriram: only for HTTP. no RDS in TCP

networking/v1alpha3/envoy_filter.proto
louiscryan: Effectively each of the patch collections is a filter over an array.

In some cases we've created a new patch collection to handle this case. E.g.  FilterChain is a nested array within Listener but we have the FilterPatch to handle relative positioning. In part this is also done because we expect there to be a Filter Discovery API in the not too distant future.

Within each of the patched types are there concrete cases for being able to mutate nested arrays in ways that require relative positional offsets? i.e cases where the ordering of elements in the array is significant and the number / complexity of the elements in the array is high.  I'm struggling to point out examples within the Envoy APIs beyond the one's we've created concrete types for.

If the cardinality of a nested array is low then the cost of overriding the complete array is generally low.

Looking at each type for nested collections I see:

**Listener:**
filter_chains - covered by the 'Filters' patch
listener_filters - low cardinality
socket_options - low cardinality

**Filters** - None

**Cluster:** 
hosts - would be odd to partially override, only used with DNS anyway
health_checks - TBD but likely to be part of ServiceEntry API over time
dns_resolvers - lightly used and cardinality very low
lb subset_selectors, subset config keys - Mostly should map into ServiceEntry API and some LB config

**Route** - I think we really mean VirtualHost here
domains - cardinality could be high, mutation is only add /replace/ remove as value is string
routes - most likely operations are prepend (prioritized routing), append (catch all), deletion seems unlikely. Has decent API coverage for more typical use-cases
route.<header mutations> - best guess would be that header injection is tied to prepending a route. Already has decent API coverage
virtual_clusters - similar to domains though for specialized telemetry annotation, not likely to be common and operations are only add/remove/replace because value is string
route.weighted_cluster - seems unlikely as weights require computation, has decent API coverage in VirtualService
route match.headers & route match.query params - Cardinality is generally low, has decent coverage in VirtualService  except for query params which we need to fix :/

networking/v1alpha3/envoy_filter.proto
rshriram: suggestions for expressing 
  `https.<server.portNumber>.<server.portName>.<gateway.Name>.<gateway.Namespace>` as structured match? Should I literally add fields in the match for all these things?

networking/v1alpha3/envoy_filter.proto
rshriram: Oh yeah. it applies to each filter chain.. we switched ages ago and people are still using this for injecting http filters or tcp filters..

networking/v1alpha3/envoy_filter.proto
rshriram: ack

networking/v1alpha3/envoy_filter.proto
louiscryan: When we have protocol sniffing we won't decorate the listener with a protocol we'll use a listener filter and then filter chain matches on the protocols to dictate routing. This is similar to what we do for permissive TLS but with sniffing it will become a LOT more common.

networking/v1alpha3/envoy_filter.proto
louiscryan: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener/listener.proto#envoy-api-msg-listener-filter

networking/v1alpha3/envoy_filter.proto
louiscryan: A ServiceEntry with resolution#NONE is original_dest right ? That would be qualified

We should create a default one during install

networking/v1alpha3/envoy_filter.proto
louiscryan: Ideally we would create ServiceEntry objects for each of these by default during install. If we have reserved names for these we can document them and enforce some convention E.g.

$passthrough
$blackhole_reject_connection
$http_blackhole_503

etc.

networking/v1alpha3/envoy_filter.proto
louiscryan: Should we call this VirtualHost then ?

networking/v1alpha3/envoy_filter.proto
louiscryan: aren't gateway=true|false and namespace covered by context and the label match above ?

If so this boils down to port matching which is similar to Listener

networking/v1alpha3/envoy_filter.proto
rshriram: Thats a very very limited list :). The real list is in buried inside the protobuf comments somewhere.

networking/v1alpha3/envoy_filter.proto
rshriram: yes, but the protocol will be empty in cases like gateway, where only the sni matches are present

networking/v1alpha3/envoy_filter.proto
rshriram: It would have to be a special invalid service entry object with empty ports, empty address, shortname host (passthrough).

networking/v1alpha3/envoy_filter.proto
rshriram: serviceentry for *.googleapis.com with resolution NONE is a cluster of the form
```outbound|*.googleapis.com|443|..
        lbType: orig_dst
```

networking/v1alpha3/envoy_filter.proto
rshriram: It not a vhost though. its the envoy Route Object that has vhosts, plus additional stuff like common headers to add/remove, common settings across all vhosts, plus list of all vhosts.

networking/v1alpha3/envoy_filter.proto
rshriram: remember this gateway.Name is the gateway CRD object name. the gateway.Namespace is the gateway CRD object namespace. the server.PortName is the portname in the port field of the Server object in the Gateway CRD.

networking/v1alpha3/envoy_filter.proto
rshriram: my rationale for putting this in the first place: we dont cover all protocols. e.g., kafka/zookeeper/Foobar are still treated as tcp. if those ports are properly named, then the user could select listeners for that portname, apply their custom codec filters. Not all codecs are in the open either.

networking/v1alpha3/envoy_filter.proto
rshriram: converted to name.. with comments

rshriram:
releaselocks/release-1.0/ has been updated to reflect the message type change from ListenerType to Context. Same for 1.1. Only the message types have changed. enum values remain the same.

rshriram:
For reference, I tried to retain the old enum and have context as another enum. But since these two enums are nested inside the same protobuf message, two enums can't share same enum value names (like ANY). WE dont want to move Context out of the protobuf message.

ramaraochavali:
@rshriram Thank you. At high level this approach makes sense. I will take a detailed look and add comments.

networking/v1alpha3/envoy_filter.proto
ramaraochavali: I do not think there is any place where it is listed completely today. It is only listed [here](https://github.com/envoyproxy/envoy/blob/b771f99f53142bbb1496dbcddccc5defaaaa8dd8/source/extensions/filters/http/well_known_names.h#L13)  and [here](https://github.com/envoyproxy/envoy/blob/b771f99f53142bbb1496dbcddccc5defaaaa8dd8/source/extensions/filters/network/well_known_names.h#L13). There was a plan to auto generate the filter names in to the docs but has not been done yet.

shakti-das:
@rshriram This feature looks great to me. With the added customization support in `EnvoyFilter`, I wonder whether the name `EnvoyFilter` does justice to it, it feels like a generic EnvoyPatch/EnvoyOverride.

networking/v1alpha3/envoy_filter.proto
ramaraochavali: I found the above list in the docs. So may be good to refer to docs?.

networking/v1alpha3/envoy_filter.proto
ramaraochavali: I think `RouteMatch` is good, if we want to be super clear we can rename this to `HttpRouteMatch`

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +215,169 @@ message EnvoyFilter {
...
+    string name = 5;
+  };
+
+  // Conditions specified in RouteMatch must be met for the patch
+  // to be applied to a route.
+  message RouteMatch {
+    // The service port number or gateway server port number for which
+    // this route was generated.  If omitted, applies to routes for
+    // all ports.
+    uint32 port_number = 1;
```
ramaraochavali: Are both `port_number` and `port_name` required? and what happens if they conflict with each other? I guess it won't be applied?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +215,169 @@ message EnvoyFilter {
...
+    // this route was generated.  If omitted, applies to routes for
+    // all ports.
+    uint32 port_name = 2;
+
+    // Selects either inbound routes, outbound routes in sidecars, or
+    // routes on a gateway.
+    Context context = 3;
+
+    // The Istio gateway config's namespace/name for which this route
+    // was generated. Applies only if the conntext is GATEWAY. Should
```
ramaraochavali: nit: `conntext` -> `context`

rshriram:
>This feature looks great to me. With the added customization support in EnvoyFilter, I wonder whether the name EnvoyFilter does justice to it, it feels like a generic EnvoyPatch/EnvoyOverride.
That is certainly true. But I am not sure if there is appetite to introduce another new CRD. Even I personally dont feel like it. And given that this is for power users who understand the thing, not normal istio users, the clunky name shouldb't be a big deal.

networking/v1alpha3/envoy_filter.proto
rshriram: Almost everything in the match condition is optional. If omitted, its equivalent to a wildcard. So, if you provide 80 & mysql, it just wont match with any listener or route/cluster ..

shakti-das:
Yes, that makes sense.

networking/v1alpha3/envoy_filter.proto
louiscryan: In essence we're treating this like a label match then ? I'm fine with that

networking/v1alpha3/envoy_filter.proto
louiscryan: The protocol would be TLS right ? In any case I'm not sure an enum makes sense here for the reasons you outline above about protocol diversity

networking/v1alpha3/envoy_filter.proto
```diff
@@ -149,21 +184,21 @@ message EnvoyFilter {
...
-    // conditions are true.  If not specified, the filters will be
+    // conditions are true. If not specified, the filters will be
     // applied to all listeners where possible, potentially resulting
     // in invalid configurations. It is recommended to specify the
     // listener match criteria for all filter insertions.
     ListenerMatch listener_match = 1;
 
-    // Insert position in the filter chain. Defaults to FIRST
+    // Insert position in the filter chain. Defaults to FIRST.
     InsertPosition insert_position = 2;
```
louiscryan: I assume we're deprecating this in favor of the Patch.Operation

networking/v1alpha3/envoy_filter.proto
louiscryan: Ok, so this is a match against the routes inside vhost as if it were a flattened collection. That's fine but then we should add match fields for common fields in VirtualHost, i.e name and domain.

It also worth nothing that by doing this we don't have the ability to mutate VHost itself and we might need to backfill for that

networking/v1alpha3/envoy_filter.proto
louiscryan: *.googleapis.com is not an FQDN but it can be a service

networking/v1alpha3/envoy_filter.proto
```diff
@@ -74,6 +105,20 @@ message EnvoyFilter {
   // the mesh.
   map<string, string> workload_labels = 1;
 
+  enum Context {
```
ramaraochavali: nit: It seems `Context` is very generic - should we call it `PatchContext` or `ConfigOverrideContext`?. NBD but just throwing it out for discussion

networking/v1alpha3/envoy_filter.proto
```diff
@@ -74,6 +105,20 @@ message EnvoyFilter {
...
+    // All listeners/routes
+    ANY = 0;
+
+    // Inbound listener/route in sidecar
```
ramaraochavali: applies to cluster as well it seems?

networking/v1alpha3/envoy_filter.proto
rshriram: There is no patch stuff for filters.. so we still need this right?

rshriram:
@louiscryan / @ramaraochavali / @utako take a look at my latest update.. Its lumping all the patch and match into one generic object. If this is okay, then i shall clean up the comments, etc. If this is too confusing, then I can go back to the individual XXXPatch objects

ramaraochavali:
@rshriram I am good with the new approach. I might propose minor naming changes but I am good with a generic message type.

Thinking through - couple of high level thoughts 
1. Upgrades - We should document clearly this might break across upgrades if Pilot changes config generation especially w.r.t deprecated fields in Envoy.
2. Are you planning to implement all of the criteria right from beginning or planning to start with minimal api surface and start expanding as people use it?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +215,251 @@ message EnvoyFilter {
...
+
+    // The SNI value used by a filter chain's match condition.  This
+    // condition will evaluate to false if the filter chain has no sni
+    // match.
+    string sni = 2;
+
+    // The ALPN protocol used by a filter chain's match
+    // condition. This condition will evaluate to false if the filter
+    // chain has no ApplicationProtocols setting.
+    string alpn = 3;
```
louiscryan: protocols won't just be determine by ALPN/NPN so I would just call this protocol

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +215,251 @@ message EnvoyFilter {
...
+
+    // The subset associated with the service. If omitted, applies to
+    // clusters for any subset of a service.
+    string subset = 3;
+
+    // Inbound vs outbound sidecar cluster or gateway outbound
+    // cluster. If omitted, applies to all clusters.
+    PatchContext context = 4;
+
+    // The exact name of the cluster to match. To match a specific
```
louiscryan: We generally want to get away from hardcoded cluster names and start representing these things as 1st class resource. For example could we represent the hardcoded clusters as entries in this API or ServiceEntry instead ?

'Passthrough' is a good example of this

networking/v1alpha3/envoy_filter.proto
louiscryan: ping?

I'm OK with calling this 'host' if we change the doc a bit.

'The host name associated with the service defined by this cluster'

networking/v1alpha3/envoy_filter.proto
louiscryan: ... can we enumerate all the hardcoded cluster names ?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +215,251 @@ message EnvoyFilter {
...
+
+    // The Istio gateway config's namespace/name for which this route
+    // configuration was generated. Applies only if the context is
+    // GATEWAY. Should be in the namespace/name format. Use this field
+    // in conjunction with the portNumber and portName to accurately
+    // select the Envoy route configuration for a specific HTTPS
+    // server within a gateway config object.
+    string gateway = 4;
+
+    // Route configuration name to match on. Can be used to match a
```
louiscryan: Do we one of these not called 'http_proxy' ?

shakti-das:
@rshriram I liked the original separation for each Listener/Route/Cluster. This is far more confusing considering we are already overloading `EnvoyFilter`. Even for advanced users, it's easier to tweak and explain in code reviews with the separate approach than with the cumulative/lumped approach.

shakti-das:
@rshriram Please ignore my previous comment, this looks good to me. I misinterpreted lumping, it's actually optimizing the structure.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +215,251 @@ message EnvoyFilter {
...
+    // internally generated "allow_any" or "block_all" virtual host
+    // configurations for all sidecars.
+    string name = 5;
+  }
+
+  // For listeners with multiple filter chains (e.g., inbound
+  // listeners on sidecars with permissive mTLS, gateway listeners
+  // with multiple SNI matches), the filter chain match can be used to
+  // select a specific filter chain to patch.
+  message FilterChainMatch {
```
louiscryan: This needs to have a match against filter name to be a useful mutation over the logical 'filters'  collection, right now it's only useful for mutating the FilterChainMatch resource itself. I'd like to avoid being dependent on JSON-Patch for complex array mutations within resources like FilterChain.

Consider the use-case of  'prepend a filter to all filter chains'

The object model is
             filter_chains                      filters
Listener [1] --> [N] FilterChain [1] --> [N] --> Filter

The API as currently defined is really a match against all the 'filter_chains' in all the listeners. This means that the mutation is done against the FilterChain resource which by implication means that to perform a prepend the patch has to have a path selector that reaches into the 'filters' array and  has some array indexing

patch_filter_chain(match=blah, path='.filters', op=PREPEND, value=<some filter resource>)

This works ok because '.filters' is referring to an array and we have an operator which says to inset something at the start of the array. Now to perform an operation like 'prepend a filter before 'http_conn_manager in all filter chains things start to get painful, we need to refer to the array and something in the array at the same time using the path expression

patch_filter_chain(chain_match=blah, path='.filters[$.name=http_conn_manager]', op=INSERT_BEFORE, value=<some sub_resource of filter_chain>)

Now the JSON-PATH expression is referring to a resource in the array so the operation has to change to be relative and the runtime has to refer back to the container of the referenced resource. 

The other way to handle this is to have the match be against all the filters in a chain, i.e were matching against the elements of the collection in the 'context of' the chain. The relative operations become easier to specify

patch_filters_in_chain(chain_match=blah, op=INSERT_BEFORE,  value=<a filter>)  // insert in first position as the first filter in the list will match
patch_filters_in_chain(chain_match=blah, filter_match.name=http_conn_manager, op=INSERT_BEFORE,  value=<a filter>)  // relative insertion

The number of operations needed is reduced, the path complexity is reduced and the potential diversity of value types is reduced.

We can keep both types of patches if we feel the need to patch the non-collection fields of FilterChain.

Essentially we have two classes of matcher+patch types, ones for key collection types (listeners, filters, clusters, vhosts & routes) and ones for their key container types (FilterChain,RouteManager etc.). 

Make sense?

networking/v1alpha3/envoy_filter.proto
rshriram: we have istio as a protocol in the mtls sniffing. :( . That doesnt warrant being called protocol. wdyt?

rshriram:
> Upgrades - We should document clearly this might break across upgrades if Pilot changes config generation especially w.r.t deprecated fields in Envoy.

@ramaraochavali I am not sure if this will break upgrades. The structure of envoy config remains the same across istio upgrades. If the user sets a non existent field on a newer proxy, it wont work [as we silently ignore unknown fields in envoy, IIRC]. But it certainly wont crash envoy unless the user sets an incorrect config. Am I missing something here?

ramaraochavali:
@rshriram 
> The structure of envoy config remains the same across istio upgrades.
In general that is true but  may not be the case always - especially when fields are deprecated in Envoy. Let us say pilot v1 is using a field in Envoy and later Envoy deprecates the field for some reason. And user has configured the patch context to refer to that field. Few releases later Envoy removes that field and Pilot code is changed to not use that field in v2 and move to the replacement. Then when user upgrades to v2, the patch might break with non existing field. While this is a very rare scenario, this might happen. While Envoy won't crash, I am assuming patch may not be applied because the path evaluation fails - Is my understanding correct with path evaluation?

rshriram:
>While Envoy won't crash, I am assuming patch may not be applied because the path evaluation fails - Is my understanding correct with path evaluation?

Yes. it wont be applied and will likely fail silently.

ramaraochavali:
> Yes. it wont be applied and will likely fail silently.

Yeah. This is what I meant when I said it might break upgrades. But since this is intended for power users, they will detect the failures in staging envs and fix their  patch. Not a big deal but may be worth documenting.

rshriram:
@ramaraochavali / @louiscryan / @utako take a look at the three doc examples I have added and let me know if it looks palatable https://github.com/istio/api/pull/899/files#diff-fa79aef151053e8b9e7be36c58eeaf39R54

networking/v1alpha3/envoy_filter.proto
```diff
@@ -50,17 +57,111 @@ import "google/protobuf/struct.proto";
...
+//   configPatches:
+//     # The first patch adds the lua filter to the listener/http connection manager
+//   - match:
+//       listener:
+//         context: SIDECAR_INBOUND # will match with the inbound listener for reviews:8080
+//         portNumber: 8080
+//         filterChain:
+//           filterName: "envoy.http_connection_manager"
+//     patches:
+//     - path: '{.http_filters}' # relative to http connection manager
```
louiscryan: What type does this match apply to? The top level specifier is 'listener' but the condition on 'filterChain' and 'filter' below are designed to select a filter and not a listener. So if the 'filterChain' didn't include a filterName field it would select the FilterChain resource ?

Having done that the mutation is on a field in the filter. We've used on mechanism for the path structure for listeners..filters  but now we use an alternate syntax for the  http_conn_manager...http_filters. Why not just hoist this into the same nesting structure above?

I'm not sure that using nested objects to perform selection and type refinement is going to work as well as flattening into a match per logical connection. For example here is what it would look like for a flattened matcher which is processed for each entry in the http_filters collection within each http_conn_manager


// kind: EnvoyFilter
// metadata:
//   name: reviews-lua
// spec:
//   workloadLabels:
//     app: reviews
//   configPatches:
//     # The first patch adds the lua filter to the listener/http connection manager
//   - match:
//       http_filters:  # Will be applied to all filters in http_filters in http_conn_managers for the defined context
//         context: SIDECAR_INBOUND # will match with the inbound listener for reviews:8080
//         portNumber: 8080
//         name: 'envoy.router'  # Match the router filter
//     patches:
//     - path: '' # No needed as we're applying it to http_filter resources
//       operation: INSERT_BEFORE  # Insert before the router filer
//        name: envoy.lua
//        config:
//          inlineCode: |
//            function envoy_on_request(request_handle)
//              -- Make an HTTP call to an upstream host with the following headers, body, and timeout.
//              local headers, body = request_handle:httpCall(
//               "lua_cluster",
//               {
//                [":method"] = "POST",
//                [":path"] = "/acl",
//                [":authority"] = "internal.org.net"
//               },
//              "authorize call",
//              5000)
//            end

The type is clearer, we can have one operation field instead of two, there is little nesting within the match and we generally don't need to have a 'path' for most use-cases

networking/v1alpha3/envoy_filter.proto
```diff
@@ -180,18 +286,226 @@ message EnvoyFilter {
...
+    // Instead of using specific port numbers, a set of ports matching
+    // a given port name can be selected. Matching is case
+    // insensitive.
+    string port_name = 3;
+
+    // Match a specific filter chain in a listener. If specified, the
+    // patch will be applied to the filter chain (and a specific
+    // filter if specified) and not to other filter chains in the
+    // listener.
+    FilterChainMatch filter_chain = 4;
```
louiscryan: So if this value is defined the selected object for mutation is now the FilterChain resource, and recursively if FilterChainMatch.filter_name is defined then the resource for mutation is now a filter ?

I think that could be OK but the matching semantics need to change a little as it should be applicable to all filter chains and not just the first matching one.

We should also nest into a type the filter matching criteria in case we need to match on more than just name

nrjpoddar:
> @nrjpoddar this has nothing to do with generic extension apis that @louiscryan talked about. This is purely for the envoy filter crd that we currently have today. It simply allows people to set specific fields in listeners/http connection managers, etc. instead of piling more and more fields into virtual service/destination rule. The latter is what causes more trouble as we have to fully support those. Not the envoy filters.

I understand that this PR only applies to generated Envoy configuration and not mutating Istio APIs themselves. Giving the ability to patch Istio generated Envoy config directly seems like a very low level interface. As the current Envoy.Filter API stands which only allows addition makes sense as it allows users to add custom Lua filters. Ability to mutate all pieces of Envoy configuration seems to me is a recipe for maintenance/support nightmares, specially around merge semantics w.r.t multiple Envoy filter resources.

rshriram:
@robertpanzer since you were playing with the envoy filters, PTAL at this and let us know if this meets some of your needs.

robertpanzer:
@rshriram Looks good to me, currently I assume that I can migrate my filter config to this new schema.

rshriram:
Yes. we will support the old filter insertion thing for one release.

mesh/v1alpha1/config.proto
```diff
@@ -221,7 +221,6 @@ message MeshConfig {
   // Fallback to old identity format(without trust domain) if not set.
   string trust_domain = 26;
 
-  // $hide_from_docs
```
louiscryan: Why are these changes in this PR ?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+  // Zero or more labels that indicate a specific set of sidecars or
+  // gateways whose envoy configurations need to be customized. The
+  // scope of label search is restricted to the configuration
+  // namespace in which the the resource is present. In other words,
+  // the EnvoyFilter resource must reside in the same namespace as the
+  // gateway/sidecar. If omitted, the EnvoyFilter patches will be
+  // applied to all workloads in the same namespace. If the
+  // EnvoyFilter is present in the config root namespace, it will be
+  // applied to all applicable workloads in any namespace.
+  map<string, string> workload_selector = 3;
```
louiscryan: We have a type for this

networking/v1alpha3/envoy_filter.proto
```diff
@@ -17,67 +17,194 @@ syntax = "proto3";
...
+// kind: EnvoyFilter
+// metadata:
+//   name: custom-protocol
+//   namespace: istio-config # as defined in meshConfig resource.
+// spec:
+//   configPatches:
+//   - applyTo: NETWORK_FILTER
+//     match:
+//       context: SIDECAR_OUTBOUND # will match outbound listeners in all sidecars
+//       lds:
```
louiscryan: Is the nesting under 'lds' adding anything? It couples the API to the protocol in ways we don't really need

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+  // ApplyTo specifies where in the Envoy configuration, the given patch should be applied.
+  enum ApplyTo {
+
+    INVALID = 0;
+
+    // Applies the patch to the listener.
+    LISTENER = 1;
+
+    // Applies the patch to the filter chain match condition.
+    FILTER_CHAIN_MATCH = 2;
```
louiscryan: Isn't the patch applying to FilterChain and not FilterChainMatch ?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
     google.protobuf.Struct filter_config = 5;
   };
 
-  // REQUIRED: Envoy network filters/http filters to be added to matching
-  // listeners.  When adding network filters to http connections, care
-  // should be taken to ensure that the filter is added before
-  // envoy.http_connection_manager.
-  repeated Filter filters = 2;
+  // ApplyTo specifies where in the Envoy configuration, the given patch should be applied.
+  enum ApplyTo {
```
louiscryan: Do we want to support all of these from the get go ? Can choose to define the schema but not implement for the the apply modes.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+    }
+
+    // The service port number or gateway server port number for which
+    // this route configuration was generated. If omitted, applies to
+    // route configurations for all ports.
+    uint32 port_number = 1;
+
+    // The gateway server port name for which this route configuration
+    // was generated. Port name is not applicable for sidecar contexts.
+    uint32 gateway_server_port_name = 2;
```
louiscryan: Why isn't this just port_name? We have such a thing in Sidecar

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+
+    // The Istio gateway config's namespace/name for which this route
+    // configuration was generated. Applies only if the context is
+    // GATEWAY. Should be in the namespace/name format. Use this field
+    // in conjunction with the portNumber and portName to accurately
+    // select the Envoy route configuration for a specific HTTPS
+    // server within a gateway config object.
+    string gateway = 3;
```
louiscryan: Does this belong here or is it really part of 'Context'.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+    // listeners on sidecars with permissive mTLS, gateway listeners
+    // with multiple SNI matches), the filter chain match can be used
+    // to select a specific filter chain to patch.
+    message FilterChainMatch {
+      // The SNI value used by a filter chain's match condition.  This
+      // condition will evaluate to false if the filter chain has no
+      // sni match.
+      string sni = 1;
+
+      // Applies only to SIDECAR_INBOUND context. If non-empty, a
```
louiscryan: Applies to GATEWAY too I believe

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+    // The service port/gateway port to which traffic is being
+    // sent/received. If not specified, matches all listeners. Even though
+    // inbound listeners are generated for the instance/pod ports, only
+    // service ports should be used to match listeners.
+    uint32 port_number = 1;
+
+    // Instead of using specific port numbers, a set of ports matching
+    // a given service's port name can be selected. Matching is case
+    // insensitive.
+    string service_port_name = 2;
```
louiscryan: elide the 'service_'

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+      // chain match.  This value will be compared against the
+      // transport protocol of a new connection, when it's detected by
+      // one of the listener filters.
+      //
+      // Accepted values include:
+      //
+      // * `raw_buffer` - default, used when no transport protocol is detected.
+      // * `tls` - set when TLS protocol is detected by the TLS inspector.
+      string transport_protocol = 2;
+
```
louiscryan: Need the 'application_protocols' from 

https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/listener/listener.proto#listener-filterchainmatch

networking/v1alpha3/envoy_filter.proto
```diff
@@ -189,9 +320,289 @@ message EnvoyFilter {
...
+      // require a value to be specified. When removing fields their
+      // values are reset to their intrinsic default.
+      REMOVE = 3;
+
+      // Insert operation on an array of named objects. This operation
+      // is typically useful only in the context of filters, where the
+      // order of filters matter. For clusters and virtual hosts,
+      // order of the element in the array does not matter. Insert at
+      // the front of the list.
+      INSERT_FIRST = 4;
```
louiscryan: Not sure we need _FIRST or _LAST. If the selector selects all elements in the array, e.g. no filter name or sub filter name match, then we can use _BEFORE or _AFTER

networking/v1alpha3/envoy_filter.proto
rshriram: Because this port name here applies only to the gateway. This is matching on the Route Configuration name. Our route configurations for sidecars are simply the port numbers. But for Gateways, we have to distinguish the rds for each server (due to https).

networking/v1alpha3/envoy_filter.proto
rshriram: This is the gateway object name.. rds for a gateway server is https.443.https1.appGw.istio-system .. We need to match on the 443 - port number, https1 which is the port Name in the `Server` object in Gateway, istio-system/appGw is the gateway name.

networking/v1alpha3/envoy_filter.proto
rshriram: We dont really do anything with ALPN for now. We dont even generate distinct filter chains for http2/http1. The idea behind having the filter chain match piece here is to allow people to tweak the two filter chains created for mTLS [with sniffing]. When it comes to gateways, we can simply match using the SNI value.

networking/v1alpha3/envoy_filter.proto
rshriram: Nope.. this is the thing that is set by the mTLS sniffing code in Envoy. its vaues are rawBuffer and tls. none of this is set for gateway.

networking/v1alpha3/envoy_filter.proto
rshriram: ack

networking/v1alpha3/envoy_filter.proto
rshriram: fixed

networking/v1alpha3/envoy_filter.proto
louiscryan: This is going to evolve though so we should be careful. Can you expand in the doc

networking/v1alpha3/envoy_filter.proto
louiscryan: Sure, don't we sniff TLS in Gateways ?

networking/v1alpha3/envoy_filter.proto
louiscryan: Its OK not to have it now but we're definitely going to need it down the road as we support more protocols.

networking/v1alpha3/envoy_filter.proto
rshriram: Nope. Only on sidecars. We dont even have API support for such sniffing at gateways.

networking/v1alpha3/envoy_filter.proto
rshriram: okay..

networking/v1alpha3/envoy_filter.proto
```diff
@@ -17,67 +17,194 @@ syntax = "proto3";
...
+//                [":method"] = "POST",
+//                [":path"] = "/acl",
+//                [":authority"] = "internal.org.net"
+//               },
+//              "authorize call",
+//              5000)
+//            end
+//   # The second patch adds the cluster that is referenced by the lua code
+//   # cds match is omitted as a new cluster is being added
+//   - applyTo: CLUSTER
```
louiscryan: @mandarjog @costinm 

This is a great example of where we would rather users used a ServiceEntry. This use-case highlights two issues that make doing that difficult that we should try to address

- We mangle names
- We can't force a cluster to be generated in Gateway (we can in Sidecar)

We should come up with a proposal to solve these as it will be a better UX than asking people to do this.

louiscryan: Created

https://github.com/istio/api/issues/955

rshriram: We can force a cluster to be generated in the gateway by adding a Service entry in the gateway namespace and scoping it with `exportTo: .` - its a bit too clunky though. Also, if someone already has figured out that they are using service entry just to generate a cluster in envoy [i.e. they know envoy internals], then why provide an alternative means of creating the cluster when they can do the same with envoyFilter ? Remember that the class of people adding an explicit service to the system just so that their sidecars or gateways or even their apps can access it - only understand things at Istio level [services, with hostnames, IPs, etc.]. If we now tell them to add service entries with names like $xyz, with some IPs, etc. then its same amount of work with the exception of using $xyz instead of a hostname.

louiscryan: Lets move discussion to issue...

networking/v1alpha3/envoy_filter.proto
louiscryan: shouldn't this be renamed to FILTER_CHAIN then?

networking/v1alpha3/envoy_filter.proto
louiscryan: Doc still out of date ?

networking/v1alpha3/envoy_filter.proto
rshriram: fixed

PR #901, opened 2019-04-15
Clarify doc for VirtualService gateways field

frankbu:
Fixes https://github.com/istio/istio/issues/11203

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -144,14 +144,12 @@ message VirtualService {
   // only for services defined via the Gateway.
   repeated string hosts = 1;
 
-  // The names of gateways and sidecars that should apply these rules. A
-  // single VirtualService can be used for sidecars inside the mesh as well
-  // as for one or more gateways. The reserved word `mesh` is used to apply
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/virtual_service.proto
```diff
@@ -144,16 +144,17 @@ message VirtualService {
...
-  // single VirtualService is used for sidecars inside the mesh as well as
-  // for one or more gateways. The selection condition imposed by this
-  // field can be overridden using the source field in the match conditions
-  // of protocol-specific routes. The reserved word `mesh` is used to imply
-  // all the sidecars in the mesh. When this field is omitted, the default
-  // gateway (`mesh`) will be used, which would apply the rule to all
-  // sidecars in the mesh. If a list of gateway names is provided, the
-  // rules will apply only to the gateways. To apply the rules to both
-  // gateways and sidecars, specify `mesh` as one of the gateway names.
+  // The names of gateways and sidecars that should apply these rules. A
```
hzxuzhonghu: why `sidecars` names ?

hzxuzhonghu: mesh has a special meaning

frankbu: good point, this is badly worded

PR #902, opened 2019-04-16
Mesh ProxyConfig: Add envoy metadata

drichelson:
This continues to address https://github.com/istio/istio/issues/11984 and is a follow-on to https://github.com/istio/istio/pull/12189

We're adding the node metadata to this struct so it can be easily plugged in to the bootstrap config when pilot-agent starts.


drichelson:
@rshriram Please see istio PR for a little more context: https://github.com/istio/istio/pull/13367/files
We already have cert file paths as args in pilot-agent. They are used for both watching custom cert paths + sending metadata to pilot-discovery. Are you suggesting we rework the cert file watcher config code to read from env vars?

rshriram:
Closing as the issue has been resolved in a different way. Thanks @drichelson for your patience!

PR #903, opened 2019-04-18
Add warning about sidecar egress

frankbu:
Fixes https://github.com/istio/istio/issues/12237

### DIFF
--- a/networking/v1alpha3/sidecar.proto
+++ b/networking/v1alpha3/sidecar.proto
@@ -300,7 +300,7 @@ message IstioEgressListener {
   //
   // **WARNING:** The list of egress hosts in a `Sidecar` must also include
   // the Mixer control plane services if they are enabled. Envoy will not
-  // function properly otherwise. For example, add host
+  // be able to reach them otherwise. For example, add host
   // `istio-system/istio-telemetry.istio-system.svc.cluster.local` if telemetry
   // is enabled, `istio-system/istio-policy.istio-system.svc.cluster.local` if
   // policy is enabled, or add `istio-system/*` to allow all services in the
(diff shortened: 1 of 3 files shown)


linsun:
/lgtm
a little unfortunate user has to do this manually but ok with it being a temp issue till mixer v2 is in place.

cc @geeknoid @louiscryan

louiscryan:
/lgtm

agreed about the pain here. Is there a tracking issue for resolving

networking/v1alpha3/sidecar.proto
```diff
@@ -289,14 +289,22 @@ message IstioEgressListener {
...
   //
   // NOTE: Only services and configuration artifacts exported to the sidecar's
   // namespace (e.g., `exportTo` value of `*`) can be referenced.
   // Private configurations (e.g., `exportTo` set to `.`) will
   // not be available. Refer to the `exportTo` setting in `VirtualService`,
   // `DestinationRule`, and `ServiceEntry` configurations for details.
+  //
+  // **WARNING:** The list of egress hosts in a `Sidecar` must also include
+  // the Mixer control plane services if they are enabled. Envoy will not
+  // function properly otherwise. For example add host
```
rshriram: I would say that if telemetry is enabled, add istio-system/telemetry.. same for policy. Else it sounds like you always need this.

howardjohn: Agreed, "function properly" is a bit vague. Your app won't break, since you can still connect with Pilot

xeviknal:
Is this warning still valid on istio 1.5?

howardjohn:
Only if you use mixer

PR #905, opened 2019-04-23
WIP: authentication v1alpha2

diemtvu:
- Add v1alpha2 authentication API.
- Add v1beta1 for some common API elements, e.g `WorkloadSelector`, `StringMatch`.

[Design doc](https://docs.google.com/document/d/1ofYk-OLcDH7410QayB-uhUYrIaHBo_k_pXQv_hTGsmE/edit#)


diemtvu:
I'm still working on few details, but this is closed to the version to bring to the final review. Please take a look to make sure I do not miss any big use cases.

diemtvu:
+ @lei-tang

costinm:
@louiscryan  - on workload selection, we must use a consistent mechanism and find if this belong to sidecar or not.

diemtvu:
> Upgrade plan ? What are the differences ?
> Is it in scope for 1.2 ?

This API will be hidden until it's ready. The new version tries to align with label selector approach, which is an important criteria to stabilize and move it to beta, imo.

More details, including migration plan, are discussed in the [doc](https://docs.google.com/document/d/1ofYk-OLcDH7410QayB-uhUYrIaHBo_k_pXQv_hTGsmE/edit#)

authentication/v1alpha2/method.proto
```diff
@@ -0,0 +1,41 @@
...
+import "authentication/v1alpha2/mtls_method.proto";
+import "authentication/v1alpha2/jwt_method.proto";
+
+// AuthenticationMethod describes the authentication method to be enforced. This specification is
+// used for both peer and origin authentication (see more in `AuthentiationRule`), though there is
+// restriction what method can be used for each type of authentication. For example, as of speaking,
+// origin  authentication can use only JWT. This logic will be enforced by validation. System will
+// also properly ignore the invalid ones.
+message AuthenticationMethod {
+  oneof params {
```
yangminzhu: Should we consider not to use `oneof` in the new API? Here and other places.

common/v1beta1/match.proto
```diff
@@ -0,0 +1,38 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.common.v1beta1;
```
yangminzhu: What's the reason to name this new `common` package as `v1beta1`, why not start with `alpha`.

Makefile
```diff
@@ -95,6 +97,34 @@ generate: \
...
+	@$(docker_gen) $(gogofast_plugin) $(protoc_gen_docs_plugin)$(config_common_path) $^
+
+generate-common-python: $(config_common_pb_pythons)
+
+$(config_common_pb_pythons): $(config_common_protos)
+	## Generate python/istio_api/mcp/v1beta1/*_pb2.py
+	@$(docker_lock) status
+	@$(docker_gen) $(protoc_gen_python_plugin) $^
+
+clean-common:
```
yangminzhu: this target should also be added to line 429.

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+//         - prefix: "/get"
+//       method: example-jwt
+// ```
+message Match {
+  // Ports to match. If not specified, it matches to any port number.
+  // Note: these are workload ports, not service ports.
+  repeated uint32 ports = 1;
+
+  // Request path to match. Available for L7 authentication method only (e.g JWT). If not
+  // specified, it matches to any paths.
```
yangminzhu: I think it's worth to mention what happens when `paths` is set on non-HTTP ports (i.e. raw TCP protocol), Is the match never matched? Or is it considered a config error that rejects connection on the port (i.e. `deny-all`)?

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+// or origin) should be used as principal. By default, it uses peer.
+// For examples, see `AuthenticationPolicy`
+message PolicyContent {
+  // List of authentication methods that can be used for peer authentication.
+  // Rule will be examined in order. Each rule may contain match conditions.
+  // The first (and only) rule that have the match conditions satisfied will
+  // be activated, and the authentication result decides whether the request is
+  // allowed or denied. In other words, all other rules are ignored. If
+  // there is no rule meet the conditions, or if the list is empty,
+  // authentication is not required.
```
yangminzhu: Can we make it clear in the comment that the `peers` will be evaluated before `origins`.

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+  // authentication policy should be applied. If omitted, the authentication policy
+  // be applied to all workload instances in the same namespace.
+  istio.common.v1beta1.WorkloadSelector workload_selector = 1;
+
+  // REQUIRED. The authentication spec to be set for the selected workloads.
+  PolicyContent policy = 2;
+}
+
+// PolicyContent specifies the authentication requirements. It is composed of 2-part authentication:
+// - peer: verify caller service credentials. This part will set source.user
```
yangminzhu: I think `source.user` is deprecated so can we only mention `source.principal` here?
See https://docs.google.com/document/d/1IcR6C7j5Vl-4Jx8ervVT4lCTBDrBgsD7Uac77K5C5h4/

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+  // See `peers` for more details.
+  repeated AuthenticationRule origins = 3;
+
+  // Define whether peer or origin identity should be used for request principal.
+  // Default value is USE_PEER.
+  // If peer (or origin) identity is not available, either because of peer/origin
+  // authentication is not defined, or failed, principal will be left unset.
+  // In other words, binding rule does not affect the decision to accept or
+  // reject request.
+  PrincipalBinding principal_binding = 6;
```
yangminzhu: Can we make it clear in the doc that this sets the attribute `request.auth.principal`?

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+      // that is equivalent to `deny-all`.
+      string method = 1;
+
+      // Defines the required authentication method inline.
+      AuthenticationMethod spec = 2;
+    }
+  }
+
+  // REQUIRED. Authentication method to be used if the match conditions are satisfied.
+  AuthenticationMethodReference authentication = 2;
```
yangminzhu: Do we need `AuthenticationMethodReference`? can we just put the `method`/`spec` in the `AuthenticationRule`?

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+// $description: Authentication policy for Istio services.
+// $location: https://istio.io/docs/reference/config/istio.authentication.v1alpha2.html
+
+// This package defines user-facing authentication policy.
+package istio.authentication.v1alpha2;
+
+option go_package = "istio.io/api/authentication/v1alpha2";
+
+// AuthenticationPolicy describes authentication requirements for the workload(s) it is attached to.
+// Example of a simple policy that enable mTLS for all workloads in the namespace scope of the
```
lei-tang: "that enable mTLS" -> "that enables mTLS"

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+// policy. Note that `istio.mtls` is a name of the pre-define `AuthenticationMethod` for Istio
+// mutual TLS.
+//
```
lei-tang: "is a name of the pre-define" -> "is the name of a pre-defined".

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+// ```
+// spec:
+//   policy:
+//     peers:
+//     - authentication:
+//         method: "istio.mtls"
+// ```
+//
+// This example set end-user authentication for workloads that match label `app`=`foo`, using JWT
```
lei-tang: "This example set" -> "This example sets".

lei-tang: "using JWT" -> "to require a JWT."

common/v1beta1/match.proto
diemtvu: I think those are mature enough to be beta (given we are using them in several places already). On the other hand, I admit the choice of version is somewhat arbitrary. Waiting for more feedback.

Makefile
diemtvu: Done

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,191 @@
...
+  PrincipalBinding principal_binding = 6;
+}
+
+// AuthenticationRule describes match conditions and the corresponding authentication action. See
+// AuthenticationPolicy for example.
+message AuthenticationRule {
+  // Define the conditions when the authentication mechanism will be activated.
+  // All conditions must be satisfied (ANDed) for the authentication mechanism
+  // to be used. If empty, it matches all requests.
+  repeated Match match = 1;
```
liminw: Do we need `match` for every AuthenticationRule? Or per-Policy is enough?

common/v1beta1/match.proto
liminw: Same question. How do we determine the version here?

common/v1beta1/selector.proto
```diff
@@ -0,0 +1,27 @@
...
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.common.v1beta1;
+
+option go_package = "istio.io/api/common/v1beta1";
```
liminw: Same here. Why is this "beta"?

common/v1beta1/selector.proto
```diff
@@ -0,0 +1,27 @@
...
+
+// WorkloadSelector specifies the criteria used to determine if a resource can be applied to a
+// workload (proxy).
+message WorkloadSelector {
+  // REQUIRED: One or more labels that indicate a specific set of pods/VMs. The scope of
+  // label search is restricted to the configuration namespace in which the resource is present.
+  map<string, string> labels = 1;
```
liminw: Use "matchLabels" based on Istio Config Model proposal.

common/v1beta1/match.proto
```diff
@@ -0,0 +1,43 @@
...
+    // prefix-based match.
+    string prefix = 2;
+
+    // suffix-based match.
+    string suffix = 3;
+
+    // `ECMAscript` style regex-based match as defined by [`EDCA-262`](
+    // `http://en.cppreference.com/w/cpp/regex/ecmascript`).
+    // Example: "^/pets/(.*?)?"
+    string regex = 4;
```
liminw: Do we want to support `regex`? This provides more flexibility, but opens up more cases we need to defend in terms of security.

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,196 @@
...
+//
+// This example set end-user authentication for workloads that match label `app`=`foo`, using JWT
+// The JWT authentication details are defined directly in the policy (`inline`).
+// ```
+// spec:
+//   workloadSelector:
+//     app: foo
+//   policy:
+//     origins:
+//     - apply:
```
lei-tang: Suggest changing "apply" to "inlineMethod" since "method" is used to refer a non-inline authentication method.

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,196 @@
...
+//         spec:
+//           jwt:
+//             issuer: "https://securetoken.google.com"
+//             audiences:
+//             - "productpage"
+//             jwksUri: "https://www.googleapis.com/oauth2/v1/certs"
+// ```
+//
+// This last example shows a policy to set peer authentication with mTLS for all requests, and
```
lei-tang: Suggest changing "This last example" -> "This example".

authentication/v1alpha2/jwt_method.proto
```diff
@@ -0,0 +1,115 @@
...
+  // Example: `https://www.googleapis.com/oauth2/v1/certs`
+  //
+  // Note: Only one of jwks_uri and jwks should be used.
+  string jwks_uri = 3;
+
+  // JSON Web Key Set of public keys to validate signature of the JWT.
+  // See https://auth0.com/docs/jwks.
+  //
+  // Note: Only one of jwks_uri and jwks should be used.
+  string jwks = 10;
```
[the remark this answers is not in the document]
yangminzhu: It's needed to pass the jwks from Galley to Pilot when we change to use Galley to fetch it.

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,196 @@
...
+// This example set end-user authentication for workloads that match label `app`=`foo`, using JWT
+// The JWT authentication details are defined directly in the policy (`inline`).
+// ```
+// spec:
+//   workloadSelector:
+//     app: foo
+//   policy:
+//     origins:
+//     - apply:
+//         spec:
```
lei-tang: Suggest changing "spec" to "inlineMethod" since "method" is used to refer a non-inline authentication method.

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,196 @@
...
+// AuthenticationRule describes match conditions and the corresponding authentication action. See
+// AuthenticationPolicy for example.
+message AuthenticationRule {
+  // Define the conditions when the authentication mechanism will be activated.
+  // All conditions must be satisfied (ANDed) for the authentication mechanism
+  // to be used. If empty, it matches all requests.
+  repeated Match match = 1;
+
+  // AuthenticationMethodReference refers to `AuthenticationMethod` by name or define it inline.
+  message AuthenticationMethodReference {
```
lei-tang: Suggest changing "AuthenticationMethodReference" to "AuthenticationMethodSpec" since this message may contain a reference or an inline authentication method definition.

authentication/v1alpha2/mtls_method.proto
```diff
@@ -0,0 +1,42 @@
...
+
+option go_package = "istio.io/api/authentication/v1alpha2";
+
+// $hide_from_docs
+// Mutual TLS authentication parameters.
+message MutualTls {
+  // Defines the acceptable connection TLS mode.
+  enum Mode {
+    // Must never be set.
+    INVALID = 0;
```
lei-tang: Why "INVALID" mode is needed? It seems that the server-side-only TLS mode is missing.

authentication/v1alpha2/jwt_method.proto
```diff
@@ -0,0 +1,116 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+// $title: Authentication Policy v1alpha2
+// $description: Authentication policy for Istio services.
```
lei-tang: Suggest changing "Authentication policy for Istio services." to "Istio JWT authentication policy."

authentication/v1alpha2/method.proto
```diff
@@ -0,0 +1,42 @@
...
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+// $title: Authentication Policy v1alpha2
+// $description: Authentication policy for Istio services.
```
lei-tang: Suggest changing "Authentication policy for Istio services." to "Istio authentication method."

authentication/v1alpha2/jwt_method.proto
```diff
@@ -0,0 +1,116 @@
...
+// A JWT for any requests:
+//
+// ```yaml
+// issuer: https://example.com
+// audiences:
+// - bookstore_android.apps.googleusercontent.com
+//   bookstore_web.apps.googleusercontent.com
+// jwksUri: https://example.com/.well-known/jwks.json
+// ```
+message Jwt {
```
costinm: Type Jwt is well defined in standards - I don't think we should define our own message with different terms.

"iss", "sub", "aud" and the other terms are defined by the RFC, I think it's confusing to rename them.

authentication/v1alpha2/jwt_method.proto
```diff
@@ -0,0 +1,116 @@
...
+  // The service name will be accepted if audiences is empty.
+  //
+  // Example:
+  //
+  // ```yaml
+  // audiences:
+  // - bookstore_android.apps.googleusercontent.com
+  //   bookstore_web.apps.googleusercontent.com
+  // ```
+  repeated string audiences = 2;
```
costinm: I don't remember JWT supporting multiple audiences. Can you confirm ?

authentication/v1alpha2/jwt_method.proto
```diff
@@ -0,0 +1,116 @@
...
+  //
+  //     2) `access_token` query parameter (see
+  //     [URI Query Parameter](`https://tools.ietf.org/html/rfc6750#section-2.3`))
+
+  // JWT is sent in a request header. `header` represents the
+  // header name.
+  //
+  // For example, if `header=x-goog-iap-jwt-assertion`, the header
+  // format will be `x-goog-iap-jwt-assertion`: $JWT.
+  repeated string jwt_headers = 6;
```
costinm: Why would the header be part of the JWT message or spec ? How is it used ?

authentication/v1alpha2/jwt_method.proto
```diff
@@ -0,0 +1,115 @@
...
+  //
+  // For example, if `header=x-goog-iap-jwt-assertion`, the header
+  // format will be x-goog-iap-jwt-assertion: <JWT>.
+  repeated string jwt_headers = 6;
+
+  // JWT is sent in a query parameter. `query` represents the
+  // query parameter name.
+  //
+  // For example, `query=jwt_token`.
+  repeated string jwt_params = 7;
```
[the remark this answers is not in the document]
costinm: OAuth standard defines how tokens should be passed, any reason to deviate from standards ?

authentication/v1alpha2/method.proto
```diff
@@ -0,0 +1,42 @@
...
+import "authentication/v1alpha2/mtls_method.proto";
+import "authentication/v1alpha2/jwt_method.proto";
+
+// $hide_from_docs
+// AuthenticationMethod describes the authentication method to be enforced. This specification is
+// used for both peer and origin authentication (see more in `AuthentiationRule`), though there is
+// restriction what method can be used for each type of authentication. For example, as of speaking,
+// origin  authentication can use only JWT. This logic will be enforced by validation. System will
+// also properly ignore the invalid ones.
+message AuthenticationMethod {
```
costinm: What is the meaning of "Method" ? Any relevant standards using this terms ?

authentication/v1alpha2/method.proto
costinm: Why one of ? Can you have both ? What about other auth methods that IStio may support - any reason to restrict to those ?

authentication/v1alpha2/mtls_method.proto
```diff
@@ -0,0 +1,42 @@
...
+// $description: Authentication policy for Istio services.
+// $location: https://istio.io/docs/reference/config/istio.authentication.v1alpha2.html
+
+package istio.authentication.v1alpha2;
+
+option go_package = "istio.io/api/authentication/v1alpha2";
+
+// $hide_from_docs
+// Mutual TLS authentication parameters.
+message MutualTls {
```
costinm: Can someone explain why we need to rev the API ? If we do - maybe we can use the opportunity to have a better UX and naming...

authentication/v1alpha2/policy.proto
```diff
@@ -0,0 +1,196 @@
...
+//         - prefix: "/create"
+//         - prefix: "/update"
+//       authentication:
+//         method: "my-jwt"
+// ```
+message AuthenticationPolicy {
+  // Criteria used to select the specific set of pods/VMs on which this
+  // authentication policy should be applied. If omitted, the authentication policy
+  // be applied to all workload instances in the same namespace.
+  istio.common.v1beta1.WorkloadSelector workload_selector = 1;
```
costinm: Where is this used ? I don't think Sidecar or Gateway use it - did we launch it for anything ? 

I would prefer if all APIs used the same terms to select workloads - Sidecar terms in particular.

common/v1beta1/match.proto
```diff
@@ -0,0 +1,43 @@
...
+// $description: Common Istio APIs.
+// $location: https://istio.io/docs/reference/config/istio.common.v1alpha1.html
+
+// This package defines common Istio APIs.
+package istio.common.v1beta1;
+
+option go_package = "istio.io/api/common/v1beta1";
+
+// Describes how to match a given string. Match is case-sensitive.
+message StringMatch {
```
costinm: Should this match exactly what we support for VirtualService  ? Any reason to use different terms and names ?  Seems bad UX.

common/v1beta1/selector.proto
```diff
@@ -0,0 +1,27 @@
...
+
+syntax = "proto3";
+
+package istio.common.v1beta1;
+
+option go_package = "istio.io/api/common/v1beta1";
+
+// WorkloadSelector specifies the criteria used to determine if a resource can be applied to a
+// workload (proxy).
+message WorkloadSelector {
```
costinm: Why ? How is it common - and why are we adding yet-another way to select workloads, specific to auth ? 
I'm not aware of any plan to rev the networking APIs again, or make arbitrary changes.  

Please use the exact same syntax that Sidecar is using in your Policy - I don't think there is any need to have a common proto to share a map.

PR #906, opened 2019-04-23
removed port name from the port selector

GregHanson:
coincides with https://github.com/istio/istio/pull/13054

resolves: https://github.com/istio/istio/issues/8754


rshriram:
to master

networking/v1alpha3/virtual_service.proto
```diff
@@ -1105,12 +1105,11 @@ message HTTPFaultInjection {
 // PortSelector specifies the number of a port to be used for
 // matching or selection for final routing.
 message PortSelector {
-  oneof port {
-    // Valid port number
-    uint32 number = 1;
-    // $hide_from_docs
-    string name = 2;
-  }
+  // Valid port number
```
costinm: Can you add a test ( can be in istio/istio ) to verify that a VirtualService that used "name" doesn't break ? 
I don't remember what was the use case and if we ever told users about it - was it never added, deprecated ? 

I assume it is not used anywhere in the code ?

geeknoid:
@GregHanson Do you want to push this over the finish line, or should we close this PR?

rshriram:
please redo pr on master

PR #911, opened 2019-04-26
Adding envoy gRPC access log service configuration to MeshConfig

hanahmily:
This change is in preparation for adding support for the envoy http gRPC access log service (ALS) API. In order to support envoy `google_grpc` client SSL, there is a `credentials` field to enable [sslcredentials](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/grpc_service.proto#core-grpcservice-googlegrpc-sslcredentials).


mesh/v1alpha1/config.proto
```diff
@@ -140,6 +140,25 @@ message MeshConfig {
   // Default value is text.
   AccessLogEncoding access_log_encoding = 27;
 
+  message EnvoyAccesslogService {
```
rshriram: Google grpc service

hanahmily: done

douglas-reid:
What is the reason for not adding this option in the same place (`proxy.proto`) as the Metrics Service address?  And should the Metrics Service address be updated to support this Grpc message?

mesh/v1alpha1/config.proto
```diff
@@ -140,6 +140,26 @@ message MeshConfig {
   // Default value is text.
   AccessLogEncoding access_log_encoding = 27;
 
+  message GoogleGrpcService {
```
costinm: In general we are using DestinationRules to specify how to connect to various clusters - including internal ones as Mixer, etc.

There is no need to define another way.

costinm: Also: if DR lacks anything for allowing definition of such service, we should probably add it.

mesh/v1alpha1/config.proto
```diff
@@ -140,6 +140,26 @@ message MeshConfig {
...
+      string cert_chain = 3;
+    }
+    // Optional: SSL credentials. If empty, client will use insecure gRPC channel.
+    SSLCredentials credentials = 2;
+  }
+
+  // The Envoy gRPC Access Log Service (ALS) implementation.
+  // See [Access Log Service](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/accesslog/v2/als.proto#config-accesslog-v2-httpgrpcaccesslogconfig)
+  // for details about Envoy's gRPC Access Log Service API.
+  GoogleGrpcService envoy_accesslog_service = 37;
```
costinm: @douglas-reid  @ostromart 
We are moving to operator and new modular install - and attempting to decouple telemetry and other settings to their own specific configs.

As was discussed, the new settings will be initially based on values.yaml but will move to separate CRDs. 
Access log - format, settings, destination - is already defined in values.yaml, and logically belongs to telemetry. 
It is inconsistent to add it to mesh config

douglas-reid: I agree that mesh config doesn't seem to be the correct place. However, there is precedent to add this to the `proxy.proto` (with metrics service).  We should clarify how that should be handled.

costinm:
There is precedent for a lot of things - but install inconsistency and
complexity is a major problem we need to fix, not make worse.

mesh/v1alpha1/config.proto
rshriram: Envoy’s google grpc client does not require explicit cluster declaration. Filters can talk directly to remote service. One less cluster to worry about.

mesh/v1alpha1/config.proto
rshriram: Adding to proxy proto is so that we can add a static cluster in bootstrap. But unlike metrics service which is not referred to any where in the envoy listener config, the access log service has to be referred to everywhere in http connection manager and network listener filter stack. 

Naturally referring to a service typically means pointing to a cluster in envoy. If we do the proxy proto way, we add a cluster with fixed name. We then hardcode that cluster name in pilot. And use that cluster name when setting up the access log service. That’s a bit clunky. 

The google grpc client in envoy allows us to define access log service endpoint in place without having to define clusters. It has the bonus of not requiring hard coded cluster names. Pilot can ship this configuration as part of lds and it works nicely.

rshriram: Btw this info is required at runtime as part of lds. So values.yaml won’t help unless you want to shoehorn an entire proto into annotation or environment variable.

costinm:
If it is used for bootstrap - there is no other choice but map it to env
that is used to generate bootstrap config.

If it is used by pilot - it can read values yaml, or it can get it from
mesh config or env or CLI - which all get generated from values.yaml -
using some tool if helm is not available.

Again- the problem is not how it gets into pilot or agent, but what will
the user configure. We have many related configs - and several efforts to
clean up the inconsistent settings ( and settings used only by tests or not
even). For telemetry/logs/tracing - Doug is working on a proposal, Martin
on another - and I'm trying to somehow be able to upgrade from 1.1 to 1.2
and the new installer without having everything break again.

costinm:
Btw - the discussion on patches (started with the filter) is yet another
aspect of the puzzle, as well as adding kustomize as an option.

Mesh config is a mess that was not exposed to users so far and I was hoping
to deprecate it for operator settings without another upgrade nightmare.

rshriram:
/hold cancel

Now that we have decided to allow changes to mesh config (and have already allowed a PR in, for mixer settings), letting this through. At somepoint, we need to recategorize all these options but thats no reason to hold up this PR.

douglas-reid:
@rshriram when was that decision made? is it recorded somewhere? i'm looking for guidance on related issues moving forward.

mesh/v1alpha1/config.proto
```diff
@@ -140,6 +140,26 @@ message MeshConfig {
...
+      string cert_chain = 3;
+    }
+    // Optional: SSL credentials. If empty, client will use insecure gRPC channel.
+    SSLCredentials credentials = 2;
+  }
+
+  // The Envoy gRPC Access Log Service (ALS) implementation.
+  // See [Access Log Service](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/accesslog/v2/als.proto#config-accesslog-v2-httpgrpcaccesslogconfig)
+  // for details about Envoy's gRPC Access Log Service API.
+  GoogleGrpcService envoy_accesslog_service = 40;
```
douglas-reid: @rshriram shouldn't this have been `envoy_access_log_service` to match `access_log_encoding` which is just above?  It'd be great to have some consistency in field names, if nothing else, in this config file.

mesh/v1alpha1/config.proto
douglas-reid: @rshriram I still don't understand why the metrics_service can't use a google grpc client. It is certainly supported in Envoy, to my understanding. I think it is inconsistent to expose one of the telemetry services with a simple uri and for the other to define it as a google grpc endpoint. Can you provide more clarification here?

hzxuzhonghu:
@douglas-reid I think this agreement is from slack channel, many guys are there talking IIRC

douglas-reid:
@hzxuzhonghu ok. we've had a long-standing policy that decisions should not be made in slack. slack is not the source of record for important decisions (it doesn't even save history at this point). we should find somewhere else to document this decision and a different way to communicate it.

PR #912, opened 2019-04-26
proxy: Reserve a inbound port field in mesh config

silentdai:
**Context**
Pilot and proxy used to assume the inbound and outbound traffic are intercepted using the same port which is 15001. By sharing the same port we are hitting problems including below

- listener mismatch: When there is no listener matching the intercepted traffic, inbound and outbound share the same decision. IIRC the current strategy is to assume it's outbound traffic. We definitely need 2 different strategies.
- Performance: inbound and outbound listeners should never have the same destination. We could save the matching time.
- Risk in develop cycle: It's relatively hard to make one flow working when the [ongoing bind_to_port field deprecation as per lds api](See https://github.com/istio/istio/pull/13666).

**Description**
This PR add a separate port for inbound traffic. Will choose 15006 as there is no conflict in existing code base.

- How to set the value?
Similar to `PilotListenPort`, there is little benefit to negotiate in the xds protocol. The value is shared by all the pilot and sidecar proxy. The value should be populated from `values.yaml`

- Upgrade/Downgrade
As described above, the field is considered as const value shared by pilot and sidecar proxy. Again, see how the existing `ProxyListenPort` works.
It's another topic how to rolling out the port split. 


rshriram:
So this should go through the same "config review" that @costinm  is talking about. Especially all the FUD about upgrade and other pains pointed out in #911 applies here as well. 

So please resolve this discussion in a fair way. :)

silentdai:
@rshriram  Thanks for the pointer! LMS what is needed...

silentdai:
The port itself is an easy one. Personally I don't even assume it is configurable. Please let me know what you think @rshriram @costinm

rshriram:
It should be configurable to maintain consistency with the proxyListenPort we have. So lets resolve the mesh config api update issue properly.

silentdai:
@rshriram is there a discussion making `proxyListenPort` configurable? Configurable at which phase? AFAIK I don't see any code changing proxyListenPort at runtime. I only find it configurable from `values.yaml`
This PR is not suppose to address the configurable issue. I just want `proxy_inbound_listen_port` follow the same configurability as `proxyListenPort`

silentdai:
Discussed with @costinm  and I am going to abandon this PR. Some notes here for future track:
0. Istio will reserve the 15000ish ports
1. The existing `ProxyListenPort` is never assumed configurable although it's in the API. No code assumes it's can be some other value. No test coverage on changing the value.
2. Align with the `ProxyListenPort` the `proxy_inbound_listen_port` will be hard coded.
3. `ProxyListenPort` will be moved out of API some other day.

rshriram:
@silentdai can you rebase this PR and fix your https://github.com/istio/istio/pull/13666 to use the global setting instead of yet another proxy level metadata?

silentdai:
@rshriram I'd love to as it was the initial plan. Just one thing: before I make the change: is there any discussion/issue/document showing we do need to make it a API?

rshriram:
proxyListenPort is already part of the API! . So if supplying a separate inboundListenPort will make things extremely confusing not to mention, make it more complicated

silentdai:
>> `proxyListenPort` is already part of the API! . So if supplying a separate inboundListenPort will make things extremely confusing not to mention, make it more complicated
True. It's the fact. However, I have the impression that it shouldn't be and it will be removed from the API.
If that happens, the `inboundListenPort` as API won't survive, either.

rshriram:
> True. It's the fact. However, I have the impression that it shouldn't be and it will be removed from the API. If that happens, the inboundListenPort as API won't survive, either.

Its not going to be removed from the API anytime soon. Lets please not overcomplicate this, for the sake of extreme flexibility. Lets go with the simpler global option first

PR #914, opened 2019-04-30
adding report_batch_max_request and report_batch_max_time to TransportConfig

huang195:
Mixerclient has the capability to send telemetry data in batches. In Isito 1.1, the default values are every 100 requests or every 1 second, whichever one comes first. Tuning these 2 parameters has a big impact on long-tail latency and CPU usage (on both mixer and envoy). Even though this capability has been implemented, the control knob to tune these 2 parameters is not there, i.e., everyone is forced to use the default values of 100 req/1 sec, regardless the application they are deploying. The only control knob we have today is to turn on and off batch reporting, which is too coarse in most cases. When batch reporting is enabled (default behavior), this PR allows fine tuning of how batch reporting is done per Istio deployment.


huang195:
The tunable parameters are to resolve the problem described in https://github.com/istio/proxy/issues/2185, as it's difficult to use default values for all situations.

mixer/v1/config/client/client_config.proto
```diff
@@ -145,6 +145,14 @@ message TransportConfig {
   // includes the "source.ip" and "source.uid" attributes. These
   // attributes are consumed by the proxy in front of mixer.
   Attributes attributes_for_mixer_proxy = 8;
+
+  // When disable_report_batch is false, this value specifies after how many
+  // requests a batch report is sent
+  uint32 report_batch_max_entries = 9;
```
mandarjog: how are defaults handled here?

mandarjog: `report_batch_max_entries == 0` will use the hardcoded defaults.

mixer/v1/config/client/client_config.proto
```diff
@@ -145,6 +145,14 @@ message TransportConfig {
...
+
+  // When disable_report_batch is false, this value specifies after how much
+  // time a batch report is sent
+  google.protobuf.Duration report_batch_max_time = 10;
```
mandarjog: Same comment as above.
Please add this to the block comments in this field and above.

mixer/v1/config/client/client_config.proto
huang195: comment is updated per suggestion

mandarjog: Nit: this value specifies the maximum number of requests that are batched in a report.

huang195:
Added more description about the 2 added fields. @mandarjog let me know if anything else is still not clear from the comments.

huang195:
@mandarjog ping

mixer/v1/config/client/client_config.proto
```diff
@@ -145,6 +145,18 @@ message TransportConfig {
   // includes the "source.ip" and "source.uid" attributes. These
   // attributes are consumed by the proxy in front of mixer.
   Attributes attributes_for_mixer_proxy = 8;
+
+  // When disable_report_batch is false, this value specifies the maximum number
+  // of requests that are batched in report. If left unspecified, the default value
```
mandarjog: Nit: that are batched in “a” report.

PR #915, opened 2019-05-01
correct default values in the destination rule

ramaraochavali:
The destination rule proto default values for `max_retries` ,  `http1_max_pending_requests` and `http2_max_requests` is not as per the current [implemementation](https://github.com/istio/istio/blob/7d61672e25a28b0fa49ef2091f5b49703307572f/pilot/pkg/networking/core/v1alpha3/cluster.go). This PR corrects the docs.


rshriram:
Oh.. good catch! thanks.

PR #916, opened 2019-05-02
add default max connections to docs

dwradcliffe:
Simple change to add the default number of max connections to the docs.

ref: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster/circuit_breaker.proto#cluster-circuitbreakers-thresholds

cc: @Stono


frankbu:
@dwradcliffe You need to run `scripts/generate-protos.sh` to regen the html, etc. with your change.

https://github.com/istio/api/blob/master/README.md#updating

dwradcliffe:
@frankbu Oops, thanks! Fixed now.

PR #917, opened 2019-05-06
Authentication v1alpha2 policy

diemtvu:
New authentication policy APIs:
* Align with common use of label selectors.
* Simpler and clearer definition of authentication input and output.

[Design doc](https://docs.google.com/document/d/1ofYk-OLcDH7410QayB-uhUYrIaHBo_k_pXQv_hTGsmE/edit#)

PR #918, opened 2019-05-07
Add authorization policy v1beta1

yangminzhu:
This PR adds the authorization v1beta1 API.

Design doc: https://docs.google.com/document/d/1diwa9oYVwmLtarXb-kfWPq-Ul3O2Ic3b8Y21Z3R8DzQ.

https://github.com/istio/istio/issues/12394


rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+  // Enable Istio Authorization when true.
+  bool enabled = 1;
+
+  // Target defines workload or namespaces.
+  message Target {
+    // A list of namespaces.
+    string namespaces = 2;
+
+    // A list of workload instances.
+    WorkloadSelector workload_selector = 3;
```
pitlv2109: Have we reached an agreement on the name for this? The doc and examples here use `selector` so it's pretty confusing for the users.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+// labels attached to the pod/VM, or any other info that the proxy provides
+// to Istio during the initial handshake. If multiple conditions are
+// specified, all conditions need to match in order for the workload instance to be
+// selected. Currently, only label based selection mechanism is supported.
+message WorkloadSelector {
+  // One or more labels that indicate a specific set of pods/VMs on which
+  // this sidecar configuration should be applied. The scope of label
+  // search is restricted to the configuration namespace in which the the
+  // resource is present.
+  map<string, string> labels = 1;
```
pitlv2109: Same here. Doc uses matchLabels.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+package istio.rbac.v1alpha2;
+
+// $hide_from_docs
+// $title: RBAC
+// $description: Configuration for Role Based Access Control.
+// $location: https://istio.io/docs/reference/config/authorization/istio.rbac.v1alpha2.html
+
+// Istio RBAC (Role Based Access Control) defines ServiceRole, AuthorizationPolicy and AuthorizationConfig objects.
+//
+// An AuthorizationConfig is a cluster-scope resource that enables or disables the role based access control for
```
pitlv2109: This does not just only replace `RbacConfig` but also support workload level (besides namespace level), right? This adds flexibility, but I wonder if this has been approved. I don't see it in the design doc.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+//   enabled: true
+//   inclusions:
+//   - namespaces: ["bookinfo"]
+//   exclusions:
+//   - workload_selectors:
+//       labels:
+//         app: test
+// ```
+//
+// A ServiceRole defines a list of permissions. Below is an example of ServiceRole object "productpage-viewer", which
```
pitlv2109: Move ServiceRole comments to its definition?

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+// metadata:
+//   name: productpage-viewer
+//   namespace: bookinfo
+// spec:
+//   permissions:
+//   - paths: ["/info"]
+//     methods: ["GET", "HEAD"]
+// ```
+//
+// An AuthorizationPolicy grants a list of principals to a list of roles. Below is an example of AuthorizationPolicy
```
pitlv2109: Move AuthorizationPolicy comments to its definition?

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+  //      app: product
+  //  allow:
+  //  - principals:
+  //    - names: ["cluster.local/ns/default/sa/frontend"]
+  //    inline_role:
+  //      permissions:
+  //      - paths: ["/info"]
+  //        methods: ["GET"]
+  // ```
+  ServiceRole inline_role = 3;
```
pitlv2109: Shouldn't it be `permissions` for inline role definition?

rbac/v1alpha2/rbac.proto
yangminzhu: it replaces the `service` in `RbacConfig`, this is part of the service -> workload label migration.

rbac/v1alpha2/rbac.proto
yangminzhu: I think `inline_role` is a better name, it's the natural comparison to the `roles` above. the same `role` indicating they're referring to the same thing. `permission` doesn't have this indication and makes people wonder it's different from the `roles`.

rbac/v1alpha2/rbac.proto
yangminzhu: either name is okay, I don't see too much difference or confusion.

rbac/v1alpha2/rbac.proto
pitlv2109: I see. I didn't read the API carefully so I didn't know `RbacConfig` support service level. I thought it's just namespace level.

rbac/v1alpha2/rbac.proto
yangminzhu: the `match` seems redundant @liminw

rbac/v1alpha2/rbac.proto
pitlv2109: Just thought it should be consistent within Istio API. I think many APIs are using different names for workload selector.

pitlv2109:
Link to https://github.com/istio/istio/issues/12394 for tracking

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+// apiVersion: "rbac.istio.io/v1alpha2"
+// kind: AuthorizationPolicy
+// metadata:
+//   name: productpage
+//   namespace: bookinfo
+// spec:
+//   workload_selector:
+//     app: productpage
+//   principals:
+//   - names: ["alice@yahoo.com"]
```
liminw: We do not have "names" field under principals. See https://docs.google.com/document/d/1diwa9oYVwmLtarXb-kfWPq-Ul3O2Ic3b8Y21Z3R8DzQ/edit#heading=h.wyshhqz8b0rt

rbac/v1alpha2/rbac.proto
liminw: I prefer using "selector" to be consistent with Istio config model proposal.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+
+// $hide_from_docs
+// AuthorizationPolicy binds roles or permissions to a list of principals for a workload.
+message AuthorizationPolicy {
+  // Optional. One or more labels that indicate a specific set of pods/VMs
+  // on which this authorization policy should be applied. Leave this empty to
+  // select all pods/VMs.
+  // The scope of label search is platform dependent. On Kubernetes, for example,
+  // the scope includes pods running in the same namespace as the authorization policy itself.
+  WorkloadSelector workload_selector = 1;
```
liminw: "workload_selector" => "selector".

rbac/v1alpha2/rbac.proto
liminw: Let's use "matchLabels" to be consistent with Istio config model.

rbac/v1alpha2/rbac.proto
liminw: I prefer "permissions". "ServiceRole" is essentially a list of "permissions". Using "permissions" should be clear enough.

liminw: Let's also call out that only "oneof" "roles" or "permissions" should be used.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,282 @@
...
+  // metadata:
+  //  name: product-frontend
+  //  namespace: ns1
+  // spec:
+  //  selector:
+  //    labels:
+  //      app: product
+  //  allow:
+  //  - principals:
+  //    - names: ["cluster.local/ns/default/sa/frontend"]
```
liminw: Same here. We do not support "names" field any more.

rbac/v1alpha2/rbac.proto
yangminzhu: The `Sidecar` API actually uses `workload_selector`, if we want to be consistent within Istio API, shouldn't we use `workload_selector` also?

rbac/v1alpha2/rbac.proto
yangminzhu: I still think `inline_role` is clearer especially if you look at it together with the `roles` field.

By the way, do we actually want `ServiceRole permission` or `repeated Permission permission`? If it's the latter, the comments will be updated to mention it's just inline permissions, as there is any connection with `role` in this case.

rbac/v1alpha2/rbac.proto
yangminzhu: I kept it in both places.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,285 @@
...
+package istio.rbac.v1alpha2;
+
+// $hide_from_docs
+// $title: RBAC
+// $description: Configuration for Role Based Access Control.
+// $location: https://istio.io/docs/reference/config/authorization/istio.rbac.v1alpha2.html
+
+// Istio RBAC (Role Based Access Control) defines ServiceRole, AuthorizationPolicy and AuthorizationConfig objects.
+//
+// An AuthorizationConfig is a cluster-scope resource that enables or disables the role based access control globally.
```
liminw: Shall we make it a namespace-scope CRD, and put it in root namespace?

yangminzhu:
@costinm 
This is targeted for 1.3 so no rush for 1.2.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,285 @@
...
+// apiVersion: "rbac.istio.io/v1alpha2"
+// kind: AuthorizationConfig
+// metadata:
+//   name: default
+// spec:
+//   enabled: true
+//   inclusions:
+//   - namespaces: ["bookinfo"]
+//   exclusions:
+//   - selectors:
```
liminw: selectors->selector

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,285 @@
...
+// apiVersion: "rbac.istio.io/v1alpha2"
+// kind: AuthorizationPolicy
+// metadata:
+//   name: productpage
+//   namespace: bookinfo
+// spec:
+//   selector:
+//     app: productpage
+//   principals:
+//   - peers: ["alice@yahoo.com"]
```
liminw: Maybe this "alice@yahoo.com" should be "requests". It represents an end user.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,285 @@
...
+  // Enable Istio Authorization when true.
+  bool enabled = 1;
+
+  // Target defines workload or namespaces.
+  message Target {
+    // A list of namespaces.
+    string namespaces = 2;
+
+    // A list of workload instances.
+    WorkloadSelector selector = 3;
```
liminw: When you use "selector" to refer to a workload instance, you have to add "namespace" information here as well. For example, "app B in namespace ns1". This is quite redundant. Maybe we can support enabling RBAC on namespaces only, until we come up with a better way to representing workloads/"services" later. What do you think?

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,285 @@
...
+  // Optional. Additional constraints of the binding. The constraints are ANDed together.
+  repeated Constraint constraints = 4;
+}
+
+// $hide_from_docs
+// Principal defines an identity.
+message Principal {
+  // Optional. A list of peer identities. This is matched to the `source.principal` attribute.
+  // Exact, prefix and suffix matches are supported.
+  repeated string peers = 1;
```
liminw: @wora We use "principals/peer" to refer to source.principal, "principals/request" to refer to request.auth.principal. Are you ok with the names, or do you have better suggestions?

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,285 @@
...
+// This is forked from the networking/v1alpha3/sidecar.proto to avoid a direct
+// dependency from the authorization API on networking API.
+// WorkloadSelector specifies the criteria used to determine if the Gateway
+// or Sidecar resource can be applied to a proxy. The matching criteria
+// includes the metadata associated with a proxy, workload instance info such as
+// labels attached to the pod/VM, or any other info that the proxy provides
+// to Istio during the initial handshake. If multiple conditions are
+// specified, all conditions need to match in order for the workload instance to be
+// selected. Currently, only label based selection mechanism is supported.
+message WorkloadSelector {
```
liminw: It seems that Diem is adding a common selector definition in https://github.com/istio/api/pull/917. We can switch to use that one once it is merged.

costinm:
Great, but please hold it until we cut the branch, to avoid confusion. And
to allow people to focus on tests for 1.2 instead of getting distracted, we
have large gaps in auth testing (sds, automating, scale)

rbac/v1alpha2/rbac.proto
yangminzhu: Yes, we can use the same selector once it's merged.

rbac/v1alpha2/rbac.proto
yangminzhu: I think it's not that redundant as customer just need to specify the `namespace` field, the `namespace` information is always needed if customer wants to specify workload in some namespace, either here or somewhere else.

Also note it will be a break change if we don't support `selector`, (there is a `service` field in the `RbacConfig`).

I'm okay to remove this at this point though, we can add it back if it's needed.

rbac/v1alpha2/rbac.proto
yangminzhu: I changed it to `requests`, but again, I feel either the `peers` and `requests` are confusing. I don't know what does it mean by only looking at the field name, and I have to keep it in mind after reading the comments.

rbac/v1alpha2/rbac.proto
yangminzhu: Updated the comment.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,250 @@
...
+// $hide_from_docs
+// $title: RBAC
+// $description: Configuration for Role Based Access Control.
+// $location: https://istio.io/docs/reference/config/authorization/istio.rbac.v1alpha2.html
+
+// Istio RBAC (Role Based Access Control) defines ServiceRole, AuthorizationPolicy and AuthorizationConfig objects.
+//
+// An AuthorizationConfig is to enable or disable the role based access control globally.
+// There should be at most 1 AuthorizationConfig resource with name `default` defined in the
+// Istio root namespace (e.g. istio-system).
```
liminw: istio-system -> istio-config

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,250 @@
...
+// An AuthorizationConfig is a resource that enables or disables the role based access control globally.
+// There must be at most one AuthorizationConfig resource defined in the Istio root namespace with name `default`.
+message AuthorizationConfig {
+  // Enable Istio Authorization when true.
+  bool enabled = 1;
+
+  // Target defines namespaces.
+  message Target {
+    // A list of namespaces.
+    string namespaces = 2;
```
liminw: namespaces = 1?

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,250 @@
...
+// AuthorizationPolicy binds roles or permissions to a list of principals for a workload.
+message AuthorizationPolicy {
+  // Optional. One or more labels that indicate a specific set of pods/VMs
+  // on which this authorization policy should be applied. Leave this empty to
+  // select all pods/VMs.
+  // The scope of label search is platform dependent. On Kubernetes, for example,
+  // the scope includes pods running in the same namespace as the authorization policy itself.
+  WorkloadSelector selector = 1;
+
+  // A list of principals that would be granted the permission to access the workload.
```
liminw: Assigns permissions to principals.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,250 @@
...
+  repeated Permission permissions = 3;
+
+  // Constraint define a key-values pair.
+  message Constraint {
+    // Key of the constraint. This is the name of an Istio attribute.
+    string key = 1;
+
+    // List of valid values of the constraint.
+    // Exact match, prefix match, and suffix match are supported.
+    repeated string values = 2;
```
liminw: Do we want to use StringMatch instead of string? It is more consistent with other APIs. 

Example: 
prefix: /info/
instead of "/info/*"

yangminzhu: Yes, we can.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,250 @@
...
+// ```yaml
+// apiVersion: "rbac.istio.io/v1alpha2"
+// kind: AuthorizationPolicy
+// metadata:
+//   name: productpage
+//   namespace: bookinfo
+// spec:
+//   selector:
+//     app: productpage
+//   principals:
```
pitlv2109: Add a `-` before `principals` to indicate list.
Also add `allow` to this example to be consistent with the API definition, i.e. `repeated ServiceRoleBinding allow = 2`

rbac/v1alpha2/rbac.proto
liminw: Let's rename it to "sources", to match "source.principal".

yangminzhu: Updated.

rbac/v1alpha2/rbac.proto
yangminzhu: Thanks, updated.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,269 @@
...
+//     roles: ["productpage-viewer"]
+// ```
+option go_package="istio.io/api/rbac/v1alpha2";
+
+// $hide_from_docs
+// An AuthorizationConfig is a resource that enables or disables the role based access control globally.
+// There must be at most one AuthorizationConfig resource defined in the Istio root namespace with name `default`.
+message AuthorizationConfig {
+  // Enable Istio Authorization when true.
+  bool enabled = 1;
```
liminw: Why do we change from enum "mode" to a boolean "enabled"? I think using enum is clearer on the state.

rbac/v1alpha2/rbac.proto
```diff
@@ -0,0 +1,269 @@
...
+  // Optional. Additional constraints of the binding. The constraints are ANDed together.
+  repeated Constraint constraints = 4;
+}
+
+// $hide_from_docs
+// Principal defines an identity.
+message Principal {
+  // Optional. A list of source identities. This is matched to the `source.principal` attribute.
+  // Exact, prefix and suffix matches are supported.
+  repeated string sources = 1;
```
liminw: Let's use StringMatch for all the fields that support prefix/suffix match.

yangminzhu: I thought about this but not sure the user experience would be good, it basically mean part of the fields in `Permission`/`Principal` will be `StringMatch` and the others will still be `string`.

rbac/v1alpha2/rbac.proto
yangminzhu: I want to simplify this as it's not clearer than a single bool flag, and we have seen many cases that user specifying the `ON`, `OFF` string without quotes leading to errors in the yaml parsing.
The other problem is it's simply to long to type (think about `ON_WITH_INCLUSION` and `ON_WITH_EXCLUSION` compared to `true` and `false`)

pitlv2109:
Just to double check, are we waiting for 1.2 release before we check in this PR?

rbac/v1alpha2/rbac.proto
yangminzhu: Discussed offline to keep using `string` at the moment.

pitlv2109:
I think we can update the API now, given that the design has been approved?

yangminzhu:
@pitlv2109 Yes, will update the PR soon.

yangminzhu:
@liminw @pitlv2109  PTAL, thanks.

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,282 @@
...
+//    to:
+//    - operation:
+//        methods: [“GET”]
+//        paths: ["/info*"]
+//    - operation:
+//        methods: ["POST"]
+//        paths: ["/data"]
+//    when:
+//    - key: request.auth.claims[iss]
+//      values: [“istio.io”]
```
liminw: NIT: Let's use "https://accounts.google.com" as example for issuer.

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,282 @@
...
+// ```
+//
+// Access control is enabled on a workload if there is any authorization policies selecting
+// the workload. When access control is enabled, the default behavior is deny (deny-by-default)
+// which means requests to the workload will be rejected if the request is not allowed by any of
+// the authorization policies selecting the workload.
+//
+// Multiple authorization policies selecting the same workload are OR-ed together.
+// The effect of authorization policy is additive, more authorization policies
+// always allow more sources to access the workload.
```
liminw: Change this paragraph to: 

Currently AuthorizationPolicy only supports "ALLOW" action. This means that if multiple authorization policies apply to the same workload, the effect is additive.

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,282 @@
...
+//
+// Authorization policy could be specified in workload, namespace and mesh level.
```
liminw: what about "selector" is not empty and the namespace is root namespace?

I would use the description from "policy scope" section in the design doc (https://docs.google.com/document/d/1diwa9oYVwmLtarXb-kfWPq-Ul3O2Ic3b8Y21Z3R8DzQ/edit#heading=h.wq9jh71vsoq0):

Policy scope (target) is determined by “metadata/namespace” and an optional “selector”. 
* “metadata/namespace” tells which namespace the policy applies. If set to root namespace, the policy applies to all namespaces in a mesh. The value of “root namespace” is configurable, and the default is “istio-config”. 
* workload “selector” can be used to further restrict where a policy applies.

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,282 @@
...
+// The following namespace level authorization policy selects all workloads in namespace foo.
+//
+// ```yaml
+// apiVersion: security.istio.io/v1beta1
+// kind: AuthorizationPolicy
+// metadata:
+//  name: policy
+//  namespace: foo
+// spec:
+//  selector:
```
liminw: We do not need "selector" in this case.

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,282 @@
...
+  // Optional. A list of request identities (i.e. “iss/sub” claims), which
+  // matches to the “request.auth.principal” attribute.
+  //
+  // If not set, any request principal is allowed.
+  repeated string request_principals = 2;
+
+  // Optional. A list of namespaces, which matches to the “source.namespace”
+  // attribute.
+  //
+  // If not set, any namespace is allowed. Must be used only when mTLS is enabled.
```
liminw: mTLS is required because of our current implementation. But this restriction could be removed in the future. Can we make this point clear in the comment?

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,282 @@
...
+
+// $hide_from_docs
+// Condition specifies additional required attributes.
+message Condition {
+  // Required. The name of an Istio attribute.
+  // Note: Check https://istio.io/docs/reference/config/ for the list of supported
+  // attribute name.
+  string key = 1;
+
+  // Required. The allowed values for the attribute.
```
liminw: For "Values" we support exact/prefix/suffix/presence match.

security/v1beta1/authorization.proto
yangminzhu: I removed it from the API, after a second thought, I feel it's better to not to include the implementation detail in the API.

security/v1beta1/authorization.proto
yangminzhu: I mentioned this in the comment for `Rule`, I feel it's better to just mention it once for all string fields that we support exact/prefix/suffix/presence matching.

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,281 @@
...
+  // attribute.
+  //
+  // If not set, any namespace is allowed.
+  repeated string namespaces = 3;
+
+  // Optional. A list of IPs, which matches to the “source.ip” attribute.
+  // Both single IP (e.g. "1.2.3.4") or CIDR (e.g. "1.2.3.0/24") are supported.
+  //
+  // If not set, any ip is allowed.
+  repeated string ips = 4;
```
liminw: ips -> ipBlocks? We can just support ipBlocks. A single IP "1.2.3.4" can be expressed as "1.2.3.4/32".

yangminzhu:
@costinm @rshriram Could you review the API change? The design is approved and already shared with the community. Thank you.

security/v1beta1/authorization.proto
```diff
@@ -0,0 +1,281 @@
...
+
+  // Optional. A list of rules to specify the allowed access to the workload.
+  //
+  // If not set, access is denied unless explicitly allowed by other authorization policy.
+  repeated Rule rules = 2;
+}
+
+// $hide_from_docs
+// WorkloadSelector specifies the scope where the authorization policy should be applied to.
+message WorkloadSelector {
```
rshriram: can you use this opportunity to move the workloadSelector from sidecar and use the same one? I do not want to duplicate the workload selector again. Second, you have a different field in workload selector than what sidecar has, which is not a good UX.

costinm:
+1

costinm:
And preferably use one of the protos and syntax for selector that already
exists - either the one in Sidecar or Envoyfilters.

liminw:
Add @louiscryan @smawson 

@costinm @rshriram Agree that we should have a single "selector" syntax. Based on [Istio Config Model](https://docs.google.com/document/d/1T22PxxLwwFnxC_yFQ_HmuKsuZgFjRcDy2BRb5enfYpk), the following is the only approved "selector" format. Shall we reconcile the existing APIs to move to the agreed syntax?

```
  selector:
     matchLabels:
          <label selector>
```

yangminzhu:
@rshriram @costinm @liminw I changed to use the WorkloadSelector in Sidecar. We could move the `WorkloadSelector` in a common directory if you prefer, I can do it in this PR.

As for Gateway, I would prefer not to change it in this PR. The syntax will change and all code and yaml need to update, and this should be done in a separate PR.

liminw:
In existing code, there are two different selector definitions.

In sidecar
```
workloadSelector:
   labels:
      app: A
```
In Gateway, 
```
selector:
   app: A
```
Both are different from the agreed on selector definition in [Istio config model doc](https://docs.google.com/document/d/1T22PxxLwwFnxC_yFQ_HmuKsuZgFjRcDy2BRb5enfYpk). Either picking one of the definitions above, or following the recommendation in Istio config model, we need a standard selector definition, and use it for all APIs consistently.

yangminzhu:
Within this PR's scope, I think we have 3 options:

1. Add new WorkloadSelector that is exactly the same as the one approved in the [Istio config model doc](https://docs.google.com/document/d/1T22PxxLwwFnxC_yFQ_HmuKsuZgFjRcDy2BRb5enfYpk)

1. Reuse the WorkloadSelector in Sidecar.proto which is a little bit different from the approved one. The difference is in the field name: `labels` vs `matchLabels`

1. Reuse the WorkloadSelector in Sidecar.proto but also move it to a new common directory.

@rshriram @costinm @louiscryan @liminw I can do any of the above in this PR, could you let me know which one you prefer? Thank you.

rshriram:
I care less about naming. We need consistency. So, if this new "selector" is the answer, then lets go for it. You can change the EnvoyFilter's workloadSelector to this new Selector thing immediately as it has not been released yet. For the gateway, you have a problem. Sidecars would need some deprecation model..

PR #920, opened 2019-05-09
Document default behavior for maxRequestsPerConnection

ssaavedra:
The default on envoy is to use MAX_STREAMS, defined internally
to be 2²⁹ connections.

This is one of the few properties missing an explanation for its default value.

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -419,7 +419,8 @@ message ConnectionPoolSettings {
     int32 http2_max_requests = 2;
 
     // Maximum number of requests per connection to a backend. Setting this
-    // parameter to 1 disables keep alive.
+    // parameter to 1 disables keep alive. Default 0, meaning "unlimited",
+    // up to 2^29.
     int32 max_requests_per_connection = 3;
 
     // Maximum number of retries that can be outstanding to all hosts in a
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/destination_rule.proto
```diff
@@ -419,7 +419,8 @@ message ConnectionPoolSettings {
     int32 http2_max_requests = 2;
 
     // Maximum number of requests per connection to a backend. Setting this
-    // parameter to 1 disables keep alive.
+    // parameter to 1 disables keep alive. Default 0, meaning "unlimited",
+    // up to 2^29.
```
hzxuzhonghu: Is there any doc says 2^29?

ssaavedra: I think this is currently an implementation detail in Envoy, but we could reach out to Envoy to assert it or even document it officially. Otherwise, we can reword that as simply "unlimited", although unlimited would not be precise.

You get that value when [checking here](https://github.com/envoyproxy/envoy/blob/b41ba5925a4e93d22a86c6501d63314ccf0d79f3/source/common/http/http2/conn_pool.cc#L114), that maxTotalStreams [is implemented as](https://github.com/envoyproxy/envoy/blob/b41ba5925a4e93d22a86c6501d63314ccf0d79f3/source/common/http/http2/conn_pool.cc#L315) MAX_STREAMS, which is defined [as (1 << 29)](https://github.com/envoyproxy/envoy/blob/97bda5098040b703dd5b84a5cd913b6e2e79ff2a/source/common/http/http2/conn_pool.h#L113) with a comment:

>   // All streams are 2^31. Client streams are half that, minus stream 0. Just to be on the safe
> // side we do 2^29.

PR #922, opened 2019-05-15
Add issue templates like in istio/istio

geeknoid:


ssuchter:
How about directing the user to discuss.istio.io as an option (for best effort support?)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
+---
+name: Bug report
+about: Report a bug to help us improve Istio
+---
+
```
mandarjog: We should point users to the `discuss forum` right here for "how do I do this" type questions

geeknoid:
@ssuchter @mandarjog Added a note about discuss.istio.io

ozevren:
/lgtm

Should we even allow bugs on the istio/api repo?

PR #923, opened 2019-05-15
Move attribute proto definitions out of the wire protocol

kyessenov:
This is a generally safe operation since there is no change in the wire format. 
De-couple wire encoding from the runtime definitions. 
Motivated by https://github.com/istio/pkg/pull/13

/assign @geeknoid 


geeknoid:
Sorry, I don't think this is right either.

First, this is moving internal protocol definitions into a user-visible package. This would result in the wire protocol protos being documented on istio.io, which we definitely don't want. At the very least, this stuff would need to be in a separate package.

Second, I don't think the attribute package needs any dependencies on istio/api at all since attributes (in a post-Mixer-V1 world) will not be on the wire. So there's no need to have stuff coming from istio/api and no need for protos. Unless I'm missing something of course.

kyessenov:
I just need definitions of attribute types and attribute references. FWIW we use proto as schema definitions, and not for the marshalling concerns. Attribute context is already a proto https://github.com/envoyproxy/envoy/blob/master/api/envoy/service/auth/v2/attribute_context.proto, so I'm just looking at schematization of that in terms of attributes. 

We can drop compressed attributes from presentation, but we'll need a bit more refactoring of istio to separate compressed attributes from reference tracking.

PR #929, opened 2019-05-22
adding fields to allow batch reporting options to be tunable as helm values

huang195:
This is the follow-up PR to https://github.com/istio/api/pull/914. These are needed to control the behavior of batch reporting in terms of helm variables. When the mixer plugin starts, these values can be passed in as Envoy filter configuration values.


huang195:
@mandarjog /ping. This is pretty much the same fields added in TransportOptions that you reviewed earlier, but needed these fields to be defined here as well so they can be passed from mixer plugin to Envoy filter.

huang195:
@mandarjog gentle /ping.

PR #930, opened 2019-05-24
H2 auto upgrade options

mandarjog:
Give operator the choice to opt-in or opt-out of http2 upgrades by destination rules.


networking/v1alpha3/destination_rule.proto
```diff
@@ -431,6 +431,22 @@ message ConnectionPoolSettings {
     // If not set, there is no idle timeout. When the idle timeout is reached the connection will be closed. 
     // Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections.
     google.protobuf.Duration idle_timeout = 5;
+
+    // Policy for upgrading http1.1 connections to http2.
+    enum H2UpgradePolicy {
+      // Use the global default.
+      DEFAULT = 0;
```
rshriram: where is the global default defined?

mandarjog: TBD. I don't want to add more stuff to mesh config. 
While that is being decided this will be an opt-in feature. So the implicit global default is `DO_NOT_UPGRADE`

rshriram: Lets not add implicit global defaults if there is no mesh config.. Might as well have this as DO_NOT_UPGRADE as the default. If we are not touching mesh config, then I see no point in even specifying text related to global_default.

mandarjog: making 0 - do_not_upgrade to h2 can create istio upgrade issues.
At some point we want to change this to opt out. In that case the default value *cannot* be opt-out.

rshriram: Now that we have decided to add stuff to mesh config, can you also add a global default in the mesh config? Then the semantics become much clear

mandarjog: Ok I will do it. Thanks.

mandarjog: @rshriram updated, PTAL.

mesh/v1alpha1/config.proto
```diff
@@ -297,8 +297,20 @@ message MeshConfig {
...
+  // Default Policy for upgrading http1.1 connections to http2.
+  enum H2UpgradePolicy {
+    // Do not upgrade connections to http2.
+    DO_NOT_UPGRADE = 0;
+    // Upgrade the connections to http2.
+    UPGRADE = 1;
+  }
+
+  // Specify if http1.1 connections should be upgraded to http2 by default.
+  // destinationRule.trafficPolicy.connectionPool.http.h2UpgradePolicy overrides this setting.
```
rshriram: I would add a note here saying that if sidecar is installed on all pods in the mesh, then this should be set to UPGRADE. If one or more services/namespaces do not have a sidecar, then this should be set to DO_NOT_UPGRADE.

mandarjog:
updated comments, PTAL @rshriram

mandarjog:
@rshriram ping!

PR #932, opened 2019-05-30
Fix invalid ServiceEntry docs

hzxuzhonghu:


networking/v1alpha3/service_entry.proto
```diff
@@ -338,19 +338,19 @@ option go_package = "istio.io/api/networking/v1alpha3";
 
 message ServiceEntry {
   // REQUIRED. The hosts associated with the ServiceEntry. Could be a DNS
-  // name with wildcard prefix (external services only). For HTTP traffic
-  // the HTTP Host/Authority header will be matched against the hosts field.
-  // For HTTPs or TLS traffic containing Server Name Indication (SNI), the SNI value
-  // will be matched against the hosts field. For all other protocols
-  // the hosts will be ignored, and the port and addresses fields
-  // will be used if present. Note that when resolution is set to type DNS
+  // name with wildcard prefix.
```
linsun: isn't this only for external services as you don't need service entry for services in k8s?

networking/v1alpha3/service_entry.proto
```diff
@@ -338,19 +338,19 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
+  // 1. The hosts field is used to select VirtualServices and DestinationRules.
```
linsun: select matching hosts in VirtualServices and DestinationRules?

linsun: I think you intended to have 2 and 3 below as sub-bullets to elaborate 1.  Is that correct?

networking/v1alpha3/service_entry.proto
```diff
@@ -338,19 +338,19 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
+  // 2. For HTTP traffic the HTTP Host/Authority header will be matched against the hosts field.
```
frankbu: Is this true when the `addresses` field is used to call the service by IP?

networking/v1alpha3/service_entry.proto
```diff
@@ -338,19 +338,19 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
+  // Note that when resolution is set to type DNS
   // and no endpoints are specified, the host field will be used as the DNS name
   // of the endpoint to route traffic to.
   repeated string hosts = 1;
 
   // The virtual IP addresses associated with the service. Could be CIDR
-  // prefix. For HTTP traffic the addresses field will be ignored and
-  // the destination will be identified based on the HTTP Host/Authority
+  // prefix. For HTTP traffic the addresses field will be used to generate http route domains
+  // along with hosts field and the destination will be identified based on the HTTP Host/Authority
```
frankbu: `along with hosts field` implies that the `hosts` field is generated, so it should not be marked `REQUIRED`, above.

networking/v1alpha3/service_entry.proto
hzxuzhonghu: it can be used for k8s headless service as well

networking/v1alpha3/service_entry.proto
hzxuzhonghu: I know what u mean. Hmm, but how should we describe this when calling service by its ip address?

networking/v1alpha3/service_entry.proto
hzxuzhonghu: can be used

hzxuzhonghu:
>but I think the hosts field is the host that clients will address, unless addresses field is set, in which case the client will call the service using the IP

I donot quite agree with this , as below config, client can also use host to access server instead of by address. 

image

networking/v1alpha3/service_entry.proto
hzxuzhonghu: I remembered there is a issue using SE for StatefulSet Service

networking/v1alpha3/service_entry.proto
hzxuzhonghu: 1 can be for all protocols 

2 and 3 are for specific protocol

networking/v1alpha3/service_entry.proto
hzxuzhonghu: ref: https://github.com/istio/istio/issues/10659

frankbu:
@hzxuzhonghu What is the use case for setting addresses in the service entry but then calling the service using the host name? It seems to me that addresses would only be used when there is no DNS host, so the client needs to call the service using IPs.

@vadimeisenbergibm You have lots of experience with this feature. WDYT?

hzxuzhonghu:
One use case Ican think of now is when we access same port TCP services , we can set both the address and host.
Then pilot will generate listeners with the specific ip instead of '0.0.0.0'. Whatever, the comments should address all use guide if possible. But in reality, it is kind of difficult now.

hzxuzhonghu:
> It seems to me that addresses would only be used when there is no DNS host, so the client needs to call the service using IPs.
I think no one want to call a service by ip address, this is definitely not the original design intention.

networking/v1alpha3/service_entry.proto
hzxuzhonghu: It is also fit for `addresses`, see L352

howardjohn:
> I think no one want to call a service by ip address, this is definitely not the original design intention

there are some cases for this,like when people use brokers and Kafka that returns the ip address to connect to

vadimeisenbergibm:
@frankbu @hzxuzhonghu @howardjohn My two cents: please note that for the TCP protocol the hostname is ignored by Envoy. Hostname is not a part of TCP, Envoy does not see it. So it does not matter how the client calls the service, by a Host name or by IP, Envoy only sees the IP.

hzxuzhonghu:
Thanks @vadimeisenbergibm , envoy does not care the hostname for tcp, only istio does.

networking/v1alpha3/service_entry.proto
```diff
@@ -338,19 +338,19 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
+  // 1. The hosts field is used to select matching hosts in VirtualServices and DestinationRules.
+  // 2. For HTTP traffic the HTTP Host/Authority header will be matched against the hosts field.
+  // 3. For HTTPs or TLS traffic containing Server Name Indication (SNI), the SNI value
+  // will be matched against the hosts field.
+  // Note that when resolution is set to type DNS
   // and no endpoints are specified, the host field will be used as the DNS name
   // of the endpoint to route traffic to.
   repeated string hosts = 1;
 
   // The virtual IP addresses associated with the service. Could be CIDR
```
frankbu: Is `Could be CIDR` correct, or is it `Must be CIDR`? I believe that somebody reported recently that `a.b.c.d` didn't work, but `a.b.c.d/32` did? Is that probably a user error? If not, we should update the doc.
@rshriram ^^^

rshriram: yes it could be cidr.

networking/v1alpha3/service_entry.proto
frankbu: I still don't understand what this sentence is trying to say. Can you give an example that shows what it means?

networking/v1alpha3/service_entry.proto
```diff
@@ -338,19 +338,19 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
 message ServiceEntry {
   // REQUIRED. The hosts associated with the ServiceEntry. Could be a DNS
-  // name with wildcard prefix (external services only). For HTTP traffic
-  // the HTTP Host/Authority header will be matched against the hosts field.
-  // For HTTPs or TLS traffic containing Server Name Indication (SNI), the SNI value
-  // will be matched against the hosts field. For all other protocols
-  // the hosts will be ignored, and the port and addresses fields
-  // will be used if present. Note that when resolution is set to type DNS
+  // name with wildcard prefix.
+  // 1. The hosts field is used to select matching hosts in VirtualServices and DestinationRules.
```
frankbu: This numbered list doesn't seem to generate a proper list in the html table. Maybe it's not supported?
@geeknoid ^^^

geeknoid: Hmm. That should work. Try putting a blank line before and after the bulleted list.

networking/v1alpha3/service_entry.proto
hzxuzhonghu: You can check the route configuration, both hostname and addresses are there.

frankbu: Got it. Maybe this would be a little clearer wording?
```
  // prefix. For HTTP traffic, generated route configurations will include http route
  // domains for both the `addresses` and `hosts` field values and the destination will
  // be identified based on the HTTP Host/Authority
```

frankbu: @hzxuzhonghu ^^^ WDYT?

hzxuzhonghu: SGTM

frankbu:
I think we want this in 1.2, so it needs to be sent to the `release-1.2` branch, instead of master.

hzxuzhonghu:
@frankbu The policy is get into master and cherry-pick to 1.2 i think

frankbu:
@hzxuzhonghu it turns out that the merge-to-release-branch bot that @geeknoid added wasn't working reliably (sometimes picking up extra commits), so he had to disable it. So for now we are back to the old process of just sending PRs that belong in 1.2 to the release-1.2 branch and then in a few weeks from now we will do a bulk cherry-pick. Martin is also going to experiment with changing the merge bot to work the other way, i.e., send PRs to release branch with a merge-to-master label to do the automatic cherry-pick back to master.

frankbu:
Sorry, didn't mean to close it

frankbu:
Actually, I forgot this is an istio/api PR, not istio/istio.io, so I guess the merge to master and then cherry pick to release branch is valid.

/lgtm
/approve

hzxuzhonghu:
opend #954

PR #933, opened 2019-05-30
Remove stale useless default_endpoint field in gateway.server

hzxuzhonghu:


geeknoid:
This is a breaking change. Do we need a deprecation story here?

linsun:
How do we know this is stale?  we definitely need release note this deprecation.

networking/v1alpha3/gateway.proto
```diff
@@ -410,10 +410,8 @@ message Server {
   // https, and the TLS modes to use.
   TLSOptions tls = 3;
 
-  // The loopback IP endpoint or Unix domain socket to which traffic should
-  // be forwarded to by default. Format should be `127.0.0.1:PORT` or
-  // `unix:///path/to/socket` or `unix://@foobar` (Linux abstract namespace).
-  string default_endpoint = 5;
+  reserved 5;
+  reserved "default_endpoint";
```
rshriram: why is this stale? this is implemented right? if not, my bad as I forgot about implementing this. This needs to be implemented.

hzxuzhonghu: It is no implemented now.

hzxuzhonghu:
@linsun @geeknoid I noticed it was in since 2017, and not used

rshriram:
I think you are looking at a wrong branch.. This field was added 5 months ago by me
https://github.com/istio/api/pull/742 
I didn't implement it (forgot to)

hzxuzhonghu:
ok, then let me close it.

PR #935, opened 2019-05-30
Import common files into this repo.

geeknoid:
- Delete circleci config as its completely redundant with what prow is doing.

- Delete other cruft


WORKSPACE
```diff
@@ -1,19 +0,0 @@
-workspace(name = "io_istio_api")
```
rshriram: can this be deleted? doesn't istio proxy depend on this?

kyessenov:
You cannot remove bazel stuff or you're going to break proxy bazel build.

scripts/run_golangci.sh
```diff
@@ -0,0 +1,39 @@
...
+ROOTDIR=$(dirname "${SCRIPTPATH}")
+cd "${ROOTDIR}"
+
+if [[ "$1" == "--fix" ]]
+then
+    FIX="--fix"
+fi
+
+# if you want to update this version, also change the version number in .golangci.yml
+GOLANGCI_VERSION="v1.16.0"
```
kyessenov: why do we need to run golangci here? it automatically skips generated files, so it's entirely a no-op.

geeknoid: This is a common file, copied from the istio/common-files repo.

geeknoid:
OK, I restored WORKSPACE and BUILD. We should fix the proxy repo so that these turds aren't needed in istio/api...

PR #937, opened 2019-06-04
Replacing envoy_google_grpc with envoy_grpc for als

hanahmily:
This change is in preparation for adding support for the envoy http gRPC access log service (ALS) API. The old solution is using envoy_google_grpc which is not quite stable, so I change it to envoy_grpc.


rshriram:
thanks for fixing this. ! @douglas-reid / @mandarjog - this is now consistent with envoy metrics service

PR #940, opened 2019-06-05
Add Thrift definitions

pnovotnak:
This is a WIP PR that adds definitions for [Thrift](https://thrift.apache.org/)

Companion PR: https://github.com/istio/istio/pull/14578


mixer/v1/config/client/client_config.proto
```diff
@@ -208,3 +208,36 @@ message TcpClientConfig {
   // should not be less than 1 second, otherwise it will be reset to 1 second.
   google.protobuf.Duration report_interval = 6;
 }
+
+// Defines the client config for Thrift.
+message ThriftClientConfig {
```
rshriram: why do you need this? This is actually used by the istio-proxy to wrap the data into specific objects. I dont think you are looking for mixer control over thrift connections, are you?

pnovotnak: Are these defs needed to get telemetry working? We'll want that at some point, but we don't need it yet.

rshriram: Sorry for the delay. You dont need this until you add support in istio-proxy for thrift stack.

geeknoid:
Closing since the matching PR in istio/istio was closed.

PR #942, opened 2019-06-06
Refine RBAC API usage for TCP and gRPC services. 

pitlv2109:
This is for the reference page on istio.io 

Related PR: https://github.com/istio/istio.io/pull/4262

### DIFF
--- a/rbac/v1alpha1/rbac.proto
+++ b/rbac/v1alpha1/rbac.proto
@@ -25,7 +25,7 @@ syntax = "proto3";
 // the following standard fields:
 //
 //   * services: a list of services.
-//   * methods: HTTP methods. In the case of gRPC, this field is ignored because the value is always "POST".
+//   * methods: A list of HTTP methods. You can set the value to `*` to include all HTTP methods. This field should not be set for TCP and gRPC services.
 //   * paths: HTTP paths or gRPC methods. Note that gRPC methods should be
 //     presented in the form of "/packageName.serviceName/methodName" and are case sensitive.
 //
@@ -161,8 +161,8 @@ message AccessRule {
   repeated string not_paths = 7;
 
   // Optional. A list of HTTP methods (e.g., "GET", "POST").
-  // It is ignored in gRPC case because the value is always "POST".
-  // If not specified, it matches to any methods.
+  // If not specified or specified as "*", it matches to any methods.  
+  // This field should not be set for TCP and gRPC services.
   repeated string methods = 3;
 
   // $hide_from_docs
--- a/rbac/v1alpha1/rbac.proto
+++ b/rbac/v1alpha1/rbac.proto
@@ -25,7 +25,9 @@ syntax = "proto3";
 // the following standard fields:
 //
 //   * services: a list of services.
-//   * methods: A list of HTTP methods. You can set the value to `*` to include all HTTP methods. This field should not be set for TCP and gRPC services.
+//   * methods: A list of HTTP methods. You can set the value to `*` to include all HTTP methods.
+//              This field should not be set for TCP services. The policy will be ignored.
+//              For gRPC services, only `POST` is allowed; other methods will result in denying services.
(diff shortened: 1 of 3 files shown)


rbac/v1alpha1/rbac.proto
```diff
@@ -25,7 +25,7 @@ syntax = "proto3";
 // the following standard fields:
 //
 //   * services: a list of services.
-//   * methods: HTTP methods. In the case of gRPC, this field is ignored because the value is always "POST".
+//   * methods: A list of HTTP methods. You can set the value to `*` to include all HTTP methods. This field should not be set for TCP and gRPC services.
```
yangminzhu: I feel this part is a little confusing, "not setting this field" is equal to "setting this field to *", right?
So "This field should not be set for TCP and gRPC services." actually means to allow all HTTP methods for gRPC? (of course it should be set for TCP, I think you don't need to mention this, like `path` is also not for TCP)

rbac/v1alpha1/rbac.proto
```diff
@@ -161,8 +161,8 @@ message AccessRule {
   repeated string not_paths = 7;
 
   // Optional. A list of HTTP methods (e.g., "GET", "POST").
-  // It is ignored in gRPC case because the value is always "POST".
-  // If not specified, it matches to any methods.
+  // If not specified or specified as "*", it matches to any methods.  
+  // This field should not be set for TCP and gRPC services.
```
yangminzhu: Same here.

rbac/v1alpha1/rbac.proto
pitlv2109: TCP and gRPC to some extent does not have the term "methods" (grpc methods are actually used in paths), right? So the users should not bother setting any of these for TCP and gRPC services.

diemtvu: Probably explain what happen if it is set for TCP and/or gRPC. If we can rename to http_methods, it may help a bit.

liminw: +1 to add explanation on what happens if not set for TCP and gRPC. 

I would prefer keeping the field name "methods" as is. "http_methods" seems too long.

pitlv2109: How about "This field should not be set for TCP services (policy will not be accepted) and gRPC services (only `POST` is allowed; other methods will result in denying services)."? But I still feel like this is unnecessary.

yangminzhu: I think you can just simplify it to: `methods: A list of HTTP methods. For gRPC, you can either omit this field or set it to "POST".`

pitlv2109: Then we will fall back into the "explain what happen if it is set of TCP". I'll let you guys decide the wording, but I think we will need this checked in for 1.2 and after. For the end users, I think saying do not use it for TCP and gRPC is fine, no explanation is needed.

yangminzhu: That's the same question for `path`, `host`, yeah, you can also just say `do not use it for TCP and gRPC`, much simpler.

pitlv2109: I think this is fine as it is. For end users, we can just say don't set/use it and be done with it. This PR is needed for release 1.2 because currently, it's misleading for the end users. The doc says "In the case of gRPC, this field is ignored" while it's actually not ignored, if the user sets `methods`, it's still being used.

pitlv2109:
Doesn't seem like we can reach an agreement for the wording. Can we merge this quick for release-1.2 for the sake of end users? @diemtvu @liminw

rbac/v1alpha1/rbac.proto
diemtvu: I think we need to explain the consequence of (incorrect) settings. For example, `if this is set for TCP, requests will be denied` (or whatever the current behavior is). Similarly for gRPC.

pitlv2109: Pushed. PTAL.

Issue #939, opened 2019-06-05
MCP resource name format is actually "namespace/name"


jeffmendoza:
https://github.com/istio/api/blob/master/mcp/v1alpha1/metadata.proto#L55

    //    "/<k8s namespace>/<k8s resource name>"

This is incorrect, we are seeing "namespace/name" (without leading slash) on live data. Seems like the Galley code is good and working, so the right thing to do is update the comment/docs.

jeffmendoza:
cc @panywei @qfel @ozevren

kaushikchaubal:
I can pick up this issue and create a PR. This is going to be my first PR to this community so if anything is incorrect, please shout out.

jeffmendoza:
Sounds good. Make sure to follow https://github.com/istio/api#updating to update the generated files.

kaushikchaubal:
I have created a PR here: https://github.com/istio/api/pull/944. Looks like it is awaiting a trusted user to review the PR and leave an /ok-to-test message.

kaushikchaubal:
this is now merged. Should i close this issue now?

PR #944, opened 2019-06-07
#939: Changing the documentation to the format 'namespace/name' (without leading slash)

kaushikchaubal:


kaushikchaubal:
I signed it!

jeffmendoza:
@kaushikchaubal Looks like the error is:

```
Repo has unstaged changes. Re-run ./scripts/generate-protos.sh or make proto-commit
```

Try that out.

jeffmendoza:
You'll also need `make clean generate` as described here: https://github.com/istio/api/blob/master/README.md#updating

kaushikchaubal:
@jeffmendoza - i have run the following commands and pushed as per your comment:
```
make proto-commit
make clean generate
```

PR #948, opened 2019-06-15
add case sensitive flag to StringMatch

ramaraochavali:
Adds case sensitive flag to `StringMatch`.


ramaraochavali:
Once this is merged, I will follow-up with implementation.

networking/v1alpha3/virtual_service.proto
```diff
@@ -907,6 +907,10 @@ message StringMatch {
     // ECMAscript style regex-based match
     string regex = 3;
   }
+
+   // Indicates that string matching should be case insensitive. The default is true.
```
rshriram: is that the default today?

ramaraochavali: Yes. If we do not specify, Envoy defaults it to true https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto.html?highlight=case_sensitive

ramaraochavali: I realized we have `ignore_uri_case ` flag already defined that allow us to override the behaviour of uris. We are good with that. I will close this PR.

PR #946, opened 2019-06-14
Fix envoy links

sdake:
commit ab50e4491b613f1ece33f297a7d49c8e5afbb477
Author: Steven Dake <sdake@users.noreply.github.com>
Date:   Fri Jun 14 14:58:20 2019 -0700

    Run make after modifying protos
    
    This generates the go commented PI as well as the proto HTML
    documentation.

commit 332d1e3a677bf7e4deded524a1a590b690d37541
Author: Steven Dake <sdake@users.noreply.github.com>
Date:   Fri Jun 14 14:56:45 2019 -0700

    Modify upstream links that changed in envoy docs
    
    See PR: https://github.com/istio/istio.io/pull/4424/files
    
    This was an emergency blockage during the last few days of release
    1.2, so the exact process of modifying API first then modifying docs
    was not done.  Instead, sadly it felt backwards (modify docs first,
    then modify this repo).

PR #951, opened 2019-06-17
Fix envoy links (#946)

geeknoid:
* Modify upstream links that changed in envoy docs

See PR: https://github.com/istio/istio.io/pull/4424/files

This was an emergency blockage during the last few days of release
1.2, so the exact process of modifying API first then modifying docs
was not done.  Instead, sadly it felt backwards (modify docs first,
then modify this repo).

* Run make after modifying protos

This generates the go commented PI as well as the proto HTML
documentation.

(cherry picked from commit 20f94249cb73ff74208276fd6763134f79adcfd0)


geeknoid:
@sdake FYI

geeknoid:
Needed for 1.2 docs.

PR #953, opened 2019-06-19
Add ProtocolOptions to DestinationRule

hap2712:
This PR adds the [Envoy http2 protocol options](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto.html) configuration to DestinationRule, as these run per-upstream-cluster.


rshriram:
Hi, can you use the EnvoyFilter crd to add these options for specific clusters?

acevedop:
I have not used Envoy Filter CRDs before, but this is part of the connection manager section in Envoy, not filters, so I would say it can not be done like that. Anyway this is already part of specific clusters because its a DestinationRule, so you have this as part of those rules.

networking/v1alpha3/destination_rule.proto
```diff
@@ -445,6 +445,28 @@ message ConnectionPoolSettings {
...
+    // buffer per stream in the HTTP2 codec buffers. Once the buffer reaches
+    // this pointer, watermark callbacks will fire to stop the flow of data to
+    // the codec buffers. 
+    uint32 initial_stream_window_size = 9;
+
+    // Similar to initial_stream_window_size, but for connection-level
+    // flow-control window.
+    uint32 initial_connection_window_size = 10;
+
+    // Allow proxying websocket and other upgrades over H2 connect.
```
costinm: What is the default ?

costinm:
Re. filter API - we do want to allow filters to apply to clusters, and to use it for special configs. But H2 standard settings are quite reasonable first class APIs, and we have few other h1 and h2 options. None of the added options seem envoy specific or special. My only concern is naming - I want to make sure the options match either the names used in H2 spec or Envoy config, I'm tired of new names.

rshriram:
/hold 
Let’s not add these options as very few people are going to use this. Please use envoy filter instead.

rshriram:
@hap2712 please check out the Istio EnvoyFilter api. Despite its naming, it actually allows you to edit all aspects of Envoy's config (clusters, routes, listeners, etc.). So this setting can be easily accomplished with the EnvoyFilter crd.

geeknoid:
Closing due to inactivity.

PR #957, opened 2019-06-24
 Check in initial openAPI spec for Istio APIs

jasonwzm:
-  APIs in envoy/ are not generated with openapi spec
-  Updated Makefile so that OpenAPI gen is integrated with API update
-  Updated protoc image tool


jasonwzm:
/hold
looks like presubmit failed when running the generator with the new checked in code. I am looking into it.

geeknoid:
Excellent!

/lgtm
/approve

jasonwzm:
@geeknoid The new Docker image resulted in a different protoc go binary and resulted in slightly different generated go files. The diffs can be seen here: https://gist.github.com/jasonwzm/c75cb8619452f94853016ee99001906d
The diffs are primarily different implementations of the sovXXX functions and UnimplementedXXX structs which we do not use in Istio code, so no impact on istio/istio. 
I will add go modules in Dockerfiles of our tools afterwards.

geeknoid:
Can you check out the docs for the gogo protoc extension to see if there's a way to disable this new output? It's just junk we don't need, so I'd rather nix it if we can.

jasonwzm:
Closing this PR now. A new PR will be created to check in the openapis.

PR #958, opened 2019-06-25
Add feature gate in global config

hzxuzhonghu:


hzxuzhonghu:
/assign @rshriram 

Added `LOCALITY_AWARE_ROUTING ` in initial pr, more features should be moved here in future.

mesh/v1alpha1/config.proto
```diff
@@ -419,3 +422,19 @@ message LocalityLoadBalancerSetting{
...
+  LOCALITY_AWARE_ROUTING = 0;
+}
+
+// Feature specify to enable or disable a specific feature.
+message Feature {
+  // Feature is the name of the feature.
+  FeatureName feature = 1;
+
+  // Specify true to enable the feature or false to disable the feature.
+  bool enable = 2;
```
rshriram: sorry for delay. This is very nice. But do you think its useful to have a enable/disable? 
Why can't we simply have a repeated enum such that in mesh config we have

```
features:
- LOCALITY_AWARE_ROUTING
- FOO_BAR
- BAZ
...
```

If someone wants to disable it, they could remove that line right?

hzxuzhonghu: Originally, i have thought about this, but this requires user to explicitly set all features if they want to enable them. 

How about we make this as type map[string]bool. If we have default enabled features in pilot like 
LOCALITY_AWARE_ROUTING,  FOO_BAR, BAZ, and we want to disable BAZ, we just need to set:

```
features:
  BAZ: false
```

hzxuzhonghu: And also have `Feature1` disabled by default in pilot code, and we want to enable it, then just need explicitly set:

```
features:
  Feature1: true
```

hzxuzhonghu: But this has a limit, the key of map in proto can not be a enum type.

rshriram: hmm.. I think then what you have is good enough.

PR #963, opened 2019-06-28
Fix documentation of supported protocols for Port.protocol

robertpanzer:
This PR updates the documentation of Gateway.port.protocol so that it matches the list of supported values in Istio 1.2.0.
Right now the documentation is missing the values `REDIS`, `MYSQL` and `UDP`.

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -422,7 +422,7 @@ message Port {
   uint32 number = 1;
 
   // REQUIRED: The protocol exposed on the port.
-  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|TCP|TLS.
+  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|REDIS|MYSQL|TCP|TLS|UDP.
   // TLS implies the connection will be routed based on the SNI header to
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -422,7 +422,7 @@ message Port {
   uint32 number = 1;
 
   // REQUIRED: The protocol exposed on the port.
-  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|TCP|TLS.
+  // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|REDIS|MYSQL|TCP|TLS|UDP.
```
hzxuzhonghu: We donot support udp now.

robertpanzer: @hzxuzhonghu Can you give some insight on what you mean with UDP is not supported?
I see several places in the codebase of pilot that explicitly handle Protocol_UDP.
My understanding is that the traffic is not routed through the istio-proxy and that no outbound clusters are generated, but still this constant is handled differently than any other random string. (And I couldn't see in the code how inbound clusters are handled)

hzxuzhonghu: You are right

PR #964, opened 2019-07-01
Fixes #12873. Add property Sidecar.OutboundTrafficPolicy to configure…

robertpanzer:
… outbound traffic policy individually per application

This PR fixes https://github.com/istio/istio/issues/12873 and adds a new field `OutboundTrafficPolicy` to the `Sidecar` resource.
That way it would be possible to override the global outbound traffic policy per namespace or application.

I created the PR for master as a backward compatible addition to the API would require an increase of the minor version number and there is no branch yet for 1.3 and I assume that Istio follows semver.

I'll create a corresponding PR for istio/istio.


hzxuzhonghu:
One small nit: it is better if we can reuse `MeshConfig_OutboundTrafficPolicy_Mode`

robertpanzer:
@hzxuzhonghu I has a similar PR already open in the past, and planned to do the same, but @rshriram made me aware of that this would introduce a circular dependency.

robertpanzer:
I had to manually update protocol.lock.
`make` created a slightly different file for me

PR #965, opened 2019-07-01
Add proto to OpenAPI to support CUE constraints pipeline

mpvl:
This CL is an alternative to PR #957.

Notable differences:

- This does not use the regular proto generation infrastructure
as the idea is to combine more than just Protobufs  as a source
for the OpenAPI schema (e.g. Go code and CUE files)

- All files are self-contained, but not do repeat imported types:
types are included as a component with a fully qualified path.

- Fixes several issues relative to #957:
  - uses JSON names
  - fixes oneOf (current representation reflects the proto accurately)
  - accurate interpretation of protobuf types, even if not
    supported by OpenAPI (e.g. uint64)

Current issues:

- Proto types with standard mappings to JSON still map
  to the proto type. Question: support both or just JSON mapping.
- Not printing document info and version yet.


mpvl:
@jasonwzm @geeknoid: as mentioned over email.

jasonwzm:
Can we just check in the json files for now and I would like to talk to you about the tools integration in the pipeline tomorrow?

mpvl:
@jasonwzm: sure, but did you mean also not check in the generation script? Note that this does not check in any of the cue files already, only the final result and the script.

jasonwzm:
The schema is checked in at https://github.com/istio/api/pull/989. Closing this one.

PR #969, opened 2019-07-03
Add default validation context options for mTLS

Phlak106:
Envoy exposes a default validation context which gets merged with the validation context from SDS via Pilot.  However, those options are not exposed to users. This would allow users to configure the default validation context similar to Envoy and let Pilot merge the contexts but with more defaults.

https://github.com/istio/istio/blob/c5ff36a66084fe7f3b72d5a15203f5783317dbec/pilot/pkg/networking/core/v1alpha3/gateway.go#L424

https://discuss.istio.io/t/mtls-default-rootca-for-client-certificates/2881/9


Phlak106:
I signed it!

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,57 @@ message Server {
...
+      // If not specified and a peer certificate is presented it will not be
+      // verified. By default, a client certificate is optional, unless one of
+      // the additional options (
+      // :ref:`require_client_certificate`
+      // :ref:`verify_certificate_spki`
+      // is also specified.
+      //
+      // See :ref:`the TLS overview <arch_overview_ssl_enabling_verification>`
+      // for a list of common system CA locations.
+      string trusted_ca = 1;
```
hzxuzhonghu: what's this? A ca file or ca string?

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,57 @@ message Server {
...
+      // server certificate for clusters or client certificate for listeners).
+      // If not specified and a peer certificate is presented it will not be
+      // verified. By default, a client certificate is optional, unless one of
+      // the additional options (
+      // :ref:`require_client_certificate`
+      // :ref:`verify_certificate_spki`
+      // is also specified.
+      //
+      // See :ref:`the TLS overview <arch_overview_ssl_enabling_verification>`
+      // for a list of common system CA locations.
```
hzxuzhonghu: you should not copy all envoy docs

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,57 @@ message Server {
...
+      //     | openssl enc -base64
+      //   NvqYIYSbgK2vCJpQhObf77vv+bQWtc5ek5RIOwPiC9A=
+      //
+      // This is the format used in HTTP Public Key Pinning.
+      repeated string verify_certificate_spki = 2;
+    
+      // An optional `certificate revocation list 
+      // <https://en.wikipedia.org/wiki/Certificate_revocation_list>`_
+      // (in PEM format). If specified, Envoy will verify that the presented
+      // peer certificate has not been revoked by this CRL. If this DataSource
```
hzxuzhonghu: nit: also fix comments

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,57 @@ message Server {
...
+      // contains multiple CRLs, all of them will be used.
+      repeated string crl = 3;
```
hzxuzhonghu: Usually, it is hard to use, have no idea when you need this

networking/v1alpha3/gateway.proto
Phlak106: Neither?  I intended it to be a secret of type kubernetes.io/tls

networking/v1alpha3/gateway.proto
Phlak106: Fair. Will remove.

Phlak106:
I apologize for the extra commits. I'm trying to amend them so they pass the CLA check.

Phlak106:
Abandoning to fix the commit history

PR #970, opened 2019-07-08
add parameter to allow specify HTTP redirect code

ckcd:
This adds a `redirectCode` parameter to `HTTPRedirect` to allow for specify HTTP redirect code. e.g. set `redirectCode` to `308` rather than `301` to avoid method change to `GET`.

See https://github.com/istio/istio/issues/15292 for further information.


networking/v1alpha3/virtual_service.proto
```diff
@@ -855,6 +855,10 @@ message HTTPRedirect {
   // On a redirect, overwrite the Authority/Host portion of the URL with
   // this value.
   string authority = 2;
+
+  // On a redirect, use HTTP statusCode 308 to avoid method change
+  // to GET.
+  bool use_permanent_redirect = 3;
```
rshriram: rather than have a boolean, why not simply have
```
redirectCode: uint32
```

and in validation, we can check if the redirect code is in 3xx range.. I believe envoy allows specifying a redirect code right?

ckcd: yes, `envoy` can specify this redirect code as `301` `302` `307` or `308`, which can be found here:
https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#enum-route-redirectaction-redirectresponsecode

I agree with your opinion ~ will do the change soon~

ckcd: BTW how can I do this validation ?  I didn't see any other `Validate()` code or `protoc-gen-validate` tools. or should I do this in where call this parameter in Istio?

ckcd: @rshriram DONE~ please review it , Thanks !!

ckcd:
@rshriram PTAL. Thanks !

ckcd:
cla test

ckcd:
seems like the CLA failed is because of the `mergify` bot.

@rshriram  can this be merge ? or should I do some rebase for it ?

ckcd:
seems like the CLA failed is because of the `mergify` bot.

@rshriram  can this be merge ? or should I do some rebase for it ?

ckcd:
seems like the CLA failed is because of the `mergify` bot do a merge from master to this branch and it has NO CLA.

@rshriram can you merge this PR ?  thanks ~

PR #973, opened 2019-07-10
Add TLS certificate default validation options

Phlak106:
Envoy exposes a default validation context which gets merged with the validation context from SDS via Pilot. However, those options are not exposed to users. This would allow users to configure the default validation context similar to Envoy and let Pilot merge the contexts but with more defaults.  This change would go hand in hand with a change to Pilot to consume the defaults.

[ ] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[x] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

https://github.com/istio/istio/blob/c5ff36a66084fe7f3b72d5a15203f5783317dbec/pilot/pkg/networking/core/v1alpha3/gateway.go#L424

https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/auth/cert.proto#auth.CertificateValidationContext

https://discuss.istio.io/t/mtls-default-rootca-for-client-certificates/2881/9

Duplicate of https://github.com/istio/api/pull/969 to fix the commit history with correct author tags


Phlak106:
@hzxuzhonghu

Phlak106:
I signed it!

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,28 @@ message Server {
     // Optional: If specified, only support the specified cipher list.
     // Otherwise default to the default cipher list supported by Envoy.
     repeated string cipher_suites = 9;
+
+    // Options for validating a certificate
+    message CertificateValidationContext
+    {
```
venilnoronha: nit: move this to the previous line

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,28 @@ message Server {
...
+      // An optional list of base64-encoded SHA-256 hashes. If specified, the
+      // proxy will verify that the SHA-256 of the DER-encoded Subject Public
+      // Key Information (SPKI) of the presented certificate matches one of
+      // the specified values.
+      repeated string verify_certificate_spki = 2;
+    }
+
+    // Optional: If specified, use these options for default TLS validation.
+    // This context will be merged with the context from SDS.
+    CertificateValidationContext default_validation_context = 11;
```
venilnoronha: Use a full name, like `certificate_validation_context`?

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,27 @@ message Server {
...
+
+    // Options for validating a certificate
+    message CertificateValidationContext {
+      // Name of a TLS certificate secret containing certificate authority
+      // certificates to use in verifying a presented peer certificate (e.g.
+      // server certificate for cluster or client certificate for listeners).
+      // If not specified and a peer certificate is presented it will not be
+      // verified. By default, a client certificate is optional unless TLS
+      // mode is MUTUAL.
+      string trusted_ca = 1;
```
rshriram: how is this different from https://github.com/istio/api/pull/973/files#diff-8aa9e28deea4d3d00698818bb407e3b0R355 ? the ca_certificates field below private_key.

networking/v1alpha3/gateway.proto
```diff
@@ -403,6 +403,27 @@ message Server {
...
+
+      // An optional list of base64-encoded SHA-256 hashes. If specified, the
+      // proxy will verify that the SHA-256 of the DER-encoded Subject Public
+      // Key Information (SPKI) of the presented certificate matches one of
+      // the specified values.
+      repeated string verify_certificate_spki = 2;
```
rshriram: is there a more simpler explanation for this rather than copying the content straight out of envoy? We have tried to simplify the explanations to make it easier for people to use.

networking/v1alpha3/gateway.proto
Phlak106: The one below the private_key is a file mount path.  The file path approach does not leverage secret rotation and doesn't work with SDS. This one is intended to be a true secret that contains namespace or cluster-wide trusted CAs that gets merged with any intermediates that come from the SDS context.

networking/v1alpha3/gateway.proto
Phlak106: I can make it more concise. Not sure about simpler. I think noting the SHA-256 is important.

Phlak106:
@wattli Can you take a look?

networking/v1alpha3/gateway.proto
rshriram: But for SDS, we actually ask the user to create two secrets.. secret, and secret-cacert, IIRC. @myidpt / @wattli ?

Phlak106: Until recently (https://github.com/istio/istio/pull/15496), for SDS you had to create a single opaque secret with server key, server cert, and CAs for client authentication. Now you can create two tls-type secrets and as long as you have the magic suffix "-cacert" SDS will do the right thing. CredentialName will only be the server tls secret and SDS will assume the existence of the -cacert secret. Also, SDS won't really manage the -cacert one though if I understand it correctly.

This approach was started before that PR was picked up by @JimmyCYJ after the original was dropped. It also attempts to avoid the magic suffix.

rshriram: Thanks for the clarification. In that case, its all the more important to not have this trusted_ca secret name inside this message. PTAL at https://github.com/istio/api/pull/973/files#diff-8aa9e28deea4d3d00698818bb407e3b0R373 .. This was the original intent several months ago. If you want to explicitly specify the secret for the cacert instead of using the magic suffix, then please add cacert_credential_name below the `string credential_name` with the appropriate explanation. Otherwise, we would be fragmenting the info across two different proto messages [it turns out strangely in yaml as well as docs].

once you move the cacert_crednetial_name or [whatever]_credential_name out, the only thing left in this message is the verify_certificate_spki field. You can hoist this field into the main proto message itself.

costinm:
And even if we do end up supporting this - adding it to the API directly doesn't make sense, we don't even have a prototype implementation of cert rotation or multicluster public cert merging.

costinm:
Also on the '2 secrets' - I don't thing this was reviewed by Env/Operator/installer - we will need to discuss because it conflicts with the plans for multicluster simplification and setup.

costinm:
To be clear: moving the trusted public keys out of ingress cert is great. It is not clear yet how the set of trusted CAs will be configured and distributed - there are several MC implications. We may allow it to be 
set globally or per namespace.  

Clearly sidecars and Gateways can't access Secrets in a different namespace, and public keys are also clearly not secret.

myidpt:
@Phlak106 Would you like to bring this up in the Istio security WG tomrrow? We can get some feedback from the community as well.

costinm:
Also in Env WG - since it impacts both operator and MC and install config.
I suspect Networking WG must also approve as it's a change to Gateway.

Phlak106:
What times are the WG meetings?

myidpt:
The WG meeting will be at 11AM tomorrow:
https://github.com/istio/community/blob/master/WORKING-GROUPS.md#working-group-meetings

costinm:
For any config and install changes - Env WG is at 10AM on Wed - tomorrow
and every 2 weeks.

rshriram:
FYI All he is doing (in the end) is literally adding verify_pki_sha field just like the way we have tls version checks. Not sure what’s there for all working groups to chime in on for this. 

We already have two secrets for private key and ca cert as mentioned in the docs.

costinm:
Distribution of the root cert is pretty critical for install - operator,
new installer, upgrade, etc. We must make sure changes are consistent and
can be safely applied.

costinm:
For reference - can you add to the PR description a quick link to envoy
reference of the field we would generate ?

costinm:
To be clear - there are several issues here. No problem with adding an
option for pki sha validation consistent with other ones.

But it should not depend on SDS, nor a particular way to configure public
keys distribution. Rotation , deployment, multicluster and other factors
are involved in the rootca distribution, we want a consistent solution.

rshriram:
@Phlak106 can you move the pki sha thing out of the validation context message into the main server message near the TLS version info fields ?

We can tackle the “addition of an api field for specifying the CA cert to verify client connections “ in a separate discussion.

costinm:
Also any reason to support spki but not verify_certificate_hash ? 

I would add it next to the subject_alt_name section, because they serve the same purpose (cipher suites and protocol versions affect the handshake selection, not client auth).

Phlak106:
Can definitely move the SPKI field out and separate from the CA cert options.

As for the has field, I omitted it because of the note on SPKI: This option is preferred over verify_certificate_hash, because SPKI is tied to a private key, so it doesn’t change when the certificate is renewed using the same private key. I can add the hash field too.

Phlak106:
@googlebot I consent

rshriram:
Bad merge

rshriram:
@costinm / @wattli - PTAL.. better to get this in before the code freeze

Phlak106:
Yes, please. Especially since there will also need to be a change to consume this in Pilot.

PR #976, opened 2019-07-12
Test change to ensure mergify is working.

geeknoid:


geeknoid:
Could someone approve this? I'm just trying to see if mergify works as expected.

geeknoid:
@howardjohn Thanks for approving. But you're not in the code owners file, so how come this was merged? Hmmm.

howardjohn:
My understanding is that mergify doesn't care about codeowners. If you have branch protection set up "right" (I am not sure what right is.. installer was this way as of this morning, may not now) it will get blocked, but not gracefully. See https://github.com/Mergifyio/mergify-engine

PR #979, opened 2019-07-15
Autogenerate common Istio annotations

nmittler:


annotation/annotations.gen.go
```diff
@@ -0,0 +1,285 @@
+// Copyright 2019 Istio Authors
```
geeknoid: No need for a copyright on a generated file.

annotation/annotations.yaml
```diff
@@ -0,0 +1,136 @@
+annotations:
```
geeknoid: This one could use a copyright.

annotation/annotations.gen.go
```diff
@@ -0,0 +1,271 @@
...
+//
+// nolint: lll
+//go:generate go run $GOPATH/src/istio.io/api/annotation/cmd/gen/main.go --input=$GOPATH/src/istio.io/api/annotation/annotations.yaml --output=$GOPATH/src/istio.io/api/annotation/annotations.gen.go
+//go:generate goimports -w annotations.gen.go
+
+package annotation
+
+var (
+	IngressClass = Instance{
+		Name: "kubernetes.io/ingress.class",
```
nmittler: @geeknoid I suspect I should remove non-Istio annotations (this is the only one) ... wdyt?

nmittler: Decided to just remove it.

PR #980, opened 2019-07-16
authn: add a field to confrol trust domain validation

yangminzhu:
Note, this changes the internal authn filter API between pilot and proxy, no user-facing API changed.


envoy/config/filter/http/authn/v2alpha1/config.proto
```diff
@@ -33,4 +33,14 @@ message FilterConfig {
...
+
+  // Skips validating the peer's trust domain.
+  // By default, the istio authn filter will reject the request if the peer and
+  // the local service is not in the same trust domain.
+  // Set this field to true to skip the validation and allows peers from any
+  // trust domains.
+  // Note, the istio authn filter only validates the trust domain when both mTLS
+  // policy and the STRICT mode is used, In other words, this field has no
+  // effect when plaintext or PERMISSIVE mode is used.
+  bool skip_validate_trust_domain = 3;
```
yangminzhu: @myidpt I'm planning to only validate the trust domain when mTLS STRICT mode is used as after a second thought, I'm not sure what's the benefit of supporting it for PERMISSIVE mode. Please let me know if I missed any reason for this, thank you.

myidpt: Yes, the concern I have is, only when the users migrate from PERMISSIVE to STRICT, they will notice the failures due to calls from different trust domains. So I would like the mTLS authentication behavior to be the same for PERMISSIVE and STRICT modes.

yangminzhu: Okay, I'm fine with either way if this is not about security but for migration.

yangminzhu:
@myidpt comments updated, PTAL, thanks.

PR #988, opened 2019-07-22
Revert "Add feature gate in global config"

hzxuzhonghu:
Reverts istio/api#958


costinm:
/approve
/lgtm

And please stop making changes to mesh config ( or approving changes ) without informing Env WG and confirming that the change can be supported by installer and operator and can be rolled (upgrade, etc).  Mesh config is not and was not user editable since 0.6, operator is replacing it with a different config.

PR #989, opened 2019-07-22
Add openapi schema to Istio protos

jasonwzm:
The PR supersedes https://github.com/istio/api/pull/965 by adding the tooling to autogenerate the openapi files in each proto directory.

/cc @mpvl


jasonwzm:
@geeknoid I am not sure why I am not able to assign reviewers, but PTAL and merge if it looks good.

mcp/v1alpha1/istio.mcp.v1alpha1.json
```diff
@@ -0,0 +1,314 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "This package defines the common, core types used by the Mesh Configuration Protocol.",
+    "version": "v1alpha1"
+  },
+  "components": {
+    "schemas": {
+      "istio.mcp.v1alpha1.SinkNode": {
+        "description": "Identifies a specific MCP sink node instance. The node identifier is\n presented to the resource source, which may use this identifier\n to distinguish per sink configuration for serving. This\n information is not authoritative. Authoritative identity should come\n from the underlying transport layer (e.g. rpc credentials).",
```
geeknoid: Is that \n in there intentional, or did we not handle linefeed conversion properly somewhere along the line?

jasonwzm: Done.

jasonwzm:
removed the docker image from the PR and updated the generated json. @ozevren PTAL

cue.yaml
```diff
@@ -0,0 +1,36 @@
...
+  mixer/adapter/model/v1beta1:
+    - mode: all
+  mixer/v1/config/client:
+    - mode: all
+  mixer/v1:
+    - mode: all
+      title: This package defines the Mixer API that the sidecar proxy uses to perform precondition checks, manage quotas, and report telemetry.
+  networking/v1alpha3:
+    - mode: perFile
+  policy/v1beta1:
```
ozevren: Shouldn't these be perFile as well?

ozevren: I would have expected that all CRD mapped files to have a per-file OpenAPI schema, so that we can include them in the CRD definitions?

jasonwzm: So the generation here is mapping what we have in doc gen and is for general consumption. The schema used by CRDs will be slightly different, for example MeshPolicy does not have a direct mapping in the current generated schema, so it will needs to be tweaked a bit and fed as values into helm. This will be in parallel to this current pipeline.

ozevren: Why can't we have a single pipeline for both? Having two mechanisms seems very confusing.

jasonwzm:
Adding Nathan when Oz is on vacation.

authentication/v1alpha1/istio.authentication.v1alpha1.json
```diff
@@ -0,0 +1,278 @@
...
+                "type": "string",
+                "format": "string"
+              }
+            }
+          }
+        ]
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
jasonwzm: Noted. This involves in changing the tool. I will follow up with another PR for it.

PR #990, opened 2019-07-22
Update mergify config approver count, to match  with GitHub has.

geeknoid:


go.sum
```diff
@@ -61,10 +62,12 @@ golang.org/x/tools v0.0.0-20190524140312-2c0ae7006135/go.mod h1:RgjU9mgBXZiqYHBn
 google.golang.org/appengine v1.1.0/go.mod h1:EbEs0AVv82hx2wNQdGPgUI5lhzA/G0D9YwlJXL52JkM=
 google.golang.org/genproto v0.0.0-20180817151627-c66870c02cf8 h1:Nw54tB0rB7hY/N0NQvRW8DG4Yk3Q6T9cu9RcFQDu1tc=
 google.golang.org/genproto v0.0.0-20180817151627-c66870c02cf8/go.mod h1:JiN7NxoALGmiZfu7CAH4rXhgtRTLTxftemlI0sWmxmc=
+google.golang.org/grpc v1.21.0/go.mod h1:oYelfM1adQP15Ek0mdvEgi9Df8B9CZIaU1084ijfRaM=
```
howardjohn: this doesn't look right? was this from `go mod tidy`? since it looks like a different version

geeknoid: I have no idea where this came from. I've removed it from the PR.

howardjohn:
nice, I cannot merge it since only 1/2 reviews, so the github side is working

PR #991, opened 2019-07-23
Add ISTIO_MUTUAL TLS mode to gateway

Monkeyanator:
[Relevant issue](https://github.com/istio/istio/issues/15485)

To support secure connections between workloads and gateways with SDS enabled, we must add a new `TLSmode`, `ISTIO_MUTUAL`, to `gateway.proto`. This is a parallel change to the addition of [the ISTIO_MUTUAL flag in `destination_rule.proto`](https://github.com/istio/api/commit/170445541e5b5a1515b52c9c9596570220679a27).

For instance, this will allow us to define a `Gateway` as follows, without explicitly specifying the specific certificate locations:
```yaml
kind: Gateway
metadata:
  name: istio-egressgateway
spec:
  selector:
    istio: egressgateway
  servers:
  - port:
      number: 80
      name: https
      protocol: HTTPS
    hosts:
    - edition.cnn.com
    tls:
      mode: ISTIO_MUTUAL
```

As opposed to

```yaml
kind: Gateway
metadata:
  name: istio-egressgateway
spec:
  selector:
    istio: egressgateway
  servers:
  - port:
      number: 80
      name: https
      protocol: HTTPS
    hosts:
    - edition.cnn.com
    tls:
      mode: MUTUAL
      serverCertificate: /etc/certs/cert-chain.pem
      privateKey: /etc/certs/key.pem
      caCertificates: /etc/certs/root-cert.pem
```
Note that with SDS enabled, there is no secret mounting happening, and it is therefore not possible to explicitly reference mounted secrets as shown in the second example.

@incfly @quanjielin fyi


networking/v1alpha3/gateway.proto
```diff
@@ -334,6 +334,14 @@ message Server {
       // their respective endpoints. Use of this mode assumes that both the
       // source and the destination are using Istio mTLS to secure traffic.
       AUTO_PASSTHROUGH = 3;
+
+      // Secure connections to the upstream using mutual TLS by presenting
```
incfly: hmm, i know you copy from `mutual` mode above, but the `upstream` and `client certificate` seems just opposite of what gateway's server section is about. it supposed to be downstream and server certificate.

@rshriram could you confirm?

networking/v1alpha3/gateway.proto
```diff
@@ -334,6 +334,14 @@ message Server {
...
+      // client certificates for authentication.
+      // Compared to Mutual mode, this mode uses certificates generated
+      // automatically by Istio for mTLS authentication. When this mode is
+      // used, all other fields in `TLSSettings` should be empty.
```
incfly: s/empty/are ignored.

networking/v1alpha3/gateway.proto
```diff
@@ -334,6 +334,14 @@ message Server {
...
+      // Allows users to make egress requests through gateway with SDS enabled
```
incfly: you probably don't need to focus on SDS. I imagine the implementation make `ISTIO_MUTUAL` works for secret mount as well.

networking/v1alpha3/gateway.proto
```diff
@@ -334,6 +334,14 @@ message Server {
       // their respective endpoints. Use of this mode assumes that both the
       // source and the destination are using Istio mTLS to secure traffic.
       AUTO_PASSTHROUGH = 3;
+
+      // Secure connections to the upstream using mutual TLS by presenting
+      // client certificates for authentication.
+      // Compared to Mutual mode, this mode uses certificates generated
+      // automatically by Istio for mTLS authentication. When this mode is
```
incfly: add, `representing gateway workload identity, generated automatically by Istio`?

incfly:
@rshriram @vadimeisenbergibm do you want to take another pass?

One thing I'm not sure about if we want to mark `<hidden-from-doc>` before we confirm this will be delivered in 1.3 release?

costinm:
I don't get it - for sidecars we autoload the internal certs. Gateway is not really an API to configure the internal loading of istio certs. I think Vadim proposed an install option to indicate a gateway is used for egress, we can also use other install options. 

The config for mtls must be loaded before pilot is contacted - so the API is not really implementable... The internal certs must be mounted or loaded at startup.

/Hold

incfly:
If you're talking about gateway's own envoy bootstrap config's SDS config used to talking to control plane, e.g. pilot, true, that requires helm/installation level option.

This FR and PR are about configuring data plane SDS for egress gateway. We need an API to configure gateway api dynamically in runtime.

Using the correct SDS talking to Pilot, I would treat that as part of control plane SDS, rather than the scope of this FR. I agree that we need to solve both though.

@vadimeisenbergibm do you want to confirm above and maybe clarify?

vadimeisenbergibm:
@incfly let me review the PR.
@costinm This PR is not related to https://github.com/istio/api/pull/556. The current PR is valid for both file-mounted certs and for SDS, it just instructs the gateway to expect Istio data-plane mTLS incoming connections on the configured port.

With SDS, the gateway gets the certs for Istio data-plane mTLS from the node agent, in the same way as sidecar proxies do, and uses them for incoming connections. The gateway receives TLS context configuration that uses SDS from pilot, as the sidecars do. The operation is the same as with the sidecar proxies.

networking/v1alpha3/gateway.proto
vadimeisenbergibm: right, it should be _secure connections from the downstream using mutual TLS, presenting server certificates_.

costinm:
Vadim, generally I don't mind configuring data-plane mtls in Gateway and
Sidecar instead of AuthPolicy. Just want to have consistency and make sure
it makes sense for the API.

Since APIs need to be supported for a very long time - let's have more
example and discuss in networking, usability.

Is there any use case where egressgateway would not use mtls ? Or would use
mtls on some ports only ?

vadimeisenbergibm:
@costinm According to the current API, you have to configure explicitly which protocol gateway expects at a particular port. In particular, you must specify TLS mode and to provide the certs. The current values for the [TLS server side mode](https://istio.io/docs/reference/config/networking/v1alpha3/gateway/#Server-TLSOptions-TLSmode) are SIMPLE, MUTUAL, PASSTHROUGH and AUTO_PASSTHROUGH . For some reason, ISTIO_MUTUAL mode is missing. For the [client side TLS mode](https://istio.io/docs/reference/config/networking/v1alpha3/destination-rule/#TLSSettings-TLSmode), in the DestinationRule, the values are  DISABLE, SIMPLE, MUTUAL and ISTIO_MUTUAL.

vadimeisenbergibm:
@costinm
> Is there any use case where egressgateway would not use mtls 

In case of passthrough: the client uses (m)TLS and the gateway passes the traffic through as is

costinm:
How would it apply to inbound gateways ?

As I mentioned I am all for consolidating tls settings in gateway and
sidecar, but I want to make sure we do it cleanly and consider all
implications, it's been a major source of peoblems and confusion.

vadimeisenbergibm:
@costinm In case of mesh federation, when you have common trust between two meshes, a workload in `mesh1` can talk to an ingress gateway in `mesh2` using Istio mutual TLS. The ingress gateway will perform TLS termination, and forward the request, for example by HTTP routing to a service in `mesh2`, and perform mTLS origination if needed.

PR #992, opened 2019-07-24
Follow a comment about the default value of `min_health_percent`.

mnktsts2:
Related to https://github.com/istio/istio/pull/15609 .

By the PR,  the default value of `min_health_percent` was changed to `0%` from `50%`.


networking/v1alpha3/destination_rule.proto
```diff
@@ -511,7 +511,9 @@ message OutlierDetection {
   // pool has at least min_health_percent hosts in healthy mode. When the
   // percentage of healthy hosts in the load balancing pool drops below this
   // threshold, outlier detection will be disabled and the proxy will load balance
-  // across all hosts in the pool (healthy and unhealthy).  The default is 50%.
+  // across all hosts in the pool (healthy and unhealthy). the threshold can be
```
venilnoronha: s/the/The

PR #993, opened 2019-07-24
Generate a doc page for the annotations defined in annotations.yaml.

geeknoid:
This ends up regenerating all the .go and .py files due to the updated container image we're using. The updated generated go code appears to be a performance improvement delivered  by the latest  gogo code.


geeknoid:
@kyessenov Need two reviewers in this repo. Mind approving please?

PR #994, opened 2019-07-25
Simplify protogen container builder

sdake:
The old continer builder was based on Golang 1.9 (according to
@geeknoid; I couldn't actually get it to build).  The container
could use some serious simplificaiton.  If we don't need to pin gRPC
and protobuf, it is better practice to let the experts in golang
distribute their own container (golang:1.12) and work from that.

I have assumed in this PR that they have the correct versions of
gRPC and protobuf, which may be a bad assumption.

Also this PR does not yet build (so don't merge) because of a
dependency problem in the tools repo.  Further, this PR likely
won't build and certainly isn't tested yet, until the tools
repo build problem is sorted out.

Depends-On: https://github.com/istio/tools/pull/222

### DIFF
--- a/tools/protoc/Dockerfile
+++ b/tools/protoc/Dockerfile
@@ -1,14 +1,8 @@
-# Based on: https://github.com/znly/docker-protobuf/blob/master/Dockerfile
-# Modified to slim down the image and adjust for Istio
-
 FROM golang:1.12 as build-env
 
 RUN mkdir -p /out_build_env/usr/bin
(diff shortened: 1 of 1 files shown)


kyessenov:
I think we do need to pin protoc-gen-go and grpc. Both are under active development. protoc-gen-go generates different files depending on the version which will break reproducibility.

sdake:
@kyessenov fair enough - your an expert in this area and I''m not :)  Are you satisified with the current pins (latest rev) or want me to backrev?

PR #997, opened 2019-07-26
Use a 80% smaller image for protoc generator

sdake:
Prow doesn't like 4.3gb images.  As a result, use a much smaller
container that uses build contexts.  The dependent container needs
to be rebuilt prior to this merging.

Depends-On: https://github.com/istio/tools/pull/229

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all: generate
 # docker_gen
 ########################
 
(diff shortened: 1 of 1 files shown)


geeknoid:
Replaced by #998 which includes a needed source update.

PR #1001, opened 2019-07-28
Use a consolidated dockerfile

sdake:
This is built based upon the consolidated build-api-site Dockerfile:
https://github.com/istio/tools/pull/230

The Makefile is particularly relevant. I am unclear if the other
changes are correct, however, this is what the tooling is generating.

Note the dockerfile itself needs be pushed prior to testing this PR
and an appropriate tag specified.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all: generate
 # setup
 ########################
 
-apitools_img := gcr.io/istio-testing/api-build-tools:2019-07-26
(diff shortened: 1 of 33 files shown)


geeknoid:
Changes look good.

I've rebuilt and pushed the new docker image. The new tag  you can use in the makefile is gcr.io/istio-testing/api-build-tools:2019-07-29

geeknoid:
Supplanted by #1004

PR #1003, opened 2019-07-29
EnvoyFilter: add ability to match on proxy metadata and version

rshriram:
Based on @ramaraochavali's concerns that EnvoyFilters would pose
a potential problem during upgrades, adding a proxy version check
in the match condition. This will allow users to apply the patches
to specific versions of the proxy


rshriram:
@ramaraochavali FYI. This should hopefully alleviate the upgrade issue. We can add more checks under the Proxy Version over time.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -366,6 +366,23 @@ message EnvoyFilter {
...
 
+  // ProxyVersion restricts the patches to sidecars or gateways with
+  // from a specific version of Istio. The Istio version for a given
+  // proxy is obtained from the node metadata field ISTIO_VERSION
+  // supplied by the proxy when connecting to Pilot. This value is
+  // embedded as an environment variable (ISTIO_META_ISTIO_VERSION) in
+  // the Istio proxy docker image. Custom proxy implementations should
+  // provide this metadata variable to take advantage of the Istio
+  // version check option.
+  message ProxyVersion {
```
ramaraochavali: @rshriram This is great. Thanks for handling this.

rshriram: actually check the proto again. I have made this a lot more generic.

howardjohn:
This is not OWNERs approved by the way. I think there is a bug in our branch protection

networking/v1alpha3/envoy_filter.proto
```diff
@@ -366,6 +366,35 @@ message EnvoyFilter {
     GATEWAY = 3;
   };
 
+  // One or more properties of the proxy to match on.
+  message ProxyMatch {
+    oneof version_match {
```
louiscryan: why a oneof ? why not a regex ?

howardjohn:
This was not approved by owners @rshriram ..

networking/v1alpha3/envoy_filter.proto
rshriram: I dont mind doing regex. But i thought you were strongly opposed to doing regexes at all. :)

rshriram:
I talked to @louiscryan offline. So no harm done..

PR #1008, opened 2019-07-30
move proxy match to the workload selector

rshriram:
Workload selector is the appropriate place to match based on
specific properties of a proxy such as labels, or the proxy metadata.
This will also come in handy for Sidecar.


networking/v1alpha3/sidecar.proto
```diff
@@ -349,12 +349,30 @@ message IstioEgressListener {
...
   map<string, string> labels = 1;
 
+  // A regular expression in golang regex format (RE2) that can be
+  // used to select proxies using a specific version of istio
+  // proxy. The Istio version for a given proxy is obtained from the
+  // node metadata field ISTIO_VERSION supplied by the proxy when
+  // connecting to Pilot. This value is embedded as an environment
+  // variable (ISTIO_META_ISTIO_VERSION) in the Istio proxy docker
+  // image. Custom proxy implementations should provide this metadata
+  // variable to take advantage of the Istio version check option.
```
mandarjog: This should be expression and it will be evaluated per semVer ?

rshriram: lol I wish.. stuff from our release branches has release-1.x as the version value. Stuff from master has some random stuff.. This is not semver, though the implicit contract is that it is

networking/v1alpha3/envoy_filter.proto
```diff
@@ -596,16 +567,13 @@ message EnvoyFilter {
     // inbound traffic to sidecar and outbound traffic from sidecar.
```
mandarjog: clarification on gateway. gateway is treated as inbound, can add that here?

rshriram: We have a separate context for gateway called GATEWAY.

networking/v1alpha3/sidecar.proto
yangminzhu: I'm not sure if it's a good idea to expose this API in the workload selector, it seems it's exposing too much internal implementation details to end users. 

The other thing I'm worried is that this makes it harder to support custom proxy implementation. This is adding a contract between Pilot and the proxy metadata, probably need some more design discussions.

I'll leave this for other TOC members to decide.

If we decided to add this, would it be better to use a CEL expression that matches directly against the node metadata?

```protobuf
message WorkloadSelector {
  map<string, string> labels = 1;

  // A CEL expression that matches on the node metadata supplied by a proxy when connecting to Istio Pilot.
  // For example, expression `ISTIO_VERSION in ["1.2", "1.1"]` that selects a specific version of istio proxy.
  string metadata = 2;
}
```
This seems more flexible and you don't need to separate `proxy_version` and `proxy_metadata`.

rbac/v1alpha1/rbac.proto
```diff
@@ -103,6 +103,24 @@ message WorkloadSelector {
   // search is restricted to the configuration namespace in which the the
   // resource is present.
   map<string, string> labels = 1;
+
```
yangminzhu: Please see my other comment, I tend to not to add this new API to RBAC unless we're sure it's going to be useful for RBAC.

cc @liminw

networking/v1alpha3/sidecar.proto
howardjohn: Maybe we should change our versioning to be semver. Daily builds could be 1.2-date, master could be set the the next version immediately (today we would release 1.3-073019 or something)

rshriram: @yangminzhu exposing Istio version is not same as exposing istio internals. This simply lets people specify some set of settings for older proxies and something else for newer proxies. The CEL expression just to match this small thing is an absolute overkill in this scenario. All we need is a simple regex or sem ver based version match

louiscryan: Given the blast radius of this change I think we should keep this out of WorkloadSelector for the time being

Agree we don't need CEL unless we make it holistically our workload selection mechanism

Agree we should be SemVer for ISTIO_VERSION

rbac/v1alpha1/rbac.proto
kyessenov: Agree. This seems out of place.

networking/v1alpha3/sidecar.proto
kyessenov: I think this should be capability-based instead of hard-coding proxy version. It's not even clear how this works with custom builds of proxy that don't follow istio versioning (e.g. raw envoy).

networking/v1alpha3/sidecar.proto
```diff
@@ -349,12 +349,30 @@ message IstioEgressListener {
...
+  string proxy_version = 2;
+
+  // Match on the node metadata supplied by a proxy when connecting
+  // to Istio Pilot. Note that while Envoy's node metadata is of
+  // type Struct, only string key-value pairs are processed by
+  // Pilot. All keys specified in the metadata must match with exact
```
kyessenov: That won't work for standard istio metadata. It's two level, under istio.io/metadata.

geeknoid:
@rshriram What's the state of this PR? Keep it, or nuke it?

rshriram:
So, I can convert this to an issue. waiting on @yangminzhu and co for some direction

yangminzhu:
> So, I can convert this to an issue. waiting on @yangminzhu and co for some direction

@rshriram I think you will need to change the common workload selector now instead of changing the old one in sidecar.proto, I'm fine with the change as long as it's approved by the TOC.

PR #1010, opened 2019-07-30
Documented per-workload proxy resource overrides

nrjpoddar:
Fixes: https://github.com/istio/istio/issues/15936

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -223,7 +223,7 @@ var (
           Name: "sidecar.istio.io/proxyCPU",
           Description: "Specifies the requested CPU setting for the istio-proxy "+
                         "sidecar.",
-          Hidden: false,
+          Hidden: true,
           Deprecated: false,
         }
 	
@@ -239,7 +239,7 @@ var (
           Name: "sidecar.istio.io/proxyMemory",
           Description: "Specifies the requested memory setting for the "+
                         "istio-proxy sidecar.",
-          Hidden: false,
+          Hidden: true,
           Deprecated: false,
         }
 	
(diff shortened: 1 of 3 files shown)


nrjpoddar:
I verified that the overrides work but I'm not sure why they were hidden to begin with? Example:

```yaml
- apiVersion: v1
  kind: Pod
  metadata:
    annotations:
      sidecar.istio.io/proxyMemory: 1Gi
```
results in the following sidecar proxy:
```yaml
      name: istio-proxy
      ports:
      - containerPort: 15090
        name: http-envoy-prom
        protocol: TCP
      readinessProbe:
        failureThreshold: 30
        httpGet:
          path: /healthz/ready
          port: 15020
          scheme: HTTP
        initialDelaySeconds: 1
        periodSeconds: 2
        successThreshold: 1
        timeoutSeconds: 1
      resources:
        requests:
          memory: 1Gi
```
```

File: annotation/annotations.gen.go [hunk 1]
```diff
@@ -239,7 +239,7 @@ var (
           Name: "sidecar.istio.io/proxyMemory",
           Description: "Specifies the requested memory setting for the "+
                         "istio-proxy sidecar.",
-          Hidden: false,
+          Hidden: true,
```
howardjohn: maybe I am missing something, but aren't you hiding this? And it is already documented currently?

From the description and issue it sounds like you are trying to add documentation but the PR looks like its removing it?

nrjpoddar: Jeez, I completely messed it up. It's already showing up correctly in the Istio 1.3 preliminary docs. It was updated recently and I had an old version of the repository. Thanks for catching this. Cheers!

nrjpoddar:
Addressed by https://github.com/istio/api/pull/979

PR #1012, opened 2019-07-31
fix proxy version match

rshriram:


rshriram:
@louiscryan fixing the proxy version to use regex as you suggested.

PR #1014, opened 2019-07-31
Point to the right version of generated RPC protos.

geeknoid:


kyessenov:
What's the reason to move away from gogo/googleapis-genproto?

geeknoid:
The situation is a mess.

We've had the google/rpc code generation in istio/gogo-genproto for a long time. Some code in Istio was using those definitions, while other code used the definitions from gogo/googleapis instead. That's not great, but it's been working.

Istio depends on a fix that was pushed into the gogo/protobuf repo. In order to  get this fix, we need to get the latest unreleased  version of the gogo code. There has been some code generation changes in gogo, making old generated code not compatible with the new generated code. You have to regenerate everything.  Well, because gogo hasn't been officially released yet, the generated code exposed by  the gogo/googleapis repo is now out of date and incompatible with code generated with the latest gogo tools.

The solution I'm pursuing is thus to consolidate all use of the googleapis stuff in Istio on the versions in gogo-genproto. This code is all generated with the latest gogo tools and so works nicely with  the rest of Istio.

The go-control-plane package was the last bit still depending on the gogo/googleapis repo.

Note that this package also depended on both istio.io/gogo-genprotos and gogo/googleapis. Now it's only depending on istio.io/gogo-genprotos.

PR #1016, opened 2019-08-01
Operator API

ostromart:


operator/v1alpha1/operator.proto
```diff
@@ -0,0 +1,679 @@
...
+spec:
+
+2. Default minimal profile install
+
+spec:
+  profile: minimal
+
+3. Default install with telemetry disabled
+
+spec:
```
costinm: What does it even mean for telemetry to be enabled ? Envoy will collect metrics. Prometheus has its own operator. 
Is it for the to-be-deprecated mixer ?

operator/v1alpha1/operator.proto
```diff
@@ -0,0 +1,679 @@
...
+      namespace: istio-policy
+  telemetry:
+    components:
+      namespace: istio-telemetry
+  config_management:
+    components:
+      namespace: istio-config-management
+  security:
+    components:
+      citadel:
```
costinm: This implies that the operator is limited by the same 'must run with cluster-wide perm (create ns, etc)' as Tiller. 
Is this the right design, or will it limit its capabilities/adoption ?

operator/v1alpha1/operator.proto
```diff
@@ -0,0 +1,679 @@
...
+        namespace: istio-cert-manager
+      node_agent:
+        namespace: istio-node-agent
+
+5. Default install with specialized k8s settings for pilot
+
+spec:
+  traffic_management:
+    components:
+      pilot:
```
costinm: Isn't this too tied to a particular component name ? Shouldn't it be more like kustomize, where you match by name and apply patches ?

operator/v1alpha1/operator.proto
```diff
@@ -0,0 +1,679 @@
...
+            successThreshold: 33
+
+6. Default install with values.yaml customizations for proxy
+
+spec:
+  traffic_management:
+    components:
+      proxy:
+        values:
+        - global.proxy.enableCoreDump: true
```
costinm: I don't know if this is example or not - but neither of them belong to operator, but to Istio networking default configs.

operator/v1alpha1/operator.proto
```diff
@@ -0,0 +1,679 @@
...
+            - path: spec.template.spec.containers.[name:galley].command.[--livenessProbeInterval]
+              value: --livenessProbeInterval=123s
+*/
+
+// IstioControlPlane is a CustomResourceDefinition (CRD) describing an Istio control plane.
+message IstioControlPlane {
+    // Spec defines the desired state of IstioControlPlane.
+    IstioControlPlaneSpec spec = 3;
+    // Status reports the status of the Istio control plane.
+    InstallStatus status = 4;
```
costinm: That's not commonly done and conflicts with plans for UDPA and general event or metric based status reporting. 
I really don't think we should start this route. 

Fine to report status in the k8s CRD - but I don't think this is part of the message.

geeknoid:
Your PR needs to include the generated .pb.go, pb.html and pb.py2 files.

Also, I see the name of the included .html file doesn't include the word operator, so some name is not being parsed/processed correctly. This is probably what's leading to the permission denied problem.

operator/v1alpha1/operator.proto
ostromart: Right now it means that no telemetry component is installed. It's a feature level setting so it can evolve.

operator/v1alpha1/operator.proto
ostromart: Yes that's the limitation and for that reason its use is likely to be limited initially. The API should work as well with the CLI since there are few situations where changes are not triggered by a change in user config. At the same time, some type of controller is needed in some cases, like managed scenarios. 
We can split up the permissions for operator controller with the tradeoff that it will lead to more complexity for installation.

operator/v1alpha1/operator.proto
ostromart: kustomize is tied to component too - just because you select a resource by name doesn't mean you don't have to know which component it belongs to. I think it actually makes more sense to specify a component for k8s settings than looking at the group of resources associated with that component.

operator/v1alpha1/operator.proto
ostromart: Agreed - this is the passthough Helm API which we'll hopefully whittle down to nothing in coming releases.

operator/v1alpha1/operator.proto
ostromart: This is just following the operator CRD pattern which does include status normally. Won't the k8s status be lacking in context/structure?

ostromart:
Added the python files - the go files still hit the permissions error. I think it's related to something going on with funky substitution because I see the following:

operator_v1alpha1_protos=operator/v1alpha1/operator.proto
operator_v1alpha1_pb_gos=operator/v1alpha1/operator.pb.go
operator_v1alpha1_pb_pythons=python/istio_api/operator/v1alpha1/operator_pb2.py
operator_v1alpha1_pb_doc=operator/v1alpha1/istio.operator.v1alpha1.pb.html

But the docs actually still get generated to operator/v1alpha1/v1alpha1.pb.html (operator is missing). 

Comparing protoc command lines between authn and operator the path for doc looks correct:

> make generate-authn

@docker run --user 325877 -v /etc/passwd:/etc/passwd:ro --rm
-v /usr/local/google/home/mostrowski/go/src/istio.io/api:/src/istio.io/api
-w /src gcr.io/istio-testing/api-build-tools:2019-07-31
protoc -I/usr/include/protobuf -Iistio.io/api
--gogofast_out=blabla
--docs_out=warnings=true,dictionary=istio.io/api/dictionaries/en-US,custom_word_list=istio.io/api/dictionaries/custom.txt,mode=html_fragment_with_front_matter:istio.io/api/authentication/v1alpha1
--python_out=,:istio.io/api/python/istio_api authentication/v1alpha1/policy.proto

> make generate-operator

@docker run --user 325877 -v /etc/passwd:/etc/passwd:ro --rm
-v /usr/local/google/home/mostrowski/go/src/istio.io/api:/src/istio.io/api
-v /usr/local/google/home/mostrowski/go/src/istio.io/api/../../k8s.io:/src/k8s.io
-w /src gcr.io/istio-testing/api-build-tools:2019-07-31
protoc -I/usr/include/protobuf -Iistio.io/api -I/src
--gogofast_out=blabla
--docs_out=warnings=true,dictionary=istio.io/api/dictionaries/en-US,custom_word_list=istio.io/api/dictionaries/custom.txt,mode=html_fragment_with_front_matter:istio.io/api/operator/v1alpha1
--python_out=,:istio.io/api/python/istio_api operator/v1alpha1/operator.proto

costinm:
Istio api is generally k8s independent - anything that describe the mesh
config needs to remain this way, can't be operator specific. Including info
on what is enabled.

I think status will be a different proto ? We never used this pattern
AFAIK, but pretty sure all api protos end up in spec portion.

costinm:
Also - first please merge values.yaml proto, and indicate which of the
fields in new proto map to what old fields, since a migration will be
needed.

And I don't think spending time on new api when we haven't even finished an
alpha is best use of review time...

costinm:
We will discuss this in the Env WG and let the ToC decide - I think having operator-specific APIs that duplicate or overlay with both 'core' Istio API areas and with the existing install methods is not acceptable. We don't want any 'operator-only' API - any configuration that touches Istio behavior like defaults for networking, or describe the mesh should be common for all install methods (including non-k8s installs).

We agreed that a proto-encoded representation for values.yaml will be used as common install configuration in 1.3 - it doesn't seem this was merged, but since Istio 0.8 it is the de-factor install API,
and the consensus seems to be to move towards 'istio-config' and proto merge as mechanism to provide defaults.

geeknoid:
@ostromart Do we want to keep this PR at this point?

ostromart:
@geeknoid, closing this, I'm going to redo the work per the API integration proposal.

PR #1020, opened 2019-08-06
Update the common files.

geeknoid:


Makefile
```diff
@@ -1,339 +1,52 @@
...
+	-w /work $(IMG)
+endif
+
+MAKE = $(RUN) make -f Makefile.core.mk
+
+%:
+	@$(MAKE) $@
+
+default:
+	@$(MAKE)
```
sdake: planning to transition this repo to the new build system with this pr?

scripts/run_gobindata.sh
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+SCRIPTPATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ROOTDIR=$(dirname "${SCRIPTPATH}")
+
+img=gcr.io/istio-testing/api-build-tools:2019-07-30
```
sdake: note for the future - we should attempt to consolidate this container into the build-tools container.  I believe it contains a subset of the build-tools container at this time.

Makefile.core.mk
```diff
@@ -0,0 +1,339 @@
+all: generate
```
sdake: Is this a git mv only?

Makefile
geeknoid: Nope. Just updating the common files.

scripts/run_gobindata.sh
geeknoid: Correct. Next step.

Makefile.core.mk
geeknoid: Yes, but sadly since I'm adding another file called Makefile, git doesn't realize I did a rename for this file.

Makefile
sdake: Oh build-with-container=0. got it.

Makefile.core.mk
sdake: thanks.

scripts/run_gobindata.sh
sdake: next of many :)

Makefile.core.mk
```diff
@@ -14,9 +14,9 @@ repo_mount := $(mount_dir)/istio.io/api
 out_path = .
 uid := $(shell id -u)
 
-protoc = docker run --user $(uid) -v /etc/passwd:/etc/passwd:ro --rm -v $(pwd):$(repo_mount) -w $(mount_dir) $(apitools_img) protoc -I/usr/include/protobuf -I$(repo_dir)
+protoc = docker run --user $(uid) -v /etc/passwd:/etc/passwd:ro --rm -v $(pwd):$(repo_mount) -w $(mount_dir) $(buildtools_img) protoc -I/usr/include/protobuf -I$(repo_dir)
```
sdake: you will need many of the same mounts to get this moving that are in the main makefile.

sdake: also the dockerfile may need permissions changes to get this working - I have the existing dockerfile pretty locked down perm wise.

PR #1021, opened 2019-08-06
Fix OpenAPI dependency tracking.

geeknoid:
Typing "make" twice would run cue twice because dependencies from *.proto files to *.json files weren't being tracked. Now with the right dependency tracking in place, the second time you run "make" nothing happens because all targets are up to date.


geeknoid:
@jasonwzm There's something funny going on. When I build the open api stuff, I get:

Building envoy_filter.json...

printed, yet there  is no envoy_filter.json file output anywhere I can see and it's not mentioned in the makefile either. So what's the deal?

geeknoid:
Question answered, envoy_filter.json is in the networking directory instead of the envoy directory, so I had missed it.

geeknoid:
Ping.

PR #1023, opened 2019-08-07
[proxy] Add TLS Settings for envoy ALS and metrics services

rshriram:
Currently these services are configured in the bootstrap without TLS settings.
Users have reported that they would like to use mTLS between envoy and ALS/metrics, just
like the way we use mTLS between envoy and mixer. This PR enables such functionality.

Note that the ALS field was introduced post 1.2. So this is not a breaking change.
I have deprecated the older Envoy metrics service and added the newer format, to maintain
consistency with the ALS message.


rshriram:
@ramaraochavali I forgot the github handle of the person from Salesforce who added the metrics service option. Can you get the appropriate people to take a look at this and let us know if this change is okay?

ramaraochavali:
@rshriram sure. @joeyb can you PTAL?

joeyb:
@rshriram - Looks great. Especially glad to see that this gives us the ability to set the client cert information.

Are there already plans to similarly expose the pilot/mixer/etc. TLS settings? That's something that we currently manually override in the bootstrap template, but it would be great to move that out to a config and get back on the unmodified template.

rshriram:
> Are there already plans to similarly expose the pilot/mixer/etc. TLS settings? That's something that we currently manually override in the bootstrap template, but it would be great to move that out to a config and get back on the unmodified template.

Thats a great idea! open up a PR, similar to this one, changing discovery_address to a RemoteService. The mixer piece is actually easy because you can create a destination rule for istio-mixer service with your custom certificate paths.

mesh/v1alpha1/proxy.proto
```diff
@@ -218,5 +217,22 @@ message ProxyConfig {
   // sent. (e.g. accesslog-service:15000). See [Access Log
   // Service](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/accesslog/v2/als.proto)
   // for details about Envoy's gRPC Access Log Service API.
-  string envoy_access_log_service_address = 22;
+  RemoteService envoy_access_log_service = 22;
```
howardjohn: Don't we need to change the number here or it will be incompatible? Or is that just for on-the-wire compatibility which doesnt matter for mesh config?

rshriram: We can replace it because this is not released yet.

yangminzhu:
The change looks good to me. cc @diemtvu @incfly in case they have other comments.

PR #1024, opened 2019-08-08
Add Keepalive settings for envoy ALS and metrics services

hanahmily:
Currently, these services are configured in the bootstrap without TCP Keepalive settings.
Some users have reported that they encountered some reconnection issues. When to resort traffic in application cluster powered by istio after stopping it some hours ago, envoy should take a few minutes to connect these services.

Signed-off-by: Gao Hongtao <hanahmily@gmail.com>


mesh/v1alpha1/proxy.proto
```diff
@@ -235,4 +235,7 @@ message RemoteService {
   // uses Istio mutual TLS and shares the root CA with Pilot, specify the TLS
   // mode as `ISTIO_MUTUAL`.
   istio.networking.v1alpha3.TLSSettings tls_settings = 2;
+
+  // If set then set SO_KEEPALIVE on the socket to enable TCP Keepalives.
+  istio.networking.v1alpha3.ConnectionPoolSettings.TCPSettings.TcpKeepalive tcp_keepalive = 3;
```
rshriram: nice!!

PR #1025, opened 2019-08-08
Switch to the build-tools image

geeknoid:


howardjohn:
also see - no codeowners required

geeknoid:
Yeah, I'm working on the lint rule. The build-tools and api-build-tools container weren't lining up.

As for code owners, it's because I'm an admin and GitHub seems to consider that sufficient I think.  The permission is set in the GitHub UI to demand CODEOWNERS reviews for this repo. But we don't have  the "enforce branch protection for admins" option on at the moment. We can turn  that on globally  once test-infra is using CODEOWNERS. Coming soon.

geeknoid:
See https://github.com/istio/api/pull/1025

PR #1026, opened 2019-08-08
Named routes in VS and few more fields in envoyfilter api

rshriram:
Add a name field to HTTP routes and matches in a virtual service. Envoy access logs can print the route names, thus helping in debugging.

Ability to match filter chains by name.
Ability to match filter chain by application protocols. After addition of sniffing logic,
we now have two filter chains [one with app protocol match and another without match]. So
people need to be able to match the specific filter chain for their HTTP connection manager tweaks,
else bad stuff will happen.

Ability to match http routes inside a virtual host by name/action.
The latter is useful for people who want to tweak the redirect responses, 
or direct responses from envoy at a gateway for things like custom 404 
messages, etc.


networking/v1alpha3/envoy_filter.proto
```diff
@@ -462,27 +492,39 @@ message EnvoyFilter {
...
     // to select a specific filter chain to patch.
     message FilterChainMatch {
+      // The name assigned to the filter chain.
+      string name = 1;
+
       // The SNI value used by a filter chain's match condition.  This
       // condition will evaluate to false if the filter chain has no
       // sni match.
-      string sni = 1;
+      string sni = 2;
```
louiscryan: I assume you're renumbering now because we haven't shipped :P

rshriram: Yep.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -462,27 +492,39 @@ message EnvoyFilter {
...
+      // the tls_inspector listener filter.
       //
       // Accepted values include:
       //
       // * `raw_buffer` - default, used when no transport protocol is detected.
       // * `tls` - set when TLS protocol is detected by the TLS inspector.
-      string transport_protocol = 2;
+      string transport_protocol = 3;
+
+      // Applies only to sidecars. If non-empty, a comma separated set
```
louiscryan: Why is this only sidecar ? Should work for Gateway too

rshriram: Because this is part of the filter chain match where we have application_protocols field in envoy. The ALPN stuff is part of the TLS context - we are not doing ALPN stuff here.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -417,13 +424,40 @@ message EnvoyFilter {
   // the patch to be applied to a route configuration object or a
   // specific virtual host within the route configuration.
   message RouteConfigurationMatch {
+    // Match a specific route inside a virtual host in a route configuration.
+    message RouteMatch {
+      // The Route objects generated by default are named as
```
howardjohn: Are we turning our implementation details into an API with this? Or are we serious that this is break-glass and are ok with breaking this at any time.

I don't want to be in a position where we can't change our internal naming.

rshriram: We only name our default inbound/outbound routes as "default". Other names are purely user specified and they come from the VirtualService.httpRoute.Name .. So we wont have to worry about exposing our internal semantics to end users..

PR #1028, opened 2019-08-08
Add feature flag enable_auto_mtls in mesh config.

incfly:
API Part for https://github.com/istio/istio/issues/14524


incfly:
/cc @linsun @GregHanson 

I use `hide_from_doc` macro, but let me know if this supposed to be merged till freeze/1.3 branch is cut.

incfly:
@rshriram Points are valid. I owe an full description of the overall plan. The doc is under polishing, will update here once rewriting is done. Briefly,

- The `auto` mode is intended to be defined in root namespace, and in case when customers' own DR not having TLSSettings, root namespace TLS auto mode will be used. Thus this is to explicitly avoid user manual configuration.
- We'll add labeling information to endpoints, for k8s case, sidecar injector automatically does that. And Pilot tracks if all endpoints having that label, for a given service. If yes, auto mode kick in.

costinm:
Agree with Shriram.

Let's postpone this for 1.4 - when we'll have more info.

costinm:
The feature can still ship - using the annotation and/or global flag.

rshriram:
> We'll add labeling information to endpoints, for k8s case, sidecar injector automatically does that. And Pilot tracks if all endpoints having that label, for a given service. If yes, auto mode kick in.

thats still not enough as we do mtls setup in cds code.. trying to enumerate all endpoints of a cluster at cds time is very costly. Second, eds has incremental updates within a single cluster but not cds. So i think you need to invest in a cluster level filter that sees if the remote endpoint has istio labels or not and initiate tls to that endpoint. @kyessenov would know better. i think we should do that first and foremost.

incfly:
@rshriram I understand your concern. We won't enumerate in CDS code. Let us have discussion on the design doc. I'll post here once I believe that's ready.

costinm:
I don't think CDS is an issue - the endpoints structures hold list of
endpoints and all labels. When an endpoint
status changes we generate the incremental EDS push. At this point we can
check a bool indicating if the
endpoint set is all mtls, and if the new endpoint(s) are mtls all good,
otherwise we'll need a full CDS push (
hopefully only to workloads that depend on that cluster only).

If the endpoints added are all mtls, and the existing set of endpoints is
empty or all mtls - but previously
we sent plain text in CDS - we push CDS.

So in normal case is just a bool + checking the label on new endpoints.

rshriram:
> otherwise we'll need a full CDS push (hopefully only to workloads that depend on that cluster only).

And this is where the complexity arises. AS such we are having a tough time handling all the concurrency issues with multiple pushes, along with a plethora of races and hard to debug issues like missing config updates. Throw in feedback loops that selectively update CDS, it will lead to more consistency issues.

costinm:
Agreed that current code is too complex - there is a lot of historic
baggage and 'backward compat'/safety code.

However long term we must be able to send pushes only to sidecars that need
the updates and nobody else.
No other way to scale while keeping resource usage sane.

Might be the right time to try few 'clean room' variants of the XDS -
possibly with some of the new UDP protocol -
I think a 'pubsub'-modeled option is likely to scale the best, planning to
do a prototype soon. In reality each
sidecar needs to 'subscribe' to a typically small number of config objects,
and may be able to do the
conversion to envoy locally if it gets the 'raw' config.

incfly:
Per offline discuss, we move it to mesh config flag and will add support on sidecar injector to annotate workloads on per deployment opt-in.

I confirmed with @rlenglet this PR is okay to proceed at this moment given 1.3 branch has been cut.

PTAL, thanks!

mesh/v1alpha1/config.proto
```diff
@@ -210,6 +210,19 @@ message MeshConfig {
...
+  // If set to true, and a given service does not have a corresponding DestinationRule configured,
+  // or its DestinationRule does not have TLSSettings specified, Istio configures client side
+  // TLS configuration appropriately. More specifically,
+  // If the upstream authentication policy is in STRICT mode, use Istio provisioned certificate
+  // for mutual TLS to connect to upstream.
+  // If upstream service is in plain text mode, use plain text.
+  // If the upstream authentication policy is in PERMISSIVE mode, Istio configures clients to use
+  // mutual TLS when server sides are capable of accepting mutual TLS traffic.
+  // If service DestinationRule exists and has TLSSettings specified, that is always used instead.
+  bool enable_auto_mtls = 43;
```
rshriram: Please use the Bool Wrapper object. And state that by default its true. This way, it wont default to false and we wont have to worry about updating it in 100 places.

incfly: What's the need to use wrapper? 

1. Throughout the meshconfig other fields I don't see any other precedence of using wrapper.
2. Benefits of using wrapper from the doc](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto#L31) is being able to distinguish from not set vs default value. Why and how we'll distinguish these two cases? For this particular feature?

incfly: The feature itself will be default off in `values.yaml` in 1.4 @costinm And we can then migrate to default true maybe in 1.5 with operator facility.

Re proto.wrapper. I did more study now understand that: if a boolean should be default to true, and we need to change the default at proto, then wrapper is needed.

But we can always update the default values from values.yaml helm chart level? Rather than touching the proto definition?

^ @costinm

howardjohn: Seems like everywhere we just control the defaults by helm/DefaultMeshConfig() function. Not sure a wrapper is needed

rshriram: How are you going to set this to default true when the time comes? Ignore the 1.4 timeframe. What if someone forgets to set this feature flag ? It then goes to off? Why does it even need to be explicitly set by the user to true or false always? Go look at linkerd. There is no configuration needed to set such basic stuff.

incfly: Fine with wrapper, although mentioned, the defaults can be toggled via values.yaml chart at that level even for primitive bool value.

@howardjohn after grepping, `DefaultMeshConfig` mostly used for test code to get a reasonable copy. It's true that without wrapper, you can't differentiate false with unset case. And we need to treat unset case to be true in future.

incfly: @rshriram Done, changed to wrapper. PTAL.

mesh/v1alpha1/config.proto
```diff
@@ -210,6 +211,19 @@ message MeshConfig {
   // can be configured for a single control plane.
   repeated ConfigSource config_sources = 22;
 
+  // This flag is used to enable mutual TLS automatically for service to service communication
+  // within the mesh, default false.
+  // If set to true, and a given service does not have a corresponding DestinationRule configured,
+  // or its DestinationRule does not have TLSSettings specified, Istio configures client side
+  // TLS configuration appropriately. More specifically,
+  // If the upstream authentication policy is in STRICT mode, use Istio provisioned certificate
```
diemtvu: We still have the plain-text mode, though it is fairly hidden at the moment (as the default installation set the whole mesh to strict or permissive). Anyway, I think we still need to handle it and mention it here.

incfly: See line 221, it's mentioned already.

PR #1029, opened 2019-08-09
add annotation for protocol sniffing

yxue:
annotation for enable/disable protocol sniffing

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -324,9 +324,10 @@ var (
                         "enabled. This is an alpha API, no backward compatibility "+
                         "guarantee. If enabled, protocol sniffing will be used for "+
                         "each listener that istio cannot parse the protocol from "+
-                        "it's name. The listener will select the filter chain by "+
+                        "its name. The listener will select the filter chain by "+
                         "inspecting the application protocol from the incoming "+
-                        "packet. Sniffing is not supported on H2 and gRPC port.",
+                        "packet. Sniffing is not supported on H2 and gRPC port in "+
+                        "1.3.",
           Hidden: false,
           Deprecated: false,
         }
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -318,8 +318,8 @@ var (
           Deprecated: false,
         }
 	
-		SidecarTrafficEnableProtocolSniffing = Instance {
-          Name: "traffic.sidecar.istio.io/enableProtocolSniffing",
+		AlphaSidecarTrafficEnableProtocolSniffing = Instance {
+          Name: "traffic.sidecar.alpha.istio.io/enableProtocolSniffing",
           Description: "A boolean value indicates that if protocol sniffing is "+
                         "enabled. This is an alpha API, no backward compatibility "+
                         "guarantee. If enabled, protocol sniffing will be used for "+
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -321,7 +321,12 @@ var (
 		SidecarTrafficEnableProtocolSniffing = Instance {
           Name: "traffic.sidecar.istio.io/enableProtocolSniffing",
           Description: "A boolean value indicates that if protocol sniffing is "+
-                        "enabled.",
+                        "enabled. This is an alpha API, no backward compatibility "+
+                        "guarantee. If enabled, protocol sniffing will be used for "+
+                        "each listener that istio cannot parse the protocol from "+
+                        "it's name. The listener will select the filter chain by "+
+                        "inspecting the application protocol from the incoming "+
+                        "packet. Sniffing is not supported on H2 and gRPC port.",
           Hidden: false,
           Deprecated: false,
         }
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -328,7 +328,7 @@ var (
                         "inspecting the application protocol from the incoming "+
                         "packet. Sniffing is not supported on H2 and gRPC port in "+
                         "1.3.",
-          Hidden: false,
+          Hidden: true,
           Deprecated: false,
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -195,6 +195,10 @@ annotations:
       (from VM) will be treated as outbound.
     deprecated: false
     hidden: false
+  - name: traffic.sidecar.istio.io/enableProtocolSniffing
```
nmittler: make this `traffic.sidecar.alpha.istio.io/enableProtocolSniffing`

annotation/annotations.yaml
```diff
@@ -195,6 +195,10 @@ annotations:
       (from VM) will be treated as outbound.
     deprecated: false
     hidden: false
+  - name: traffic.sidecar.istio.io/enableProtocolSniffing 
+    description: A boolean value indicates that if protocol sniffing is enabled.
+    deprecated: false
+    hidden: false
```
nmittler: hidden: true

This means that we don't include it in documentation.

annotation/annotations.yaml
```diff
@@ -195,6 +195,10 @@ annotations:
       (from VM) will be treated as outbound.
     deprecated: false
     hidden: false
+  - name: traffic.sidecar.istio.io/enableProtocolSniffing 
+    description: A boolean value indicates that if protocol sniffing is enabled.
```
nmittler: Move this annotation up with the other alpha annotations at the top.

annotation/annotations.yaml
costinm: I'm not sure - we never did this, and it would create user problems as we move it.

The doc should indicate it is alpha, but it doesn't have to be in the name.

annotation/annotations.yaml
costinm: And please add more description. This is the doc that the user will see, how is a Istio user supposed to know when to set it, what is sniffed, what impact it'll have, etc

annotation/annotations.yaml
costinm: We want to include it in the doc - but the doc should include the "this is an alpha API, no backward compatibility guarantee" or whatever boilerplate we apply to all alpha APIs.

Hidden is for 'internal' annotations that we don't expect users to use or are not yet ready (  WIP - not even alpha )

annotation/annotations.yaml
nmittler: @costinm 

> we never did this

That's not true ... there are several annotations that use `alpha` in the name already.  Just scroll to the top of this file.

annotation/annotations.yaml
nmittler: @costinm 

That seems to be exactly the opposite of our discussions over the past week.  The the docs meeting last week I thought we had agreed that anything alpha would simply not be documented.  Or did this change since then?

costinm:
Sorry for the confusion: of course alpha needs to be documented, almost all
of Istio is alpha :-)

I don't know about the annotations with alpha in names - did we ship any in
1.2 ? Did we document ?
I don't think it's a good idea, happy to raise it with doc/toc - we're
going to spend a lot of effort
and user pain on changing all the alpha CRDs. It doesn't help anyone -
people still use
istio API despite the alpha in the CRD.

What we don't want to document is internal stuff - either dev (not ready
yet) or stuff that is not intended
for users.  This particular feature is very user facing. And I was
concerned about env variables and
CLI flags - APIs and annotations ( which are implicitly part of API ) are
less clear cut.

There is a difference between alpha on the implementation stability - it's
brand new code - and
our intent to support the annotation moving forward.

I think alpha in annotation names is more of an exception, at least most
commonly used annotations
don't use that pattern.

nmittler:
@costinm completely agreed with the distinction of alpha vs dev.  Let's discuss at the next docs meeting and see if we can sort this out.

annotation/annotations.yaml
nmittler: @costinm based on previous discussions, this annotation may not exist for long. It seems more like a dev or internal annotation rather than alpha. Maybe we should hide it for now and then unhide it when we agree on API for protocol sniffing in general (e.g. 1.4?).  WDYT?

costinm: It needs to be visible to user so he can turn it on. 

It is not clear how long it'll exist - we don't have data. We hope to turn it on for all, just like removing the container ports. It is no different - and the container ports are annotations with same pattern.

annotation/annotations.gen.go
```diff
@@ -318,6 +318,20 @@ var (
           Deprecated: false,
         }
 	
+		SidecarTrafficEnableProtocolSniffing = Instance {
+          Name: "traffic.sidecar.istio.io/enableProtocolSniffing",
+          Description: "A boolean value indicates that if protocol sniffing is "+
```
costinm: We may want to revisit the text - indicate the interaction with the values.yaml setting that enables by default ( i.e. annotation overrides the default value ), and indicate that if a ports is labeled explicitly it'll still respect the label ( I think that's the behavior ?).

costinm: But we can do it in separate PR

crazyxy: Yes. The default value will be overridden by the annotation. The default value is false, i.e., protocol sniffing is disabled.

costinm: The default value is given by the installer flag - if pilot setting is to default to on, the annotation can disable it for individual workloads.

This should end up in the generated documentation.

kyessenov:
Should there be a warning about the security implications?
What if I send a malformed HTTP request that tricks the sniffer to think it's TCP while the backend treats it as HTTP, doesn't that escape all the HTTP policies?
E.g. something like https://portswigger.net/blog/http-desync-attacks-request-smuggling-reborn

PR #1030, opened 2019-08-09
Update CODEOWNERS

rshriram:


CODEOWNERS
```diff
@@ -1 +1,15 @@
...
+mesh/* @costinm @mandarjog
+annotation/* @costinm @mandarjog
+
+mixer/* @kyessenov @mandarjog @geeknoid
+
+authentication/* @diemtvu @wattli
+security/* @diemtvu @wattli
+rbac/* @diemtvu @wattli
+
+mcp/* @ozevren
```
howardjohn: this is impossible to merge.. We require 2 approvals and only have 1 owner

CODEOWNERS don't have inheritance like owners did

duderino: how about we add @andraxylia as co-owner for mcp?

CODEOWNERS
```diff
@@ -1 +1,15 @@
 *   @geeknoid, @rshriram, @andraxylia, @costinm, @diemtvu, @frankbu, @GregHanson, @mandarjog, @ozevren, @wattli
+# @frankbu added as he frequenlty edits comments/docs
+
+networking/* @rshriram @louiscryan @frankbu
```
duderino: I think @costinm needs to be on this list.  This is the second PR attempting to remove him without justification

CODEOWNERS
```diff
@@ -1 +1,15 @@
...
+
+mesh/* @costinm @mandarjog
+annotation/* @costinm @mandarjog
+
+mixer/* @kyessenov @mandarjog @geeknoid
+
+authentication/* @diemtvu @wattli
+security/* @diemtvu @wattli
+rbac/* @diemtvu @wattli
```
duderino: wattli has left the project.  I would recommend @liminw and @myidpt for these

costinm:
I still believe API should be approved by people from TOC and WG leads -
and have cross WG review, to ensure consistency.

howardjohn:
I think one of the problems is what Shriram mentioned in the TOC meeting
about owner granularity. We don't every wg lead and TOC member to approve
some fix for s typo or something, but it makes sense for a large API
change. But there isn't any (easy) way to enforce this with owners. If
there is a good way to enforce this that would be ideal, but not sure how
that would eork. It seems that we should have owners that will recognize
when a change has large implications and allow others to review as well,
but even if this happened not everyone may understand the impact a change
may have on other components.

CODEOWNERS
rshriram: They are already there in the top level list. I am not removing anyone in this PR. All the lines are additional specializations on top of the `*` thing.

howardjohn: That is not how codeowners works. The last matching row is used there is no merging. Additionally, we require 2 approvals and you can approve your own pr (maybe we can change this - I'll look into it). So what this means is if you submit a PR to network you need both Louis and frank to approve it. All of the ones with 2 owners are impossible to be approved if submitted by the owners, or require both if submitted by non owners

howardjohn: that is why there is all the copy and paste in the Istio file

CODEOWNERS
ozevren: @geeknoid can be added here.
Also, @ayj  is a good candidate, if he can be added.

PR #1031, opened 2019-08-09
remove the out-dated authorization policy

yangminzhu:
Now we have the new approved `AuthorizationPolicy` in the `security` directory, it's best to remove the out-dated one to avoid confusions.

The corresponding code had been removed in https://github.com/istio/istio/pull/16119.

istio/istio#12394

### DIFF
--- a/proto.lock
+++ b/proto.lock
@@ -6853,35 +6853,6 @@
         ],
         "messages": [
           {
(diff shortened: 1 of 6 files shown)


yangminzhu:
@rshriram @costinm Could you take a look? Thanks.

PR #1032, opened 2019-08-09
Add WorkloadSelector definition in a common directory.

liminw:


Makefile.core.mk
```diff
@@ -71,12 +71,14 @@ protoc_gen_python_plugin := $(protoc_gen_python_prefix):$(repo_dir)/$(python_out
...
 protoc_gen_docs_plugin_for_networking := --docs_out=warnings=true,dictionary=$(repo_dir)/dictionaries/en-US,custom_word_list=$(repo_dir)/dictionaries/custom.txt,per_file=true,mode=html_fragment_with_front_matter:$(repo_dir)/
 protoc_gen_docs_plugin_for_security := --docs_out=warnings=true,dictionary=$(repo_dir)/dictionaries/en-US,custom_word_list=$(repo_dir)/dictionaries/custom.txt,per_file=true,mode=html_fragment_with_front_matter:$(repo_dir)/
+protoc_gen_docs_plugin_for_type := --docs_out=warnings=true,dictionary=$(repo_dir)/dictionaries/en-US,custom_word_list=$(repo_dir)/dictionaries/custom.txt,per_file=true,mode=html_fragment_with_front_matter:$(repo_dir)/
 
 #####################
 # Generation Rules
 #####################
 
 generate: \
+    generate-type \
```
yangminzhu: I think this needs to be tab instead of whitespaces.

type/v1beta1/selector.proto
```diff
@@ -0,0 +1,36 @@
...
+// $hide_from_docs
+// $title: Workload Selector
+// $description: Definition of a workload selector.
+// $location: https://istio.io/docs/reference/config/type/v1beta1/workload-selector.html
+
+package istio.type.v1beta1;
+
+option go_package="istio.io/api/type/v1beta1";
+
+// WorkloadSelector specifies the criteria used to determine if a policy can be applied
```
yangminzhu: If you want to hide the doc, I think you also need $hide_from_docs here.

cue.yaml
```diff
@@ -7,6 +7,8 @@ openapi:
   fieldFilter: "min.*|max.*"
 
 directories:
+  type/v1beta1:
+    - mode: all
```
yangminzhu: I think it's better to use `perFile`

liminw: Done.

type/v1beta1/selector.proto
liminw: Done.

Makefile.core.mk
liminw: Done.

costinm:
Do we have a plan to migrate the other selectors to this ? Is it identical
with any of the in-use selectors ? Did the TOC agree to deprecate the
3 selector APIs and switch to this one ( or to have 4 selector methods ) ?

I would really like to see an official TOC review and decision on the
deprecation or migration to new selector.

liminw:
@costinm I think there is consensus we need a common "selector" definition. This PR just added one following [Istio Config Model](https://docs.google.com/document/d/1T22PxxLwwFnxC_yFQ_HmuKsuZgFjRcDy2BRb5enfYpk). 
The existing selector definitions are already different. 

In sidecar
```
workloadSelector:
   labels:
      app: A
```

In Gateway,
```
selector:
   app: A
```
As @rshriram [commented in this PR](https://github.com/istio/api/pull/918#issuecomment-519764790), if we choose consistency, some form of deprecation cannot be avoided. 

I don't care too much about the exact naming, as long as there is consensus. Since we are introducing beta API now, I think it is important to make sure API stability and consistency.  

@louiscryan Any recommendation?

type/v1beta1/selector.proto
rshriram: Can you also update the EnvoyFilter to use this workload selector [and update the comments in envoyfilter accordingly] ? The workloadSelector there has not been released yet. 
There is a workloadSelector in the Sidecar.proto.. Please remove it and make the sidecar refer to this. Since the sidecar uses workloadSelector.labels, i suggest you add a labels field in this proto and mark it as deprecated. 
In the implementation, its as simple as copying the value from the existing labels field into the new matchLabels field.

type/v1beta1/selector.proto
```diff
@@ -0,0 +1,37 @@
...
+// $hide_from_docs
+// $title: Workload Selector
+// $description: Definition of a workload selector.
+// $location: https://istio.io/docs/reference/config/type/v1beta1/workload-selector.html
+
+package istio.type.v1beta1;
+
+option go_package="istio.io/api/type/v1beta1";
+
+// $hide_from_docs
```
rshriram: no hide from docs if this is going to be used by Sidecar/envoyfilter etc.

liminw: Done.

type/v1beta1/selector.proto
liminw: I will do EnvoyFilter/Sidecar proto change in a separate PR.

PR #1033, opened 2019-08-13
Protocol Detection Timeout

rshriram:
This setting controls how long Envoy will wait before
timing out on TLS sniffing/ HTTP sniffing. This is very similar to 
the connection timeout feature that we have. Will default to 10ms
in the charts.


mesh/v1alpha1/config.proto
```diff
@@ -74,6 +74,17 @@ message MeshConfig {
   // Connection timeout used by Envoy. (MUST BE >=1ms)
   google.protobuf.Duration connect_timeout = 6;
 
+  // Automatic protocol detection uses a set of heuristics to
+  // determine whether the connection is using TLS or not (on the
+  // server side), as well as the application protocol being used
+  // (e.g., http vs tcp). These heuristics rely on the client sending
+  // the first bits of data. For server first protocols like MySQL,
```
howardjohn: aren't you basically just adding a `x ms` overhead on each request, since server first protocols will always hit the timeout?

rshriram: For each connection and not each request in the connection. And yes, all server first protocols will incur this additional overhead. And matter of fact, they already pay this price today with TLS sniffing. Though today, the price is pretty bad - stuck forever. The timeout approach bounds the waiting time to say 10ms at most, maintaining a decent balance. If someone is really concerned about performance, they should properly name the ports - which automatically eliminates sniffing.

howardjohn: Good point. The only possible issue is if it doesn't work at all it's obvious, I'd it's just slower you may not notice.

I do think this is a good idea, just one that we need to make well documented

rshriram:
>This makes the protocol detection trivial to bypass, since you can just wait x ms before sending your request. Thats ok if and only if we have very very explicit documentation not to use this in production

I dont think this is an attack vector. The real attack vector is the whole idea of TLS sniffing. Our docs should be saying that in a production system, one should always have strict MTLS and should always name all the ports properly to avoid any sniffing related issues. 

Things like sniffing are useful for the new comers, people trying out Istio on some existing dev/staging environments. Never an option for something going to production

mesh/v1alpha1/config.proto
rshriram: I agree. @crazyxy 's next task is to write lots of FAQ entries in istio.io as well as operational notes, on the sniffing piece explaining the various behaviors.

PR #1037, opened 2019-08-15
typo fix rbac/v1alpha1/rbac.pb.go

zhranklin:
Typo fix for `*`s not displayed correctly because they are treated as italic markers.


zhranklin:
@geeknoid should this pr be accepted? Another approving review is needed.

PR #1040, opened 2019-08-15
add the mtlsReady label for istio autopilot

GregHanson:
add the mtlsReady annotation to enabled mtls autopilot

related issue:
https://github.com/istio/istio/issues/14524

related design doc:
https://docs.google.com/document/d/1yEMDRO2FZCyZnDK1AzNmjQtbtQqE7wN8YoM65FAH7uA/edit#heading=h.74liym2vo8bc

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -46,6 +46,15 @@ var (
           Deprecated: false,
         }
 	
+		AuthenticationMtlsReady = Instance {
+          Name: "authentication.istio.io/mtlsReady",
+          Description: "Label injected by either istioctl or webhook which "+
+                        "specifies whether a service instance is ready to accept "+
+                        "mTLS traffic.",
+          Hidden: false,
+          Deprecated: false,
+        }
+	
 		IoKubernetesIngressClass = Instance {
           Name: "kubernetes.io/ingress.class",
           Description: "Annotation on an Ingress resources denoting the class of "+
--- a/annotation/annotations.pb.html
+++ b/annotation/annotations.pb.html
@@ -28,6 +28,15 @@ Istio supports to control its behavior.
 				
 					<tr>
 				
+					<td><code>authentication.istio.io/mtlsReady</code></td>
+					<td>Label injected by either istioctl or webhook which specifies whether a service instance is ready to accept mTLS traffic.</td>
+				</tr>
+			
+		
+			
+				
+					<tr>
+				
 					<td><code>kubernetes.io/ingress.class</code></td>
 					<td>Annotation on an Ingress resources denoting the class of controllers responsible for it.</td>
 				</tr>
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -223,3 +223,8 @@ annotations:
     description: Annotation on an Ingress resources denoting the class of controllers responsible for it.
     deprecated: false
     hidden: false
+  - name: authentication.istio.io/mtlsReady
+    description: Label injected by either istioctl or webhook which specifies whether a
+      service instance is ready to accept mTLS traffic.
+    deprecated: false
+    hidden: false


annotation/annotations.gen.go
```diff
@@ -46,6 +46,15 @@ var (
           Deprecated: false,
         }
 	
+		AuthenticationMtlsReady = Instance {
+          Name: "authentication.istio.io/mtlsReady",
+          Description: "Label injected by either istioctl or webhook which "+
```
rshriram: If this is a label, then this should not be an annotation. And I would suggest using the label instead of annotation to maintain cross platform compatibility.

rshriram: And if this is a label, then you wont need this PR at all.

GregHanson: Agree, description wording is incorrect.  However @costinm was against labels because they are costly in kubernetes.  I see even the design docs goes back and forth between calling it a label vs annotation

howardjohn: Shriram my understanding is we dump the annotations and labels into NodeMetadata, so is there a difference between annotation and label for non-k8s?

incfly: This should be a annotations.

IIRC, label is used for k8s apiserver itself, annotations are used for 3rd party application, like istio.

incfly: @diemtvu I think we should call it "security.istio.io" rather than "authentication.istio.io", given the possibility of moving away from authentication policy?

linsun: Do you have some performance data on label vs annotation we can use to make decisions?

GregHanson: cc @costinm ^^

costinm:
Confused here: is it a label or an annotation ? I thought it is a label,
happy with annotation too - but not both.

costinm:
On istio side, there should be no difference in performance between labels
and annotations - in pilot or injector.

On K8S apiserver: labels are indexed, while annotations are not. I haven't
seen any perf test on Apiserers - but we did
and do have problems when apiserver is slow. I would really like to see
some numbers - but won't hold this PR
for it.

On non-K8S: I don't think annotations are used, so it'll have to be a label
( in consul for example ). Also in the
ServiceEntry for mesh expansion.

The bootstrap currently mixes annotations and labels - there are proposals
to separate them, but I'm not sure
how much K8S-specific semantic should leak into Istio node metadata.

Only labels can be used in selectors - which is the likely reason for
Apiserver to index them.

We generally use annotations because most of the metadata we need doesn't
need to be indexed and
used for selection.

For mtlsReady: I think it could be a label, since it's used is consistent
with indexing. But happy with annotation too.

annotation/annotations.gen.go
diemtvu: > @diemtvu I think we should call it "security.istio.io" rather than "authentication.istio.io", given the possibility of moving away from authentication policy?

SGTM.

rshriram: This code needs to be cross platform and not a kubernetes only hack. so, no annotations please. Especially for a core functionality like this. How do you represent this annotation in a service entry?

incfly: For ServiceEntry, I think we can make `security.istio.io/mtlsReady` as label key, "true" as value in ServiceEntry here.
https://github.com/istio/api/blob/master/networking/v1alpha3/service_entry.proto#L449

That should work, I assume?

I remember k8s annotations ultimately are transformed into the entry in `IstioEndpoint.Labels` entries. Istio data structure make it encapsulate the different platforms' implementations. I admitted the overloading `label` can cause some confusion, but that does not seems to be an issue to support different platform.

I'm fine with label as well, just feel not needed.

costinm:
Agree, it's not authentication specific - mtls may be used even when auth
is not the user intention. The behavior is to initiate tls connection
instead of plain text.

Do labels support full names or just single words ?

annotation/annotations.gen.go
howardjohn: can we confirm that you can set the annotation outside if k8s? if so I think annotation is much more appropriate, if not let's make it a label?

incfly:
Per offline discussion, we should use labels, since annotations are not converted as label on each endpoint within Pilot.

We can use `security.istio.io/mtlsReady": "true"/"false"

Given that, we don't need this PR maybe? Or are we keep track of all the labels we support somewhere in this repo as wel?

GregHanson:
@incfly agreed that this PR is no longer needed, closing this PR and refactoring https://github.com/istio/istio/pull/16296 for labels instead of annotations

>Or are we keep track of all the labels we support somewhere in this repo as wel?

I don't believe we are tracking istio labels in istio/api or a common package in istio/istio, but I will open an issue since this should probably be addressed.  I will not try to address this in https://github.com/istio/istio/pull/16296 in order to keep that PR as small and direct as possible

GregHanson:
opened https://github.com/istio/istio/issues/16439 issue for discussion on whether we need labels tracked in a common package same as we do for annotations

incfly:
Thanks! SGTM

PR #1042, opened 2019-08-15
Add feature phase enum.

nmittler:
This will be used in the annotation structure in a follow-on PR. The content of this is taken from https://istio.io/about/feature-stages/.

### DIFF
--- a/feature/phase.go
+++ b/feature/phase.go
@@ -1,0 +1,76 @@
+// Copyright 2019 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


geeknoid:
@nmittler Do we still want this?

nmittler:
@geeknoid sorry sort of lost track of this. That's a good question ... has something change since we decided to do this? I assume we still want to document the support level for annotations, no?

geeknoid:
Yeah, I still think this is useful. I guess I don't understand how you expect this go file to be used? I would have imagined extending https://github.com/istio/tools/tree/master/cmd/annotations_prep to support these phases, so that they could be specified in https://github.com/istio/api/blob/master/annotation/annotations.yaml.

PR #1053, opened 2019-08-22
Modify to a half-width quotation mark to fix the garbled document

AsCat:
In some auto generated docs like html, it will appear garbled when using full-width quotation mark.
It seems to be some typing error, i just fixed them all.
This PR should better to be reviewed. Thanks.


geeknoid:
This PR is experiencing an unrelated CI problem, probably related to the protoc version in our build-tools container. We're investigating.

AsCat:
It seems like this PR need at least two reviewers to approve it but we have only one, can @howardjohn or someone else be added to the reviewers?

PR #1056, opened 2019-08-26
add stat name pattern

ramaraochavali:
Adds support for specifying stat name pattern in Destination Rule which will be used while emitting stats for the related clusters.


networking/v1alpha3/destination_rule.proto
```diff
@@ -140,6 +140,36 @@ message DestinationRule {
...
+  // %SERVICE_PORT% - Will be substituted with port of the service.
+  // %SERVICE_PORT_NAME% - Will be substituted with port name of the service.
+  //  
+  // Following are some examples of supported patterns for reviews.
+  // %SERVICE_FQDN%_%SERVICE_PORT% will use reviews.prod.svc.cluster.local_7443 as the stats name.
+  // %SERVICE% will use reviews.prod as the stats name.
+  string inbound_stat_name = 5;
+
+  // Name to be used while emitting statistics for outbound traffic. 
+  // By default, Istio emits statistics with the pattern outbound|<port>|<port-name>|<service-FQDN>.
```
mandarjog: the default is
`outbound|port|subsetname|hostname` for example `outbound|8080|v2|svc01-0.service-graph01.svc.cluster.local`
We need to add subset name in the substitution.

ramaraochavali: Yeah. Good catch. Added subset name and corrected the default pattern. PTAL.

ramaraochavali:
@howardjohn can you PTAL and merge this? I will follow-up with Pilot side of implementation

ramaraochavali:
@rshriram sure. MeshConfig works for us as well. But was wondering with destination rule, we get lot more flexibility. But If it is going to add more confusion than flexibility, I will move it to MeshConfig

PR #1065, opened 2019-08-28
Use protos from common-files repository

sdake:
Depends-On: https://github.com/istio/common-files/pull/42


geeknoid:
Subsumed by #1075

PR #1067, opened 2019-09-03
add tls settings for pilot and tracer config

ramaraochavali:
Adds support for specifying TLS settings for Pilot and tracer.  Implements https://github.com/istio/istio/issues/13221. This was also discussed here https://github.com/istio/api/pull/1023#issuecomment-519126493


ramaraochavali:
@rshriram @howardjohn can you please take look when you get chance? I have started making pilot side changes.

mesh/v1alpha1/proxy.proto
```diff
@@ -173,6 +182,7 @@ message ProxyConfig {
   // Authentication policy defines the global switch to control authentication
   // for Envoy-to-Envoy communication for istio components Mixer and Pilot.
   AuthenticationPolicy control_plane_auth_policy = 13;
```
rshriram: what happens to this then? it is largely obsolete isnt it?

ramaraochavali: Yes. good catch.  Should I deprecate that as well?

ramaraochavali: It seems it may be still be used by mixer atleast. What should I do? Should I change the doc to indicate it is used by mixer?

ramaraochavali: updated the doc. Please take look or if you have any better suggestions, let me know - will change.

ramaraochavali:
@rshriram can you PTAL when you get chance?

rshriram:
Will get another person to approve. Please start with Pilot impl

rshriram:
? All he is doing is providing a way to specify the paths to the control plane TLS certificates (which are hard coded currently to wherever we set certs). There is no upgrade issue because nothing is being removed. 

Matter of fact, its extremely inconsistent before the PR:  we allow customizable TLS certs for the data plane communication but not for control plane, even though both use istio mTLS. 

The alternative option is to add 3 more items of config that specify where the mTLS certs to talk to a Pilot are present. On VM platforms, we cannot expect to force people to stick the certificates on /etc/certs/..

costinm:
It's replacing 'discoveryAddress' - which is what injector and everything is using. 

Using RemoteService instead of DestinationRule to configure client TLS settings is also important from a design perspective because we want to control canary / version split for control plane components as well. 

I don't think DR is perfect - but if we want to replace it I think it deserve a design doc. 
And deprecating/replacing widely used settings - like discoveryAdress - deserves a design doc and proper review of the migration process and compat, UX, etc.

costinm:
And about VMs - this is obviously not the best design because we can't expect certificates to be uniform across the mesh. A global setting that has the certs set to the same path is not going to work well. 
We do have a mechanism to specify a base dir, and we have SDS - again a proper design on how certs will be distributed to VMs and in cases where the file system layout is not consistent would be better.

The 'alternative' is to have a clear design on how certs will be handled on VMs - SDS or detecting
the location from env - and a design on how control plane components can be configured so we can also handle traffic split and be consistent with the rest of the networking APIs. It's tricky - I don't claim 
to have a good solution - so the request for a design doc that explains the plans and options.

ramaraochavali:
I am not sure how DR will work with Pilot cluster for two reasons
- Since the control plane cluster is statically defined in the bootstrap, Envoy does not allow updating it via CDS (I think DR updates has to go via CDS Api)
- The initial connection has to be established to get the DR pushed to Envoy which requires certs.

Am I missing some thing here or are you thinking of some other way we can achieve with DR?

costinm:
An option is to move bootstrap generation to pilot. Would not help with
pilot own address, but this is rarely configured. For Tls to pilot we can
use apiserver-signed certs, avoiding the pubkey problem.

Current code handling bootstrap is too complex, in large part due to not
having access to real config, and mangling it via injector.

rshriram:
> An option is to move bootstrap generation to pilot. Would not help with
pilot own address, but this is rarely configured. 

moving bootstrap to pilot wont help because this is for the sidecar to start. We wont have push contexts or any other thing setup. IF we start hitting the service registry directly (env.List) for every sidecar, then it increases load on pilot. All of this is very unnecessary in my opinion.

just because something has TLS does not mean it has to be destination rule. We have TLS settings for galley sources [been there for 2 releases]. So, the remoteServices is simply abstracting this away nicely and not polluting the mesh config with 100s of fields. 

And finally, discovery_address is "deprecated". It is not `removed`. We can fix the installer, and helm charts so that they are all in sync. We have so many of these kinds of changes happening.

costinm:
On Wed, Sep 11, 2019 at 7:08 AM Shriram Rajagopalan <

> An option is to move bootstrap generation to pilot. Would not help with
> pilot own address, but this is rarely configured.
>
> moving bootstrap to pilot wont help because this is for the sidecar to
> start. We wont have push contexts or any other thing setup. IF we start
> hitting the service registry directly (env.List) for every sidecar, then it
> increases load on pilot. All of this is very unnecessary in my opinion.
>

That's why we should have a design doc, some prototype to get data on the
impact. Agree env.List would be to expensive, but there are ways to do it
with O(1) and very cheap operations.

And there are benefits - the current inject using global settings + env
variables mangling + template is not ideal either.
Having pilot generate the bootstrap might work better for mesh expansion,
may allow per-namespace settings and more control.

Or maybe not - the point is that we should have a proper design on how to
improve the bootstrap configuration ( and that covers
not only the remote services used in bootstrap, but the bigger picture )..




> just because something has TLS does not mean it has to be destination
> rule. We have TLS settings for galley sources [been there for 2 releases].
> So, the remoteServices is simply abstracting this away nicely and not
> polluting the mesh config with 100s of fields.
>

Agreed, and also believe current DestinationRule + ServiceEntry are far
from perfect.

That's why it's important to have a design doc and have consensus on how we
want to represent the settings of a client connecting to a server ( and
result in an envoy cluster settings). Maybe RemoteService is the right
solution - but it should be a documented decision, with a plan to migrate
to it
and maybe a plan to use it as a replacement for DR if it is clearly better.

The goal is to simplify Istio config - and having 3 different ways to
express TLS settings I don't think we're improving.




> And finally, discovery_address is "deprecated". It is not removed. We can
> fix the installer, and helm charts so that they are all in sync. We have so
> many of these kinds of changes happening.
>

And the plan on how deprecation will be handled, timing, when will it be
removed, how it impacts operator, etc need to be captured and discussed.
Deprecation is not a trivial thing to do.

We do have many of these changes happening - without sufficient review and
'big picture' thinking, and it is a big cause for the current install and
operation
complexity.


Costin

rshriram:
I agree that we need a separate effort to move bootstrap into Pilot but isn't that independent of this PR? I feel that this is simply following the conventions that we already have in code (for access log, metrics, tracing, galley, etc.)

PR #1070, opened 2019-09-05
fix the sidecar egress order in the example

linsun:
The sidecar config caused the following error when I applied it.
	* sidecar: the egress listener with empty port should be the last listener in the list


linsun:
would like to merge to 1.2 and 1.3 doc as well.

PR #1073, opened 2019-09-06
Use realpath instead of readlink

jwendell:
Output of readlink on linux is something like
```
$ readlink /etc/localtime
../usr/share/zoneinfo/America/New_York
```

Which breaks the `-v` docker cli option

realpath works fine:
```
$ realpath /etc/localtime
/usr/share/zoneinfo/America/New_York
```


jwendell:
> I think this file comes from istio/common-files

Thanks! Istio is really getting complicated :/
I'll close this one and create another.

PR #1063, opened 2019-08-28
add stat name patterns to mesh config

ramaraochavali:
Adds support for specifying stat name pattern in MeshConfig which will be used while emitting stats for the related clusters.

PR #1074, opened 2019-09-06
[cherry-pick-1.3] add stat name patterns to mesh config

ramaraochavali:
Cherry pick of #1063 . This is minor api change with no risk.


ramaraochavali:
@howardjohn  PTAL. Merge conflicts with auto cherry pick. So created a manual PR.

ramaraochavali:
@rlenglet This is a very minor change with no risk. We (Salesforce) are deploying Istio in production and this is critical to get the service migration from our in-house control plane to Istio. Can we merge this API change to `1.3` and implementation to a patch of `1.3.1` as I understand API changes to patches may not be allowed? Or if you have any better suggestions to land this before `1.4` as it would be too late also is fine. Please suggest.

rshriram:
@rlenglet this change is harmless IMO as its purely optional and enabled if and only if user specifies this value. So, the risk is limited to just this user/company and not to others.

So from networking point of view, LGTM as a one time exception. I am only approving this for this time so that the salesforce folks can get acclimated to our release processes.
@ramaraochavali for future reference, we ask everyone to diligently cherrypick changes that they need well before the release date, especially the API changes.

ramaraochavali:
@rshriram Thank you. Sure will follow the suggested process in future.

PR #1076, opened 2019-09-07
refine NetworkEndpoints fromRegistry description

styblope:
Description in Network.NetworkEndpoints field fromRegistry is misleading by guiding to use a Kubernetes secret name rather than a kubeconfig filename used as a data key inside the secret.

[x] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

Make-up for #4880

### DIFF
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
@@ -850,7 +850,7 @@ inside a mesh and how to route to endpoints in each network. For example</p>
 <pre><code class="language-yaml">networks:
   network1:
   - endpoints:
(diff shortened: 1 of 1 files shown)


geeknoid:
Getting closer :-).

The change actually needs to go in the .proto file instead. Then you run "make" which generates the .html file. You then check in all the generated files along with your .proto change. Once it's all checked in, then we have a process to collect those files from this repo and insert them into the istio.io repo where they get published on the web site.

Sorry for all the hoops just for a typo...

styblope:
Fixed in #1078. Hope this one will do it. Thanks for your patience and advise!

PR #1077, opened 2019-09-08
Enable all linters on this repo.

geeknoid:
- Fix a few linting errors.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,9 +7,9 @@ about: Report a bug to help us improve Istio
   To ask questions about how to use Istio, please visit <https://discuss.istio.io>
 )
 
-**Bug description**
+<em>Bug description</em>
```
howardjohn: This doesn't look right? Why change this?

geeknoid: The markdown linter complains about "using emphasis instead of a header". So then I switched from the "**" notation to using "#####" instead. Then the linter complained that I should be using "#" instead of "#####" which makes the titles too big.

In theory, `<em>` should produce the same result as the original "**" did in the final rendered  issue page. It doesn't look the same while editing the issue though, that's true.

geeknoid: I reverted that change and instead added .github to the list of directories we don't look into for linting.

PR #1079, opened 2019-09-09
Add mirror_percent and deprecate mirror in HTTPRoute

jwendell:
This allows us to mirror only a percentage of the traffic, instead
of always mirroring 100%.
    
The only difference between the two structs is the `weight` field,
which will be used in pilot side to implement the percentage mirroring.
    
This partially fixes https://github.com/istio/istio/issues/14437, in
conjuction with https://github.com/istio/istio/pull/16939.


jwendell:
This is going to be used on https://github.com/istio/istio/pull/16939

rshriram:
What does mirror2 mean?  specifically to the end user?

jwendell:
@rshriram yeah, that's why it's marked as `wip`. I'm not sure if just adding another mirror field (and calling it `mirror2`) is the best approach.

The idea is to enhance the `mirror` functionality with a `weight` field (see https://github.com/istio/istio/issues/14437 and https://github.com/istio/istio/pull/16939).

We can't just change the current `Mirror` field as that would break API, correct? So I thought of adding a new field instead, and deprecate the current one. Agreed that `Mirror2` is a terrible name. Any other suggestion?

Thanks.

networking/v1alpha3/virtual_service.proto
```diff
@@ -400,13 +400,16 @@ message HTTPRoute {
...
+  Destination mirror = 9;
+
   // Mirror HTTP traffic to a another destination in addition to forwarding
   // the requests to the intended destination. Mirrored traffic is on a
   // best effort basis where the sidecar/gateway will not wait for the
   // mirrored cluster to respond before returning the response from the
   // original destination.  Statistics will be generated for the mirrored
   // destination.
-  Destination mirror = 9;
+  RouteDestination mirror2 = 18;
```
venilnoronha: Name this as `mirror_percent` or `mirror_fraction`?

jwendell: changed it to `mirror_percent`

networking/v1alpha3/virtual_service.proto
```diff
@@ -400,13 +400,18 @@ message HTTPRoute {
...
   // the requests to the intended destination. Mirrored traffic is on a
   // best effort basis where the sidecar/gateway will not wait for the
   // mirrored cluster to respond before returning the response from the
   // original destination.  Statistics will be generated for the mirrored
-  // destination.
-  Destination mirror = 9;
+  // destination. You can mirror only a percentage of the traffic using the
+  // `weight` field. If weight has the zero value, mirroring is effectively
+  // disabled.
+  RouteDestination mirror_percent = 18;
```
rshriram: This is not what I meant. There is no need to deprecate that field. Just add the mirror_percent as an UintValue

networking/v1alpha3/virtual_service.proto
```diff
@@ -408,6 +408,12 @@ message HTTPRoute {
   // destination.
   Destination mirror = 9;
 
+  // Percentage of the traffic to be mirrored by the `mirror` field.
+  // If this field has the value of zero, all the traffic (100%)
+  // will be mirrored. In order to disable mirroring, the `mirror`
+  // field should be removed. Max value is 100.
+  uint32 mirror_percent = 18;
```
rshriram: google.protobuf.UintVal or whatever that wrapper object is. Otherwise, you will not be able to differentiate between a mirror percent 0 vs missing mirror percent field.

jwendell: For compatibility, 0 (or absence) should be handled as 100%, agree?

jwendell: nah, absence makes sense, 0 no.

PR #1083, opened 2019-09-16
Add synthetic serviceEntry option to config source

Nino-K:
This option allows to distinguish if a given address is responsible for MCP config or Synthetic service entries only. It is related to PR https://github.com/istio/istio/pull/12276


mesh/v1alpha1/config.proto
```diff
@@ -379,6 +379,11 @@ message ConfigSource {
   // uses Istio mutual TLS and shares the root CA with Pilot, specify the TLS
   // mode as `ISTIO_MUTUAL`.
   istio.networking.v1alpha3.TLSSettings tls_settings = 2;
+
+ // This flag is to indicate that the address in only forwarding synthetic service entries.
+ // Use this option if the config source server is configured to only forward platform generated
+ // service entries.
+  bool synthetic_service_entries_only = 3;
```
rshriram: how about `contains_service_registry_info` ?

costinm: Agree, synthetic_service_entry is an impementation detail. 

discovery_only ? registry_only ?

costinm: Alternative: repeated list of types to watch ?

costinm:
To be clear: I don't know if 'only synthetic' is what we want long term. 
We might want to eventually watch other things from the 'secondary' clusters, so an API that allows configuring the list of watched resources might be more future-proof.

mesh/v1alpha1/config.proto
```diff
@@ -379,6 +379,11 @@ message ConfigSource {
   // uses Istio mutual TLS and shares the root CA with Pilot, specify the TLS
   // mode as `ISTIO_MUTUAL`.
   istio.networking.v1alpha3.TLSSettings tls_settings = 2;
+
+ // This flag is to indicate that the address in only forwarding synthetic service entries.
```
hzxuzhonghu: `the address in only forwarding ` ?

mesh/v1alpha1/config.proto
```diff
@@ -379,6 +379,11 @@ message ConfigSource {
...
+ // Use this option if the config source server is configured to only forward platform generated
```
hzxuzhonghu: s/Use/Set ?

hzxuzhonghu: Does ` platform generated ` ones mean those generated from services and endpoints in k8s?

mesh/v1alpha1/config.proto
hzxuzhonghu: I am not sure what's this field actually mean when this flag is true? 

1. send only  synthetic_service_entry not native service entries?

2. send both synthetic_service_entry and native service entries?

Nino-K: Synthetic Service entries are the ones that are generated by the platform (k8s) the ones are generated by the users are just synthetic service entires (main difference is the collection type). So when the flags is set it indicates that only synthetic service entries are retrieved from this configsource. I totally agree with not exposing "synthetic service entry" via API as it is just an implementation details.

mesh/v1alpha1/config.proto
Nino-K: correct

mesh/v1alpha1/config.proto
```diff
@@ -379,6 +379,11 @@ message ConfigSource {
   // uses Istio mutual TLS and shares the root CA with Pilot, specify the TLS
   // mode as `ISTIO_MUTUAL`.
   istio.networking.v1alpha3.TLSSettings tls_settings = 2;
+
+  // This flag is to indicate that the address is only forwarding service entries that are generated
+  // by the platform. These auto generated service entries are combination of services and endpoints
+  // that are generated by a specific platform e.g. k8
+   bool service_registry_only = 3;
```
rshriram: per @costinm 's  comment, I think you can expand this a bit more.. Something like this:
resources:
- KUBERNETES_REGISTRY
- CONSUL_REGISTRY
- SERVICE_ENTRIES
- VIRTUAL_SERVICES
- ...

So, you could define an enum like
```
enum resource {
KUBERNETES_REGISTRY = 1,
CONSUL_REGISTRY = 2,
}
```
such that in the ConfigSource you can do
```
repeated subscribed_resources = 3
```

we dont have to enumerate all the resources in the enum - we can add them over time. For now, we simply need to default to all if not specified.

Nino-K: @rshriram in the context of MCP we can have galley processing configs in two modes, one is to forward all sort of configs the other just to forward `Synthetic service entries only`. How do you distinguish the two modes? are you suggesting in your example above that `KUBERNETES_REGISTRY` forwards all configs and `SERVICE_ENTRIES` are only forwarding `syntheti_service_entries`?

Nino-K:
@costinm @rshriram PTAL to see if it captures your suggestions.

rshriram:
need another person to rubber stamp

PR #1085, opened 2019-09-17
feat(mixerclient): add field to ignore forwarded attributes

douglas-reid:
This PR provides a way to configure mixerclient to drop/ignore attributes received as part of a request.

Motivations:
- ingress gateway should not accept forwarded attributes. there are several open issues related to this, including traffic from inside a mesh going through the gateway (duplicated source info)
- egress gateway should ignore forwarded attributes. it acts as a client, leading to incorrect source and destination information (and overrides the correct attributes provided in route configuration).


mixer/v1/config/client/client_config.proto
```diff
@@ -181,6 +181,12 @@ message HttpClientConfig {
   // Default attributes to forward to upstream. This typically
   // includes the "source.ip" and "source.uid" attributes.
   Attributes forward_attributes = 5;
+
+  // Whether or not to use attributes forwarded in the request headers to 
+  // create the attribute bag to send to mixer. For intra-mesh traffic,
+  // this should be set to "false". For ingress/egress gateways, this
+  // should be set to "true".
+  bool ignore_forwarded_attributes = 6;
```
kyessenov: There is a long list of precedence for attributes above (see comment for forwarded_attributes). Can you clarify the impact on it there, too?
By the way, why not just skip forwarded attributes for traffic to egress?

douglas-reid: we could skip the forwarded attributes just for egress, but that ignores the need to do this also on ingress.

That precedence list does not seem to be working as described, based on observations of making the same request from an injected pod and an uninjected pod. It looks like route config is not given a higher priority than forwarded config attributes.

But using *any* attributes from the client in this case, while also acting as a client, seems counter-intuitive (and in the case of ingress, potentially wrong). Feedback has been that we want to correctly attribute actions of the gateways to the gateways themselves. Any scheme that takes forwarded attributes and uses them would subvert that expectation.

rshriram:
need ack from some more mixer folks to approve

kyessenov:
1. Ingress should not accept forwarded attributes.
2. Egress routes should not have forwarded attributes in the RDS tables.

Other than that, fine to me. Keep in mind that after https://github.com/istio/istio/pull/17132 lands, you'd need to manually generate protobufs with 1.3.2 protobuf in istio/istio.

douglas-reid:
@rshriram how can we move this forward?

PR #1086, opened 2019-09-18
copy golang and gogo protos to two places

kyessenov:
/assign @geeknoid 
/assign @ozevren 

Sorry, I couldn't figure out makefiles. Done by hand since this is just temporary.


geeknoid:
Can't we just generate these go in istio/istio using a // go:generate thingie in Pilot?

kyessenov:
We could, but then we still need to revert the previous change for those protos. So either revert, or merge this.

kyessenov:
Can you revert my previous commit here please?

PR #1090, opened 2019-09-20
Add an annotation option to opt-in autopilot.

incfly:
https://github.com/istio/istio/issues/17276

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -58,6 +58,11 @@ annotations:
       reachable within its namespace.
     deprecated: false
     hidden: false
+  - name: security.istio.io/autoMTLS
+    description: Determines whether the client proxy uses auto mTLS. This overrides the mesh
+      default specified in MeshConfig.enable_auto_mtls.
+    hidden: true
+    deprecated: false
   - name: sidecar.istio.io/inject
     description: Specifies whether or not an istio-proxy sidecar should be automatically
       injected into the workload.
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -74,6 +74,11 @@ annotations:
       istio-proxy sidecar.
     deprecated: false
     hidden: false
+  - name: security.istio.io/OptInAutoMTLS
+    description: Determined whether the client Envoy opt-in auto mTLS behavior. This overrides
```
linsun: s/Determined/Determine/?

annotation/annotations.yaml
```diff
@@ -74,6 +74,11 @@ annotations:
       istio-proxy sidecar.
     deprecated: false
     hidden: false
+  - name: security.istio.io/OptInAutoMTLS
```
linsun: what do you think of just calling it security.istio.io/autoMTLS

annotation/annotations.yaml
incfly: done.

smawson: Determines whether the client proxy uses auto mTLS. This overrides the mesh default specified in MeshConfig.enable_auto-mtls.

annotation/annotations.yaml
incfly: Sounds good. Done.

Just to confirm, the value pairs is still `true/false`. right?

annotation/annotations.yaml
incfly: Done. reworded.

incfly:
Thanks for the comments. Updated, PTAL.

@smawson @linsun

annotation/annotations.yaml
```diff
@@ -74,6 +74,11 @@ annotations:
       istio-proxy sidecar.
     deprecated: false
     hidden: false
+  - name: security.istio.io/autoMTLS
```
smawson: Can you move this so it is not in the middle of the sidecar annotations? Basically keep the annotations sorted in this file would be helpful.

annotation/annotations.yaml
linsun: yes +1

annotation/annotations.yaml
incfly: Done.

incfly:
@linsun  @smawson 
Mind taking another look to see if this is ready to be merged? Thanks!

PR #1094, opened 2019-09-23
Add the traffic.istio.io/locality annotation

venilnoronha:
This adds a new `traffic.istio.io/locality` annotation which will allow users to configure the locality of sidecars and other components (egress gateway, etc.). Currently, this functionality is achieved by applying the `istio-locality` label on sidecars, and by additionally setting the `ISTIO_META_istio-locality` environment variable on egress gateways. The label only supports values in the `region.zone.sub_zone` format while this annotation will allow users to configure locality in the well-known `region/zone/sub_zone` format.

/cc @istio/wg-networking-maintainers

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -327,6 +327,14 @@ var (
           Deprecated: false,
         }
 	
+		TrafficLocality = Instance {
+          Name: "traffic.istio.io/locality",
+          Description: "Specifies the component locality in the "+
+                        "region/zone/sub_zone format.",
+          Hidden: true,
+          Deprecated: false,
+        }
+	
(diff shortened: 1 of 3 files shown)


venilnoronha:
okay, I'll do that.

rshriram:
what is the plan to reconcile this with the k8s node labels? Also, shouldn't we already have one of this in the node metadata already? Not that it makes a difference because Envoy will require its own locality right before start [such a pain]

venilnoronha:
If specified, this will override the locality obtained from K8s node labels. I noticed that Envoy node metadata isn't configured with locality details in some scenarios. That may be a bug. In any case, I think it would be useful to have the proxy locality configurable.

PR #1095, opened 2019-09-23
update common and common-protos

rcernich:
Signed-off-by: rcernich <rcernich@redhat.com>


rcernich:
test is failing because prow is using an older build-tools image, one that doesn't have the renamed `cue-gen`.

rcernich:
This duplicates #1093.

PR #1096, opened 2019-09-24
fix retry policy links

frankbu:
Fixes: https://github.com/istio/istio.io/issues/5032


geeknoid:
Huh, something is broken in the job, it didn't catch the fact you forgot to do "make" before your commit.

I'll look into why the job didn't flag this as a problem.

PR #1097, opened 2019-09-24
Fix build process.

geeknoid:


geeknoid:
Needs https://github.com/istio/test-infra/pull/1834

PR #1100, opened 2019-09-24
removed port name from the port selector

GregHanson:
re-opening https://github.com/istio/api/pull/906 against master

coincides with istio/istio#13054

resolves: istio/istio#8754


GregHanson:
Regarding @costinm's original concerns raised here:
https://github.com/istio/api/pull/906#pullrequestreview-231301996

There already exists validation logic to throw an error about the presence of a Name in the selector since 1.2 was released
https://github.com/istio/istio/blob/master/pkg/config/validation/validation.go#L2252

geeknoid:
@GregHanson This requires work on istio/istio to get things to work. Can you take care of this? Just do "go get istio.io/api" and "go get istio.io/gogo-genproto" in the istio repo, and then iterate until everything builds and passes tests.

PR #1101, opened 2019-09-24
Integrate IstioControlPlaneSpec with MeshConfig

ostromart:
First pass for integrating these two protos per the [integration proposal doc](https://docs.google.com/document/d/19IKo6DDv0x6DKtollcbFivjqLkQGwts3dFCzIFm2swM/edit#heading=h.u2k8rqep28j3).
Once there's basic agreement, there will be follow-on work items:
1. Translate between v1alpha1<->v1alpha2 MeshConfig. Required for backwards compatibility with existing components that expect v1alpha1, and in the other direction for user migration to v1alpha2.
2. Translate from v1alpha2->values.yaml. 
3. Port of the remainder of values.yaml into v1alpha2.


mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+  // declarations in that namespace first and if none are found it will
+  // search in the root namespace. Any matching declaration found in the root
+  // namespace is processed as if it were declared in the leaf namespace.
+  //
+  // The precise semantics of this processing are documented on each resource
+  // type.
+  string default_namespace = 30;
+
+  // Configuration for access logs.
+  AccessLogConfig access_log_config = 50;
```
howardjohn: this seems out of place here.. why is this top level?

ostromart: It seems to straddle traffic and telemetry features but that's a good point. If no one objects I will put it into telemetry.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+  // Disable policy checks by the Mixer service. Default
+  // is false, i.e. Mixer policy check is enabled by default.
+  bool disable_policy_checks = 11;
+
+  // Allow all traffic in cases when the Mixer policy service cannot be reached.
+  // Default is false which means the traffic is denied when the client is unable
+  // to connect to Mixer.
+  bool policy_check_fail_open = 12;
+
+  message OutboundTrafficPolicy {
```
howardjohn: I don't think this belongs under policy, because if a user disables policy (which is the vast majority of users) then this still applies.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+  // make the default 'Current namespace' so that services are only visible
+  // within their own namespaces by default. Operators can then expand the
+  // visibility of services to other namespaces as needed. Use of 'No Namespace'
+  // is expected to be rare but can have utility for deployments where
+  // dependency management needs to be precise even within the scope of a single
+  // namespace.
+  //
+  // For further discussion see the reference documentation for ServiceEntry,
+  // Sidecar, and Gateway.
+  repeated string default_service_export_to = 15;
```
howardjohn: same as above regarding being under policy

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+  // A Pattern can be composed of various pre-defined variables. The following variables are supported.
+  // %SERVICE% - Will be substituted with name of the service.
+  // %SERVICE_FQDN% - Will be substituted with FQDN of the service.
+  // %SERVICE_PORT% - Will be substituted with port of the service.
+  // %SERVICE_PORT_NAME% - Will be substituted with port name of the service.
+  //
+  // Following are some examples of supported patterns for reviews.
+  // %SERVICE_FQDN%_%SERVICE_PORT% will use reviews.prod.svc.cluster.local_7443 as the stats name.
+  // %SERVICE% will use reviews.prod as the stats name.
+  string inbound_cluster_stat_name = 16;
```
howardjohn: Logically this *is* telemetry, so it makes sense to be under telemetry section. But what does it mean if we set telemetry.enabled=false? We still need this configuration even if mixer is disabled (and mixer will be disabled typically in the near future).

Does `telemetry.enabled=false` == `mixer.enabled=false`? 

I guess to me its just not clear exactly what it means to "disable telemetry", because there are so many things that are telemetry but are distinct things (mixer, mixerv2, telemetry lite, access logs, ....)

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+    // a secondary ingress controller (e.g., in addition to a
+    // cloud-provided ingress controller).
+    STRICT = 2;
+  }
+
+  // Defines whether to use Istio ingress controller for annotated or all ingress resources.
+  IngressControllerMode ingress_controller_mode = 12;
+
+  // Locality based load balancing distribution or failover settings.
+  LocalityLoadBalancerSetting locality_lb_setting = 13;
```
howardjohn: This applies to sidecars as well, so probably shouldn't be here and instead be in traffic management?

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+
+  message Components {
+    // Namespace that auto injections components are installed into.
+    string namespace = 1;
+
+    // Ingress/egress gateway configuration.
+    IngressGatewayComponentSpec ingress_gateway = 31;
+    EgressGatewayComponentSpec egress_gateway = 32;
```
howardjohn: How do we support multiple gateways? Should we make this a repeated field?

This may apply to other components as well in the future for canary, etc, but as far as I know people are only really running multiple gateways

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+  }
+
+  Components components = 50;
+}
+
+// Configuration for proxy access logs.
```
howardjohn: Is this just for proxy? I thought we we would have a common config for control plane as well, or is that already in all the `_Specs`?

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+  // Releases are published to docker hub under 'istio' project.
+  // Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly
+  string hub = 12;
+  // Version tag for docker images e.g. 1.0.6
+  string tag = 13;
+
+  // Overrides for default global values.yaml.
+  TypeMapStringInterface values = 50;
+  // Unvalidated overrides for default global values.yaml.
+  TypeMapStringInterface unvalidatedValues = 51;
```
howardjohn: When does a user use this vs `values`?

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+// Configuration options for the pilot component.
+message PilotComponentSpec {
+  google.protobuf.BoolValue enabled = 1;
+  string namespace = 2;
+  KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for the proxy.
+message ProxyComponentSpec {
+  google.protobuf.BoolValue enabled = 1;
```
howardjohn: what does it mean for this to be disabled?

mesh/v1alpha2/config.proto
ostromart: There's a difference between feature and component settings. Feature setting does not require the presence of any components, nor does it map to any particular component. 
Is traffic policy a part of traffic or policy feature? If we have a policy feature at all, I'd argue it should be the latter.

howardjohn: (this applies to a bunch of my comments as well)

I don't get what happens if I set `policy.enabled=false` though. I have to have an OutboundTrafficPolicy, so what does disabling policy do? 

It seems maybe we only need enabled for the individual components, not for the top level stuff?

mesh/v1alpha2/config.proto
ostromart: As per the other comment, feature settings are compatible with the feature being disabled. Feature disabled generally means "no components or resources are installed for this feature". But certainly there's a grey area where it could mean more (if we want it to), and this is where we may need further clarification. The down side of of the field meaning more is the overhead associated with maintenance and documentation, and also possible errors in semantics in the implementation.

mesh/v1alpha2/config.proto
ostromart: Gateways are likely to be user config so it's not clear if this needs to cover all the use cases. For now it's there to support demo/out of the box installs but for serious use, users would turn these off and create their own. 
I'd argue for leaving it as is, otherwise this may get very complex.

mesh/v1alpha2/config.proto
ostromart: We could, do you know if this mesh config setting currently applies to other components? If not, I'd leave it as is to reflect reality and add it to other components when supported.

mesh/v1alpha2/config.proto
ostromart: It is used for custom charts that have values not in the official schema.

mesh/v1alpha2/config.proto
ostromart: This is a weird one. It's probably the only one where this could be dropped since it's really autoInjector feature that controls it and we don't want that control to be duplicated.

mesh/v1alpha2/config.proto
ostromart: I get your point, but if we go with the definition of feature disabled == no resources installed for a feature it's probably still a useful abstraction, otherwise we have to set all components to be disabled (it's one component in policy but three in security), plus it's not clear which component enablement controls any common resources. 
But if disabled means something else than we need to rethink exactly what it means for every feature - it will be different in every case. For that reason it's probably better to stick with the first, simpler definition.

howardjohn: I think its pretty confusing to have this config

```yaml
policy:
  enabled: false
  outboundTrafficPolicy: ALLOW_ANY
```

In our existing helm charts, if a field is ever `enabled: false` then none of the other config does (or even can) have any impact. But in this case the `outboundTrafficPolicy` does matter. I think this pattern is fairly common in other helm charts too.

It seems like the `enable` field is directly linked to the components logically, but not in the api.

Would it make more sense to have

```yaml
policy:
  components:
    enabled: false
  outboundTrafficPolicy: ALLOW_ANY
```

or even


```yaml
policy:
  outboundTrafficPolicy: ALLOW_ANY
components:
  policy:
    enabled: false
```

mesh/v1alpha2/config.proto
howardjohn: will the operator not support users configuring their own gateways?

mesh/v1alpha2/config.proto
howardjohn: I think all the control plane read the log settings from command line flags for now

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,694 @@
...
+  // Default namespace if feature or component namespaces are not set.
+  // The namespace to treat as the administrative root namespace for
+  // Istio configuration. When processing a leaf namespace Istio will search for
+  // declarations in that namespace first and if none are found it will
+  // search in the root namespace. Any matching declaration found in the root
+  // namespace is processed as if it were declared in the leaf namespace.
+  //
+  // The precise semantics of this processing are documented on each resource
+  // type.
+  string default_namespace = 30;
```
ayj: MeshConfig itself is namespace'd. Can we use that for the default instead of duplicating it here?

mesh/v1alpha2/config.proto
ayj: Agreed with @howardjohn. If full-time Istio developers find it confusing, users are going to be even more confused.

Do we need to nest component under feature? We'll end up interleaving k8s specific resource customization with Istio specific features. Keeping feature/component separate makes it more obvious what "enabled" means based on the context IMO. 

Validation can reject invalid combinations and offer suggestions to fix the problem (e.g. pilot component is disabled but user has enabled a traffic management feature). Validation logic can change over time as features responsibility shifts between components. 

For example:

```yaml
components:
  policy:
    enabled: false
  telemetry:   
    enabled: true
    namespace: istio-telemetry
  pilot:
    enabled: true
    namespace: istio-control
    k8s:
      env:
      - name: GODEBUG
        value: gctrace=1  
    # etc...

 features:
    trafficManagement:
      outboundTrafficPolicy: ALLOW_ANY
```

rshriram:
90% of the stuff in Mesh config is for Pilot's consumption only.  Even if its authN or mixer report server or what not. Given that, I find it confusing to have settings used by pilot be present under citadel, combined with helm chart stuff for citadel.

All in all, I am not convinced that this is the right approach to reconciling values.yaml with mesh config. We should move stuff from values into meshConfig if it doesn't exist. And then we can simply have a ComponentSpec that captures the settings that we currently have in pilot/citadel/gateway/telemetry, etc. IF we go down this path, we wont have to have a v1alpha2 of MeshConfig. We just need a k8s.proto and the componentSpec.proto

mesh/v1alpha2/config.proto
ostromart: I was thinking about this some more last night and I agree with you guys. It is confusing so I will take it out. Regarding nesting the components under feature, there is a benefit to that - we can set a namespace per feature. This is much more likely than setting per component.

mesh/v1alpha2/config.proto
ostromart: They would do it outside of the operator, basically it would be treated like all other Istio user config. The reason is that it has been a struggle to create a sufficiently broad definition for gateway to accommodate all cases. But if the existing Ingress/Egress specs above are sufficient to cover most user cases, we can make these a repeated field and call it a day.

mesh/v1alpha2/config.proto
ostromart: I think we may want to keep the namespace separate for the config from the default namespace for Istio installation - that is the feedback I've had from several people.

ostromart:
@rshriram the fact that most of the parameters are consumed by pilot is an implementation detail that most users don't know or care about. The issue here is that we have a huge, totally flat structure, which is only going to get bigger as we migrate values.yaml parameters into it. This is one way to impose some structure on it that follows relatively well established feature groupings. 
I agree there's a migration cost involved but if we want to take the pain this is the right moment to do it imo.

mesh/v1alpha2/config.proto
ayj: Repeating these as-is results in multiple gateway LBs with identical config (e.g. ports for for multi-cluster passthrough and VM expansion). We probably want two different things:

1. A default set of gateway(s) with appropriate Istio config to satisfy the MeshConfig contract. And some way to disable these defaults.

2. A method for users to create their own Istio gateway LBs.

I believe (2) is technically possible with helm by disable all components except the gateway. I don't think thats appropriate here. We need to differentiate default from user gateways if they both exist in MeshConfig. Alternatively, we add another API for creating gateway LB instances. In either case, the operator could manage the gateway lifecycle.

mesh/v1alpha2/config.proto
ostromart: Setting namespace for a feature would mean that any component that happens to be a part of that feature would be installed in that namespace. Also any resources associated with that feature would go in there. If we accept the concept of a feature, I don't see a problem with grouping components under it.

mesh/v1alpha2/config.proto
ostromart: sgtm.

mesh/v1alpha2/config.proto
ostromart: IMO the gateway question requires significantly more work on both the API and resulting operator code. Given the short amount of time for 1.4, I propose we defer this work till 1.5.
Since we still have the helm passthrough in this API, we could just remove the gateway component completely for this release and ask users to use the helm API for gateway components.

mesh/v1alpha2/config.proto
ostromart: Removed.

mesh/v1alpha2/config.proto
ostromart: Ok, but without this field how could we specify a default istio namespace if it's different from MeshConfig CR? User would have to specify it for each feature.

mesh/v1alpha2/config.proto
ostromart: Removed the enablement at feature level per other thread.

mesh/v1alpha2/config.proto
ayj: Can you provide an example of when we'd want the MeshConfig's namespace to be different than the default?

ostromart: Managed Istio, or cases where admin wants to give users ability to configure Istio installation without granting all the privileges of Istio itself.

ayj: I see. Admins will probably need some additional level of admission controller to prevent users from setting the default_namespace (or component namespaces) to the CR's own namespace.

costinm: We need to stop adding features and options because someone might need it. It's confusing users, increases testing surface, etc. 

This is clearly not part of 1.4 scope, so please remove it for now. If it's needed - we'll need e2e tests, docs, etc.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+// $description: Configuration affecting the service mesh as a whole.
+// $location: https://istio.io/docs/reference/config/istio.mesh.v1alpha2.html
+
+// Configuration affecting the service mesh as a whole.
+package istio.mesh.v1alpha2;
+
+option go_package="istio.io/api/mesh/v1alpha2";
+
+// MeshConfig defines a top level configuration for all Istio control plane components.
+message MeshConfig {
```
costinm: Don't we already have the top level proto ? I don't get it - the agreement was to use the existing mesh config, and add what is required.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+  // Selection and configuration of core Istio features.
+  TrafficManagementFeatureSpec traffic_management = 10;
```
costinm: The agreement was to add what is required to the existing mesh config - with a migration plan if a setting already existed and it really needed. Since policy, telemetry are on the way out - I don't see a reason to make changes to the existing settings or introduce user pain ( since they are familiar with the current settings ).

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+  PolicyFeatureSpec policy = 11;
+  TelemetryFeatureSpec telemetry = 12;
+  SecurityFeatureSpec security = 13;
```
costinm: For the others - do we need any of them in 1.4 ? I want to make sure what we add is required and has long-term future.
That was the plan - any deprecation or new API should be reviewed as any new API, with use case, etc.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+
+  // Installation time parameters, used in Helm and operator.
+  InstallSpec install = 60;
+}
+
+// Configuration options for traffic management.
+message TrafficManagementFeatureSpec {
+  // Address of the server that will be used by the proxies for policy report
+  // calls.
+  string mixer_report_server = 10;
```
costinm: See above - no reason to deprecate old setting, and no time to do proper migration, testing, etc. 1.4 is a short release,
and adding APIs that are not implemented ( when by 1.5 it's likely to be deprecated anyways ) doesn't make sense.

costinm: Besides - this is not a config for traffic management.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+
+  // Enable session affinity for Envoy Mixer reports so that calls from a proxy will
+  // always target the same Mixer instance.
+  bool sidecar_to_telemetry_session_affinity = 11;
```
costinm: Same, this is telementry and on its way out.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+
+  // Port on which Envoy should listen for incoming connections from
+  // other services.
+  int32 proxy_listen_port = 12;
```
costinm: We already have those options in mesh config - if you want to deprecate and migrate  - I don't think we have time in 1.4, we can do it in 1.5 if it is prioritsed. Networking WG should review any changes related to networking, we need test/migration plan, etc. And not sure it's is worth it for settings that are well-established and broadly used in code.

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+  // MongoDB, etc., Envoy will timeout on the protocol detection after
+  // the specified period, defaulting to non mTLS plain TCP
+  // traffic. Set this field to tweak the period that Envoy will wait
+  // for the client to send the first bits of data. (MUST BE >=1ms)
+  google.protobuf.Duration protocol_detection_timeout = 15;
+
+  // If set then set SO_KEEPALIVE on the socket to enable TCP Keepalives.
+  istio.networking.v1alpha3.ConnectionPoolSettings.TCPSettings.TcpKeepalive tcp_keepalive = 16;
+
+  // Unix Domain Socket through which Envoy communicates with NodeAgent SDS to get key/cert for mTLS.
```
costinm: Another setting that is likely to go away ( even in 1.4)

mesh/v1alpha2/config.proto
```diff
@@ -0,0 +1,669 @@
...
+  // Use secret-mount files instead of SDS if set to empty.
+  string sds_uds_path = 18;
+
+  // ConfigSource describes a source of configuration data for networking
+  // rules, and other Istio configuration artifacts. Multiple data sources
+  // can be configured for a single control plane.
+  repeated ConfigSource config_sources = 19;
```
costinm: I don't get it - we just added them few releases ago. Why do they need to move ?

mesh/v1alpha2/config.proto
costinm: Multiple namespace support is not in 1.4 roadmap and was dropped as a feature for installer. We can add it back - but not in 1.4, too little time to test it properly.

ostromart:
@costinm I thought the agreement was actually to restructure the config to be per-component. I haven't added or deleted anything from MeshConfig, simply moved the knobs to the appropriate component. Let's sync up offline to see where the disconnect is.

rshriram:
Rahter than mixing the component helm deployment stuff with the component settings under one component box, can you separate the two? like under existing mesh config, you could have a simple

repeated ComponentSettings
where componentSettings is one of PilotSettings, MixerSettings, etc.. 

Most istio users are familiar with the mesh config. Moving this out will cause lots of confusion. In addition, people who do not plan on using the operator will not have to pay the penalty of migrating to new crds or new tooling. Its just additional fields in the same crd.

costinm:
+1 on Shriram's point. People are familiar with current mesh config, it
should be a high bar to make changes to APIs we already support.

For missing configs: I would rather not use 'pilot' or 'mixer', the API is
supposed to be 'mesh' and independent of current component internal names.
Besides, since Mixer is going away we should not spend too much time on it.

We were talking on weather operator and new installer should even support
mixerv1 and policyv1 - users who  need them can continue to use
the old installer, and telemetry v2 is already pretty good shape and much
faster.

But my main concern here is making sure we don't bulk-add changes - and
review carefully each new API/config, since we'll need to support it
for quite a while.

mesh/v1alpha2/config.proto
ostromart: Removed in new integration attempt.

mesh/v1alpha2/config.proto
ostromart: Leaving this alone in second merge attempt.

mesh/v1alpha2/config.proto
ostromart: Removed in new integration.

mesh/v1alpha2/config.proto
ostromart: Changed in pass 2.

mesh/v1alpha2/config.proto
ostromart: See pass 2.

mesh/v1alpha2/config.proto
ostromart: See pass 2.

mesh/v1alpha2/config.proto
ostromart: See pass 2.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    //     - minimal (looks in profiles dir for a file called minimal.yaml)
+    //     - /tmp/istio/install/values/custom/custom-install.yaml (local file path)
+    // default profile is used if this field is unset.
+    string profile = 100;
+    // Path for the install package. e.g.
+    //     - /tmp/istio-installer/nightly (local file path)
+    string install_package_path = 102;
+    // Root for docker image paths e.g. docker.io/istio-release.
+    // Releases are published to docker hub under 'istio' project.
+    // Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly
```
howardjohn:
```suggestion
```

mesh/v1alpha1/config.proto
```diff
@@ -356,14 +356,16 @@ message MeshConfig {
...
   // %SUBSET_NAME% - Will be substituted with subset.
-  //  
+  //
   // Following are some examples of supported patterns for reviews.
   // %SERVICE_FQDN%_%SERVICE_PORT% will use reviews.prod.svc.cluster.local_7443 as the stats name.
   // %SERVICE% will use reviews.prod as the stats name.
   string outbound_cluster_stat_name = 45;
 
+  // Installation parameters.
+  IstioControlPlaneSpec install = 46;
```
costinm: Once again, operator and mesh config are NOT about install. It is about controlling ( or operating ) on an already installed control plane.

Yes, a marginal feature is the initial installation - but that's not what mesh config is about.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+              value: --livenessProbeInterval=123s
+    ```
+*/
+
+// IstioControlPlaneSpec defines the desired installed state of IstioControlPlane.
+// The spec is a used to define a customization of the default profile values that are supplied with each Istio release.
+// Because the spec is a customization API, specifying an empty InstallSpec results in a default Istio control plane.
+message IstioControlPlaneSpec {
+    // Default namespace if feature or component namespaces are not set.
+    string default_namespace = 5;
```
costinm: We are not going to support this in 1.4, please remove or comment out until we support it ( non istio-system is not on the P0 or P1)

costinm: Also: the namespace is actually the namespace where the mesh config is defined - having the resource in one namespace operate on a control plane in a different namespace doesn't make sense. So just remove this option.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+
+    PilotComponentSpec pilot = 30;
```
costinm: Many of those components are likely to be gone and deprecated, and also use a mix of internal names ( galley, pilot ) and feature names ( telemetry, etc).

What is the minimal feature that we must have in 1.4 ? requirement/limits ? Patches ? Can we use an unstructured set of patches ( like kustomize ), without hardcoding internal component names ?

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    NodeAgentComponentSpec node_agent = 37;
+    GalleyComponentSpec galley = 38;
+    IngressGatewayComponentSpec ingress_gateway = 39;
+    EgressGatewayComponentSpec egress_gateway = 40;
+    CNIComponentSpec cni = 41;
+
+    // Overrides for default values.yaml.
+    TypeMapStringInterface values = 50;
+    // Unvalidated overrides for default values.yaml. Used for custom templates where new parameters are added.
+    TypeMapStringInterface unvalidatedValues = 51;
```
costinm: I thought we don't want values.yaml to leak into the mesh config ? If we do this, then the list above of components is not needed in 1.4, and we can define once the deprecation and possible reorg of the deployments is more clear. As I mentioned, telemetry/policy will be gone soon, sidecar injector likely to move into pilot, etc.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    // Path or name for the profile e.g.
+    //     - minimal (looks in profiles dir for a file called minimal.yaml)
+    //     - /tmp/istio/install/values/custom/custom-install.yaml (local file path)
+    // default profile is used if this field is unset.
+    string profile = 100;
```
costinm: Profile is used in the old installer - new installer is 'a-la-carte', there is no profile.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    // Path for the install package. e.g.
+    //     - /tmp/istio-installer/nightly (local file path)
```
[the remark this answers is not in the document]
costinm: +1, we can't use local file paths in configs. Docker image / kubernetes / etc can use well-known paths, user doesn't have to know.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    string install_package_path = 102;
+    // Root for docker image paths e.g. docker.io/istio-release.
+    // Releases are published to docker hub under 'istio' project.
+    // Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly
+    string hub = 110;
```
costinm: Technically docker.io is the 'hub',  and istio-release/istio-pilot is the image. If we want to add this to the API - we should probably use the right words.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+
+// Configuration options for the pilot component.
+message PilotComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for the proxy.
+message ProxyComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
```
costinm: How can the proxy be disabled ?

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for the sidecar injector component.
+message SidecarInjectorComponentSpec {
```
costinm: The sidecar injection mode can be part of the 'main' mesh config - it is not an install override but a runtime config.

I would hold on resource overrides for sidecar ( using only existing values.yaml overrides for 1.4 ) - since this is most likely to be merged into pilot.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for the policy enforcement component.
+message PolicyComponentSpec {
```
costinm: Same for the next 2, and for cert manager which we agreed we'll no longer ship.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for certificate manager component.
+message CertManagerComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for node agent component.
```
costinm: Node agent is not yet supported ( upgrade/downgrade don't work), and there are few proposal to fold it into either pilot or pilot agent. No point in adding it to the API.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for egress gateways.
+message EgressGatewayComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for cni component.
```
costinm: I would also leave this out from 1.4 - requires pretty high permissions, not sure operator is ready for this ( in-cluster) or it's the right thing to do. 

Also most likely this will be replaced by a CRI.

costinm: And I don't think we have time to test turning this on/off dynamically on config change. 

For install we can use the values.yaml escape hatch.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+message CNIComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// KubernetesResourcesConfig is a common set of k8s resource configs for components.
+message KubernetesResourcesSpec {
```
costinm: This is a part I like a lot - didn't look at all details, but it's a good replacement for the values.yaml and template conflicting styles. 

Would it be possible to have a map<component, KubernetesResourceSpec> ?

howardjohn: +1 this is great. does it make sense to also have a global setting here ? useful for stuff like imagePullPolixy

mesh/v1alpha1/config.proto
ostromart: Feel free to suggest a different name - I honestly don't care.

mesh/v1alpha1/install.proto
ostromart: When they are deprecated we will also deprecate them in the API. They are part of 1.4 and hence they will remain here too.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,391 @@
...
+// Selects whether this component is installed.
+// google.protobuf.BoolValue enabled = 1;
+
+// Kubernetes resource spec.
+// KubernetesResourcesSpec k8s = 80;
+
+// Configuration options for the pilot component.
+message PilotComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
```
howardjohn: should be call this kubernetes instead of k8s? I remember seeing someone complain about using k8s but maybe that's just for docs

mesh/v1alpha1/install.proto
ostromart: How is a user going to pass values.yaml for the ones that are not in MeshConfig otherwise?
I keep hearing "will be gone soon". It's not gone in 1.4 and hence it needs to be supported through any API for 1.4. Let's talk again in 1.5.

mesh/v1alpha1/install.proto
ostromart: Why is it in istio/installer 1.3 released charts?

mesh/v1alpha1/install.proto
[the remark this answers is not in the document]
ostromart: The controller would run from built-in charts so this would usually be unset in the controller case. We could change it to a flag for the CLI I suppose.

mesh/v1alpha1/install.proto
ostromart: I had actually removed it but hadn't pushed. Done.

mesh/v1alpha1/install.proto
ostromart: I thought at least default, demo and minimal would still be there?

mesh/v1alpha1/install.proto
ostromart: Any suggestions? Repo maybe?

mesh/v1alpha1/install.proto
ostromart: Agreed, I forgot to remove this.

mesh/v1alpha1/install.proto
ostromart: You mean remove enable field? I agree that's runtime config so it makes sense.
For the resources, I think we can easily remove this if sidecar doesn't go into 1.4. I have some doubts that we'll see such dramatic changes in this short release.

mesh/v1alpha1/install.proto
ostromart: Removed.

mesh/v1alpha1/install.proto
ostromart: I don't understand why the high privileges would be more problematic for controller vs installed via CLI?

mesh/v1alpha1/install.proto
ostromart: We could do map instead but:
- we lose compile time checks
- it still looks the same to the user. We just save a little repetition in the proto.
Let me take a look at other global settings we can include (imagePull is definitely a good one).

mesh/v1alpha1/install.proto
ostromart: I think there were complaints only in the docs. kubernetes is just more characters...

mesh/v1alpha2/config.proto
ostromart: Removed.

mesh/v1alpha1/install.proto
ostromart: Done.

mesh/v1alpha1/config.proto
linsun: icp? )

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+import "k8s.io/api/autoscaling/v2beta1/generated.proto";
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "github.com/gogo/protobuf/protobuf/google/protobuf/wrappers.proto";
+
+package istio.mesh.v1alpha1;
+
+option go_package="istio.io/api/mesh/v1alpha1";
+
+/*
+IstioControlPlaneSpec is a schema for both defining and customizing Istio control plane installation resources.
```
linsun: remove "installation" as it does more than install?

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+
+The simplest install specialization is to point the user InstallSpec profile to a different values file, for
```
linsun: also have issue with InstallSpec...

mesh/v1alpha1/install.proto
linsun: i thought so too, it is also in operator

linsun:
@costinm @ostromart @howardjohn I would like to have some clarification on this proposal.  When I first heard it, I had thought the proposal is to merge values.yaml configuration that aren't represented in IstioControlPlaneSpec (e.g. traceSampling) to MeshConfig so that folks can modify it via MeshConfig if needed.  

If we are merging everything from IstioControlPlaneSpec to MeshConfig (This PR seems to propose this if I'm reading correctly), how does user make changes like number of replicas for pilot?  There seem to be 2 ways:
1. user modify istio control plane spec resource and redeploy it.  Our istio operator controller would watch and react to the change.
2. User updates the meshconfig directly.

Why would we want to expose 2 ways to do things?  How would users choose 1 way over the other?

cc @sdake

ostromart:
@linsun what we have here is the first step, not the final product. The final product will be:
- values.yaml will be deprecated (it can continue to be supported outside of istio/api as long as we have helm users)
- any settings that we want from values.yaml that are not already in here will be migrated gradually into MeshConfig. Since we already cover all the k8s related settings in IstioControlSpec (which probably needs a different name now) and the really important runtime config is in MeshConfig, I'm not sure what that actually leaves. 

For 1.4, there will indeed be multiple paths to access some parameters (through MeshConfig and values.yaml field embedded in it). At least I don't see how we can do away with values.yaml at this point without migrating a bunch of config knobs. Some of these are fairly easy but some (like user defined gateways) are quite hard.

mesh/v1alpha1/config.proto
costinm: My point was that all fields are about setting config options. We can add a section 'telemetry' under mesh config, or 
'autoinjection' - no need to have a layer called 'install'. 

The options should be about mesh behavior or properties - not about specific components.

mesh/v1alpha1/install.proto
costinm: That's not how things work - the API is supposed to be supported on the long run, and unlike helm it describe the properties of the mesh. 

For example - we already have a well documented/known mechanism to disable telemetry, by setting the mixer address to "".  For 1.3 - I think this is sufficient control, no need to spend resources testing enable/disable on telemetry server. 

Since we seem to still support values.yaml as fallback - users can use it for those settings without us creating API flakiness.

mesh/v1alpha1/install.proto
costinm: We can use a separate config map with the values.yaml overrides - without adding this to MeshConfig and creating the impression that we'll support this.

The idea of having this in API is to make sure any option we add to the API is properly reviewed, tested and users can assume it'll be supported for few releases ( even if the API is marked as alpha - most fields have been stable for a long time).

mesh/v1alpha1/install.proto
costinm: I thought we agreed demo is not the goal for operator - we have the kubectl install, old installer - let's focus on prod users and stop the demo. 

We do want to primarily support 'default', and possibly minimal - but this is not really a 'profile', we should 
make it clear it's a-la-carte, i.e. users chooses different features  ( not components ) and it's not a fixed-menu.

mesh/v1alpha1/install.proto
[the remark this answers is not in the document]
costinm: CLI flag sounds good, it's certainly not a 'mesh config' property.

mesh/v1alpha1/install.proto
costinm: I don't know - what terms are used in other operators or in docker hub docs ?

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+// KubernetesResourcesSpec k8s = 80;
+
+// Configuration options for the pilot component.
+message PilotComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for the proxy.
+message ProxyComponentSpec {
```
costinm: Proxy settings are already expressed in a structure - no need to create a parallel one, you can add it there.

mesh/v1alpha1/install.proto
costinm: It's not about when/if the change happens - again, please stop thinking in terms of 'install only' - this is a proper API, focused on the behavior of the mesh not on what is installed and where. 

In top level mesh config we can have a filed 'injection', with settings like 'mode' ( all / label / none ), and possibly in future a URL  for having an out-of-cluster injector, maybe a signing CA. This will be used by the webhook - as well as any other component that wants to know how mesh is behaving. 

That applies to most of the other settings in this PR - they are still 'install oriented' - not mesh or feature behavior. 

It should not matter where the injection code runs or which implementation is used - or if we even remove auto-injection via webhook and use CRI/CNI to manage the sidecars - it's a setting controlling Istio behavior, not install or operator specific.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+
+// Configuration options for the policy enforcement component.
+message PolicyComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for the telemetry component.
+message TelemetryComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
```
costinm: How about: for telemetry and policy just use the value of mixerAddress (already existing) in 1.4. This is how users are trained to control telemetry, no need to add an API just to remove it soon after. 

Besides - telemetry will remain enabled even if mixer is not running. This is clearly not correct - users still get most of the telemetry from envoy.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for Citadel component.
+message CitadelComponentSpec {
```
costinm: Same as with the other - I would run it by security WG, but I think the proper way would be to have a "security" section at top level, and maybe a provider: with citadel as an option, or the non-citadel CA provider URL or other related settings. There is a proposal to add trust aliases as well, sds may migrate there as well. 

Meanwhile -  if users want to disable citadel in 1.3 they can use the values.yaml override.

mesh/v1alpha1/install.proto
costinm: A lot of things in 1.3 release charts are alpha or don't work properly for upgrade. 

I would assume that for each of those 'enabled' option and settings you add there will be a test to verify 
that turning on/off the new API works ? If yes - will be pretty easy to see that this one in particular doesn't 
work ( there are some manual steps to install a PSP, etc)

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+// Configuration options for node agent component.
+message NodeAgentComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for galley component.
+message GalleyComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
```
costinm: Same as in the other changes - please use the feature or behavior, not component/install name. 

mcp_server ? 

There is already a section on mcp addresses - I suspect we can just check if istio-galley.istio-system is on the list to 
determine if galley should run, do maybe no need to add this redundant section ?

If you just need a place to attach k8s - I mentioned using a map<string,k8sresourceSpec> - keyed by component name. Or keeping them in a separate file, to not make the mesh config dependent on internal code names.
( see the discussion in Chiron in env wg)

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+}
+
+// Configuration options for ingress gateways.
+message IngressGatewayComponentSpec {
+    google.protobuf.BoolValue enabled = 1;
```
costinm: This is interesting because Istio supports multiple ingress gateways. Same for egress.

mesh/v1alpha1/install.proto
costinm: Because CLI runs from a machine where cluster-wide secrets are offline and protected by password/second-factor/etc.

While in-cluster controller will need permanent high privileges, with relatively simple ways for other pods in same namespace to run with the same account and escalate any attack. 

AFAIK that was one of the major reasons for people not running tiller and the helm3 changes.

mesh/v1alpha1/install.proto
costinm: We still have all compile checks - on the value. 

The internal component names are not 'mesh config'.

You can also have it as a list - with component name as a field. I believe that's how kustomize works, match is based on the resource name which is part of the patch.

costinm: Actually - list of k8sresourcespec with Meta: Name: ... is better for consistency with kustomize.

costinm: And in the same direction: please make sure the fields are following the same structure, I have a feeling they are not. 

No reason to use a different structure for things that express the same thing.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,380 @@
...
+    k8s.io.api.core.v1.ServiceSpec service = 12;
+    // k8s deployment strategy.
+    // https://kubernetes.io/docs/concepts/workloads/controllers/deployment/
+    DeploymentStrategy strategy = 13;
+    // k8s toleration
+    // https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+    repeated k8s.io.api.core.v1.Toleration tolerations = 14;
+
+    // Overlays for k8s resources in rendered manifests.
+    repeated k8sObjectOverlay overlays = 100;
```
costinm: Maybe this is also something to keep for 1.5 ? Little time, too much to test ?

linsun:
@costinm thanks, I think I agree with that assessment for 1.4

@ostromart re your comment, This is exactly what I'm confused...."any settings that we want from values.yaml that are not already in here will be migrated gradually into MeshConfig. Since we already cover all the k8s related settings in IstioControlSpec (which probably needs a different name now) and the really important runtime config is in MeshConfig, I'm not sure what that actually leaves."  Why are we moving everything from values.yaml to mesh config, vs just moving things from values.yaml but not covered in IstioControlPlaneSpec to MeshConfig?

mesh/v1alpha1/install.proto
sdake: I was of the impression we wanted profiles as well.

a-la-carte is great for super-advanced users.  For those that are not trialing Istio (the demo use case), but instead deploying somewhere in the middle, do we want profiles or not?  I could see how profiles could quickly get out of hand, although if we keep the scope of profiles small, that should not be a concern.

@linsun do you have a reference to the original profiles we defined?

mesh/v1alpha1/install.proto
sdake: there are hubs, projects-or-orgs,images,tags.

hub/project/image:tag is the cononical way to describe these properties.

mesh/v1alpha1/install.proto
sdake: @costinm I don't understand the conflation with the operator and cni.  CNI does require a high degree of privileges (to intall IIUC), however, I don't understand what that has to do with the conttroller vs the installer CLI.

Can you elaborate on CRI plans to replace CNI please.

mesh/v1alpha1/install.proto
sdake: this should be spelled as kubernetes or Kubernetes.

Thanks :)

sdake:
I am a little confused as well.  I thought the agreement was to have runtime configurable options go to MeshCnfig, and have install time configurable options go to IstioControlPlane.  It may be that IstiioControlPlane additionally contains runtime configuration options (via `values.yaml`) until they are migrated to `IstioControlPlane` - not `MeshConfig`.

rshriram:
LGTM after all the comments from others are addressed.

mesh/v1alpha2/config.proto
ostromart: Removed.

mesh/v1alpha1/install.proto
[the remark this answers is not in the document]
ostromart: It's same as the chart path - likely to be unset for controller case but I don't see why it couldn't be a mount. This could be turned into a flag for CLI only if we want to explicitly disallow this.

mesh/v1alpha1/config.proto
ostromart: Are you saying to put the feature sections back, but without moving the MeshConfig parameters into the feature structs?
I actually removed the feature layer since there was nothing in there anymore and just have components now, but I can put it back.

mesh/v1alpha1/install.proto
ostromart: The runtime config is expected to be stable over time, no debate there.
You could make the component installation resources part more durable by making it
dynamic (your map suggestion elsewhere) but I don't think that's better than managing the components using the regular API management practice of adding and deprecating, which is syntax checked and makes it explicit that a component is going away in the future. The problem with a dynamic map without hard coded component names is that user is always wondering whether a particular value is valid for a particular build of Istio - there's no explicit contract of any kind. I think that's best reserved for things like extending the API, rather than managing a not-particularly-fast changing list of components.

mesh/v1alpha1/install.proto
ostromart: What's the advantage of defining a separate ConfigMap for values and where would it be defined? 
If we want to make it clear that this is going away, the normal way to do it is would be to mark it as deprecated. I think that would convey exactly the right message in this case.

mesh/v1alpha1/install.proto
ostromart: I don't understand what you mean by "install oriented". We can split all these parameters into two parts: 1. runtime config (mesh components react to changes in these) and 2. what I'm calling install config (which is outside the components, things like HPA, k8s resources etc.) We can call it something else if this is a problem. If there's an overlap like "enabled" being in both runtime and install like for sidecar injector, let's eliminate it. That's what I was trying to clarify with my question above.
Defining (or not) resources for sidecar injector for 1.4 is orthogonal to this. It comes under the other discussion we're having which is whether components should be explicitly named or not -we should continue there.

mesh/v1alpha1/install.proto
ostromart: The things that are alpha or experimental should be explicitly documented. Operator follows and uses exactly the installer charts and we should avoid creating a split - if a chart is present in installer but experimental, let's still expose it in operator but document it as such. The installer charts should be authoritative and not selectively held back in operator, otherwise we'll create confusion between helm and operator users. 
Another option is to keep a chart in installer master but exclude it from release-1.4 (say if it's completely broken), in which case we can also easily remove it from the API.

mesh/v1alpha1/install.proto
ostromart: Which fields following the same structure are you referring to?

mesh/v1alpha1/install.proto
ostromart: I'd do it for you Steve :-) 
I'll just point out that the kubernetes protos themselves use the abbreviation e.g. even in this file we have k8s.io.api. Changing our fields to kubernetes will create an inconsistency where we're mixing the two spellings. 
But if you feel very strongly about it, I can live with it.

mesh/v1alpha1/install.proto
ostromart: Done.

mesh/v1alpha1/install.proto
ostromart: Done I think?

mesh/v1alpha1/install.proto
ostromart: Which structure?

mesh/v1alpha1/install.proto
ostromart: Done.

mesh/v1alpha1/install.proto
ostromart: Sounds good, but I feel like this is something we should tackle incrementally rather than this PR, since it's a real API change. The original changes I made were a purely mechanical reshuffling of parameters into groups, your proposal has some actual meat in it.

mesh/v1alpha1/install.proto
ostromart: We're conflating two things here: "feature" settings and "component" settings. I had feature structs in the first attempt to integrate the protos but I've since taken them out, since there's nothing left there anymore. The feature settings should not have any internal names.
The component settings necessarily name internal components because they map resources 1:1 with those components. Again I think this boils down to listing the components vs. your map proposal, let's continue in that thread.

mesh/v1alpha1/install.proto
ostromart: Gateways anyway require values.yaml API to be used. This just allows the built-in gateways' resources to be controlled through this new API. User gateways can still be defined, but their resources must be set using values.yaml (for now).

ostromart:
@costinm @linsun the operator will not work fine without this PR. This PR represents the minimal amount of integration we need for operator to use the converged MeshConfig that we agreed to. 
What I don't want this PR to become is a complete overhaul of the values.yaml API. That's not what we agreed to and such a PR will never complete in the forseeable future. 
I propose we agree to a minimum integration - which I think this PR is now very close to - and leave the bigger task of migrating values.yaml incrementally since that involves different WGs and is better handled through smaller, incremental PRs rather than a single giant one.

ostromart:
@linsun sorry, that's exactly what I meant - we will move any fields in values.yaml _that are not already in IstioControlPlaneSpec_ into MeshConfig, unless we decide to deprecate them instead.

ostromart:
@sdake that's correct - values.yaml config that covers "install time" (k8s resources, HPA etc) is moving into IstioControlPlaneSpec in this PR. Going forward, only the fields in values.yaml not already in MeshConfig or IstioControlPlaneSpec will be considered for migration into MeshConfig or deprecated. Finally, we will only have MeshConfig (for runtime) and ICPS (for install resources). 
Operator will use at least ICPS, although it's possible for it to act on changes in runtime config if there's a need to do so.

mesh/v1alpha1/install.proto
ostromart: docker.io/istio-release doesn't seem to be covered by these canonical terms. The closest would be HubProject I suppose.

mesh/v1alpha1/install.proto
smawson: istio.mesh.v1alpha1.ProxyConfig

Available as the field "default_config", unfortunate choice of name.

It does seem easy to just throw the k8s resource spec in there.

mesh/v1alpha1/config.proto
smawson: Also the proto file shouldn't be "install.proto" it should be "control_plane.proto" or "control_plane_components.proto" or something. That is if we need it to be separate.

mesh/v1alpha1/install.proto
smawson: I think costin is trying to propose just adding the kubernetes resource spec to existing mesh config locations, rather than introducing a new top-level grouping for the components.

So if we already have a top-level group for something (like proxy or injector), we reuse that, otherwise we should create a top-level grouping and put the stuff there.

BTW there isn't a top-level field for injector that I see, nor any way to enable/disable it in meshconfig. Maybe I'm looking in the wrong place. I think the name "sidecar" in the message is confusing the issue, this is the injector, not the sidecar, right? Sidecar resource overrides are in the proxy component spec.

mesh/v1alpha1/install.proto
smawson: We maybe should add Mixer to these names to make it clear these are configs for MixerTelemetry, not Telemetry overall.

mesh/v1alpha1/install.proto
smawson: It seems totally reasonable to have a top-level "citadel" field in the spec and configure the component. That should hold citadel-specific config, and a "security" section could have security-specific config (maybe you nest citadel under security if we feel the need to).

But generally I agree on holding off a major refactor of mesh config in this PR.

mesh/v1alpha1/install.proto
smawson: :+1: to being explicit about the difference between configuration of a feature (such as whether config validation is on) and configuration of a component (such as whether we run galley, and if so how it is run). These are different concerns and should remain separate.

mesh/v1alpha1/install.proto
smawson: If adding this to the spec requires increasing the controller's permissions it does seem a bit scary. It probably makes sense to leave it out and have the user manually enable or disable CNI themselves (controller won't manage that for now).

mesh/v1alpha1/install.proto
smawson: I prefer having the component names as fields in the proto, it seems more explicit to the user what we support (and we have a path to deprecation). Maps are too dynamic. We're changing the meaning of mesh config to include management of the components, so this is inline with the increased scope of the API.

mesh/v1alpha1/install.proto
linsun: @costinm we didn't have time to reach agreement.  While I do agree focusing on default profile (production), I don't agree we should get rid of demo profile.   demo profile is mostly used for user onboard to istio and it is used in many tutorials as well.  It is in our operator today if you do `mesh profile list`.

personally, I think a-la-carte is pretty hard to use.  It took me a few days to sort out and understand the new installer.  Typical users don't want to spend that long to learn installation of Istio.

@sdake we have a WI: https://github.com/istio/istio/issues/14523

@smawson if profile changes runtime configuration (I think it does, for example demo profile has tracesampling to 100), i agree, it should apply to mesh config as well.

mesh/v1alpha1/install.proto
linsun: +1 on leaving cni out for now.

mesh/v1alpha1/install.proto
linsun: Thanks

mesh/v1alpha1/install.proto
linsun: Do we want to use InstallSpec profile here?  Seems best not to imply this is for install only.

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +487,19 @@ message LocalityLoadBalancerSetting{
   // Note: if no OutlierDetection specified, this will not take effect.
   repeated Failover failover = 2;
 }
+
+// Configuration options for traffic management.
+message TrafficManagementFeatureSpec {
```
smawson: Rather than creating empty messages can we move a few fields into each of these (or rather copy them over to do the migration)?

Or if we don't want to do that yet maybe leave MeshConfig alone and have this PR just clean up control plane spec?

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,379 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+import "k8s.io/api/core/v1/generated.proto";
+import "k8s.io/api/autoscaling/v2beta1/generated.proto";
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "github.com/gogo/protobuf/protobuf/google/protobuf/wrappers.proto";
+
+package istio.mesh.v1alpha1;
```
smawson: Should rename this proto to components.proto or controlplane.proto or something.

mesh/v1alpha1/install.proto
```diff
@@ -0,0 +1,379 @@
...
+
+option go_package="istio.io/api/mesh/v1alpha1";
+
+/*
+IstioControlPlaneSpec is a schema for both defining and customizing Istio control plane resources.
+
+The simplest specialization is to point the user IstioControlPlaneSpec profile to a different values file, for
```
smawson: Different profile? We maybe don't want to talk about values since those will go away?

ostromart:
not stale, just sleeping...

ostromart:
Continued in https://github.com/istio/api/pull/1177.

PR #1103, opened 2019-09-25
Docs for namespace of VirtualService gateways

kvudata:
It looks like specifying gateways in other namespaces is an undocumented feature of the VirtualService API.


kvudata:
@googlebot I signed it!

geeknoid:
@howardjohn is correct. Please make the corresponding change in the .proto file, run make from the top of the source tree, and commit anything that's changed.

kvudata:
Thank you for the instructions and sorry for the delay, I'll update this

PR #1104, opened 2019-09-26
Hide ServiceEntry.Location from docs

jwendell:
As it's not used anymore.

### DIFF
--- a/networking/v1alpha3/service_entry.json
+++ b/networking/v1alpha3/service_entry.json
@@ -85,7 +85,7 @@
         }
       },
       "istio.networking.v1alpha3.ServiceEntry.Location": {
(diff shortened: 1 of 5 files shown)


jwendell:
I can also nuke all references to this field on istio/istio, as part of https://github.com/istio/istio/issues/17157, what do you think?

howardjohn:
We can keep it around forever as hidden -- is that better than mentioning
its not used? I am fine with either one, I think we shouldn't actually
remove it though

rshriram:
that also works. just dont remove it yet

jwendell:
@rshriram done.

rshriram:
Given how we have started using this heavily for auto mTLS, i dont think we can hide this anymre

jwendell:
ok, closing then.

PR #1105, opened 2019-09-26
K8sCertificateSetting configures the provision of k8s CA signed certificates

lei-tang:
This PR is for issue: https://github.com/istio/istio/issues/17061
Related PR: https://github.com/istio/istio/pull/17400

Design documents:
[1]https://docs.google.com/document/d/1VHyBre8943USOx_YEVtA18KfEoGTmLT1iBHESO5bzcA/edit#heading=h.qex63c29z2to
[2] https://docs.google.com/document/d/1v8BxI07u-mby5f5rCruwF7odSXgb9G8-C9W5hQtSIAg/edit#heading=h.xw1gqgyqs5b

TODO (lei-tang): allow specifying multiple certificate issuers.
Related issue: https://github.com/kubernetes/kubernetes/issues/69836
CertificateIssuer configures the issuer of a certificate.
message CertificateIssuer {
// The name of the issuer.
  string name = 1;
 // The kind of the issuer, e.g., k8s.
  string kind = 2;
}
message Certificate {
  ....
  // The issuer of the certificate. Multiple issuers may be defined;
  repeated CertificateIssuer issuer_ref = 3;
}


mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,21 @@ message LocalityLoadBalancerSetting{
...
+
+  // The path to the public certificate bundle file of the k8s signing CA.
+  // This parameter needs to be consistent with the k8s CA signing certificate, which
+  // may be customized through --cluster-signing-cert-file.
+  string ca_bundle_file = 2;
+
+  // The services that will be provisioned with certificates.
+  // E.g., istio-galley.istio-system, istio-sidecar-injector.istio-system, and istio-pilot.istio-system.
+  // The format is service-name.service-namespace, with '.' as the delimiter, which
+  // means the service-name and service-namespace should not contain '.'.
```
howardjohn: you already create a service or namespace with a `.`, so no need to warn about this

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,21 @@ message LocalityLoadBalancerSetting{
   // Note: if no OutlierDetection specified, this will not take effect.
   repeated Failover failover = 2;
 }
+
+
+// K8sCertificateSetting configures the provision of k8s CA signed certificates.
+message K8sCertificateSetting{
+  // Whether enable provisioning k8s CA signed certificates
```
howardjohn: Can we add some docs about why this is useful? How should a user decide if they want to enable or disable this?

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,21 @@ message LocalityLoadBalancerSetting{
...
+  bool enabled = 1;
+
+  // The path to the public certificate bundle file of the k8s signing CA.
+  // This parameter needs to be consistent with the k8s CA signing certificate, which
+  // may be customized through --cluster-signing-cert-file.
```
howardjohn: If I were trying to configure this, I wouldn't know what to put here based on the comment. Can we make the comment more clear how to find what this should be? What is `--cluster-signing-cert-file` a parameter to?

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done: added more details on this setting.

mesh/v1alpha1/config.proto
lei-tang: Done: added more details on configuration this parameter and clarification.

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,39 @@ message LocalityLoadBalancerSetting{
...
 }
+
+
+// K8sCertificateSetting configures the provision of k8s CA signed certificates.
+message K8sCertificateSetting{
+  // Whether enable provisioning k8s CA signed certificates.
+  // When it is enabled, the services configured in this setting with certificates
+  // signed by k8s CA. One example usage is to provision certificates to Control Plane
+  // components such as Pilot, Galley, and Sidecar Injector.
+  bool enabled = 1;
```
costinm: This pattern is not currently used in mesh.yaml ( it is specific to helm, an artifact on how we load subcharts in old installer ). Let's keep things consistent. In particular - there is no need for it since the presence of the k8scert settings is sufficient to enable.

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,39 @@ message LocalityLoadBalancerSetting{
...
+
+  // The path to the public certificate bundle file of the k8s signing CA.
+  // This parameter needs to be consistent with the k8s CA signing certificate, which
+  // may be customized through --cluster-signing-cert-file (the Kubernetes CA certificate
+  // used to sign certificates on kube-controller-manager,
+  // details in https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).
+  // In the case that k8s signing certificate is the same as that of the k8s API server,
+  // this parameter may be configured as the pod certificate at /var/run/secrets/kubernetes.io/serviceaccount/ca.crt.
+  string ca_bundle_file = 2;
```
costinm: Let's not add this until it is clear we need it. K8S has it hardcoded, and pilot-agent can probe or use environment-specific detection (like we do for SDS) - it is not something we need the user to worry about, and certainly not something that can be a global setting ( since the only use case for using a different value is to support mesh expansion, and setting this string will break k8s workloads)

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,39 @@ message LocalityLoadBalancerSetting{
...
+
+  // The Pilot service that will be provisioned with a certificate.
+  // E.g., istio-pilot.istio-system.
+  // The format is service-name.service-namespace, with '.' as the delimiter.
+  // If it is an empty, no Pilot certificate is generated.
+  string pilot_service = 3;
```
costinm: - Mesh config should not use internal service names ( it does use some - mixer - but that's a bug ). 
- we already have a discovery address in mesh.yaml - the cert should match that ( so DNS verification works out of box).
- no need to make this optional.

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,39 @@ message LocalityLoadBalancerSetting{
...
+
+  // The directory path to store the Pilot key and certificate (e.g. key.pem, cert-chain.pem,
+  // and root.pem).
+  // Different from other services, Pilot certificate is generated in a directory
+  // rather than mounting from a secret. This eliminates the waiting for mounting the secret.
+  string pilot_certificate_path = 4;
```
costinm: No need to have this exposed to user, until we have clear use case for user changing it. Please keep the default for now. This is also a setting that doesn't work as global.

mesh/v1alpha1/config.proto
```diff
@@ -362,8 +362,11 @@ message MeshConfig {
   // %SERVICE% will use reviews.prod as the stats name.
   string outbound_cluster_stat_name = 45;
 
+  // Configure the provision of k8s CA signed certificates.
+  K8sCertificateSetting k8s_certificate_setting = 46;
```
costinm: Can you rename this to 'certificates' ? And remove K8s ( see the last comment on why )

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,39 @@ message LocalityLoadBalancerSetting{
...
+  // The directory path to store the Pilot key and certificate (e.g. key.pem, cert-chain.pem,
+  // and root.pem).
+  // Different from other services, Pilot certificate is generated in a directory
+  // rather than mounting from a secret. This eliminates the waiting for mounting the secret.
+  string pilot_certificate_path = 4;
+
+  // Other services that will be provisioned with certificates.
+  // E.g., istio-galley.istio-system, istio-sidecar-injector.istio-system, and etc.
+  // The format is service-name.service-namespace, with '.' as the delimiter.
+  repeated string services = 5;
```
costinm: This is probably the only setting we need. However, I think it is better to have a structure for each managed certificate, 
possibly 'inspired' from https://docs.cert-manager.io/en/latest/reference/certificates.html

So I would create a 'message Certificate' - use it as a repeated field in mesh.yaml, under certificates, and add an 'issuerRef' similar with certmanager.

Keep in mind K8S may allow multiple issuers in the same cluster - so we'll have to adapt to that. 

Example structure:

mesh:
   certificates:
      - secretName: galley-cert
        dnsNames: 
        - galley.istio-system.svc
        - galley.mydomain.com
        issuerRef: 
           name: k8s

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,48 @@ message LocalityLoadBalancerSetting{
   // Note: if no OutlierDetection specified, this will not take effect.
   repeated Failover failover = 2;
 }
+
+// CertificateIssuer configures the issuer of a certificate.
+// The path to the CA bundle is hard-coded.
```
costinm: You can remove the comment about the path ( not relevant to the user, implementation detail until an API is exposed to make it configurable)

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,48 @@ message LocalityLoadBalancerSetting{
...
+message CertificateIssuer {
+  // The name of the issuer.
+  string name = 1;
+  // The kind of the issuer, e.g., k8s.
```
costinm: Maybe hide it in 1.4 - since we don't yet support multiple issuers, and k8s is not yet clear on what they want to do  ? 
It's clear we'll need it, but if not implemented or used - we can leave it for 1.5 (and hide )

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,48 @@ message LocalityLoadBalancerSetting{
...
+//       kind: k8s
+// }
+message Certificate {
+  // Which secret the certificate and its key will be stored into.
+  // If it is empty, it will not be stored into a secret.
+  // Instead, the certificate and its key will be stored into a hard-coded directory.
+  string secret_name = 1;
+  // The namespace of the secret into which the certificate will be stored.
+  // If it is empty, it will not be stored into a secret.
+  string secret_namespace = 2;
```
costinm: I would not include secret_namespace - we should only generate in same namespace, we want to stop holding secret permissions to other ns.

Once SDS is used we can use istio-system to hold some secrets for workloads.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Per discussion, it is changed as TODO.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,40 @@ message LocalityLoadBalancerSetting{
...
+//     - galley.mydomain.com
+// }
+// Example 2: key and cert stored in a directory
+// { dnsNames:
+//     - pilot.istio-system
+//     - pilot.istio-system.svc
+//     - pilot.mydomain.com
+// }
+message Certificate {
+  // Which secret the certificate and its key will be stored into.
```
smawson:
```suggestion
  // Name of the secret the certificate and its key will be stored into.
```

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,40 @@ message LocalityLoadBalancerSetting{
...
+  // If it is empty, it will not be stored into a secret.
+  // Instead, the certificate and its key will be stored into a hard-coded directory.
+  string secret_name = 1;
+  // The DNS names for the certificate. A certificate may contain
+  // multiple DNS names.
+  repeated string dns_names = 2;
+  // TODO (lei-tang): allow specify multiple certificate issuers.
+  // The issuer of the certificate. Multiple issuers may be defined;
+  // repeated CertificateIssuer issuer_ref = 3;
```
smawson: Can you take these out? Just stick them in the issue tracking this TODO rather than putting in commented-out code. Same for the message definition.

mesh/v1alpha1/config.proto
```diff
@@ -483,3 +486,40 @@ message LocalityLoadBalancerSetting{
   // Note: if no OutlierDetection specified, this will not take effect.
   repeated Failover failover = 2;
 }
+
+// TODO (lei-tang): allow specify multiple certificate issuer.
```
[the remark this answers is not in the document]
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done.

mesh/v1alpha1/config.proto
lei-tang: Done.

PR #1108, opened 2019-09-30
Doc for VirtualService sourceLabels field

frankbu:


geeknoid:
Please run 'make' and update the PR.

I don't know why the gate is failing to detect that make wasn't run, it's supposed to...

PR #1110, opened 2019-09-30
security: do not generate envoy xds protos here

kyessenov:
Generated files are in https://github.com/istio/istio/tree/master/security/proto/envoy/config/filter


geeknoid:
Shouldn't the associated proto file be deleted too?

kyessenov:
The proto needs to be imported into proxy so it should stay.

kyessenov:
@howardjohn can this be merged? it's confusing to generate protos here that cannot be used by pilot.

PR #1114, opened 2019-10-02
Fix the gen-check target

geeknoid:


geeknoid:
Needs https://github.com/istio/test-infra/pull/1880 to pass the gencheck job.

PR #1118, opened 2019-10-08
Remove unimplemented/unused labels in authn target selector

diemtvu:
It's decided to introduce new, proper API instead of doing this ad-hoc change. The field is not actively use in Istio, and is marked hidden all the time so it should be safe to remove.

Issue: https://github.com/istio/istio/issues/16986


diemtvu:
ping

geeknoid:
Do we want this in 1.4? If so, please set the cherry pick label accordingly.

PR #1120, opened 2019-10-08
add API for ALPN filter

yxue:


kyessenov:
FYI, do not use these gogo protos in an xDS server. I'm trying to delete them unsuccessfully from here https://github.com/istio/api/pull/1110. You need to use standard golang protos.

yxue:
> FYI, do not use these gogo protos in an xDS server. I'm trying to delete them unsuccessfully from here #1110. You need to use standard golang protos.

Are we going to maintain two proto files? One in istio/api, one in istio/istio.

yxue:
Are we going to maintain two proto files? One in istio/api and another one in istio/istio

kyessenov:
The proto can stay here.

PR #1123, opened 2019-10-10
Configure whether Operator manages webhook configurations

lei-tang:
This PR is for issue: istio/istio#17061
The matching PR in istio/installer is: https://github.com/istio/installer/pull/428

Design documents:
[1]https://docs.google.com/document/d/1VHyBre8943USOx_YEVtA18KfEoGTmLT1iBHESO5bzcA/edit#heading=h.qex63c29z2to
[2] https://docs.google.com/document/d/1v8BxI07u-mby5f5rCruwF7odSXgb9G8-C9W5hQtSIAg/edit#heading=h.xw1gqgyqs5b


mesh/v1alpha1/config.proto
```diff
@@ -380,8 +380,12 @@ message MeshConfig {
   // Configure the provision of certificates.
   repeated Certificate certificates = 47;
 
+  // Configure whether Operator manages webhook configurations. The current behavior
+  // of Galley and Sidecar Injector is that they manage their own webhook configurations.
+  bool operator = 49;
```
smawson: Needs a much better name than just "operator", that can mean all kinds of things. Also since this is more about turning off self-managed webhooks, we should call it something that describes exactly what setting it to true does.

We should also make the comment explain what happens if it is true vs. false.

lei-tang: Done: changed the name of the field and the comments.

lei-tang:
After a discussion with Costin, this PR is closed for now.

PR #1125, opened 2019-10-11
Use field_behavior as an option for required fields

jasonwzm:
Use `[(google.api.field_behavior) = REQUIRED]` as an option for required fields. So that our docs will be cleaner in the reference doc (without `required` in different formats and places) and we added a new column in the reference doc to show if the field is required or not. In addition, this helps us with api client tooling for required fields.

For WG reviewers, please note that (from geeknoid@) the end state we want is that anything that's not marked required is implicitly optional.


geeknoid:
Is this sufficient?

We have many fields that are marked as optional. Should we be removing these from the comments?

And then, any fields not currently explicitly marked as optional should receive the new `required` annotation.

jasonwzm:
@geeknoid My assumption was that if the comment mentions REQUIRED then the field is marked as required. Is it true that if a field is by default required, unless OPTIONAL is specified in the comment?

geeknoid:
The existing state is:

- If marked as optional, the field is optional.
- If marked as required, the field is required.
- If not marked, the field is required.

So we need to remove optional/required from the comments, and add the required annotations as you've done and to any fields that didn't previously have any indication of required/optional.

jasonwzm:
WIP: going through the protos to identify required fields with each working group.

mandarjog:
@kyessenov can you help review this from policy and telemetry standpoint ?

policy/v1beta1/cfg.proto
```diff
@@ -136,7 +137,7 @@ message Rule {
   // * `match(destination.service.host, "ratings.*")` selects any request targeting a service whose
   // name starts with "ratings"
   // * `attr1 == "20" && attr2 == "30"` logical AND, OR, and NOT are also available
-  string match = 1;
+  string match = 1 [(google.api.field_behavior) = REQUIRED];
```
kyessenov: This is not really required. I think the code will skip over empty match fiels.

policy/v1beta1/cfg.proto
```diff
@@ -250,27 +251,27 @@ message Action {
...
-  string compiled_template = 67794676;
+  string compiled_template = 67794676 [(google.api.field_behavior) = REQUIRED];
 
-  // Required. The name of the template this instance creates instances for. For referencing compiled-in
+  // The name of the template this instance creates instances for. For referencing compiled-in
   // templates, use the `compiled_template` field instead.
   //
   // The value must match the name of the available template in scope.
-  string template = 2;
+  string template = 2 [(google.api.field_behavior) = REQUIRED];
```
kyessenov: Not required. It's a one-of with compiled_template I think. Both should be optional.

policy/v1beta1/cfg.proto
```diff
@@ -350,22 +351,22 @@ message Instance {
...
-  string compiled_adapter = 67794676;
+  string compiled_adapter = 67794676 [(google.api.field_behavior) = REQUIRED];
 
-  // Required. The name of a specific adapter implementation. For referencing compiled-in
+  // The name of a specific adapter implementation. For referencing compiled-in
   // adapters, use the `compiled_adapter` field instead.
   //
   // An adapter's implementation name is typically a constant in its code.
-  string adapter = 2;
+  string adapter = 2 [(google.api.field_behavior) = REQUIRED];
```
kyessenov: Same here. It's a one-of between adapter and compiled_adapter. Both are optional.

policy/v1beta1/cfg.proto
jasonwzm: Done.

policy/v1beta1/cfg.proto
jasonwzm: Done.

policy/v1beta1/cfg.proto
jasonwzm: Done.

jasonwzm:
@kyessenov can you approve if changes looks good in the policy/ and mixer/ directories?

geeknoid:
@kyessenov Just to be clear, the end state we want is that anything that's not marked required is implicitly optional. Our generated reference docs now show a Required Yes/No column for each field.

networking/v1alpha3/virtual_service.proto
```diff
@@ -770,15 +771,15 @@ message HTTPMatchRequest {
...
-  Destination destination = 1;
+  Destination destination = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // REQUIRED. The proportion of traffic to be forwarded to the service
+  // The proportion of traffic to be forwarded to the service
   // version. (0-100). Sum of weights across destinations SHOULD BE == 100.
   // If there is only one destination in a rule, the weight value is assumed to
   // be 100.
-  int32 weight = 2;
+  int32 weight = 2 [(google.api.field_behavior) = REQUIRED];
```
rshriram: you can't make this as required.. ```If there is only one destination in a rule, the weight value is assumed to
  // be 100.```

jasonwzm: Done.

PR #1128, opened 2019-10-15
update alpn filter api

yxue:


envoy/config/filter/http/alpn/v2alpha1/config.proto
```diff
@@ -23,5 +23,22 @@ option go_package = "istio.io/api/envoy/config/filter/http/alpn/v2alpha1";
 // FilterConfig is the config for Istio-specific filter.
 message FilterConfig {
   // A list of ALPN that will override the ALPN for upstream TLS connections.
-  repeated string alpn_override = 1;
+  repeated string alpn_override = 1 [ deprecated = true ];
```
rshriram: why does this have to be deprecated? We have not released this filter yet. So please remove this completely

yxue: I tried to remove the field but it failed to make. 
```
CONFLICT: "FilterConfig" ID: "1" has been removed, but is not reserved [envoy/config/filter/http/alpn/v2alpha1/config.proto]
CONFLICT: "FilterConfig" field: "alpn_override" has been removed, but is not reserved [envoy/config/filter/http/alpn/v2alpha1/config.proto]
```

It requires to reserve the field and ID.

PR #1129, opened 2019-10-15
Change quota charge to int32

jasonwzm:
In proto to [JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json), `int64` is represented as a `string` field. This is not an issue for users when they specify `int` for the field in yaml. This creates confusion in our mock tests where we convert proto to json and the output of the field becomes a `string`. This will also fail schema validation.

Changing this field for now as this is a user setting and is user visible in docs. `int32` should be more than enough for this field amount.


jasonwzm:
@howardjohn this needs a force merge, so feel free to approve and Martin will be able to merge it.

jasonwzm:
@howardjohn it is not a backward compatible change. however, the type change should be sane to make.

howardjohn:
I don't think we can make backwards incompatible changes like this, won't this break everyone using this api?

jasonwzm:
yes if they are using a charge value bigger than max of int32.

howardjohn:
What about wire protocol? If two different services have different istio/api versions won't this generate an invalid quota spec?

geeknoid:
@howardjohn This is a breaking change for protos, but this is not used as a proto but strictly to parse customer-authored YAML. The specific case is for a quota token count which would is typically a value of 1 or 2, never >2^31.

I'm OK with this change given it'll fix some downstream problems we're having with serializing of 64 integers which are coming out as strings (due to the proto->json serialization standard)

howardjohn:
Isn't this sent over MCP as a proto?

jasonwzm:
@howardjohn int32 and int64 should be wire compatible: https://developers.google.com/protocol-buffers/docs/proto#updating

howardjohn:
Thanks, I'll need to bookmark that page, very useful

PR #1133, opened 2019-10-17
Updated paths for networking docs.

geeknoid:


frankbu:
Should we also physically move the files up in the istio/api repo so the doc structure matches the source structure? The v1alpha3 dir isn't needed anymore. It was only needed when we changed the API and were temporarily supporting two APIs.

rshriram:
oh we can't.. if we move, we have to change a ton of code. Not to mention everyone who has imported the istio/api as a library internally to generate these API crds, for their CI/CD systems.

PR #1139, opened 2019-10-18
Disable circuit breaking by default

dgn:
This updates ConnectionPoolSettings comments according to https://github.com/istio/istio/pull/17346

### DIFF
--- a/networking/v1alpha3/destination_rule.gen.json
+++ b/networking/v1alpha3/destination_rule.gen.json
@@ -329,7 +329,7 @@
         "type": "object",
         "properties": {
           "maxConnections": {
-            "description": "Maximum number of HTTP1 /TCP connections to a destination host. Default 1024.",
+            "description": "Maximum number of HTTP1 /TCP connections to a destination host. Default 2^32-1.",
             "type": "integer",
             "format": "int32"
           },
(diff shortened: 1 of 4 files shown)


howardjohn:
Somewhat related, would be useful is if we could provide a default
destination rule that is merged, the  regardless of what defaults we set
folks could set global defaults. For cases where a central team controls
the istio config it's less useful, but for other setups where application
teams are configuring istio it would allow admins to set their own defaults

Stono:
I'm in favour of removing default configuration like this.  The defaults caught us out too.

Out of curiosity, why are things like `http2MaxRequests` being set to int32 max (2^32-1) rather than simply disabled by default?

dgn:
> I'm in favour of removing default configuration like this. The defaults caught us out too.
> 
> Out of curiosity, why are things like `http2MaxRequests` being set to int32 max (2^32-1) rather than simply disabled by default?

Envoy doesn't have an option to disable connection pool limits, you can only set really high limits to achieve a similar result, see this FAQ entry: https://www.envoyproxy.io/docs/envoy/latest/faq/disable_circuit_breaking#is-there-a-way-to-disable-circuit-breaking

jwendell:
Code changes were merged in istio/istio. Can this be merged too?

PR #1140, opened 2019-10-21
Fix default doc for idle timeout

howardjohn:
This was changed in https://github.com/envoyproxy/envoy/pull/8629


howardjohn:
@rshriram

PR #1141, opened 2019-10-21
Add locality load balancer settings on destination rule

aattuluri:
API changes for https://github.com/istio/istio/issues/17802

**Note**
I had to run `protolock commit --force` because I had to move `LocalityLoadBalancerSetting` message to `networking/destination_rule.proto` and then import it in `mesh/config.proto`, to avoid a cyclic dependency between `mesh/config.proto` and `networking/destination_rule.proto`.

_Please note that this is not backward incompatible since there is no change in structure of the message being used (`LocalityLoadBalancerSetting`)_


aattuluri:
@rshriram @hzxuzhonghu @howardjohn

networking/v1alpha3/destination_rule.proto
```diff
@@ -194,6 +194,11 @@ message TrafficPolicy {
   // overridden by port-level settings, i.e. default values will be applied
   // to fields omitted in port-level traffic policies.
   repeated PortTrafficPolicy port_level_settings = 5;
+
+  //locality load balancer settings, this will override mesh wide settings in entirety, meaning no merging would be performed
```
hzxuzhonghu: `// Locality `

networking/v1alpha3/destination_rule.proto
```diff
@@ -194,6 +194,11 @@ message TrafficPolicy {
   // overridden by port-level settings, i.e. default values will be applied
   // to fields omitted in port-level traffic policies.
   repeated PortTrafficPolicy port_level_settings = 5;
+
+  // Locality load balancer settings, this will override mesh wide settings in entirety, meaning no merging would be performed
+  // between this object and the object one in MeshConfig
+  LocalityLoadBalancerSetting locality_lb_setting = 6;
```
rshriram: shouldn't this also be under the port rraffic policy?

aattuluri: @rshriram Yes, good catch. Made the change.

networking/v1alpha3/destination_rule.proto
```diff
@@ -186,6 +186,10 @@ message TrafficPolicy {
 
     // TLS related settings for connections to the upstream service.
     TLSSettings tls = 5;
+
+    // Locality load balancer settings, this will override mesh wide settings in entirety, meaning no merging would be performed
+    // between this object and the object one in MeshConfig
+    LocalityLoadBalancerSetting locality_lb_setting = 6;
```
hzxuzhonghu: What is the relation with `LoadBalancerSettings ` in line 179? How about moving into it?

rshriram: That is a good point. Why can’t we move this under load balancer setting rather than have a separate one ?

aattuluri:
@rshriram @hzxuzhonghu Done.

hzxuzhonghu:
lgtm

aattuluri:
@rshriram can we please merge this?

hzxuzhonghu:
awesome !

PR #1036, opened 2019-08-13
Auto Generate Structural Schema for CRDs

jasonwzm:
We added functionality to autogenerate structural schema based on our protos in Cue generator. This PR is to generate the CRDs for the first time. The generated file is at kubernetes/customresourcedefinitions.gen.yaml. 

With the above change, the generated json files are also structural which still align with the openapi schema standard.

Also removed the cue image in the Makefile (now integrated with the build-tool image).

PR #1142, opened 2019-10-22
Add CRD structural schema generation

jasonwzm:
This supersedes https://github.com/istio/api/pull/1036, as there are so many conflicts in that branch and it is easier to just create a new one. The goal is still the same.

From the previous PR:
We added functionality to autogenerate structural schema based on our protos in Cue generator. This PR is to generate the CRDs for the first time. The generated file is at kubernetes/customresourcedefinitions.gen.yaml.

With the above change, the generated json files are also structural which still align with the openapi schema standard.


jasonwzm:
@ayj, can you help take a look? Oz had a chance to review the previous PR.

cue.yaml
```diff
@@ -39,3 +39,636 @@ all:
   title: All Istio types.
```
geeknoid: Is this huge file hand-written or generated?

jasonwzm: Hand written for now. I plan to move to a comment annotation based generation. So we don’t have to have such a file. This one here is temporary to keep CRDs generated.

Makefile.core.mk
```diff
@@ -453,14 +454,25 @@ all_openapi := \
 	$(security_v1beta1_openapi) \
 	$(type_v1beta1_openapi)
 
+all_openapi_crd := kubernetes/customresourcedefinition.gen.yaml
+
 $(all_openapi): $(all_protos)
 	@$(cue) -f=$(repo_dir)/cue.yaml
 
+# The fields are added at the end to generate snake cases. This is a temporary solution to accommodate some wrong namings currently exist.
```
ayj:
```suggestion
# The fields are added at the end to generate snake cases. This is a temporary solution to accommodate some wrong namings that currently exist.
```

cue.yaml
```diff
@@ -39,3 +39,636 @@ all:
...
+        name: virtualservices.networking.istio.io
+        labels:
+          app: istio-pilot
+          chart: istio
+          heritage: Tiller
+          release: istio
+        annotations:
+          "helm.sh/resource-policy": keep
+      spec:
+        preserveUnknownFields: false
```
ayj: Per https://github.com/istio/istio/pull/18188, won't this break k8s < 1.15?

kubernetes/customresourcedefinitions.gen.yaml
```diff
@@ -0,0 +1,5099 @@
...
+              description: The revision of this document.
+              format: string
+              type: string
+          type: object
+      type: object
+  versions:
+  - name: v1alpha2
+    served: true
+    storage: true
+status:
```
ayj: Is there anyway to prevent cue from adding the `status` subresource in the generated CRD definition? Not sure if this will cause problems when applied.

cue.yaml
jasonwzm: Yes, removing this. Thanks!

kubernetes/customresourcedefinitions.gen.yaml
jasonwzm: John raised similar question. Currently it is printing the json output of the CRD API directly. I will make the improvement in cue generator. Created https://github.com/istio/istio/issues/18259

PR #1148, opened 2019-10-25
Add TLSRequirementType to allow trusted lb to conditionally terminate…

lambdai:
… tls

We may use EnvoyFilter as workaround but IMHO this should be promote to gateway api.

Fix https://github.com/istio/istio/issues/16778


lambdai:
Will run script to generate the rest of the files so that the review is easier.

lambdai:
close for now in favor of another solution

PR #1150, opened 2019-10-29
Add support for arbitrary extensions to Gateways

rshriram:
Adding an opaque Any field to the Gateway servers. This
will be completely ignored by Pilot unless someone compiles
a custom plugin into Pilot to handle the OnOutboundListener() calls.
While EnvoyFilter can partially meet the needs of custom extensions, this
technique offers a more structured experience where the user can configure
a high level Gateway server and not have to worry about Envoy level features.
Also, EnvoyFilters have certain limitations due to the proto merge semantics, that
can be overcome by the custom extensions. This is in similar spirit to CoreDNS plugins.


rshriram:
cc @louiscryan per our earlier discussions. The nice part is that this does not require ANY changes in Pilot code whatsoever. All the necessary functionality to support this idea is already present.

rshriram:
A simple example:

```yaml
Kind: Gateway
spec:
   selector:
       istio: ingressgateway
   servers:
   - port:
         number: 443
         name: https
         protocol: HTTPS
      hosts:
      - api.example.com
      tls:
         mode: SIMPLE
         credentialName: api-secret
      extensions:
      - "@type": "type.googleapis.com/my.internal.authN.validator"
         jwt:
             issuer: https://jwts.example.com
             allowedScopes:
             - query
             - update
      - "@type": "type.googleapis.com/my.internal.ratelimit.processor"
         config:
             perSessionLimit: 100
             perAPIKeyLimit: 1000
   - port:
         number: 443
         name: https-web
         protocol: HTTPS
      hosts:
      - www.example.com
      tls:
         mode: SIMPLE
         credentialName: www-secret
```

costinm:
Can we keep Sidecar consistent ?

PR #1151, opened 2019-10-29
Add RequestAuthetication API

diemtvu:
Add v1beta1 API to configure request authentication. Only JWT is supported at the moment.

Issues: https://github.com/istio/istio/issues/18399


diemtvu:
Ping

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,112 @@
...
+// $aliases: [/docs/reference/config/security/v1beta1/jwt.html]
+
+// JSON Web Token (JWT) token format for authentication as defined by
+// [RFC 7519](https://tools.ietf.org/html/rfc7519). See [OAuth 2.0](https://tools.ietf.org/html/rfc6749) and
+// [OIDC 1.0](http://openid.net/connect) for how this is used in the whole
+// authentication flow.
+//
+// For example:
+//
+// A JWT for any requests:
```
smawson: What do you mean by "for any requests"?

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,112 @@
...
+// - bookstore_android.apps.googleusercontent.com
+//   bookstore_web.apps.googleusercontent.com
+// jwksUri: https://example.com/.well-known/jwks.json
+// ```
+message JWT {
+  // Identifies the issuer that issued the JWT. See
+  // [issuer](https://tools.ietf.org/html/rfc7519#section-4.1.1)
+  // Usually a URL or an email address.
+  //
+  // Example: https://securetoken.google.com
```
smawson: Add a non-Google example?

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,112 @@
...
+  //
+  // Example: `https://www.googleapis.com/oauth2/v1/certs`
+  //
+  // Note: Only one of jwks_uri and jwks should be used.
+  string jwks_uri = 3;
+
+  // JSON Web Key Set of public keys to validate signature of the JWT.
+  // See https://auth0.com/docs/jwks.
+  //
+  // Note: Only one of jwks_uri and jwks should be used.
```
smawson: Document what happens if both are set.

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,112 @@
...
+  string jwks = 10;
+
+  // Two fields below define where to extract the JWT from an HTTP request.
```
smawson:
```suggestion
  // The next two fields define how to extract the JWT from an HTTP request.
```

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,112 @@
...
+  //
+  // If no explicit location is specified the following default
+  // locations are tried in order:
+  //
+  //     1) The Authorization header using the Bearer schema,
+  //        e.g. Authorization: Bearer <token>. (see
+  //        [Authorization Request Header
+  //        Field](https://tools.ietf.org/html/rfc6750#section-2.1))
+  //
+  //     2) `access_token` query parameter (see
```
smawson:
```suggestion
  //     2) The `access_token` query parameter (see
```

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,112 @@
...
+  //     [URI Query Parameter](https://tools.ietf.org/html/rfc6750#section-2.3))
+
+  // JWT is sent in a request header. `header` represents the
```
smawson: The comment on jwt_headers is confusing, can you reword? What is `header`? The field is called jwt_headers. I assume that is just a repeated list of header names?

Also you should start this comment with: "The JWT is extracted from a request header."

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,112 @@
...
+  // header name.
+  //
+  // For example, if `header=x-goog-iap-jwt-assertion`, the header
+  // format will be x-goog-iap-jwt-assertion: <JWT>.
+  repeated string jwt_headers = 6;
+
+  // JWT is sent in a query parameter. `query` represents the
```
smawson: Same comment here, what is `query`? The field name is jwt_params.

"The JWT is extracted from query parameters."

Also in both of these you should specify the order of extraction since they are repeated fields.

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+import "type/v1beta1/selector.proto";
+import "security/v1beta1/jwt.proto";
+
+package istio.security.v1beta1;
+
+option go_package="istio.io/api/security/v1beta1";
+
+// $schema: istio.security.v1beta1.RequestAuthentication
+// $title: RequestAuthentication
+// $description: Request authentication configuration for an workloads.
```
smawson:
```suggestion
// $description: Request authentication configuration for workloads.
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+// $location: https://istio.io/docs/reference/config/security/request_authentication.html
+// $aliases: [/docs/reference/config/security/v1beta1/request_authentication.html]
+
+// RequestAuthentication defines what request authentication methods should be supported by an workload.
```
smawson:
```suggestion
// RequestAuthentication defines what request authentication methods are supported by a workload.
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+// If a request carries a token at the location recognized by one of the rules in the configuration,
```
smawson: This is overly specific, it belongs in the JWT file.

Here you can just say that authentication policy will reject a request if the request contains invalid authentication information, based on the configured authentication rules.

If you need to use a JWT example, say it is an example. "For example, if a JWT rule is specified and the JWT token is invalid, the request will be rejected."

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+// that token must be valid in order for a request to be accepted.
+// On the other hand, a request that doesn't haven any token is also accepted but will not have any
+// authenticated identity. To restrict access to authenticated requests only, this should be accompanied
+// with an authorization rule.
```
smawson:
```suggestion
// by an authorization rule.
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+// $schema: istio.security.v1beta1.RequestAuthentication
+// $title: RequestAuthentication
+// $description: Request authentication configuration for an workloads.
+// $location: https://istio.io/docs/reference/config/security/request_authentication.html
+// $aliases: [/docs/reference/config/security/v1beta1/request_authentication.html]
+
+// RequestAuthentication defines what request authentication methods should be supported by an workload.
+// If a request carries a token at the location recognized by one of the rules in the configuration,
+// that token must be valid in order for a request to be accepted.
+// On the other hand, a request that doesn't haven any token is also accepted but will not have any
```
smawson:
```suggestion
// A request that does not contain any authentication credentials will be accepted but will not have any
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+// authenticated identity. To restrict access to authenticated requests only, this should be accompanied
+// with an authorization rule.
+// Examples:
+//
+// - Require JWT for all request for workspace `httpbin`
```
smawson: What is a workspace? This is all workloads labelled with app=httpbin from what I can tell.

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+//  selector:
+//    matchLabels:
+//      app: httpbin
+//  rules:
+//  - from:
+//    - source:
+//        requestPrincipals: ["*"]
+// ```
+//
+// - The next example shows how to set different JWT requirement for different `host`. The `RequestAuthentication`
```
smawson:
```suggestion
// - The next example shows how to set a different JWT requirement for a different `host`. The `RequestAuthentication`
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+// declares it can accpet JWT issuer by either `issuer-foo` and `issuer-bar` (the public key set are implicted
```
smawson:
```suggestion
// declares it can accept JWTs issued by either `issuer-foo` or `issuer-bar` (the public key set is implicitly
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+// set with OpenID Connect spec).
```
smawson:
```suggestion
// set from the OpenID Connect spec).
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+//      hosts: ["example.com"]
+//  - from:
+//    - source:
+//        requestPrincipals: ["issuer-bar/*"]
+//    to:
+//      hosts: ["another-host.com"]
+// ```
+//
+// - You can fine tune the authorization policy to set different requirement per path. For example,
+// to require JWT on all paths, except /healthz, the same `RequestAuthentication` can be used, but the
```
smawson: The policy below matches on /, there is nothing for /healthz.

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+//  rules:
+//  - from:
+//    - source:
+//        requestPrincipals: ["*"]
+//  - to:
+//    - operation:
+//        paths: ["/]
+//
+message RequestAuthentication {
+  // Optional. Workload selector decides where to apply the RequestAuthentication policy.
```
smawson:
```suggestion
  // Optional. The selector determines the workloads to apply the RequestAuthentication on.
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+  // If not set, the policy will be applied to all workloads in the
+  // same namespace as the policy.
+  istio.type.v1beta1.WorkloadSelector selector = 1;
+
+  // Define the list of JWTs that can be validated at the selected workloads' proxy.
```
smawson: Not just validated, also extracts the authenticated user from the JWT.

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+  // Each rule will be activated only when a token is presented at the location recorgnized by the
+  // rule. The token will be validated based on the JWT rule config. If validation fails, request will
```
smawson:
```suggestion
  // rule. The token will be validated based on the JWT rule config. If validation fails, the request will
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+  // be rejected.
+  // Note: while it is supported to validate all tokens if more than one prensented (at different locations
```
smawson:
```suggestion
  // Note: if more than one token is presented (at different locations), they will all be validated
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+  // but still recognized by one of the rules), the output principal is nondeterministic.
```
smawson:
```suggestion
  // but the output principal is nondeterministic.
```

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,141 @@
...
+  repeated JWT jwt_rules = 2;
+}
```
smawson: nit: newline to make review tool happy

diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/request_authentication.proto
diemtvu: Done

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+// - "x-goog-iap-jwt-assertion"
+// ```
+message JWT {
+  // Identifies the issuer that issued the JWT. See
+  // [issuer](https://tools.ietf.org/html/rfc7519#section-4.1.1)
+  // Usually a URL or an email address.
+  //
+  // Example: https://foobar.auth0.com
+  // Example: 1234567-compute@developer.gserviceaccount.com
+  string issuer = 1 [(google.api.field_behavior) = REQUIRED];
```
costinm: The RFC defines a JSON format to represent JWT tokens - and uses the key 'iss' to represent the issuer.

Any reason for our representation of the JWT to use a different key ( and effectively create an incompatible
representation of the JWT ) ? Can you rename all the fields in the JWT to match the ones in the RFC, to 
avoid confusion and learning pain ?

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+// ```yaml
+// issuer: https://example.com
+// jwksUri: https://example.com/.secret/jwks.json
+// jwtHeaders:
+// - "x-goog-iap-jwt-assertion"
+// ```
+message JWT {
+  // Identifies the issuer that issued the JWT. See
+  // [issuer](https://tools.ietf.org/html/rfc7519#section-4.1.1)
+  // Usually a URL or an email address.
```
costinm: Never saw a JWT with email address for issuer. You mean domain ?

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+
+// $schema: istio.security.v1beta1.JWT
+// $title: JWT
+// $description: JWT configuration.
+// $location: https://istio.io/docs/reference/config/security/jwt.html
+// $aliases: [/docs/reference/config/security/v1beta1/jwt.html]
+
+// JSON Web Token (JWT) token format for authentication as defined by
+// [RFC 7519](https://tools.ietf.org/html/rfc7519). See [OAuth 2.0](https://tools.ietf.org/html/rfc6749) and
+// [OIDC 1.0](http://openid.net/connect) for how this is used in the whole
```
costinm: Would it make sense to optimize the API for the OIDC case ?

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+option go_package="istio.io/api/security/v1beta1";
+
+// $schema: istio.security.v1beta1.JWT
+// $title: JWT
+// $description: JWT configuration.
+// $location: https://istio.io/docs/reference/config/security/jwt.html
+// $aliases: [/docs/reference/config/security/v1beta1/jwt.html]
+
+// JSON Web Token (JWT) token format for authentication as defined by
+// [RFC 7519](https://tools.ietf.org/html/rfc7519). See [OAuth 2.0](https://tools.ietf.org/html/rfc6749) and
```
costinm: https://swagger.io/specification/#securitySchemeObject - can you also link to the OpenAPI spec, and maybe add some comments explaining the differences in terms ? Do we need to use different keywords and schema from OpenAPI ? K8S is using openAPI quite a bit - it would be quite convenient for users if Istio used a subset/superset that is compatible with OpenAPI.

diemtvu: For now, this API is very limited for JWT validation, and I'm not sure referring to OpenAPI here make sense. If I understand correctly, the scheme you mentioned provides generic security API, covers both authN and authZ. If we want to go that way, we need to make change across the board and also support it fully.

security/v1beta1/jwt.proto
diemtvu: Google service account is one example.

security/v1beta1/jwt.proto
diemtvu: Our API instructs how to validate a JWT, so I'm not sure if it is necessary to use the exact keys used in RFC (which I believe use the abbreviation to make it compact). Mapping between these two are somewhat trivial, and have a better readability for the API, imo.

smawson: Agree we don't need to match exactly, since this is defining validation not defining JWTs.

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+  //
+  //     1) The Authorization header using the Bearer schema,
+  //        e.g. Authorization: Bearer <token>. (see
+  //        [Authorization Request Header
+  //        Field](https://tools.ietf.org/html/rfc6750#section-2.1))
+  //
+  //     2) The `access_token` query parameter (see
+  //     [URI Query Parameter](https://tools.ietf.org/html/rfc6750#section-2.3))
+
+  // List of header locations from which JWT is expected. For example, below is the location spec if
```
smawson:
```suggestion
  // List of header locations from which JWT is expected. For example, below is the location spec
```

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+
+// This message specifies a header location to extract JWT token.
+message JwtHeader {
+  // The HTTP header name.
+  string name = 1 [(google.api.field_behavior) = REQUIRED];
+
+  // The value prefix. The value format is "value_prefix<token>"
+  // For example, for "Authorization: Bearer <token>", value_prefix="Bearer " with a space at the
+  // end.
+  string value_prefix = 2;
```
smawson: Why is this vaue_prefix and not just prefix?

If value_prefix, in the above example need to change it to valuePrefix: "Bearer "

But I don't see the value (hah) of having value_prefix and not just prefix.

diemtvu: Done. Drop `value`.

security/v1beta1/jwt.proto
costinm: As issuer ? For normal accounts it's "iss": "accounts.google.com"
(https://developers.google.com/identity/protocols/OpenIDConnect) 

Maybe worth adding some comment about this - any doc on how the email address of the issuer is generated or how it should be checked ? Usually "sub" is an email, and issuer 
a URI or domain.

I spent quite a few years working with OIDC and JWTs and I'm confused by email - please link more docs, doubt many people will know how to use it...

security/v1beta1/jwt.proto
costinm: My concern is using the same keywords that standards are using instead of inventing an istio specific vocabulary. 

However - OpenAPI and the OIDC discovery document is using 'issuer' as well, so maybe just linking to those standards so people can put it into context.

security/v1beta1/jwt.proto
costinm: That's not what I'm suggesting. Just to use the same words that OpenAPI is using, so someone familiar with OpenAPI can easily understand.

This API is indeed a subset of OpenAPI - which obviously does far more ( it defines the entire API semantics, not only the AuthN). But the section of OpenAPI related to AuthN is very relevant and similar with this API, so if we have to use different terms and semantics it would be great to have an explanation of why our words are better.

K8S is using OpenAPI extensively, and many others.

costinm:
For more context: in Istiod we are starting to use JWT tokens and OIDC to generate certs and finally authenticating workloads to Pilot. So I did a bit of research - in particular around JWT tokens used in K8S.

In networking, we are attempting to adopt 'upstream' APIs - like EndpointSlice, K8S Gateway, etc - and reduce the istio-specific APIs if a broader API exists. 

On the subject of 'iss' for example, OIDC does actually require https:// in the iss: field - there is a note https://openid.net/specs/openid-connect-core-1_0.html#GoogleIss  that indicates google is an exception with the "accounts.google.com", but email issuers are pretty clearly not supported by OIDC.

K8S JWT tokens (at least on GKE) are using proper OIDC, and look like:

```
{
  "issuer": "https://container.googleapis.com/v1/projects/costin-istio/locations/us-central1-a/clusters/istiod-single",
  "jwks_uri": "https://container.googleapis.com/v1/projects/costin-istio/locations/us-central1-a/clusters/istiod-single/jwks",
  "response_types_supported": [
    "id_token"
  ],
  "subject_types_supported": [
    "public"
  ],
  "id_token_signing_alg_values_supported": [
    "RS256"
  ],
  "claims_supported": [
    "iss",
    "sub",
    "kubernetes.io"
  ],
  "grant_types": [
    "urn:kubernetes:grant_type:programmatic_authorization"
  ]
}
```

While it is possible to support or invent new schemes - it would be ideal if we had simple and clear 
support for OIDC, ID_TOKENS and OpenAPI ( which are pretty broadly adopted standards, including
in K8S ).

I understand some arbitrary extensions and schemes may need to be supported - but I think we have
more than enough confusing options in the APIs: can we have a clean and simple API that follows 
OIDC + OpenAPI patterns/vocabulary, and separate extensions or APIs for other standards ? 

I'm sorry I missed the design doc for this proposal, but since it's a 'beta' API I think it's important to get it right...

diemtvu:
I'm not really understand your point. Can you give an example for your proposed API?

So far, this API contains all the essential parameters to do basic JWT validation (verifying signature, check timestamp, audience). Afaik, it should work with all types of JWT.

security/v1beta1/jwt.proto
diemtvu: Rephrase the comment. We do not dictate what value the issuer must be. We simply use it to verify the issuer from the token matches with the intended one.

security/v1beta1/jwt.proto
diemtvu: Already is. All fields, except issuer are optional.

diemtvu:
@costinm Can you see if this can be unblocked based on our offline discussion. Tkx

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+// This example specifies token in non-default location (`x-goog-iap-jwt-assertion` header). It also
+// defines the URI to fetch JWKS explicitly.
+//
+// ```yaml
+// issuer: https://example.com
+// jwksUri: https://example.com/.secret/jwks.json
+// jwtHeaders:
+// - "x-goog-iap-jwt-assertion"
+// ```
+message JWT {
```
costinm: JWT is usually the name given to the token ( jwt token ) - can we call this proto something else ? 
JWTPolicy, JWTAuthConfig ?

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+  // Example: https://foobar.auth0.com
+  // Example: 1234567-compute@developer.gserviceaccount.com
+  string issuer = 1 [(google.api.field_behavior) = REQUIRED];
+
+  // The list of JWT
+  // [audiences](https://tools.ietf.org/html/rfc7519#section-4.1.3).
+  // that are allowed to access. A JWT containing any of these
+  // audiences will be accepted.
+  //
+  // The service name will be accepted if audiences is empty.
```
costinm: What does it mean ( service name will be accepted )? 

I assume it means if audiences is empty the token must have an audience matching the service name ? 
And which service name  - I assume we won't support  unqualified ( without namespace at least). 

Please be very clear about this - most APIs allow unqualified service name ( productpage ), in this case it would
be a big security problem.

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+  //
+  // Example:
+  //
+  // ```yaml
+  // audiences:
+  // - bookstore_android.apps.googleusercontent.com
+  //   bookstore_web.apps.googleusercontent.com
```
costinm: Can use use example.com ?

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+  // Example: `https://www.googleapis.com/oauth2/v1/certs`
+  //
+  // Note: Only one of jwks_uri and jwks should be used. jwks_uri will be ignored if it does.
+  string jwks_uri = 3;
+
+  // JSON Web Key Set of public keys to validate signature of the JWT.
+  // See https://auth0.com/docs/jwks.
+  //
+  // Note: Only one of jwks_uri and jwks should be used. jwks_uri will be ignored if it does.
+  string jwks = 10;
```
costinm: Can we hide this - or confirm it is a requirement and if it is document how this can be maintained ( key rotation, etc).

I expect most open id providers to support the URI - we have enough troubles with managing public keys.

security/v1beta1/jwt.proto
```diff
@@ -0,0 +1,141 @@
...
+  repeated JWTHeader from_headers = 6;
+
+  // List of query parameters from which JWT is expected. For example, if JWT is provided via query
+  // parameter `my_token` (e.g /path?my_token=<JWT>), the config is:
+  // ```
+  //   fromParams:
+  //   - "my_token"
+  // ```
+  repeated string from_params = 7;
+}
```
costinm: Can you add a comment on the behavior of 'forward' ( i.e. if workload will receive the token ) and 'forward_payload_header' ( if workload gets the verified decoded token from envoy ) ? 

At least one of them must either be default or configurable - otherwise the apps have no way to find the
identity of the caller.

costinm:
And looks good otherwise.

security/v1beta1/jwt.proto
costinm: Great.

security/v1beta1/jwt.proto
costinm: Actually - JWTRule seems the right name ( since the field is jwt_rules )

diemtvu: Done. Change to `JWTRule`

security/v1beta1/jwt.proto
diemtvu: Rewrite the comment. It was some cut and paste error. In short, if this field is not set, we will not check the audience claim.

security/v1beta1/jwt.proto
diemtvu: Done

security/v1beta1/jwt.proto
diemtvu: I was thinking to delay the decision how to deal with this. A single forward behavior may be not enough. So I've just added new fields to control how to forward the token, inline with what can be supported by Envoy API at the moment.

security/v1beta1/jwt.proto
diemtvu: Add comment to mention the key will be used as is. This can be used for the case Istio cannot access the jwks_uri. User will responsible to rotate the key.

security/v1beta1/jwt.proto
costinm: Well, I assume if the URL is not set, the issuer is also used to download the OIDC discovery and get the public keys - which will work only for URL. 

So maybe mention of issuer is not a URL, user must supply the key URL.

security/v1beta1/jwt.proto
costinm: Hmm, are you sure this is what user will expect ? I guess since we can no longer know the service names (since the rule is on the workload ) - there is no choice.

security/v1beta1/jwt.proto
costinm: How ? It's not as simple - just like saying user is responsible for rotating Citadel key is not enough. Will there be traffic loss ? With the URL the impl can probably automate rotations.

And AFAIK JWT public keys are rotated quite frequently, and would be dangerous not to do so. 

So until we can tell users how to do it - can we hide or remove this option ?

security/v1beta1/jwt.proto
costinm: I'm ok to not add all fields - or just document a default behavior. 

Also ok with adding fields to match envoy API. 

Either way - I think it's critical for the app to know the user, many apps won't work otherwise (including Pilot/istiod if it sits behind an Envoy - we need the user identity)

diemtvu: I've added options similar to Envoy API.

security/v1beta1/jwt.proto
diemtvu: Add more comment to recommend users to use `jwks_uri` when possible.

diemtvu:
@howardjohn Can you sign off on behalf of Test & Release WG. Tkx

security/v1beta1/authorization.proto.orig
```diff
@@ -0,0 +1,277 @@
...
+  repeated string methods = 3;
+
+  // Optional. A list of paths, which matches to the "request.url_path" attribute.
+  //
+  // If not set, any path is allowed. Must be used only with HTTP.
+  repeated string paths = 4;
+}
+
+// Condition specifies additional required attributes.
+message Condition {
```
howardjohn: what is this file?

diemtvu: Oh, good catch. Removed.

diemtvu:
ping

security/v1beta1/request_authentication.proto
```diff
@@ -0,0 +1,139 @@
...
+//        requestPrincipals: ["issuer-bar/*"]
+//    to:
+//      hosts: ["another-host.com"]
+// ```
+//
+// - You can fine tune the authorization policy to set different requirement per path. For example,
+// to require JWT on all paths, except /healthz, the same `RequestAuthentication` can be used, but the
+// authorization policy could be:
+//
+// apiVersion: security.istio.io/v1beta1
```
howardjohn: missing the ```yaml

PR #1152, opened 2019-11-01
Regen based on new container

jasonwzm:
- `.json` becomes `.gen.json`
- `_json.gen.go` has `MarshalJSON` and `UnmarshalJSON` for all proto messages
- customresourcedefinitions generated file does not output status field
- updated license file


nrjpoddar:
@jasonwzm @geeknoid we need this cherry-picked into release-1.4 or the existing release-1.4 needs to have the marshalers regenerated so that istio.io/client-go release-1.4 can point to the appropriate release-1.4 SHA from api repo.

jasonwzm:
See https://github.com/istio/tools/pull/501 for release 1.4 tools. Will regen once the build tool is updated.

PR #1156, opened 2019-11-04
Add support of http retry priority

ckcd:
Add parameter for `envoy.retry_priorities.previous_priorities` in `retry_priority`.

See https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#arch-overview-http-retry-plugins

related issue: https://github.com/istio/istio/issues/18418


networking/v1alpha3/virtual_service.proto
```diff
@@ -1006,6 +1006,11 @@ message HTTPRetry {
   // See the [retry policies](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on)
   // and [gRPC retry policies](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-grpc-on) for more details.
   string retry_on = 3;
+
+  // This will target priorities in subsequent retry attempts that haven’t been already used,
+  // it's value decides how often the priority load should be recalculated.
```
howardjohn: having read this comment I would have no clue what to set this value to.

Also "priority" is not an istio concept, maybe should rename this?

ckcd: > Also "priority" is not an istio concept, maybe should rename this?

what about `update_endpoint_group` ? or just named something like `update_locality`?

howardjohn: I'm bad at naming things... maybe failover? locality_failover? 

and can it be a bool? I'm not sure what int is supposed to be

ckcd: > I'm not sure what int is supposed to be

the `int` indicated the `update_frequency` below:

```
  retry_priority:
    name: envoy.retry_priorities.previous_priorities
    config:
      update_frequency: 2
```

> locality_failover

ok, I will using this name and make it bool if `int` is unnecessary.

howardjohn: what does "how often mean"? seconds? number of request? hopefully we can set a sane default and make user not have to configure this

ckcd: Cool, will do this change~

howardjohn: > This will target priorities in subsequent retry attempts that haven’t been already used. The update_frequency parameter decides how often the priority load should be recalculated.

Even the envoy docs don't really explain what it means. but does seem like we can probably pick a default

ckcd: > Even the envoy docs don't really explain what it means. but does seem like we can probably pick a default

yes, a default value is good enough.

already do the change, please review again, thanks !

ckcd:
Seems like we need a `ok-to-test` label. @howardjohn

ckcd:
@howardjohn Can we merge this ? Thanks.

howardjohn:
this still needs approval from @rshriram

ckcd:
@rshriram PTAL.

ckcd:
@rshriram Please take a look, thanks a lot~

networking/v1alpha3/virtual_service.proto
```diff
@@ -1006,6 +1006,10 @@ message HTTPRetry {
   // See the [retry policies](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on)
   // and [gRPC retry policies](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-grpc-on) for more details.
   string retry_on = 3;
+
+  // Flag to specify whether the retries should retry to another locality.
+  // See the [retry plugin configuration](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#retry-plugin-configuration) for more details.
+  bool locality_failover = 4;
```
rshriram: Will this only retry to another locality or will it also retry the current one ?

rshriram: Rename to “retry_remote_localities”

ckcd: > Will this only retry to another locality or will it also retry the current one ?

Both the current one and other localities, the frequency is determined by `update_frequency` and @howardjohn and I agree that we should set a default value for it.

> Rename to “retry_remote_localities”

DONE, please review again, thanks !

rshriram: if it is both current and remote locality, then retry_remote_localities does not sound right. Isn't it more like `retry_endpoints_in_any_locality` in that case?

ckcd: Sounds good, how about change this name to `retry_any_locality` since `retry_endpoints_in_any_locality` is too long ?

howardjohn: Its not retrying any locality is it?

if we have
P0 - ep0
P1 - ep1
P2 - ep2

We will send all requests to ep0
If we retry which do we retry to? [ep0, ep1]? [ep1]? [ep1, ep2]?

Can we verify this before we merge since we should document it and it impacts the name

ckcd: I think the retry should be: (if we set `update_frequency=2`)

- normal packet: ep0
- retry 1: ep0
- retry 2: ep0
- retry 3: ep1
- retry 4: ep1
- retry 5: ep2
...

howardjohn: I checked with some envoy folks, I think thats right:

> snowp 6:02 PM
> It tracks the priorities that were hit and recomputes the priority to load to treat those priorities as empty
> There’s an update frequency you can configure to say how often to update
> So if you set it to 2 and everything is healthy
> You hit p0 p0 p1 p1 p2 p2 etc
> It takes healthy % into account, so if p1 is unhealthy it might spill over to p2 after two attempts
> John Howard 6:04 PM
> So with frequency=1 it would be p0 p1 p2. Is that the first retry is to p0 or the initial request is to p0 then the first retry is to p1?
> And just wondering, what frequency are you guys using?
> snowp 6:05 PM
> I think first retry to p1
> 2, just to preserve existing legacy behavior
> 2 seems reasonable though, it means you try another host in the same region first
> So one bad host won’t cause cross region retries

ckcd: Great!

> I think first retry to p1

So it won't retry current P0, then the current name `retry_remote_localities` is cool, right ? @howardjohn @rshriram 


> 2 seems reasonable though

Good, will set `update_frequency=2` in istio.

ckcd:
push again to solve conflict.

networking/v1alpha3/virtual_service.proto
hzxuzhonghu: I think the above desc can be added to docs too

ckcd:
Any changes should I do ? @howardjohn @rshriram

ckcd:
force push to solve conflict.

ckcd:
@rshriram PTAL.

networking/v1alpha3/virtual_service.proto
```diff
@@ -1035,6 +1035,10 @@ message HTTPRetry {
   // See the [retry policies](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on)
   // and [gRPC retry policies](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-grpc-on) for more details.
   string retry_on = 3;
+
+  // Flag to specify whether the retries should retry to other localities.
+  // See the [retry plugin configuration](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#retry-plugin-configuration) for more details.
+  bool retry_remote_localities = 4;
```
rshriram: should this be a wrapper so that we can have defaults? Otherwise, it will always default to false even if we want to default it to true in future.

rshriram:
can you convert it to a wrapper please?

networking/v1alpha3/virtual_service.proto
ckcd: Sure, will do the convert.

ckcd:
> can you convert it to a wrapper please?

Done~ PTAL. @rshriram 

thanks !

howardjohn:
@ckcd this just needs a rebase

PR #1158, opened 2019-11-05
update the RBAC and authz doc for 1.4

yangminzhu:
For istio/istio#12394


yangminzhu:
@geeknoid I couldn't find the authorization page and it seems we do have the `$location` in `authorization.proto`, I'm not sure if it's caused by invalid path used in `$location`, could you take a look? Thanks.

geeknoid:
I think this is what you're looking for: https://preliminary.istio.io/docs/reference/config/istio.authentication.v1alpha1/

Right?

geeknoid:
@yangminzhu Please cherry pick this by hand, the tool isn't able to do it automatically.

yangminzhu:
> I think this is what you're looking for: https://preliminary.istio.io/docs/reference/config/istio.authentication.v1alpha1/
> 
> Right?

No, I'm looking for a new page for authorization at  `https://istio.io/docs/reference/config/authorization/authorization-policy.html`

PR #1160, opened 2019-11-05
Add operator annotations

howardjohn:
for https://github.com/istio/istio/issues/18621

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -319,13 +319,13 @@ annotations:
     resources:
       - Any
   - name: install.operator.istio.io/owner-generation
-    description: Represents the generate to which the resource was last reconciled.
+    description: Represents the generation to which the resource was last reconciled.
     deprecated: false
     hidden: false
     resources:
       - Any
   - name: install.operator.istio.io/version
-    description: Represents the Istio version associate with the resource
+    description: Represents the Istio version associated with the resource
     deprecated: false
     hidden: false
     resources:
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -312,3 +312,21 @@ annotations:
     hidden: false
     resources:
       - Ingress
+  - name: install.operator.istio.io/chart-owner
+    description: Represents the name of the chart used to create this resource.
+    deprecated: false
+    hidden: false
+    resources:
+      - Any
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -312,3 +312,21 @@ annotations:
...
+    resources:
+      - Any
+  - name: install.operator.istio.io/owner-generation
+    description: Represents the generate to which the resource was last reconciled.
+    deprecated: false
+    hidden: false
+    resources:
+      - Any
+  - name: install.operator.istio.io/version
+    description: Represents the Istio version associate with the resource
```
geeknoid:
```suggestion
    description: Represents the Istio version associated with the resource
```

annotation/annotations.yaml
```diff
@@ -312,3 +312,21 @@ annotations:
...
     resources:
       - Ingress
+  - name: install.operator.istio.io/chart-owner
+    description: Represents the name of the chart used to create this resource.
+    deprecated: false
+    hidden: false
+    resources:
+      - Any
+  - name: install.operator.istio.io/owner-generation
+    description: Represents the generate to which the resource was last reconciled.
```
geeknoid: The generate?

howardjohn: I can't type :slightly_smiling_face: fixed

sushicw:
Would it be possible to add `func All() []*annotation.Instance` to `annotations.gen.go`? This would allow us to update the annotations analyzer to not use a hand-maintained list.

howardjohn:
> Would it be possible to add func All() []*annotation.Instance to annotations.gen.go? This would allow us to update the annotations analyzer to not use a hand-maintained list.

 probably so, sound useful. I think the code is in the tools repo

annotation/annotations.yaml
```diff
@@ -312,3 +312,21 @@ annotations:
...
+    resources:
+      - Any
+  - name: install.operator.istio.io/owner-generation
+    description: Represents the generation to which the resource was last reconciled.
+    deprecated: false
+    hidden: false
+    resources:
+      - Any
+  - name: install.operator.istio.io/version
+    description: Represents the Istio version associated with the resource
```
rshriram: nit: full stop missing.

PR #1162, opened 2019-11-07
Gen for annotations

jasonwzm:


geeknoid:
This will need an update in test-infra to the new image before it can pass CI.

PR #1175, opened 2019-11-20
Remove unknown field hosts in sidecar

ChangyuWang:
Remove unknown field hosts in sidecar


networking/v1alpha3/sidecar.proto
```diff
@@ -121,8 +120,6 @@ import "networking/v1alpha3/gateway.proto";
 //       name: egresshttp
 //     hosts:
 //     - "prod-us1/*"
-//   - hosts:
-//     - "istio-system/*"
```
howardjohn: I don't understand. Hosts is a field under egress, see https://istio.io/docs/reference/config/networking/sidecar/#IstioEgressListener. Its also on line 121 but you didn't remove that

ChangyuWang: sorry, I make a mistake, I think the hosts to removed in sidecar spec section.  Ignore it.

PR #1177, opened 2019-11-20
Refactor and integrate MeshConfig and IstioControlPlaneSpec

ostromart:
Fresh attempt at https://github.com/istio/api/pull/1101.


mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,160 @@
...
+              - path: spec.template.spec.containers.[name:galley].command.[--livenessProbeInterval]
+                value: --livenessProbeInterval=123s
+    ```
+*/
+
+// IstioOperatorSpec defines the desired installed state of Istio components.
+// The spec is a used to define a customization of the default profile values that are supplied with each Istio release.
+// Because the spec is a customization API, specifying an empty IstioOperatorSpec results in a default Istio
+// component values.
+message IstioOperatorSpec {
```
richardwxn: why we change from IstioControlPlaneSpec

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,160 @@
...
+    //     - minimal (looks in profiles dir for a file called minimal.yaml)
+    //     - /tmp/istio/install/values/custom/custom-install.yaml (local file path)
+    // default profile is used if this field is unset.
+    string profile = 10;
+    // Path for the install package. e.g.
+    //     - /tmp/istio-installer/nightly (local file path)
+    string install_package_path = 11;
+    // Root for docker image paths e.g. docker.io/istio-release.
+    // Releases are published to docker hub under 'istio' project.
+    // Daily builds from prow are on gcr.io, and nightly builds from circle on docker.io/istionightly
```
richardwxn: this part of description is outdated

mesh/v1alpha1/operator.proto
ostromart: The intention is to create a new operator API which not only includes the various parts (meshconfig, component resource definitions) but is also a starting point for adding operator specific API knobs (e.g. for upgrade). These would be only used only by operator.

mesh/v1alpha1/operator.proto
ostromart: Done.

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,302 @@
...
+    string kind = 2;
+    // Name of resource.
+    // Namespace is always the component namespace.
+    string name = 3;
+
+    // List of patches to apply to resource.
+    repeated PathValue patches = 4;
+}
+
+// Observed state of IstioControlPlane.
```
smawson: Maybe reword this comment and describe what status will contain (map from version to status for that version?).

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,158 @@
...
+import "k8s.io/api/autoscaling/v2beta1/generated.proto";
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "github.com/gogo/protobuf/protobuf/google/protobuf/wrappers.proto";
+
+package istio.mesh.v1alpha1;
+
+option go_package="istio.io/api/mesh/v1alpha1";
+
+/*
+IstioOperatorSpec is an API for managing as Istio control plane. It is used to define both profiles that are included
```
smawson:
```suggestion
IstioOperatorSpec is an API for managing an Istio control plane. It is used to define profiles that are included
```

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,158 @@
...
+with Istio releases, and user overlays, which are used to override profile values.
+
+The simplest overlay is to point the user IstioOperatorSpec profile to a different values file, for
```
smawson:
```suggestion
The simplest overlay is to point the profile to a different values file, for
```

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,302 @@
...
+    PilotComponentSpec pilot = 30;
+    ProxyComponentSpec proxy = 31;
+    SidecarInjectorComponentSpec sidecar_injector = 32;
+    PolicyComponentSpec policy = 33;
+    TelemetryComponentSpec telemetry = 34;
+    CitadelComponentSpec citadel = 35;
+    NodeAgentComponentSpec node_agent = 37;
+    GalleyComponentSpec galley = 38;
+    IngressGatewayComponentSpec ingress_gateway = 39;
+    EgressGatewayComponentSpec egress_gateway = 40;
```
rshriram: We should just have a generic gateway spec. Else people can’t add new gateways

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,302 @@
...
+    CNIComponentSpec cni = 41;
+    CoreDNSFeatureSpec coreDNS = 33;
```
rshriram: This is an extension isn’t it? How will people add additional specs if we make this a proto ?

rshriram: I think you are better off collapsing this into a generic componentSpec that has enabled bool and k8s resource spec.

mesh/v1alpha1/component.proto
ostromart: Actually it represents the map of component name to status. I've completely reworked the status and moved it to operator.proto file, ptal.

mesh/v1alpha1/operator.proto
ostromart: Done.

mesh/v1alpha1/operator.proto
ostromart: Done.

mesh/v1alpha1/component.proto
ostromart: Yes, we struggled with how to add this to the spec because user and built-in gateways map differently into the charts and there's no enable/disable concept for user gateways, either they exist or they don't. I think this new version should cover all cases.

mesh/v1alpha1/component.proto
ostromart: We discussed this in the TOC and I believe the consensus was to keep components explicitly named. The advantages of doing that are syntax validation and ability to customize each component. I agree we need some extension mechanism so I've added a map which uses resource name as the key.

mesh/v1alpha1/component.proto
```diff
@@ -31,15 +31,20 @@ message IstioComponentSpec {
...
+    NodeAgentComponentSpec node_agent = 36;
+    GalleyComponentSpec galley = 37;
+    GatewayComponentSpec ingressGateway = 38;
+    GatewayComponentSpec egressGateway = 39;
+    // Map key is the name of the resources for the component. All k8s resources accessible through
+    // KubernetesResourcesSpec must have the same name.
+    map<string, CustomGatewayComponentSpec> customGateways = 40;
+    CNIComponentSpec cni = 41;
+    CoreDNSComponentSpec coreDNS = 42;
+    IstiodComponentSpec istiod = 43;
```
howardjohn: I don't think we will have a separate component. See https://docs.google.com/document/d/1jO2oM9zdm9h7tUqOpnAG3dh1xavtW1L7zQ5LyraYRAs/edit

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,302 @@
...
+
+package istio.mesh.v1alpha1;
+
+option go_package="istio.io/api/mesh/v1alpha1";
+
+// IstioComponentSpec defines the desired installed state of Istio components.
+message IstioComponentSpec {
+    PilotComponentSpec pilot = 30;
+    ProxyComponentSpec proxy = 31;
+    SidecarInjectorComponentSpec sidecar_injector = 32;
```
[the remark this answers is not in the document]
louiscryan: Agreed. Lets try and avoid using names like pilot / galley / citadel and use functional names instead.

mesh/v1alpha1/component.proto
smawson: Is it possible to use a common proto message to start and then specialize if we need additional fields?

With proto at least it is possible to do that, not sure how well the yaml and CRD stuff would support that. Basically only introduce specific types as needed.

mesh/v1alpha1/component.proto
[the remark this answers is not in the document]
smawson: Louis and I discussed, we agreed that we'll leave these in here for now with an understanding they'll likely be deprecated in the future (realistically deprecated in 1.6 and removed in 1.7).

These are for providing break glass to the actual components so it is OK if they have some churn, they reflect the underlying implementation and should actively break the users config if those components no longer exist.

Maybe we can put some warnings on this stuff to let users know they are responsible for these changes, and that the supported feature-level configuration is over in mesh config, not in component config.

mesh/v1alpha1/component.proto
smawson: +1 for leaving this out for now.

mesh/v1alpha1/component.proto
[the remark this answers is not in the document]
costinm: What is the point of adding an API when we know even in 1.5 (where it'll ship) it won't be accurate ? 
More complexity and churn on one side - nothing on the other. We already have the break glass - values.yaml.

mesh/v1alpha1/component.proto
costinm: +1 - in 1.5 istiod will be istio pilot.

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,294 @@
...
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "github.com/gogo/protobuf/protobuf/google/protobuf/wrappers.proto";
+
+package istio.mesh.v1alpha1;
+
+option go_package="istio.io/api/mesh/v1alpha1";
+
+// IstioComponentSpec defines the desired installed state of Istio components.
+message IstioComponentSpec {
+    PilotComponentSpec pilot = 30;
```
costinm: How about map<string, ComponentSpec> ? 

And remove all the FooComponentSpec - they seem to have the same content in most cases, and when they don't - they should, because the rest of the config should be in mesh.yaml.

This will also allow multiple instances/revisions of control plane.

mesh/v1alpha1/component.proto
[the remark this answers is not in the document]
ostromart: Which part will not be accurate in 1.5?

mesh/v1alpha1/component.proto
ostromart: There is a map for extension components (see below). 
We know that these components will exist in 1.5. Making them explicit gives us syntax checking and the ability to deprecate explicitly in the API (vs. in code). 
I thought we already discussed this point in the TOC and agreed on keeping explicit component name - why are we reopening this discussion?

costinm:
Because we also discussed istiod and removing some of the components, and
the model doesn't fit with any scheme of
upgrade-by-traffic-shifting.

Defining an API for the wrong model - just to replace it in 1.6 and confuse
users - serves no purpose. You can keep the names
as operator specific, but it's really the wrong way to define an Istio
API.  Does it mean any implementation of Istio must have something
called 'pilot' or 'galley' ?  Internal names should never leak into APIs.

I also fail to see what 'syntax checking' this provides - all components,
regardless of names, should allow the same type of
resource options and patching. We don't need 20 almost identical protobufs
- and we agreed that anything that affects the real
mesh behavior should not be defined in an (internal) component name.

costinm:
I really don't understand why you insist on pushing this API. Maybe I'm
missing something - but I don't thing anyone
complained we have too few APIs or that we must standardize the internal
name and the old config - just before
we try to get rid of it.

The API does nothing to address config updates / upgrade ( traffic shift or
in-place where possible ), we're just
standardizing an install mode that was clearly wrong in the first place.

One of the main reasons for the new installer was to decouple and support
'a-la-carte' - what this API does is manages
to move us back to a monolith install equivalent with the old installer,
losing the main benefit of the refactoring
(ability to independently install and operate each component). And of
course now the number of components is
reduced and the boundaries better defined, so that part of the new
installer is also mostly wasted.

mesh/v1alpha1/component.proto
ostromart: It's possible but proto doesn't have inheritance so the user is forced to add an extra field e.g. 
```yaml
cni:
  common: <-- this gets added
    k8s:
```

mesh/v1alpha1/config.proto
```diff
@@ -382,7 +382,7 @@ message MeshConfig {
   repeated Certificate certificates = 47;
 
   // $hide_from_docs
-  // Next available field number: 49
+  // Next available field number: 53
```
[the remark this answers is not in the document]
ostromart: Done.

mesh/v1alpha1/component.proto
ostromart: Done.

ostromart:
>> I really don't understand why you insist on pushing this API.

I'm not pushing this API - it was already agreed on weeks ago. It's you who keeps pushing your own contrary view.

>> Does it mean any implementation of Istio must have something called 'pilot' or 'galley' ?  Internal names should never leak into APIs.

You're taking a general guideline (which I agree with) and turning it into a dogma (which I don't). Pilot and galley exist for better or worse at the highest level in Istio docs and the API is a representation of that. Once they disappear (in 1.6 or later) they will be removed from the API. 
I don't see the point in debating this endlessly. We don't agree, the TOC resolved the disagreement and it's time to move on.

costinm:
It was agreed to add what is needed - not this specific form. 10 protos
with almost same content to 'validate' ( whatever that means ) the obsolete
components, and usable only with operator. Why not leave it where it is, in
operator repo ?

In any case - toc is responsible to approve this, I can't hold it - but
it's a perfect example of what is wrong with Istio API process and why we
should stick to implementing APIs with better stability and process, like
k8s or envoy...

costinm:
To be clear: I don't have a problem with adding an API for patching the
config, or extending
mesh.yaml with things that are clearly needed.

But adding a dozen protos - one for each component we know we want to
deprecated in the next release, and subjecting
the user to another churn and migration/upgrade/deprecation pain - with us
knowing ahead of time that this is
not going to hold, yet still asking users to adopt it - is simply wrong.
None of this API deals with
actually operating Istio - safe or automatic config change or anything
'operation' related. It's just
a more complicated way to do what Helm was doing - all the pain of the helm
template, all
the countless old config options - plus another large set of protos on top.

A simple map keyed by component name can provide exactly the same behavior
- and validating
a the config from a map is not rocket science, proto is for serialization
and all schemas can handle
a map. At least we wouldn't launch a new API for mixer and policy in the
same release we are deprecating
them and switching to mixer v2.

ostromart:
The YAMLs the user would pass in would be 100% identical with either implementation. There is no migration cost. The difference is that with the map implementation, the user would have to guess what is a valid name for a component because it would depend on what's implemented in the binary. That's why the API, not the binary should specify the valid value and valid values should be explicitly managed as the binary changes. The API should be the contract with the user, not what's in the binary and that's why I feel a map implementation is wrong in this case.
We discussed precisely this point in the TOC at the time that istiod design was already firm so I don't buy your assertion that we agreed to anything other than exactly this, under exactly the same set of assumptions. If we reopen this debate then something with the TOC resolution process is not working.

mesh/v1alpha1/component.proto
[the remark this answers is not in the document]
howardjohn: > when will istiod become standard for our customers?

I am very confident this will land in 1.5. There are at least 3 people willing to devote 100% effort to ensuring this happens.

> when will we be confident in removing the microservices model?

Presumably istiod will be a huge success and we can just kill this in 1.6 :slightly_smiling_face: but only time can tell for this

> We can't force our customers to adopt istiod. We can certainly head in that direction, however, the outcome is uncertain.

I don't think we will have to force people. I spoke with at least ~10 users at kubecon and 10/10 were asking for istiod to come ASAP.

I don't see why we need to follow a deprecation model. "Sidecar Injector" is not a feature we offer users, sidecar *injection* is. We can remove the microservices at any time as long as we stay within the restraints of not removing functionality we offer users.

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,160 @@
...
+
+    // Config used by control plane components internally.
+    MeshConfig meshConfig = 40;
+
+    // Kubernetes resource settings, enablement and component-specific settings that are not internal to the
+    // component.
+    IstioComponentSpec components = 50;
+
+    // Overrides for default values.yaml.
+    TypeMapStringInterface values = 100;
```
howardjohn: One thing that I still have confusion with is when do I use values and when do I use components? This could use a lot more docs. I assume if it confuses me it confuses users

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,293 @@
...
+    GatewayComponentSpec egressGateway = 39;
+    // Map key is the name of the resources for the component. All k8s resources accessible through
+    // KubernetesResourcesSpec must have the same name.
+    map<string, CustomGatewayComponentSpec> customGateways = 40;
+    CNIComponentSpec cni = 41;
+    CoreDNSComponentSpec coreDNS = 42;
+
+    // Non-core components. Map key is the name of the resources for the component. All k8s resources accessible through
+    // KubernetesResourcesSpec must have the same name.
+    map<string, ExtensionComponentSpec> = 100;
```
sdake: invalid proto syntax - need a name for the map.

mesh/v1alpha1/config.proto
```diff
@@ -434,3 +439,19 @@ message Certificate {
   // multiple DNS names.
   repeated string dns_names = 2;
 }
+
+// Configuration options for traffic management.
+message TrafficManagementFeatureSpec {
+}
```
howardjohn: why are we adding empty messages?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,302 @@
...
+// Configuration options for cni component.
+message CNIComponentSpec {
+    TypeBoolValueForPB enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for CoreDNS component.
+message CoreDNSComponentSpec {
+    TypeBoolValueForPB enabled = 1;
+    string namespace = 2;
```
howardjohn: Why does this have namespace and others don;t?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,302 @@
...
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for CoreDNS component.
+message CoreDNSComponentSpec {
```
howardjohn: this is duplicated?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,293 @@
...
+    TypeIntOrStringForPB maxSurge = 2;
+}
+
+message ObjectMeta {
+    // From k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta.
+    string name = 5;
+    string namespace = 6;
+}
+
+// GOTYPE: map[string]interface{}
```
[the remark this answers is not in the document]
ostromart: Yes this is unfortunate :-/ This has been a thorn in our side for a very long time and so far no one has found a better solution  - and not for lack of trying. The issue is that there's no way to specify an interface{} in a proto. The way that k8s works around this is to have a custom unmarshaler, which we could try to do but it would be a lot of work.

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,183 @@
...
+    // Unvalidated overrides for default values.yaml. Used for custom templates where new parameters are added.
+    TypeMapStringInterface unvalidatedValues = 101;
```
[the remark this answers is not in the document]
ostromart: It's here as an escape hatch so that if users fork the charts and add their own values to the schema they can set them without having to also fork and build values_types.proto and istioctl binary. I think it's a matter of documenting the purpose of the field clearly, that this is for break the glass use.

mesh/v1alpha1/operator.proto
ostromart: Ack, we can add more examples around the docs for this. The general rule of thumb is, if it's available through components, then use components.

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,293 @@
...
+}
+
+// Configuration options for CoreDNS component.
+message CoreDNSComponentSpec {
+    TypeBoolValueForPB enabled = 1;
+    KubernetesResourcesSpec k8s = 80;
+}
+
+// Configuration options for the istiod component.
+message IstiodComponentSpec {
```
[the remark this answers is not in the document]
ostromart: Forgot to remove this.

mesh/v1alpha1/config.proto
ostromart: Already removed per prior comments.

mesh/v1alpha1/component.proto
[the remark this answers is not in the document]
ostromart: Done.

mesh/v1alpha1/component.proto
ostromart: Done.

mesh/v1alpha1/component.proto
ostromart: Already fixed.

mesh/v1alpha1/component.proto
ostromart: Fixed.

costinm:
It will not be 100% identical - user may use a component named 'mcpserver'
with a custom MCP server, "pilot-canary", etc.
Installing Istio does not require singleton components with the internal
names.

Operator doesn't exist to manage a bookinfo demo install with fixed
configuration and in-place upgrade.

The contract with the user should not be 'Istio design allow multiple
custom MCP config sources - but they must be named
galley and really be just the one we ship".

User cares about having a 'control plane' - with possibly multiple
revisions/configs ( == canary, etc) and should be able to
bring its own MCP (or any other component) if he needs to without caring
about internal implementation names.

mesh/v1alpha1/operator.proto
smawson: The long term goal is to remove values altogether right? We just have these while we're adding additional flags to components and mesh config?

We should provide additional documentation here (and in our docs) describing what this is for. Ideally we'd also have a list of what values still go here and what values should not be included here and instead handled some other way.

mesh/v1alpha1/operator.proto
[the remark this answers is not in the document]
smawson: Ideally we'd have a better model than just "unvalidatedValues" as a way to let platform owners extend our installation system. I'd like them to be able to offer type-safe configuration that is validated through a schema rather than just having a blob of additional values. Not-blocking for this PR but we really should think through what we want that to look like.

mesh/v1alpha1/component.proto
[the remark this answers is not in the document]
smawson: This thread is probably too long for a review discussion and should move to email, but a couple quick points anyway:

1) istiod being available != microservices being deprecated. Done anyone actually think we're going to deprecate microservices deployment mode in 1.5? Before we deprecate something the replacement needs to be in production and stable with enough users to be confident, I don't see that happening with istiod before 1.5. We don't want to tie the operator to istiod (just as we decided not to tie SDS to istiod). We generally try to avoid blocking ourselves, even if it costs us technical debt we need to clean up later.

2) Simply disagreeing with an API is not useful. Much more useful is providing an alternative proposal that meets all of the requirements that the API is trying to address (and ideally additional requirements). So if there is a proposal for an operator API that works better for some use cases, great, please write up a doc with a proposal and discuss in environments WG. We have until 1.5 ships to make this better so you have time to propose something better. When you do so, please focus on the user experience of the API more than the .proto definition since that is what users will face.

3) I think the primary useful feedback in this thread is that we should consider how to handle multi-control-plane installs. Is the plan to have multiple instances of the operator spec, one per control plane installation, or is the plan to have multiple control plane installations managed by a single operator? As designed right now the operator spec represents a single instance of the control plane, and the intention is to handle multiple instances through multiple specs. That seems reasonable, but maybe it needs to be better understood and documented?

4) What would be very useful is to write up some set of requirements/user journeys so we can make sure the design supports what we want it to support. I think this is a key part of the disagreement here, Costin is arguing that the design does not meet some set of requirements, but those are not written down anywhere so it isn't clear what the requirements are that are not being met. So again I'd say the best way to disagree is to write up the requirements that are not being met, get agreement that we need to meet those requirements, and then work towards designing an api and implementation that meets those requirements.

Generally I'm in favor of submitting this and starting to play around with it, and in parallel let's explore what requirements we think we are missing and alternative api designs that meet those requirements.

costinm:
On Mon, Dec 2, 2019 at 5:36 PM Sven Mawson <notifications@github.com> wrote:

> *@smawson* approved this pull request.
>
> In mesh/v1alpha1/operator.proto
> <https://github.com/istio/api/pull/1177#discussion_r352942212>:
>
> > +    IstioComponentSpec components = 50;
> +
> +    // Overrides for default values.yaml.
> +    TypeMapStringInterface values = 100;
>
> The long term goal is to remove values altogether right? We just have
> these while we're adding additional flags to components and mesh config?
>
> We should provide additional documentation here (and in our docs)
> describing what this is for. Ideally we'd also have a list of what values
> still go here and what values should not be included here and instead
> handled some other way.
>
> In mesh/v1alpha1/operator.proto
> <https://github.com/istio/api/pull/1177#discussion_r352942347>:
>
> > +    // Root for docker image paths e.g. docker.io/istio
> +    string hub = 12;
> +    // Version tag for docker images e.g. 1.0.6
> +    string tag = 13;
> +
> +    // Config used by control plane components internally.
> +    MeshConfig meshConfig = 40;
> +
> +    // Kubernetes resource settings, enablement and component-specific settings that are not internal to the
> +    // component.
> +    IstioComponentSpec components = 50;
> +
> +    // Overrides for default values.yaml.
> +    TypeMapStringInterface values = 100;
> +    // Unvalidated overrides for default values.yaml. Used for custom templates where new parameters are added.
> +    TypeMapStringInterface unvalidatedValues = 101;
>
> Ideally we'd have a better model than just "unvalidatedValues" as a way to
> let platform owners extend our installation system. I'd like them to be
> able to offer type-safe configuration that is validated through a schema
> rather than just having a blob of additional values. Not-blocking for this
> PR but we really should think through what we want that to look like.
>
> In mesh/v1alpha1/component.proto
> <https://github.com/istio/api/pull/1177#discussion_r352946579>:
>
> > +syntax = "proto3";
> +
> +import "k8s.io/api/core/v1/generated.proto";
> +import "k8s.io/api/autoscaling/v2beta1/generated.proto";
> +import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
> +import "github.com/gogo/protobuf/protobuf/google/protobuf/wrappers.proto";
> +
> +package istio.mesh.v1alpha1;
> +
> +option go_package="istio.io/api/mesh/v1alpha1";
> +
> +// IstioComponentSpec defines the desired installed state of Istio components.
> +message IstioComponentSpec {
> +    PilotComponentSpec pilot = 30;
> +    ProxyComponentSpec proxy = 31;
> +    SidecarInjectorComponentSpec sidecar_injector = 32;
>
> This thread is probably too long for a review discussion and should move
> to email, but a couple quick points anyway:
>
>    1.
>
>    istiod being available != microservices being deprecated. Done anyone
>    actually think we're going to deprecate microservices deployment mode in
>    1.5? Before we deprecate something the replacement needs to be in
>    production and stable with enough users to be confident, I don't see that
>    happening with istiod before 1.5. We don't want to tie the operator to
>    istiod (just as we decided not to tie SDS to istiod). We generally try to
>    avoid blocking ourselves, even if it costs us technical debt we need to
>    clean up later.
>
> Yes. But Istiod is a proof that control plane is not a fixed set of
(singleton) components named "pilot, policy, mixer,...".
And it is not tied in any way with any deprecation.

Not sure what you mean about 'SDS not tied to istiod' - AFAIK the plan
remains for Istiod to _only_ implement SDS, with
migration from file-based mount implemented on agent. So it is quite tied.

We are not blocked in any way on istiod - the only thing that is unblocked
by istiod is simpler 'traffic shift' upgrades for the
control plane, which was too difficult (but still possible) before.

>
>    1.
>
>    Simply disagreeing with an API is not useful. Much more useful is
>    providing an alternative proposal that meets all of the requirements that
>    the API is trying to address (and ideally additional requirements). So if
>    there is a proposal for an operator API that works better for some use
>    cases, great, please write up a doc with a proposal and discuss in
>    environments WG. We have until 1.5 ships to make this better so you have
>    time to propose something better. When you do so, please focus on the user
>    experience of the API more than the .proto definition since that is what
>    users will face.
>
> I don't think anyone 'simply disagrees with an API'.

There are concrete flaws in the API:
 - the dependency on a set of component names ( policy, mixer, pilot, etc)
- the fact the components are singleton ( you can have a single pilot - so
no traffic shifting mode, in-place is the only option )
- internal names instead of functional ( Louis feedback in an early comment
- you seem to have convinced him it's not an issue based on the followup,
but I haven't seen any rationale and I believe it is still a very valid
concern). Why would users care about internal names (pilot or istiod ) ?

I think you raise a very important point - we need to start with the
requirements.

And my understanding is that _operating_ - i.e. config changes including
version change (aka upgrade) are the primary requirements
for the operator API.

I have a proposal in env for traffic-shift based config changes / upgrades
- happy to discuss how it can address production safety,
but regardless of that proposal, I think it is reasonable to disagree with
an API codifying a proven bad practice, the in-place
upgrade we are doing - and the list of hardcoded internal names - have
clear problems and it doesn't serve anyone to
have users adopt an API based on those names and practices.


>
>    1.
>    2.
>
>    I think the primary useful feedback in this thread is that we should
>    consider how to handle multi-control-plane installs. Is the plan to have
>    multiple instances of the operator spec, one per control plane
>    installation, or is the plan to have multiple control plane installations
>    managed by a single operator? As designed right now the operator spec
>    represents a single instance of the control plane, and the intention is to
>    handle multiple instances through multiple specs. That seems reasonable,
>    but maybe it needs to be better understood and documented?
>
> The 'one or many' operators is not relevant for the API, but
implementation details.

That's the main problem with this API really - tied to implementation
names/details instead of features and behaviors.
And as a consequence creates dozen of protos with almost same content.



>
>    1.
>    2.
>
>    What would be very useful is to write up some set of requirements/user
>    journeys so we can make sure the design supports what we want it to
>    support. I think this is a key part of the disagreement here, Costin is
>    arguing that the design does not meet some set of requirements, but those
>    are not written down anywhere so it isn't clear what the requirements are
>    that are not being met. So again I'd say the best way to disagree is to
>    write up the requirements that are not being met, get agreement that we
>    need to meet those requirements, and then work towards designing an api and
>    implementation that meets those requirements.
>
>
The main requirement I'm concerned about is 'multiple control planes' use
case, where different control planes may run
different implementations - i.e. may run istiod instead of
pilot+injector+ca, or may run a completely custom XDS server
and a custom CA provider ( implementing our defined APIs ).

Again - forget the schedule for istiod, or when we deprecate mixer v1 - the
main point is that  APIs should not codify
internal component names or assume the control plane runs as a fixed set of
microservices.

As a consequence of this - the model should be generic and not tied to the
hardcoded list of components - or require
12 protos.

If there is any 'pilot' specific setting - it does not belong in the
operator API, but mesh config (as agreed). And it's really
not a pilot setting - but a 'network' or 'telemetry' or 'security' setting.



>    1.
>
> Generally I'm in favor of submitting this and starting to play around with
> it, and in parallel let's explore what requirements we think we are missing
> and alternative api designs that meet those requirements.
>
I don't disagree - but it is already submitted in the operator repo. So we
can play around with it.
There are other APIs that are not yet ready for istio/api - like the
istio/security/proto, or values.yaml
proto.

Costin

costinm:
If I wasn't clear:  I don't have any problem with handling upgrades/config
changes with multiple instances of this
proto. It is actually a proven model - by KNative for example.

The problem is that the structure of the API is forcing a strict
structure on each control plane.

I would strongly suggest people play a bit with KNative and their Revision
CRD.

smawson:
Can we move this discussion to a doc or email thread or something?

The requirements I'm seeing Costin raise:

1) We need to support incremental rollouts of control plane changes. Something like the knative revision model (updates to configuration generate revisions, configuration can specify which revisions are active). Needs a design discussion, but seems relatively simple to layer on top (strawman: control plane spec generates a revision for each change, operator can be configured to manage N live revisions, data plane can be configured to talk to specific revision).

2) We need a design for extensibility of components, so alternative components can be used and configured. We can't require additional components to be added through code changes. The current structure of the API bakes in a particular set of components in a release, which makes it very hard for alternative control plane components to be controlled via the operator. Needs a design discussion.

Both of these are valid requirements that we should pursue solutions to, but we don't need to solve them in this PR. We should try to have a solution by 1.5 though.

The last thing raised, the separation of feature config and component config, I think we've generally reached agreement that most stuff should be feature config, and it should be in a telemetry or security or networking block in mesh config. I don't think anyone disagrees with that as a goal.

The problem is that we do actually have some component-specific configuration, such as number of replicas, memory requirements, component-specific flags, etc. We need a place to put that stuff. We reached agreement in TOC that we will put those as per-component config blocks in a component section of the operator config, with a warning to users that this is not necessarily stable and will potentially break upgradability. It is break-glass control over low-level details that most people should not use. Our goal is to keep that minimal.

smawson:
Or maybe we can create an issue to track these requirements (or separate issues per requirement)?

costinm:
On the last point - "we need a place to put the replicas/memory": that's my
main concern with merging this PR as is.

I am perfectly file with this PR if it has a:
 map<component, ComponentSetting>  - and a single proto for the
replicas/memory/patches/etc.

I am not with

 pilot PilotComponentSettings
 policy PolicyComponentSetting
 mixer MixerSettings
....

I don't see the argument that we need the individual fields and protos for
'validation'.

If we can switch to the map<string,SingleProto> - instead of the many
duplicated protos- I'm +1 on merging it.

ostromart:
This is an issue that was explicitly discussed in the TOC and it was agreed that naming components was the way to go. To reiterate some reasons for doing so:
1. without explicit naming, user does not know what the valid values are for component names for a particular binary - there's no contract for it. 
2. some components will likely require specialized fields as we migrate values.yaml. The map approach does not accommodate that. 

I've started moving some of the points discussed here into a doc: https://docs.google.com/document/d/1XF-9HswZUcORuBiDdCoIl1TSVcfTYTBTQeJyOC2U_YM
@costinm and @smawson - feel free to edit that.

smawson:
There is a middle ground where we have:

Component pilot = 1;
Component galley = 2;
...

So we at least don't introduce a bunch of duplicate types. When we need to add component-specific fields we can just replace:

Component pilot = 2;
GalleyComponent galley = 2;
...

Assuming the structure is the same the yaml, json, and proto representations are forward compatible with doing that AFAIK.

That doesn't answer the full concern but at least would reduce some of the noise.

mesh/v1alpha1/operator.proto
ostromart: I added a pointer to the current Helm schema. We do have a proto representation of the same in the istio/operator repo but the docs are the source of truth so it seems more appropriate to point to that. 
Yes, we're going to gradually migrate all the values to the feature fields in MeshConfig - once that's done this field will be deprecated and removed.

costinm:
It is certainly better than a dozen of identical protos, but we still
define an api for policy and components that we plan to deprecate - and tie
the API to internal code names of components.

A map has same representation in yaml - and it gives us back the
flexibility to use a different control plane components, without baking the
current list into API and subject users to deprecation and migration.

costinm:
But to compromise :-), ok with Sven proposal for the components we're not
planning to depreacate (mixer, policy.)

And we can add a map for extra components that don't use the fixes names.
The tight coupling doesn't benefit anyone.

smawson:
Yes it is intentional that we are tying the API to internal names of components, that's what this API is, a component API. So that's WAI. Also subjecting users to deprecation and migration is also WAI. If a user specifies a flag to galley, they MUST do a migration. It doesn't matter if the flag was in a map or in a field, they still need to do a migration. In that sense it is better to have it in a field so they can see that the field is deprecated, get a warning, etc. So deprecation and migration is WAI.

costinm:
On Tue, Dec 3, 2019, 17:42 Sven Mawson <notifications@github.com> wrote:

> Yes it is intentional that we are tying the API to internal names of
> components, that's what this API is, a component API. So that's WAI. Also
> subjecting users to deprecation and migration is also WAI. If a user
> specifies a flag to galley, they MUST do a migration. It doesn't matter if
> the flag was in a map or in a field, they still need to do a migration. In
> that sense it is better to have it in a field so they can see that the
> field is deprecated, get a warning, etc. So deprecation and migration is
> WAI.
>

The difference is that with the map you can have a different MCP server -
not galley. Or a different ca - certmanager instead of citadel, if they
implement the equivalent interface.

Rest is the same - if they specify an option for citadel and switch to
certmanager, they need to migrate. But at least it is possible.

Map or hardcoded names doesn't make a difference in what the user types or
how it's implemented - except the hardcoded is restricted to a specific set
of names.


It is certainly better than a dozen of identical protos, but we still

costinm:
Regarding the concern that "some components will require specialized
fields" - if they do, they should be moved into mesh.yaml
as feature specific, not component specific.

The file is for overriding cpu/memory and patching components - a user
should only have to touch this if they need
to fine tune a component, in which case they probably know what component
needs to fine tune.

Most users should interact with the mesh config to control the behavior of
the mesh.

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,157 @@
...
+// Configuration common to most components.
+message ComponentSpec {
+    // Selects whether this component is installed.
+    TypeBoolValueForPB enabled = 1;
+    // Namespace for the component.
+    string namespace = 2;
+    // Arbitrary install time configuration for the component.
+    TypeInterface spec = 3;
+    // Schema to validate spec against.
+    google.protobuf.Any schema = 4;
```
smawson: I don't love having the schema be passed in by the user, that seems weird. If they already know the schema they can validate it themselves.

Can we have ComponentSpec just be "enabled" and KubernetesResourcesSpec for now?

Or do we have existing use cases for schema, spec, and namespace?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,157 @@
...
+    // Kubernetes resource spec.
+    KubernetesResourcesSpec k8s = 5;
+}
+
+// Configuration options for the proxy.
+message ProxyComponentSpec {
```
smawson: Was this only to remove "enabled"? Can we just make that an error?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,157 @@
...
+    ComponentSpec telemetry = 34;
+    ComponentSpec citadel = 35;
+    ComponentSpec node_agent = 36;
+    ComponentSpec galley = 37;
+
+    ComponentSpec ingressGateway = 38;
+    ComponentSpec egressGateway = 39;
+    // Map key is the name of the resources for the component. All k8s resources accessible through
+    // KubernetesResourcesSpec must have the same name.
+    map<string, ComponentSpec> customGateways = 40;
```
howardjohn: I am not sure these should be separate from `ingressGateway`. You can just have multiple gateways, one of them isn't more "first class" than others

ostromart:
Ok, I've applied Sven's suggestion. I also tried doing two maps (one with enums of component names, the other with string as keys) but enum keys are not supported in proto unfortunately. Anyway, I think the current structure is pretty compact and we can add "deprecated" to whatever component is on the way out (mixer) if appropriate. I'd leave that to the mixer team to decide.

mesh/v1alpha1/component.proto
ostromart: sgtm. We still need separate ingress and egress gateways since they use different charts.

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,157 @@
...
+
+    ComponentSpec cni = 41;
+    ComponentSpec coreDNS = 42;
+
+    // Non-core components. Map key is the name of the resources for the component. All k8s resources accessible through
+    // KubernetesResourcesSpec must have the same name.
+    map<string, ComponentSpec> extraComponents = 100;
```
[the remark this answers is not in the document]
ostromart: If it's not explicitly named above, it's not a core component. As to what exactly is a core component it's a bit vague but one definition could be a component that may have some logic embedded for it in the operator, and can always be found in the accompanying released charts.
We bundle prometheus charts but we'll not be in the business of embedding prom logic in operator code.

mesh/v1alpha1/component.proto
howardjohn: Yep. How I would envision it is just something like
```
ingressGateways:
- name: istio-ingressgateway
  namespace: istio-system
- name: my-custom-gateway
  namespace: istio-system
  someOtherField: foo
```

Something like that maybe? The exact shape isn't as much as a concern as making it possible though

mesh/v1alpha1/component.proto
ostromart: I'm confused - I thought you were complaining about the lack of schema in unvalidatedValues field? If it's up to the user to validate we are basically back in that model. 
We can make the schema field optional. Anyway, I'm also happy with no schema, I just put it in here because I thought you were asking for it.

mesh/v1alpha1/component.proto
ostromart: Yes. I've created a new type because this enablement issue also applies to map entries.

mesh/v1alpha1/component.proto
ostromart: Ok, yes, I think we do need name field too, updated.

mesh/v1alpha1/component.proto
smawson: I don't want the schema passed in this API, I want a separate API for registering component schemas which then result in it being validated here.

Think about it like CRDs, we don't pass in the CRD schema next to the spec for the CRD :)

costinm:
To use Sven's advice of focusing on requirements:

I think the 'validation' should verify that the names in the API are
matching names in the templates.
If user is using custom template, with a 'bob' component implementing a
fancy MCP service as part of the control plane - the API
should validate this.

I don't know if we require an extra 'registration' step for components -
the presence of the templates is part of this.

Names in a proto are NOT a validation - protos are for serialization, not
validators and don't allow the needed flexibility, while the
code in istioctl can easily check if "pilot" or "mixer" in the API are
matching valid components in the templates.

In other words: I still believe using a map and validating in code ( with
explicit registration as Sven suggests - or using implicit
one based on templates )

costinm:
How did coredns become core component ? It's in ecosystem and not beta.
While prometheus is more core than even mixer ( with v2 telemetry you just
need prom).

costinm:
At the same time - prom should not be managed by istio operator, it has its
own. We just need to install istio-related configs, not manage the prom
deployment.

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.12
 
 require (
 	github.com/gogo/protobuf v1.3.0
-	github.com/golang/protobuf v1.3.2 // indirect
-	golang.org/x/text v0.3.2 // indirect
 	google.golang.org/grpc v1.23.1
 	istio.io/gogo-genproto v0.0.0-20190930162913-45029607206a
+	k8s.io/api v0.0.0-20191206001707-7edad22604e1
+	k8s.io/apimachinery v0.0.0-20191203211716-adc6f4cd9e7d
```
howardjohn: Why do we have a golang dependency on k8s? This is a giant dependency that makes this repo much heavier

howardjohn: Is this because we are depending on their protos, and the codegen is smart enough to depend on the golang deps? How does this work for the other client languages like python? Is there any way to avoid this?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+    // Namespace for the component.
+    string namespace = 2;
+
+    // Arbitrary install time configuration for the component.
+    TypeInterface spec = 10;
+
+    // Chart path for addon components.
+    string chart_path = 30;
+    // Optional schema to validate spec against.
+    google.protobuf.Any schema = 35;
```
howardjohn: How does this work? A user is providing the schema?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Configuration for gateways.
+message GatewaySpec {
+    // Enablement is implied by the existence of this spec.
+
+    // Namespace for the gateway.
+    string namespace = 2;
+    // Name for the gateway.
+    string name = 3;
```
howardjohn: Gateways are selected by label. Should KubernetesResourcesSpec allow labels? Otherwise we need to add a labels here or I don't think this would work

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+    string namespace = 2;
+
+    // Arbitrary install time configuration for the component.
+    TypeInterface spec = 10;
+
+    // Kubernetes resource spec.
+    KubernetesResourcesSpec k8s = 50;
+}
+
+// Configuration for external components.
```
linsun: are these the add-ons?

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,179 @@
...
+    // Root for docker image paths e.g. docker.io/istio
+    string hub = 12;
+    // Version tag for docker images e.g. 1.0.6
+    string tag = 13;
+    // Resource suffix is appended to all resources installed by each component. Used in upgrade scenarios where two
+    // Istio control planes must exist in the same namespace.
+    string resource_suffix = 14;
+
+    // Config used by control plane components internally.
+    MeshConfig mesh_config = 40;
```
linsun: where is this MeshConfig specified?

linsun:
@ostromart can you update the desp to indicate what you intend to do?  Is MeshConfig definition outside of this PR?

cc @rcernich PTAL

linsun:
In general, I like the direction of this and bring consistency to mesh config and istio control config.

mesh/v1alpha1/component.proto
ostromart: Could you give some more explicit guidance about how you'd expect that to look? For istioctl case, this CR is the entirety of config - there's no separate registration path. So afaics this means somewhere we have to statically provide a list of the schemas within this CR. It could be a separate list but it seems more clean to combine it with the data it's actually validating.

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+    ComponentSpec pilot = 30;
+    ComponentSpec proxy = 31;
+    ComponentSpec sidecar_injector = 32;
+    ComponentSpec policy = 33;
+    ComponentSpec telemetry = 34;
+    ComponentSpec citadel = 35;
+    ComponentSpec node_agent = 36;
+    ComponentSpec galley = 37;
+    ComponentSpec cni = 38;
+    ComponentSpec coreDNS = 39;
```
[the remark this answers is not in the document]
ostromart: fixed.

mesh/v1alpha1/component.proto
ostromart: Yes, this would be stuff like prometheus, grafana, kiali etc.

mesh/v1alpha1/operator.proto
ostromart: MeshConfig is defined in the same dir in config.proto: https://github.com/istio/api/blob/master/mesh/v1alpha1/config.proto

mesh/v1alpha1/component.proto
ostromart: Yes, exactly.

howardjohn: If I know the schema why do I need a schema? Can you please provide an example of how the user is expected to use this?

mesh/v1alpha1/component.proto
ostromart: Good point, added.

mesh/v1alpha1/component.proto
ostromart: It's optional and makes it a one step check in the operator rather than having to validate externally.
If user wants to plug in an external chart and they happen to have a schema for it, they could do something like:
extraComponents:
  myComponent:
    spec:
      foo: bar
    schema:
      typeUrl: <some URL to your schema> (see any.proto for details)

We would try to unmarshal spec: into the schema. 
Sven is talking about registering the schema in a comment elsewhere but I'm not sure how that would in practice.

go.mod
ostromart: Yes this is because protoc generates Go files and these need the deps to work. 
If this becomes burdensome for this repo we have the option of mirroring more structs.

howardjohn: How much of k8s api are we using? I recall seeing some structs mirrored, its hard to tell though since there are 67 files in this PR.

The pain of taking on a k8s dependency is large, so if feasible we should avoid it. Its not just the burden for THIS repo, its a burden for client-go, istio, proxy, and any other dependency (including non-istio ones).

mesh/v1alpha1/kubernetes.proto
```diff
@@ -0,0 +1,225 @@
...
+    uint32 maxUnavailable = 3;
+}
+
+// Mirrors k8s.io.api.apps.v1.DeploymentStrategy for unmarshaling.
+message DeploymentStrategy {
+    string type = 1;
+    RollingUpdateDeployment rollingUpdate = 2;
+}
+
+// Mirrors k8s.io.api.apps.v1.RollingUpdateDeployment for unmarshaling.
```
howardjohn: If we are already importing k8s why do we need to mirror these?

mesh/v1alpha1/component.proto
howardjohn: In the schema do they ONLY need to specify the typeUrl? they don't need to specify something like 
```yaml
extraComponents:
  myComponent:
    spec:
      foo: bar
    schema:
      typeUrl: "my.proto.foo"
      foo: string
```

or something?

How do they register the type? Without an idea of the implementation its hard to understand what makes sense for the API

Can we implicitly determine the schema? Defining the schema in the config object we are defining isn't great... If not should we borrow Envoy's syntax? See https://github.com/istio/istio/wiki/EnvoyFilter-Samples#tracing-and-access-logging for an example

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+    ComponentSpec node_agent = 36;
+    ComponentSpec galley = 37;
+    ComponentSpec cni = 38;
+    ComponentSpec core_dNS = 39;
+
+    repeated GatewaySpec ingress_gateways = 40;
+    repeated GatewaySpec egress_gateways = 41;
+
+    // Extra addon components which are not explicitly specified above.
+    map<string, ExternalComponentSpec> extra_components = 100;
```
howardjohn: In my yaml, I should be able to specify:
```yaml
pilot:
  some config
prometheus:
  some other config
```

With this current config, I would need to do
```yaml
pilot:
  some config
extraComponents:
  prometheus:
    some other config
```
right?

I am pretty sure no one, even people in the env WG, will be able to figure out what is an extra component and what is a "first class" component without a few trial and errors or digging through the docs.

costinm:
I still don't get the schema and validation part.

Protos are a serialization format - what is that schema and how does it
work ?

go.mod
ostromart: Don't those repos already have a k8s dependency regardless? It's probably not hard to mirror all the structs we need, I can look into it. The concern is more the maintenance overhead of the copies.

howardjohn: istio does if course, but I didn't think client go did - now that I think about it it must though. Maybe it's not a big deal then. I am definitely not saying we should not add this, just we should make sure we don't do it without good reason. I saw we did copy some structs already so wasn't sure what we were important and what we were copying.

In the past we imported like megabytes of dependencies for a single string constant so trying to avoid that case

mesh/v1alpha1/component.proto
ostromart: It depends how the schema is defined. If it's a http URL (type server) then you don't need "foo: string" info. If it's defined together with the message if would have "foo: string" type info. 
We could define the schema elsewhere using the Any serialization format rather than Envoy's. It really depends on whether we want to steer users towards defining their schemas in proto format or something else (OpenAPI is another possibility).

mesh/v1alpha1/kubernetes.proto
ostromart: It's a bug. There are problems with importing k8s protos and have everything unmarshal as you'd expect, but there are solutions to those, just a question of time and willpower. In the operator repo very few messages are mirrored.

mesh/v1alpha1/component.proto
howardjohn: Even after having this discussion with you and a bit of googling I still have no clue how I would use this. Can you update the comment here with a description of how to use this and a concrete example?

ostromart:
@costinm proto is also a schema format of sorts, which at least gives you field types and layout. That's sufficient to catch typos and simple bad data types, which are the most common problems. 
Schema formats like OpenAPI are richer but are harder to read and work with - overkill in this case imo.

howardjohn:
For what it's worth, helm uses json schema https://github.com/helm/community/blob/master/helm-v3/001-charts.md#schematized-values-files

mesh/v1alpha1/component.proto
ostromart: Yes, right. I guess the problem is we don't have a good definition for what is a "first class" vs "extra" component. If we say that "first class" is whatever charts we distribute then a bunch of other components would become "first class". I think that's ok but it somewhat breaks our definition of addon - prom is an addon but it's also first class?

howardjohn: Users don't really care do they? There should just be one way to do things (everything at the same top level). I don't see a need to distinguish between them, is there something I am missing? If it's a about how proto serialization is handled we should work around that and not push complexity to the user just because we decided to use proto

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,179 @@
...
+
+    // Kubernetes resource settings, enablement and component-specific settings that are not internal to the
```
howardjohn: What does "internal to a component mean"?

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,179 @@
...
+    IstioComponentSetSpec components = 50;
+
+    // Overrides for default values.yaml. This is a validated pass-through to Helm templates.
+    // See the Helm installation options for schema details: https://istio.io/docs/reference/config/installation-options/.
+    TypeMapStringInterface values = 100;
+    // Unvalidated overrides for default values.yaml. Used for custom templates where new parameters are added.
+    TypeMapStringInterface unvalidated_values = 101;
+
+    // Status describes the current state of a component.
+    enum Status {
```
howardjohn: can you add docs for each of these?

mesh/v1alpha1/operator.proto
```diff
@@ -0,0 +1,179 @@
...
+    // Config used by control plane components internally.
+    MeshConfig mesh_config = 40;
+
+    // Kubernetes resource settings, enablement and component-specific settings that are not internal to the
+    // component.
+    IstioComponentSetSpec components = 50;
+
+    // Overrides for default values.yaml. This is a validated pass-through to Helm templates.
+    // See the Helm installation options for schema details: https://istio.io/docs/reference/config/installation-options/.
+    TypeMapStringInterface values = 100;
```
howardjohn: Can we doc when I should configure something through here and when it's available in components?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+
+// Configuration for internal components.
+message ComponentSpec {
+    // Selects whether this component is installed.
+    TypeBoolValueForPB enabled = 1;
+    // Namespace for the component.
+    string namespace = 2;
+
+    // Arbitrary install time configuration for the component.
+    TypeInterface spec = 10;
```
howardjohn: how do I know what I can put here?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+// Configuration for gateways.
+message GatewaySpec {
+    // Enablement is implied by the existence of this spec.
+
+    // Namespace for the gateway.
+    string namespace = 2;
+    // Name for the gateway.
+    string name = 3;
+    // Label for the gateway.
+    string label = 4;
```
howardjohn: How is label a string? shouldn't it be a map?

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+    // k8s toleration
+    // https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
+    Toleration tolerations = 14;
+
+    // Overlays for k8s resources in rendered manifests.
+    repeated K8sObjectOverlay overlays = 100;
+}
+
+// Patch for an existing k8s resource.
+message K8sObjectOverlay {
```
howardjohn: This could use some examples. I could not figure out how to use this just from the docs here

mesh/v1alpha1/component.proto
```diff
@@ -0,0 +1,177 @@
...
+
+
+// GOTYPE: map[string]interface{}
+message TypeMapStringInterface {}
+
+// GOTYPE: interface{}
+message TypeInterface {}
+
+// GOTYPE: *BoolValueForPB
+message TypeBoolValueForPB {}
```
howardjohn: Why do we need this? Doesn't proto have native bool wrappers

mesh/v1alpha1/operator.proto
ostromart: Internal to component means used by the component itself (i.e. in the code). External means related to the component but not used by the component (e.g. k8s resources). It is a bit difficult to explain so I'm open to suggestions how to word this better.

mesh/v1alpha1/operator.proto
ostromart: Done.

mesh/v1alpha1/operator.proto
ostromart: I added some more guidance. In general this sort of stuff would be documented in user docs  (it is right now for ICP) and I don't want to copy too much of that here.

mesh/v1alpha1/component.proto
ostromart: Yes, it should, fixed.

mesh/v1alpha1/component.proto
ostromart: Did you see the examples at the top of the file? If you could not figure it out that's a problem.

mesh/v1alpha1/component.proto
ostromart: Yes, but it caused some problems with jsonpb unmarshal. I don't recall the details but the bottom line is we had to switch to our own definition.

mesh/v1alpha1/component.proto
ostromart: It's defined by the component owner so it would be up to them to provide the schema and docs to the users.

mesh/v1alpha1/component.proto
ostromart: I assume you're asking about the dynamically registered case. Did you see https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/type.proto?
That is a proto for schema definition. We just need to pass it to the operator somehow and I can think of a couple of different ways to do it:
1. The type server idea that any.proto is talking about where you get a protobuf.Type response from it. Operator would query it using http to look up types dynamically.
2. Use a YAML serialization of protobuf.Type and pass that to the operator through some other means. Probably the cleanest way would be to define a CRD for component schemas that the operator would watch.  
Once operator has this schema, it can use proto.Any to try to unmarshal some user YAML and see if there are any errors.

mesh/v1alpha1/component.proto
ostromart: Yes, agreed. Then let's make all components that we distribute charts for "first class" and not extraComponents. 
The only thing that would go into extraComponents would be for new charts that some fork of Istio wants to add.

howardjohn: Can we just make everything first class?

ostromart: I don't think so. Unless we make schemas compulsory for external components (and that's a big ask), we couldn't validate the syntax of the first class components.

howardjohn: We can allow users to specify any config "shape" we want, and we can validate it any way we want. Lets not push the limitations of using protobuf for validation onto the users and make the experience worse for them.

We should start with the user experience, and use that to decide how we implement the API, not the other way around. If we do that,  its clear (to me, at least) we shouldn't have things like `extraComponents`.

ostromart: We can always write a custom unmarshaler I suppose. At least for the time being I will move everything that has a chart into the named fields section.

PR #1179, opened 2019-11-22
Add proto for analysis messages

zerobfd:
See https://docs.google.com/document/d/1lCnDBVzRzkr22BWjzPtGk6EK6EvGOnVIe2qBpaiQ3SM/ for more context and discussion around the change.

I split the struct up a bit differently than the doc to make things (hopefully) more clear and make the developer UX a bit easier.


analysis/v1alpha1/message.proto
```diff
@@ -0,0 +1,113 @@
...
+    WARNING = 8;
+    INFO = 12;
+  }
+
+  // Represents how severe a message is. Required.
+  Level level = 3;
+
+  enum MessageTypeVersion {
+    V1ALPHA1 = 0;
+  }
```
ayj: Presumably this matches the version in the file path? This is part of the protobuf type.

analysis/v1alpha1/message.proto
```diff
@@ -0,0 +1,113 @@
...
+  // Required.
+  string documentation_url = 5;
+
+}
+
+// AnalysisMessageSchema is the set of information that's needed to define a
+// weakly-typed schema. The purpose of this proto is to provide a mechanism for
+// validating istio/istio/galley/pkg/config/analysis/msg/messages.yaml to make
+// sure that we don't allow committing underspecified types.
+message AnalysisMessageSchema {
```
ayj: The naming here seems reversed. I would expect a generic message to be weakly typed and the analysis to be strongly typed.

analysis/v1alpha1/message.proto
```diff
@@ -0,0 +1,113 @@
...
+  // A human readable description of what the error means. Required when
+  // defining a schema in YAML.
+  string description = 2;
+
+  // A go-style template string defining how to combine the args for a
+  // particular message into a log line. Required when defining a schema in
+  // YAML.
+  string template = 3;
+
+  message ArgType {
```
ayj: Did you consider https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#struct? The template's format parameters would need to specify the full path in the struct.

analysis/v1alpha1/message.proto
```diff
@@ -0,0 +1,113 @@
...
+option go_package="istio.io/api/analysis/v1alpha1";
+
+// AnalysisMessageType describes some common information that is needed for both
+// creating a message schema or a reified message. Weakly typed message schemas
+// are defined in istio/istio/galley/pkg/config/analysis/msg/messages.yaml
+message AnalysisMessageType {
+  // A human-readable name for the message type. e.g. "InternalError",
+  // "Deprecated". This should be the same for all messages of the same type.
+  // Required.
+  string name = 1;
```
ayj: Why is this already deprecated? Can we just omit it entirely?

zerobfd: "Deprecated" here was just an example of the name of a message, see https://github.com/istio/istio/blob/master/galley/pkg/config/analysis/msg/messages.yaml

I'll pick a better example to prevent confusion, I just picked the two at the top of the list.

geeknoid:
There's a few things missing here.

#1. This new proto isn't being built. It needs to be built and linted. So it needs to be integrated into the make system.

#2. If you want this content to be documented on istio.io, then you need to add the needed voodoo in the makefile in order to produce a .pb.html file. If you do not intend to document this on istio.io, then you should remove the $ directives in the comments at the top of the file, since those are only relevant for the doc publishing process.

#3. Since I believe this proto will be used to describe status in Istio CRDs, then this probably will need to be integrated with the build of the .json files that described our schema to the API server. And will probably need to be integrated with the client-go libraries in order to enable programmatic access to the status state. @jasonwzm can help with the JSON stuff, @richardwxn can help with the client-go stuff.

jasonwzm:
@zerobfd Can you rebase again to make gencheck work?

analysis/v1alpha1/message.proto
```diff
@@ -0,0 +1,128 @@
...
+// new analyzers we offer a path that doesn't require committing to two
+// different repos and solidifying the interface.
+// Finally, we can create a new proto message of a specific message type and
+// commit it to istio/api when we need a strong guarantee for cross platform
+// communication.
+
+// AnalysisMessageBase describes some common information that is needed for all
+// messages. All information should be static with respect to the error code.
+message AnalysisMessageBase {
+  // A human-readable name for the message type. e.g. "InternalError",
```
louiscryan: These read a little bit like pseudo-enum values in strings. While we're not guaranteeing stability I presume the intent is to keep these relatively stable as they are more descriptive of. This seems a little at odds with 'code' which is designed for uniqueness but maps to these names in an unspecified way - can't tell if the mapping is unique or non-unique.

See https://cloud.google.com/apis/design/errors#http_mapping for a similar pattern

including the 'detail' section which seems similar to the documentation link for example.

analysis/v1alpha1/message.proto
louiscryan: This seems very implementation centric to use 'go' as the typing and declaration system.

analysis/v1alpha1/message.proto
```diff
@@ -0,0 +1,128 @@
...
+// validating istio/istio/galley/pkg/config/analysis/msg/messages.yaml to make
+// sure that we don't allow committing underspecified types.
+message AnalysisMessageWeakSchema {
+  // Required
+  AnalysisMessageBase message_base = 1;
+
+  // A human readable description of what the error means. Required.
+  string description = 2;
+
+  // A go-style template string defining how to combine the args for a
```
louiscryan: Link to template spec

analysis/v1alpha1/message.proto
```diff
@@ -0,0 +1,128 @@
...
+// list of args at runtime. Developers can also create stronger-typed versions
+// of GenericAnalysisMessage for well-known and stable message types.
+message GenericAnalysisMessage {
+  // Required
+  AnalysisMessageBase message_base = 1;
+
+  // Any message-type specific arguments that need to get codified. Optional.
+  google.protobuf.Struct args = 2;
+
+  // A list of strings specifying the path for resources which were the cause of
```
louiscryan: can you clarify the meaning of 'path for resources' ?

analysis/v1alpha1/message.proto
zerobfd: The code and the name field are intended to be a one-to-one mapping. If I group them into an embedded message and add comments, does it make the intent clearer?

analysis/v1alpha1/message.proto
zerobfd: I'm hoping that before this gets out of alpha this can be changed to something less pinned to Go, but as it stands I want it to be compatible with what's already in istio/istio. Since it's used in code generation as-is right now, I'd like there to be an alpha API to work off of before adding in a translation layer.

louiscryan: fair enough. maybe add this as a comment ?

analysis/v1alpha1/message.proto
louiscryan: don't strictly have to use an embedded message as it wouldn't guarantee enforcement, sufficient to describe contract in comments

analysis/v1alpha1/message.proto
zerobfd: I intended this to mean the "kind/namespace/resource" construction which identifies a resource, but then someone told me that wasn't a path, it was a qualified resource name. Although when I google around I don't see anything about that--is there a concept of a resource ID somewhere that I can link to?

PR #1182, opened 2019-11-28
add enable for locality lb setting

ckcd:
See https://github.com/istio/istio/issues/18903


networking/v1alpha3/destination_rule.proto
```diff
@@ -728,4 +728,7 @@ message LocalityLoadBalancerSetting{
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
+
+    // disable locality load balancing.
```
hzxuzhonghu: Doc clearly the relationship between mesh Config and DR

ckcd: DONE.

networking/v1alpha3/destination_rule.proto
```diff
@@ -728,4 +728,8 @@ message LocalityLoadBalancerSetting{
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
+
+    // disable locality load balancing, this is DestinationRule-level and will override mesh wide settings in entirety.
+    // e.g. true means that turn off locality load balancing for this DestinationRule no matter what mesh wide settings is.
+    bool disable = 3;
```
yingzhuivy: what if I have locality load balancing turned off mesh wide, and I want to turn it on per DR?

howardjohn: Should we just make this a field, `enabled`, such that

`enabled=true` use locality lb
`enabled=false` don't use locality lb
not specified - use mesh config

?

We can get "not specified" by using the wrapper bool object instead of just primitive `bool`

yingzhuivy: 👍 i agree

ckcd: Sounds good! will do this change ~

ckcd: DONE, PTAL. @howardjohn

ckcd:
`panic: runtime error: slice bounds out of range [:-1]`

Seems like the `gencheck_api` test has some problem ... since most PR here https://github.com/istio/api/pulls is failed ...

howardjohn:
Yes someone is working on fixing this, it's not related to this change

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: But use `enabled` requires user to additionally set one more field, because by default it is false for a bool value

ckcd: Uh, sorry for previous bad code, I have already re-push to correct the `bool` to `google.protobuf.BoolValue` so that we can get `not specified`.

@hzxuzhonghu @howardjohn

howardjohn: :+1:

networking/v1alpha3/destination_rule.proto
```diff
@@ -728,4 +729,8 @@ message LocalityLoadBalancerSetting{
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
+
+    // enable locality load balancing, this is DestinationRule-level and will override mesh wide settings in entirety.
+    // e.g. true means that turn on locality load balancing for this DestinationRule no matter what mesh wide settings is.
+    google.protobuf.BoolValue enable = 3;
```
howardjohn: nit: on our helm settings we always used `enabled` not `enable`. We may want to use the same here. Can you look if we have other APIs that have a field like this? We should just make sure we are consistent with them

ckcd: > on our helm settings we always used enabled not enable.

Uh yes, your are right. will do this change.

ckcd: DONE. PTAL.

ckcd:
force push to solve conflict.

ckcd:
@howardjohn with 2 approve can we merge this ?

howardjohn:
Need @rshriram or someone else from TOC to approve any API changes.

@louiscryan can as well

yingzhuivy:
@rshriram @louiscryan can someone take a look at this? 🙏 

@ckcd could you rebase?

networking/v1alpha3/destination_rule.proto
```diff
@@ -744,4 +745,8 @@ message LocalityLoadBalancerSetting{
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
+
+    // enable locality load balancing, this is DestinationRule-level and will override mesh wide settings in entirety.
+    // e.g. true means that turn on locality load balancing for this DestinationRule no matter what mesh wide settings is.
+    google.protobuf.BoolValue enabled = 3;
```
hzxuzhonghu: enable

ckcd: > on our helm settings we always used enabled not enable. We may want to use the same here.

as @howardjohn said, so i use `enabled` here.

ckcd:
thanks @yingzhuivy 

force push to resolve conflict.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: IC， we use `enabled` for all install options

rshriram:
sorry for the delay. approved.

PR #1185, opened 2019-11-28
Clarify location of timeout field

frankbu:
Fixes https://github.com/istio/istio.io/issues/5867


geeknoid:
in protos, you should use paths that point to istio.io instead of relative paths like /docs/... The proto doc process takes are of the mapping. This makes it so you get useful links if you're just looking at the protos directly.

PR #1186, opened 2019-11-29
Add istio api extensions definitions

jasonwzm:
Proposing some extensions for Istio API protos. With these extensions, we are able to specify additional rules to the API fields. e.g.

``` protobuf
message Destination {
...
string subset = 2 [(extension.rules).string.pattern = "^[a-zA-Z0-9](?:[-a-z-A-Z0-9]*[a-zA-Z0-9])?$"];
...
}
```

This will allow us to express richer schema syntax, more docs and better API/client tooling.


geeknoid:
Beautiful.

Issue #909, opened 2019-04-25
OutlierDetection documentation wrong for error types


howardjohn:
https://istio.io/docs/reference/config/networking/v1alpha3/destination-rule/#OutlierDetection

Under consecutiveErrors: 502, 503, and 504s only, 500 codes will not trigger the outlier detection

But in the docs above:

For HTTP services, hosts that continually return 5xx errors for API calls are ejected from the pool for a pre-defined period of time.

So sounds like the docs are incorrect here. It will only work for 502/503/504 from looking at the code and https://github.com/istio/api/pull/617

codependent:
@howardjohn Thanks for clearing it up. 

Just one question why isn't 500 included in the circuit breaking? In my view a service that returns continuously Internal Server Errors should also be also ejected from the pool. Otherwise we are keeping a service that is failing for whatever reason.

howardjohn:
Some context from the change #617 that introduced it:

> We try not to expose the plethora of sometimes confusing Envoy options to end users, in the routing api.
> 
> Within a mesh, gateway errors will be more common (502/503/504) while most sensible external services will return a 503 to shed load.
> 
> Secondly we just made the outlier detection generic to both tcp and http. The consecutive gateway error applies only to http and will make no sense in tcp context.
> 
> I also feel that 500 error code is not something indicative of overload. The whole idea behind outliers is to remove overloaded servers from the lb pool.
> 
> We don’t have very many users relying on this behavior I think. We kept it intentionally generic so that we can switch to a more specific error code in future (which happens to be now).

codependent:
I understand the motivation now however I think there's another valid scenario that is not supported.

In our microservice infrastructure we were relying on Hystrix as an application circuit breaker. When downstream services are crashing (500 errors) we don't want to keep sending traffic to them, and  Hystrix removed them temporaly from the pool.

Right now with Istio, this failing services are being invoked despite being actually down. That is to say we are protected from latency/overloading but not from failure. We can always translate 500 errors to 502/503/504 at the microservices but this seems like a work around. 

Does this make sense?

codependent:
Envoy explicitly distinguishes between consecutive gateway failures (502,503,504) errors and consecutive errors (5xx), supporting both of them.

https://www.envoyproxy.io/docs/envoy/v1.5.0/intro/arch_overview/outlier

Istio’s consecutiveErrors field is therefore misleading, shouldn’t it be aligned with Envoys terminology? Also, why not allow both kinds of configurations?

[...]

PR #1189, opened 2019-12-02
deprecate consecutive_errors and add consecutive_gateway_errors & consecutive_5xx_errors

yingzhuivy:
Related to https://github.com/istio/api/issues/909
This PR deprecates consecutive_errors and adds consecutive_gateway_errors and consecutive_5xx_errors field in OutlierDetection.
@frankbu @howardjohn


yingzhuivy:
@frankbu ptal

yingzhuivy:
@frankbu @howardjohn what's next step?  should i submit a pr for the actual implementation now or wait for this to merge

howardjohn:
@yingzhuivy you are free to work on the implementation if you prefer, with the caveat that there may be some wasted time if others object to this. You can use a `replace` directive in go.mod until this merges for development purposes if you were not aware.

@rshriram can you take a look?

yingzhuivy:
cc @jhahn21 @brianwolfe

networking/v1alpha3/destination_rule.proto
```diff
@@ -522,7 +523,26 @@ message OutlierDetection {
   // 502, 503, or 504 return code qualifies as an error. When the upstream host
   // is accessed over an opaque TCP connection, connect timeouts and
   // connection error/failure events qualify as an error.
-  int32 consecutive_errors = 1;
+  // $hide_from_docs
+  int32 consecutive_errors = 1 [deprecated=true];
+
+  // Number of gateway errors before a host is ejected from the connection pool.
+  // When the upstream host is accessed over HTTP, a 502, 503, or 504 return
+  // code qualifies as an error. When the upstream host is accessed over an
```
rshriram: as a gateway error

networking/v1alpha3/destination_rule.proto
```diff
@@ -522,7 +523,26 @@ message OutlierDetection {
...
+  // opaque TCP connection, connect timeouts and connection error/failure
+  // events qualify as a gateway error.
+  // If set to 0, this feature is disabled. Defaults to 5.
```
rshriram: remove this. There are no defaults.

rshriram: ignore

networking/v1alpha3/destination_rule.proto
```diff
@@ -522,7 +523,26 @@ message OutlierDetection {
...
+  google.protobuf.UInt32Value consecutive_gateway_errors = 6;
+
+  // Number of 5xx errors before a host is ejected from the connection pool.
+  // When the upstream host is accessed over TCP, errors are internally mapped
+  // to HTTP 5xx codes and treated as such.
```
rshriram: For TCP, its essentially just connect timeout and connection reset/error, unless there is more nuancing on the types of errors. So, I suggest repeating the above sentence "When the upstream host is accessed over an opaque TCP connection, connect timeouts, ... events qualify as an error".

networking/v1alpha3/destination_rule.proto
```diff
@@ -522,7 +523,26 @@ message OutlierDetection {
...
+  // If set to 0, this feature is disabled. Defaults to 0.
```
rshriram: This defaults to 5 as per Envoy docs. https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster/outlier_detection.proto#cluster-outlierdetection

yingzhuivy: The default in envoy is same as consecutive_5xx_errors=5 and consecutive_gateway_errors=0 (disabled) 
My questions are:
1. do we want to change the default behavior? e.g., a user without any setting will see a different behavior if we set gateway default to 5. 
2. if we want to set the both defaults to 5, we don’t need to set the default for gateway since it has no meaning (since 5xx includes gateway).

rshriram: 1. user without any setting but with just outlier detection probably falls under the legacy user using the older consecutive_errors format. 
That said, Istio behavior for last two releases has been 

```
if outlier.ConsecutiveErrors > 0 {
		// Only listen to gateway errors, see https://github.com/istio/api/pull/617
		out.EnforcingConsecutiveGatewayFailure = &wrappers.UInt32Value{Value: uint32(100)} // defaults to 0
		out.EnforcingConsecutive_5Xx = &wrappers.UInt32Value{Value: uint32(0)}             // defaults to 100
		out.ConsecutiveGatewayFailure = &wrappers.UInt32Value{Value: uint32(outlier.ConsecutiveErrors)}
	}
```

i.e. if user did not set the consecutive error, then we were still doing consecutive_5xx by default. So, I think what you have is good.?

yingzhuivy: so
1. there is discrepancy between the legacy consecutive_errors documentation and actual implementation for the last two releases: i.e. if user did not set the consecutive error, we were still doing 5xx while we should actually be doing gateway errors. Do we want to fix the implementation?
2. for the new consecutive_5xx_errors and consecutive_gateway_errors, what do we want for default. do we want to use 5xx as default or gateway.

frankbu: 1. It's too late to change the implementation of the old `consecutive_errors` field. Right or wrong, it was what it was. Just keep it working the same, only deprecated now.

2. I think the new default also needs to stay the same: 5 5xx errors. So to get the same behavior as setting the old field to some value, e.g., `consecutive_errors = 10`, the user will now need to set both fields:
```
consecutive_gateway_errors = 10
consecutive_5xx_errors = 0 // disable
```
@rshriram WDYT?

rshriram: This API has become very confusing unfortunately. Not to mention, envoy's behavior of assuming some values for unspecified fields is making it even more hard to reason about the behavior. I suggest this:

1. deprecate consecutive_errors and max_ejection_percent (both are ints). Instead convert them into Int wrappers (consecutive_gateway_errors, consecutive_5xx_errors, max_ejection_percentage). If not specified, that feature should be disabled (there is no 5 or 5000 default). If specified, we take the value.
2. In the outlier implementation in cluster.go in pilot, we should keep the existing deprecated field implementations as is.  In addition, (thinking out aloud), I think we should also explicitly DISABLE or set high defaults for all the rest of the knobs in Envoy (https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster/outlier_detection.proto#cluster-outlierdetection) so as to get a predictable behavior. There are soo many knobs there that are confusingly worded. Some candidates to disable include the failure_precentage_minimum_hosts, consecutive_local_origin_failure, etc. that have default set to 5.

frankbu: @rshriram, you're suggestion #1 is a breaking change. Currently, if nothing is set, the default Envoy 5 consecutive 5xx errors will apply. My suggestion is to live with that behavior, but now users can also change 5xx errors by setting it to some other value. The user can also still set gateway errors but if setting it to a value greater than 5, you need to explicitly disable the 5xx one. I don't think it's really that complicated, and it is simpler but pretty well aligned with the underlying Envoy API. Just need to document the two fields clearly.

frankbu: Actually, now that I think of it, it can be simpler than I suggested. The user won't need to explicitly disable 5xx errors (`consecutive_5xx_errors = 0 // disable`) if we say that when `consequtive_gateway_errors` is set, it overrides the value of `consecutive_5xx_errors` (explicit or default value). I'm guessing that's how Envoy handles it too. So this config:
```
consecutive_gateway_errors = 10
```
will result in a timeout after 10 gateway errors but also after 5 (default) of any other 5xx errors. You would still need to set both fields, like I said above, if you want the exact same behavior as setting the old `consecutive_errors` field.

jhahn21: @rshriram, @frankbu I'd like to clarify two things first.

- What kind of default behavior we want?
  - Do we want to follow the Envoy's default behavior? IIUC, the default is only enabling `consecutive_5xx_errors` with `5`.
  - Or do we want to have Istio own default behavior regardless of Envoy's one?

- How do we want to fix the current bug?
  - The current [doc](https://istio.io/docs/reference/config/networking/destination-rule/#OutlierDetection) says `consecutiveErrors` is enabled by default with `5`.
  - But the actual default behavior is enabling `consecutive_5xx_errors` with `5`.  `consecutive_5xx_errors` is disabled only when `consecutiveErrors` (i.e.,  `consecutive_gateway_errors`) is set  to a positive value. (It looks like [this PR](https://github.com/istio/istio/commit/af73923d620cfa52e4433a0102d0bbb80500d278) introduce it.)
  - How do we want to resolve this inconsistency? Do we want to fix the documentation or the implementation?

frankbu: @jhahn21 I'm suggesting the Istio default behavior should be the same as Envoy's (5 5xx errors). That is how it works today, and has been for a couple of releases already, so it would be a breaking change if we do something different. I also think there's no advantage to having different default behavior than Envoy's, it's just another complication.

As far as the current bug is concerned, I'm saying the doc is wrong and would need to be fixed, but since we are deprecating the old field, we will remove that doc and therefore it doesn't need to be fixed. We just need to clearly document the behavior for the 2 new fields.

jhahn21: @frankbu yes, that makes sense and I agree with you.

@rshriram do you also agree on it? @yingzhuivy will update her PR as soon as both of you agree on it.

rshriram: > The user won't need to explicitly disable 5xx errors (consecutive_5xx_errors = 0 // disable) if we say that when consequtive_gateway_errors is set, it overrides the value of consecutive_5xx_errors (explicit or default value). I'm guessing that's how Envoy handles it too. 

Can we validate this assumption ? @lizan is this true?

yingzhuivy: I don't think consequtive_gateway_errors overrides consecutive_5xx_errors. Users can set separate thresholds on the two, for example gateway=3, 5xx=5. When both of them are set, ejection will be triggered as soon as one of them reaches threshold. https://github.com/envoyproxy/envoy/blob/1a7b32de759939f82af2cf7e83471aff5d57c93d/source/common/upstream/outlier_detection_impl.cc#L66

frankbu: @yangminzhu That's too bad, because it means that users can never set the gateways one greater than the 5xx one, since the gateway codes overlap with the 5xx ones. We could do something fancy in our code (i.e., disable `consecutive_5xx_errors` when `consecutive_gateway_errors` is set, similar to the way we have implemented `consecutive_errors` today), but for simplicity I think we should just make Istio work the same as Envoy. The user will need to set both fields and understand that the codes overlap. That's the way Envoy users need to do it. We just need to make this all clear in the doc.

frankbu: Also, if we just make Istio work like Envoy does, maybe we should open an Envoy issue suggesting that it gateway errors should override 5xx, as I've suggested. That would be a more useful design, IMO. If they change the implementation, then Istio would get the same improvement in the future.

yingzhuivy: @frankbu If the user wants to disable 5xx, they could set consecutive_5xx_errors to 0. Because we are now using wrapper class, we could distinguish between 0 and unset.

yingzhuivy: I might have misunderstood your point, but I am not sure why would a user want to set the gateway greater than 5xx. In my mind, there are three cases:
1. user want to set 5xx only
2. user want to set gateway only
3. user want to set both gateway and 5xx (gateway should be smaller than 5xx otherwise it is same as case 1)

I think the only caveat is that in case 2, the user needs to disable 5xx explictly by setting it as 0

frankbu: Yes, I understand that by distinguishing between 0 and unset we can cover the 3 cases, but I think there is a valid 4th case that neither Envoy nor Istio support:

4. user wants to set gateway count to a higher number (e.g., 10) because they are likely to correct themselves but wants other 5xx errors (e.g., 500) to time out much earlier, for example after 1 or 2.

frankbu: Anyway, just to be clear, I think the Istio implementation should just do what Envoy does, whether they change to support case 4 or not.

frankbu: It also almost seems like a bug that if the user sets both 5xx to 5 and gateway to 10, that the gateway field is just ignored. If I set gateway to 10, I expect that to do something, instead of be ignore because it overlaps with another config parameter. That's why I think it might be worth opening an Envoy issue to see if they want to change their implementation.

rshriram: Lets go with what frank suggests for now!

frankbu:
@yingzhuivy It looks like we are now all in agreement on the API. Are you planning to update this PR accordingly, and then provide the implementation?

frankbu:
I think that all that needs to change in this PR is to correct the documented default values of the two new fields and to add a sentence in the `consecutive_gateway_errors` doc to say that its setting will have no effect if the value is greater than the value of `consecutive_5xx_errors`.

jhahn21:
@frankbu Yes, @yingzhuivy will update this PR as soon as she resolves an issue in connecting from China.

networking/v1alpha3/destination_rule.proto
yingzhuivy: done

networking/v1alpha3/destination_rule.proto
yingzhuivy: looking at envoy code, for TCP
connection timeout -> 504
connection failure -> 503
request failure -> 500

https://github.com/envoyproxy/envoy/blob/1a7b32de759939f82af2cf7e83471aff5d57c93d/source/common/upstream/outlier_detection_impl.cc#L96

updated the comment

yingzhuivy:
@frankbu @rshriram just updated the PR, please take a look thanks!

networking/v1alpha3/destination_rule.proto
```diff
@@ -522,7 +523,35 @@ message OutlierDetection {
...
-  int32 consecutive_errors = 1;
+  // $hide_from_docs
+  int32 consecutive_errors = 1 [deprecated=true];
+
+  // Number of gateway errors before a host is ejected from the connection pool.
+  // When the upstream host is accessed over HTTP, a 502, 503, or 504 return
+  // code qualifies as a gateway error. When the upstream host is accessed over
+  // an opaque TCP connection, connect timeouts and connection error/failure
+  // events qualify as a gateway error.
+  // If set to 0, this feature is disabled. Defaults to 0.
```
frankbu:
```suggestion
  // This feature is disabled by default or when set to the value 0.
  //
  // Note that
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -522,7 +523,35 @@ message OutlierDetection {
...
+  // if the value of consecutive_gateway_errors is greater than or equal to
+  // the value of consecutive_5xx_errors, consecutive_gateway_errors will have
+  // no effect.
+  google.protobuf.UInt32Value consecutive_gateway_errors = 6;
+
+  // Number of 5xx errors before a host is ejected from the connection pool.
+  // When the upstream host is accessed over an opaque TCP connection, connect
+  // timeouts, connection error/failure and request failure events qualify as a
+  // 5xx error.
+  // If set to 0, this feature is disabled. Defaults to 5.
```
frankbu:
```suggestion
  // This feature defaults to 5 but can be disabled by setting the value to 0.
  //
  // Note that
```

frankbu:
@yingzhuivy a couple of minor suggestions, but otherwise LGTM.

networking/v1alpha3/destination_rule.proto
yingzhuivy: changed.

one thing to note is that the generated json file doesn't include description after a blank line. not sure if that's intended or a bug.

geeknoid: @jasonwzm Was that intentional? I think it was, right?

PR #1190, opened 2019-12-04
Update policy.proto

sureshvis:
Even with permissive mode, the filter chain adds "require_client_certificate": true" and required you to pass client cert.


sureshvis:
closing this pr

PR #1192, opened 2019-12-04
Use annotations for CRD gen

jasonwzm:


jasonwzm:
Test fails because the order of the output is non-deterministic. Fixing it here: https://github.com/istio/tools/pull/605

PR #1194, opened 2019-12-05
Few small updates to the mesh config API

costinm:
Added a mechanism to specify the 'env' that gets injected - to be used instead of values.yaml.


costinm:
Happy with any name - whatever name is selected will also be used in the inject template for consistency.

mesh/v1alpha1/config.proto
```diff
@@ -208,7 +208,8 @@ message MeshConfig {
 
   // Unix Domain Socket through which Envoy communicates with NodeAgent SDS to get key/cert for mTLS.
   // Use secret-mount files instead of SDS if set to empty.
-  string sds_uds_path = 20;
+  // @deprecated - istiod agent will detect and send the path to envoy.
```
louiscryan: s/istiod agent/istio-agent

mesh/v1alpha1/proxy.proto
```diff
@@ -223,6 +224,10 @@ message ProxyConfig {
   // See [Metric Service](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/metrics/v2/metrics_service.proto)
   // for details about Envoy's Metrics Service API.
   RemoteService envoy_metrics_service = 23;
+
+  // Additional env variables to inject into sidecar.
+  // Names starting with ISTIO_META_ will be included in the generated bootstrap and sent to the XDS server.
+  map<string,string> meta = 24;
```
louiscryan: name is a little opaque, 'sidecar_env' ?

I don't have a strong opinion about this, otherwise. Shriram / Lin are likely better TOC reviewers than I.

costinm: sidecar_env sounds good - unless gRCP expects to use it as well.
Things like CA_ADDR will go here, might be relevant for sidecar-less. 

env ? It'll go under container env: section

costinm: To be clear: this is intended for break glass, vendor or advanced features - I will add caAddr as a top-level setting.

mesh/v1alpha1/config.proto
rshriram: Do not add docs here. It will show up in webpage. Just add a $hide_from_docs

mesh/v1alpha1/proxy.proto
rshriram: Proxy config is processed by the pilot agent before generating envoy bootstrap. It seems confusing to specify env vars here that will be consumed by platform (added to process env var) before pilot agent starts.

howardjohn: Does a user really care how it's used? It seems like more of an implementation detail - to the user it's all just configuring the proxy?

mesh/v1alpha1/config.proto
costinm: Done

mesh/v1alpha1/proxy.proto
costinm: The 'default' proxy has all the (default) settings for the sidecar. For now we do indeed do a crazy complicated thing - but the plan (according to the istiod doc) is to simplify things a bit. 

One option is to have Pilot return the template and mesh.yaml to agent, so agent doesn't need the crazy stuff. Longer term Pilot can do the rendering of the bootstrap - and return a final bootstrap, plus some env variables needed by pilot-agent.

But in the end - logically this is config for the proxy, and seems reasonable to keep it here. DiscoveryAddress is the same.

costinm:
PTAL. 

I have few other related changes so we can have injection without values.yaml and helm - trying to use small PRs to avoid problems...

costinm:
PTAL - I keep spending time resolving conflicts.

mesh/v1alpha1/proxy.proto
rshriram: let us call this proxy_metadata as it applies to gateways as well.

rshriram:
lgtm after renaming to proxy_metadata

mesh/v1alpha1/proxy.proto
costinm: Can we call it proxy_env ? Metadata is a subset - not all the env are passed to envoy bootstrap and node metadata inside. Some apply to pilot-agent/istio-agent itself. 

Not a big deal - I'll change it to proxy_metadata for now.

costinm: @rshriram - to be clear, the gateway can't easily support it (right now), it's used in injection. 

We are discussing adding a handler in Pilot to return the 'active' mesh config, and istio-agent using that - either in gateway or sidecar - instead of having the env and other options handled by injector. But in terms of implementation it'll take a bit longer, the use in injection is simpler.

PR #1195, opened 2019-12-06
update envoyfilter docs

rshriram:


networking/v1alpha3/envoy_filter.proto
```diff
@@ -37,22 +37,17 @@ import "networking/v1alpha3/sidecar.proto";
 // namespace](https://istio.io/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig),
 // followed by all matching EnvoyFilters in the workload's namespace.
 //
-// **NOTE 1**: Since this is break glass configuration, there will not
-// be any backward compatibility across different Istio releases. In
-// other words, this configuration is subject to change based on
-// internal implementation of Istio networking subsystem.
```
howardjohn: Why remove this?

mandarjog: I think we should strive to keep this api stable, but do we want to make firm commitments?

The other problem is that this is 2 APIs in one, a generic xds patch api and a “filter api”
The stability guarantees can be made about the outer API, but path and version management should be done by the user. This part we need to document thru examples.

howardjohn: Its not about the stability of the EnvoyFilter API itself, its about the stability of Pilot's XDS output. At the very minimum, once the XDS api changes, your EnvoyFilter breaks (this happened at least 4 times with telemetry v2, for example). This warning (by my interpretation) is not about the stability of the proto schema, but about the stability of the effect of the API

kyessenov: It's fairly difficult to qualify compatibility between versions. We would need to test all possible ways config is generated for patch compatibility. Even the choice of Any vs Struct breaks this patch for a single version I think. Unless we can find a way to say that certain patches are safe, it's hard to justify testing them in production.

kyessenov: My suggestion would be to carve out a subset of this API that is aligned with filterchain/filter discovery service xDS and ingress v2 extensions and commit to to that solely.

howardjohn: The XDS output that pilot produces is an implementation detail. Attempting to turn this into an API basically means we will never make any changes to pilot ever. If we want to commit to support something, we should make it first class, such that pilot has knowledge about the extension and can be smart about how to handle it for various versions (like all of our plugins)

kyessenov: @howardjohn Pilot cannot possible know about all the extensions since they are not always available at the time of the release. So the only job that Pilot can promise is _where_ to stick an extensions. That really restricts us to filter configs since that's the only part of Envoy that is actually dynamic (because of Any).

howardjohn: We could just have a very minimal set of things that can be added, such as "Add Filter API". Pilot knows where to put the filters for different versions of pilot, but doesn't need to know what the filter is. But having a completely generic xds patch api will not be backwards compatible

mandarjog: Using the EnvoyFilter api means that you are using xds directly .
Managing versions and compatibility of the inner proto is up to the user.
Since we are temporarily using it for telemetry v2 we are going thru that process.
We need to give them tools to manage it via proxy version, which we do.
What we have not done is to come up with a full story around how to do upgrades in those cases.

So out api stability is up to Istio. Inner API stability is up to xds.

kyessenov: The problem with that approach is that it fails as an API if you need N*M config files for N version of istio and M versions of xDS. It's not much better than compiling in it into pilot. It is be nice, however, to make changes to Pilot behavior without re-build, but there are better ways (scripting and so on).

rshriram: I removed this because we are still supporting the old envoy Filter API for backward compat reasons. But I agree with @howardjohn that the whole API is centered around Pilot's implementation details (the semantics of route/listener names). That said, I do not want to call this as Break Glass, especially when the stance is to use EnvoyFilter for all extensibility stuff from xff, use remote address, common Lua stuff, external authZ, etc. I will reword this to provide a "warning" when using the name fields to rely on the semantics.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -37,22 +37,22 @@ import "networking/v1alpha3/sidecar.proto";
...
+// **NOTE 1**: This API is deeply tied to the internal implementation
+// in Istio networking subsystem as well as Envoy's XDS API. While
+// every effort will be made to preserve backward compatibility where
+// possible, EnvoyFilter API objects should be carefully vetted across
+// Istio version upgrades to ensure that there are no
+// breakages. Specifically, any envoy configuration provided through
```
howardjohn: Should we recommend using proxyMatch on the version? This is what we are doing with telemetry v2

networking/v1alpha3/envoy_filter.proto
```diff
@@ -37,22 +37,22 @@ import "networking/v1alpha3/sidecar.proto";
...
+// this mechanism should be carefully monitored across Istio proxy
+// version upgrades, to ensure that deprecated fields are removed and
```
howardjohn: nit: its not just if we remove fields, we can change the whole behavior. Like when we moved to single inbound listener, and will likely do the same for outbound

networking/v1alpha3/envoy_filter.proto
```diff
@@ -37,22 +37,22 @@ import "networking/v1alpha3/sidecar.proto";
 // namespace](https://istio.io/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig),
 // followed by all matching EnvoyFilters in the workload's namespace.
 //
-// **NOTE 1**: Since this is break glass configuration, there will not
-// be any backward compatibility across different Istio releases. In
-// other words, this configuration is subject to change based on
-// internal implementation of Istio networking subsystem.
+// **NOTE 1**: This API is deeply tied to the internal implementation
```
mandarjog: Can we be more explicit?
There are 3 versions involved here

1. Proto of the EnvoyFilter itself and its evolution.
2. Structure of the xds output produced by pilot, so the patch target can be located.
3. The actual patch that patches the inner proto.

Different versions have different guarantees.
As for 2, so long as patch targets are identified by  hostnames, ports , metadata etc that is likely to be safer than otherwise

networking/v1alpha3/envoy_filter.proto
mandarjog: Yes, and explain that if the inner xds proto is experimental then it is crucial to use.

PR #1197, opened 2019-12-12
add StringMatcher in CorsPolicy

ZereChen:
AllowOrigin and AllowOriginRegex are deprecated in route.CorsPolicy. So we can use alloworiginStringMatch instead of them. because route.CorsPolicy has a variable alloworiginStringMatch,
i want to add  alloworiginStringMatch in networking.CorsPolicy. Exact is the same as AllowOrigin and regex is the same as AllowOriginRegex


ZereChen:
@googlebot I signed it!

linsun:
context:  https://godoc.org/github.com/cilium/proxy/go/envoy/api/v2/route#CorsPolicy.GetAllowOrigin

networking/v1alpha3/virtual_service.proto
```diff
@@ -1053,7 +1053,11 @@ message CorsPolicy {
   // The list of origins that are allowed to perform CORS requests. The
   // content will be serialized into the Access-Control-Allow-Origin
   // header. Wildcard * will allow all origins.
-  repeated string allow_origin = 1;
+  repeated string allow_origin = 1 [deprecated=true];
+  // The list of origins that are allowed to perform CORS requests. The
+  // content will be serialized into the Access-Control-Allow-Origin
```
louiscryan: The Access-Control-Allow-Origin does not allow regex's or other wildcards other than '*'

What is the use-case for something like an arbitrary regex or suffix match. Please include in the PR description

ZereChen: see: https://github.com/envoyproxy/data-plane-api/blob/master/envoy/api/v2/route/route.proto#L472

louiscryan: Sure, but then we need to update the documentation to match the intent described by

https://github.com/envoyproxy/envoy/issues/2526

ZereChen: @louiscryan ok, I've updated it. Look at it again

networking/v1alpha3/virtual_service.proto
```diff
@@ -1096,6 +1101,24 @@ message CorsPolicy {
   google.protobuf.BoolValue allow_credentials = 6;
 }
 
+message StringMatcher {
+  oneof match_pattern {
```
rshriram: We have a string match already in the api

ZereChen: @rshriram see:https://github.com/envoyproxy/data-plane-api/blob/master/envoy/type/matcher/string.proto
Because there's no suffix  and safe_regex in string match. So I created a new StringMatcher. StringMatcher in virtual_service.proto and StringMatcher in string.proto are individually matched to each other

howardjohn: we don't need a safe_regex field. This is an istio API, not envoy. we translate regex to safe regex.

Please let's not add another API for this

ZereChen: > .. just use the existing StringMatch object. Add a allow_origins (plural) with StringMatch
ok, I've updated it. Look at it again

ZereChen: @rshriram  @howardjohn  I've updated it. Look at it again

ZereChen:
@howardjohn @rshriram @louiscryan What should I do now?

networking/v1alpha3/virtual_service.proto
```diff
@@ -1074,7 +1074,12 @@ message CorsPolicy {
   // The list of origins that are allowed to perform CORS requests. The
   // content will be serialized into the Access-Control-Allow-Origin
   // header. Wildcard * will allow all origins.
-  repeated string allow_origin = 1;
+  repeated string allow_origin = 1[deprecated=true];
```
rshriram: add a $hide_from_docs in a separate comment line to hide this field.

ZereChen:
@linsun @louiscryan  @geeknoid What should I do?

PR #1200, opened 2019-12-13
update to individual owners - attempt to fix 2 approvers issue

linsun:
Discussed with @geeknoid on why our 2 approvers aren't working today.   @geeknoid suggested trying individual owners

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,4 @@
-*               @istio/technical-oversight-committee
+*                @linsun @rshriram @dcberg @louiscryan @geeknoid @smawson
 /Makefile*       @istio/wg-test-and-release-maintainers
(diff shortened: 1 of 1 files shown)


howardjohn:
What's the desired behavior? 2 TOC members or 1 TOC+1 other? The first is not possible as far as I know

howardjohn:
and the second seemed to be working form what I've seen

linsun:
In talking to @geeknoid @sdake my understanding is that TOC approved 2 approvers for API repo a while back, but it is not enforced today.

cc @istio/technical-oversight-committee

howardjohn:
It does require 2 approvers, but only one has to be in the TOC. At least it used to a week or 2 ago

PR #1203, opened 2019-12-16
Fix copy out of build container

ostromart:
There was a bug with the post-processed generated Go structs not being correctly copied out of the container into the repo. This led to:
```
go build ./mesh/v1alpha1/...
go: finding k8s.io/klog v1.0.0
go: finding golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
# istio.io/api/mesh/v1alpha1
mesh/v1alpha1/component.pb.go:175:11: undefined: BoolValueForPB
```
See also https://github.com/istio/istio/issues/19616


ostromart:
There were several additional problems to work through: started from scratch in https://github.com/istio/api/pull/1211

PR #1206, opened 2019-12-17
Deprecate obsolete fields in authN alpha policy.

diemtvu:
- Deprecate target selector. Only mesh & namespace level policy is supported.
- Deprecate origin authentication and related artifacts (JWT configuration, principal binding etc). It is replaced by [RequestAuthentication](https://github.com/istio/api/blob/master/security/v1beta1/request_authentication.proto).

Issue 

- https://github.com/istio/istio/issues/18399


authentication/v1alpha1/policy.proto
```diff
@@ -57,18 +58,21 @@ message MutualTls {
     PERMISSIVE = 1;
   };
 
-  // WILL BE DEPRECATED, if set, will translates to `TLS_PERMISSIVE` mode.
+  // $hide_from_docs
+  // if set, will translates to `TLS_PERMISSIVE` mode.
```
smawson:
```suggestion
  // if set, will translate to `TLS_PERMISSIVE` mode.
```

smawson: Also should you say deprecated here as well and point them to what they should do instead?

authentication/v1alpha1/policy.proto
```diff
@@ -230,10 +235,12 @@ message PeerAuthenticationMethod {
 
     // $hide_from_docs
     // Set if JWT is used. This option is not yet available.
```
smawson: It was never available? Can we reword this better?

authentication/v1alpha1/policy.proto
```diff
@@ -366,13 +376,16 @@ message Policy {
   // Leave the list empty if peer authentication is not required
   repeated PeerAuthenticationMethod peers = 2;
 
+  // $hide_from_docs
   // Set this flag to true to accept request (for peer authentication perspective),
```
smawson: Should say deprecated and indicate what to do instead.

authentication/v1alpha1/policy.proto
```diff
@@ -57,18 +58,21 @@ message MutualTls {
...
   // be left unset.
   // When the flag is false (default), request must have client certificate.
-  bool allow_tls = 1;
+  bool allow_tls = 1 [deprecated=true];
 
   // Defines the mode of mTLS authentication.
   Mode mode = 2;
 }
 
+// $hide_from_docs
```
louiscryan: @jasonwzm do we hide deprecated resources in generated docs or do we mark them as 'deprecated' in the docs somehow ?

jasonwzm: we don't hide the deprecated fields now, instead we just show them at the end of the fields in the docs. We will be better at documenting deprecated fields when we have multi versions of the APIs.

authentication/v1alpha1/policy.proto
diemtvu: Done

authentication/v1alpha1/policy.proto
diemtvu: Done

authentication/v1alpha1/policy.proto
diemtvu: Done

authentication/v1alpha1/policy.proto
diemtvu: I've reverted some $hide_from_doc so that it still shows in doc (though hope it will be at the end, like Jason said). I think it is better for 1.4 -> 1.5 users know what happen with those fields.

geeknoid: I think deprecated fields are also rendered in a different color, if I recall correctly.

diemtvu:
@louiscryan @smawson Can you PTAL? Thanks.

authentication/v1alpha1/policy.proto
```diff
@@ -242,6 +249,9 @@ message OriginAuthenticationMethod {
   Jwt jwt = 1;
 }
 
+// $hide_from_docs
+// Deprecated. When using security/v1beta1/RequestAuthentication, the request principal is always
```
smawson:
```suggestion
// Deprecated. When using security/v1beta1/RequestAuthentication, the request principal always
```

authentication/v1alpha1/policy.proto
```diff
@@ -242,6 +249,9 @@ message OriginAuthenticationMethod {
...
+// come from request authentication (i.e JWT).
```
smawson:
```suggestion
// comes from request authentication (i.e. JWT).
```

authentication/v1alpha1/policy.proto
smawson: There are still a lot of fields marked hide from docs, have you not updated or are you intentionally hiding some but not others?

diemtvu:
I unhide the deprecated fields, but keep the definition of those deprecated message hidden, as I think it's cleaner for the doc. Let me know if you think otherwise.

authentication/v1alpha1/policy.proto
diemtvu: Do you mean the one on line 74? It's for the JWT message, so should be hidden. And it's also commented deprecated.

diemtvu:
ping @louiscryan

PR #1210, opened 2019-12-18
Put back doc for gateways field in http match

frankbu:
Fixes https://github.com/istio/istio/issues/17105

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -704,12 +704,12 @@ message HTTPMatchRequest {
 
   // One or more labels that constrain the applicability of a rule to
   // workloads with the given labels. If the VirtualService has a list of
-  // gateways specified at the top, it must include the reserved gateway
+  // gateways specified in the top-level `gateways` field, it must include the reserved gateway
   // `mesh` for this field to be applicable.
   map<string, string> source_labels = 7;
 
   // Names of gateways where the rule should be applied to. Gateway names
-  // at the top of the VirtualService (if any) are overridden. The gateway
+  // in the top-level `gateways` field of the VirtualService (if any) are overridden. The gateway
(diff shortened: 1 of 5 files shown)


hzxuzhonghu:
@frankbu Personally, i think it is good to remove the gateway field eventually, it is weird to allow a VS binding to more than one gateway. We should treat this as a conflict.

dcberg:
I agree that it may not be common to have a VS assigned to multiple gateways but it is not weird. For example you may have a public and private ingress gateways and expose an overlapping set of services to both. In this case VS’s would be bound to multiple gateways.

networking/v1alpha3/virtual_service.proto
```diff
@@ -708,7 +708,9 @@ message HTTPMatchRequest {
   // `mesh` for this field to be applicable.
   map<string, string> source_labels = 7;
 
-  // $hide_from_docs
+  // Names of gateways where the rule should be applied to. Gateway names
+  // at the top of the VirtualService (if any) are overridden. The gateway
```
geeknoid: What does "at the top" mean? At the top of the yaml?

frankbu: Yes, at the top level in the VS yaml. I'll reword it more clearly.

frankbu:
Actually, @hzxuzhonghu might have a point about the match-specific gateways field being "weird" in the sense that it overrides the top-level gateways field of the VS. For example, it would be weird if the VS had no gateways listed at the top-level, but every rule had a specific gateway binding. In that case you couldn't tell what gateways a VS is bound to without looking at the entire structure. It probably would be less weird if we restricted the match-specific gateways field to be a subset of the top-level gateways field, but that would be a breaking change. Bottom line is that it's not ideal, but probably still not worth changing.

PR #1211, opened 2019-12-19
Fixes for synthetic and k8s types

ostromart:


mesh/v1alpha1/proxy.pb.go
```diff
@@ -1,16 +1,15 @@
...
 
 package v1alpha1
 
 import (
 	fmt "fmt"
-	proto "github.com/gogo/protobuf/proto"
-	types "github.com/gogo/protobuf/types"
-	io "io"
+	proto "github.com/golang/protobuf/proto"
+	duration "github.com/golang/protobuf/ptypes/duration"
```
yingzhuivy: got multiple errors after updating api, one example: 
https://github.com/istio/istio/blob/master/pkg/config/validation/validation.go#L1137
complaining type inconsistency between google protobuf and gogo protobuf

is this intended? @ostromart

ostromart: no, it's not intended - could you open an issue with details?

howardjohn: See my revert PR https://github.com/istio/api/pull/1215

You switched mesh config from gogo proto to golang protobuf. These are not compatible, so you see failures in 100s of files in istio/istio. If you diligently go through and fix all of them (if you are going down this route let me know, I fixed most of them), you will eventually find it wont work because mesh config references other Istio APIs which are gogo protos, and you can't mix the two.

PR #1212, opened 2019-12-19
Fix link to attribute vocabulary

nacx:
Fix the link to the attribute vocabulary from the Authorization Policy Conditions.


nacx:
This has already been fixed and this is no longer needed.

PR #1213, opened 2019-12-20
[WIP] Initial v1alpha1 Troubleshooting API

incfly:
Design doc [link](https://docs.google.com/document/d/11P6OXlITExJvFqzsWfFu9WhFLjY6uqOcCUKqPVgSSoc/edit#)

API design [link](https://docs.google.com/document/d/1g_ZGOtSJ-DqTlpkiEJHTK_U9JvvTnco3h6rsDK-sOvA/edit#)


rlenglet:
> i removed the timeout_duration in the response. since i expect this can be set at gRPC library runtime config, rather than the payload itself.

@incfly You need to add it back. There's a timeout that can be set in a gRPC _request_, but this is not a request, it's technically a gRPC _response_.

rlenglet:
> ManagedNode GetServer methods are not part of the rpc service since we maybe ale to do it in istioctl client side.

This is still under discussion. Keep it there for consistency with the current design doc.

rlenglet:
> renamed the NodeType to ProxyType.

"NodeType" is an Istio term.

incfly:
timeout is added back, and changed back to nodetype (my original concern is that this confuses users, but given user facing UX is istoctl, which makes it less of a problem)

i'm still not quite sure about the `GetServers`. For an API, I think it's generally easier to add compared with remove. I'm not strong about this, I can add that back if  needed.

troubleshooting/v1alpha1/troubleshooting.proto
```diff
@@ -0,0 +1,145 @@
...
+  repeated LoggerLogLevel logger_levels = 2;
+}
+
+message SetLogLevelResponse {
+  // ManagedNode managed_node = 1;
+  repeated LoggerLogLevel logger_levels = 2;
+}
+
+// ProxyTroubleshootingService is exposed by the proxy management component on each node.
+// Currently this is pilot.
```
rshriram: ?? The agreement was that this won’t be pilot but a separate component. Either way nuke this statement.

troubleshooting/v1alpha1/troubleshooting.proto
```diff
@@ -0,0 +1,145 @@
...
+  repeated NodeType types = 1;
+
+  // The namespaces of the proxy reside in.
+  repeated string namespaces = 2;
+
+  // the names of the proxy to be selected. For Kubernetes, this is the pod id.
+  repeated string names = 3;
+
+  // Selector of proxy based on workload labels.
+  istio.type.v1beta1.WorkloadSelector selector = 4;
```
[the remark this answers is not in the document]
incfly: sorry, i missed this. start from what? what're proposed change, as opposed to use workload selector?

troubleshooting/v1alpha1/troubleshooting.proto
incfly: I can remove the comment.

From the [conversation](https://docs.google.com/document/d/1g_ZGOtSJ-DqTlpkiEJHTK_U9JvvTnco3h6rsDK-sOvA/edit?disco=AAAAD_Z2F60), it's seems whether it's istiod by itself is pending on authz decision, which now is resolved given the aggregation api.

My preference:

- We can't make this on and off at users' choices. Otherwise, something wrong, need troubleshooting information, you don't know which set of pods/namespaces need to be troubleshooted in advance. This has to be on all the time. The rollout phase can span multi release, but the end goal should be always on to make itself actual useful.
- Separate deployment or as same as istiod. but from implementation point of view, I guess we'll share the code base, deployment model can be different.

@rlenglet thoughts || confirm?

troubleshooting/v1alpha1/troubleshooting.proto
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+// ProxySelector specifies the proxy filtering of the troubleshooting API targetted scope.
+// All the fields in this selector are and-ed together. Empty fields will selector all the
+// proxy.
+message ProxySelector {
+  // The type of the proxy runs as.
+  enum NodeType { SIDECAR = 0; ROUTER = 1; }
+
```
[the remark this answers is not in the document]
incfly: should plan to make this will an user facing API @rlenglet  e.g. 3rd party vendors developping their software based on this API later on.

if so, yeah, making sense to change to `GATEWAY` to avoid confusion, but then we need to doc/comment for ourselves, "this maps ROUTER internally".

troubleshooting/v1alpha1/troubleshooting.proto
rlenglet: This comment is incorrect. There is no service exposed by the agent in the current design.

incfly: done. reworded.

ramaraochavali:
FWIW, there is a new API being proposed in XDS https://github.com/envoyproxy/envoy/pull/9383 - This seems similar to what is being done here.

troubleshooting/v1alpha1/troubleshooting.proto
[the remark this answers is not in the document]
hzxuzhonghu: IMO, we can get the specific proxy by ns.name. Why do we need other fields?

troubleshooting/v1alpha1/troubleshooting.proto
```diff
@@ -0,0 +1,150 @@
...
+  istio.type.v1beta1.WorkloadSelector selector = 4;
+}
+
+// GetConfigDumpRequest is the request of GetConfigDump.
+message GetConfigDumpRequest{
+  // Selector selects a subset of the proxies for troubleshooting.
+  ProxySelector selector = 1;
+}
+
+// GetConfigDumpResponse is the request of GetConfigDump.
```
hzxuzhonghu: response

troubleshooting/v1alpha1/troubleshooting.proto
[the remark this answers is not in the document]
hzxuzhonghu: Is there a need to support for multiple proxies? What's the config dump response for multiple proxies?

rlenglet:
> FWIW, there is a new API being proposed in XDS envoyproxy/envoy#9383 - This seems similar to what is being done here.

The whole troubleshooting API can't be done in Envoy. Using xDS to Envoy is not the solution.

troubleshooting/v1alpha1/troubleshooting.proto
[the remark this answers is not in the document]
rlenglet: > Is there a need to support for multiple proxies? 

Yes.

> What's the config dump response for multiple proxies?

One response per proxy.

howardjohn:
What is the "whole troubleshooting API"?In the doc it's just config dump (obviously can be done by envoy) and changing log level (which can be done through runtime Discovery afaik), are there others?

rlenglet:
> What is the "whole troubleshooting API"?In the doc it's just config dump (obviously can be done by envoy) and changing log level (which can be done through runtime Discovery afaik), are there others?

The scope of the whole troubleshooting API is specified in this design proposal, which you already reviewed and commented on: https://docs.google.com/document/d/11P6OXlITExJvFqzsWfFu9WhFLjY6uqOcCUKqPVgSSoc/edit#heading=h.h4dnnbkwzgzw

This PR is specifying the 1st version of the API, which is a subset of the whole API, as specified in the detailed doc which you also already reviewed and commented on:
https://docs.google.com/document/d/1g_ZGOtSJ-DqTlpkiEJHTK_U9JvvTnco3h6rsDK-sOvA/edit#heading=h.fxpk50cps4zs

howardjohn:
@rlenglet I understand, what I am confused is what part of that cannot be solved in envoy

troubleshooting/v1alpha1/troubleshooting.proto
incfly: done.

troubleshooting/v1alpha1/troubleshooting.proto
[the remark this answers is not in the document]
incfly: changed to gateway. will have impl notes to say this match to router internally. fyi, @rlenglet

troubleshooting/v1alpha1/troubleshooting.proto
incfly: done.

PR #1214, opened 2019-12-20
Remove some fields in virtualservice

hzxuzhonghu:
Fields like `  "websocket_upgrade", "append_headers", "remove_response_headers", "append_response_headers", "remove_request_headers", "append_request_headers"`, etc were mostly  deprecated more than one year. It is safe to remove them.

And it is a pre-cleanup work for VS chaining.


howardjohn:
we will deny unknown fields now, so should we really remove them?

networking/v1alpha3/virtual_service.proto
```diff
@@ -891,7 +865,7 @@ message TLSMatchAttributes {
   // IPv4 or IPv6 ip address of source with optional subnet. E.g., a.b.c.d/xx
   // form or just a.b.c.d
   // $hide_from_docs
-  string source_subnet = 4;
+  string source_subnet = 4 [deprecated=true];
```
rshriram: just remove it please

hzxuzhonghu: ok

hzxuzhonghu:
>we will deny unknown fields now, so should we really remove them?

@howardjohn That is to say, we donot deny now. Also they are deprecated a year ago, So i think it is safe to remove them. BTW

howardjohn:
we want to support multiple versions of control plane running at once
since validation is a Singleton, that means we need to make just one of those do validation, which means it should he backwards compatible. This means we shouldn't delete fields

howardjohn:
@jasonwzm

rshriram:
What does control plane version have to do with removal of deprecated fields? If the support for those fields have ended, there is no point in keeping these fields around. Second, this fear of removal will then prevent us from adding anything to any Api (including all existing api prs) and changing anything in the api as at some point in future if we were to remove these fields due to deprecation, we can’t do so. So might as well not add anything at all and do everything with backdoors like environment variable hacks.

howardjohn:
Kubernetes has never removed a field from an API and doesn't do everything with environment variable hacks. This is the entire reason we have API versions. If we want to remove a field, we should increment the API version

jasonwzm:
According to [Kubernetes](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning), alpha version APIs could change in incompatible ways, including removing fields. According to our versioning plan, we plan to have alpha support of multi version different schema of an API in 1.5 and in 1.6 this will be fully supported. If this can wait util 1.6, we can clean up the deprecated fields by incrementing the version and we are able to provide a conversion story for the users.

Alternatively, we can mark the fields as deprecated (as done in https://github.com/istio/api/pull/1206) and our docs will reflect that. Then in 1.6, we increment the version and remove the deprecated fields. I think it is the better approach as it provides time of a release for users to move away from the deprecated fields.

hzxuzhonghu:
@jasonwzm Good link, the fields here were deprecated long ago.

rshriram:
Yes. We can keep the fields around but I don’t want to keep the implementation around. It complicated the implementation and creates a lot of tech debt. These changes were made a long time ago. Even envoy’s api fields in v2 underwent changes - deleted field numbers were marked as reserved. I don’t see a convincing reason for this objection given we are simply removing minor fields deprecated two or three releases ago (this is very much in line with our api policies).

hzxuzhonghu:
@howardjohn Any other concern?

howardjohn:
what is the benefit of removing the fields rather than flowing what Jason suggested?

hzxuzhonghu:
It tells users not to use deprecated fields, one day they will definitely be removed, change to new recommended ones ASAP.

hzxuzhonghu:
I think it is similar to the current implementation in pilot when bumping to new envoy but have a limited older versions support.

dgn:
> According to [Kubernetes](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning), alpha version APIs could change in incompatible ways, including removing fields. 

Note that in Kubernetes, alpha APIs are used to acces alpha functionality, whereas in Istio, core functionality of 1.x releases is only available through alpha APIs - so they're not really alpha. I don't think we should treat them as such, given they're being used in production right now.

> According to our versioning plan, we plan to have alpha support of multi version different schema of an API in 1.5 and in 1.6 this will be fully supported. If this can wait util 1.6, we can clean up the deprecated fields by incrementing the version and we are able to provide a conversion story for the users. Alternatively, we can mark the fields as deprecated (as done in #1206) and our docs will reflect that. Then in 1.6, we increment the version and remove the deprecated fields. I think it is the better approach as it provides time of a release for users to move away from the deprecated fields.

I like this. Let's deprecate and remove in the next version.

hzxuzhonghu:
@dgn Mostly i would agree with you.  But if you look carefully, these fields were hidden or deprecated long ago.

rshriram:
Yes. @dgn we actually support deprecated fields for a very long time (in this case, about a year!!). It is not even documented or visible in the docs site if you look at istio.io/docs. So, we are really just following the proper practices and removing deprecated fields after 3+ releases :).

PR #1215, opened 2019-12-22
Revert "Fixes for synthetic and k8s types"

howardjohn:
Reverts istio/api#1211

This change requires massive changes to istio/istio -- at least 1000 lines changed. It should not be merged unless we have a corresponding istio/istio PR to update. 

I am pretty sure its not possible to do so anyways though. The change modifies mesh config to be a gogo protobuf. However, mesh config references other APIs, which are in gogo protobuf, which doesn't work.

We have the same problem with the new operator config, which is a golang protobuf which references mesh config, which is why the original change was made. I don't think its possible to change mesh config without also changing DestinationRule, which likely means changing every single one of our APIs to golang protobuf, which is not acceptable due to performance impact in Mixer (and, it would require a lot of changes to our code)


ostromart:
See https://github.com/istio/api/pull/1216.

PR #1219, opened 2019-12-27
Add kubeType annotation for mixer CRD client registration

jasonwzm:
This is the API annotation change to fix https://github.com/istio/istio/issues/19638


howardjohn:
if you cherry pick to 1.4 you'll need to do the same in tools?

jasonwzm:
> if you cherry pick to 1.4 you'll need to do the same in tools?

Yes, will do it in another PR.

PR #1221, opened 2019-12-30
sketch of an extension manifest

kyessenov:
API to add envoy filters to istio:
- restricted to filters (as the core envoy extension mechanism), major reason why EnvoyFilter is not adequate (since it allows everything).
- no merging of protobufs (too expensive in CPU)
- basic admin control over the manifests using the filter name as the primary key (will be free-form once work in https://github.com/envoyproxy/envoy/pull/9391 completed)
- basic resolution semantics to place filters in chains
- match conditions are not fully specified since they are likely to evolve
- extension configuration can be white-listed with a separate extension resource and a config reference

cc @mandarjog @howardjohn @louiscryan @rshriram 

### DIFF
--- a/extension/v1alpha1/manifest.proto
+++ b/extension/v1alpha1/manifest.proto
@@ -50,7 +50,9 @@ option go_package = "istio.io/api/extension/v1alpha1";
 import "google/protobuf/struct.proto";
 
 // Opaque configuration for an extension (wire and JSON compatible with UDPA).
-message TypedStruct {
+// The extension resource allows white-listing pre-defined configuration for
+// extensions.
+message Extension {
   // A URL that uniquely identifies the type of the serialize protocol buffer
   // message. This has same semantics and format described in
   // google.protobuf.Any:
@@ -59,6 +61,12 @@ message TypedStruct {
 
   // A JSON representation of the above specified type.
   google.protobuf.Struct value = 2;
+}
+
+// Reference to extension resource.
+message ExtensionRef {
+  string name = 1;
+  string namespace = 2;
 }
 
 // A condition limiting the extent of the extension application.
@@ -182,14 +190,19 @@ message ExtensionManifest {
     string name = 3;
 
     // Opaque configuration for the filter.
-    TypedStruct config = 4;
+    oneof configuration {
+      // Inlined extension configuration.
+      Extension config = 4;
+      // Configuration reference.
+      ExtensionRef config_ref = 5;
+    }
 
     // Modification operation (function is optional in case of replacement or
     // removal).
-    Operation operation = 5;
+    Operation operation = 6;
 
     // Purpose of the extension used to place the filter in a chain.
-    Function function = 6;
+    Function function = 7;
   };
 
   // A list of extensions to be applied. The filter extensions are processed in
--- a/extension/v1alpha1/manifest.proto
+++ b/extension/v1alpha1/manifest.proto
@@ -35,7 +35,7 @@ syntax = "proto3";
 //       context: SIDECAR_OUTBOUND
 //       portNumber: 8080
 //     type: HTTP_FILTER
-//     name: envoy.filters.http.gzip
+//     name: compress-outbound
 //     config:
 //       typeUrl: config.filter.http.gzip.v2.Gzip
 //       value:
@@ -49,7 +49,7 @@ option go_package = "istio.io/api/extension/v1alpha1";
 
 import "google/protobuf/struct.proto";
 
-// Opaque configuration for an extension (wire-compatible with UDPA).
+// Opaque configuration for an extension (wire and JSON compatible with UDPA).
 message TypedStruct {
   // A URL that uniquely identifies the type of the serialize protocol buffer
   // message. This has same semantics and format described in
@@ -121,12 +121,16 @@ message MatchContext {
   // namespace, it will be applied to all applicable workloads in any
   // namespace.
   map<string, string> workload_labels = 8;
+
+  // Namespace of the proxy.
+  string namespace = 9;
 }
 
 // ExtensionManifest specifies a set of custom extensions.
-// Manifests can be applied in any namespace, in which case, a per-namespace
-// manifest overrides the root manifest of the same name. Beyond that, the
-// order of application of the manifests is unspecified.
+// Manifests can be applied in any namespace. Per-namespace manifests
+// may override the root manifests for the overlapping filter names if
+// both match conditions are satisfied.
+// Beyond that, the order of application of the manifests is unspecified.
 // Additional control over the delegation of the extension injection should
 // be enforced with either resource-level RBAC or an admission controller.
 message ExtensionManifest {
@@ -139,7 +143,19 @@ message ExtensionManifest {
     LISTENER_FILTER = 4;
   };
 
-  // Extension function serving as a hint to determine the order of the
+  // Operation performed with the supplied config when the match condition is
+  // satisfied. By default, the operation is to displace the filter with the
+  // same name. If no filter with the same name is present, then the filter is
+  // added to chain using the provided function hint to locate the position of
+  // insertion. If the operation is to remove, then the filter with the same
+  // name is removed (config is optional). If no filter with the same name is
+  // found then, then it is a no-op.
+  enum Operation {
+    ADD = 0;
+    REMOVE = 2;
+  };
+
+  // Extension purpose serving as a hint to determine the order of the
   // application of an extension in a chain of extensions. Istio recognizes the
   // following categories of extensions and applies them in the order specified
   // here.
@@ -151,39 +167,34 @@ message ExtensionManifest {
     POST_PROCESSING = 4;
   };
 
-  // Allows adding, replacing or removing of the extension by name. Note that
-  // for replacement, the function field is optional, and for removal, the
-  // config field is optional.
-  enum Operation {
-    ADD = 0;
-    REPLACE = 1;
-    REMOVE = 2;
-  };
-
   // Extension description for a data plane filter.
   message FilterExtension {
-    // Conditions to limit the scope of the extension.
-    // Any of the conditions must be satisfied for the filter to be applied.
+    // Optional conditions to limit the scope of the extension.
+    // At least one of the conditions must be satisfied for the filter to be
+    // applied.
     repeated MatchContext match = 1;
 
     // Specifies the type of the filter.
     FilterType type = 2;
 
-    // The name of the filter (corresponding to the envoy extension name).
+    // The required name of the filter instance. The name should generally be
+    // unique in a filter chain.
     string name = 3;
 
     // Opaque configuration for the filter.
     TypedStruct config = 4;
 
-    // Purpose of the extension.
-    Function function = 5;
-
     // Modification operation (function is optional in case of replacement or
     // removal).
-    Operation operation = 6;
+    Operation operation = 5;
+
+    // Purpose of the extension used to place the filter in a chain.
+    Function function = 6;
   };
 
-  // A list of extensions to be applied. If the order of extensions is
-  // ambiguous, then this list serves as the canonical order of application.
+  // A list of extensions to be applied. The filter extensions are processed in
+  // the order specified here. If multiple extension conditions are satisfied
+  // for the same filter name within a manifest, then the last filter config
+  // applies.
   repeated FilterExtension filters = 1;
 }
--- a/extension/v1alpha1/manifest.proto
+++ b/extension/v1alpha1/manifest.proto
@@ -1,0 +1,189 @@
+// Copyright 2019 Istio Authors
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+// $schema: istio.extension.v1alpha1.ExtensionManifest
+// $title: Extension Manifest
+
+// `ExtensionManifest` specifies additional extensions to be
+// applied in a service mesh. Extensions are injected into the
+// data plane processing based on a set of criteria, such as
+// the workload labels, service ports, protocols, etc.
+//
+// ```yaml
+// apiVersion: extension.istio.io/v1alpha1
+// kind: ExtensionManifest
+// metadata:
+//   name: zip
+//   namespace: istio-config
+// spec:
+//   filters:
+//   - match:
+//       proxyVersion: '1\.4.*'
+//       context: SIDECAR_OUTBOUND
+//       portNumber: 8080
+//     type: HTTP_FILTER
+//     name: envoy.filters.http.gzip
+//     config:
+//       typeUrl: config.filter.http.gzip.v2.Gzip
+//       value:
+//         compression_level: BEST
+//     function: POST_PROCESSING
+//```
+//
+package istio.extension.v1alpha1;
+
+option go_package = "istio.io/api/extension/v1alpha1";
+
+import "google/protobuf/struct.proto";
+
+// Opaque configuration for an extension (wire-compatible with UDPA).
+message TypedStruct {
+  // A URL that uniquely identifies the type of the serialize protocol buffer
+  // message. This has same semantics and format described in
+  // google.protobuf.Any:
+  // https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto
+  string type_url = 1;
+
+  // A JSON representation of the above specified type.
+  google.protobuf.Struct value = 2;
+}
+
+// A condition limiting the extent of the extension application.
+message MatchContext {
+  // A regular expression in golang regex format (RE2) that can be
+  // used to select proxies using a specific version of istio
+  // proxy. The Istio version for a given proxy is obtained from the
+  // node metadata field ISTIO_VERSION supplied by the proxy when
+  // connecting to Pilot. This value is embedded as an environment
+  // variable (ISTIO_META_ISTIO_VERSION) in the Istio proxy docker
+  // image. Custom proxy implementations should provide this metadata
+  // variable to take advantage of the Istio version check option.
+  string proxy_version = 1;
+
+  // Match on the node metadata supplied by a proxy when connecting
+  // to Istio Pilot. Note that while Envoy's node metadata is of
+  // type Struct, only string key-value pairs are processed by
+  // Pilot. All keys specified in the metadata must match with exact
+  // values. The match will fail if any of the specified keys are
+  // absent or the values fail to match.
+  map<string, string> metadata = 2;
+
+  // Configuration type.
+  enum ProxyContext {
+    UNSPECIFIED = 0;
+    SIDECAR_INBOUND = 1;
+    SIDECAR_OUTBOUND = 2;
+    GATEWAY = 3;
+  };
+
+  // Match on the proxy type (applicable for network and HTTP filters).
+  ProxyContext context = 3;
+
+  // The service port of the upstream or downstream configuration resource. If
+  // omitted, applies to clusters and listeners for any port.
+  uint32 port_number = 4;
+
+  // The Istio gateway config's namespace/name for which this route
+  // configuration was generated. Applies only if the context is GATEWAY. Should
+  // be in the namespace/name format. Use this field in conjunction with the
+  // portNumber to accurately select the Envoy route configuration for a
+  // specific HTTPS server within a gateway config object.
+  string gateway = 5;
+
+  // The fully qualified service name for this cluster or listener. For
+  // services defined through service entries, the service name is same as the
+  // hosts defined in the service entry.
+  string service = 6;
+
+  // The subset associated with the service. If omitted, applies to
+  // clusters for any subset of a service.
+  string subset = 7;
+
+  // Criteria used to select the specific set of pods/VMs on which
+  // this patch configuration should be applied. If omitted, the set
+  // of patches in this configuration will be applied to all workload
+  // instances in the same namespace.  If omitted, the EnvoyFilter
+  // patches will be applied to all workloads in the same
+  // namespace. If the EnvoyFilter is present in the config root
+  // namespace, it will be applied to all applicable workloads in any
+  // namespace.
+  map<string, string> workload_labels = 8;
+}
+
+// ExtensionManifest specifies a set of custom extensions.
+// Manifests can be applied in any namespace, in which case, a per-namespace
+// manifest overrides the root manifest of the same name. Beyond that, the
+// order of application of the manifests is unspecified.
+// Additional control over the delegation of the extension injection should
+// be enforced with either resource-level RBAC or an admission controller.
+message ExtensionManifest {
+  // Extension filter type.
+  enum FilterType {
+    UNSPECIFIED_FILTER_TYPE = 0;
+    HTTP_FILTER = 1;
+    NETWORK_FILTER = 2;
+    UPSTREAM_NETWORK_FILTER = 3;
+    LISTENER_FILTER = 4;
+  };
+
+  // Extension function serving as a hint to determine the order of the
+  // application of an extension in a chain of extensions. Istio recognizes the
+  // following categories of extensions and applies them in the order specified
+  // here.
+  enum Function {
+    UNSPECIFIED_FUNCTION = 0;
+    TELEMETRY = 1;
+    PRE_PROCESSING = 2;
+    AUTHORIZATION = 3;
+    POST_PROCESSING = 4;
+  };
+
+  // Allows adding, replacing or removing of the extension by name. Note that
+  // for replacement, the function field is optional, and for removal, the
+  // config field is optional.
+  enum Operation {
+    ADD = 0;
+    REPLACE = 1;
+    REMOVE = 2;
+  };
+
+  // Extension description for a data plane filter.
+  message FilterExtension {
+    // Conditions to limit the scope of the extension.
+    // Any of the conditions must be satisfied for the filter to be applied.
+    repeated MatchContext match = 1;
+
+    // Specifies the type of the filter.
+    FilterType type = 2;
+
+    // The name of the filter (corresponding to the envoy extension name).
+    string name = 3;
+
+    // Opaque configuration for the filter.
+    TypedStruct config = 4;
+
+    // Purpose of the extension.
+    Function function = 5;
+
+    // Modification operation (function is optional in case of replacement or
+    // removal).
+    Operation operation = 6;
+  };
+
+  // A list of extensions to be applied. If the order of extensions is
+  // ambiguous, then this list serves as the canonical order of application.
+  repeated FilterExtension filters = 1;
+}


extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+  // for replacement, the function field is optional, and for removal, the
+  // config field is optional.
+  enum Operation {
+    ADD = 0;
+    REPLACE = 1;
+    REMOVE = 2;
+  };
+
+  // Extension description for a data plane filter.
+  message FilterExtension {
```
kyessenov: @htuch : I'm curious to hear your thoughts on this API sketch to globally apply a filter across xDS configs.

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+message MatchContext {
+  // A regular expression in golang regex format (RE2) that can be
+  // used to select proxies using a specific version of istio
+  // proxy. The Istio version for a given proxy is obtained from the
+  // node metadata field ISTIO_VERSION supplied by the proxy when
+  // connecting to Pilot. This value is embedded as an environment
+  // variable (ISTIO_META_ISTIO_VERSION) in the Istio proxy docker
+  // image. Custom proxy implementations should provide this metadata
+  // variable to take advantage of the Istio version check option.
+  string proxy_version = 1;
```
howardjohn: should we consider making this semver instead of regex?

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+  // Extension description for a data plane filter.
+  message FilterExtension {
+    // Conditions to limit the scope of the extension.
+    // Any of the conditions must be satisfied for the filter to be applied.
+    repeated MatchContext match = 1;
+
+    // Specifies the type of the filter.
+    FilterType type = 2;
+
+    // The name of the filter (corresponding to the envoy extension name).
```
kyessenov: NB: this will become free form (and generally useful for replace/removal).

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+  // node metadata field ISTIO_VERSION supplied by the proxy when
+  // connecting to Pilot. This value is embedded as an environment
+  // variable (ISTIO_META_ISTIO_VERSION) in the Istio proxy docker
+  // image. Custom proxy implementations should provide this metadata
+  // variable to take advantage of the Istio version check option.
+  string proxy_version = 1;
+
+  // Match on the node metadata supplied by a proxy when connecting
+  // to Istio Pilot. Note that while Envoy's node metadata is of
+  // type Struct, only string key-value pairs are processed by
```
howardjohn: do we expect this to be a problem at anytime in the future?

mandarjog:
@kyessenov can you update the design doc and discussion about the analysis of this API, otherwise the same question will have to rehashed here.

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,175 @@
...
+// Opaque configuration for an extension (wire-compatible with UDPA).
+message TypedStruct {
+  // A URL that uniquely identifies the type of the serialize protocol buffer
+  // message. This has same semantics and format described in
+  // google.protobuf.Any:
+  // https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto
+  string type_url = 1;
+
+  // A JSON representation of the above specified type.
+  google.protobuf.Struct value = 2;
```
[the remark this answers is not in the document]
kyessenov: Yes, envoy is using typed struct. Any is not usable in k8s since it's JSON based. Pilot can convert to any internally once it leaves JSON world and it has the descriptor.

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,175 @@
...
+// $title: Extension Manifest
+
+// `ExtensionManifest` specifies additional extensions to be
+// applied in a service mesh. Extensions are injected into the
+// data plane processing based on a set of criteria, such as
+// the workload labels, service ports, protocols, etc.
+//
+// ```yaml
+// apiVersion: extension.istio.io/v1alpha1
+// kind: ExtensionManifest
```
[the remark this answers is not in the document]
kyessenov: We have to support EnvoyFilter as long as it is used. If this API gets more usage we can consider deprecation.

extension/v1alpha1/manifest.proto
kyessenov: Sounds good to me.

extension/v1alpha1/manifest.proto
kyessenov: I'm not sure why it is the way it is but it sounds like it should be consistent throughout Istio. This is a verbatim copy from similar APIs.

extension/v1alpha1/manifest.proto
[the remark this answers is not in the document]
howardjohn: Give the major overlap would it make sense to have these APIs merged in some manner? This is effectively a subset of EnvoyFilter, right? I am not sure if there is any prior art to have APIs that are a subset of another API as a standalone resource?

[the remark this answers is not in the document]
mandarjog: It is not a strict subset. Crucially envoyfilter API misses filter ordering.
EnvoyFilter API has become a generic xds patch API, even though the name suggests otherwise. We are explicitly trying *not* to be generic xds patching here and focus on filters only.

extension/v1alpha1/manifest.proto
howardjohn: some more context - back when we made this in EnvoyFilter we had versions like master-latest-daily. We now call these `1.5-dev.SHA` for example (technically its not semver without 3 numbers but close enough to work) so it should work

mandarjog: This was a requests in the original envoyfilter API as well, but we did not have correct proxy version that time.

extension/v1alpha1/manifest.proto
howardjohn: I am not sure we should look at EnvoyFilter as a place to copy from. At the time we made that, metadata was pretty flat. With all the recent changes for telemetryv2 its actually fairly nested? Not sure if those are really needed to be selected on though

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+    FilterType type = 2;
+
+    // The name of the filter (corresponding to the envoy extension name).
+    string name = 3;
+
+    // Opaque configuration for the filter.
+    TypedStruct config = 4;
+
+    // Purpose of the extension.
+    Function function = 5;
```
bianpengyuan: How is order controlled for the filters with the same function?

kyessenov: Via the explicit order in the manifest. Generally, we would add a new function if the order becomes important (e.g. multiple pre-processing phases, or post-processing phases).

kyessenov:
Design doc updated. Please comment there https://docs.google.com/document/d/1jFuZM4UCKw6AMDmXMIYqbdlyJk95ut2zMPLSNdrUtW8/edit#

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+//     function: POST_PROCESSING
+//```
+//
+package istio.extension.v1alpha1;
+
+option go_package = "istio.io/api/extension/v1alpha1";
+
+import "google/protobuf/struct.proto";
+
+// Opaque configuration for an extension (wire-compatible with UDPA).
```
htuch: Why is this being redefined here? Is it possible to consume from UDPA directly?

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+  // used to select proxies using a specific version of istio
+  // proxy. The Istio version for a given proxy is obtained from the
+  // node metadata field ISTIO_VERSION supplied by the proxy when
+  // connecting to Pilot. This value is embedded as an environment
+  // variable (ISTIO_META_ISTIO_VERSION) in the Istio proxy docker
+  // image. Custom proxy implementations should provide this metadata
+  // variable to take advantage of the Istio version check option.
+  string proxy_version = 1;
+
+  // Match on the node metadata supplied by a proxy when connecting
```
htuch: How does this interact with the match criteria enabled by https://github.com/envoyproxy/envoy/pull/9301? Specifically the ability to match on the presence of a given filter?

extension/v1alpha1/manifest.proto
htuch: I think it would also be valuable to hear from @alexburnos here, as he has experience with similar systems.

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+
+    // Opaque configuration for the filter.
+    TypedStruct config = 4;
+
+    // Purpose of the extension.
+    Function function = 5;
+
+    // Modification operation (function is optional in case of replacement or
+    // removal).
+    Operation operation = 6;
```
htuch: I'm not an Istio API expert, but this strikes me as providing an imperative vs. a declarative API. I was under the impression we tend to describe what the system should be like, rather than how to get there.

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,189 @@
...
+    AUTHORIZATION = 3;
+    POST_PROCESSING = 4;
+  };
+
+  // Allows adding, replacing or removing of the extension by name. Note that
+  // for replacement, the function field is optional, and for removal, the
+  // config field is optional.
+  enum Operation {
+    ADD = 0;
+    REPLACE = 1;
```
howardjohn: I am a bit confused what the use case for remove and replace is? This feels like we are just going to add more and more options to this API and soon enough we have the full EnvoyFilter spec

extension/v1alpha1/manifest.proto
mandarjog: This part is debatable. We have a choice here.
We can forgo these operations, and then leave “how to get there” to CI/CD
In that case you install a whole new named manifest in your namespace, so someone else must do copy and update operation outside the istio API.
Please take a look at the linked doc where we have listed the use cases this needs to support.

extension/v1alpha1/manifest.proto
mandarjog: From my other comment, we can forgo all operations, so effectively the entire manifest must be replaced.
Or optimize for the most common use case over api purity.

extension/v1alpha1/manifest.proto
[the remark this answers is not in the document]
kyessenov: Closing. We can discuss the future of envoy filter in the doc.

extension/v1alpha1/manifest.proto
kyessenov: It's a tooling issue. Istio uses gogo so we can't really import UDPA go protos. Also, this surfaces as k8s JSON so proto types are completely erased in JSON. Next, we add new annotations to protos for istio tooling purposes, which would be difficult to apply in a different repo.

extension/v1alpha1/manifest.proto
kyessenov: @htuch We really only need the removal operation since there is no other way to provide an override for a filter that disables it. Think of a per-namespace declaration that disables a root default. .

extension/v1alpha1/manifest.proto
kyessenov: The intent is to provide an override (canarying or namespace-specific config).

extension/v1alpha1/manifest.proto
kyessenov: I'm not sure we need that match criterion in istio. We generally know what filters are packaged, and don't select on presence or absence of compiled filters. In the future, we expect filters to be dynamically loadable.

kyessenov:
Updated to provide an override semantics based on the _name_ of the filter as the primary key / URL of a filter config. This allows a more fine-grained selection instead of the whole manifest override, and is in fact, required to express config canarying (e.g. use config A if label is matched, but then fallback to config B).

extension/v1alpha1/manifest.proto
kyessenov: Match condition can evolve. I copied verbatim since I'm not sure exactly what this means :)

extension/v1alpha1/manifest.proto
kyessenov: Semver in UDPA is just a numeric version. Don't we need to handle ASM qualifier in addition to semver?

howardjohn: I guess that may be an issue, may want to select on `1.4.2-custom` vs `1.4.2`. Or we could say that is overloading the version and should just set another metadata field and use that?

htuch: The Envoy `BuildVersion` in https://github.com/envoyproxy/envoy/pull/9301 includes metadata capable of expressing these labels.

extension/v1alpha1/manifest.proto
htuch: I think it would be cleaner (and very simple) to add a per-filter disable override in Envoy. We've discussed this before and it just needs O(1 day) of effort to add if someone is inclined.

mandarjog: If this is a bool inside the filter proto, it would mean merging the default proto with the override. We want to avoid proto merges.

 a “conditional” dispatch of filters is a useful feature, and we have an open feature request in envoy for that.

However wouldn’t it be better to not send a filter configuration at all instead of sending it with a disabled flag?

rshriram:
I like the direction of this PR. That said, I am not fully convinced that it warrants a new API. There are several incremental usability improvements that are directly applicable to the EnvoyFilter API [this needs renaming]. Secondly, the notion that

>restricted to filters (as the core envoy extension mechanism), major reason why EnvoyFilter is not adequate (since it allows everything).

is short sighted. Most people who use some external authZ callout may also need a cluster to talk to the external service wont they? So, simply defining the extension without defining the cluster seems moot. 

Also, 
> no merging of protobufs (too expensive in CPU)
In the current API, we do merge ONLY if the user wants to merge. We do a simple list append if the user wants to add/insert. So, I dont see this as an improvement. It feels like simply removing a feature as it is expensive. 

Thirdly, extension is IMO subjective: people who want to tweak XFF settings are also extending the system because Istio does not provide a way to do so. Same goes for people who are tweaking some stream timeout (one of the N different timeouts) in a route because Istio exposes just one single knob for the common case use. If we go with a dedicated API for just filters (as was the case with envoyFilter v1), then you still have the gap for people who want to control these basic parameters. Incidentally, the entire reason for a merge operation is to support this feature (ability to tweak the listener/cluster/hcm) - most people use merge only for this, not for filters.

Finally, how will you handle unknown type URLs in Pilot? We need to pre-process all the extensions into their respective go protobuf structs so that we can insert into the list of filters. Is there some magic that allows us to read in the json config and ship it as is to envoy?

extension/v1alpha1/manifest.proto
kyessenov: ISTIO_VERSION is part of node metadata, so we would be regex matching on a field in metadata. We could express it as a map from a field to a string matcher, but that feels too level in istio, since it does not convey the purpose of using the version.

extension/v1alpha1/manifest.proto
kyessenov: @htuch Are you thinking of adding a bool field to filter config protos to disable it? I think that might be useful with filter chain discovery, to be able to dynamically disable filters. If so, we could mirror it here with a bool field instead of operation enum.

kyessenov:
> Most people who use some external authZ callout may also need a cluster to talk to the external service wont they? So, simply defining the extension without defining the cluster seems moot.

They should be using native APIs to define services and service entries to add an endpoint. I don't think we need a third way to add clusters to Envoy.

> In the current API, we do merge ONLY if the user wants to merge. We do a simple list append if the user wants to add/insert. So, I dont see this as an improvement. It feels like simply removing a feature as it is expensive.

Yes, it's a categorical rejection of merging functionality since it makes it imperative and difficult to write an admission controller for. It's hard to tell the effects of a merge without execution, so difficult for an admin to ensure that bad things are not allowed. We don't want to allow arbitrary scripting changes to xDS.

> Thirdly, extension is IMO subjective: people who want to tweak XFF settings are also extending the system because Istio does not provide a way to do so. Same goes for people who are tweaking some stream timeout (one of the N different timeouts) in a route because Istio exposes just one single knob for the common case use. If we go with a dedicated API for just filters (as was the case with envoyFilter v1), then you still have the gap for people who want to control these basic parameters. Incidentally, the entire reason for a merge operation is to support this feature (ability to tweak the listener/cluster/hcm) - most people use merge only for this, not for filters.

I think it's a known issue that HTTP connection manager is a monolith and is becoming unwieldy to configure (cc @htuch). The solution is to break it apart around features, and allow those features to be applied independently from one another. This API is not aimed to solving that problem, however. The use case to replace a HCM config or any other core config (tcp_proxy) seems out of scope.

> Finally, how will you handle unknown type URLs in Pilot? We need to pre-process all the extensions into their respective go protobuf structs so that we can insert into the list of filters. Is there some magic that allows us to read in the json config and ship it as is to envoy?

Yes, it's called UDPA typed struct.

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,213 @@
...
+      // Configuration reference.
+      ExtensionRef config_ref = 5;
+    }
+
+    // Modification operation (function is optional in case of replacement or
+    // removal).
+    Operation operation = 6;
+
+    // Purpose of the extension used to place the filter in a chain.
+    Function function = 7;
```
mandarjog: name: `filter_phase`

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,213 @@
...
+    FilterType type = 2;
+
+    // The required name of the filter instance. The name should generally be
+    // unique in a filter chain.
+    string name = 3;
+
+    // Opaque configuration for the filter.
+    oneof configuration {
+      // Inlined extension configuration.
+      Extension config = 4;
```
mandarjog: Supporting both ref and inline means that k8s rbac cannot be used. We need to use a more sophisitcated admission control.

extension/v1alpha1/manifest.proto
alexburnos: Essentially you need the version to check capabilities of the proxy itself (will it accept the extension or not). What is the reason not to rely on BuildVersion, which would make Istio more UDPA compliant, vs Istio-specific API such as using ISTIO_VERSION metadata?

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,213 @@
...
+// Beyond that, the order of application of the manifests is unspecified.
+// Additional control over the delegation of the extension injection should
+// be enforced with either resource-level RBAC or an admission controller.
+message ExtensionManifest {
+  // Extension filter type.
+  enum FilterType {
+    UNSPECIFIED_FILTER_TYPE = 0;
+    HTTP_FILTER = 1;
+    NETWORK_FILTER = 2;
+    UPSTREAM_NETWORK_FILTER = 3;
```
alexburnos: What's the difference between NETWORK_FILTER and UPSTREAM_NETWORK_FILTER?

extension/v1alpha1/manifest.proto
howardjohn: I don't think we have any reason other than that build version doesn't exist yet as far as I can tell? We would likely need to fallback to istio_version for old proxies even when it does though

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,213 @@
...
+package istio.extension.v1alpha1;
+
+option go_package = "istio.io/api/extension/v1alpha1";
+
+import "google/protobuf/struct.proto";
+
+// Opaque configuration for an extension (wire and JSON compatible with UDPA).
+// The extension resource allows white-listing pre-defined configuration for
+// extensions.
+message Extension {
```
alexburnos: Naming nit: maybe "ExtensionConfig"?

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,213 @@
...
+    // At least one of the conditions must be satisfied for the filter to be
+    // applied.
+    repeated MatchContext match = 1;
+
+    // Specifies the type of the filter.
+    FilterType type = 2;
+
+    // The required name of the filter instance. The name should generally be
+    // unique in a filter chain.
+    string name = 3;
```
alexburnos: I assume it is not a UDPA Filter name field (as it has uniqueness requirement)? In this case, where is UDPA filter name will be specified, in the Extension?

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,213 @@
...
+  enum Operation {
+    ADD = 0;
+    REMOVE = 2;
+  };
+
+  // Extension purpose serving as a hint to determine the order of the
+  // application of an extension in a chain of extensions. Istio recognizes the
+  // following categories of extensions and applies them in the order specified
+  // here.
+  enum Function {
```
alexburnos: What would be the mechanism for two independent providers of filters to coordinate relative order?

I.e. I have filters A and B, is there a way to specify that B should go before A?

kyessenov: We don't want to allow specific total order of filters. A partial order constraint like this enum is all we want to support. That is we can express that authorization filters should come before post-processing filters, but no specific order between authorization filters.

extension/v1alpha1/manifest.proto
kyessenov: I'm working to change UDPA filter name to be free-form in envoy. UDPA filter type URL is sufficient to determine an extension.

extension/v1alpha1/manifest.proto
kyessenov: NETWORK_FILTER is a downstream network filter, UPSTREAM_NETWORK_FILTER is an upstream network filter a.k.a cluster filter.

rshriram:
> > Most people who use some external authZ callout may also need a cluster to talk to the external service wont they? So, simply defining the extension without defining the cluster seems moot.
> 
> They should be using native APIs to define services and service entries to add an endpoint. I don't think we need a third way to add clusters to Envoy.
> 
> > In the current API, we do merge ONLY if the user wants to merge. We do a simple list append if the user wants to add/insert. So, I dont see this as an improvement. It feels like simply removing a feature as it is expensive.
> 
> Yes, it's a categorical rejection of merging functionality since it makes it imperative and difficult to write an admission controller for. It's hard to tell the effects of a merge without execution, so difficult for an admin to ensure that bad things are not allowed. We don't want to allow arbitrary scripting changes to xDS.
> 
> > Thirdly, extension is IMO subjective: people who want to tweak XFF settings are also extending the system because Istio does not provide a way to do so. Same goes for people who are tweaking some stream timeout (one of the N different timeouts) in a route because Istio exposes just one single knob for the common case use. If we go with a dedicated API for just filters (as was the case with envoyFilter v1), then you still have the gap for people who want to control these basic parameters. Incidentally, the entire reason for a merge operation is to support this feature (ability to tweak the listener/cluster/hcm) - most people use merge only for this, not for filters.
> 
> I think it's a known issue that HTTP connection manager is a monolith and is becoming unwieldy to configure (cc @htuch). The solution is to break it apart around features, and allow those features to be applied independently from one another. This API is not aimed to solving that problem, however. The use case to replace a HCM config or any other core config (tcp_proxy) seems out of scope.
> 

I am not fully convinced. This PR is effectively rolling back EnvoyFilter to its older incarnation (that allowed only add/remove of network and http filters). We improved it solely based on user feedback that such an API was insufficient for their needs (including the idea of adding clusters as many complained that using Service Entries to force cluster creation was unweildy and created unwanted listeners/routes, etc.). Editing HCM or tcp proxy should be well within the scope because they are also *filters*. Special casing these two filters will lead to a brittle solution as we would have to add more special cases in future (thrift, dubbo, etc.) with no end in sight. Nor is it useful to create one api for one set of filters and another API for a different set of configs in envoy. 

Fwiw, writing an admission controller to deny any EnvoyFilter with merge op is not that hard either. So, I dont see why it is necessary to remove it from the API when people who dont want it can write their own admission controllers to reject such objects.

> > Finally, how will you handle unknown type URLs in Pilot? We need to pre-process all the extensions into their respective go protobuf structs so that we can insert into the list of filters. Is there some magic that allows us to read in the json config and ship it as is to envoy?
> 
> Yes, it's called UDPA typed struct.

LOL! So we are back to shipping structs. So much for all the work around using any typed filters.

kyessenov:
> I am not fully convinced. This PR is effectively rolling back EnvoyFilter to its older incarnation (that allowed only add/remove of network and http filters). We improved it solely based on user feedback that such an API was insufficient for their needs (including the idea of adding clusters as many complained that using Service Entries to force cluster creation was unweildy and created unwanted listeners/routes, etc.). Editing HCM or tcp proxy should be well within the scope because they are also _filters_. Special casing these two filters will lead to a brittle solution as we would have to add more special cases in future (thrift, dubbo, etc.) with no end in sight. Nor is it useful to create one api for one set of filters and another API for a different set of configs in envoy.

This PR is not trying to solve every problem with extensions. We can add additional extensions since the top-level field is a list of filter extensions. Each API needs a purpose, not an un-scoped fits-all solution.

> Fwiw, writing an admission controller to deny any EnvoyFilter with merge op is not that hard either. So, I dont see why it is necessary to remove it from the API when people who dont want it can write their own admission controllers to reject such objects.

I beg to disagree. It's not easy to write an admissions controller for operations that include protobuf merge. You would need to have a very solid idea what pilot generates (which is also changing per version of Istio). Merge operations are not commutative, so you would also need to think through sequences of operations which is against the principle of a declarative API. Finally, for core filters, it's simply not future-proof since we don't specify the number of HTTP connection managers (one per listener, one per chain, one across all listeners?) or how they are partitioned.

> LOL! So we are back to shipping structs. So much for all the work around using any typed filters.

Struct is the only way to ensure that both Pilot and k8s can ingest config without a descriptor. We can add an optimization later for Pilot to load additional descriptors and encode struct as any before shipping to Envoy.

howardjohn:
> Struct is the only way to ensure that both Pilot and k8s can ingest config without a descriptor. We can add an optimization later for Pilot to load additional descriptors and encode struct as any before shipping to Envoy.

This is just for the extension anyways, right? I don't think its horrible to have extensions as struct, but everything being a struct would be horrible

rshriram:
> This PR is not trying to solve every problem with extensions. We can add additional extensions since the top-level field is a list of filter extensions. Each API needs a purpose, not an un-scoped fits-all solution.

When you start adding Clusters, Listeners and route edits to this extension manifest, you get the existing EnvoyFilter API. And there is no escaping from adding clusters and listener edits if you want to solve the basic use case of wanting to enable certain fields that Istio does not care about, or needing to add clusters used by the extensions with custom tls settings or dns settings. Put another way, if you move the spec.Filters under the existing EnvoyFilter and tweak some existing semantics like no editing filters in listeners, you get the same output. You can still write a very simple admission controller that says only admin namespaces can have listener edits while consumer namespaces can have extensions, along with the approved/banned extensions stuff talked about earlier. So the question again, is why a new CRD? I dont mind renaming the EnvoyFilter to Extensions, but I certainly do not want to different ways of editing the envoy config and all the associated confusion that it creates.

extension/v1alpha1/manifest.proto
```diff
@@ -0,0 +1,213 @@
...
+  // The service port of the upstream or downstream configuration resource. If
+  // omitted, applies to clusters and listeners for any port.
+  uint32 port_number = 4;
+
+  // The Istio gateway config's namespace/name for which this route
+  // configuration was generated. Applies only if the context is GATEWAY. Should
+  // be in the namespace/name format. Use this field in conjunction with the
+  // portNumber to accurately select the Envoy route configuration for a
+  // specific HTTPS server within a gateway config object.
+  string gateway = 5;
```
rshriram: a gateway can have many Servers with its own hosts. Each of which ends up in its own filter chain if using TLS. So you need a way to specify the hostnames here as well, so that you can match a filter chain based on the sni. 

The match as it stands is insufficient as people have most commonly wanted to add a lua filter only for certain paths and not for all routes (like /user/.. and not for /index.html). I have seen this way too many times to keep track. This then requires route.PerFilterconfig in envoy. You cant do that with the way you have expressed the matches as it stand today. You would also now need to match on a specific route by name so that you can add this config to the perFilterConfig. 

Finally, since we are in the land of designing for future, we also need to take into account the passthrough listeners. People will want to write loggers/trackers/enforces for the traffic that is passthrough while using standard rbac for the rest.

rshriram:
also, fwiw, here is the deprecated EnvoyFilter API https://github.com/istio/api/blob/1.2.10/networking/v1alpha3/envoy_filter.proto that is essentially doing the same thing as the extension stuff

kyessenov:
> When you start adding Clusters, Listeners and route edits to this extension manifest, you get the existing EnvoyFilter API. And there is no escaping from adding clusters and listener edits if you want to solve the basic use case of wanting to enable certain fields that Istio does not care about, or needing to add clusters used by the extensions with custom tls settings or dns settings. Put another way, if you move the spec.Filters under the existing EnvoyFilter and tweak some existing semantics like no editing filters in listeners, you get the same output. You can still write a very simple admission controller that says only admin namespaces can have listener edits while consumer namespaces can have extensions, along with the approved/banned extensions stuff talked about earlier. So the question again, is why a new CRD? I dont mind renaming the EnvoyFilter to Extensions, but I certainly do not want to different ways of editing the envoy config and all the associated confusion that it creates.

I don't understand why listeners edits are essential. Again, the purpose of this effort to define a _declarative_ intent-based API for extensions. EnvoyFilter is a scripting language in disguise which I hope you agree is not amenable to static analysis (or you have some magic way to prove programs are safe). That's the primary reason for this new CRD, to make something more declarative that is also hopefully not entirely tied to a specific istio and envoy versions.

I also don't understand about the different way to edit envoy config argument. Every single networking API is editing envoy config in some ways. The reasons why I'm restricting to extensions is because I'm not convinced we need to model cluster patches the same way. They more naturally belong to other istio resources, with some opaque extensions. Same for route per-filter config, it naturally belongs to virtual service extension similar to ingress v2.

extension/v1alpha1/manifest.proto
kyessenov: Shouldn't per filter config be part of virtual service or other service/route-specific resource? This API is written from an administrator point of view (MUST enable telemetry and authorization) so opting out should be an override.

kyessenov:
The previous iteration is closed but it's not handling two issues:
1. We didn't have UDPA typed configs / structs.
2. We didn't assign names to filter instances that allow override semantics between cluster-level and namespace-level.

extension/v1alpha1/manifest.proto
mandarjog: There is no impl under way yet. We are still in the process of getting agreement. 

Re: ordering, integers priority helps, but what is easier is to preserve the manifest order in the actual output within a filter type.

dgn: thanks, yeah I remember now that this was talked about in the last meeting. I'm not sure that's very user-friendly- it's non-obvious that this implicit order matters, as the filters could be matching on completely separate sidecars. It also leaves out how this would behave if different `ExtensionManifests` add filters to the same Envoy config. Which one wins?

dgn:
- As a user, I will want to create an `ExtensionManifest` (or whatever the name) resource to define all filters on a given workload. I would therefore expect to have the `match` section at the root of the resource, and then a list of filters that are applied to that resource.
- I would also expect to be able to specify filters of a different `FilterType` for the same workload, in the same resource - this is not possible using the current design. Having to split filters that match the same workload across multiple `ExtensionManifest` resources just because they're of a different `FilterType` seems arbitrary
- as a user, I would expect that there's no difference between defining 1 filter each in 2 separate `ExtensionManifest` resources and defining 2 filters in 1 `ExtensionManifest` resource
- deriving the relative ordering of filters from the order of filters in the list is non-obvious (because they could be matching on completely separate workloads) and leads to non-deterministic behaviour when multiple `ExtensionManifest` instances add filters of the same `FilterType` matching the same workload

dgn:
my suggestions, derived from the problems outlined in my previous comment would therefore be:
- move the `match` section into the `spec`
- move the `FilterType` field into the filter list items (every filter in the list can have different type)
- add a `priority` field to the filter list items to establish relative ordering within `FilterType`

kyessenov:
This is stale.

PR #1222, opened 2019-12-30
add deny and negative match in authorization policy

yangminzhu:
The design doc is here: https://docs.google.com/document/d/1aJ1hffHz9JYGXIM9btnKaWmwVdn1Qg34FJScnDyZPw4/edit?folder=0BzW5bSyKst8Jb2hhRWQ2eTJYVzQ#

For https://github.com/istio/istio/issues/19894


yangminzhu:
@liminw Could you do a first round of the review for the PR? thank you.

cc @linsun @louiscryan @smawson @diemtvu @incfly

security/v1beta1/authorization.proto
```diff
@@ -224,56 +274,100 @@ message Rule {
...
+//
+// For example, the following operation matches if the host has suffix ".example.com"
+// and the method is "GET" or "HEAD" and the path doesn't have prefix "/admin".
+//
+// ```yaml
+// hosts: ["*.example.com"]
+// methods: ["GET", "HEAD"]
+// not_paths: ["/admin*"]
+// ```
 message Operation {
```
howardjohn: possible idea would be instead of making each of these have not_ versions, just make two top level fields, one representing not and one not not (don't call the field `not_not` though :slightly_smiling_face:

security/v1beta1/authorization.proto
```diff
@@ -67,16 +76,41 @@ import "type/v1beta1/selector.proto";
...
+//  - from:
+//    - source:
+//        namespaces: ["dev"]
+//    to:
+//    - operation:
+//        methods: ["POST"]
+// ```
+//
+// Safe-list and Block-list access control could be used on the same workload at the same
+// time. When both DENY policies and ALLOW policies are exist for a workload, the DENY policies
```
liminw: are exist => exist

security/v1beta1/authorization.proto
```diff
@@ -181,16 +215,32 @@ message AuthorizationPolicy {
   // same namespace as the authorization policy.
   istio.type.v1beta1.WorkloadSelector selector = 1;
 
-  // Optional. A list of rules to specify the allowed access to the workload.
+  // Optional. A list of rules to match the request. A match occurs when at least
+  // one rule matches the request.
   //
-  // If not set, access is denied unless explicitly allowed by other authorization policy.
+  // If not set, matches nothing. In other words, it is always not matched.
```
liminw: This description is not clear. You can just mention "If not set, requests to the target workload are denied by default."

security/v1beta1/authorization.proto
```diff
@@ -25,18 +25,30 @@ import "type/v1beta1/selector.proto";
 
 // Istio Authorization Policy enables access control on workloads in the mesh.
 //
-// For example, the following authorization policy applies to workloads matched with
-// label selector "app: httpbin, version: v1".
+// The authorization policy supports both safe-list style and block-list style access
+// control.
+//
```
liminw: I feel that mentioning "safe-list"/"block-list" and "allow"/"deny" interchangeably are not helping to clarify things. Can we just mention "allow policy" and "deny policy"? They are already self-explained. No need to introduce more terminologies here.

security/v1beta1/authorization.proto
```diff
@@ -67,16 +76,41 @@ import "type/v1beta1/selector.proto";
...
+// Safe-list and Block-list access control could be used on the same workload at the same
+// time. When both DENY policies and ALLOW policies are exist for a workload, the DENY policies
+// are evaluated first. The evaluation is determined by the following rules:
 //
-// Currently AuthorizationPolicy only supports "ALLOW" action. This means that
-// if multiple authorization policies apply to the same workload, the effect is additive.
+// 1. Are there any DENY policies selecting the workload? If Yes goto step 2, otherwise goto step 3.
+// 2. Are any of these DENY policies matched the request? If Yes then deny (block-list style), otherwise goto step 3.
+// 3. Are there any ALLOW policies selecting the workload? If Yes goto step 4, otherwise allow.
+// 4. Are any of these ALLOW policies matched the request? If Yes then allow, otherwise deny (safe-list style).
```
liminw: No need to mention "safe-list style" or "block-list style". Deny/allow is already very clear. 

"If Yes then allow" => "If Yes, allow".

security/v1beta1/authorization.proto
liminw: The proto would be shorter. But that would introduce another level indention in the policy definition (e.g. matches/not_matches), which makes the policy more verbose. 

For example, for the above example, you will have write:
```
operation:
    matches:
        hosts: ["*.example.com"]
        methods: ["GET", "HEAD"]
    notMatches:
        paths: ["/admin*"]
```

security/v1beta1/authorization.proto
yangminzhu: done, I think we need add validation to reject policies that has action `DENY` but doesn't set `rules` to avoid special handling in this case.

security/v1beta1/authorization.proto
yangminzhu: Updated.

security/v1beta1/authorization.proto
yangminzhu: The other thing is it requires user to change their existing policies if they want to use the new top-level fields.

security/v1beta1/authorization.proto
```diff
@@ -25,18 +25,29 @@ import "type/v1beta1/selector.proto";
 
 // Istio Authorization Policy enables access control on workloads in the mesh.
 //
-// For example, the following authorization policy applies to workloads matched with
-// label selector "app: httpbin, version: v1".
+// The authorization policy supports both allow policy and deny policy.
+//
+// - The allow policy means a request is allowed only if it matches any of the
+// allow policies, otherwise it is denied.
+// - The deny policy means a request is allowed if it doesn't match any of
```
liminw: This is a little misleading. If there is no explicit deny, it does not mean "allow". The request still has to be explicitly allowed. Since you already explained in the later section that "deny policy is evaluated before allow policy", you can eliminate this part. If you prefer, you can move that section up here.

yangminzhu: Updated.

yangminzhu:
@louiscryan @smawson @linsun @rshriram @dcberg @geeknoid Could you review the API change, this has been approved by @liminw. Thank you.

yangminzhu:
ping.

yangminzhu:
@louiscryan @smawson @linsun @rshriram @dcberg @geeknoid Can someone take a look, I'm trying to get this to 1.5 (cut in ~2 weeks) and all the following changes are blocked by this API change, thank you.

security/v1beta1/authorization.proto
```diff
@@ -25,18 +25,31 @@ import "type/v1beta1/selector.proto";
 
 // Istio Authorization Policy enables access control on workloads in the mesh.
 //
-// For example, the following authorization policy applies to workloads matched with
-// label selector "app: httpbin, version: v1".
+// The authorization policy supports both allow policy and deny policy. When
```
smawson:
```suggestion
// Authorization policy supports both allow and deny policies. When
```

security/v1beta1/authorization.proto
```diff
@@ -25,18 +25,31 @@ import "type/v1beta1/selector.proto";
...
+// allow policies and deny policies are used for a workload at the same time,
+// the deny policies are evaluated first. The evaluation is determined by the following rules:
+//
+// 1. Are there any DENY policies for the workload? Yes -> step 2, No -> step 3.
```
smawson: I think you can simplify this:

1. If there are any DENY policies that match the request, deny the request.
2. If there are no ALLOW policies for the workload, allow the request.
3. If any of the ALLOW policies match the request, allow the request.
4. Deny the request.

security/v1beta1/authorization.proto
```diff
@@ -25,18 +25,31 @@ import "type/v1beta1/selector.proto";
 
 // Istio Authorization Policy enables access control on workloads in the mesh.
 //
-// For example, the following authorization policy applies to workloads matched with
-// label selector "app: httpbin, version: v1".
+// The authorization policy supports both allow policy and deny policy. When
+// allow policies and deny policies are used for a workload at the same time,
```
smawson:
```suggestion
// allow and deny policies are used for a workload at the same time,
```

security/v1beta1/authorization.proto
```diff
@@ -25,18 +25,31 @@ import "type/v1beta1/selector.proto";
...
+// the deny policies are evaluated first. The evaluation is determined by the following rules:
+//
+// 1. Are there any DENY policies for the workload? Yes -> step 2, No -> step 3.
+// 2. Is the request matched with any of the DENY policies? Yes -> deny, No -> step 3.
+// 3. Are there any ALLOW policies for the workload? Yes -> step 4, No -> allow.
+// 4. Is the request matched with any of the ALLOW policies? Yes -> allow, No -> deny.
+//
+// For example, the following authorization policy sets the `action` to "ALLOW"
```
smawson: We should mention that the default action is ALLOW but that it is useful to be explicit.

security/v1beta1/authorization.proto
```diff
@@ -67,16 +77,29 @@ import "type/v1beta1/selector.proto";
 //      values: ["https://accounts.google.com"]
 // ```
 //
-// Access control is enabled on a workload if there is any authorization policies selecting
-// the workload. When access control is enabled, the default behavior is deny (deny-by-default)
-// which means requests to the workload will be rejected if the request is not allowed by any of
-// the authorization policies selecting the workload.
+// The following is another example that sets `action` to "DENY" to use the deny policy.
```
smawson:
```suggestion
// The following is another example that sets `action` to "DENY" to create a deny policy.
```

security/v1beta1/authorization.proto
```diff
@@ -25,18 +25,31 @@ import "type/v1beta1/selector.proto";
...
+// allow policies and deny policies are used for a workload at the same time,
+// the deny policies are evaluated first. The evaluation is determined by the following rules:
+//
+// 1. Are there any DENY policies for the workload? Yes -> step 2, No -> step 3.
+// 2. Is the request matched with any of the DENY policies? Yes -> deny, No -> step 3.
+// 3. Are there any ALLOW policies for the workload? Yes -> step 4, No -> allow.
+// 4. Is the request matched with any of the ALLOW policies? Yes -> allow, No -> deny.
+//
+// For example, the following authorization policy sets the `action` to "ALLOW"
+// to use the allow policy.
```
smawson:
```suggestion
// to create an allow policy.
```

security/v1beta1/authorization.proto
```diff
@@ -67,16 +77,29 @@ import "type/v1beta1/selector.proto";
 //      values: ["https://accounts.google.com"]
 // ```
 //
-// Access control is enabled on a workload if there is any authorization policies selecting
-// the workload. When access control is enabled, the default behavior is deny (deny-by-default)
-// which means requests to the workload will be rejected if the request is not allowed by any of
-// the authorization policies selecting the workload.
+// The following is another example that sets `action` to "DENY" to use the deny policy.
+// It denies requests from namespace "dev" to access the workload with "POST" method.
```
smawson:
```suggestion
// It denies requests from the "dev" namespace to the "POST" method on all workloads in the "foo" namespace.
```

security/v1beta1/authorization.proto
```diff
@@ -181,16 +204,31 @@ message AuthorizationPolicy {
   // same namespace as the authorization policy.
   istio.type.v1beta1.WorkloadSelector selector = 1;
 
-  // Optional. A list of rules to specify the allowed access to the workload.
+  // Optional. A list of rules to match the request. A match occurs when at least
+  // one rule matches the request.
   //
-  // If not set, access is denied unless explicitly allowed by other authorization policy.
+  // If not set, requests to the target workload are denied by default.
```
smawson: I feel like this is still a little confusing because of the complexity.

Basically what we want to say is that if you want to enable deny-by-default, create an authorization policy without any rules, since the default if there is no authorization policy is to allow, but the default if there is an authorization policy is to deny. So maybe something like:

If not set, this policy changes the default to deny requests if they do not match any rules.

Also we need to think about what the meaning of the action field is with a policy with no rules. Should it be used to set the default? So could they say action = ALLOW with no rules and that changes the default to allow?

security/v1beta1/authorization.proto
```diff
@@ -181,16 +204,31 @@ message AuthorizationPolicy {
...
+    DENY = 1;
+  }
+
+  // Optional. The action to take if the request is matched with the rules.
+  Action action = 3;
 }
 
-// Rule allows access from a list of sources to perform a list of operations when
-// the condition is matched.
+// Rule matches requests from a list of sources that perform a list of operations when
```
smawson:
```suggestion
// Rule matches requests from a list of sources that perform a list of operations subject to a
```

security/v1beta1/authorization.proto
```diff
@@ -181,16 +204,31 @@ message AuthorizationPolicy {
...
+// the condition is matched. A match occurs when at least one source, operation and condition
```
smawson:
```suggestion
// list of conditions. A match occurs when at least one source, operation and condition
```

security/v1beta1/authorization.proto
```diff
@@ -224,56 +262,100 @@ message Rule {
...
   // Optional. A list of paths, which matches to the "request.url_path" attribute.
   //
   // If not set, any path is allowed. Must be used only with HTTP.
   repeated string paths = 4;
+
+  // Optional. A list of negative match of paths.
+  repeated string not_paths = 8;
 }
 
 // Condition specifies additional required attributes.
```
smawson: Should we specify somewhere that either values or not_values is required?

yangminzhu: Yes, I added a comment: `Note: at least one of the values or not_values field must be set.`

security/v1beta1/authorization.proto
yangminzhu: > If not set, this policy changes the default to deny requests if they do not match any rules.

I feel this is a little bit hard to understand, fundamentally, the `rules` are used to match the request, if the `rules` is not set, it means it will never match on any of the request. And this turns out to be deny all.

> Also we need to think about what the meaning of the action field is with a policy with no rules. Should it be used to set the default? So could they say action = ALLOW with no rules and that changes the default to allow?

Like I mentioned before, probably the best way is to think about the `action` separately from the `rules`: the `rules` is used to match the request and the `action` happens when the match occurs.

- `rules` not set and `action` set to "ALLOW": the request is never matched and the allow action never happens, in other words, it's deny-all
- `rules` not set and `action` set to "DENY": the request is never matched and the deny action never happens, in other words, it's meaningless as it essentially does nothing, we could let the validation to reject such policy.

I updated the comment please let me know wdyt, thanks.

security/v1beta1/authorization.proto
```diff
@@ -181,16 +206,34 @@ message AuthorizationPolicy {
   // same namespace as the authorization policy.
   istio.type.v1beta1.WorkloadSelector selector = 1;
 
-  // Optional. A list of rules to specify the allowed access to the workload.
+  // Optional. A list of rules to match the request. A match occurs when at least
+  // one rule matches the request.
   //
-  // If not set, access is denied unless explicitly allowed by other authorization policy.
+  // If not set, the match will never occur. You can leave this field unset and
+  // set the action to "ALLOW" to create a deny-all policy. It is meaningless to
```
smawson: Well according to what you said, setting Action with no rules is basically useless.  So I'd rephrase to say that the action field is ignored if there are no rules, since it happens only when a rule is matched.

Also there is a little technicality here between no rules and a blanket deny-all rule, which is that if you have no rules, but someone creates another authorization policy which has rules, your authorization policy is basically ignored.

So it is really just about setting the default, not about creating a deny-all policy. It is kind of a weird way to set the default though, especially since setting action doesn't do anything. I'm fine with leaving this somewhat nebulous here, or even taking it out rather than being not-quite-correct.

Maybe a reword would be:

If not set, the match will never occur. This is equivalent to setting a default of deny for the target workloads.

security/v1beta1/authorization.proto
```diff
@@ -282,6 +369,11 @@ message Condition {
   // See the [full list of supported attributes](https://istio.io/docs/reference/config/security/conditions/).
   string key = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // The allowed values for the attribute.
-  repeated string values = 2 [(google.api.field_behavior) = REQUIRED];
+  // Optional. A list of allowed values for the attribute.
+  // Note: at least one of the values or not_values field must be set.
```
smawson:
```suggestion
  // Note: at least one of values or not_values must be set.
```

security/v1beta1/authorization.proto
```diff
@@ -282,6 +369,11 @@ message Condition {
...
+  repeated string values = 2;
+
+  // Optional. A list of negative match of values for the attribute.
+  // Note: at least one of the values or not_values field must be set.
```
smawson:
```suggestion
  // Note: at least one of values or not_values must be set.
```

security/v1beta1/authorization.proto
yangminzhu: I see, updated, thanks.

PR #1225, opened 2020-01-07
Add a flag to configure whether using k8s CA to sign istiod certificate

lei-tang:
This PR is for the issue: https://github.com/istio/istio/issues/19950


mesh/v1alpha1/config.proto
```diff
@@ -387,8 +387,11 @@ message MeshConfig {
   // Configure the provision of certificates.
   repeated Certificate certificates = 47;
 
+  // Configure whether istio uses Kubernetes CA APIs to sign Pilot certificate.
+  bool sign_cert_at_kubernetes_ca = 49;
```
howardjohn: @costinm mentioned something about an RFC for a generic cert signing API. I am not sure of the details, but we may want to future proof this to allow things other than a true/false

costinm:
Let's start with autodection and maybe an env variable - before making it an API.

I think at some point we should have an API to select the CA used for DNS certs and for spiffee certs - and that will probably need to have the full address of the provider, could be kubernetes:// for using the local k8s api server.  

But in 1.5 - given that API signing in k8s is not broadly available and is beta and might change - I think we could as well just detect it and have a env-based override.

lei-tang:
Per the comments, this PR is closed and env variables will be used to configure whether using k8s CA or not.

PR #1227, opened 2020-01-09
Add annotation for suppressing analyzer messages per-resource

selmanj:
Note that this is slightly different than the annotation originally decided in https://docs.google.com/document/d/1r1hOBIwkDesZ8vYaHMj2YThEcUxegB-fLMPBst_yXIU/edit#heading=h.xw1gqgyqs5b. Originally the annotation was called `galley.istio.io/noanalyze`, but I was worried that the name itself might be potentially confusing (for example, the presence of this annotation appears to imply that the resource is never analyzed, rather than a specific list of codes).

Due to this change i've added approvers from the doc as approvers on this PR as well.

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -79,6 +79,21 @@ var (
 		  Resources: []ResourceTypes{ Service, },
         }
 	
+		GalleyAnalyzeSuppress = Instance {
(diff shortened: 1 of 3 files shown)


selmanj:
Can a code owner take a look?

PR #1229, opened 2020-01-09
use custom type for k8s intorstring for operator

richardwxn:
Use custom type for k8s intorstring for operator with custom marshaller/unmarshaller, otherwise the unmarshaling would not work for targetport.

This change together with the operator PR would enable setting k8s service specs for IstioOperatorSpec


howardjohn:
why doesn't it work? if I'm remembering correctly there was some bug with this that for fixed, but may be different can you give more details

richardwxn:
@howardjohn the k8s intorstring type has custom unmarshaller defined but not called when unmarshalling the k8s service type for jsobpb, so we need to mirror that type and its unmarshaller. I was trying to address that in previous operator PR: https://github.com/istio/operator/pull/711. But since now we are using the new API in istio/api so I need to make the change here

richardwxn:
@howardjohn @ostromart @linsun Can you PTAL? This is blocking another operator fix:  https://github.com/istio/operator/pull/711

PR #1230, opened 2020-01-09
add service annotation to KubernetesResourcesSpec in operator api.

morvencao:
resubmit PR to resolve istio/istio#19959 after https://github.com/istio/operator/pull/713 and https://github.com/istio/api/pull/1223/


morvencao:
@ostromart PTAL

ostromart:
Thanks @morvencao. We also need approval from TOC member (@smawson and @linsun are familiar with this API).

PR #1231, opened 2020-01-10
move status out of spec for operator api

morvencao:
resolve: https://github.com/istio/istio/issues/20021


ostromart:
LGTM, however I'm not completely sure this is the root cause of the problem. Anyway, I agree with this change so let's see what happens.

morvencao:
@ostromart Currently `e2e-operator` test is broken because of  failing status update, and the reconcile loop can't be stopped. 
I think https://github.com/istio/operator/pull/741 can fix that temporarily.
However, after this PR is merged, we need to enhance the status update logic further.

PR #1232, opened 2020-01-10
Add v1beta1 for Networking APIs

jasonwzm:


hzxuzhonghu:
One question: how should we support two versions simultaneously? CRD conversion?

jasonwzm:
> One question: how should we support two versions simultaneously? CRD conversion?

For v1beta1 and 1.5, they are the same schema with v1alpha3 APIs. In 1.6, we will use API conversion.

hzxuzhonghu:
>For v1beta1 and 1.5, they are the same schema with v1alpha3 APIs.
But we can not get v1alpha resource with v1beta1 client.

>In 1.6, we will use API conversion.

It seems this is the only feasible solution

jasonwzm:
> > For v1beta1 and 1.5, they are the same schema with v1alpha3 APIs.
> 
> But we can not get v1alpha resource with v1beta1 client.
> 
Since they are the same schema, you are able to get the resource with either version.

hzxuzhonghu:
ah, good. Long time no tracing it. https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions

Issue #1220, opened 2019-12-28
Extending StringMatch to support "if-not-present" in the header match


igsong:
**Describe the feature request**

I would like to suggest to extend header match capability. For example, Route of Envoy is capable to express non-existence of a header, but Istio is not. In some case, an existence of a header is not acceptable to a specific endpoint.

One easy way is to add fields into StringMatch of VirtualService as follows:
```
// StringMatch of VirtualService
message StringMatch {
  oneof match_type {

    // exact string match
    string exact = 1;

    // prefix-based match
    string prefix = 2;

    // ECMAscript style regex-based match
    string regex = 3;

    // Newly added to express "present of a header"  
    bool present = 4;
  }

  reserved 5, 6, 7, 8;
  // Newly added to express negation
  bool invert = 9;
}
```

In addition, similar with above way, `suffix` matching may be able to be implemented.

**Describe alternatives you've considered**

What should be noted is that `StringMatch` is used for "header match" as well as "query parameter match", while the header match and query parameter match are differently handled by Envoy. 
(Basically, Envoy does not support `invert` semantics for the "query parameter match")

To keep consistency between Envoy and Istio, it would be better to define and use new data structure for the header match, in order to use different data structure for the header match and query parameter match.
However, it may cause incompatibility with previous API version, so I think that the addition of fields as above is easier way.

[X] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience

**Additional context**

rshriram:
this is fine but dont do it in string match as other fields like url match use it. One option is to create a richer HeaderMatch that has present and inverted. Another one is to do a not_match field with the old string match (and a not present field with just header names)

igsong:
Thanks for comments! I'll make a pull request for this in several weeks.

rshriram:
are you planning to send an istio API Pr for this as well ?

igsong:
Do you mean that I have a plan for submitting PR to istio/istio regarding the changes happen on https://github.com/istio/api/pull/1233 ? If right, I will do soon after the changed version of API is properly included in istio/istio.

igsong:
BTW, it would be better to keep this issue until corresponding changes is made on istio/istio.

PR #1233, opened 2020-01-12
Add the HeaderMatch field and type for supporting all the header match types of Envoy

igsong:
Istio header matching scheme is not supporting all the header match type of Envoy.
So, I would like to suggest to add new field and type `HeaderMatch`.
For more details of motivation, please refer to https://github.com/istio/api/issues/1220#issue-543293746


networking/v1alpha3/virtual_service.proto
```diff
@@ -969,6 +971,19 @@ message StringMatch {
...
+    string exact = 2;
+    string regex = 3;
+    string prefix = 4;
+    string suffix = 5;
+    Int64Range range = 6;
+    bool present = 7;
+  }
+  bool invert = 10;
+}
+
```
rshriram: Why do we need a new match message when we already have a Header match field? Can't we do the following as I suggested in the comment:
1. Fix the validation (and the doc comments in this proto) to say that if the StringMatch is nil, then we just test for the presence of the header. This would then allow the user to specify
```
match:
    headers:
        foo: 
```
2. For the if-not-present case, add a separate `without_headers` field that is also a map<string, StringMatch>. 
This should allow you to do 
```
match:
 withoutHeaders:
     foo:
```

3. The whole int64range is a super weird use case that microsoft added for their internal use. I have not seen a single person ask for that in Istio. Do you really need the integer range matching? If not, skip it. 
4. Do you need suffix match ? If not, skip it.

Essentially, the entirety of your change will be to add a new without_headers field and change the docs to indicate that header matches without a value will be treated as a presence check.

igsong: First of all, thank you for comments, and I'll make a change regarding your suggestions as follows:
 
* I don't need 3 and 4 right now, so I'll skip them, as your suggestion.
* For presence of the header, it seems OK, but it could be also presented with regex as follows:
```
match:
  headers:
    foo:
      regex: .*
```
* For if-not-present case, `withoutHeaders` is effectively working for my case, so it will be done.

networking/v1alpha3/virtual_service.proto
```diff
@@ -719,6 +719,10 @@ message HTTPMatchRequest {
   // **Note:** The case will be ignored only in the case of `exact` and `prefix`
   // URI matches.
   bool ignore_uri_case = 10;
+
+  // withoutHeader has the same syntax with the header, but has opposite meaning.
+  // If a header is matched with a matching rule among withoutHeader, the traffic becomes not matched one.
```
rshriram: dont forget to add a note in headers saying if value is absent, it is a presence check

networking/v1alpha3/virtual_service.gen.json
```diff
@@ -243,6 +243,13 @@
           "ignoreUriCase": {
             "description": "Flag to specify whether the URI matching should be case-insensitive.",
             "type": "boolean"
+          },
+          "withoutHeaders": {
```
rshriram: this is not in sync with the additional doc comment in the proto. Thats why gen check s failing

rshriram:
you need to sign the CLA.

rshriram:
@howardjohn cc

PR #1236, opened 2020-01-17
fix toleration type for operator

richardwxn:
1. change it to be repeated 
2. use k8s core toleration type


richardwxn:
@smawson @linsun Can you PTAL?

PR #1237, opened 2020-01-21
Change to camelCase to make the marshal/unmarshal work

irisdingbj:
https://github.com/istio/istio/issues/20328

according to our api guideline, we should not camelCase for field name. https://github.com/istio/api/blob/master/GUIDELINES.md#fields

But this change will make sure we use `ingressGateways` as a uniform name both in iop.yaml and also all the builtin profiles we have in operator.

This is actually have another way to fix the problem that we can keep the filed name unchanged in API. But it also means we need to change to `ingress_gateways` in iop and also all profiles we have in operator.

From a User's perspective, I think `ingressGateways` is better than `ingress_gateways` so this is the reason for this PR.


ostromart:
@irisdingbj could you provide a bit more explanation why this is needed? since we are using jsonpb for [un]marshaling, it seems to me it should not be required. what is the specific problem you're trying to fix here?
btw. we use camelcase solely in the case of kubernetes.proto because those messages are mirrored from k8s.

irisdingbj:
@ostromart Thanks for looking into this.
If I have below in my iop:
 components:
    ingressGateways:
      - namespace: istio-system
When in-cluster operator reads the IOP, it will get an empty ingressGateways. This is because k8s is using `ingress_gateways` as we defined in our component.proto.

As you mentioned, all kubernetes.proto is using cacmelCase, So this is actually the same situation as we are running in cluster-operator.

I can confirm this is not a problem is you use `istioctl manifest apply`.

morvencao:
I just realized the issue should not just happen in `gateways`, there are still others fields with snake_case json tag, I think we can fix these in https://github.com/istio/api/blob/master/operator/fixup_structs/main.go 
I raised a PR for this: https://github.com/istio/api/pull/1277

irisdingbj:
> I just realized the issue should not just happen in `gateways`, there are still others fields with snake_case json tag, I think we can fix these in https://github.com/istio/api/blob/master/operator/fixup_structs/main.go
> I raised a PR for this: #1277

yes, this happens for all snake case json tag

ostromart:
Closing this since https://github.com/istio/api/pull/1277 is making a similar fix.

PR #1241, opened 2020-01-22
Peer authentication beta API, aka mTLS beta

diemtvu:


rshriram:
@diemtvu take a look at my PR for sidecar TLS termination (one way tls) #1248 . I think this API can be nicely folded into the Sidecar api so that the entire TLS thing is coherent. We can relax certain semantics of the sidecars (like mandatory egress, etc.) to accommodate the tls policy

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,100 @@
...
+
+    // Connection is in mTLS tunnel (TLS with client cert must be presented).
+    STRICT = 1;
+
+    // Connection is not tunneled.
+    DISABLE = 2;
+  }
+
+  // Defines the mode of peer authentication.
+  Mode mode = 2;
```
rshriram: you need to add per port override here as well. Currently we have a bug. if the authn policy is strict, it applies to all declared ingress ports on a sidecar even if the user wanted custom TLS on that ingress port, or no TLS for that matter.

rshriram: You also need to surface the mechanism to identify the mesh mTLS using ALPNs or in case of @mdhume's use case, based on SNIs (wherein non sidecar clients are using istio mTLS certs to talk to servers with sidecar, but they cant send the ALPNs).

diemtvu: I'd like to postpone sni/alpn later, after we delivery the essential features. I've add MutualTls wrapper message, so that we can add configuration for them later.

For port override, I add a map from port (number) to MutualTls. I am a bit hesitant on this too, and would like to hide it first until we are sure it works well in all aspect. Wdyt?

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,109 @@
...
+  // If not set, the policy will be applied to all workloads in the same namespace as the policy.
+  istio.type.v1beta1.WorkloadSelector selector = 1;
+
+  // Mutual TLS settings.
+  message MutualTLS {
+    enum Mode {
+      // Connection can be either plaintext or mTLS tunnel.
+      PERMISSIVE = 0;
+
+      // Connection is in mTLS tunnel (TLS with client cert must be presented).
```
smawson:
```suggestion
      // Connection is an mTLS tunnel (TLS with client cert must be presented).
```

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,109 @@
...
+  istio.type.v1beta1.WorkloadSelector selector = 1;
+
+  // Mutual TLS settings.
+  message MutualTLS {
+    enum Mode {
+      // Connection can be either plaintext or mTLS tunnel.
+      PERMISSIVE = 0;
+
+      // Connection is an mTLS tunnel (TLS with client cert must be presented).
+      STRICT = 1;
```
costinm: Strict - and all access checks that are required on a connection - should be in the authorization API. 

I think we have consensus that this is the place to have it - and we have time to implement it there (it should be the same effort as implementing this new API ).

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,109 @@
...
+message PeerAuthentication {
+  // The selector determines the workloads to apply the ChannelAuthentication on.
+  // If not set, the policy will be applied to all workloads in the same namespace as the policy.
+  istio.type.v1beta1.WorkloadSelector selector = 1;
+
+  // Mutual TLS settings.
+  message MutualTLS {
+    enum Mode {
+      // Connection can be either plaintext or mTLS tunnel.
+      PERMISSIVE = 0;
```
costinm: I don't think we want to continue to expose permissive as a user preference. AutoMTLS removes the need for permissive to be exposed and managed by user.

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,109 @@
...
+
+      // Connection is an mTLS tunnel (TLS with client cert must be presented).
+      STRICT = 1;
+
+      // Connection is not tunneled.
+      DISABLE = 2;
```
costinm: What does it mean ? We already have a mechanism to opt out a port from injection. 

You mean we don't do auto-detection and other filters ? Plain text ? This is quite confusing for user, and should be part of Sidecar or networking APIs, since it doesn't apply to authentication ( there is no authentication on the plain text port).

security/v1beta1/peer_authentication.proto
diemtvu: autoMtls solves different problem. We need this API to do what the existing alpha API doing today.

security/v1beta1/peer_authentication.proto
diemtvu: It means mTLS is disable. I'm not sure what do you mean by the mechanism to opt out a port from injection and what does it have any thing to do in here.

security/v1beta1/peer_authentication.proto
diemtvu: The consensus actually is this API. Please check with @smawson and @louiscryan

costinm:
Adding authorizations checks for the transport properties has been
discussed and I don't remember anyone objecting to it.
Adding a CRD and an API with just one setting doesn't make sense to me and
I don't think any WG (security or networking)
had discussed.

At least the interactions to AutoMTLS and the direction of the networking
should be covered in a networking WG or doc, Louis
has an entire doc about better transport security, etc.

Enforcing strict with a separate CRD instead of authorization - I think the
Security WG can decide if it's the best option for the users.

But exposing control over 'permissive' and non-mTLS settings: I don't think
this is the right way to do it.

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,109 @@
...
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=security.istio.io/v1beta1
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message PeerAuthentication {
+  // The selector determines the workloads to apply the ChannelAuthentication on.
+  // If not set, the policy will be applied to all workloads in the same namespace as the policy.
+  istio.type.v1beta1.WorkloadSelector selector = 1;
```
costinm: Sorry, just realized: this will have very 'dramatic' impact on behavior, meaning that will select ports that are currently not selected - and break most of the stateful set. 

We can't have users of the old API just switch to this API  - and it's not even clear how it can work in general for those cases.

diemtvu: Do you mean the problem related to the pass-through inbound listener? I think we will need to secure all ports on workload with this API, and user may need to use permissive mode in those cases. It's better than today behavior, when users might not realize that their workloads are not protected and traffic is actually in plain-text.

security/v1beta1/peer_authentication.proto
rshriram: Lets call it as UNSET. We can then fallback to permissive internally if needed. But if we are going ot have permissive as a dedicated option, I certainly do not want this to be the default enum. The default enum in that case should be unset=0, as it allows for proper inheritance.

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,109 @@
...
+      DISABLE = 2;
+    }
+
+    // Defines the mTLS mode used for peer authentication.
+    Mode mode = 1;
+  }
+
+  // Mutual TLS settings for workload.
+  MutualTLS mtls = 2;
+}
```
rshriram: We really need the per port setting. Else it will be a regression from the old API. If I want to disable mTLS for a specific port (say the db port 3306 used by mysql clients with their own one-way TLS but leave it on for the replication port like 8899, or for a VM that wants to use one-way TLS for the legacy app port but use mTLS for the newer ports), there is no way to do that in the new API (we can do that in the old API though).

rshriram:
I am 80% okay with costin's proposal of moving the "enforce mesh security" mode to the authZ api (with the ability to add exceptions per port). We can define inheritance mechanisms for this setting alone in the authZ api. One less CRD to contend with. And paves the way for incremental upgrade to the new model. 

How we do the mesh security is our internal implementation detail (mtls, proxy protocol plus some other wrapper, etc.). But from user point of view, it is a 3 way set: mesh-transport is UNSET, REQUIRED, DISABLED. Internally, we can assume unset == permissive at global level, and at individual namespace/workload level, it can inherit from the parent.

security/v1beta1/peer_authentication.proto
diemtvu: I can add UNSET, but want to use it for validation to make sure the mode must be set explicitly  instead. We still need permissive, otherwise, overwrite policy for namespace/workload to permissive seems odd.

Inheritance can base on whether the message (MutualTLS) is set or not.

Wdyt?

security/v1beta1/peer_authentication.proto
diemtvu: We'll do that in the next cycle. For now, I'd like to have a minimum feature set that we can deliver.

Work-around for port-level setting at the moment is set workload to permissive and use authZ to lock down those needed.

diemtvu:
I also agree that authZ can address 80% of the need. But the remain 20% (e.g disable mTLS, the need for sni/alpn config) is the reason I think we still need this. No matter if they are edge cases or not, if we need to invent a new way to do so, the cost is the same, if not more.

It's true that when we have better security transport, more intuitive authZ or networking API, this one can be diminished. But it could easily be a few quarters from now, and I don't see why we have to stop this.

security/v1beta1/peer_authentication.proto
rshriram: I dont think you understand. These are ports receiving TLS traffic to VMs. Setting entire workload to permissive still doesn't help. What people need is a way to disable authZ per port. I dont see the technical limitation to supporting the feature that already existed in the older API.

diemtvu: It's just a matter of time. Anyway, I've added port level settings, hope we can move fast to deliver everything.

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,132 @@
...
+// -->
+message PeerAuthentication {
+  // The selector determines the workloads to apply the ChannelAuthentication on.
+  // If not set, the policy will be applied to all workloads in the same namespace as the policy.
+  istio.type.v1beta1.WorkloadSelector selector = 1;
+
+  // Mutual TLS settings.
+  message MutualTLS {
+    enum Mode {
+      INVALID = 0;
```
rshriram: can you change this to UNSET ? Because this is an inherited field. So invalid doesn't make sense

nrjpoddar:
It looks this PR and https://github.com/istio/api/pull/1248 have overlapping concerns. Additionally, these are big API changes which are pretty close to release-1.5, can we ratify them after the release?

security/v1beta1/peer_authentication.proto
diemtvu: I actually don't want inherit in enum, but inherit if the message type field `mtls` is not set. Using INVALID to require user to explicitly set it.

smawson: With proto3 we can't distinguish unset from set to 0, so the default would be INVALID, which doesn't make sense.

With port settings people can leave this unset just to change a port setting without changing the default.

So rename to UNSET seems fine.

diemtvu: I mean, if users want to inherit, they should not define mtls at all, e.g
```
  selector:
    match:
      app: foo
  portLevelMtls:
    8080:
       mode: DISABLE
```

Instead of 
```
  selector:
    match:
      app: foo
  mtls: {}
  portLevelMtls:
    8080:
       mode: DISABLE
```

The INVALID is not accepted value (i.e we will reject with validation), which force the yaml must explicitly set the mode, so `mtls: {}` is not legit.

Anyway, I'm ok to change to `UNSET` if you think it's better to treat this as inherit.

diemtvu: FYI, I add above example to the API comment. Let me know if it make sense. I happy to change to UNSET otherwise.

smawson: Just change it to UNSET.

Also is there a reason other than distinguishing between UNSET and missing that we're nesting mode inside MutualTLS message? Could we just inline that and just have the enum?

diemtvu: Done. 
The reason for MutualTLS message is for future extension. Not sure if we will need any, but I don't want to get into a dead end.

smawson: OK in that case UNSET is definitely right, because if you extend it we wouldn't be able to distinguish if they leave that unset but set some other field in MutualTLS. So LGTM.

smawson:
We'd like to get these changes into 1.5 and handle #1248 after, since this is really just moving authentication policy to Beta. This is all likely to change with the introduction of improved transport security, but in the meantime users need a beta-level way to control the behavior of the mTLS code.

@rshriram can you give this another look?

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,151 @@
...
+
+// $schema: istio.security.v1beta1.PeerAuthentication
+// $title: PeerAuthentication
+// $description: Peer authentication configuration for workloads.
+// $location: https://istio.io/docs/reference/config/security/peer_authentication.html
+// $aliases: [/docs/reference/config/security/v1beta1/peer_authentication]
+
+// PeerAuthentication defines how traffic will be tunneled (or not) to the sidecar.
+//
+// Examples:
```
liminw: Probably worth to give a mesh level policy example (i.e., use root namespace).

security/v1beta1/peer_authentication.proto
```diff
@@ -0,0 +1,153 @@
...
+//     mode: PERMISSIVE
+// ---
+// apiVersion: security.istio.io/v1beta1
+// kind: PeerAuthentication
+// metadata:
+//   name: default
+//   namespace: foo
+// spec:
+//   selector:
+//     matchLabels:
```
incfly: doe this mean we need to do additional look up for auto mTLS? could that be done efficiently?

rshriram:
> We'd like to get these changes into 1.5 and handle #1248 after, since this is really just moving authentication policy to Beta. This is all likely to change with the introduction of improved transport security, but in the meantime users need a beta-level way to control the behavior of the mTLS code.

I spoke with @diemtvu offline. To make progress, I am okay with this API. I will trim the sidecar API to simply specify TLS settings per port for ports that have mTLS disabled in this API - as that is useful in VM context.

costinm:
My primary requirement is to not include it by default in istio-system. If
you absolutely can't have the default in mesh config and must have the
extra complexity of inheritance ( on top of migration/interop with prev
version) - at least don't force-install this new CR in istio-config.

User may set it, like any other config - and I assume istioctl and tests
will exist for cases of old version still installed and/or for migration.

costinm:
And to be clear - safe upgrade has been a requirement since 1.0.

And multiple control planes has been approved by env wg - so any
install-time resource should be compatible or made compatible in 1.6.

PR #1242, opened 2020-01-23
Update dependencies to latest patch versions

dgn:
As part of https://github.com/istio/istio/issues/20361


dgn:
I changed `gogo/protobuf` back to `1.3.0` as we had issues with that in istio/istio (see https://github.com/istio/istio/pull/20451)

PR #1248, opened 2020-01-24
custom TLS termination at sidecar

rshriram:
When used in conjunction with the PeerAuthentication security api, this enhancement allows the user to define custom certs for one-way or mutual TLS termination at the sidecar (useful in VM context). Note that for the TLS settings defined here to work, the user has to have explicitly declared an exception either for the entire workload or the specific port for which the TLS settings are being applied.


rshriram:
@louiscryan I don't think the Sidecar is the right place to define this TLS termination. The user would have to ensure that the ingress port matches the port in the service definition (or service entry). We would also have to be explicit about the override semantics (auth policy today applies to all ports on a sidecar - whether declared via k8s or added via the Sidecar crd, and sets up mtls).

networking/v1alpha3/gateway.gen.json
```diff
@@ -102,7 +102,7 @@
             }
           },
           "httpsRedirect": {
-            "description": "If set to true, the load balancer will send a 301 redirect for all http connections, asking the clients to use HTTPS.",
+            "description": "If set to true, the load balancer will send a 301 redirect for all http connections, asking the clients to use HTTPS. Not applicable in Sidecar ingress context.",
```
costinm: Not sure why, but seems reasonable for now.

networking/v1alpha3/gateway.proto
```diff
@@ -351,24 +353,27 @@ message Server {
...
-      // gateway workload identity, generated automatically by Istio for
-      // mTLS authentication. When this mode is used, all other fields in
-      // `TLSOptions` should be empty.
+      // Secure connections from the downstream using mutual TLS by
+      // presenting server certificates for authentication.  Compared
+      // to Mutual mode, this mode uses certificates, representing
+      // gateway workload identity, generated automatically by Istio
+      // for mTLS authentication. When this mode is used, all other
+      // fields in `TLSOptions` should be empty. Not applicable in
+      // Sidecar ingress context.
```
costinm: Isn't this the same thing with what Sidecar does by default ( i.e. STRICT mode ) ?

networking/v1alpha3/gateway.gen.json
rshriram: https redirect. Implementing it in sidecar will take some work and thinking. I would rather not advertise this now.

networking/v1alpha3/gateway.proto
rshriram: Yes, but I remember louis saying we shouldn't support istio_mutual here. Or did I get it wrong? I dont have any preference.

networking/v1alpha3/sidecar.proto
```diff
@@ -300,6 +311,10 @@ message IstioIngressListener {
   // or Unix domain socket where the application workload instance is listening for
   // connections. Format should be `127.0.0.1:PORT` or `unix:///path/to/socket`
   string default_endpoint = 4 [(google.api.field_behavior) = REQUIRED];
+
+  // Set of TLS related options that allow a listener to terminate
+  // SIMPLE or MUTUAL TLS connections at the sidecar.
+  Server.TLSOptions tls = 5;
```
diemtvu: Does this make sense to have a default one for all ports?

rshriram:
> My main concern with using this API for the purpose enabling/disabling Istio mTLS is that users need to set egress and ingress (port, capture mode, endpoint etc). Is this possible to make all those optional?

Yes that is what I meant. We can relax those restrictions for the sake of TLS and provide some form of inheritance semantics where needed.

rshriram:
@sureshvis / @mdhume PTAL.

diemtvu:
> > My main concern with using this API for the purpose enabling/disabling Istio mTLS is that users need to set egress and ingress (port, capture mode, endpoint etc). Is this possible to make all those optional?
> 
> Yes that is what I meant. We can relax those restrictions for the sake of TLS and provide some form of inheritance semantics where needed.

This is not exactly comment on the API, but I'd like to see how users can use this for the purpose of mTLS enablement.

I've realized that this API (nor any workload-oriented API for mTLS) will not work well with auto-mTLS (TL;DR: endpoint with annotation mtls is not necessary mtls-capable, as it can be disable by API). We may also need to make server side honor this annotation and overwrite the mTLS setting to give consistent auto-mtls behavior.

For mTLS enablement purpose, I now think a better choice is:
- Have a global flag to switch between permissive and strict. Default is permissive.
- No namespace level settings.
- Use annotation as a mechanism to disable mTLS. This work well with auto-mtls.
- Future works can infer strict vs permissive for authorization for some optimization
- This sidecar API can be considered as advance option. User will get what they set, and should config the companion DR for the client side. 

Wdyt?

mdhume:
@rshriram Had a couple of questions
1. Doesn't the new setting of inboundTrafficPolicy overlap with the Authentication policy that is currently supported? Or will inboundTrafficPolicy be used to specify workload specific overrides while Auth Policy can only be specified at a global or namespace level?
2. The inboundTrafficPolicy modes only seem to support istio_mutual or disabled modes which means that the server will expect `istio` application_protocol as a filter chain match for incoming client requests. Our use-case is to support external non-istio clients which are not able to pass `istio` alpn but still want to make a mutual tls connection to the server which gets terminated at the istio-proxy sidecar. How would we go about specifying that option?
We would prefer to use something like `inboundTrafficPolicy` to specify that as opposed to istioIngressListener to avoid having to list out every inbound port that the server is listening on

rshriram:
>Doesn't the new setting of inboundTrafficPolicy overlap with the Authentication policy that is currently supported? Or will inboundTrafficPolicy be used to specify workload specific overrides while Auth Policy can only be specified at a global or namespace level?

this is meant as a replacement for the auth policy.

>We would prefer to use something like inboundTrafficPolicy to specify that as opposed to istioIngressListener to avoid having to list out every inbound port that the server is listening on

ack. Let me try to rework the API

rshriram:
> I've realized that this API (nor any workload-oriented API for mTLS) will not work well with auto-mTLS (TL;DR: endpoint with annotation mtls is not necessary mtls-capable, as it can be disable by API). We may also need to make server side honor this annotation and overwrite the mTLS setting to give consistent auto-mtls behavior.

@diemtvu currently, even in auto mTLS code, we check if there is a authN policy for that namespace. Only if it is permissive or strict do we enable auto mTLS. Otherwise, we dont enable auto mtls. So, if a user declares a namespace wide sidecar with auto mtls, we can easily detect and adapt. Only when they use a workload specific sidecar, do we need user intervention in the form of a destination rule to set the appropriate TLS mode.

The API as I have in this PR allows us to do 
1. Global mTLS setting using the default Sidecar 
2. Per namespace mTLS setting using namespace wide Sidecar 
3. Auto mTLS will adapt based on the global and namespace wide sidecar.
4. Workload specific sidecar, if it overrides the TLS policy, will then need an associated DestinationRule. But this is no different from what we have today with auto mTLS and workload specific authN policy.

networking/v1beta1/gateway.proto
```diff
@@ -334,43 +334,56 @@ message Server {
...
+      // applicable in Sidecar API.
       AUTO_PASSTHROUGH = 3;
 
-      // Secure connections from the downstream using mutual TLS by presenting
-      // server certificates for authentication.
-      // Compared to Mutual mode, this mode uses certificates, representing
-      // gateway workload identity, generated automatically by Istio for
-      // mTLS authentication. When this mode is used, all other fields in
-      // `TLSOptions` should be empty.
+      // Secure connections from the downstream using mutual TLS by
```
louiscryan: @diemtvu @smawson @costinm @rlenglet 

I really don't want to co-mingle application level TLS concerns with the details of our underlying secure tunneling mechanism. i.e in the medium term I want to deprecate ISTIO_MUTUAL. 

With auto-mtls we are mostly trying to remove the need to configure a secure-underlay from the service owners. I'd rather separate service owner facing TLS concerns from the requirement that a secure underlay is used or not.

i.e

Sidecar
   ingress:
       networkSecurity : **strict** | permissive 
       tls : 
         mode : simple | mutual | ...

Where networkSecurity==strict is defined to mean that all traffic into the sidecar must arrive over a network that provide identity, confidentiality etc. (open to suggestions on terminology) i.e what auto-mtls would provide.

networkSecurity==permissive means that traffic can be received over networks that do not provide this property.


Separating these APIs is important. Firstly it allows app-TLS over a secure underlay to be declared which is important for migrations. It also avoids the conflation of intent between application TLS concerns and Istio's ability to provide a secure underlay network.

rshriram: I had it setup this way previously. But see the comment from mdhume below that said they wanted to specify SIMPLE tls settings for all ports in the ingress, without having to enumerate each and every port of a workload. We could make this a one-of though:
 oneof transportMode {
  IstioTransport {enum networkSecurity ..}
  CustomTransport {TLSOptions tls}
 }

which then clearly conveys that if the developer opts into our underlay, we offer a choice of strict/permissive, and if she chooses the custom transport, then its the standard tls settings from gateway

diemtvu:
> > I've realized that this API (nor any workload-oriented API for mTLS) will not work well with auto-mTLS (TL;DR: endpoint with annotation mtls is not necessary mtls-capable, as it can be disable by API). We may also need to make server side honor this annotation and overwrite the mTLS setting to give consistent auto-mtls behavior.
> 
> @diemtvu currently, even in auto mTLS code, we check if there is a authN policy for that namespace. Only if it is permissive or strict do we enable auto mTLS. Otherwise, we dont enable auto mtls. So, if a user declares a namespace wide sidecar with auto mtls, we can easily detect and adapt. Only when they use a workload specific sidecar, do we need user intervention in the form of a destination rule to set the appropriate TLS mode.
> 
> The API as I have in this PR allows us to do
> 
> 1. Global mTLS setting using the default Sidecar
> 2. Per namespace mTLS setting using namespace wide Sidecar
> 3. Auto mTLS will adapt based on the global and namespace wide sidecar.
> 4. Workload specific sidecar, if it overrides the TLS policy, will then need an associated DestinationRule. But this is no different from what we have today with auto mTLS and workload specific authN policy.

My point is that we need to document well when users are responsible to setting up DR (i.e case 1+2 vs 4). On top of that, we also need to take into account annotation to make sure auto-mTLS work correctly which make that document even more complicate. In my opinion, it is a bad UX.

Instead of that:

1. Have a mesh-config flag to set mTLS of the whole mesh to permissive or strict. *No plaintext*
2. `security.istio.io/tlsMode` will be considered when setting listener TLS. In particular, if the mode is set (by users) to different value that sidecar injector use, then mTLS should be disable. 
3. User can use sidecar for more advance configuration, but they should be responsible to setup DR accordingly. It's possible to discourage using this at namespace and mesh level too (which also simplify the API, and avoid relaxing requirements)

Bottomline, the order of mTLS (server) settings will be sidecar > label > mesh config.

Note:
- We can use the current authN API for mesh-config in 1.5, if we don't have time to make that change.
- (2) should be done to fix the problem for auto-mTLS incorrectly infer the mTLS mode of the service.
- A small correction of your previous comment: for auto-mTLS, at the moment we check the authN policy for the *destination service*, which is doable as API using service-selector.

rshriram:
Removing annotation causes the workload to be redeployed? if not, it may be possible. If it does, then its a no go as it requires people to redploy the workload just to enable mtls. 

either way, I dont think this is bad UX as the auto mtls is hidden from end user. They do not control it mostly as we inject the label. So I dont see how it can interfere with the sidecar. If today, most people use the authN policy to disable mtls on a namespace level (or at workload level), the same UX continues tomorrow with this API. nothing changes except a new API. (including creating the destination rules when auto mtls is disabled for a specific workload). 

Finally, from a security perspective, annotations are much harder to access control compared to the API objects that can be monitored (or control given to security admin and not the developer committing the yaml in the repo).

networking/v1alpha3/sidecar.proto
rshriram: yes I have one for all ports.

networking/v1beta1/gateway.proto
rshriram: Please see the updated protos.

networking/v1alpha3/sidecar.proto
```diff
@@ -427,3 +525,74 @@ enum CaptureMode {
...
+  };
+
+  // Determines the method to use to authenticate the remote
+  // endpoint. To enable secure communications between all VMs and
+  // pods, this should be set to REQUIRED or OPTIONAL. When set to
+  // disabled, the sidecar would not verify the identity of the remote
+  // endpoint. It can still provide transport level security by
+  // terminating the application's TLS traffic using certificates
+  // provided in the `tls` option below.
+  IstioPeerAuthentication istio_peer_authentication = 1;
```
louiscryan: @smawson @rlenglet @costinm @diemtvu opinions on nomenclature requested.

I think this arrangement is pretty sound so my primary remaining concern is nomenclature and documentation. Starting with nomenclature here are some options and their various pros/cons.

- peer_authentication - Doesn't capture the other facets of security in it's name. E.g. confidentiality, integrity, ... Doesn't collide with TLS too badly.
- peer_security - Similar to above though might be a little hard to grok. 
- peer_communication : secure | permissive | insecure   -Variant on above
- network_security - Pretty vanilla but relatively easy to document and covers a lot of ground
- secure_network - Similar to above 
- mesh_security - Similar to above if you consider mesh to be a synonym for networking.
- transport_security - Most descriptive but definitely collides with TLS
- secure_transport - Similar to above but collides a little less with TLS
- secure_tunnel - Sounds a little too much like describing an implementation than an intent but distinguishing between 'tunnel' and 'transport' might help users grok

diemtvu: I find peer_authentication is ok, given it is the term we have used for a while in Istio. Just my 2c.

diemtvu:
> Removing annotation causes the workload to be redeployed? if not, it may be possible. If it does, then its a no go as it requires people to redploy the workload just to enable mtls.
> 
> either way, I dont think this is bad UX as the auto mtls is hidden from end user. They do not control it mostly as we inject the label. So I dont see how it can interfere with the sidecar. If today, most people use the authN policy to disable mtls on a namespace level (or at workload level), the same UX continues tomorrow with this API. nothing changes except a new API. (including creating the destination rules when auto mtls is disabled for a specific workload).
> 
> Finally, from a security perspective, annotations are much harder to access control compared to the API objects that can be monitored (or control given to security admin and not the developer committing the yaml in the repo).

auto-mTLS (at least with the existing implementation) will fail to tell if sidecar (or any workload-selector API) change the mTLS settings. In other words, auto-mTLS will stop working correctly once users set workload-specific sidecar. Given auto-mTLS become default in 1.5, we should be careful that the new API won't break it.

rshriram:
Which is why I defined limited inheritance semantics. Unless I override it per Sidecar, users won’t have to do much. Secondly, if someone is explicitly disabling mtls for a workload, then the configs required are same as today. There is nothing we can do about it.

nrjpoddar:
It looks this PR and https://github.com/istio/api/pull/1241 have overlapping concerns. Additionally, these are big API changes which are pretty close to release-1.5, can we ratify them after the release?

rshriram:
To all the viewers and reviewers, this is how the API will look (the diffs are badly rendered)

```yaml
kind: Sidecar
  metadata:
   name: ratings
   namespace: prod-us1
 spec:
   workloadSelector:
     labels:
       app: ratings
   ingress:
   - port:
       number: 9080
       protocol: HTTP
       name: somename
     defaultEndpoint: unix:///var/run/someuds.sock
   - port:
       number: 9443
       protocol: HTTPS
       name: httpsport
     tls:
       mode: SIMPLE # overrides namespace default
       serverCertificate: /etc/certs/servercert.pem
       privateKey: /etc/certs/privatekey.pem
     defaultEndpoint: unix:///var/run/someuds.sock
   egress:
   - port:
       number: 9080
       protocol: HTTP
       name: egresshttp
     hosts:
     - "prod-us1/*"
   - hosts:
     - "istio-system/*"
---
# and the associated PeerAuthentication security policy to ensure
# that mutual TLS based authentication is not configured for ports
# 9080 and 9443:

 kind: PeerAuthentication
 metadata:
   name: ratings-istio-mtls-exception
   namespace: prod-us1
 spec:
   selector:
     matchLabels:
       app: ratings
   # other ports inherit the settings from namespace-wide policy.
   portLevelMtls:
     9080:
       mode: DISABLE
     9443:
       mode: DISABLE
---
# and the associated DestinationRule to ensure that the clients use
# the appropriate TLS settings:

 kind: DestinationRule
 metadata:
   name: ratings-istio-mtls-exception
   namespace: prod-us1
 spec:
   host: ratings.prod-us1.svc.cluster.local
   trafficPolicy:
    portLevelSettings:
    - port:
        number: 9080
      tls:
        mode: DISABLE
    - port:
        number: 9443
      tls:
        mode: SIMPLE
        caCertificates: /etc/certs/ca-certs.pem
```

rshriram:
>Additionally, these are big API changes which are pretty close to release-1.5, can we ratify them after the release?
@nrjpoddar this is actually a very small API change (simply adding a TLS field in the ingress). Pilot level code changes are also straightforward (reusing the gateway's code as is). It looks big because we have to copy the change into both v1alpha3 and v1beta1

networking/v1beta1/sidecar.proto
```diff
@@ -94,26 +95,47 @@ import "networking/v1beta1/gateway.proto";
...
-// egress direction, in addition to the `istio-system` namespace, the sidecar
-// proxies only HTTP traffic bound for port 9080 for services in the
-// `prod-us1` namespace.
+// The following example declares a `Sidecar` configuration in the
+// `prod-us1` namespace for all pods with labels `app: ratings`
+// belonging to the `ratings.prod-us1` service.  The workload accepts
+// inbound HTTP traffic on port 9080 without any authentication, and
+// HTTPS traffic on port 9443 with one-way TLS termination using
+// custom certificates. _To accomplish custom TLS termination on this
+// workload, the `PeerAuthentication` security policy must be declared
```
diemtvu: How do we enforce this *must be declared to disable* ? Could it be easier if we setup the precedence rule, e.g sidecar > authn ?

rshriram: we dont have to enforce. Remember logically two different personas. If Sidecar owner tries to do one-way TLS but mesh admin says ithas to be ONLY mTLS, then the admin level policy takes precedence. This is how it is done today.

networking/v1alpha3/sidecar.proto
```diff
@@ -44,22 +44,23 @@ import "networking/v1alpha3/gateway.proto";
 // `workloadSelector` that selects this workload instance, over a `Sidecar` configuration
 // without any `workloadSelector`.
 //
-// NOTE 1: *_Each namespace can have only one `Sidecar` configuration without any
+// **NOTE 1**: *_Each namespace can have only one `Sidecar` configuration without any
```
costinm: Can we add 'the recommended name is 'default' ' ( or maybe even require it ). 

Easier to fix.

networking/v1alpha3/sidecar.proto
```diff
@@ -125,18 +147,66 @@ import "networking/v1alpha3/gateway.proto";
...
-// `127.0.0.1:8080`. It also allows the application to communicate with a
-// backing MySQL database on `127.0.0.1:3306`, that then gets proxied to the
-// externally hosted MySQL service at `mysql.foo.com:3306`.
+// and the associated PeerAuthentication security policy to ensure
+// that mutual TLS based authentication is not configured for ports
+// 9080 and 9443:
+//
+// ```yaml
+// apiVersion: security.istio.io/v1beta1
+// kind: PeerAuthentication
```
costinm: 1. can we rename it to networking.istio.io - since this is now mainly handling transport-level settings, and disabling or fine tuning TLS?

2. Can we call it Transport(Config|Setting) ?

3. Can we move the TLS settings from Sidecar to this new config ?

networking/v1alpha3/sidecar.proto
diemtvu: +1 on fix the namespace level to `default`. It's easy to have webhook validation to enforce that, and/or ignore at runtime.

networking/v1alpha3/sidecar.proto
rshriram: This is squarely in the security folks domain. I am simply referring to that config here to get the right property working. In terms of having a separate CRD, I am less convinced that we need a separate CRD just with TLS. @louiscryan 's motivation was that we should separate the Istio's secure transport from the application level concerns. And application level stuff should be additive on top of whatever Istio transport is. Yes the implementation is not there today, but once we have it in place, we wont have to change the API at all, if we follow this strict separation model.

costinm: Setting up plain text on a port ? Or in general configuring the network ports ? Seems squarely networking :-), as much as the TLS settings in gateway, dr and the new ones in sidecar. 

The 'security' part is in authorization already. 

I agree with @louiscryan that we should separate application level from transport concerns. 
But we should not separate one setting of TLS - disable - from the rest of the settings. 
If we add TLS settings in Sidecar and user specifies a SNI and cert - will the 'disable' in PeerAuthentication have priority ?

networking/v1beta1/sidecar.proto
costinm: And the authorization policy takes precedence over both - if the authz admin sets "only user 'admin' is allowed to access" - anything with plaintext will be denied.

costinm: The problem is that users may have to deal with 3 possibly conflicting configs. And that's the last thing we want in security, we have too much confusion already.

costinm:
I'm sorry, but the example is extremely confusing. 

User explicitly sets in Sidecar that he wants TLS SIMPLE mode. 
He needs to also go to PeerAuthentication to set 

portLevelMtls:
     9443:
       mode: DISABLE

The result will be plain TLS.

However if he only does PeerAuthentication - the result is plaintext.

In the first case - why would the user need to edit both CRDs, when the intent is clear from the first config ?

If we add the Authorization settings on top - it gets very confusing.

costinm:
Unfortunately we can't fix to default - will conflict with the
mutli-control plane and canary plans.

networking/v1beta1/sidecar.proto
hzxuzhonghu: +1 also think this limitation is too confusing. How about making it take preference over the peer auth?

diemtvu: Per comment, peer auth is preferred over sidecar, unless it is `disable`. IMHO, this is a reasonable behavior. AuthZ is a different thing, and even though today `principal` is tightly related to mTLS, its semantic won't necessary be the same.

In the previous draft of this PR, where we have a separate `IstioPeerAuthentication` block, we have the same complexity in deciding what takes over. The fact they are on the same CRD doesn't help, as you still need to define a similar preference rule, (e.g `IstioPeerAuthentication` has higher precedent)

If using the TLSOption offer in this PR for both peer auth and applicaiton TLS, we can get rid of such rule. But then, we won't be able to distinguish application TLS and Istio channel security, mixing operation persona. On top of that, being a big fat API, but the combination rule is limited at "the latest win" could be a trouble. We have seen such an issue in DR, when defining one thing nullify the other.

networking/v1alpha3/sidecar.proto
rshriram: Its the other way.. TLS settings in sidecar with sni+cert will only be valid if peer authentication is disabled. I added a line in the doc explicitly stating so.

rshriram:
> I'm sorry, but the example is extremely confusing.
> 
> User explicitly sets in Sidecar that he wants TLS SIMPLE mode.
> He needs to also go to PeerAuthentication to set
> 
> portLevelMtls:
> 9443:
> mode: DISABLE
> 
> The result will be plain TLS.
> 
> However if he only does PeerAuthentication - the result is plaintext.
> 
> In the first case - why would the user need to edit both CRDs, when the intent is clear from the first config ?
> 
> If we add the Authorization settings on top - it gets very confusing.

As I mentioned elsewhere, if it is a single persona doing everything, we could just have one CRD for all. If it is two different personas, then we need separate controls over the peer authentication and over the application level authentication. In the near future, the user will be able to specify TLS mode as simple for his application irrespective of the peer authentication, i.e. we should be able to do both peer authentication as well as the simple TLS termination requested by the user. The API, in its current form allows for this future functionality, and thus remains stable as we evolve the data plane to do both. Until then, it will simply be a doc level note that due to envoy limitations today, users will have to disable peer authn to do simple TLS. Long story short, as @louiscryan said,

```which is better ? an unstable API or some invalid permutations of a stable one? e.g (peerauthN=strict & tls=<on>) is not valid today but could be in the future```

**Comment by costinm** [comment]
But that's worse, and it breaks one of the most important use cases for
this feature.
SNI+cert allows you to migrate an existing app that is already using TLS or
mTLS to Istio certs.
You configure SNI - with both existing and new certificate, both in mTLS
mode. And you use
authorization policy to allow both. In both cases the peer is authenticated
- maybe non-Istio
auth will use JWT or something else, but still TLS and authenticated.
If this mode requires peer auth policy to be disabled - it'll be very
confusing, because in fact peer authentication
is used.

In general - the single per-port on or off 'peer authentication policy' is
very poor for expressing what we need
for orgs already using TLS - that's one of the reasons we need the SNI and
full control over the TLS
settings in the first place. And it's also why we need the fine control we
have in authz.

**Comment by costinm** [comment]
And as I mentioned, I am now fully in support for 2 personas and 2 configs,
this thread and other discussions make it
clear to me that TLS is tricky and we need to minimize confusion :-)

In other words: leave Sidecar as it is (stable), and add the (stable) TLS
settings from Gateway to a standalone resource
dedicated to port-level TLS settings.

**Comment by costinm** [comment]
I'm now pretty sure in 1.5 we can express this using EnvoyFilter - adding
or removing the config.
Shriram - can you confirm, I haven't yet tested. That will allow people to
use any of the Envoy
options to fine tune custom TLS configuration, including more advanced
modes like 'plain TLS on a different
port, forwarding to the same inbound cluster', or turning off TLS on
specific port.

 This gives us more time - so we can add the proper TLS API in 1.6.

I don't mind too much if we merge this and add it as part of Sidecar, but
I'm not more convinced
that creating a top level CRD specialized on TLS settings would be cleaner
and more future-proof,
considering K8S Gateway and other use cases ( aligned with the EnvoyFilter
patch model - which
I think is remarkably clear).

**Comment by rshriram** [comment]
> I'm now pretty sure in 1.5 we can express this using EnvoyFilter - adding
or removing the config.
Shriram - can you confirm, I haven't yet tested. That will allow people to
use any of the Envoy
options to fine tune custom TLS configuration, including more advanced
modes like 'plain TLS on a different
port, forwarding to the same inbound cluster', or turning off TLS on
specific port.

It is very complicated to do this in envoy filter partly because one has to manipulate the inbound filter chains, under VirtualInbound to add this. Given that this is such a common use case for VM based users, it makes sense to have an API driven way to enable this. This is an application level TLS concern where the application wants to offload TLS to Sidecar. Nothing to do with peer authN etc. Besides, the folks at Yahoo! are also looking to use this with their custom CA (athens I think). So we have multiple people demanding a common feature/

**Comment by costinm** [comment]
To be clear: I approved the PR and I still believe it's a good thing and should move on. The NetWG also discussed and approved this.

My point is that in 1.6 time frame we should consider a separate CRD, with only the TLS settings, mostly identical with what we do in Sidecar and Gateway, and using 'patch' semantic like EnvoyFilter. 
This will allow separating the TLS configs - for users who want to use different RBAC filters or not confuse regular developers using Sidecar mostly for isolation with TLS-related stuff. 
We can also hold it until K8S GAteway is better defined.

File: networking/v1alpha3/sidecar.proto [hunk 9]
```diff
@@ -133,28 +137,49 @@ import "networking/v1alpha3/gateway.proto";
...
+// belonging to the `ratings.prod-us1` service.  The workload accepts
+// inbound HTTP traffic on port 9080 without any authentication, and
+// HTTPS traffic on port 9443 with one-way TLS termination using
+// custom certificates. _To accomplish custom TLS termination on this
+// workload, the `PeerAuthentication` security policy must be declared
+// to disable Istio mutual TLS on these two ports. Any other
+// auto-generated listener for this workload will still obey the
+// mutual TLS termination requirements set forth in the
+// PeerAuthentication policy_. The traffic is then forwarded to the
+// attached workload instance listening on a Unix domain socket. In
```
linsun: Unix domain socket, as specified in defaultEndpoint?

networking/v1alpha3/sidecar.proto
```diff
@@ -194,18 +231,94 @@ import "networking/v1alpha3/gateway.proto";
...
-// on port 9080 and forward it to the application listening on
-// `127.0.0.1:8080`. It also allows the application to communicate with a
-// backing MySQL database on `127.0.0.1:3306`, that then gets proxied to the
-// externally hosted MySQL service at `mysql.foo.com:3306`.
+// and the associated PeerAuthentication security policy to ensure
+// that mutual TLS based authentication is not configured for ports
+// 9080 and 9443:
+//
+// ```yaml
+// apiVersion: security.istio.io/v1beta1
```
linsun: should this be alpha first?

networking/v1alpha3/sidecar.proto
```diff
@@ -194,18 +231,94 @@ import "networking/v1alpha3/gateway.proto";
...
-// container) and the proxy metadata `ISTIO_META_INTERCEPTION_MODE` is set to
-// `NONE`, the specification, below, allows such pods to receive HTTP traffic
-// on port 9080 and forward it to the application listening on
-// `127.0.0.1:8080`. It also allows the application to communicate with a
-// backing MySQL database on `127.0.0.1:3306`, that then gets proxied to the
-// externally hosted MySQL service at `mysql.foo.com:3306`.
+// and the associated PeerAuthentication security policy to ensure
+// that mutual TLS based authentication is not configured for ports
+// 9080 and 9443:
+//
```
linsun: Is this disablement for both inbound and outbound for the ports declared (9080 & 9443)?

linsun:
@rshriram I do have a general questions on overall user's flow for this.  For the flow @costinm summarized earlier, this only applies if user has mesh wide or namespace wide or service specific mTLS strict enabled right?  Essentially, when that happens, users need to use Peer Authentication to disable the strict mTLS and allow TLS simple.  Am I interpreting the steps correctly?


---//----
User explicitly sets in Sidecar that he wants TLS SIMPLE mode.
He needs to also go to PeerAuthentication to set

portLevelMtls:
9443:
mode: DISABLE

The result will be plain TLS.
---//----

networking/v1alpha3/sidecar.proto
rshriram: This is the new security Peer AuthN api (which is v1beta1)

networking/v1alpha3/sidecar.proto
rshriram: So this is not stuff in my Pr. I am simply repeating what is in Peer Authn for a complete example :);

rshriram:
```
---//----
User explicitly sets in Sidecar that he wants TLS SIMPLE mode.
He needs to also go to PeerAuthentication to set

portLevelMtls:
9443:
mode: DISABLE

The result will be plain TLS.
---//----
```

Yep.

networking/v1alpha3/sidecar.proto
linsun: ok I missed that. I had thought your PR also proposed this peer authN API

networking/v1alpha3/sidecar.proto
linsun: ok fair - but the questions triggers when you put them together with sidecar examples that uses ingress and egress.

networking/v1alpha3/sidecar.proto
linsun: from Shriram: Yes

networking/v1alpha3/sidecar.proto
linsun: from Shriram: Peer Authn is just for inbound. Not for outbound. Outbound is auto mtls

mesh/v1alpha1/config.proto
```diff
@@ -192,14 +192,17 @@ message MeshConfig {
...
-  // dependencies, instead of using allow_any, so that traffic to these
-  // services can be monitored.
+  // Set the default behavior of the sidecar for handling outbound
+  // traffic from the application.  If your application uses one or
+  // more external services that are not known apriori, setting the
+  // policy to ALLOW_ANY will cause the sidecars to route any unknown
+  // traffic originating from the application to its requested
+  // destination.  Users are strongly encouraged to use ServiceEntries
+  // to explicitly declare any external dependencies, instead of using
+  // allow_any, so that traffic to these services can be
```
louiscryan: s/allow_any/ALLOW_ANY

PR #1250, opened 2020-01-25
Revert "Add without_headers field to HTTPMatchRequest for specifying …

howardjohn:
…if-not-present semantics (#1233)"

This reverts commit b255104af796489597addc7f64fd1c9934efb7f5.

This is a workaround for https://github.com/istio/istio/issues/20386

I think our CRD generation is broken

Alternatives:
* revert v1beta1
* wait for proper fix, and block all other API changes until we get htat


howardjohn:
note - there is nothing inherintly wrong with this pr, it just happened to trigger some bug in our CRD generation code. given this is blocking critical 1.5 changes and the owner of the CRD gen code is ooo, a revert seems appropriate

ramaraochavali:
@howardjohn Is having it in `v1alpha3` and NOT in `v1beta1` could be the cause of the bug in CRD generation? 

I fixed it in https://github.com/istio/api/pull/1249 - I copied it `v1beta1` as well. Not sure if it helps though, but FYI - that is one difference I can see with this PR.

howardjohn:
Rama, I tried copying it to beta as well and it still broke. Are you sure
yours worked? it fails when you apply the crd but will pass all tests here

ramaraochavali:
No. I am not sure I haven’t tried it. I was just wondering if that is the
reason- if you have tried then that is not the problem then. Let us just
revert it

howardjohn:
@igsong fyi. Sorry this isn't the fault of your PR

PR #1254, opened 2020-01-27
Add source_namespace to match attributes

jackkleeman:
It would be useful to be able to route on source namespace additional to source labels. This should be fairly straightforward to implement, with very similar properties to source_labels, using `sourceMatchHTTP` and equivalents


jackkleeman:
Not sure how to ensure I don't trigger https://github.com/istio/api/pull/1250

networking/v1beta1/virtual_service.proto
```diff
@@ -719,6 +719,11 @@ message HTTPMatchRequest {
   // **Note:** The case will be ignored only in the case of `exact` and `prefix`
   // URI matches.
   bool ignore_uri_case = 10;
+
+  // Source namespace constraining the applicability of a rule to workloads in that namespace.
+  // If the VirtualService has a list of gateways specified in the top-level `gateways` field,
+  // it must include the reserved gateway `mesh` for this field to be applicable.
+  string source_namespace = 12;
```
rshriram: changes need to go to both v1alpha3 and v1beta1

PR #1255, opened 2020-01-27
Sync between v1alpha3 and v1beta1 and add back reverted commit

jasonwzm:
Add https://github.com/istio/api/pull/1233 back. Note that in order to make same schema work in CRD generation, the fields need to be exactly the same between v1alpha3 and v1beta1.


rshriram:
I think we need to revert that PR in all versions. The author @igsong has not responded yet about implementing it in code. So, we dont want to release that api change without a backing implementation

jasonwzm:
The original commit was reverted in all versions (master and release-1.5). 

@igsong Are you working on an implementation for this?

igsong:
> The original commit was reverted in all versions (master and release-1.5).
> 
> @igsong Are you working on an implementation for this?

Oh, I was waiting for adding back my commit reverted. I've done to implement the corresponding behaviors on istio/istio.

rshriram:
Where is the PR in istio/istio ?

igsong:
There is no PR yet. As I mentioned above, I was waiting for adding the reverted commit in order to set dependencies in go.mod properly.

igsong:
I've just submitted PR for istio/istio. https://github.com/istio/istio/pull/20629

PR #1263, opened 2020-01-30
Add json codegen for Operator

jackkleeman:
We have a pretty nasty bug where fields with two names, eg
UnvalidatedValues, aren't getting unmarshaled properly in k8s clients
because we haven't set up jsonpb unmarshaling for the Operator. This
should solve it.


jackkleeman:
Had to do some pretty dodgy stuff in the makefile because our GOTYPE annotations don't place nice with json codegen

morvencao:
This is needed for release-1.5 to fix: https://github.com/istio/istio/issues/21554

PR #1266, opened 2020-01-31
Add networking v1beta1 example to docs

jasonwzm:
- Networking v1beta1 API examples are shown in another tab along with v1alpha3 API examples.
- v1beta1 `.pb.html` files are not generated to avoid confusion.


networking/v1alpha3/destination_rule.proto
```diff
@@ -43,13 +45,31 @@ import "gogoproto/gogo.proto";
...
+// metadata:
+//   name: bookinfo-ratings
+// spec:
+//   host: ratings.prod.svc.cluster.local
+//   trafficPolicy:
+//     loadBalancer:
+//       simple: LEAST_CONN
+// ```
+// {{</tab>}}
+// {{</tabset>}}
```
rshriram: why is this being duplicated o the v1alpah3 api?

jasonwzm: Our istio.io imports and is built on top of the v1alpha3 generated html files. So to have things keep working, the v1beta1 examples are put together with v1alpha3 examples.

PR #1270, opened 2020-02-05
Use v1alpha3 as the storage version for networking APIs

jasonwzm:
To support the downgrade scenario to istio 1.4. 

fixes: https://github.com/istio/istio/issues/20861


jasonwzm:
cc/ @costinm

PR #1274, opened 2020-02-07
Add mesh ID to ProxyConfig

carolynhu:
This work is based on the [Adding meshID for telemetry](https://docs.google.com/document/d/1URXkFu2-cSg-g2fJin666M_QJvkumdn-C1NY9XkYgIM/edit#heading=h.vc0c4rza296u) proposal.

We would like to be able to aggregate telemetry from multiple Istio meshes into a single place, and then be able to query it, distinguishing which mesh particular entries came from.

An Istio mesh currently knows of no unique ID that represents itself, so it has no way to label telemetry so that it can be aggregated across multiple meshes as desired.


mesh/v1alpha1/config.proto
```diff
@@ -105,6 +105,9 @@ message MeshConfig {
   // Name of theKubernetes service used for the istio ingress controller.
   string ingress_service = 8;
 
+  // ID of the mesh
+  string mesh_id = 50;
```
rshriram: ? Any explanation on why this PR is needed?

carolynhu: I'm exporting pilot metrics to stackdriver. Where in the Stackdriver Monitoring labels, we want to add something like this: `labels.Set("mesh_id", "xxx", "ID for Mesh")`. Currently, we cannot get mesh_id from MeshConfig, that's why I'm adding this PR. Thanks.

carolynhu: /cc @douglas-reid @howardjohn

douglas-reid: I think we should clarify the relationship to `trust_domain`, as I believe those are used somewhat interchangeably in the install artifacts. We definitely want to add a description of the purpose for a mesh identifier to the comment.

Also, given that this is already `MeshConfig`, would it make sense to just use `id` here?  Part of me thinks it would, but I can also see that being too generic.

linsun: I agree with @rshriram we need to be clear on why this is used.  Is it used for multi cluster or multi-mesh telemetry scenario to differentiate which meshes?  Or is this for multi-tenancy scenarios?  or both?

cc @knrc

douglas-reid: @linsun mesh ids as an Istio concept were approved in [this proposal](https://docs.google.com/document/d/1URXkFu2-cSg-g2fJin666M_QJvkumdn-C1NY9XkYgIM/edit#heading=h.x9snb54sjlu9). the reasoning for their existence is provided there.

This PR is about exposing them more formally as a part of the MeshConfig.

knrc: @linsun and adding to @douglas-reid's comment it's not related to our multi-tenancy work.

smawson: Can we update this with a better comment clarifying exactly what it is for? This is the documentation that ends up in the reference right? We should say something about how this identifies the mesh as part of telemetry reporting so services can be grouped across clusters or whatever.

Basically the "motivation" from the mesh ID proposal.

linsun: Thanks @douglas-reid @knrc!  Agree with @smawson - I'll approve when the comment is clear to user.

mesh/v1alpha1/config.proto
```diff
@@ -105,6 +105,16 @@ message MeshConfig {
...
   string ingress_service = 8;
 
+  // ID of the mesh
+  // There are several usages of mesh ID:
+  // - Telemetry: add mesh ID as part of labels for metrics and logs to
+  // aggregate telemetry/logs from multiple meshes into a single place.
+  // If not specified, we will use the value of the mesh's Trust Domain.
+  // - Multicluster: mesh ID can be set on all clusters to differentiate meshes.
+  // - Security: mesh ID can be added as part of identity in certificates, to
+  // support federation of meshes that have the same trust domain.
```
linsun: for the 3 scenarios, can you clarify if trust domain will be used in absent of mesh id?  Or is that just for telemetry?

costinm:
In 'globals' we use meshID - any reason to change it to mesh_id ( or to
have used meshID in the install API ) ?
I don't mind either - but it's confusing to keep changing, and error prone.

I am still not clear of the relation with trust domain - and the suffix for
global names ( podDNSSearchNamespace / ".global" )
I understand there are corner cases where each would have a different
value, but it would hardly be a best practice.

If a user wants the easiest to understand and most consistent
configuration, as a best practice - which name should he
set ? For example, in a MC mesh I could use "prod.example.com" and global
namespace (which would also be used with the
MC DNS resolver), and this would also be used in telemetry and as trust
domain. Which setting should the user
make to get to this configuration ? Set mesh_id and leave the other 2 blank
? Set all 3 to same value ? Set trustDomain
and leave the other 2 blank ?

Please also remove the last part - I don't think we don't have a clear
understanding/agreement of the format of the cert
or how 'mesh federation with same trust domain' would behave, nor do we
have code to support either -
I think it would confuse users.

For multi-cluster - please add something about the relation to the DNS name
suffix. It's particularly important since we
can now support both control plane and workloads with DNS/ACME certificates.

carolynhu:
After offline discussions with @costinm and @douglas-reid , we decided to keep this PR open until we figured out and have an agreement of the relationship between trust domain and mesh ID and some specific usages/needs of mesh ID.

douglas-reid:
@carolynhu this has come up again. I think we had some consensus around mesh id that I'd like to preserve. Having the ID accessible and stored in a retrieval place seems like a worthy goal.

How can I help move this forward?

costinm:
There is a very active discussion in environments about multi-tenant.

I think it's likely we'll have a 'tenant ID' - which may or may not be the
same with the trustedDomain.
It is not clear at all if a 'mesh' will have multiple tenants, or a tenant
will have multiple meshes.

If anyone can finds a definition for what a mesh is - that everyone agrees
on - I'm all for adding the mesh id.

douglas-reid:
> There is a very active discussion in environments about multi-tenant. I think it's likely we'll have a 'tenant ID' - which may or may not be the same with the trustedDomain. It is not clear at all if a 'mesh' will have multiple tenants, or a tenant will have multiple meshes. If anyone can finds a definition for what a mesh is - that everyone agrees on - I'm all for adding the mesh id.

We already have a concept of mesh id. It is used and exposed as an install option: [`global.meshID`](https://github.com/istio/istio/blob/master/manifests/charts/global.yaml#L297). It comes with the following description:

```
Mesh ID means Mesh Identifier. It should be unique within the scope where 
meshes will interact with each other, but it is not required to be
globally/universally unique. For example, if any of the following are true,
then two meshes must have different Mesh IDs:
- Meshes will have their telemetry aggregated in one place
- Meshes will be federated together
- Policy will be written referencing one mesh from the other
```

We discuss it in our public facing concept docs as well :

> To avoid service naming collisions, you can give each mesh a globally unique mesh ID, to ensure that the fully qualified domain name (FQDN) for each service is distinct.

-- https://istio.io/latest/docs/ops/deployment/deployment-models/

But we have no way to retrieve it once set -- other than parsing out the sidecar injector template.

smawson:
Our current definition of mesh is here: https://istio.io/latest/docs/reference/glossary/#service-mesh

Probably doesn't actually help this discussion much, but a mesh is a single universe of names, a single root of trust and a single discovery domain.

Our current tenancy models are described here: https://istio.io/latest/docs/ops/deployment/deployment-models/#tenancy-models

We support namespaces as tenants within a mesh, or mesh-per-tenant if stronger isolation is needed.

In either case meshID seems completely reasonable, and each control plane should be in a single mesh.

If in the future we come up with some new tenant model that is not namespace or mesh, we can add that on top, since tenant would not be equal to mesh, so we would need both IDs.

So yeah we should update this PR and get it in.

mesh/v1alpha1/config.proto
```diff
@@ -105,6 +105,18 @@ message MeshConfig {
   // Name of the Kubernetes service used for the istio ingress controller.
   string ingress_service = 8;
 
+  // ID of the mesh
+  // In general, a mesh is a single universe of names, a single root of trust and a single discovery domain.
+  // It helps with better isolation when needed under multi-tenant or multi-cluster scenarios.
```
carolynhu, line 110: Not sure if this is a good description or we need something else to describe our purpose of adding meshID?

@smawson @douglas-reid @costinm Please take a look, thanks!

smawson: Can you copy the definition from istio.io glossary or point to it?

I wouldn't actually define mesh here and instead rely on definition elsewhere. This description should just say what mesh id is for.

So something like:

// The unique identifier for the service mesh (link to glossary definition).
// There are several usages of mesh ID:
// ...

mesh/v1alpha1/config.proto
```diff
@@ -105,6 +105,18 @@ message MeshConfig {
...
+  // There are several usages of mesh ID:
+  // - Telemetry: add mesh ID as part of labels for metrics and logs to
```
smawson, line 112: - Telemetry: mesh ID will be included in labels on metrics and logs to allow aggregating telemetry from multiple control planes.

Also the "if not specified" thing should maybe go earlier, that isn't a usage of mesh ID but information about how it is derived.

mesh/v1alpha1/config.proto
```diff
@@ -105,6 +105,18 @@ message MeshConfig {
...
+  // aggregate telemetry/logs from multiple meshes into a single place.
+  // If not specified, we will use the value of the mesh's Trust Domain.
+  // - Multicluster: mesh ID can be set on all clusters to differentiate meshes.
```
smawson, line 115: This doesn't really describe the usage, what is it actually used for?

mesh/v1alpha1/config.proto
```diff
@@ -105,6 +105,18 @@ message MeshConfig {
...
+  // - Security: mesh ID can be added as part of identity in certificates, to
```
smawson, line 116: I don't think this is actually true today is it? So this is a future usage? Are we actually planning on doing this?

Last I had heard we were relying on trust domain being unique and not having both trust domain and mesh ID in certs. So this may not be accurate.

I'd prefer we only include the actual uses of mesh ID here. If it is only for telemetry, just list the telemetry use case.

mesh/v1alpha1/config.proto
```diff
@@ -105,6 +105,18 @@ message MeshConfig {
...
+  // support federation of meshes that have the same trust domain.
+  string id = 54;
```
smawson, line 118: Call it mesh_id rather than just id. I know this is mesh config but id could mean the id of the config rather than the id of the mesh. Especially since the comments talk about mesh ID.

costinm:
Yes, please don't use 'id' - there are many ids, we can't take over the key
for the least user-visible use.

Also - it may be better to have it in ProxyConfig. Proxyes likely will need
to get this info.

MeshConfig is really a istiod revision config - the name is confusing, but
it's no longer associated with mesh, nor do we require that an istiod
serves only one tenant or only one mesh.

So 'default' ProxyConfig will be used by istiod to configure proxies, and
overrides may allow additional trustdomains/meshes to be served.

costinm:
For additional background.. Before envoy gets config from Istiod, we need
to generate a cert - so we can connect to istiod.

The cert must include trustDomain - which means trustDomain must be
available in sidecar. Currently it's a flag, and we need
to move it to ProxyConfig.
Another parameter is 'CA_ADDR' - which also needs to get into ProxyConfig,
and determines the CA root.

For tenant - the namespace is part of the cert as well, and it works with
the definition 'tenant == namespace'.

My question is: if mesh ID is 'a single universe of names, a single root of
trust and a single discovery domain' - how will
it be reflected in the certificate, and indirectly into Policy and other
APIs ?

Single root of trust generally means CA public key plus trust domain (for
spiffee names at least). Single discovery domain
is more complicated - at least I don't know even what it means, so no clue
how would be implemented/enforced.

Will 2 tenants be allowed to have different mesh IDs ? We provide options
for them to get different root CAs/trust domains.

It is very nice to define 'abstract' APIs and names - but at least
internally we need to understand if there is any hope
of actually implementing them correctly before users see them and assume
they work. So it would be great for
Security WG to sign off on whatever changes are needed to spiffee format we
use to include mesh ID ( or confirm
that mesh ID can't be used to securely identify the mesh ). And it would be
great for Networking+Env WG to confirm if
'single universe of names'/'single discovery domain' can be implemented and
how.  For example if I create to
Istiod revisions ( in same or different namespaces) and set different
meshID - would they somehow get a separate
naming ? I can see the mesh ID getting into the Spifee ID ( with a large
effort ! ), but would it be used in cluster name
(i.e. imply different domain ) ?

carolynhu:
okay, so for now, I will only add telemetry usages of meshID and move meshID to ProxyConfig. Later, if we have other clear usages of meshID will add those accordingly.

carolynhu:
@smawson please take another look

mesh/v1alpha1/proxy.proto
```diff
@@ -389,6 +389,13 @@ message ProxyConfig {
   // sleeps for the termination_drain_duration and then kills any remaining active Envoy processes.
   // If not set, a default of 5s will be applied.
   google.protobuf.Duration  termination_drain_duration = 29;
+
+  // The unique identifier for the [service mesh](https://istio.io/latest/docs/reference/glossary/#service-mesh)
+  // All control planes running in the same service mesh should specify with the same mesh ID.
```
smawson, line 394:
```suggestion
  // All control planes running in the same service mesh should specify the same mesh ID.
```

mesh/v1alpha1/proxy.proto
```diff
@@ -389,6 +389,13 @@ message ProxyConfig {
...
+  // There are several usages of mesh ID:
```
smawson, line 395: If we're only listing one we shouldn't say there are several uses. Just state that this is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.

costinm:
If this is mainly for telemetry - can we use a less generic term ?
MetricsID for example ?

This will both be more clear - and more flexible. Maybe 2 namespaces want
to separate their metrics. No need to tie telemetry id with a big concept
as mesh.

costinm:
I mean - we say mesh is a unit of naming and discovery uniqueness, and we
may use it this way in future. Metric separation seems a much different
space.

mesh/v1alpha1/proxy.proto
carolynhu: yes, for now, we only have clear usages for telemetry. Sure, let me update it.

carolynhu:
@douglas-reid what's your opinion? do we name it mesh_id or metrics_id?

douglas-reid:
> @douglas-reid what's your opinion? do we name it mesh_id or metrics_id?

This is called `mesh id` everywhere -- including the install params. Having it be `metrics id` would be needlessly confusing, imho.

carolynhu:
@smawson updated the description, please take another look.

costinm:
Install params are not really an API - and mesh id has its purpose, as
documented.

The question is if metrics id use cases are equivalent with mesh id - i.e.
a customer will never use 2 metrics id ( segregate metrics )
in the same mesh, etc. I would see use cases for that - and I suspect if
'isolated telemetry' is one of the goals, we
kind of support it for some of the providers. For example datadog and other
providers take a secret that I think effectively
directs the telemetry to that account - allowing even 2 pods to isolate
their metrics.

We would add metrics id to the install options as well ( in practice
anything in ProxyConfig is an install option as well ).

costinm:
To be clear: I'm not against adding meshID to ProxyConfig, it is one of the
values.yaml settings we need to migrate.

I'm just worried that the use cases may not align, and it may be better to
be more explicit. For example we have
a 'trustDomain' for security plus the CA_ADDR which determines the root,
dns suffix for networking, etc.
Isolation has many dimensions, with different granularities - meshID has
pretty broad semantics.

PR #1275, opened 2020-02-07
Tag must be interface{} type for unmarshaling

ostromart:
Fixes https://github.com/istio/istio/issues/19921.


proto.lock
```diff
@@ -44239,7 +44239,7 @@
               {
                 "id": 13,
                 "name": "tag",
-                "type": "string"
+                "type": "TypeInterface2"
```
richardwxn: why some referring to typeInterface2 while others referring to typeInterface

ostromart: It's because fixup_structs deals with files rather than packages, meaning we have to redefine synthetic types in every file. But protoc _does_ deal with package scope, which means we cannot call it the same thing. There's a comment at the bottom. It would be nice to fix it, but quite a lot of work.

linsun:
LG (assuming no user impact on the APIs)

ostromart:
I've created https://github.com/istio/istio/issues/20962 to track the file vs. package scope issue. Hope someone will pick it up, it's a good starter issue :-)

PR #1277, opened 2020-02-07
fix operator go struct tag

morvencao:
More context: https://github.com/istio/api/pull/1237#issue-365133001

### DIFF
--- a/operator/fixup_structs/main.go
+++ b/operator/fixup_structs/main.go
@@ -125,16 +125,18 @@ func main() {
 
 			tmp = append(tmp, lines[i+1])
 			i += 2
-		case strings.Contains(l, "json") && strings.Contains(l, "_"):
-			lv := strings.Split(l, goFieldJsonTagPrefix)
-			if len(lv) == 2 && strings.Contains(lv[1], "_") {
-				jsonTag := lv[1]
-				rgx := regexp.MustCompile("_[a-z]")
-				lv[1] = rgx.ReplaceAllStringFunc(jsonTag, func(m string) string {
-					return strings.ToUpper(strings.TrimPrefix(m, "_"))
-				})
-			}
-			tmp = append(tmp, strings.Join(lv, goFieldJsonTagPrefix))
+		case strings.Contains(l, goFieldJsonTagPrefix) && strings.Contains(l, "_"):
+			rgx := regexp.MustCompile(`json\:"[a-z|0-9]+_[a-z|0-9]+,`)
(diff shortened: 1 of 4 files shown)


morvencao:
```
```

morvencao:
```
IngressGateways      []*GatewaySpec     `protobuf:"bytes,40,rep,name=ingress_gateways,json=ingressGateways,proto3" json:"ingressGateways,omitempty"
```
Let's take the `IngressGateways` as an example, the tags for this field consists with two parts, `protobuf` and `json`; and also there is another `json` attribute in `protobuf` part. 

From our investigation, the `json` attribute in `protobuf`  tag is used for jsonpb lib to unmarshal/marshal, while `json` tag is used for json lib to unmarshal/marshal, which is used by kubernetes' go client AFAIK, so if we have go field with snake case： `json:"ingress_gateways,omitempty"`

```
IngressGateways      []*GatewaySpec     `protobuf:"bytes,40,rep,name=ingress_gateways,json=ingressGateways,proto3" json:"ingress_gateways,omitempty"
```
Then we have to write the IOP yaml like this:
```
components:
    ingress_gateways:
      - namespace: istio-system
```

ostromart:
@morvencao are you saying the purpose of the PR is to be able to unmarshal using regular yaml/json unmarshalers (rather than jsonpb?). i think we are using jsonpb consistently throughout the codebase so is this a requirement or just to try to eliminate use of jsonpb?
i'm concerned about this change because in general we want to avoid messing around with the generated files unless we absolutely have to.
feel free to ping my during my evening to discuss further.

morvencao:
We are using both the yaml/json unmarshalers and jsonpb unmarshaler.
For all the commands integrated with istioctl, we use jsonpb unmarshaler, while for controller, we need to retrieve and store the IOP instance from kube-apiserver with go-client lib, it will use the yaml/json unmarshalers bedefault, see: https://github.com/istio/istio/blob/master/operator/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go#L120
```
if err := r.client.Get(context.TODO(), reqNamespacedName, iop); err != nil {
...
}
```

ostromart:
@morvencao we might as well in that case change everything to regular yaml unmarshal. it's dangerous to use two different unmarshal paths because we will lose test coverage. i've created https://github.com/istio/istio/issues/21039 to track.

ostromart:
@morvencao this fix is needed in release-1.5, could you please cherrypick there?

morvencao:
@ostromart https://github.com/istio/api/pull/1301

PR #1278, opened 2020-02-07
annotation for using TCP loopback for Envoy admin endpoint

yxue:
https://github.com/istio/istio/issues/19684

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -315,6 +315,16 @@ var (
 		  Resources: []ResourceTypes{ Pod, },
         }
 	
+		SidecarHttpAdminEndpoint = Instance {
+          Name: "sidecar.istio.io/httpAdminEndpoint",
+          Description: "Specifies whether or not using TCP loopback address as "+
+                        "Envoy admin endpoint. UDS path will be used if it's set "+
+                        "to false.",
+          Hidden: false,
+          Deprecated: false,
+		  Resources: []ResourceTypes{ Pod, },
+        }
+	
 		SidecarInject = Instance {
           Name: "sidecar.istio.io/inject",
           Description: "Specifies whether or not an Envoy sidecar should be "+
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -188,6 +188,13 @@ annotations:
     hidden: false
     resources:
       - Pod
+  - name: sidecar.istio.io/httpAdminEndpoint
```
ramaraochavali: IMO `httpAdminEndpoint` is little confusing and not providing enough context to figure out what this annotation will do. How about `adminEndpointOnUDS` or `secureAdminEndpoint` ?

yxue: `adminEndpointOnUDS` SGTM

howardjohn:
I think this should be in mesh config

annotation/annotations.gen.go
```diff
@@ -269,6 +269,16 @@ var (
 		  Resources: []ResourceTypes{ Pod, },
         }
 	
+		SidecarAdminEndpointOnUDS = Instance {
+          Name: "sidecar.istio.io/adminEndpointOnUDS",
+          Description: "Specifies whether or not using TCP loopback address as "+
+                        "Envoy admin endpoint. Unix domain socket will be used if "+
```
rlenglet: I think this is the wrong logic. Envoy should be configured to use UDS unconditionally. This annotation should only enable a TCP proxy to it in pilot-agent.

yxue: The intention is to make the UDS as an option and default to TCP loopback.

duderino:
@yxue can you add an item to discuss this in a networking control plane wg meeting?  

This one is important but not urgent so you can wait until it works for your timezone ;)

yxue:
> @yxue can you add an item to discuss this in a networking control plane wg meeting?
> 
> This one is important but not urgent so you can wait until it works for your timezone ;)

Sure. I am going to add the item to the WG meeting once I am back to the GMT-8 zone.

PR #1279, opened 2020-02-07
Sidecar Outbound Traffic Policy allow_any mode tweaks

rshriram:
Allow users to specify a proxy through which allow_any traffic should be sent.
People can use forward proxies like Squid, Envoy's dynamic forward proxy, etc.


networking/v1alpha3/sidecar.proto
```diff
@@ -566,8 +566,32 @@ message OutboundTrafficPolicy {
...
+    // 15443 if omitted.
+    uint32 port = 2;
+  }
+
+  // Specifies the details of the proxy through which unknown traffic
+  // should be sent. This field is applicable only when the mode is
+  // set to ALLOW_ANY_VIA_PROXY. If omitted, defaults to
+  // istio-egressgateway.istio-system.svc.cluster.local:15443 when
+  // applicable.
+  Proxy proxy = 2;
```
hzxuzhonghu: Isn't there a http proxy setting in istio? 
I donot think there should be a default value here. Egress gateway only can be used by specifying explicitly.

rshriram: Yes we have a proxy port. It only allows the sidecar to setup a proxy port that the application can use (127.0.0.1:<port>). This one is for unknown traffic in ALLOW_ANY mode. With regard to the default, this setting will be used ONLY if the mode is set to ALLOW_ANY_VIA_PROXY. If the mode is anything else, this setting will make no difference.

linsun:
@rshriram This seems reasonable, will let others to have time to review too.   I have a few questions:

1. allow_any_via_proxy seems to assume the proxy will always be running in k8s.  Correct?
2. I can also configure egress gateway via allow_any_via_proxy if I use egress gateway host name as the host.  How is this different than using registry_only?

networking/v1alpha3/sidecar.proto
```diff
@@ -714,8 +714,32 @@ message OutboundTrafficPolicy {
     // Outbound traffic to unknown destinations will be allowed, in case
     // there are no services or `ServiceEntry` configurations for the destination port.
     ALLOW_ANY = 1;
+
+    // Outbound traffic to unknown destinations will be captured and
+    // forwarded to the specified proxy. Applicable only for HTTP and
+    // TLS traffic.
+    ALLOW_ANY_VIA_PROXY = 2;
```
howardjohn: I would expect that we just keep two modes, registry or allow, and then if proxy is non nil we forward through the proxy. is there any reason to add another option here?

networking/v1alpha3/sidecar.proto
```diff
@@ -714,8 +714,32 @@ message OutboundTrafficPolicy {
...
   }
   Mode mode = 1;
+
+  message Proxy {
```
howardjohn: is this just in sidecar? Is this message duplicated in mesh config?

networking/v1alpha3/sidecar.proto
```diff
@@ -714,8 +714,32 @@ message OutboundTrafficPolicy {
...
+    string host = 1;
+    // The port to which the traffic should be forwarded to. Defaults to
+    // 15443 if omitted.
+    uint32 port = 2;
+  }
+
+  // Specifies the details of the proxy through which unknown traffic
+  // should be sent. This field is applicable only when the mode is
+  // set to ALLOW_ANY_VIA_PROXY. If omitted, defaults to
+  // istio-egressgateway.istio-system.svc.cluster.local:15443 when
```
howardjohn: I suggest we do not do this. This name, as well as istio-system, are coincidental and should not be a part of our core API

networking/v1alpha3/sidecar.proto
rshriram: hmm. So that is a good point. I thought, this would be "explicit".

networking/v1alpha3/sidecar.proto
rshriram: Just in sidecar. MeshConfig has its own OutboundTrafficPolicy :(. I am not sure if we should be duplicating it. People can set the global sidecar config with the OutboundtrafficPolicy for the same effect as the mesh config

networking/v1alpha3/sidecar.proto
rshriram: Fair enough. We can remove the defaults and let people explicitly configure it.

networking/v1alpha3/sidecar.proto
howardjohn: I thought the same at first, but providing the `proxy` field is really explicit to me. Otherwise we are just going to add validation to enforce that proxy field is present iff VIA_PROXY is selected.

networking/v1alpha3/sidecar.proto
howardjohn: Its not exactly the same though, right? If I define a global sidecar, then defined a sidecar like
```
egress:
  - foo
... # no outbound traffic policy
```

for a specific namespace, that namespace would not have the outbound traffic policy of the sidecar, it would use the meshconfig default? Basically there is no merging

rshriram:
@linsun I have updated the API. There is no built in assumption about whether the proxy is running inside the mesh or not. Users can specify a valid hostname outside of the mesh and we should be able to forward the traffic to the proxy.

> I can also configure egress gateway via allow_any_via_proxy if I use egress gateway host name as the host. How is this different than using registry_only?

not sure I understand. When you use `registry_only` plus a bunch of egress gateway configs (gateway, virtual service, etc.), you have to `explicitly` specify each and every destination host that needs to pass through the egress gateway. So, in situations where the user does not know all the external accesses involved, she can use the egress proxy proposed here. She can simply forward ANY unknown traffic (https/http) to the egress proxy that is setup as a dynamic forward proxy (Squid/Envoy/nginx etc.).

networking/v1alpha3/sidecar.proto
rshriram: Actually there is :). This is the foundation that I laid in the previous PR stating that the outbound traffic policy will inherit from the parent if present. It needs to be implemented, but very straightforward.

networking/v1beta1/sidecar.proto
```diff
@@ -517,6 +518,34 @@ message OutboundTrafficPolicy {
...
+  message EgressProxy {
+    // REQUIRED: The hostname of the proxy to which the unknown
+    // traffic should be forwarded to. If the hostname is in the
+    // namespace/fqdn format, it will be assumed to be a service in
+    // the mesh and Envoy will load balance the traffic to one of the
+    // endpoints of the service. If the hostname is simply a FQDN
+    // corresponding to a service outside the mesh, Envoy will resolve
+    // the host using DNS and forward the traffic to the IP address
+    // returned by the DNS.
+    string host = 1;
```
hzxuzhonghu: what is the protocol of the proxy? ss, HTTP, TLS. etc?

rshriram: There is a TLS setting that tells the sidecar how to encrypt the traffic to the proxy. App simply calls http://google.com.. Sidecar will consider this as allow_any and forwards it to the egress gateway. Similary, app calls https://google.com. Sidecar will look at the SNI, it wont match anything, and it will end up being forwarded to the egress gateway. In the TLS case, the egress gateway use the SNI to resolve the DNS and then forward the traffic to that host. We can accomplish all of this using Envoy's dynamic_forward proxy

PR #1282, opened 2020-02-11
Use gogo protobuf for operator

jackkleeman:
Fixes https://github.com/istio/istio/issues/20986


jackkleeman:
Hmm, this is trickier than I'd hoped

PR #1283, opened 2020-02-11
Use golang protobuf in operator json generated code

jackkleeman:
The json generation stuff is gogo, but the operator proto is golang. This is causing issues.
Fixes https://github.com/istio/istio/issues/20986


Makefile.core.mk
```diff
@@ -217,6 +217,7 @@ $(operator_v1alpha1_pb_gos) $(operator_v1alpha1_pb_doc) $(operator_v1alpha1_pb_p
 	@cp -r /tmp/istio.io/api/operator/* operator
 	@sed -i -E '/MarshalJSON is a custom marshaler for TypeMapStringInterface2/,+10d' $(operator_v1alpha1_path)/operator_json.gen.go
 	@sed -i -E '/MarshalJSON is a custom marshaler for TypeInterface2/,+10d' $(operator_v1alpha1_path)/operator_json.gen.go
+	@sed -i 's/gogo/golang/g' $(operator_v1alpha1_path)/operator_json.gen.go
```
howardjohn: this feels like the wrong way to do this but I'm not a proto expert

jackkleeman:
The operator protos are in golang, and changing this seems not super easy due to GOTYPE

Makefile.core.mk
jackkleeman: https://github.com/istio/tools/tree/master/cmd/protoc-gen-jsonshim produces gogo:
https://github.com/istio/tools/blob/master/cmd/protoc-gen-jsonshim/jsonshim/plugin.go#L63
So theres no other way :(

jackkleeman:
It would be preferable to move operator to gogo, but I think that will require some surgery. I don't really understand the GOTYPE stuff we are doing and why its needed

PR #1286, opened 2020-02-12
Sync v1alpha3 and v1beta1 APIs

jasonwzm:
Having a test/mechanism that makes sure the networking v1alpha3 and v1beta1 APIs are in sync can be a little tricky. Plus we also need to make sure developers are not making backward incompatible changes.

I am leaning toward a nagger bot for now to remind the developer to change in both versions, as we will deprecate v1alpha3 in a near future version and this situation may no longer exists. What do reviewers think?


jasonwzm:
> Nag bot seems like high effort for minimal gain. Can we just validate the CRDs? `kubectl apply --dry-run` or something?

That can be part of the check of the submit to make sure we don't break the install with each API change. But we also need to make sure two versions are schematically identical, which kubernetes does not validate.

PR #1290, opened 2020-02-14
Deprecate authentication alpha policy

diemtvu:
This is replaced by security/v1beta1/peer_authentication.proto


diemtvu:
@louiscryan , can I have another stamp to have this merged? Tkx

PR #1295, opened 2020-02-20
Fix front matter generation in some API docs

jasonwzm:
fixes: https://github.com/istio/istio/issues/21152

For per-package generation, protoc needs a package comment to determine the package comment to include. For per-file generation, comment placement is important to get the front matter.

/cc @richardwxn


jasonwzm:
htmlgenerator relies on protoc to retrieve those fields. It is a requirement of protoc instead of htmlgenerator itself. I am not sure if this is a good candidate for linter rules, as a developer can put comments anywhere in a proto file and that mean something to them. I think we can add a tip for developers who make changes to protos about this.

PR #1296, opened 2020-02-24
Add status port to proxy config

howardjohn:
For https://github.com/istio/istio/issues/21222

This is part of the effort to move configuration out of flags/env vars
into mesh config for agent (and other fields). These fields are
currently only passed via helm, I am moving them to mesh config in the
same form they were in.


mesh/v1alpha1/proxy.proto
```diff
@@ -229,6 +229,18 @@ message ProxyConfig {
...
+  STS sts = 25;
+
+  // Port on which the agent should listen for administrative commands such as readiness probe.
+  int32 status_port = 26;
+}
+
+message STS {
+  // The service port used by Security Token Service (STS) server to handle token exchange requests.
+  // Setting this port to a non-zero value enables STS server.
+  string service_port = 1;
```
hzxuzhonghu: set type int32 and call it port

costinm:
- please don't create a sub-message for a single field.

- I don't get the comment on sts. Is it the URL of the server ? We have a
pattern we use for tracing and other remote services we use, we discovered
we need more than url. And who is listening ?

howardjohn:
> - please don't create a sub-message for a single field.

If we don't then we run the risk of having a bunch of flat messages like sts_port, sts_foo, sts_bar if we want to extend it. Is that something we want, or do we assume there will be no more sts related settings?

I don't mind either way

> - I don't get the comment on sts. Is it the URL of the server ? We have a
> pattern we use for tracing and other remote services we use, we discovered
> we need more than url. And who is listening ?

pilot-agent is listening on localhost so I don't think we need tls settings, etc. Its some proxying type thing where envoy -> agent -> real STS server. The real STS server part is determined by the "plugin" settings on the agent I believe

costinm:
On Wed, Feb 26, 2020, 08:08 John Howard <notifications@github.com> wrote:

>
>    - please don't create a sub-message for a single field.
>
> If we don't then we run the risk of having a bunch of flat messages like
> sts_port, sts_foo, sts_bar if we want to extend it. Is that something we
> want, or do we assume there will be no more sts related settings?
>

Hopefully not - we can have an Authentication message and move all the
other fields (sds, etc ) - but it's better to just remove them.



I don't mind either way
>
>
>    - I don't get the comment on sts. Is it the URL of the server ? We
>    have a
>    pattern we use for tracing and other remote services we use, we
>    discovered
>    we need more than url. And who is listening ?
>
> pilot-agent is listening on localhost so I don't think we need tls
> settings, etc. Its some proxying type thing where envoy -> agent -> real
> STS server. The real STS server part is determined by the "plugin" settings
> on the agent I believe
>
Then why is it even a setting ? Why would a user set it to a different
value and API ?

Not sure what the plugin section is, maybe we should discuss it in a WG.
The problem is that people just add random stuff to values.yaml and helm,
with no usability or operation or testing consideration - so we ended up
with 100s of install options that don't really work.

If an operator needs to configure STS - we should treat it as an API and
not leak implementation details into the API.

howardjohn:
Well we need some way to tell the agent to run STS or to not run STS server. I don't really care if the port is configurable, we can make this `bool sts_enabled` if preferred. But we do need some way to do it, and currently the preferred method to configure the proxy is through mesh config?

costinm:
To take a step back - why does the proxy need to run the STS server ? I
assume it is only if:
- external CA is configured ( not istiod )
- external XDS server is configured ( not Istiod )
- a remote STS server is used to to connect to the external CA or to  the
XDS server

In either case, the remote STS server URL will need to be configured - and
that should be enough to enable
or not the sts. If there is no STS server URL - sts_enabled=true will be
broken.

I would also guess we need to know the authentication mode for the CA
and/or XDS servers - to know
if we need to call STS or just send the JWT tokens directly.

JimmyCYJ:
Here is the STS flow, it is not related to the CA or SDS.
proxy-->STS server ---->token service backend
|------------------------|
istio-proxy container


STS server has a Google token exchange plugin, which calls cloud GAIA for token. The URL is hard coded into the token exchange plugin. The request parameters in the calls are specific to Google cloud GAIA. If we want to let the plugin to call different backend, e.g. AWS, the request parameters could be quite different. 
We can add another token exchange plugin, and let that plugin call different backend.

costinm:
Why would the STS server be inside proxy ? If google supports STS - we
should call the standard STS backend.

If it is a google-specific feature - it is possible to have code to
auto-detect google environment and enable it
automatically. We do this already for the google CA - if the CA_ADDR
contains google.com we use the google-specific
code. We have similar plugins for other vendors.

But the implementation details of a vendor integration should not be
reflected into the mesh API, in particular if
there are simpler alternatives.

We do have a generic mechanism to configure external services - and if we
claim to support STS (as an open
standard ) - we should allow consistent configuration.

JimmyCYJ:
The STS server is an agent, and envoy talks to this STS server following [standard STS spec](https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-16). 
This STS server enables a google plugin if it is running in a google environment, and the plugin calls cloud GAIA for token. Cloud GAIA does not serve STS requests directly. 

Right now we add this because we need a token to connect Google services, but the STS part is a standard. We could extend this to get token from other places

The benefit of running STS server inside proxy is that we can let STS server to do a token prefetch, the prefetch requires the mounted trustworthy JWT.

When the STS port is non-zero, Pilot-agent enables STS server, and passes the port information to proxy metadata. When it is zero, we don't enable the STS server, and proxy will not call STS server for token.

There is no plan to add more flags in the short term, so it is okay if we use flag message sts_port.
I use sub-message in case that we want to add more flags to support more token services.

costinm:
On Tue, Mar 3, 2020 at 10:19 AM Jimmy Chen <notifications@github.com> wrote:

> The STS server is an agent, and envoy talks to this STS server following standard
> STS spec <https://tools.ietf.org/html/draft-ietf-oauth-token-exchange-16>.
>
This STS server enables a google plugin if it is running in a google
> environment, and the plugin calls cloud GAIA for token. Cloud GAIA does not
> serve STS requests directly.
>
Right now we add this because we need a token to connect Google services,
> but the STS part is a standard. We could extend this to get token from
> other places
>

If this can be used to get a token from any STS server - I have no problem
to treat it in the same way we treat other external services, like
telemetry and logging - and following the same pattern.

If it is google specific - either autodetect google env ( like we do for
managed CA ) and not pollute mesh config, or move this setting to
the google distribution.



> The benefit of running STS server inside proxy is that we can let STS
> server to do a token prefetch, the prefetch requires the mounted
> trustworthy JWT.
>

But that means envoy is not actually following the STS standard - if I
understand it correctly.

STS requires sending a token ( the mounted JWT ) to the STS server, and
getting back another token.

It seems this is complex and most complex designs are likely to have subtle
bugs - in particular around security.



> When the STS port is non-zero, Pilot-agent enables STS server, and passes
> the port information to proxy metadata. When it is zero, we don't enable
> the STS server, and proxy will not call STS server for token.
>

But that doesn't mean we need an API for this internal behavior.

You could as well detect you are in a google environment - for example by
checking the trust domain in the JWT token -
and automatically enable this.


> There is no plan to add more flags in the short term, so it is okay if we
> use flag message sts_port.
> I use sub-message in case that we want to add more flags to support more
> token services.
>
That's not a good enough reason to not follow the existing patterns, or to
add settings that are not generally needed.

For values.yaml - we have a huge mess because everyone adds settings left
and right with no good review - but
mesh config is treated as an API.

JimmyCYJ:
Enabling STS port by auto detecting a google env sounds good to me. I will modify the Pilot-agent code for master branch, so that we can get rid of this setting.

howardjohn:
I dropped STS from this PR

PR #1297, opened 2020-02-24
Change int or strings fields to interface type for yaml unmarshal

ostromart:
Unmarshaling with jsonpb leads to annoying error noise output to console. This PR modifies some field types so IstioOperator can be unmarshaled using regular yaml package. There's little loss in syntax constraint because int or string already allows a very broad range of values.


sdake:
> LGTM.
> 
> This needs TOC approval prior to the shipping of 1.5 if possible.

@ostromart has indicated this can wait for 1.6.

PR #1299, opened 2020-02-27
[release-1.5] Change int or strings fields to interface type for yaml unmarshal (#1297)

ostromart:
Manual cherry pick of https://github.com/istio/api/pull/1297.
This also merges back a fix https://github.com/istio/api/pull/1251 which looks to have been removed accidentally.

* Change int or strings fields to interface type for yaml unmarshal

* Update 1.5 proto lock


dgn:
In the original PR, @sdake said:
> LGTM.
> 
> This needs TOC approval prior to the shipping of 1.5 if possible.

Has anything changed about that?

ostromart:
I'm not sure what Steve meant by this, I assumed it was just the formality of approving the review. This is not a user visible API change.

sdake:
> In the original PR, @sdake said:
> 
> > LGTM.
> > This needs TOC approval prior to the shipping of 1.5 if possible.
> 
> Has anything changed about that?

The PR looks good to me but can wait until 1.6. This PR is tied to another PR that is on hold until 1.5 ships.

sdake:
@dgn here was the original "on hold" interpretation: https://github.com/istio/istio/pull/21434#issuecomment-590603310

However, after review of that PR, I really do believe that PR needs to merge in 1.5 and this is a dependent PR of #https://github.com/istio/istio/pull/21434. I am not sure where @ostromart stands on this issue.

ostromart:
I'm very much in favor of merging this. My comment in the other issue to wait for 1.6 is just for moving all the other unmarshals to regular yaml, rathet than just the one that causes the nasty error to pop up. 
To set some context, this fixes the problem where for every manifest command the user will see the confusing "proto: tag has too few fields: "-"" message, even on success.

operator/v1alpha1/component.proto
```diff
@@ -113,7 +113,7 @@ message KubernetesResourcesSpec {
     repeated EnvVar env = 2;
     // k8s HorizontalPodAutoscaler settings.
     // https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/
-    k8s.io.api.autoscaling.v2beta1.HorizontalPodAutoscalerSpec hpa_spec = 3;
+    HorizontalPodAutoscalerSpec hpa_spec = 3;
```
fpesce: @ostromart proto-linter is chocking on the fact you removed k8s.io.apiu.autoscaling without getting rid of the import.
Your PR should remove 
import "k8s.io/api/autoscaling/v2beta1/generated.proto";

in the header of this file.

ostromart: that must be it - good catch, thank you!

ostromart: indeed that was the problem.

dgn:
@sdake @ostromart thank you for the discussion, I just wanted to be safe

PR #1301, opened 2020-02-27
fix operator go struct tag.

morvencao:
cherrypick: https://github.com/istio/api/pull/1277


morvencao:
Should also fix: https://github.com/istio/istio/issues/21490

PR #1314, opened 2020-03-09
add annotation to redirect from destination 127.0.0.1 to local podIP

ChenLingPeng:
There are some services only listen on pod IP in k8s for some reason instead of 0.0.0.0 or 127.0.0.1.
After using istio, the inbound traffic will be intercepted to envoy and redirect to 127.0.0.1, istio assuming that user service should listen on lo address, then the traffic will fail if it is not the truth.

This change introduce an annotation `traffic.sidecar.istio.io/bindPodIPPorts`, istio should handle this annotation and redirect those port's traffic to podIP:port than then 127.0.0.1:port.

Signed-off-by: Lingpeng Chen <forrest0579@gmail.com>

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -426,6 +426,19 @@ var (
 		  Resources: []ResourceTypes{ Pod, },
         }
 	
+		SidecarTrafficBindPodIPPorts = Instance {
+          Name: "traffic.sidecar.istio.io/bindPodIPPorts",
+          Description: "A Comma separated list of inbound ports for which are "+
+                        "binding to pod IP only. By default, Envoy will redirect "+
+                        "inbound traffic to 127.0.0.1:${port}, if the port is only "+
+                        "bind to pod IP instead of 127.0.0.1, this annotation "+
+                        "should be addressed to redirect 127.0.0.1:${port} to "+
+                        "${podIP}:${port}.",
+          Hidden: false,
+          Deprecated: false,
+		  Resources: []ResourceTypes{ Pod, },
+        }
+	
 		SidecarTrafficExcludeInboundPorts = Instance {
           Name: "traffic.sidecar.istio.io/excludeInboundPorts",
           Description: "A comma separated list of inbound ports to be excluded "+
@@ -530,6 +543,7 @@ func AllResourceAnnotations() []*Instance {
 		&SidecarUserVolume,
 		&SidecarUserVolumeMount,
 		&SidecarStatusPort,
+		&SidecarTrafficBindPodIPPorts,
 		&SidecarTrafficExcludeInboundPorts,
 		&SidecarTrafficExcludeOutboundIPRanges,
 		&SidecarTrafficExcludeOutboundPorts,
--- a/annotation/annotations.pb.html
+++ b/annotation/annotations.pb.html
@@ -391,6 +391,16 @@ Istio supports to control its behavior.
 				
 					<tr>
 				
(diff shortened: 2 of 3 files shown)


annotation/annotations.gen.go
```diff
@@ -442,6 +442,19 @@ var (
 		  Resources: []ResourceTypes{ Pod, },
         }
 	
+		SidecarTrafficBindPodIPPorts = Instance {
+          Name: "traffic.sidecar.istio.io/bindPodIPPorts",
+          Description: "A Comma separated list of inbound ports for which are "+
+                        "binding to pod IP only. By default, Envoy will redirect "+
+                        "inbound traffic to 127.0.0.1:${port}, if the port is only "+
+                        "bind to pod IP instead of 127.0.0.1, this annotation "+
+                        "should be addressed to redirect 127.0.0.1:${port} to "+
```
howardjohn: We should either leave out the implementation details or finalize the implementation before merge (see comments on the other pr)

howardjohn:
In general though I don't have any issues with this

rshriram:
>After using istio, the inbound traffic will be intercepted to envoy and redirect to 127.0.0.1, istio assuming that user service should listen on lo address, then the traffic will fail if it is not the truth.

Remember that the application is expected to listen on 0.0.0.0 or 127.0.0.1. I dont see why this requires an annotation. You can easily use an EnvoyFilter to replace the InboundCluster with 127.0.0.1 to orig_dst cluster (which then routes to the pod IP).

ChenLingPeng:
@rshriram I think simply change the destination IP is not enough, the iptables rules may conflict with it.
see istio/istio#22117 for more details

hzxuzhonghu:
This is a good start to reduce the unfriendly requirements. I think it is good to go.

louiscryan:
Why is default_endpoint in the Sidecar ingress listener insufficient for this ?

hzxuzhonghu:
It can not, the default_endpoint can be only declared as `127.0.0.1:PORT` or `unix:///path/to/socket`

Also we can not predict the pod ip

ramaraochavali:
We could say `default_endpoint: %POD_IP%:port ` some thing like that like how we are processing `unix:` specifically, we could process POD_IP differently, if we want to reuse `default_endpoint`?

ChenLingPeng:
@ramaraochavali simply changing the destination ip from 127.0.0.1 to PODIP is not enough. the iptables rules will redirect this kind of traffic back to envoy which will cause an infinite loop.

ramaraochavali:
@ChenLingPeng Sure. I am just saying this as an alternative to adding a new annotation to @louiscryan 's question above.

rshriram:
As Rama suggested, you can allow :port in the ingress section. In implementation we can easily handle this. The annotation is essentially doing the same thing.

ChenLingPeng:
@rshriram from implementation aspect, defaultEndpoint can also handle this. But I think it is a little complicated for user experience. People should create an additional Sidecar crd instead of one line annotation. Right?

ChenLingPeng:
@hzxuzhonghu @howardjohn I would like to see more opinions about using annotation or  Sidecar.Ingress to implement this feature.

hzxuzhonghu:
@rshriram In reality, it is a per sidecar config.  As you can see from the implementation,  it also need to add a specific iptables rule handle this.  We can not get this using `SideCar` when initializing iptables. 
Another point this annotation is similar to `traffic.sidecar.istio.io/excludeInboundPorts`, etc.

PR #1316, opened 2020-03-09
[WIP] add workloadEntry into API

irisdingbj:
@smawson Please take a look at this change to see whether this is what we want. Thanks.

For
https://docs.google.com/document/d/1VDV4-prbMbzW06hslfp12sn7V6oUvnIAst6mT6FtSWc/edit#heading=h.xw1gqgyqs5b


hzxuzhonghu:
What is it diff from ServiceEntry? Any design link?

irisdingbj:
> What is it diff from ServiceEntry? Any design link?

https://docs.google.com/document/d/1VDV4-prbMbzW06hslfp12sn7V6oUvnIAst6mT6FtSWc/edit#heading=h.uakvommwzrq2

ServiceEntry can select workloadEntry.

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
+// Copyright 2018 Istio Authors
```
smawson: 2020

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+
+syntax = "proto3";
+
+import "google/api/field_behavior.proto";
+import "networking/v1alpha3/gateway.proto";
+
+
+// $schema: istio.networking.v1alpha3.WorkloadEntry
+// $title: Workload Entry
+// $description: Configuration affecting workload  entry.
```
smawson:
```suggestion
// $description: Configuration of a workload entry.
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+
+// `WorkloadEntry` enables adding additional workloads into Istio's internal
+// service registry, so that auto-discovered services in the mesh can
```
smawson:
```suggestion
// service registry, so that services in the mesh can
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+// access/route to these manually specified workload. A workload entry
```
smawson:
```suggestion
// route to these workloads. A workload entry
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+// describes the properties of a workload.
```
smawson:
```suggestion
// describes the properties of a workload such as its identity and endpoints.
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+//
+// The following example declares a  workloadEntry running on both 1.2.3.4 and 1.2.3.5.
```
smawson:
```suggestion
// The following example declares a workloadEntry running on two IP addresses: 1.2.3.4 and 1.2.3.5.
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+
+import "google/api/field_behavior.proto";
+import "networking/v1alpha3/gateway.proto";
+
+
+// $schema: istio.networking.v1alpha3.WorkloadEntry
+// $title: Workload Entry
+// $description: Configuration affecting workload  entry.
+
+// `WorkloadEntry` enables adding additional workloads into Istio's internal
```
smawson:
```suggestion
// `WorkloadEntry` enables adding additional workloads such as VMs into Istio's internal
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+// describes the properties of a workload.
+//
+// The following example declares a  workloadEntry running on both 1.2.3.4 and 1.2.3.5.
+// expose ports on 27018
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: WorkloadEntry
+// metadata:
+//   name: myvm
```
smawson: Should stick it in a namespace just as best practice?

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+//   - number: 27018
+//     name: mongodb
+//     protocol: MONGO
+//   endpoints:
+//     - addresses:
+//       - 1.2.3.4
+//       - 1.2.3.5
+// ```
+//
+// The following example declares a workloadEntry running on  1.2.3.4
```
smawson:
```suggestion
// The following example declares a WorkloadEntry running at 1.2.3.4
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+// and expose ports on 27018 and 8888
```
smawson:
```suggestion
// which exposes ports 27018 and 8888.
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+// $title: Workload Entry
+// $description: Configuration affecting workload  entry.
+
+// `WorkloadEntry` enables adding additional workloads into Istio's internal
+// service registry, so that auto-discovered services in the mesh can
+// access/route to these manually specified workload. A workload entry
+// describes the properties of a workload.
+//
+// The following example declares a  workloadEntry running on both 1.2.3.4 and 1.2.3.5.
+// expose ports on 27018
```
smawson:
```suggestion
// Both endpoints expose port 27018.
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+//
+
+package istio.networking.v1alpha3;
+
+option go_package = "istio.io/api/networking/v1alpha3";
+
+
+message WorkloadEntry {
+
+  string serviceAccountName = 1 [(google.api.field_behavior) = REQUIRED];
```
smawson: Add comment describing the meaning of this field.

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+  // One or more labels associated with the workloadEntry.
+  map<string, string> labels = 2;
```
smawson: Shouldn't this be part of metadata? I don't think we need additional labels.

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+  repeated Port ports = 3;
+
+  // One or more endpoints associated with the workloadEntry.
+  repeated Endpoint endpoints = 4 [(google.api.field_behavior) = REQUIRED];
+
+  };
+
+  message Endpoint {
+
+    repeated string addresses = 1 [(google.api.field_behavior) = REQUIRED];
```
smawson: Describe what types of addresses are supported.

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,102 @@
...
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: WorkloadEntry
+// metadata:
+//   name: myvm
+//   labels:
+//      app: myapp
+//      version: v1
+// spec:
```
howardjohn: Do we need the topology here as well? Otherwise we will have no way to extract this information.

networking/v1alpha3/workload_entry.proto
howardjohn: What types of address are supported? IP only or DNS as well? We will need to document how this should be used

howardjohn: In particular the difference between
```
endpoints:
- addresses: [a, b, c]
```
and 

```
endpoints:
- addresses: [a]
- addresses: [b]
- addresses: [c]
```

Which should be that the first is one instance with multiple addresses (dual stack, etc), the second being 3 different instances

networking/v1alpha3/workload_entry.proto
howardjohn: Do we want labels defined here, given its already in k8s object meta?

networking/v1alpha3/workload_entry.proto
irisdingbj: fixed

networking/v1alpha3/workload_entry.proto
irisdingbj: fixed

networking/v1alpha3/workload_entry.proto
irisdingbj: fixed

networking/v1alpha3/workload_entry.proto
irisdingbj: fixed

networking/v1alpha3/workload_entry.proto
irisdingbj: fixed

networking/v1alpha3/workload_entry.proto
irisdingbj: removed

irisdingbj: removed

networking/v1alpha3/workload_entry.proto
irisdingbj: added explanation we support both formats, and also we support both DNS and static IP

networking/v1alpha3/workload_entry.proto
irisdingbj: fixed

networking/v1alpha3/workload_entry.proto
irisdingbj: fixed

networking/v1alpha3/workload_entry.proto
irisdingbj: added  topoloty field like the one from k8s.endpoint

irisdingbj:
@smawson @howardjohn  Thanks for the review comments. PTAL.

networking/v1alpha3/workload_entry.proto
smawson: Wasn't changed, what do you mean by "auto-discovered" here?

networking/v1alpha3/workload_entry.proto
```diff
@@ -20,15 +20,15 @@ import "networking/v1alpha3/gateway.proto";
...
 // $schema: istio.networking.v1alpha3.WorkloadEntry
 // $title: Workload Entry
-// $description: Configuration affecting workload  entry.
+// $description: Configuration of a workload entry.
 
 // `WorkloadEntry` enables adding additional workloads into Istio's internal
 // service registry, so that auto-discovered services in the mesh can
-// access/route to these manually specified workload. A workload entry
-// describes the properties of a workload.
+// route to these workloads. A workload entry. A workload entry
```
smawson:
```suggestion
// route to these workloads. A workload entry
```

networking/v1alpha3/workload_entry.proto
smawson: Not fixed

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
 //       - 1.2.3.4
 // ```
 //
+// The following example declares a workloadEntry running with dual IP stack on 1.2.3.4 and 1.2.3.5
+// and expose ports on 27018 and 8888
```
smawson:
```suggestion
// which exposes ports 27018 and 8888.
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
...
+//   - number: 8888
+//     name: tcp-vm
+//     protocol: tcp
+//   endpoints:
+//     - addresses:
+//       - 1.2.3.4,1.2.3.5
+// ```
+//
+// The following example declares a workloadEntry running with DNS name httpbin.testcompany.com
+// and expose ports on 27018 and 8888
```
smawson:
```suggestion
// which exposes ports 27018 and 8888.
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
...
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
 
 message WorkloadEntry {
 
-  string serviceAccountName = 1 [(google.api.field_behavior) = REQUIRED];
-  // One or more labels associated with the workloadEntry.
-  map<string, string> labels = 2;
+  // Identity information for this workloadEntry
```
smawson: How about:

The service account that this WorkloadEntry is running as. Services calling this WorkloadEntry will verify that the identity matches the service account specified here.

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
...
+  string service_account_name = 1 [(google.api.field_behavior) = REQUIRED];
 
   // The ports associated with the workloadEntry.
```
smawson: We should describe what these are used for.

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
...
   repeated Port ports = 3;
 
-  // One or more endpoints associated with the workloadEntry.
+  // One or more endpoints associated with the service.
```
smawson: We should say something about how these are used too. These are the endpoints that traffic will be routed to for Services that select this WorkloadEntry.

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
...
   repeated Endpoint endpoints = 4 [(google.api.field_behavior) = REQUIRED];
 
   };
 
   message Endpoint {
 
+    // The endpoint addresses associated with the workloadEntry. Could be a DNS
+    // name as well as IP address
```
smawson: Mention that multiple addresses are possible for things like dual stack.

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
...
     repeated string addresses = 1 [(google.api.field_behavior) = REQUIRED];
 
+    // topology contains arbitrary topology information associated with the
```
smawson:
```suggestion
    // Topology information associated with the
```

networking/v1alpha3/workload_entry.proto
```diff
@@ -75,28 +75,112 @@ import "networking/v1alpha3/gateway.proto";
...
+    // endpoint. These key/value pairs must conform with the label format.
```
smawson: Rather than copy-pasting the K8s topology comment, maybe we can just point over at it? We can say this is identical to the topology information on an endpointslice that was derived from a Pod.

We can do something similar for the endpoints btw, point over to k8s endpointslice for additional details.

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,186 @@
...
+//   name: myvm
+//   labels:
+//      app: myapp
+//      version: v1
+// spec:
+//   serviceAccountName: vmtest
+//   ports:
+//   - number: 27018
+//     name: mongodb
+//     protocol: MONGO
```
rshriram: why does this have ports with protocols? How is this going to be used in Pilot?

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,186 @@
...
+//   - number: 8888
+//     name: tcp-vm
+//     protocol: tcp
+//   endpoints:
+//     - addresses:
+//       - 1.2.3.4,1.2.3.5
```
rshriram: Is this some comma separated string? What is a dual IP stack? If you want to declare one endpoint with multiple IPs, then this is not the place. Nothing in Istio is equipped to handle the dual IP in an endpoint (all the filter chains we setup in inbound, the iptables loop prevention, etc. will break down. There is nothing in place to prevent one IP from sending traffic to another and avoid traffic loop)

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,186 @@
...
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1alpha3
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+
+message WorkloadEntry {
+
+  // Identity information for this workloadEntry
+  string service_account_name = 1 [(google.api.field_behavior) = REQUIRED];
```
rshriram: Why is this required? What happens if the end user is not using mtls to this workload as its a legacy VM? And secondly, what is the significance of this service account as far as Istio is concerned? How does this interact with the service entry's SAN field?

networking/v1alpha3/workload_entry.proto
rshriram: we can't handle dual stack today. A whole lot of system is designed based on the assumption that there is only one IP (specifically config generated by Pilot)

networking/v1alpha3/workload_entry.proto
rshriram: btw, we have our own notion of locality. Let us please not create yet another topology alternative

louiscryan: What alignment can we achieve with

https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#topology

can we say that any of our labels are synonyms with those above ?

@howardjohn

irisdingbj:
@rshriram Thanks for taking care of this. Noticed you have made it work so I will close this.

howardjohn:
Thanks for getting this started Iris!

PR #1320, opened 2020-03-11
Fix MeshConfig field tags for JSON unmarshal

ostromart:
Fixes https://github.com/istio/istio/issues/21794. Regular yaml unmarshal doesn't work with MeshConfig because json struct tags are rendered as snake_case by protoc.

### DIFF
--- a/mesh/v1alpha1/config.pb.go
+++ b/mesh/v1alpha1/config.pb.go
@@ -242,7 +242,7 @@ type MeshConfig struct {
 	// Port on which Envoy should listen for HTTP PROXY requests if set.
 	ProxyHttpPort int32 `protobuf:"varint,5,opt,name=proxy_http_port,json=proxyHttpPort,proto3" json:"proxy_http_port,omitempty"`
 	// Connection timeout used by Envoy. (MUST BE >=1ms)
-	ConnectTimeout *types.Duration `protobuf:"bytes,6,opt,name=connect_timeout,json=connectTimeout,proto3" json:"connect_timeout,omitempty"`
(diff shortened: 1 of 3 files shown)


mesh/v1alpha1/config.pb.go
```diff
@@ -461,7 +461,7 @@ type MeshConfig struct {
 	// Configure the provision of certificates.
 	Certificates []*Certificate `protobuf:"bytes,47,rep,name=certificates,proto3" json:"certificates,omitempty"`
 	// Set configuration for Thrift protocol
-	ThriftConfig         *MeshConfig_ThriftConfig `protobuf:"bytes,49,opt,name=thrift_config,json=thriftConfig,proto3" json:"thrift_config,omitempty"`
+	ThriftConfig         *MeshConfig_ThriftConfig `protobuf:"bytes,49,opt,name=thrift_config,json=thriftConfig,proto3" json:"thriftConfig,omitempty"`
```
howardjohn: What is the impact on pilot? we have camelCase in mesh config so I am a bit confused, I would have expected this to be snake case here. Is it because we use jsonpb?

ostromart: Actually it's because we don't use jsonpb in the controller - due to using k8s controller framework which doesn't use jsonpb internally. We hacked around this in the past but k8s as a whole doesn't treat proto as a first class citizen so their code is not fully compatible. 
Pilot will continue to work because the rest of Istio uses jsonpb, which translates the snake_case used in the api server to camelCase on the client.

PR #1324, opened 2020-03-12
Fix bug with JSON tag camelCase conversion

ostromart:
https://github.com/istio/api/pull/1301 introduced a bug where some fields are not converted from snake to camel case because the regex expected at most a single underscore - so fields like my_long_named_field would not match it.

### DIFF
--- a/operator/fixup_structs/main.go
+++ b/operator/fixup_structs/main.go
@@ -71,7 +71,7 @@ var (
 		goFieldToken: "",
 	}
 
-	regexJSONTag = regexp.MustCompile(`json\:"[a-z|0-9|_]+,`)
+	regexJSONTag        = regexp.MustCompile(`json\:"[a-z|0-9|_]+,`)
 	regexJSONTagIllegal = regexp.MustCompile(`json\:"_`)
 )
 
--- a/operator/fixup_structs/main.go
+++ b/operator/fixup_structs/main.go
@@ -70,6 +70,9 @@ var (
 		"github.com/gogo/protobuf/protobuf/google/protobuf": "github.com/gogo/protobuf/types",
 		goFieldToken: "",
 	}
+
+	regexJSONTag = regexp.MustCompile(`json\:"[a-z|0-9|_]+,`)
+	regexJSONTagIllegal = regexp.MustCompile(`json\:"_`)
 )
 
 //nolint: gocognit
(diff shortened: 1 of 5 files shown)


operator/fixup_structs/main.go
```diff
@@ -126,7 +126,7 @@ func main() {
 			tmp = append(tmp, lines[i+1])
 			i += 2
 		case strings.Contains(l, goFieldJSONTagPrefix) && strings.Contains(l, "_"):
-			rgx := regexp.MustCompile(`json\:"[a-z|0-9]+_[a-z|0-9]+,`)
+			rgx := regexp.MustCompile(`json\:"[a-z|0-9|_]+,`)
```
l8huang: I also encountered this issue. The old one doesn't match string begins with `_`, how about:
```suggestion
			rgx := regexp.MustCompile(`json\:"[a-z|0-9]+(_[a-z|0-9]+)+,`)
```

ostromart: I'm not sure I follow your suggestion. Could you provide an example of where this would work and the one in the PR would fail?

l8huang: The old one's problem is it can't match strings with more than one "_", e.g "xxx_yyy_zzz", it should be fixed to work with them; the one in this PR matches "xxx_yyy_zzz", but it also matches strings begins with "_", e.g. "_xxx". 

And the old one doesn't match strings without "_" in it, e.g "xxxx"; but the one in the PR matches strings without "_", that's, the replacing function is always called.

Currently I guess there are no fields begin with "_", so the one in the PR doesn't cause any issue, but it's not necessary to extend the matching space.

ostromart: Generally fields starting with _ are not allowed, at least these are caught by the linter since the spec is not explicit about this. But to be extra safe I've put in a check that aborts the tool if such a tag name is found.

operator/fixup_structs/main.go
```diff
@@ -126,8 +129,11 @@ func main() {
 			tmp = append(tmp, lines[i+1])
 			i += 2
 		case strings.Contains(l, goFieldJSONTagPrefix) && strings.Contains(l, "_"):
-			rgx := regexp.MustCompile(`json\:"[a-z|0-9|_]+,`)
-			l = rgx.ReplaceAllStringFunc(l, func(m string) string {
+			if regexJSONTagIllegal.MatchString(l) {
+				fmt.Printf("Illegal leading _ in json field tag: %s\n", l)
+				os.Exit(1) //nolint: gomnd
+			}
+			l = regexJSONTag.ReplaceAllStringFunc(l, func(m string) string {
```
l8huang: It's good to have a check for leading `_`, it makes the naming rule more clear than before.

When json tag doesn't include `_`, e.g. 'xxx', the regex still matches it and replacing function is called, is it better to avoid calling a function when it doesn't really do anything? ;)

ostromart: In this case I'd argue it's more important to avoid more complex regex, this is not a performance critical path.

PR #1329, opened 2020-03-14
Add Gateway Class annotation

rshriram:
As of now, we are not doing any special processing in Pilot based
on this annotation but we will soon set some special settings for
gateways depending on their class (ingress/egress/midtier).

For now, I opted to set this as a service level annotation rather
than pod level annotation. I dont mind switching. But either way, this
is not an API field as we have multiple gateway objects applying to 
the same workload. It doesn't make sense for one object to be ingress and
the other one to be egress.

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -11,7 +11,6 @@ const (
     Ingress
     Pod
     Service
-    ServiceEntry
 )
 
 func (r ResourceTypes) String() string {
@@ -24,8 +23,6 @@ func (r ResourceTypes) String() string {
 		return "Pod"
 	case 4:
 		return "Service"
-	case 5:
-		return "ServiceEntry"
 	}
 	return "Unknown"
 }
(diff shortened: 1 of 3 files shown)


hzxuzhonghu:
Just FYI, this is how k8s `Ingress` declares ingress class.

```
// IngressSpec describes the Ingress the user wishes to exist.
type IngressSpec struct {
	// IngressClassName is the name of the IngressClass cluster resource. The
	// associated IngressClass defines which controller will implement the
	// resource. This replaces the deprecated `kubernetes.io/ingress.class`
	// annotation. For backwards compatibility, when that annotation is set, it
	// must be given precedence over this field. The controller may emit a
	// warning if the field and annotation have different values.
	// Implementations of this API should ignore Ingresses without a class
	// specified. An IngressClass resource may be marked as default, which can
	// be used to set a default value for this field. For more information,
	// refer to the IngressClass documentation.
	// +optional
	IngressClassName *string

```

rshriram:
Our notion of class is ingress/egress/mid-tier. I dont see this classification being applicable at the API level. It seems strange to say that a single workload will serve as both ingress gateway for some set of hosts, egress gateway for another set and a mid-tier gateway for the rest. This may be theoretically possible but for the sake of simplicity and avoiding the cognitive overload on the user's part, I chose to go with the service-wide annotation so that we can automatically set things like XFF for the entire gateway workload, or envoy's dynamic forward proxy for the entire egress gateway.

PR #1331, opened 2020-03-14
VMs: Separate endpoints from service entry into WorkloadEntry

rshriram:
Alternative to https://github.com/istio/api/pull/1316

cc @smawson


networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,181 @@
...
+// +kubetype-gen:groupVersion=networking.istio.io/v1alpha3
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message WorkloadEntry {
+  // The service account associated with the workloads if sidecars are present in the workload.
+  string service_account = 1;
+
+  // A set of endpoints describing properties of the workloads. 
+  repeated ServiceEntry.Endpoint endpoints = 2 [(google.api.field_behavior) = REQUIRED];
```
hzxuzhonghu: From the first glance at the SE api, i thought WorkloadEntry should be with labels and just one instance. So that it works as representing a pod.

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,181 @@
+// Copyright 2018 Istio Authors
```
hzxuzhonghu: 2020

networking/v1alpha3/service_entry.proto
```diff
@@ -826,9 +826,19 @@ message ServiceEntry {
...
   repeated Endpoint endpoints = 6;
 
+  // Applicable only for MESH_INTERNAL services. If VM workloads are
+  // defined separately using the WorkloadEntry resource, use the
+  // workloadSelector to select the set of workloads that belong to
+  // the service described by this ServiceEntry. Only one of
+  // `endpoints` or `workloadSelector` can be specified. The
+  // WorkloadEntry object representing the VMs should be defined in
+  // the same namespace as the ServiceEntry.
+  WorkloadSelector endpoint_selector = 9;
```
howardjohn: do we want to name this `selector` or `endpointSelector`? or `workloadSelector` matches `Sidecar` I guess. The comment is misleading, it implies the field name is `workloadSelector` but its actually `endpointSelector` right?

networking/v1alpha3/service_entry.proto
```diff
@@ -826,9 +826,19 @@ message ServiceEntry {
     uint32 weight = 6;
   };
 
-  // One or more endpoints associated with the service.
+  // One or more endpoints associated with the service. Only one of
+  // `endpoints` or `workloadSelector` can be specified.
```
howardjohn: This is an arbitrary restriction. I am fine with it, just wondering if there was some context for why you made this decision?

networking/v1alpha3/service_entry.proto
howardjohn: So this will select only WorkloadEntries, not k8s endpoints as well?

networking/v1alpha3/sidecar.proto
```diff
@@ -663,16 +663,18 @@ message IstioEgressListener {
...
 
-// `WorkloadSelector` specifies the criteria used to determine if the `Gateway`,
-// `Sidecar`, or `EnvoyFilter` configuration can be applied to a proxy. The matching criteria
-// includes the metadata associated with a proxy, workload instance info such as
-// labels attached to the pod/VM, or any other info that the proxy provides
-// to Istio during the initial handshake. If multiple conditions are
-// specified, all conditions need to match in order for the workload instance to be
-// selected. Currently, only label based selection mechanism is supported.
+// `WorkloadSelector` specifies the criteria used to determine if the
+// `Gateway`, `Sidecar`, or `EnvoyFilter` or `ServiceEntry`
```
howardjohn: nit: maybe we should just stop having a full list here

networking/v1alpha3/workload_entry.proto
smawson: Oh I like this, interesting. Having labels on the endpoints makes selection slightly harder (have to scan all endpoints rather than just selecting matching based on metadata), but otherwise this is an improvement.

We should discuss whether we move Endpoint out of ServiceEntry at some point to make it a generic concept, but that's just some proto refactoring.

We should also discuss adding topology information to our endpoints to support topology (per nate's proposal), but that can come in a followup.

howardjohn:
We do not use the ports from pods at all anymore

smawson:
What do we use the pods on ServiceEntry.Endpoint for? Just curious as this point since it seems like if we don't need the Pod ports we don't need WorkloadEntry ports. Can always add later if we discover it is a problem so I'd rather leave things out and add later (topology is in the same camp I think).

howardjohn:
From the code, it looks like it allows overriding the ServiceEntry port? https://github.com/istio/istio/blob/80ca197e1c51b477a234300829531608a74e069c/pilot/pkg/serviceregistry/external/conversion.go#L138 But the documentation contradicts this and it doesn't seem particularly useful

networking/v1alpha3/service_entry.proto
rshriram: With regard to ServiceEntry enhancing the details of an existing k8s service, I would like to keep this aspect outside of this PR. It is, IMO a bug fix in Pilot that would enable us to merge the richer spec in the service entry with the k8s spec for the same service [have to work out a few issues there].

I went through the combos you specified when coming up with this. few problems surfaced:
1. If services select workload entries in addition to k8s pods, then we will have a series of problems in Pilot with regard to setting up EDS, Clusters, etc.
    1. If I create workload entries with DNS names for a headless service (that is usually orig dst cluster), what happens? [common solution: add a rider to say that this is not allowed and do runtime enforcement]
    2. If I create workload entries with DNS names for a non headless k8s service, we will have to do some complicated dance to dynamically create the Envoy aggregate cluster for this service (we already determined that there are a bunch of technical issues to doing that, due to API inconsistencies along the lines of Destination Rule subsets, tls settings, etc.) [solution: disallow this and only allow IP endpoint based workloadEntry for k8s services]
    3. Biggest problem: As someone introducing a bunch of VM workloads into the system without any knowledge of other k8s services, we now have to ensure that none of the labels in the VM workloads "accidentally" match the selectors of any k8s services in the namespace, unintentionally. If this were to happen, a lot of users will be caught by surprise in production. 
2. If a service entry starts selecting pods in addition to workload entries, and the service entry has a completely different hostname for example, we will have similar problems (DNS and endpoints). 

All said, I am not completely against this hybrid model. It would be good to do this in baby steps once we get this piece sorted out properly in Pilot.

networking/v1alpha3/service_entry.proto
rshriram: For clarity. I see no convenience reason to allow this kind of inline and out of line endpoint sets and then us having to deal with the code complexities of both models, not to mention having to explain this to end users. If we want this to look like K8S service, then it would be good to adhere to the way it looks [services have selectors like k8s or inline endpoints].

networking/v1alpha3/service_entry.proto
rshriram: I remembered more. If I want the facility for a k8s service to refer to an arbitrary set of Endpoints from VMs, why wouldn't I use an EndpointSlice with a k8s service? What is the rationale/additional functionality that WorkloadEntry offers over EndpointSlice? When I thought about this, it started to feel like repeating the same mistake as the umpteen set of multicluster options (single/multi-network, single/multi control plane, shared/no shared CA, auto import all or explicit import/export). The k8s community picking up the dumbest and most straightforward approach seemed to indicate (to me) that it is preferable to provide *one* clear way of doing X rather than multiple ways of doing the same thing.

networking/v1alpha3/workload_entry.proto
rshriram: We have had topology in the endpoints since 1.0 (called "locality" which is a region/zone/sub-zone hierarchy). Also, the label selection may be harder but by now, we have a very well oiled machinery in pilot to efficiently select a set of endpoints based on a service's label selectors.

rshriram:
With regard to the ports, the model followed in ServiceEntry is that if you omit ports in the endpoint, then the endpoint is assumed to have the same ports as the Service. If you have custom target ports in the endpoint, then you are supposed to provide a mapping like `servicePortName: endpointTargetPort` [e.g., `http1: 8080` where http1 is a port on the main service entry object]. From what we have seen, most people dont use the ports field in the endpoints.

networking/v1alpha3/sidecar.proto
vikaschoudhary16: s/or EnvoyFilter or/, Envoyfliter or/

networking/v1alpha3/service_entry.proto
vikaschoudhary16: @smawson 
> ServiceEntry can describe a lot more about a service than K8s Service, and can select both Pods and WorkloadEntry.

What is the use case for ServiceEntry selecting pods?

vikaschoudhary16: > What is the rationale/additional functionality that WorkloadEntry offers over EndpointSlice?
WorkloadEntry would be used to register the VM in the mesh. Then a controller can create EndpointSlices automatically using WorkloadEntry and Services/ServiceEntries.

rshriram: Yeah but why have another API? Why not directly register endpoint slice

louiscryan: One note about accidental selection into K8S services. An EndpointSlice generated by the K8S controller has a back-pointer to the Service from which it was generated.

This can be considered to act like a constraint in the reverse direction. If we had a controller that generated EndpointSlices by evaluating over workload entries we could similarly annotate the service / serviceentry it was composed from.

networking/v1alpha3/service_entry.proto
```diff
@@ -850,10 +860,12 @@ message ServiceEntry {
   // "." or "*" (i.e., the current namespace or all namespaces).
   repeated string export_to = 7;
 
-  // The list of subject alternate names allowed for workload instances that
-  // implement this service. This information is used to enforce
-  // [secure-naming](https://istio.io/docs/concepts/security/#secure-naming).
-  // If specified, the proxy will verify that the server
-  // certificate's subject alternate name matches one of the specified values.
+  // If specified, the proxy will verify that the server certificate's
```
louiscryan: @howardjohn 

what is the current state in the K8S service improvement spec to carry service account identities around on endpoint slice and/or make it a matching criteria in Service selection

networking/v1alpha3/workload_entry.proto
louiscryan: Its worth noting that this change makes it even more similar to EndpointSlice in structure.

networking/v1alpha3/service_entry.proto
hzxuzhonghu: Is there sa associated with EndpoinySlice/Service or plan to do that? 
BTW, this field is not ever supported.

networking/v1alpha3/service_entry.proto
smawson: Sounds reasonable to me, I'm OK with the restriction (old model or new model but not a mix).

networking/v1alpha3/service_entry.proto
smawson: re: why not directly register endpoint slices, it is because endpointslice is one per service, so if you have a vm with a few different services pointing at it you need to register them each independently. With workload entry + a controller creating endpoint slices you only need to track the endpoints in a single place.

re:Why ServiceEntry could select pods, it would be for cases where you don't need a Service object. You could have just a ServiceEntry and Pods.

That said I agree with the baby steps model, let's start with simple and straightforward (serviceentry selects workloadentry, service selects pod), we can get experience with that and evolve it based on feedback.

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,181 @@
...
+// $schema: istio.networking.v1alpha3.WorkloadEntry
+// $title: Workload Entry
+// $description: Configuration affecting VMs onboarded into the mesh.
+// $location: https://istio.io/docs/reference/config/networking/workload-entry.html
+// $aliases: [/docs/reference/config/networking/v1alpha3/workload-entry]
+
+// `WorkloadEntry` enables operators to describe the properties of
+// non-Kubernetes workloads such as VMs and bare metal servers as they
+// are onboarded into the mesh. A `WorkloadEntry` must be accompanied
+// by an Istio `ServiceEntry` that selects the workloads and provides
```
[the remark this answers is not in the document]
smawson: Per other conversation I'm fine doing this later if we have need.

networking/v1alpha3/workload_entry.proto
rshriram: endpoint slice still doesnt have labels, service accounts, weights, etc. These are vital to us for variety of purposes (version routing, authn/authz, etc).

networking/v1alpha3/service_entry.proto
howardjohn: I have not heard anything about this until now

rshriram:
@smawson / @hzxuzhonghu PTAL at the updates following discussions with @louiscryan 
The workload entry will now be one per endpoint (not multiple as before). This will allow us to auto create workload entries upon registration with istiod (or some other entity that manages lifecycle of the crd), along with updating the status field in the CRD.

Secondly, this update also allows the service entry to select both pods and VMs for the common vm to k8s migration scenario but not the other way around (k8s services selecting workload entries as the endpoint slice is supposed to automatically cover that use case. Any controller converting workload entries into endpoint slices will then allow both istio and non istio workloads to access such services using the k8s service name).

networking/v1alpha3/service_entry.proto
```diff
@@ -689,9 +812,22 @@ message ServiceEntry {
...
+  // supplies its own set of endpoints, the ServiceEntry will be
+  // treated as a decorator of the existing Kubernetes
+  // service. Properties in the service entry will be added to the
+  // Kubernetes service if applicable. Currently, the only the
+  // following additional properties will be considered by `istiod`:
+  //
+  // 1. subjectAltNames: In addition to verifying the SANs of the
+  //    service accounts associated with the pods of the service, the
+  //    SANs specified here will also be verified.
+  //
```
rshriram: To the reviewers: This semantic is introduced as a foundation for moving subsets into the ServiceEntry (wherein the service entry's host will refer to the k8s fqdn)

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,257 @@
...
+// Populated by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata"
+// -->
+//
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1alpha3
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+
```
ramaraochavali: proto definition for workload entry is missing?

ramaraochavali:
> This will allow us to auto create workload entries upon registration with istiod (or some other entity that manages lifecycle of the crd), along with updating the status field in the CRD.

Sorry, I am definitely missing some thing here, can you explain more on "upon registration with istiod" - is this same as connecting to Istiod with proxy when it has proxy to get config or is there some other registration involved?

networking/v1beta1/service_entry.proto
```diff
@@ -823,11 +959,25 @@ message ServiceEntry {
...
+ 
+    // The service account associated with the workload if a sidecar
+    // is present in the workload. The service account must be present
+    // in the same namespace as the configuration ( WorkloadEntry or a
+    // ServiceEntry)
+    string service_account = 7;
   };
 
   // One or more endpoints associated with the service.
   repeated Endpoint endpoints = 6;
```
hzxuzhonghu: Add `Only one of
  // `endpoints` or `workloadSelector` can be specified.` here

networking/v1alpha3/workload_entry.proto
rshriram: The workloadentry is simply the serviceEntry.Endpoint message. Do I have to create a separate message here to get this to compile?

rshriram:
@ramaraochavali the idea is that we can automatically create workload entry CRDs when a VM with sidecar connects to istiod. Once created, other istiods can use this info in their endpoint computations. the only downside to istiod doing all this is the lifecycle management of the workload entry CRD. We wont know when to delete the workload entry CRD [is it when proxy disconnects from an istiod or is there some other event]. Another alternative is for the istio agents on VMs to talk to some central health service that can then automatically create this crd. If the agent fails to provide heartbeat after a period of time, this health service can delete the CRD (and hence remove the endpoint from all clusters).

hzxuzhonghu:
>the idea is that we can automatically create workload entry CRDs when a VM with sidecar connects to istiod. Once created, other istiods can use this info in their endpoint computations. the only downside to istiod doing all this is the lifecycle management of the workload entry CRD. We wont know when to delete the workload entry CRD [is it when proxy disconnects from an istiod or is there some other event]. Another alternative is for the istio agents on VMs to talk to some central health service that can then automatically create this crd. If the agent fails to provide heartbeat after a period of time, this health service can delete the CRD (and hence remove the endpoint from all clusters).

@rshriram I would suggest doing this by the operator who deploys the vm app, who has the responsibility to create WorkloadEntry. Not to couple the process with pilot-agent or pilot. 
The defect is large quantities of connections to kube-apiserver if do this by pilot-agent, and it is not direct to infer the health status of the app with the xds connection to pilot/istiod.

networking/v1alpha3/service_entry.proto
```diff
@@ -824,11 +960,26 @@ message ServiceEntry {
     // The load balancing weight associated with the endpoint. Endpoints
     // with higher weights will receive proportionally higher traffic.
     uint32 weight = 6;
+
+    // The service account associated with the workload if a sidecar
+    // is present in the workload. The service account must be present
+    // in the same namespace as the configuration ( WorkloadEntry or a
+    // ServiceEntry)
+    string service_account = 7;
```
hzxuzhonghu: This should be the attribute of a specific workload. Different workloads may have different serviciceaccounts in theory

howardjohn:
> I would suggest doing this by the operator who deploys the vm app, who has the responsibility to create WorkloadEntry

That is exactly what we are trying to avoid, is the manual process required for VM registration. Maintaining these is a burden for users that isn't present on k8s

networking/v1alpha3/service_entry.proto
rshriram: Yes. Thats why I put it inside the endpoint object. One sa per endpoint if desired

rshriram:
>That is exactly what we are trying to avoid, is the manual process required for VM registration. Maintaining these is a burden for users that isn't present on k8s

As long as we do the registration as a separate health check controller outside of istiod, I dont mind. It can be turned off by people who dont want this. Doing in istiod is unnecessary complexity due to managing crd lifecycle as the workload's state changes (when to delete, when to recreate, handling flaky networks, etc )

hzxuzhonghu:
Yes, and with xds connection, we can not actually infer the workload health.

networking/v1alpha3/workload_entry.proto
ramaraochavali: Hmm.. Not sure if we need to have a separate message here - if it becomes an issue, we can do it later

ramaraochavali:
@rshriram Thanks for the details. One more question, when VM does not have sidecar and may be used for egress in mesh (I think that is talked about as one of the use case in this ?) - we assume that it is created by operators/some other process?

In any case, I agree with giving the option to turn off this auto creation would be useful (wherever this logic lives) at least during migration till people fully migrate from existing systems that they may have built that generates the service entry endpoints based on VMs.

howardjohn:
> In any case, I agree with giving the option to turn off this auto creation would be useful (wherever this logic lives) at least during migration till people fully migrate from existing systems that they may have built that generates the service entry endpoints based on VMs.

Auto creation and the API are not strictly coupled, as well, the API has value without auto creation, as it decouples workloads from Services, just like Pods.

ramaraochavali:
> Auto creation and the API are not strictly coupled, as well, the API has value without auto creation, as it decouples workloads from Services, just like Pods.

Yeah. Completely agree with this. Definitely not at all opposed to API change and it has definite advantage as it brings the right decoupling.

networking/v1alpha3/workload_entry.proto
smawson: I'm guessing this won't work, since the generator has no idea what message this should be applied to. We could maybe put these annotations on the Endpoint object?

Maybe we should consider renaming this to "ServiceEndpoint" given that it is now a single endpoint? then just pull Endpoint message out of ServiceEntry and call it ServiceEndpoint, assuming renaming proto message types is OK (doesn't affect yaml, doesn't affect serialized proto).

rshriram:
How about serviceInstance? Just throwing it out there. Don’t mind serviceEndpoint (though it slightly conflates with the api endpoint notion).

With regard to the grouping, I must have argued for grouping as well but Louis Ryan had a good point about being able to use the crd status message to record endpoint status. (Though keep  in mind, the status is a generic object, ie we can stick an array there for all we care).

smawson:
Suggested ServiceEndpoint so the name changes less, but ServiceInstance is closer to what we call this in our glossary (WorkloadInstance), so yeah go with that.

smawson:
Or WorkloadInstance if we are OK with CRD and proto having different names.

rshriram:
For now, i just retained it as WorkloadEntry to rhyme with ServiceEntry.. Let me know

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,311 @@
...
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1alpha3
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message WorkloadEntry {
+  // Address associated with the network endpoint without the
+  // port.  Domain names can be used if and only if the resolution is set
+  // to DNS, and must be fully-qualified without wildcards. Use the form
```
howardjohn: We no longer have `resolution` that can be set

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,311 @@
...
+  // unix:///absolute/path/to/socket for Unix domain socket endpoints.
+  string address = 1 [(google.api.field_behavior) = REQUIRED];
+
+  // Set of ports associated with the endpoint. The ports must be
+  // associated with a port name that was declared as part of the
+  // service. Do not use for `unix://` addresses.
+  map<string, uint32> ports = 2;
```
howardjohn: We briefly touched on this earlier, but I am not sure this makes sense independently, since there is no ServiceEntry.Port for us to reference the name of?

rshriram: yes this is the clunky part. But we need a mechanism to refer to the service ports implemented by the endpoint. One option is to do the `targetPort` thing in the port object but that loses the flexibility for different endpoints to select different service ports - this flexibility is useful when VMs migrate to pods and there are port changes. My gut feeling is that we should table this for now until we get more data, since not many are likely to even choose the ports per endpoint. Those who want to, will now have the burden of co-ordinating with the service entry port names.

networking/v1alpha3/workload_entry.proto
rshriram: This is in the workload entry. What this statement is saying is that if you declare DNS hostnames in your workloadEntry and your service is expecting STATIC resolution, prepare for this workloadEntry to be rejected. Remember that this is the first iteration. Once we get this going, we can start exploring the use of envoy aggregate clusters if we figure out how to scalably emit it.

howardjohn: Ah right. so we need to reference the SE here, makes sense. Thanks

networking/v1alpha3/workload_entry.proto
howardjohn: So for the normal case, no ports are mentioned here and they inherit the ServiceEntry.Port? That makes sense to me

howardjohn:
LGTM just one question. What are we selecting? the config meta or the spec label?
```yaml
kind: WorkloadEntry
metadata:
  name: foo-1
  metadata:
    labels:
      app: foo
spec:
  address: 1.2.3.5
  labels:
    app: bar
```

Do we match foo or bar?

rshriram:
We match the app:bar, i.e. the workload labels. Not the config meta. This is inline with k8s semantics.

howardjohn:
>  This is inline with k8s semantics.

K8s service is selecting the pods ConfigMeta labels though, it doesn't have the duplicated labels like we have here?

rshriram:
we cant remove endpoint labels though (backward compat). Plus all existing code is using the endpoint labels. Secondly, in serviceEntry we need a way to inline endpoints with labels.

rshriram:
@smawson this is ready to ship. Smaller nits we can do as we go.

howardjohn:
> we cant remove endpoint labels though (backward compat). Plus all existing code is using the endpoint labels. Secondly, in serviceEntry we need a way to inline endpoints with labels.

Yeah I think we just need to make it clear which one is used I guess?

louiscryan:
LGTM

networking/v1alpha3/workload_entry.proto
```diff
@@ -0,0 +1,311 @@
...
+
+  // The load balancing weight associated with the endpoint. Endpoints
+  // with higher weights will receive proportionally higher traffic.
+  uint32 weight = 6;
+
+  // The service account associated with the workload if a sidecar
+  // is present in the workload. The service account must be present
+  // in the same namespace as the configuration ( WorkloadEntry or a
+  // ServiceEntry)
+  string service_account = 7;
```
costinm: Can we rename this to service_account_name, matching the word everyone is using in Pod ? 

It is really hurtful for users to use a slightly different key for something that is so established, the
Pod spec has been around for a very long time.

networking/v1beta1/service_entry.proto
```diff
@@ -774,59 +911,17 @@ message ServiceEntry {
...
-
-  // One or more endpoints associated with the service.
-  repeated Endpoint endpoints = 6;
+  // Applicable only for MESH_INTERNAL services. Only one of
+  // `endpoints` or `workloadSelector` can be specified. Selects one
+  // or more Kubernetes pods or VM workloads (specified using
+  // `WorkloadEntry`) based on their labels. The `WorkloadEntry` object
+  // representing the VMs should be defined in the same namespace as
+  // the ServiceEntry.
+  WorkloadSelector workload_selector = 9;
```
costinm: What if the user has a Service object with a selector ? Do you mean the Service selector will not apply at all for WorkloadEntry ? This is extremely confusing and may lead to problems - in particular since the syntax of the selector is slightly different, so cut-and-pasting the selector will 
not work. 

Since a Service is still needed to get a ClusterIP - do we really need this, instead of just saying 
'WorkloadEntry is selected by the same rules as Pods'.

costinm:
Sorry for the late comments - I was very happy with the change and focused on backward compat, and missed the naming missmatch and duplication.

howardjohn:
WorkloadSelector is used by other Istio objects though, right? So its a
tradeoff between aligning with Service and with Istio configs?

PR #1333, opened 2020-03-15
Localhost TLS termination and initiation in Sidecar

rshriram:


networking/v1alpha3/sidecar.proto
```diff
@@ -737,3 +758,83 @@ enum CaptureMode {
...
+//       privateKey: /etc/legacy/private.pem
+//   egress:
+//   - hosts:
+//     - "./"
+// ```
+// {{</tab>}}
+// {{</tabset>}}
+//
+message Localhost {
+  // TLS settings to be used by the sidecar (client) when
```
rshriram: We can use this message to pile in future stuff like DNS settings (@costinm )

networking/v1alpha3/sidecar.proto
```diff
@@ -561,7 +562,9 @@ message Sidecar {
   // PeerAuthentication policy disables mTLS tunneling for one or more
   // ports in the workload, the TLS settings specified here will be
   // applied.
-  Server.TLSOptions inbound_tls = 5;
+  ServerTLSSettings inbound_tls = 5;
+
+  Localhost localhost = 6;
```
howardjohn: comment?

networking/v1alpha3/sidecar.proto
```diff
@@ -454,6 +455,12 @@ message Sidecar {
   // inherits the system detected defaults from the namespace-wide or
   // the global default Sidecar.
   OutboundTrafficPolicy outbound_traffic_policy = 4;
+
+  // `Localhost` describes the sidecar settings related to the
+  // communication between the user's services and the sidecar on a
+  // Kubernetes Pod or a VM. These settings apply by default to all
+  // ingress and egress listeners in a sidecar unless overridden.
+  Localhost localhost = 6;
```
hzxuzhonghu: How can distinguish different tls settings for multiple port?

rshriram: In each ingressListener, we have a ClientTLSSetting (for traffic from sidecar to app). and in each egress, we have a ServerTLSsetting for traffic from app to sidecar.

rshriram:
I had to update the release lock status on all releases as the proto messages were renamed for clarity.

networking/v1beta1/sidecar.proto
```diff
@@ -622,3 +647,83 @@ enum CaptureMode {
...
+//   egress:
+//   - hosts:
+//     - "./"
+// ```
+// {{</tab>}}
+// {{</tabset>}}
+//
+message Localhost {
+  // TLS settings to be used by the sidecar (client) when
+  // forwarding traffic to the user's process (server) on the localhost.
```
ramaraochavali: nit: It may be a good idea to document the default behaviour of "Localhost" settings?

rshriram: done

networking/v1alpha3/gateway.proto
```diff
@@ -563,126 +563,10 @@ message Server {
...
-    // Optional: If specified, only support the specified cipher list.
-    // Otherwise default to the default cipher list supported by Envoy.
-    repeated string cipher_suites = 9;
-  }
-
   // Set of TLS related options that govern the server's behavior. Use
   // these options to control if all http requests should be redirected to
   // https, and the TLS modes to use.
-  TLSOptions tls = 3;
+  ServerTLSSettings tls = 3;
```
nrjpoddar: Thanks for moving them into a usable struct. Cheers!

nrjpoddar: If client/server is confusing we can also use "outbound/inbound"

networking/v1alpha3/sidecar.proto
```diff
@@ -454,6 +455,14 @@ message Sidecar {
...
   // the global default Sidecar.
   OutboundTrafficPolicy outbound_traffic_policy = 4;
+
+  // `Localhost` describes the sidecar settings related to the
+  // communication between the user's services and the sidecar on a
+  // Kubernetes Pod or a VM. These settings apply by to all ingress
+  // and egress listeners in a sidecar unless overridden. There are no
+  // built in defaults for this setting. If not specified, the
+  // features will be disabled.
+  Localhost localhost = 6;
```
nrjpoddar: This is just convenience, I would rather remove this and only have ingress/egress settings. We can also add this later but I don't see a reason to provide this to start with.

rshriram: It is not just convenience. Having to enumerate all the inbound ports and their protocols is a chore. You also cannot do this at namespace wide level without the localhost field.

releaselocks/release-1.1/proto.lock.status
```diff
@@ -2,7 +2,24 @@ CONFLICT: "AuthorizationPolicy" field: "allow" has been removed, but is not rese
 CONFLICT: "AuthorizationPolicy" field: "workload_selector" has been removed, but is not reserved [rbac/v1alpha1/rbac.proto]
 CONFLICT: "AuthorizationPolicy" ID: "1" has been removed, but is not reserved [rbac/v1alpha1/rbac.proto]
 CONFLICT: "AuthorizationPolicy" ID: "2" has been removed, but is not reserved [rbac/v1alpha1/rbac.proto]
+<<<<<<< HEAD
```
hzxuzhonghu: conflict

hzxuzhonghu: check all lock.status

networking/v1alpha3/sidecar.proto
```diff
@@ -454,6 +455,14 @@ message Sidecar {
   // inherits the system detected defaults from the namespace-wide or
   // the global default Sidecar.
   OutboundTrafficPolicy outbound_traffic_policy = 4;
+
+  // `Localhost` describes the sidecar settings related to the
+  // communication between the user's services and the sidecar on a
+  // Kubernetes Pod or a VM. These settings apply by to all ingress
```
louiscryan: s/apply by/apply

networking/v1alpha3/sidecar.proto
```diff
@@ -454,6 +455,14 @@ message Sidecar {
   // inherits the system detected defaults from the namespace-wide or
   // the global default Sidecar.
   OutboundTrafficPolicy outbound_traffic_policy = 4;
+
+  // `Localhost` describes the sidecar settings related to the
+  // communication between the user's services and the sidecar on a
```
louiscryan: is there a better description than 'users services'. 
Something like:

`Localhost` defines settings for communication between the sidecar and the workload it is attached to

networking/v1alpha3/sidecar.proto
```diff
@@ -623,3 +648,83 @@ enum CaptureMode {
...
+// ```
+// {{</tab>}}
+// {{</tabset>}}
+//
+message Localhost {
+  // TLS settings to be used by the sidecar (client) when
+  // forwarding traffic to the user's process (server) on the localhost.
+  //
+  // **NOTE**: DISABLE, SIMPLE and MUTUAL are the only valid TLS modes.
+  ClientTLSSettings client_tls = 1;
```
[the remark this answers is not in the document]
rshriram: We have Sidecar wide settings under the localhost and per ingress/egress setting as well. That covers the variation by ports doesn’t it ?

networking/v1alpha3/sidecar.proto
```diff
@@ -17,6 +17,7 @@ syntax = "proto3";
 import "google/api/field_behavior.proto";
 import "networking/v1alpha3/gateway.proto";
 import "networking/v1alpha3/virtual_service.proto";
+import "networking/v1alpha3/destination_rule.proto";
```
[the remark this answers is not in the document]
rshriram: Jason told me that for now we have to replicate functionality in both apis. Until the conversion logic is in place.

nrjpoddar:
Are multiple Sidecar config resources attached to the same workload currently merged? What will happen if there are conflicting localhost TLS settings for the same workload?

For VM use case, do you think localhost is the right term to use? I’m assuming yes since IPtables interception is done over localhost iirc. Do we support other kinds of interception where localhost won’t make sense? I’m thinking about different proxy deployment topologies like per-node. @louiscryan @rshriram wdyt?

rshriram:
There can be only one sidecar for a workload. It is clearly specified in our docs as well.

rshriram:
Per-node is a much bigger story for the moment - everything relating to trust boundaries, shared fate, etc. will come into play. And it will still be "localhost" wrt the node and not across nodes.

PR #1340, opened 2020-03-17
Add Josh to master code owner list

rshriram:
To reflect TOC changes


louiscryan:
agree with @howardjohn comment. Other than that - approved.

PR #1342, opened 2020-03-19
Add UNSPECIFIED option for ingress mode

howardjohn:
Otherwise, its impossible to tell if something is set or if its OFF,
meaning we cannot change the default. This blocks
https://github.com/istio/istio/pull/22326

I am not sure if this is something that is legal to change though...


smawson:
I think this needs to change on a version boundary, although not sure, are we actually using the proto on-wire format for storage anywhere or are things stored via JSON representation?

If all storage is in JSON representation then this change would be safe I think?

kyessenov:
Enums are open ended so it can be parsed by older pilots if we enable allow unknown fields in JSON. Don't renumber fields though. That and changing field types are two things you should never do in a proto.

howardjohn:
We do allow unknown fields so that's not an issue. the problem is we messed made the default value have meaning. I want to be able to distinguish between off and unset

mesh/v1alpha1/config.proto
```diff
@@ -106,22 +106,25 @@ message MeshConfig {
   string ingress_service = 8;
 
   enum IngressControllerMode {
+    // Unspecified Istio ingress controller.
+    UNSPECIFIED = 0;
+
```
rshriram: Since we are not using Galley and no one is streaming stuff over the wire to Pilot, this should be okay

PR #1344, opened 2020-03-23
Remove old envoyfilter fields

rshriram:
These were deprecated 3 releases ago. Even the code has been removed from Pilot.


networking/v1alpha3/envoy_filter.proto
```diff
@@ -704,11 +583,8 @@ message EnvoyFilter {
     Patch patch = 3;
   }
 
-  // Deprecated. Use workload_selector instead.
-  // $hide_from_docs
-  map<string, string> workload_labels = 1 [deprecated = true];
-  // $hide_from_docs
-  repeated Filter filters = 2 [deprecated = true];
+  reserved 1, 2;
```
duderino: This means that when we add new fields the positions for the removed fields will not be reused?

rshriram: Yeah

PR #1351, opened 2020-03-25
Hide v1alpha1 RBAC and Authentication policy

yangminzhu:
https://github.com/istio/istio/issues/22060


diemtvu:
> Is there any reason we cannot remove this completely? At least for authN, we will remove all code related to v1alpha1 soon.

Nvm. I saw it is still used (internally) in authn filter. It's a tech debt for not having this filter rewrite when doing beta api :(

PR #1354, opened 2020-03-25
Add configuration for prometheus aggregation

howardjohn:
Design: https://docs.google.com/document/d/1TTeN4MFmh4aUYYciR4oDBTtJsxl5-T5Tu3m3mGEdSo8/edit
Issue: istio/istio#21366
Implementation: https://github.com/istio/istio/pull/22318 - complete other than enabling/disabling based on this API


annotation/annotations.yaml
```diff
@@ -13,6 +13,13 @@
 #   limitations under the License.
 
 annotations:
+  - name: prometheus.istio.io/aggregate
```
douglas-reid: fwiw, i worry that `aggregate`  is a bit overloaded when it comes to metrics and prometheus.  I don't know what a better name would be though. maybe: `joinExports` ?

howardjohn: dumping ideas:

* aggregate
* aggregate-metrics
* join-exports
* merge-exports
* join-metrics
* agent-metrics
* aggregate-metrics-in-the-agent-by-rewriting-prometheus-annotations-and-processing-in-the-agent
* rewrite-prometheus-scraping (similar to rewriteAppPropers)
* rewrite-prometheus-annotations
* rewrite-prometheus

I don't have a strong opinion here. I agree aggregate is pretty overloaded. Maybe rewrite makes sense?

howardjohn: @istio/wg-policies-and-telemetry-maintainers any strong opinions?

dcberg: Agree about the comment about `aggregate`. The term `rewrite` seems too technical about the internal functionality versus a user facing descriptive term for what is happing.
While looking at the changes I just kept thinking that as a developer I am expecting that the metrics from my application, envoy, and the agent are going to be merged together if I use this annotation.

What about `merge-metrics`?

mesh/v1alpha1/config.proto
```diff
@@ -417,8 +417,17 @@ message MeshConfig {
   // "kube-system" will be cluster-local.
   repeated string cluster_local_namespaces = 50;
 
+  // If enabled, Istio agent will aggregate metrics exposed by the application with metrics from Envoy
+  // and Istio agent. The sidecar injection will replace prometheus.io annotations present on the pod
+  // and redirect them towards Istio agent, which will then handle the aggregation of metrics.
+  // If your prometheus is not setup to scrape based on prometheus.io/scrape, prometheus.io/port, and
+  // prometheus.io/path annotations, you may need to disable this.
```
linsun, line 424: How could this happen?  I assume the default prometheus we ship already have this configured.

mesh/v1alpha1/config.proto
```diff
@@ -417,8 +417,17 @@ message MeshConfig {
   // "kube-system" will be cluster-local.
   repeated string cluster_local_namespaces = 50;
 
+  // If enabled, Istio agent will aggregate metrics exposed by the application with metrics from Envoy
```
linsun, line 420: Can we say the default behavior is enabled?

annotation/annotations.yaml
linsun: or just merge?

mesh/v1alpha1/config.proto
```diff
@@ -417,8 +417,17 @@ message MeshConfig {
...
+  // and Istio agent. The sidecar injection will replace prometheus.io annotations present on the pod
+  // and redirect them towards Istio agent, which will then handle the aggregation of metrics.
+  // If your prometheus is not setup to scrape based on prometheus.io/scrape, prometheus.io/port, and
+  // prometheus.io/path annotations, you may need to disable this.
+  // If you application exposes the same metrics as Envoy, this may cause issues. In this case, it is recommended
+  // to disable aggregation on that deployment with the `prometheus.istio.io/aggregate: "false"` annotation.
```
linsun, line 426: If user disables the aggregate, would user see the metrics from envoy or from application when both exposes the same metrics?

howardjohn: It is dependant on their scraping configuration. If they use the out of the box Prometheus they would see the application metrics only. If they opt out of aggregation they are expected to configure there scraping to scrape the envoy themselves (which is fairly easy to do and will be documented in istio.io - probably too verbose for the API docs?)

mesh/v1alpha1/config.proto
howardjohn: Yes, good idea

mesh/v1alpha1/config.proto
howardjohn: users may configure Prometheus however they want - not everyone configures using these annotations, although our demo Prometheus will

annotation/annotations.yaml
mandarjog: Merge or proxy.
Definitely not aggregate

mesh/v1alpha1/config.proto
mandarjog: This does not make sense. If someone is using non standard annotations, we need to ask them what they are and set them up accordingly,  not disable.

mesh/v1alpha1/config.proto
mandarjog: “Applications exposing same metrics as envoy” is such. Fringe case that it only serves here to confuse.

Let’s be more specific: example if you are running a separately managed envoy Ingres with an istio sidecar and agent then both envoys will emit the same metrics. In that case the metrics specified by the application envoy will be used.

mandarjog: Typo: Start of sentence 
“If your application”

annotation/annotations.yaml
howardjohn: Changed to merge. Thanks!

howardjohn:
@dcberg @linsun can you take another look? I updated to merge as suggested

howardjohn:
@istio/technical-oversight-committee can you take a look at the revised changes? Thanks

mesh/v1alpha1/config.proto
```diff
@@ -417,8 +417,19 @@ message MeshConfig {
   // "kube-system" will be cluster-local.
   repeated string cluster_local_namespaces = 50;
 
+  // If enabled, Istio agent will merge metrics exposed by the application with metrics from Envoy
+  // and Istio agent. The sidecar injection will replace prometheus.io annotations present on the pod
+  // and redirect them towards Istio agent, which will then merge metrics of from the application with Istio metrics.
+  // This releases on the annotations prometheus.io/scrape, prometheus.io/port, and
```
linsun, line 423: @howardjohn what do you mean by releases?  Do you mean user no longer needs to use these annotations to annotate the pod? https://istio.io/faq/metrics-and-logs/#prometheus-application-metrics

linsun:
LG to me, just had a small english questions.  would prefer to change to eliminate the need....

mesh/v1alpha1/config.proto
howardjohn: typo, I meant relies.

And no, the user needs to specify those for **application** metrics. This is how prometheus works, even without Istio. For Istio metrics they don't need to do anything

PR #1358, opened 2020-03-28
add status for addon components.

morvencao:
resolve: https://github.com/istio/istio/issues/22180


operator/v1alpha1/operator.proto
```diff
@@ -210,6 +215,9 @@ message InstallStatus {
     Status status = 1;
     // Individual status of each component controlled by the operator. The map key is the name of the component.
     map<string, VersionStatus> component_status = 2;
+
+    // Individual status of each addon component controlled by the operator. The map key is the name of the addon component.
+    AddonComponentStatus addon_component_status = 3;
```
howardjohn: any reason to not just use `component_status`? And if not, why is this its own message but component_status is not?

morvencao: `component_status` can only show aggregated status for all addon components:
```
  status:
    componentStatus:
      AddonComponents:
        status: HEALTHY
```

we may expect it should provide more detailed info for each add component status:

    status:
      addonComponentStatus:
        prometheus:
          status: HEALTHY
        kiali:
          status: HEALTHY
...

howardjohn: why not just change it to not aggregate?

morvencao: emm, yeah
my intention is to make the `status` consistent with `spec`, since we have dedicated field for `addonComponent` in spec: https://github.com/istio/api/blob/master/operator/v1alpha1/operator.pb.go#L99

rshriram:
Waiting for approval from Howard or someone else who knows this best

morvencao:
Closing as we have new design for addons.

PR #1359, opened 2020-03-30
Adding ClusterLocalNamespaces to mesh config

nmittler:
This is currently kept as a hidden/undocumented property


mesh/v1alpha1/network.pb.go
```diff
@@ -157,10 +157,10 @@ type isNetwork_NetworkEndpoints_Ne interface {
 }
 
 type Network_NetworkEndpoints_FromCidr struct {
-	FromCidr string `protobuf:"bytes,1,opt,name=from_cidr,json=fromCidr,proto3,oneof"`
+	FromCidr string `protobuf:"bytes,1,opt,name=from_cidr,json=fromCidr,proto3,oneof" json:"fromCidr,omitempty"`
 }
 type Network_NetworkEndpoints_FromRegistry struct {
-	FromRegistry string `protobuf:"bytes,2,opt,name=from_registry,json=fromRegistry,proto3,oneof"`
+	FromRegistry string `protobuf:"bytes,2,opt,name=from_registry,json=fromRegistry,proto3,oneof" json:"fromRegistry,omitempty"`
```
howardjohn: seems like this probably shouldn't have changed and is a result of a tool mismatch on gen or something?

mesh/v1alpha1/config.proto
```diff
@@ -410,7 +410,10 @@ message MeshConfig {
   ThriftConfig thrift_config = 49;
 
   // $hide_from_docs
-  // Next available field number: 50
+  repeated string ClusterLocalNamespaces = 50;
```
howardjohn: A comment here would be useful

howardjohn: Seems like snake_case is the standard here, not sure if its enforced but may make sense to follow that

nmittler:
@howardjohn PTAL

mesh/v1alpha1/config.proto
```diff
@@ -410,7 +410,10 @@ message MeshConfig {
   ThriftConfig thrift_config = 49;
 
   // $hide_from_docs
-  // Next available field number: 50
+  repeated string cluster_local_namespaces = 50;
```
smawson: Can we at least add some comment here on what this is for? Just so if someone peeks at the proto they know why it exists...

rshriram:
Why do we need another field in the mesh config? We have the exportTo setting. users can add an annotation to the namespaces indicating what is and what is not exported.

Issue #1360, opened 2020-03-30
Annotations code generator produces invalid code


esnible:
To reproduce, add

```
  - name: istio.io/rev
    description: The Istio control plane revision associated with the resource
    deprecated: false
    hidden: false
    resources:
      - Service
      - Pod
```

to _istio.io/api/annotation/annotations.yaml_ and do `make`.

The produced output includes `Resources: []ResourceTypes{ ServicePod, },` which is invalid.  (It should be `Resources: []ResourceTypes{ Service, Pod, },`.

knight42:
@esnible Hi, this is a bug of `istio/tools`, and I have filed https://github.com/istio/tools/pull/1075 to address this issue.

knight42:
This issue should be fixed now, please feel free to reopen it if you encounter any problems.

PR #1361, opened 2020-03-30
Label for the Istio Control Plane revision and the labels created by the Operator

esnible:
We need the new `istio.io/rev` annotation to appear in our documentation.

This annotation can be applied to Namespaces (controls injection), Services (may be used informationally by future analyzers and istioctl), and Pods (may be used informationally by future istioctl).  This PR claims the annotation can apply to `Any` because of https://github.com/istio/api/issues/1360

We also desire a constant for Go code to implement https://github.com/istio/istio/pull/22511 correctly.

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -335,10 +335,3 @@ annotations:
     hidden: false
     resources:
       - Any
-  - name: istio.io/rev
-    description: The Istio control plane revision associated with the resource
-    deprecated: false
-    hidden: false
-    resources:
-      # TODO change this to Service, Pod, Namespace when that combination is supported; https://github.com/istio/api/issues/1360
-      - Any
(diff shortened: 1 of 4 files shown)


annotation/annotations.yaml
```diff
@@ -335,3 +335,10 @@ annotations:
     hidden: false
     resources:
       - Any
+  - name: istio.io/rev
```
howardjohn: This is not an annotation, its label. Maybe should be in the labels/ package?

esnible:
@howardjohn Thanks for catching that this is now a label.

While moving this to _labels_ I noticed that the labels created by the Operator were not defined in API and decided to add them.  @ostromart do you concur that with the merge of istio/operator into istio/istio it makes sense to include the Operator's API in istio/api?

ostromart:
@esnible yes I think any operator label definitions should move to istio/api. Thanks for pointing that out.

esnible:
@smawson Please re-review.

PR #1366, opened 2020-04-03
Align operator API with reality

howardjohn:
These components no longer exist

WIP until https://github.com/istio/istio/pull/22718 merges


operator/v1alpha1/component.proto
```diff
@@ -26,17 +26,15 @@ option go_package="istio.io/api/operator/v1alpha1";
 message IstioComponentSetSpec {
     BaseComponentSpec base = 29;
     ComponentSpec pilot = 30;
-    ComponentSpec proxy = 31;
-    ComponentSpec sidecar_injector = 32;
     ComponentSpec policy = 33;
     ComponentSpec telemetry = 34;
-    ComponentSpec citadel = 35;
-    ComponentSpec node_agent = 36;
-    ComponentSpec galley = 37;
```
ostromart, line 35: I believe these must be deprecated before they can be deleted - per API policy but also to avoid potential upgrade problems.

howardjohn: What upgrade issues will we run into? If you have `citadel` specified, you are going to be confused I would imagine, as it does nothing?

ostromart: For in-place upgrade, user is likely to reuse the old CR - we may want to do something other than just ignore those fields, like issuing a warning.

howardjohn: @ostromart what do you want to do here? I am not 100% sure what the UX would be for istioctl/operator if we mark this deprecated or delete? Do you have some info here?

ostromart: after thinking about this some more, marking as deprecated may confuse users just as much as the generic error message. i suggest then that we proceed with removing these, but add some check in the code only if a user attempts to access these components (as they may do in an upgrade scenario when they roll over a previous API version).

howardjohn:
@istio/technical-oversight-committee can you take a look?

PR #1368, opened 2020-04-08
add default extra stat option

kyessenov:
Injects the pod annotations by default so that each pod does not need to be annotated. 
cc @nrjpoddar @mandarjog 


mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,12 @@ message MeshConfig {
   // If not specified, this will be enabled by default.
   google.protobuf.BoolValue enable_prometheus_merge = 51;
 
+  // Default list of stat tags to extract for the in-proxy telemetry. The value is injected into each pod as
+  // `sidecar.istio.io/extraStatTags` annotation if it is not already present in the pod.
```
howardjohn, line 432: nit: I don't think we need to describe the implementation here unless it has user impact?

In general though I don't think we should implement it from annotation, why not just handle it in pilot-agent directly? It has access to this config now.

Also this seems like it should be in proxyConfig not meshConfig

kyessenov: Annotation is needed for a per-pod override. So it deserves a mention I think. This change is just a minimal step to fix user pain.

howardjohn: I am fine with mentioning that this will be the default if the annotation is not set but I don't think we should say " The value is injected into each pod as .. annotation". Mostly because I don't think it should be implemented that way (unless there is some reason not mentioned here. I don't have any context for this change)

kyessenov: It is implemented by reading the annotation. So it's already there. I don't think it's worth removing the annotation now since that means we need write more code to propagate it to bootstrap. Basically, with this change I just need to change injector and not touch pilot agent at all.

howardjohn: We are trying to make the injection template dumb and pilot agent smart. making it implemented by an annotation means it won't work for gateways either

kyessenov: How does pilot agent fetch meshconfig?

howardjohn: The injector adds it as an environment variable. For gateway it mounts the configmap directly. We no longer are doing the weird hacks we used to where we go from mesh config -> Env var/flag -> back to mesh config, instead just directly give agent the whole mesh config

kyessenov: How does it not break 4k environment variable limit? that sounds fine, that's how mesh config was mounted as configmap back in 0.1 or so

howardjohn: It might if we start adding a bunch of really verbose stuff to mesh config I guess. For now its fine. We also ignore any default settings so for most users it will be almost empty

kyessenov:
Updated to move into proxy config.

mesh/v1alpha1/proxy.proto
```diff
@@ -232,6 +232,10 @@ message ProxyConfig {
 
   // Port on which the agent should listen for administrative commands such as readiness probe.
   int32 status_port = 26;
+
+  // An additional list of tags to extract from the in-proxy Istio telemetry. These extra tags can be
+  // added by configuring the telemetry extension. Each additional tag needs to be present in this list.
+  repeated string extra_stat_tags = 27;
```
nrjpoddar, line 238: Can you add example here for the format for this annotation?

kyessenov: It's a repeated string. So it's literally a string of tag names. Are you looking for EnvoyFilter sample here? I'd hope this is just temporary until the tag names could be declared in the code.

nrjpoddar: I see. These are just tag names irrespective of metric name. I thought you need to associate tags with specific metrics.

kyessenov: Tags can be associated, but this setting just needs tag names.

mesh/v1alpha1/config.proto
kyessenov: Ok, as long as it propagates to agents. I'll let someone else resolve the upper limit issue in the environment variable.

kyessenov:
This is ready. PTAL.

mesh/v1alpha1/proxy.proto
```diff
@@ -232,6 +232,10 @@ message ProxyConfig {
 
   // Port on which the agent should listen for administrative commands such as readiness probe.
   int32 status_port = 26;
+
+  // An additional list of tags to extract from the in-proxy Istio telemetry. These extra tags can be
+  // added by configuring the telemetry extension. Each additional tag needs to be present in this list.
```
howardjohn, line 237: nit: example of one? Or a link with more info? Just reading this I don't know how I would use this.

kyessenov: I can't add a link because I'd need to reference this field in any docs :) This is the only user facing config apart from EnvoyFilter wiki documentation.

kyessenov: https://github.com/istio/istio/wiki/Configurable-V2-Metrics

@nrjpoddar @jacob-delgado are you interested in getting this in 1.6? I am open to using annotations, label on istiod, whatever to avoid labelling each pod.

nrjpoddar: We need this or we can't move to v2 in 1.6 also for our customers. Pushing this Pod label burden to customers is not acceptable.

What's the hold up here?

kyessenov: I don't know if there's a hold up. No one seems to be able to approve this.

nrjpoddar: It seems there's a merge conflict, so can you fix that?

I will ping people to get this moving.

nrjpoddar: do you have the corresponding istio/istio changes?

kyessenov: No, but I don't estimate it to be hard. There was a churn in how mesh config is propagated that is hopefully settled now.

linsun: can we have some example for this and why user would need to add this?  Based on the wiki link, I assume this is for adding additional dimensions for prometheus, for example destination_port,request_host are not covered by default (BTW where can user finds the default) and can use this `extraStatTags` to add them globally or use pod annotation to add it per pod.

mandarjog: @linsun we hope that we don’t stay with this annotation for long and simply changing stats config will suffice just like it did in a Mixer.
The defaults are specified in the bootstrap configuration.
The way I see this used is that a user will first go to the “configuring telemetry” page and then add this annotation when instructed to do so. I don’t think it will go the other way .

@kyessenov 
We should add one extra line:
Extra tags emitted by the telemetry extension must be listed here so that they can be correctly processed and exposed as Prometheus metrics.

kyessenov: Updated.

kyessenov:
Can we make some progress here? It's been two weeks just to add a single string option. You're asking for docs but this is not the place to write docs, and the docs are being delayed because we can't use this option.

kyessenov:
CI failing for unrelated reasons `Makefile.core.mk:528: recipe for target 'kubernetes/customresourcedefinitions.gen.yaml' failed
make: *** [kubernetes/customresourcedefinitions.gen.yaml] Error 127`
/retest

PR #1369, opened 2020-04-08
Add mesh-wide support for custom tags in tracing

jacob-delgado:
To partially address https://github.com/istio/istio/issues/22457

Until there is a first class API support this is a partial solution for implementing custom spans when using tracing.

Based on https://github.com/envoyproxy/envoy/blob/master/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L148 and https://github.com/envoyproxy/envoy/blob/master/api/envoy/type/tracing/v3/custom_tag.proto


mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,54 @@ message MeshConfig {
...
+
+  // Literal type custom tag with static value for the tag value.
+  message Literal {
+    // Static literal value to populate the tag value.
+    string value = 1;
+  }
+
+  // Environment type custom tag with environment name and default value.
+  message Environment {
+    // Environment variable name to obtain the value to populate the tag value.
```
howardjohn, line 454: which env? proxy? or control plane?

mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,54 @@ message MeshConfig {
   // If not specified, this will be enabled by default.
   google.protobuf.BoolValue enable_prometheus_merge = 51;
 
+  // Describes custom tags for the active span.
+  message CustomTags {
```
nrjpoddar, line 432: ProxyConfig which is inside MeshConfig already has a Tracing section, let's put all of this inside that.

mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,54 @@ message MeshConfig {
...
+
+    // When the header does not exist,
+    // the tag value will be populated with this default value if specified,
+    // otherwise no tag will be populated.
+    string default_value = 2;
+  }
+
+  // Specify mesh-wide custom tags to be added to each tracing span
+  // for each sidecar that is deployed.
+  repeated CustomTags custom_tags = 52;
```
nrjpoddar, line 475: can we make this a map instead of list where "tag name" is the key? It just reduces duplicates and more checks in the implementation.

mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,54 @@ message MeshConfig {
...
+    // Environment variable name to obtain the value to populate the tag value.
+    string name = 1;
+    // When the environment variable is not found,
+    // the tag value will be populated with this default value if specified,
+    // otherwise no tag will be populated.
+    string default_value = 2;
+  }
+
+  // Header type custom tag with header name and default value.
+  message Header {
```
nrjpoddar, line 463: Header -> RequestHeader

mesh/v1alpha1/config.proto
nrjpoddar: Environment is the proxy environment afaik. To use this you would need to inject the environment somehow, add documentation for that.

jacob-delgado: Changed name to `ProxyEnvironment`. Will clarify documentation

kyessenov:
What's the policy on embedding xDS protos here? It's a lot of redundancy with xDS source of truth.

nrjpoddar:
> What's the policy on embedding xDS protos here? It's a lot of redundancy with xDS source of truth.

That's a great question! I don't see any current example of embedding Envoy protos here. I think we should maintain that decoupling.

jacob-delgado:
> What's the policy on embedding xDS protos here? It's a lot of redundancy with xDS source of truth.

That is definitely something that bothered me, but I couldn't find an import to an already existing xDS proto in the api folder. As we iterate on the api, we should see what the final outcome looks like. But some of it depends on how much we intend to keep. For instance. I didn't bring over the "metadata" field and @nrjpoddar has requested a change of repeated list to a map.

kyessenov:
There used to be opposition to just directly embedding xDS. The technical obstacle is gogo/golang proto divergence. It is painful to have two versioned APIs that are basically the same thing in this case.

mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,54 @@ message MeshConfig {
...
+    // Used to populate the tag name.
+    string tag = 1;
+    // Used to specify what kind of custom tag.
+    oneof type {
+      // A literal custom tag.
+      Literal literal = 2;
+      // An environment custom tag.
+      Environment environment = 3;
+      // A request header custom tag.
+      Header header = 4;
```
jacob-delgado: Do we wish to support these 3 different types? Metadata from the [xDS api](https://github.com/envoyproxy/envoy/blob/f84440dc4f95890f14e2e0686b07258f030b54b3/api/envoy/type/tracing/v3/custom_tag.proto#L99) is missing. I wasn't sure how we'd support that.

mesh/v1alpha1/config.proto
jacob-delgado: Changed

mesh/v1alpha1/config.proto
jacob-delgado: Done

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,59 @@ message Tracing {
     // Use a Stackdriver tracer.
     Stackdriver stackdriver = 4;
   }
+
+  // Describes custom tags for the active span.
```
nrjpoddar, line 103: Can we make this more clear? May be something like:

// Configure custom tags that will added to any active span.
// Tags can be generated via literals, environment variables or an incoming request header.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,59 @@ message Tracing {
...
+  message Environment {
+    // Environment variable used to populate the tag's value.
+    string name = 1;
+    // When the environment variable is not found,
+    // the tag's value will be populated with this default value if specified,
+    // otherwise the tag will not be populated.
+    string default_value = 2;
+  }
+
+  // RequestHeader type represents the http header to grab the value from
```
nrjpoddar, line 135: http --> HTTP

// RequestHeader is the HTTP request header which will be used to populate the span tag.
// A default value can be configured if the header doesn't exist.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,59 @@ message Tracing {
...
+  // and populate the tag value, as well as a default if the header does not
+  // exist
+  message RequestHeader {
+    // HTTP header name used to obtain the value from to populate the tag value.
+    string name = 1;
+
+    // When the incoming http header does not exist,
```
nrjpoddar, line 142: // Default value to be used for the tag when the named HTTP header doesn't exist.
// The tag will be skipped if no default value is provided.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,59 @@ message Tracing {
...
+    }
+  }
+
+  // Literal type represents a static value.
+  message Literal {
+    // Static literal value used to populate the tag value.
+    string value = 1;
+  }
+
+  // Environment type represents the proxy's environment to populate
```
nrjpoddar, line 124: Environment is the proxy's environment variable to be used for populateing the custom span tag.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,59 @@ message Tracing {
...
+  // tag values from
+  message Environment {
+    // Environment variable used to populate the tag's value.
```
nrjpoddar, line 127: Name of the environment variable

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,59 @@ message Tracing {
...
+
+    // When the incoming http header does not exist,
+    // the tag value will be populated with this default value if specified,
+    // otherwise tag will not be populated.
+    string default_value = 2;
+  }
+
+  // Specify mesh-wide custom tags to be added to each tracing span
+  // for each sidecar that is deployed.
+  map<string, CustomTag> custom_tags = 5;
```
nrjpoddar, line 150: // Configures the custom tags to be added to active span by all proxies.
// The key represents the name of the tag. Add an expanded example here.

nrjpoddar: Does it affect gateways & sidecars? I'm assuming yes.

mesh/v1alpha1/config.proto
nrjpoddar: That's not needed.

mesh/v1alpha1/proxy.proto
jacob-delgado: Yes. See **https://github.com/istio/istio/blob/35e75428b127d6aaf5a9b0aff88b001baa383d0c/pilot/pkg/networking/core/v1alpha3/listener.go#L442**

jacob-delgado:
@douglas-reid raised an interesting point. What should we do when there is a mesh wide setting and EnvoyFilter based config? Correct me if I'm wrong Doug.

It's conceivable that users would want to have different tags and with EnvoyFilter they could create one that removes/alters or overrides the meshconfig setting on a namespace (or istio root ns) setting.

nrjpoddar:
> @douglas-reid raised an interesting point. What should we do when there is a mesh wide setting and EnvoyFilter based config? Correct me if I'm wrong Doug.
> 
> It's conceivable that users would want to have different tags and with EnvoyFilter they could create one that removes/alters or overrides the meshconfig setting on a namespace (or istio root ns) setting.

EnvoyFilter patching is a dump operation i.e. if users needs to preserve mesh wide settings & override it via EnvoyFilter then users will need to provide all the config in one place or the mesh wide settings will get replaced with what EnvoyFilter provides. This is my understanding.

douglas-reid:
@nrjpoddar do we need a design doc for the UX here? 

thinking out loud: if we expect a first class API to appear at some point, are we intro'ing more complexity here just in an attempt to avoid `EnvoyFilter`?  If I want namespace-specific settings and that means `EnvoyFilter` in my namespace anyway (vs. one in `istio-system` or whatever), have we saved anything by adding this to the API surface?

I worry about a future world in which we have MeshConfig, EnvoyFilter, and a first-class Telemetry API all for selecting tags for span annotation. But maybe that's the direction everything is headed?

kyessenov:
Maybe we need EnvoyFilter template inside MeshConfig for tracing spans? It sounds like we are heading that way.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,66 @@ message Tracing {
...
+  // The key represents the name of the tag.
+  // Ex:
+  // ```yaml
+  // custom_tags:
+  //   new_tag_name:
+  //     header:
+  //       name: custom-http-header-name
+  //       default_value: defaulted-value-from-custom-header
+  // ```
+  map<string, CustomTag> custom_tags = 5;
```
douglas-reid, line 157: Do we expect to add other runtime options here as well?  Should we go ahead and support the two other commonly requested options: url length and sampling? Or do we want to keep this narrow?

howardjohn:
If we have all these changes going on for telemetry API hacks
(extraStatsTag, this one, and baked in EnvoyFilters), seems like it might
be time for a Telemetry API? I am sure there's plenty of context that I am
missing behind this but there are a lot of telemetry related apis going on
right now

douglas-reid:
> If we have all these changes going on for telemetry API hacks (extraStatsTag, this one, and baked in EnvoyFilters), seems like it might be time for a Telemetry API? I am sure there's plenty of context that I am missing behind this but there are a lot of telemetry related apis going on right now

I would love to see a Telemetry API, fwiw. I think the issue is balancing it with the Extensions API proposal and the service config API proposal.

kyessenov:
@howardjohn We definitely need extension APIs that serves telemetry well. It must be open-ended (no content, just a wrapper for opaque configs). It's impossible to mirror all xDS here, it's just way too powerful to control. EnvoyFilter achieves that goal but it's not very usable, but it's fine as an implementation in absence of user-facing APIs.

mesh/v1alpha1/proxy.proto
kyessenov: Keeping narrow has a problem that someone will need it and will revert to EnvoyFilters making this moot.
Keeping it wide has a problem that we're at the mercy of xDS evolution.

bianpengyuan: So probably provide explicit options for several most commonly used ones, and also provide an opaque option to do xds override?

nrjpoddar: Sampling is already a mesh wide setting afaik. URL length I haven't seen but we can always add it later too if requested by users.

nrjpoddar:
> @nrjpoddar do we need a design doc for the UX here?
> 
> thinking out loud: if we expect a first class API to appear at some point, are we intro'ing more complexity here just in an attempt to avoid `EnvoyFilter`? If I want namespace-specific settings and that means `EnvoyFilter` in my namespace anyway (vs. one in `istio-system` or whatever), have we saved anything by adding this to the API surface?
> 
> I worry about a future world in which we have MeshConfig, EnvoyFilter, and a first-class Telemetry API all for selecting tags for span annotation. But maybe that's the direction everything is headed?

@douglas-reid the idea is to have meshconfig + annotations (which are considered second class APIs in Istio) until a first class API lands. I want incremental progress and not have more complications and the current customers we have will be ok with a global meshconfig setting. Next evolution is this would be annotations (namespace or workload level) if someone needs it. If not, we should go to the Telemetry API (with or without extensions).

I agree that there will be a release which will have MeshConfig/Annotations + Telemetry API + EnvoyFilter. At that point we will start deprecating MeshConfig/Annotations like we have done before. I know it's tactical but unblocks customers.

nrjpoddar:
@howardjohn you're preaching to the choir here about first class APIs, folks in Extensions WG have spent around 3-4 hours discussing it without much progress.

I would like to make progress on this tactical enhancement that unblocks users and provide a migration path when we arrive at the solution we all want.

mesh/v1alpha1/proxy.proto
nrjpoddar: @bianpengyuan we can always add more options if this doesn't meet user needs. Let's add this incremental improvement and get feedback.

I would not try to make this API all extensible and do everything for now, knowing that we want Telemetry API in future.

douglas-reid: @nrjpoddar url length is a long-standing request: https://github.com/istio/istio/issues/14563. we explicitly added support in Envoy for that. I think it makes sense to add it if we are putting any runtime config in MeshConfig.

I don't believe sampling is in meshconfig, right?  It is an env var in Pilot, or controlled via `EnvoyFilter` still, right? I'd add support adding it here.

douglas-reid:
> @douglas-reid the idea is to have meshconfig + annotations (which are considered second class APIs in Istio) until a first class API lands. I want incremental progress and not have more complications and the current customers we have will be ok with a global meshconfig setting. Next evolution of this would be annotations (namespace or workload level) if someone needs it. If not, we should go to the Telemetry API (with or without extensions).
> 
> I agree that there will be a release which will have MeshConfig/Annotations + Telemetry API + EnvoyFilter. At that point we will start deprecating MeshConfig/Annotations like we have done before. I know it's tactical but unblocks customers.

Should we think about hiding this from the API docs, but still adding it?  And have it as the break-glass way of setting things up but note explicitly that we don't intend to support it beyond the advent of a proper API?

nrjpoddar:
@douglas-reid I'm fine hiding it or marking it experimental.

mesh/v1alpha1/proxy.proto
nrjpoddar: sampling can be set at install time via: pilot.traceSampling

Makes sense, let's add this option too. Is it this option `max_path_tag_length` in Envoy?

jacob-delgado:
I've hidden `custom_tags` from docs.

@nrjpoddar and @douglas-reid Feedback for `max_path_tag_length` in this PR or should we get it put in a new one to track?

nrjpoddar:
> I've hidden `custom_tags` from docs.
> 
> @nrjpoddar and @douglas-reid Feedback for `max_path_tag_length` in this PR or should we get it put in a new one to track?

Let's just add it here, more API PRs more churns. Pick a good user facing name, add comments and let's get this PR merged.

jacob-delgado:
Added support for `max_path_tag_length` in Envoy. I kept it as a uint32 since that is what is defined at https://github.com/envoyproxy/envoy/blob/894832b70f0e72a3de7d30b9b2a136a6d2e1dea4/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L145.

Please review the comments. I've also hidden it from docs as well.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,77 @@ message Tracing {
     // Use a Stackdriver tracer.
     Stackdriver stackdriver = 4;
   }
+
+  // Configure custom tags that will added to any active span.
```
louiscryan, line 103: s/will/will be

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,77 @@ message Tracing {
...
+  // Tags can be generated via literals, environment variables or an incoming request header.
+  // $hide_from_docs
+  message CustomTag {
+    // Specify how to populate the value in a custom tag
+    oneof type {
+      // The custom tag's value is the specified literal.
+      Literal literal = 1;
```
louiscryan, line 110: @kyessenov thoughts about using CEL expressions to generate the K,V map? CEL supports the literal and  header expressions? In theory it could have handled path truncation too.

kyessenov: It could be implemented but hasn't been done so. There is a [header format string](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/base.proto#envoy-api-msg-core-headervalue) support that is not exposed in this API, which is similar.

mesh/v1alpha1/proxy.proto
rshriram: lets add sampling please. pilot env var is a crude hack. If anything, people want per namespace customization for sampling.

nrjpoddar: @rshriram we are not adding per namespace overrides as part of this PR. MeshConfig vs Helm option for sampling won’t make make of a difference.

nrjpoddar: We can add it here if it’s better UX but it won’t solve your use case.

douglas-reid: Agree that it won't solve namespace overrides. Which is why the prior proposals were for Sidecar-driven API chanes. But in the short-term, since we are already adding runtime config to meshconfig, I think it makes sense to replace the Pilot env var with a setting in mesh config.

jacob-delgado:
cc @nrjpoddar @rshriram @douglas-reid 

Added `client_sampling`, `random_sampling` and `overall_sampling` from [Envoy proto](https://github.com/envoyproxy/envoy/blob/f84440dc4f95890f14e2e0686b07258f030b54b3/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L119).  Percentage is a double defined [here](https://github.com/envoyproxy/envoy/blob/f84440dc4f95890f14e2e0686b07258f030b54b3/api/envoy/type/v3/percent.proto#L20).

I can remove `client_sampling` and `overall_sampling` as that is not user-configurable when I was browsing the code in pilot found [here](https://github.com/istio/istio/blob/4f8c15f2829344ad41daa86f33bc737b3c00065b/pilot/pkg/security/model/trace.go#L46:6). Let me know if you guys think that is too many knobs as it's not something people can configure now.

mesh/v1alpha1/proxy.proto
jacob-delgado: Added sampling variables. I may have added too much configurability. Let me know what you think. Pilot only lets users configure `random_sampling`.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,93 @@ message Tracing {
...
+  // $hide_from_docs
+  map<string, CustomTag> custom_tags = 5;
+
+  // Configures the maximum length of the request path to extract and include in the
+  // HttpUrl tag. Used to truncate length request paths to meet the needs of tracing
+  // backend. If not set, then a length of 256 will be used.
+  // $hide_from_docs
+  uint32 max_path_tag_length = 6;
+
+  // The percentage (0.0 - 100.0) of requests manged by the HTTP connection manager
```
jacob-delgado, line 170: Need review over variables and comments.

douglas-reid:
> cc @nrjpoddar @rshriram @douglas-reid
> 
> Added `client_sampling`, `random_sampling` and `overall_sampling` from [Envoy proto](https://github.com/envoyproxy/envoy/blob/f84440dc4f95890f14e2e0686b07258f030b54b3/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L119). Percentage is a double defined [here](https://github.com/envoyproxy/envoy/blob/f84440dc4f95890f14e2e0686b07258f030b54b3/api/envoy/type/v3/percent.proto#L20).
> 
> I can remove `client_sampling` and `overall_sampling` as that is not user-configurable in istio, found when I was browsing the code in pilot found [here](https://github.com/istio/istio/blob/4f8c15f2829344ad41daa86f33bc737b3c00065b/pilot/pkg/security/model/trace.go#L46:6). Let me know if you guys think that is too many knobs as it's not something people can configure now.

I'd be fine with just `random_sampling` and calling it `sampling` to keep the surface small and digestible for MeshConfig. When we transition to a first-class API, we can provide more detailed samplers, etc.

mesh/v1alpha1/proxy.proto
nrjpoddar: let's just expose whatever pilot.traceSampling is instead of adding all the options.

nrjpoddar:
> > cc @nrjpoddar @rshriram @douglas-reid
> > Added `client_sampling`, `random_sampling` and `overall_sampling` from [Envoy proto](https://github.com/envoyproxy/envoy/blob/f84440dc4f95890f14e2e0686b07258f030b54b3/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L119). Percentage is a double defined [here](https://github.com/envoyproxy/envoy/blob/f84440dc4f95890f14e2e0686b07258f030b54b3/api/envoy/type/v3/percent.proto#L20).
> > I can remove `client_sampling` and `overall_sampling` as that is not user-configurable in istio, found when I was browsing the code in pilot found [here](https://github.com/istio/istio/blob/4f8c15f2829344ad41daa86f33bc737b3c00065b/pilot/pkg/security/model/trace.go#L46:6). Let me know if you guys think that is too many knobs as it's not something people can configure now.
> 
> I'd be fine with just `random_sampling` and calling it `sampling` to keep the surface small and digestible for MeshConfig. When we transition to a first-class API, we can provide more detailed samplers, etc.

Yes, that's what I was suggesting. Let's not make it too complex. What Doug makes perfect sense.

mesh/v1alpha1/proxy.proto
```diff
@@ -99,6 +99,82 @@ message Tracing {
...
+  // Ex:
+  // ```yaml
+  // custom_tags:
+  //   new_tag_name:
+  //     header:
+  //       name: custom-http-header-name
+  //       default_value: defaulted-value-from-custom-header
+  // ```
+  // $hide_from_docs
+  map<string, CustomTag> custom_tags = 5;
```
jacob-delgado, line 162: @nrjpoddar One possible issue I've found when implementing is that order cannot be preserved since maps are not ordered when iterating over it versus using a list. I don't know if that presents an issue. But it will mean that the order listed won't be listed in the same order in the configuration of the http connection manager filter.

nrjpoddar: Just sort it after reading it.

jacob-delgado: If the user specifies this in the meshconfig.defaultconfig

```yaml
 custom_tags:
     new_tag_name:
       header:
         name: custom-http-header-name
         default_value: defaulted-value-from-custom-header
     first_tag_name:
       header:
         name: custom-http-header-name-2
         default_value: defaulted-value-from-custom-header-2
```

nothing can be done to make sure its in the order the user listed order. Doing what you suggested would have the filter come out to

```yaml
            custom_tags:
            - tag: first_tag_name
              request_header:
                  name: custom-http-header-name-2
                  default_value: defaulted-value-from-custom-header-2
            - tag: new_tag_name
              request_header:
                  name: custom-http-header-name
                  default_value: defaulted-value-from-custom-header
```

nrjpoddar: We just want stable Envoy filter serialization. We use maps in lots of places in other APIS, I don’t think it should be an issue.

kyessenov:
Now that tracing driver is dynamically configurable via xDS, where should it go in your opinion?

nrjpoddar:
Can we get this merged? We have hashed out all the review items and need to deliver this in 1.6.

jacob-delgado:
Any updates? I am working on the implementation of this feature and I'd like to get it merged before code freeze next week.

jacob-delgado:
> Now that tracing driver is dynamically configurable via xDS, where should it go in your opinion?

For 1.6 we will probably delay this addition. I'm not sure if it also belongs in MeshConfig or if we should wait for the telemetry API people are working on. It should be revisited when we are planning for istio 1.7.

shwsingh:
@jacob-delgado are you still waiting for TOC approval

PR #1378, opened 2020-04-13
Add short alias for authorizationpolicy resource

rolandkool:
Updated/new PR for https://github.com/istio/istio/issues/19292

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -6219,6 +6219,8 @@ spec:
     kind: AuthorizationPolicy
     listKind: AuthorizationPolicyList
     plural: authorizationpolicies
+    shortNames:
+    - ap
     singular: authorizationpolicy
   scope: Namespaced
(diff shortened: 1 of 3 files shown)


stewartbutler:
@rolandkool Can you add a bit explaining why this PR is important to you? Some additional context regarding the request would help in getting the necessary approvals. Thanks!

rolandkool:
Original motivation was because of multiple CRDs with similar name making tabcompletion useless. This has been fixed. However, other Istio resources that are used often also have a shortname, so for convenience, I would like to see one for authorizationpolicies as well. Networkpolicies also has a shortname (`netpol`), so if `ap` is too short, something like `authzpol` or `authpol` will work as well.

peterj:
@rolandkool any plans to rebase and get this merged? 

It's frustrating typing out authorizationpolicy each time when interacting with the resource. Shortname `ap` would be awesome. The other CRDs (`peerauthentication` and `requestauthentication`) have short names set to `pa` and `ra`.

PR #1380, opened 2020-04-14
Create enum for inboundClusterStatName and outboundClusterStatName

shamsher31:
As per the discussion here https://github.com/istio/istio/pull/22799#discussion_r407593905, we need enum values for `inboundClusterStatName` and `outboundClusterStatName`. This will be helpful for generating possible options that the user can pass when they are trying to set `inboundClusterStatName` and `outboundClusterStatName`.
```
eg: %SERVICE_FQDN%_%SERVICE_PORT% 
```

Note that the `outboundClusterStatName` also includes all the options from `inboundClusterStatName` plus `SUBSET_NAME`.


mesh/v1alpha1/config.proto
```diff
@@ -374,6 +390,15 @@ message MeshConfig {
   // - `%SERVICE%` will use reviews.prod as the stats name.
   string inbound_cluster_stat_name = 44;
```
howardjohn, line 391: I am not sure how this will be used. Its not referenced anywhere in the proto. `inbound_cluster_stat_name` does not take an input like `SERVICE_FQDN` it takes in a string like `foo-%SERVICE_FQDN%`

shamsher31: I agree, From the [docs of MeshConfig](https://preliminary.istio.io/docs/reference/config/istio.mesh.v1alpha1.html#MeshConfig-inbound_cluster_stat_name), as per the supported pattern,  I run the test locally and I can see the following works fine.

```cli
$ istioctl manifest apply --set \
 values.meshConfig.inboundClusterStatName="%SERVICE_FQDN%_%SERVICE_PORT%"
```

shamsher31:
Any feedback from the team.

PR #1383, opened 2020-04-14
Expanding cluster-local support to include specific services.

nmittler:
Also adding support for future host-based attributes.


nmittler:
@howardjohn I expanded the comment ... PTAL

nmittler:
Not sure why gencheck is failing ... I ran `make gen` with `BUILD_WITH_CONTAINER=1`

networking/v1beta1/destination_rule.proto
```diff
@@ -516,6 +516,28 @@ message LoadBalancerSettings {
...
+  //     having services cluster-local and then slowly transition them to mesh-wide. They could do
+  //     this service-by-service (e.g. mysvc.myns.svc.cluster.local) or as a group
+  //     (e.g. *.myns.svc.cluster.local).
+  //
+  // Note that if a destination is cluster-local, the LocalityLoadBalancer settings will only apply
+  // if the cluster spans multiple availability zones.
+  //
+  // By default, Istio will consider all services in the kube-system namespace to be cluster-local,
+  // unless explicitly overridden here.
+  bool cluster_local = 4;
```
rshriram, line 540: Is this a service level property across an entire cluster or is this a per-client preference for a given service? If it is latter, dest rule is okay. But the explanation seems like it is for former. In that case, it is better suited in service entry and annotations for k8s services to indicate that they are local to a cluster

nmittler: It is a service-level property across the entire cluster, but we want a way to be able to configure a single service or a group of services at once. Service/ServiceEntry wouldn't give us quite the flexibility that we're looking for here.

howardjohn: The logical followup here is what happens if in NS A I say `foo.bar` is cluster local and in NS B I say `foo.bar` is not cluster local. Now its no longer across the entire cluster. This seems perfectly fine to me, but does counterdict what you said in a way

hzxuzhonghu: Another choice is to expand the ExportTo field.

howardjohn: That would require the client having access to the Service they are calling, which is often not the case. For example, you probably won't convince the cluster admin to add some istio annotation to kube-dns service. In some environment this isn't even possible

hzxuzhonghu: Make sense

costinm:
I understand the use case - but what is the general case here ? This seems
to be about a filter on the EDS response
and load balancing selection - where the 'cluster' label is matched against
the client.

Should we expose the general case  in the API ? We do a lot of label
matching in Istio and K8S - and
we can treat 'cluster' and 'network' as any other label.

So we could add a list of label keys that will need to match in order for
an endpoint to be used  -
with 'cluster' as an example. I assume 'network' will be the next
important one - for example
in stateful sets. And 'canary' or other labels will also be nice to match (
so a canary workload will
only get endpoints for other canary service endpoints ).

Regarding DestinationRule vs. Sidecar - I think it can go in either. I
don't know what's the current
plan on deprecating DR - if we're serious about it maybe we shouldn't add
more. IMO we should
just keep it, so I don't mind.

nmittler:
@costinm as John described in https://github.com/istio/api/pull/1383#pullrequestreview-394171077, I'm not sure labels are really an option here, since the client may not have permission to write to the service or namespace.

nmittler:
@howardjohn @louiscryan @rshriram @costinm 

I just chatted with Costin and he clarified his idea. His ask was to see if we could make this more generic so that it could also address matching network and possibly other things in the future.  So the api would look something like this:

```proto
// Match provides a mechanism to filter destination endpoints from the
// client. If specified, requires that the attributes in the client and destination
// must match in order to include the destination in the client's outbound
// endpoints.
//
// The only values currently supported are: "cluster".
//
// "cluster":
//
//   Specifies that the destination should only be accessible from within the same
//   cluster as the client.
//
//   By default, Istio assumes a service endpoint to be accessible from any cluster
//   in the mesh. By specifying this value, only upstream destinations within the
//   same cluster will be used for the host. Cluster-local destinations will not be
//   visible to workloads residing in other clusters.
//
//   There are some common scenarios when this can be useful:
//
//     - A service (or group of services) is inherently local to the cluster and has local
//       storage for that cluster. For example, the kube-system namespace (e.g. the
//       Kube API Server).
//     - A mesh administrator wants to slowly migrate services to Istio. They might
//        start by first having services cluster-local and then slowly transition them to
//        mesh-wide. They could do this service-by-service
//        (e.g. mysvc.myns.svc.cluster.local) or as a group (e.g. *.myns.svc.cluster.local).
//
//   Note that if a destination is cluster-local, the LocalityLoadBalancer settings will
//   only apply if the cluster spans multiple availability zones.
//
//   By default, Istio will consider all services in the kube-system namespace to be
//   cluster-local, unless explicitly overridden here.
repeated string match;
```

WDYT?

howardjohn:
@nmittler I like the idea in a vacuum, how will it work when we have the k8s service topology? That API adds something somewhat similar set of configuration. I think this is not conflict with that implementation at all, and they would likely share the same/similar implementation internally. Does that sound right?

nmittler:
@howardjohn 

> @nmittler I like the idea in a vacuum, how will it work when we have the k8s service topology? That API adds something somewhat similar set of configuration. I think this is not conflict with that implementation at all, and they would likely share the same/similar implementation internally. Does that sound right?

Yeah, that sounds right. The one thing I would add is that while k8s service topology is on a per-service basis, having this option in `DestinationRule` allows us more flexibility in that we can configure multiple destinations as once, and it also allows us to override the service topology on a per-client basis. So I see the two options as working together.

networking/v1beta1/destination_rule.proto
rshriram: Why is it not possible ? We have a global defaultServiceExportTo value. We can add exceptions or explicit config for kube system

nmittler:
@rshriram 

> At this point, you just need per service field - you can use exportTo or add another annotation which excludes the service.

Again, per-service fields aren't an option if you don't have control over the service (i.e. kube-system)

nrjpoddar:
@nmittler do you have a RFC for this, if you do can you point me to it?

Since it’s an API change we need to go through the same motions as other API changes are going through and get approval for WG leads. I understand the use case but it looks like there’s no prior agreement.

nmittler:
I've just created an RFC: https://docs.google.com/document/d/1qfRpp7ujfi4qa77tqbxcAMOK6E8h-_xLXFcbvWhABCg/edit#

This needs to get into 1.6 before existing MeshConfig-based API is released. I'll add it to the agenda for today's TOC meeting.

howardjohn:
One thing that I think isn't clear in this is that this intends to replace
the meshconfig setting, not be an additional setting

nmittler:
@howardjohn I've updated the objective in the doc to say exactly this.

nrjpoddar:
> I've just created an RFC: https://docs.google.com/document/d/1qfRpp7ujfi4qa77tqbxcAMOK6E8h-_xLXFcbvWhABCg/edit#
> 
> This needs to get into 1.6 before existing MeshConfig-based API is released. I'll add it to the agenda for today's TOC meeting.

Thanks @nmittler for creating the RFC!

nmittler:
@costinm @nrjpoddar @rshriram @howardjohn I've updated the PR to use MeshConfig. PTAL.

mesh/v1alpha1/config.proto
```diff
@@ -409,13 +409,49 @@ message MeshConfig {
...
+
+  // Match specifies a number of attributes that must match between the client and the service
+  // endpoints in order for the endpoints to be callable from the client.
+  message Match {
+    // Specify the name of the service to which the matches apply. If not specified or "*", the
+    // match applies to the endpoints for any service in the namespace.
+    string service_name = 1;
+
+    // Specify the namespace of the service.
+    string namespace = 2;
```
nrjpoddar, line 446: is "*" supported here?

nrjpoddar: namespace is mandatory field correct?

mesh/v1alpha1/config.proto
```diff
@@ -409,13 +409,49 @@ message MeshConfig {
...
+    // unless explicitly overridden here.
+    CLUSTER = 0;
+  }
+
+  // Match specifies a number of attributes that must match between the client and the service
+  // endpoints in order for the endpoints to be callable from the client.
+  message Match {
+    // Specify the name of the service to which the matches apply. If not specified or "*", the
+    // match applies to the endpoints for any service in the namespace.
+    string service_name = 1;
```
nrjpoddar, line 443: should this be an array to indicate multiple service names within a namespace?

mesh/v1alpha1/config.proto
```diff
@@ -409,13 +409,49 @@ message MeshConfig {
...
+    //     having services cluster-local and then slowly transition them to mesh-wide. They could do
+    //     this service-by-service (e.g. mysvc.myns.svc.cluster.local) or as a group
+    //     (e.g. *.myns.svc.cluster.local).
+    //
+    // Note that if a destination is cluster-local, the LocalityLoadBalancer settings will only apply
+    // if the cluster spans multiple availability zones.
+    //
+    // By default, Istio will consider all services in the kube-system namespace to be cluster-local,
+    // unless explicitly overridden here.
+    CLUSTER = 0;
```
costinm, line 435: Any reason to be static ? I think a list of strings is better - we will need to document the subset of attributes we support, but no need to make it enum - user could match on his own defined attributes ( like labels )

mesh/v1alpha1/config.proto
```diff
@@ -409,13 +409,49 @@ message MeshConfig {
...
+  message Match {
+    // Specify the name of the service to which the matches apply. If not specified or "*", the
+    // match applies to the endpoints for any service in the namespace.
+    string service_name = 1;
+
+    // Specify the namespace of the service.
+    string namespace = 2;
+
+    // Specify the set of attributes required to match for the endpoints of this service.
+    repeated MatchAttribute attributes = 3;
```
costinm, line 449: This is way to generic name for a narrow feature. 

endpoint_selection_rules ? 
endpoint_rules ?

mesh/v1alpha1/config.proto
```diff
@@ -409,13 +409,49 @@ message MeshConfig {
...
-  // Set the namespaces whose services will be considered local to the cluster. Cluster
-  // local services will not be exported to and will, therefore, not be accessible by
-  // workloads residing in other clusters. A common use case is the Kube API Server, where
-  // workloads expect to always reach the in-cluster server. By default, the namespace
-  // "kube-system" will be cluster-local.
-  repeated string cluster_local_namespaces = 50;
+  // MatchAttribute is an enumeration of attributes that can be used to a comparison between
+  // a client and a service endpoint.
+  enum MatchAttribute {
+    // Custer specifies that the client and endpoint cluster must match. By default,
```
nrjpoddar, line 415: Nit: Custer -> Cluster

We should try and make this documentation more clear, here's my suggestion (feel free to ignore):

// Cluster match attribute specifies that the client and service endpoints must reside in the same cluster. By default,
// in multi-cluster deployments, the Istio control plane assumes all service endpoints to be reachable from any
// client in any of the clusters which are part of the mesh. This configuration option limits the set of service endpoints visible to a client to be cluster scoped.

There's a lot of different terms host/services used interchangeably which will confuse users.

mesh/v1alpha1/config.proto
```diff
@@ -409,13 +409,49 @@ message MeshConfig {
...
+    //   - A mesh administrator wants to slowly migrate services to Istio. They might start by first
+    //     having services cluster-local and then slowly transition them to mesh-wide. They could do
+    //     this service-by-service (e.g. mysvc.myns.svc.cluster.local) or as a group
+    //     (e.g. *.myns.svc.cluster.local).
+    //
+    // Note that if a destination is cluster-local, the LocalityLoadBalancer settings will only apply
+    // if the cluster spans multiple availability zones.
+    //
+    // By default, Istio will consider all services in the kube-system namespace to be cluster-local,
+    // unless explicitly overridden here.
```
nrjpoddar, line 434: How do you override "kube-system"?

mesh/v1alpha1/config.proto
costinm, line 443: I think fully qualified name is better - we'll support 'bring your name', etc.

This is really matching the cluster name - in the new scheme of naming.

I'm not entirely sure what's the best way to select here - there is EnvoyFilter schema as well, the new naming that @louiscryan  is discussing.

nrjpoddar: +1 I was thinking the same thing as essentially this ends up enforced on the client side and most client-side config in Istio is specified via FQDN (via host fields)

costinm: How about using ClusterMatch here ( reusing the one from EnvoyFilter ) ? 

The selection you want is really specific to clusters, it allows for subset names to be included - and as EnvoyFilter evolves into the new schema, we'll replace/migrate this as well.

mesh/v1alpha1/config.proto
nmittler: > Any reason to be static ? I think a list of strings is better - we will need to document the subset of attributes we support, but no need to make it enum

It's static for the purposes of clarity in the API. Why would we *not* make it an enum?

> user could match on his own defined attributes ( like labels )

If we supported labels, we'd need a way to make that explicit. I'm not tackling that in this PR. Cluster, for example, is an internal attribute of sorts, not a real label.

mesh/v1alpha1/config.proto
nmittler: Namespace is required, correct.

mesh/v1alpha1/config.proto
nrjpoddar: EndpointLocalityMatch?

mesh/v1alpha1/config.proto
nmittler: Ah  maybe. That would solve the wildcard problem as well (i.e. empty list). Let me switch to that.

mesh/v1alpha1/config.proto
nmittler: Ah you can't, anymore. Will fix.

mesh/v1alpha1/config.proto
nmittler: I think we have to leave more clever matching to when (if) this feature is promoted to a top-level API (e.g DestinationRule)

howardjohn:
Sorry I realize I agreed to this in TOC but I completely misunderstood what I agreed to. In my opinion this does not make any sense. We are debating here 5 different ways to match a service, and how to handle things like matching multiple services, wildcards, wildcard namespaces, subsets, etc. We already have an API to do exactly that - DestinationRule!

We need to be able to add fields to the right location regardless of the stability. It seems very clear in my mind that DestinationRule is the correct place for this*. Based on that, I think we should be finding a way to put an unstable/unproven field into an existing CR. Possible options are to mark the field explicitly as experimental/alpha/hidden, or even use an annotation on the DR (I don't think we should do annotation, just an idea).

\* all of load balancing settings already exist here, it has the semantics we need for flexibility like wildcards and subsets. Mesh config has never been used to select services like this before either. If we go down this route, why even have CRDs - we can just have one huge mesh config that defines the entire configuration for the mesh.

nmittler:
@nrjpoddar @costinm @rshriram mind replying to John's comment ^^?

mesh/v1alpha1/config.proto
costinm: > > Any reason to be static ? I think a list of strings is better - we will need to document the subset of attributes we support, but no need to make it enum
> 
> It's static for the purposes of clarity in the API. Why would we _not_ make it an enum?
> 
> > user could match on his own defined attributes ( like labels )
> 
> If we supported labels, we'd need a way to make that explicit. I'm not tackling that in this PR. Cluster, for example, is an internal attribute of sorts, not a real label.

Both telemetry and Security define a number of attributes - and neither is using enums. 
I don't think our attributes should use different names or style.

mesh/v1alpha1/config.proto
costinm: It's not locality related - the general form is that it's a filter on the returned endpoints. Can be any attribute - for example 'return only canary endpoints'

mesh/v1alpha1/config.proto
costinm: Bring your own name is supported - and should remain supported. 
So namespace should be optional ( and maybe even not included ), and service_name  qualified. 

But I think we must converge with Louis naming system - and EnvoyFilter matching.

mesh/v1alpha1/config.proto
costinm: It's not really more clever matching - but more consistent matching API.

costinm:
Yes, I 100% agree with John. But we don't have a way to add 'experimental' stuff - and as proven by this discussion, defining a good API is hard. 

That's where annotations were helping. It is clear we are going to modify the naming schema and matching in filters and extensibility. It is clear this proposal is non consistent with the rest of the Istio APIs (if we define another kind of 'attribute' with enum name - while security/telemetry have other representation). Yet it's very useful to move ahead to get feedback from users. 

Even worse: it is very likely K8S APIs would be sufficient and we could just use them.

howardjohn:
>  But we don't have a way to add 'experimental' stuff - and as proven by this discussion, defining a good API is hard.

That is my main concern here. We need a way to do this. For an example, k8s recently added appprotocol field to Service which had similar concerns here. They didn't add some global config where you define a mapping of service ports to protocol. Instead, they marked the field as alpha, despite it being in a v1 stable API, and even put a feature flag behind it. Can we do something similar here?

nrjpoddar:
Here are my recommendations:

* I looked at all the options at MeshConfig, which are too many but none of them try to do in-cluster service filtering like the proposal in this API. If we are keeping anything in MeshConfig let's just keep a list of namespaces to be excluded or if you need service based control replace the list of namespaces to be list of FQDNs without more complex matching logic.
* I don't think any user needs this arbitrary match based API just yet, so we should think more about where this should be housed and the right semantics. Given the evolution of K8s APIs and we have matured DestinationRules to v1beta, any additions in the API will need to follow 3-6 months of deprecation strategy.

tl; dr
Keep a list of namespaces or FQDNs in meshconfig for exclusion, anything else requires more thought and agreement.

costinm:
I like Neeraj's proposal for short term - a list of services to use
'cluster local' is a good start, and consistent with other places where we
use an inclusion/exclusion list.
And it's super easy to type for users - compared with a deep structure.
I still think a generic solution that can address future needs is better
than a different API for each use case. It's a common problem to have to
exclude/include specific K8S objects
from Istio features - when we can't annotate the service or is too narrow
case for bloating our API.

So how about:
  MeshConfig {
    Overrides { // or featureOverrides, etc
      "EndpointClusterFilter":  [ "apiserver.kube-system",
"productpage.bookinfo", ... ]
      "...": [.... list of names
   }
  }
We have a bunch of other whitelists we probably lost (citadel, watched
namespaces, etc)


I also like John's proposal - if K8S has a well-defined way to add
'experimental' fields to stable APIs - we should do the same.
However: I don't think this is an excuse  to add things like we did in
global.yaml and values.yaml, just because it's marked experimental it can go
without future consideration.

I don't think that just because we have a way to make them experimental -
we can keep churning unstable APIs just to avoid doing the harder work of
properly considering the big picture and future needs.


Good API takes time.

costinm:
After more thinking - I think we should do both !

On Neeraj's proposal - extended a bit: in mesh config we want to let the
admin to do 'overrides' per service, namespace or workload level, without
changes in K8S Services. There are  many use cases:
- ignoring namespaces, or selectively including namesapaces to injection or
watching.
- new features
- existing features where we want to let the admin have ultimate control
- merging of global/values.yaml into mesh config - there are many variants
of this pattern.

We have a spaghetti of APIs in mesh config, global.yaml, values.yaml that
do this in very different ways. We still need to
merge global.yaml/values.yaml into mesh config - so if we can define a
consistent representation of per/svc,workload mesh
configuration we can make everything consistent. Including Mandar's
telemetry filters.

On John's proposal: we can add a repeated 'experimental' Any to top level
APIs (ServiceEntry, Gateway, VirtualService, Sidecar). Instead
of one-offs for each new feature - we can say that the same proto will be
used in both mesh config and the actual resource.
We document that 'experimental' is for new APIs, equivalent or less stable
than EnvoyFilter - no need to do it per-field.

The code can mechanically merge the 'overrides' from mesh config with the
'experimental' from the APIs. Anything per-host
will go to ServiceEntry, per-route to VirtualService and per-workload to
Sidecar. To top it off, we can make annotations consistent as well -
each annotation will also use same proto for each experimental feature, and
same rule - Service annotations will be merged to
ServiceEntry or VirtualService, Pod/Deployment annotations to Sidecar.

This combination will end all discussions on where should a new feature go
- the answer will be 'everywhere'. We still need
to define a .proto with the fields specific to the feature - and that will
be the only review we need to focus on.
Mesh admin can start with enabling the features in mesh.config - specific
to a CP revision, and for select namespaces/workloads
and eventually mesh wide.  If they are happy, they can start using the
'experimental' in the APIs or annotations.
And after 1-2 releases we can decide to move them into the proper place, if
they have sufficient usage. I suspect we can
get stats from vendors.

Back to the topic of this feature:
1. we can either treat it as a narrow 'cluster local' feature, in which
case original proposal (a boolean) is sufficient.
For any 'boolean' - or primitive - experimental feature we should use a
simple, consistent syntax:
   DestinationRule:
       - experimental:
            - cluster-local: true

2. or we can treat it as a 'filtering' feature, in which case a proto will
need to be defined - consistent with other attribute matching/filtering
we do in security, networking and telemetry.

I'm ok with the first as 'experimental' - as long as there is no plan to
promote it to 'stable' until we have the consistent filtering.

Note that I used ServiceEntry instead of DestinationRule: both apply to
hosts. We had many discussions on changing DR, so
maybe we should attach 'experimental' features to SE.

costinm:
Adding an API for each use case leads to global.yaml and values.yaml, i.e.
an inconsistent mess . Take a look at the 100s of options - each is has a
use case just like this.

Excluding a list of namespaces from cluster.local will fix this user case.
But as an API - Neeraj proposal to have a generic way to exclude namespaces
or services from _any_ feature consistently seems better.

Agrred that kube-system should just be off by default.

Nate - how does it even work, if strict mode is enabled you wouldn't be
able to connect to it. I thought we already exclude it.

nrjpoddar:
As much as I like all the debates, the comments on this PR are overwhelming and not giving clear direction to a a dev (@nmittler in this case), who wants to make progress and create a useful features for users.

I'm again going to suggest this for short term purposes:
* kube-system should be by default excluded (and is afaict)
* Service level control is needed for 2 reasons:
  + Multiple services within a namespace might have different behaviors so a namespace level default doesn't work.
  + Service annotations doesn't work for "system" level services where you might not have edit permissions.
* If service level control is needed:
  + Make the exclusion field a FQDN list instead of namespace list (with wildcards and kube-system excluded) and that will meet all the requirements afaict.

nmittler:
@nrjpoddar @costinm @howardjohn @rshriram PTAL

Given that TOC only approved changes to meshconfig, I'm keeping the changes there for now. DestinationRule or any other CR is not really an option for 1.6, since it would require another round of TOC approval. The same thing is true of service/namespace annotations.

I've modified the structure to be more future friendly by accommodating various attributes and basing it on host (rather than namespace or service name), which will hopefully make it easier to port these changes to DestinationRule in the future.

mesh/v1alpha1/config.proto
```diff
@@ -410,12 +410,38 @@ message MeshConfig {
...
+    message ClusterLocal {
+      // If true, specifies that the client and service endpoints must reside in the same cluster
+      bool value = 1;
+    }
+
+    // Specifies whether or not the client and service endpoints must reside in the same cluster.
+    ClusterLocal cluster_local = 1;
+  }
```
costinm, line 439: if it's just a boolean - why add a message ? The yaml syntax will be pretty bad:

 serviceOverrides:
  - foo.default.cluster.local:
     cluster_local:
       value: false

nrjpoddar: To be honest, the inverted UX is not so awesome. By inverted what I mean is this is keeping services first in API hierarchy and then overrides based on each feature.

This API requires a lot of repetition if we only have 1 or 2 features and lots of services that have to provide overrides.

The other way to think about this feature first -> map to FQDNs. This model fits more cleanly with how other "feature" oriented APIs are designed in Istio.

nmittler: @costinm 

> if it's just a boolean - why add a message ? The yaml syntax will be pretty bad:

proto3 doesn't have the notion of unset anymore, so we need the ability to have this field be unspecified. That's why I wrapped it in a message, so it can be nil. I'm open to other ideas.

@nrjpoddar:

> To be honest, the inverted UX is not so awesome. By inverted what I mean is this is keeping services first in API hierarchy and then overrides based on each feature.

The original thought was to simplify the eventual transition of this to DestinationRule, hence host is the key. But I'm open to alternative data structures. I'm not sure if we should care too much about repetition here, however, since in practice it's unlikely that anyone will configure more than one or two of these.

nmittler: @nrjpoddar so let me see if I understand your ask.  Would something like this be better in your mind?

```proto
  message ServiceOverrides {
    message Attributes {
      bool cluster_local = 1;
    }

    Attributes attributes = 1;
    repeated string host = 2;
  }

  repeated ServiceOverrides service_overrides = 50;
```

nrjpoddar: yes!

I'm sure others will have their preferences but for me it's very clear to a user what they are trying to do. If you write out YAML for this vs the other option you will see the difference.

nmittler: sgtm ... will change. Thanks!

nmittler: @nrjpoddar PTAL

nmittler: Also @costinm  PTAL

mesh/v1alpha1/config.proto
```diff
@@ -410,12 +410,41 @@ message MeshConfig {
...
+    Attributes attributes = 1;
+
+    // The hosts to which the Attributes should be applied. Supported host names are identical to
+    // that of DestinationRule.
+    repeated string hosts = 2;
+  }
+
+  // $hide_from_docs
+  // Various overrides for service-level attributes.
+  repeated ServiceOverrides service_overrides = 50;
```
nrjpoddar, line 447: Is ServiceOverrides the right name? I don't really have a better suggestion for you since the name fits this use case. 
I'm not sure if the name will work in general without more concrete use cases in front of me.

nmittler: Ack. Same for me ... I don't love it.

louiscryan: What do we think would be other per-service settings of a similar ilk?

I can live with settings/options

service_options
service_config
service_overrides
service_settings

mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
```diff
@@ -451,13 +450,40 @@
           }
         }
       },
+      "istio.mesh.v1alpha1.MeshConfig.ServiceOverrides": {
+        "description": "A set of attributes that override the default behavior for a set of selected services.",
```
louiscryan, line 454: Not everything that goes in here will be an override. Can always name the field

cluster_local_override

but I'm not even sure that is necessary if the doc for the field is good

mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
```diff
@@ -451,13 +450,40 @@
...
+        "type": "object",
+        "properties": {
+          "hosts": {
+            "description": "The hosts to which the Attributes should be applied. Supported host names are identical to that of DestinationRule.",
```
louiscryan, line 458: The services to which the Settings should be applied. Services are selected using the hostname matching  rules used by DestinationRule

ideally would link to that doc or provide an example inline

mesh/v1alpha1/config.proto
nmittler: Something like `ServiceSettings` seems fine.  WDYT @costinm @nrjpoddar ?

mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
nmittler: dropping "override" throughout. I agree, the comments should be sufficient.

mesh/v1alpha1/config.proto
nrjpoddar: Yes, overrides is too restrictive and places us in a corner as it's tied to implementation.

Settings or Config both work for me.

nmittler:
@nrjpoddar @louiscryan PTAL

mesh/v1alpha1/config.proto
```diff
@@ -410,12 +410,43 @@ message MeshConfig {
...
+    // The services to which the Settings should be applied. Services are selected using the hostname
+    // matching rules used by DestinationRule.
+    //
+    // For example: foo.bar.svc.cluster.local, *.baz.svc.cluster.local
+    repeated string hosts = 2;
+  }
+
+  // $hide_from_docs
+  // Settings to be applied to select services.
+  repeated ServiceSettings service_settings = 50;
```
nrjpoddar, line 449: I would add an example of how to make few services cluster_local in docs.

```yaml
serviceSettings:
  - settings:
       cluster_local: true
       hosts:
       - *.foo.svc.cluster.local
       - bar.baz.default.svc.cluster.local
```
In this example, all services in the namespace foo and service bar in namespace baz are considered local.

Something like that, I might have got the YAML wrong.

After writing out the example, I still don't like the naming schemas settings nested inside of ServiceSettings doesn't feel right.

mesh/v1alpha1/config.proto
```diff
@@ -410,12 +410,43 @@ message MeshConfig {
...
+    }
+
+    // The settings to apply to the selected services.
+    Settings settings = 1;
+
+    // The services to which the Settings should be applied. Services are selected using the hostname
+    // matching rules used by DestinationRule.
+    //
+    // For example: foo.bar.svc.cluster.local, *.baz.svc.cluster.local
+    repeated string hosts = 2;
```
hzxuzhonghu, line 444: From the implementing view, this will need to parse namespace. How can we make sure the host is of the format we want

mesh/v1alpha1/config.proto
nmittler: Yeah I tend to agree. Just looking at our docs for `DestinationRule`, the first sentence of the description is `defines policies that apply to traffic intended for a service after routing has occurred`.

It sure feels like we're defining the same thing here. Would there ever be a case where something in this structure would *not* eventually belong in DestinationRule? If not, I wonder if we should just be more explicit and say something like `DestinationRuleOverrides`.

mesh/v1alpha1/config.proto
nmittler: Why would it be any different than how we handle DestinationRule?

mesh/v1alpha1/config.proto
```diff
@@ -410,12 +410,43 @@ message MeshConfig {
...
+      // to be cluster scoped.
+      //
+      // There are some common scenarios when this can be useful:
+      //
+      //   - A service (or group of services) is inherently local to the cluster and has local storage
+      //     for that cluster. For example, the kube-system namespace (e.g. the Kube API Server).
+      //   - A mesh administrator wants to slowly migrate services to Istio. They might start by first
+      //     having services cluster-local and then slowly transition them to mesh-wide. They could do
+      //     this service-by-service (e.g. mysvc.myns.svc.cluster.local) or as a group
+      //     (e.g. *.myns.svc.cluster.local).
```
linsun, line 430: could we also add multicluster scenario here?  e.g. a mesh administrator wants to restrict certain services for local cluster only and not automatically exposed and routed to the remote cluster in this model - https://istio.io/docs/setup/install/multicluster/shared/#main-cluster?

mesh/v1alpha1/config.proto
linsun: could this be label selector based where admin can define the corresponding labels and users can choose the correct labels to apply without the need to update mesh config file?   @nmittler @nrjpoddar

costinm:
Ok with DR override - except not sure if DR will survive or be folded into
ServiceEntry or new K8S service extensions.

But yes - my intention was to keep it as close as possible to what we would
put in DR. And even in future allow admin to embed real DR in mesh config,
to override user settings.

nmittler:
@costinm 

> Host identical to what we use in DR. I would not exclude 'bring your own' names.

Was this in response to something? Perhaps https://github.com/istio/api/pull/1383#pullrequestreview-396919384?

mesh/v1alpha1/config.proto
nmittler: @linsun The label-based approach has been discussed (i.e. labels on services and namespaces). The main issue there is that users may not have permission to modify those resources (e.g. kube-system). This sort of thing better aligns with client-side load balancing behavior ... the sort of thing defined in DestinationRule today.

That being said, using labels in this way also leads to a rather clumsy API. Putting it in meshconfig is also less than desirable, but it at least keeps the hack in one place. Also, meshconfig was the agreed landing spot for this from the last TOC meeting. Changing that decision would require another TOC discussion, which means this would likely miss v1.6.

mesh/v1alpha1/config.proto
nmittler: @linsun yes, absolutely. As soon as we agree on the structure, I'll add an example. In fact, a multicluster example would be the only one that would make sense for this setting :).

mesh/v1alpha1/config.proto
nmittler: @costinm @nrjpoddar what do you think about `ExperimentalDestinationRule` for the name? Avoiding the controversial term "override".

costinm:
Why is it controversial ? I would say 'override' is quite intentional -
it's been requested for quite some time to allow the admin to
override. The 'defaults' and the config namespace are very useful - but in
some cases we want specific settings to be forced by
admin to certain values, regardless of user intent.

I would resolve the controversy rather than use 'experimental'. For
consistency I think we'll likely expand this pattern to
other config types.

Also as mentioned in a different comment, I would use ServiceOverride or
ServiceEntryOverrides - the key is a service name,
and most of the discussions around DestinationRule indicated the intent to
move some of the fields to ServiceEntry and
deprecate. I don't feel very strongly about it - but I think there is
consensus that DR is not ideal.

nmittler:
@costinm I'm personally fine with `DestinationRuleOverrides` 

@louiscryan @nrjpoddar thoughts?

mesh/v1alpha1/config.proto
louiscryan: We also have a standardized syntax for contextualizing to namespace. See Sidecar API

[<namespace>/]<host-match>

where the assumed default if <namespace> is omitted is that we match all namespaces

mesh/v1alpha1/config.proto
linsun: Thank you, SG!

nrjpoddar:
> @costinm I'm personally fine with `DestinationRuleOverrides`
> 
> @louiscryan @nrjpoddar thoughts?

I think keeping it general and not tied to overrides is better if we want to extend this field. Individual attributes can mention overrides or have comments explaining if it's an override or default.

mesh/v1alpha1/config.proto
linsun: @nmittler thanks, I forgot the scenario that users don't have permission to modify the resources.  

I would prefer override than experimental.

nrjpoddar: Adding `ExperimentalDestinationRules` ties to an API i.e. we would need to need to redo this config if we deprecate DR.

nmittler:
@nrjpoddar 

> I think keeping it general and not tied to overrides is better if we want to extend this field. Individual attributes can mention overrides or have comments explaining if it's an override or default.

The problem with a generic name like ServiceSettings is that it's not descriptive enough to be immediately obvious. Just about every API we have in Istio could fit under the umbrella of ServiceSettings. I think the approach we're trying to take here is to be specific enough that the name has meaning, while keeping it abstract enough that we can reuse it going forward. With that said, the current problem we're trying to solve is to provide a way to incubate new feature config into `DestinationRule`. This is a actually a fairly broad brush, and might even provide us a consistent migration story for all things headed toward DR. I'm not sure if there is much of a difference here between `DestinationRuleOverrides` or `ExperimentalDestinationRules`, etc. But having `DestinationRule` embedded in the name somewhere is useful for understanding.

Does that seem reasonable?

nrjpoddar:
> @nrjpoddar
> 
> > I think keeping it general and not tied to overrides is better if we want to extend this field. Individual attributes can mention overrides or have comments explaining if it's an override or default.
> 
> The problem with a generic name like ServiceSettings is that it's not descriptive enough to be immediately obvious. Just about every API we have in Istio could fit under the umbrella of ServiceSettings. I think the approach we're trying to take here is to be specific enough that the name has meaning, while keeping it abstract enough that we can reuse it going forward. With that said, the current problem we're trying to solve is to provide a way to incubate new feature config into `DestinationRule`. This is a actually a fairly broad brush, and might even provide us a consistent migration story for all things headed toward DR. I'm not sure if there is much of a difference here between `DestinationRuleOverrides` or `ExperimentalDestinationRules`, etc. But having `DestinationRule` embedded in the name somewhere is useful for understanding.
> 
> Does that seem reasonable?

I'm not totally convinced, the important thing for users to help migrate from MeshConfig -> DR (or whatever API we chose) is that the internal schema remains the schema and their semantic meaning is consistent. As there might be many fields within MeshConfig which are already providing defaults/overrides for DR fields, how do you propose reconciling them?

Why not just add a comment, saying this is experimental and we might move this field to other APIs like DR in future?

nmittler:
@costinm ^^

Honestly, I don't really care at this point. I just need to land this ASAP. If @costin and @nrjpoddar can reach an agreement, I'm happy.

nrjpoddar:
> @costinm ^^
> 
> Honestly, I don't really care at this point. I just need to land this ASAP. If @costin and @nrjpoddar can reach an agreement, I'm happy.

You already have my approval and I won't block this PR. I think Louis also approved the current version afaict.

costinm:
Neeraj - entire MeshConfig is alpha1.

This is the first step - to maintain consistency we'll add as needed
overrides for other APIs, and for existing fields in the API that
need admin override. It's a pattern that will work for most of our needs -
DR and ServiceEntry override can apply to any service or cluster,
Sidecar or Gateway overrides can apply to workloads, while
VirtualServiceOverrides will apply to routes.

I would note that another thing that was discussed is merging semantics for
MeshConfig and ProxyConfig - so we don't have to
keep all the overrides in one big resource. And we still want to define a
CRD for each of them.

kyessenov:
FYI fields should never be removed from API. This will block istio update until you're done with the implementation.

mesh/v1alpha1/config.proto
```diff
@@ -410,12 +410,43 @@ message MeshConfig {
   ThriftConfig thrift_config = 49;
 
   // $hide_from_docs
-  // Set the namespaces whose services will be considered local to the cluster. Cluster
-  // local services will not be exported to and will, therefore, not be accessible by
-  // workloads residing in other clusters. A common use case is the Kube API Server, where
-  // workloads expect to always reach the in-cluster server. By default, the namespace
-  // "kube-system" will be cluster-local.
-  repeated string cluster_local_namespaces = 50;
```
douglas-reid, line 418: @nmittler Deleting this field is apparently causing a break in `istio/istio`. Is someone working on addressing that failure?

costinm:
It never shipped - so we're not worried about upgrade.

PR #1384, opened 2020-04-14
Make operator.meshConfig an interface type

howardjohn:
Otherwise we lose information as the round trip of json -> meshconfig ->
json is lossy. See https://github.com/istio/istio/issues/22948 for more
information


howardjohn:
See https://github.com/istio/istio/pull/22951 which is dependent on this PR and both adds additional validation and fixes a bug

PR #1389, opened 2020-04-16
Add nodeselector annotation for nodeport gateway services

rshriram:
In large clusters, when VMs talk to nodeport gateway services,
it may not be preferable to allow all kubernetes nodes to be
reachable from any VM. We need a global default (any node or use
only nodes with certain labels) as well as per service override
in case the service is in a tightly firewalled zone.

This PR adds service level annotation as well as mesh config that
specifies global defaults. Note that the functionality for this PR
already exists in istiod. With this PR, the selection annotations
defined inside istiod are now properly documented and made part of the
API.


mesh/v1alpha1/network.proto
```diff
@@ -52,11 +52,14 @@ message Network {
...
-    // multicluster) or supplied by MCP server.
-    string from_registry = 2;
+    // $hide_from_docs
+    string from_registry = 2 [deprecated=true];
+
+    // Add all endpoints from the specified cluster into this network.
+    // In a multicluster setup, the names of the clusters should
+    // correspond to the kubeconfig file name inside the secret that
+    // was used to configure the cluster (Kubernetes multicluster).
+    string from_cluster = 3;
```
rshriram, line 62: I renamed this as the word registry is no longer accurate.

networking/v1alpha3/sidecar.proto
```diff
@@ -462,6 +462,7 @@ message Sidecar {
   // and egress listeners in a sidecar unless overridden. There are no
   // built in defaults for this setting. If not specified, the
   // features will be disabled.
+  // $hide_from_docs
```
rshriram, line 465: hiding this as the feature is unimplemented due to implementation hurdles.

mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,42 @@ message MeshConfig {
   // If not specified, this will be enabled by default.
   google.protobuf.BoolValue enable_prometheus_merge = 51;
 
+  message NodePortGatewayIPSelection {
+    // A shortcut for specifying how to select the node IPs for nodeport Gateways.
+    enum Mode {
+      // Nodeport gateway services will not be routable automatically from other clusters.
+      // Note: you can still explicitly specify one or more gateway node IPs in the mesh networks.
+      DISALLOW_NODE_IPS = 0;
```
rshriram, line 436: note that this is the behavior in 1.5. Though its buggy. We were only considering Gateways of LB types preventing users of nodeport gateways from taking advantage of the functionality.

mesh/v1alpha1/config.proto
```diff
@@ -428,8 +428,42 @@ message MeshConfig {
   // If not specified, this will be enabled by default.
   google.protobuf.BoolValue enable_prometheus_merge = 51;
 
+  message NodePortGatewayIPSelection {
```
linsun, line 431: Is this only when gateway service deployed as node port service type?  we normally deploy it as LB service type, how can we leverage USE_NODES_WITH_LABEL?

linsun:
overall a useful feature.  Need to understand how it can be used for LB service type.   

cc a few others to review as well @morvencao @howardjohn @costinm @nrjpoddar @knrc

annotation/annotations.yaml
```diff
@@ -36,14 +36,14 @@ annotations:
...
-    hidden: true
+    hidden: false
     resources:
       - Service
   - name: alpha.istio.io/identity
     description: Identity for the workload.
       NOTE This API is Alpha and has no stability guarantees.
     deprecated: false
-    hidden: true
+    hidden: false
```
howardjohn, line 46: is this even implemented?

annotation/annotations.yaml
```diff
@@ -54,12 +54,21 @@ annotations:
     hidden: false
     resources:
       - Service
+  - name: networking.istio.io/gatewayNodeSelector
+    description: This is a service-level annotation for gateway services on Kubernetes of type NodePort.
+      The annotation value is a json map of node labels (e.g., "{'key':'value'}" ) that select specific
```
howardjohn, line 59: nit: should we make it yaml? Then they can use json as given or
```
networking.istio.io/gatewayNodeSelector: |
  foo: bar
```

annotation/annotations.yaml
```diff
@@ -54,12 +54,21 @@ annotations:
...
+      nodes in a cluster. In mesh expansion setups, VMs will use the node IPs from these selected nodes
```
howardjohn, line 60: is this only mesh expansion? See https://docs.google.com/document/d/1sLw8N62JlkAmJsYMqU6MUuXp2Jz-ExoEs1f1XsrxDy4/ @costinm

mesh/v1alpha1/config.proto
howardjohn: I don't think what you are describing is possible. What are you trying to accomplish?

annotation/annotations.yaml
rshriram: If not I can nuke it.

annotation/annotations.yaml
rshriram: That works. As long as I can provide an example

annotation/annotations.yaml
rshriram: We need to state the motivation to be something concrete.

mesh/v1alpha1/config.proto
rshriram: We have it already today. If I specify all, we won’t use a label selection criteria in the lister interface in the node informer. If I specify none, then we can even avoid starting the node informer or stop bothering about any updates to the node informer.

rshriram: @hzxuzhonghu do we even need this global selector now that we have managed to squash node updates?

rshriram:
@linsun we have built in support for lb services. No user configuration necessary.

mesh/v1alpha1/config.proto
hzxuzhonghu: I think it's better to have this to reduce the number of nodes that are used as a loadbalancer in multi-cluster scenario.

Otherwise, the eds size could be super large and the node event callback could even be super busy.

annotation/annotations.yaml
nrjpoddar: @rshriram how will this interact with `service.spec.externalTrafficPolicy`? AFAICT it's a bit confusing for users; if you set the `externalTrafficPolicy` to `local` you already constrain the nodes that will be selected for load distribution. Will this filter be applied after `service.spec.externalTrafficPolicy` has already pruned the list?

costinm:
Not sure if you saw the recent proposal to expose Istiod as NodePort or
Headless.
Wondering if we can use a consistent approach here.

I'm not sure I understand the full picture: for NodePort Services we do
have
the IPs of nodes running the pods. In theory all other nodes should also
forward -
and we do have their IP as well.

The real problem I can see is that not all nodes in a cluster have
externally routable IP
addresses - so the real problem IMO is how to identify the nodes that are
reachable.
And that's not solved by annotations.

That's the part that confuses me - the only information that is needed is
really if a NodePort
is reachable from outside, which is a combination of public IP address (
easy to determine )
 and firewall rules / network policies - which are trickier.

rshriram:
@costinm the service level annotation is really for the purpose you mentioned. First of all, we dont have proper node port support for VM to k8s, when VMs do not have direct access to k8s pods (multi-network). We consider gateways with LB IPs only and ignore the node port types. I just fixed this recently in istio/istio. 

Secondly, with regard to the annotation, we have customers who have very large onprem clusters (100s of nodes).  But they dont have Load balancer service type on k8s. But these nodes are divided into various firewall zones such that only nodes within a zone can send traffic to one another. So when VMs in one zone talk to k8s gateways in the same zone, today, without this annotation, VMs (i.e. the sidecars) end up load balancing the traffic to all nodes in the cluster and this fails due to the firewall restrictions. The NodeSelector annotation on the gateway service is set by the end user to indicate the set of nodes where the pods of the gateway service has been launched (while we can try to auto infer this, it is much more harder and lot more work for istio). With this annotation, only nodes IPs matching the user specified labels will be used in the clusterExternalAddresses field of the Gateway service object (this cluster External addresses field is what gets delivered to VM sidecars when they are in different networks).

@nrjpoddar this annotation has nothing to do with externalTrafficPolicy. This annotation is about multi-network case where VMs are in one network and k8s pods are in another. It controls the set of k8s node IPs that a VM can reach when it contacts a k8s Gateway of NodePort type. Similar reasoning applies in the multi-cluster scenario as well.

costinm:
How is a service annotation helping ?

The problem seems to be that the node ip may be reachable or not. The
service doesn't know this.

For a large cluster users may have only a subset with public ip - or
firewall rules allowing access. Usually the nodes have labels/taints so
pods allowed to run are controlled.

So I was thinking a label selector for nodes would be needed, to allow user
to indicate which of the nodes ips to use from remote clusters.

costinm:
I mean:gateway can be launched on any node.

Unless you want to restrict the gate to the reachable nodes - but in this
case you don't need a service annotation, istiod knows the node ip for the
pods.

If Gateway used host network - restricting it to the reachable nodes would
be ideal - but still we won't need any extra config from user.

nrjpoddar:
@rshriram get your changes for hiding the unimplemented API in if you can. The new APIs doesn't seem to have consensus. Might be easier to create a separate PR for just hiding from docs.

hzxuzhonghu:
This is not necessary if we have set externalTrafficPolicy=Local

nrjpoddar:
> This is not necessary if we have set externalTrafficPolicy=Local

I also thought that, look @rshriram answer:

this annotation has nothing to do with externalTrafficPolicy. This annotation is about multi-network case where VMs are in one network and k8s pods are in another. It controls the set of k8s node IPs that a VM can reach when it contacts a k8s Gateway of NodePort type. Similar reasoning applies in the multi-cluster scenario as well.

hzxuzhonghu:
If set nodeSelector on gateway pods, then the network limit isn't a problem

nrjpoddar:
> If set nodeSelector on gateway pods, then the network limit isn't a problem

That might work, if GW workloads are only used for this purpose, you can set node selectors + local external load balancing policy.

costinm:
Sure, but why not pick a simpler solution ?

The problem is to identify the _Nodes_ that are reachable. NodePort opens
the targetPort on all nodes in the pool, but only a select set is reachable
- either because
they have public IP or are allowed by NetworkPolicy or firewall rules.

So we can add a selector to IstioNetworkGateway to identify the reachable
nodes, maybe using the same selector that NetworkPolicy has ( or a map of
selectors - with the cluster name
as a key, for MC). Right now it has 'registryServiceName' - to identify the
Gateway. The extra selector will further restrict the set of IPs identified.
Default can be any node with a public IP ( which the current code already
does).

The other option that was not mentioned is deploying the Gateway with
'hostNetworking', and use taint/toleration to make sure the pod is running
on specific
nodes that are reachable.

Annotating the service doesn't make sense to me: a Service can be deployed
by CI/CD in multiple clusters, may be part of a helm chart installed by the
user, etc.
Not the right place to have annotations that are specific to where it is
deployed.

I would have suggested using the newly added 'Overrides' mechanism in mesh
config - but IstioNetworkGateway seems a better place.

rshriram:
>Annotating the service doesn't make sense to me: a Service can be deployed
by CI/CD in multiple clusters, may be part of a helm chart installed by the
user, etc. Not the right place to have annotations that are specific to where it is
deployed.

This is not a strong argument to not having the annotation at the service level. As I mentioned, this is not a new practice we are introducing. We have customers with very large openshift clusters on-prem, using service level annotations along with some custom controllers that update the hardware LBs with the node port IPs. They have been operating like this for few years now. So, going to these folks and asking them to figure out yet another way of providing the node selector annotations just for the sake of "Istio" is not helpful. If this were a brand new practice, sure. 

btw, given that we are willing to go to excessive lengths to accommodate some customers as seen in #1383 recently, this PR has much less complexity to worry about :). So let us not bike shed on where this annotation lands. I would much rather that this be an annotation than an API field which is much harder to deprecate or move around. In future, when we optimize the kube controller implementation to automatically pick out the node port IPs corresponding to the pods, we can remove this annotation and no harm will be done. Much more easier than adding and removing an API field for a point in time implementation.

costinm:
If there are real customers using this pattern - I'm not going to say no.
But if there is a standard practice already - and they do
annotate their services this way for other purpose: I wouldn't mind to just
use their existing annotations, just like we
accept all kind of Ingress and k8s oddities for the sake of serving
customers.

My objection was for the Service annotation to be the primary or
recommended API. It is a very useful feature - let's
have a proper API defined, consistent with the rest of networking - and
it's perfectly fine to mention that this is equivalent
with an annotation on Service or even Pods. We do this for other
annotations as well.

In general: I think our general approach should be that everything we
define in the API can be (even automatically)
expressed with an annotation. So as long as we use the same keywords for
the istio API and annotation - I'm not
going to say no to anything, and if someone can write some generic code so
we do this consistently - even better.

Just Service annotation - I don't think it's the right API, even if
customers are doing this now. Ingress had plenty of annotations -
yet we defined Gateway as primary API.

costinm:
But please include some link or example on how it is used, in particular if
it's a practice on openshift or other vendors.

rshriram:
https://clouddocs.f5.com/containers/v2/kubernetes/ dig through the docs.. these controllers mostly work on nodeport mode only.. and in large clusters with 500+ nodes, there is a lot of micro-segmentation - which is where these service level selectors come in. 

And to reiterate, this is *temporary* - that is why its annotation and not API. Once we automatically select the node IPs based on the pods, this will go away.

nrjpoddar:
Ah, it’s good to see someone else refer docs (clouddocs.f5.com) I created like years ago ;)

costinm:
I don't see an annotation in a quick scan - but if there is one, I'm very
happy to support it. Just please link to that doc in the comments and/or
PR. I don't think I ever said no to more interop.
But make it clear what it is - not a istio api, but a 'foreign' api we
support for interop with vendors like the Internal LB annotations.

At the same time - we do want a proper istio config, it is an important use
case.  See my doc for adding a nodeport option for istiod.

nrjpoddar:
I talked with @rshriram about this and the use case is valid. Basically on-prem cases where service type loadbalancer is not supported (which is very very common), and customers use "node port" service type, is there a way to send a pruned node ip list to VMs w/sidecars.

Can you tell me why only 2 modes (i.e. a flag) is not sufficient in this case where "istiod" will either send all node IPs or only the node IPs where Gateway workloads are installed. I'm suggesting that we simulate the behavior of a externalLBType: local for node ports?

rshriram:
>Can you tell me why only 2 modes (i.e. a flag) is not sufficient in this case where "istiod" will either send all node IPs or only the node IPs where Gateway workloads are installed. I'm suggesting that we simulate the behavior of a externalLBType: local for node ports?

We can do this in next release. In this release, we have implemented and perf optimized the annotation based selector model. That is one of the reasons why I insist on annotations and not an API. When we add the mode you are suggesting, the annotation will automatically become a no-op and it can be removed without any impact to the end user.

costinm:
The use case is extremely valid - that's why I'm very interested :-)

The problem is that there is no guarantee that Gateways will run on the
right nodes. Yes, there is a performance benefit if you
run the Gateway in the nodes with public IP - but in that case you are
better using host networking ( so you can use 443, 80 directly,
which is useful if you don't have something in front ).  NodePort creates a
port - in the 3xxxx range - on ALL nodes. So you need
 something to map from 443/... to 3xxxx - but you can  hit any node in the
pool.

The Gateway can be deployed on any of the nodes - but only nodes in pools
with public IPs  and proper firewall should be
used. The benefit is that Gateways can run in the larger set of nodes
without public IPs.

The second mode is to control (taints/tolerations) where the Gateway are
deployed, so Gateway are only running on the
nodes with public IPs. In this case - you don't need a label selector or
annotation. Simply use the the IPs of nodes where
Gateways are installed.  In this case you can also use host networking (
making sure max 1 gateway/node is deployed),
and you get best possible performance with no translation or LB in front.

I agree with you - we should also have a flag for the simple use cases, but
to send 'all node IPs' or 'all public IPs' if the
Gateway is type NodePort.

costinm:
As long as we are moving in this direction - I don't mind how the changes
are scheduled to releases. Just to have clarity
in the docs - so users are aware how things will evolve and what we are
planning. When I read the PR it was not
clear to me :-)

Maybe a RFC with the summary of this thread would help ?

It's not only Gateway, but also Istiod that should be exposed this way. I
have an RFC for that, and I think this discussion
creates even better options for that.

rshriram:
I am going to close this. the implementation is already in istio/istio. There seems to be an inordinate amount of noise to document an existing implementation.

howardjohn:
@rshriram I think if there is pushback to this we should not just close this PR and have  the feature to istio/istio without documenting it. The objection isn't to documenting an existing feature its the feature itself. (note - I didn't follow any of this thread closely, just a general comment)

costinm:
I'm not objecting to the feature itself - I think it's _extremely_ useful
and important.

PR #1394, opened 2020-04-18
Configuration option for gateway topology

nrjpoddar:
Related [RFC](https://docs.google.com/document/d/1JJXvQSINaAww-Hl_cdfs1QnkvfaIb6omNqzg4RXcQRY/edit)


nrjpoddar:
@howardjohn @costinm PTAL.

mesh/v1alpha1/proxy.proto
```diff
@@ -188,6 +188,57 @@ message SDS {
...
+  // forwarding it to the upstream services in the cluster.
+  // The default value of num_trusted_proxies is 0.
+  uint32 num_trusted_proxies = 1;
+
+  // $hide_from_docs
+  // ForwardClientCertDetails controls how the x-forwarded-client-cert (XFCC)
+  // header is handled by the gateway proxy.
+  enum ForwardClientCertDetails {
+    // Do not send the XFCC header to the next hop. This is the default value.
+    SANITIZE = 0;
```
howardjohn, line 217: Can we add an `UNDEFINED = 0` field? Otherwise we cannot distinguish unset and default, so we can never change the default. We probably don't want to change the default here but its caused us pain in the past

howardjohn:
We should merge https://github.com/istio/istio/pull/22879 in order to properly implement this

mesh/v1alpha1/proxy.proto
nrjpoddar: Will do.

nrjpoddar: I wanted this to be consistent with Envoy options to keep parity. I can add UNKNOWN=0 which I think should be added for enums but we will lose parity. @howardjohn wdyt is preferable here?

howardjohn: I don't have a super strong opinion, but if we keep it as is we can never change the default. so if we never need to do that, its not an issue

nrjpoddar: I think SANITIZE is a safe (and will be) default. I'm leaning more towards leaving it as is.

nrjpoddar:
> Requesting at least changing the broad term, it'll make it confusing to add (real) topology settings.

Very reasonable, name bike shedding time:

* GatewayTopology (my preference)
* GatewayProxySettings
* Topology
* GatewayForwardProxySettings

Anything else you like, please add to the list.

costinm:
I can live with your preference.

My preference is ProxyChaining - because it should also be used for
Sidecars and other proxies, and is more specific than 'toplogy'.
But given that K8S may treat sidecars as a kind of Gateway class, I think
we can also use GatewayChain.

mesh/v1alpha1/proxy.proto
```diff
@@ -188,6 +188,57 @@ message SDS {
...
+  // X-Forwarded-For (XFF) header from the incoming request. If the
+  // X-Forwarded-For (XFF) header is missing or has fewer than N addresses, the
+  // gateway proxy falls back to using the immediate downstream connection's
+  // source address as the trusted client address.
+  // Note that the gateway proxy will append the downstream connection's source
+  // address to the X-Forwarded-For (XFF) address and set the
+  // X-Envoy-External-Address header to the trusted client address before
+  // forwarding it to the upstream services in the cluster.
+  // The default value of num_trusted_proxies is 0.
+  uint32 num_trusted_proxies = 1;
```
costinm, line 210: Could you add a link to the envoy documentation, and some info about what exactly will happen in the generated config ?

Hard to find and debug otherwise. 

Also are they using the same term (num_trusted_proxies ) ? Generally if envoy (or K8S) is using a word I think its better to reuse it.

mesh/v1alpha1/proxy.proto
```diff
@@ -188,6 +188,57 @@ message SDS {
...
+    // information to the request’s XFCC header and forward it.
+    APPEND_FORWARD = 2;
+
+    // When the client connection is mTLS, reset the XFCC header with the client
+    // certificate information and send it to the next hop.
+    SANITIZE_SET = 3;
+
+    // Always forward the XFCC header in the request, regardless of whether the
+    // client connection is mTLS.
+    ALWAYS_FORWARD_ONLY = 4;
```
costinm, line 233: Same as above.

mesh/v1alpha1/proxy.proto
nrjpoddar: Envoy uses the term `xff_num_trusted_hops` which scopes it to xff, i feel like this setting `num_trusted_hops` applies to other protocols too. I like calling it `num_trusted_proxies` as it's more clear vs hops.

I will add link to Envoy settings.

mesh/v1alpha1/proxy.proto
nrjpoddar: These settings are taken exactly from Envoy; I will add link to Envoy docs

nrjpoddar:
> In general looks good - asked for few more comments and double-checking the terminology is consistent with envoy.

@costinm updated PR based on your feedback. PTAL.

nrjpoddar:
@duderino @linsun @rshriram can I get approval? This has been hashed out extensively.

PR #1397, opened 2020-04-22
Add TLS setting to Tracing message

hanahmily:
These changes intend to add generic TLS setting to proxy tracing client, related to https://github.com/istio/istio/issues/22741. The tracer, for instance, zipkin, datadog and etc, can build the encrypted connection to tracing server by it.

I use `istio.networking.v1alpha3.ClientTLSSettings` as TLS setting type due to most of the tracers pick up envoy static cluster as the client. And envoy_metrics_service and envoy_als_service also adopt the same approach.

Signed-off-by: Gao Hongtao <hanahmily@gmail.com>


mesh/v1alpha1/proxy.proto
```diff
@@ -172,8 +169,14 @@ message Tracing {
   // $hide_from_docs
   double sampling = 7;
 
+
+  // Use the tls_settings to specify the tls mode to use. If the remote tracing service
+  // uses Istio mutual TLS and shares the root CA with Pilot, specify the TLS
+  // mode as `ISTIO_MUTUAL`.
+  istio.networking.v1alpha3.ClientTLSSettings tls_settings = 8;
```
hzxuzhonghu, line 176: It seems more reasonable to put it under specific Tracer

There is already CaCertPath under `Tracing_Lightstep`

hanahmily: That's my original plan(I intended to implement zipkin TLS), but after getting feedback from @howardjohn 
I also prefer a generic TLS setting for all of the tracers. That means most of the tracers could set up a TLS connection by it. You can find more details on this thread  https://github.com/istio/istio/issues/22741.

And in this PR, I also drop the CaCertPath from `Lightstep` so that it can use the generic or tracing scope TLS settings.

mesh/v1alpha1/proxy.proto
```diff
@@ -172,8 +169,14 @@ message Tracing {
   // $hide_from_docs
   double sampling = 7;
 
+
+  // Use the tls_settings to specify the tls mode to use. If the remote tracing service
```
nrjpoddar, line 173: Add hide_from_docs. 

Can you share the Envoy configuration corresponding to this option? Is that going to be bootstrap or xDS?

rshriram: This is going into bootstrap. We have it already for ALS. This PR is simply bring it in sync by adding same functionality to tracing as well.

rshriram: https://github.com/istio/istio/blob/master/tools/packaging/common/envoy_bootstrap_v2.json#L521-L523

nrjpoddar: I see, I would still say hide it from docs until you have finished implementation, then unhide it as 1.6 code freeze is imminent.

nrjpoddar:
Directionally ok, I won't remove the hold, until the option is hidden or I see the corresponding PR in istio/istio for implementation.

hanahmily:
@rshriram @nrjpoddar I pushed the corresponding PR in istio/istio

PR #1398, opened 2020-04-22
Adding example for ClusterLocal services

nmittler:


mesh/v1alpha1/config.proto
```diff
@@ -411,6 +411,18 @@ message MeshConfig {
...
   // $hide_from_docs
   // Settings to be applied to select services.
+  //
+  // For example, the following configures all services in namespace "foo" as well as the
+  // "bar" service in namespace "baz" to be considered cluster-local:
+  //
+  // ```yaml
+  // serviceSettings:
+  //  - settings:
+  //       cluster_local: true
```
nrjpoddar, line 421: Spacings is off.

PR #1401, opened 2020-04-24
Add proxy config annotation

howardjohn:
This is already implemented in istio/istio. No problem changing the
syntax if people have concerns as it has not yet shipped - just need to make sure it makes it
into 1.6.

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -343,3 +343,10 @@ annotations:
     hidden: false
     resources:
       - Any
+  - name: istio.io/proxyConfig
+    description: Overrides for the proxy configuration for this specific proxy. Available options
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -343,3 +343,10 @@ annotations:
     hidden: false
     resources:
       - Any
+  - name: istio.io/proxyConfig
```
rshriram, line 346: sidecar.istio.io

howardjohn: this applies to gateways as well, should we still use sidecar annotation?

rshriram: maybe call it as proxy.istio.io?

howardjohn: @rshriram I updated to proxy.istio.io/config

PR #1403, opened 2020-04-26
Added UNDEFINED for Gateway Topology

nrjpoddar:
We set "SANITIZE_SET" as the default option for Gateways and not having an
undefined setting makes it difficult to parse user intent. Note than Envoy has
"SANITIZE" as default 0 unset setting.

PR #1407, opened 2020-04-28
Manual cherry-pick of #1403

nrjpoddar:
Added UNDEFINED for Gateway Topology

We set "SANITIZE_SET" as the default option for Gateways and not having an
undefined setting makes it difficult to parse user intent. Note than Envoy has
"SANITIZE" as default 0 unset setting.


nrjpoddar:
Manual cherry-pick of #1403

nrjpoddar:
@jacob-delgado @brian-avery PTAL.

PR #1414, opened 2020-04-30
Initial draft to allow customization of Certificate signing

costinm:
Fix the broken 'secretName' option of the Certificate option.

Add fields to indicate the principals that are allowed to request certificates with custom format.

Indicate what namespaces a principal requesting a certificate is allowed to use. Note that granularity
of namespace matches typical user needs - we can also use specific Spiffee URLs or service accounts, but maybe when this feature is promoted from mesh config. 

The implementation in istio-api will use this config and the authenticated principal on the CSR signing
request to determine which SANs are allowed in the CSR. 

https://docs.google.com/document/d/1mL658zm9CWzVSjm3yuk6O5vIJDlVi0LU0Tv1in5zGKg/edit#heading=h.xw1gqgyqs5b

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -594,13 +594,18 @@ message Certificate {
   // OIDC is used, or a k8s identity, if K8S tokens are used.
   //
   // Treated as an opaque string in matching a request.
+  //
+  // If a Gateway or Istiod instance is requesting certificates, this is the
+  // service account running the Istiod or Gateway deployment.
   string principal = 3;
 
   // List of URI SANs allowed in the CSR request.
   // This is primarily used for explicit SPIFFE identities.
   repeated string uri_SANs = 4;
 
-  // List of allowed namespaces for issuing SPIFFE CSRs.
+  // List of allowed namespaces for issuing SPIFFE CSRs. A CI/CD, admin
+  // or Istiod tenant instance may be restricted to only specific
+  // namespaces where they're allowed to provision identity.
   //
   // This allows a CI/CD or admin to provision certificates for
   // a set of managed namespaces, or a managed VM to use istio-agent
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -393,7 +393,18 @@ message MeshConfig {
   // - `%SERVICE%` will use reviews.prod as the stats name.
   string outbound_cluster_stat_name = 45;
 
-  // Configure the provision of certificates.
+  // Configure the provision of certificates by Istiod, if certificate management
+  // is enabled in the current revision.
+  // Users may use external certificate management, or enable only specific
+  // Istiod revisions to handle certificates - for example only a subset of clusters,
+  // or a subset of revisions.
+  //
+  // Istiod can also delegate to other CA - like K8S. In such cases the config
+  // is used as a filter, but ultimate decision rests with the primary CA.
+  // When delegation happens, Istiod may use its own credentials and permissions
+  // when authenticating to the external CA - so user will still need to use
+  // the Istiod configuration to limit the certificates that workloads can retrieve.
+  //
   repeated Certificate certificates = 47;
 
   message ThriftConfig {
@@ -506,8 +517,8 @@ message ConfigSource {
 // CSR signing service and to specify what identities are allowed
 // in the CSR.
 //
-// Starting with Istio 1.6 certificates are no longer saved as
-// secrets.
+// Starting with release 1.6, Istio-generated certificates are no longer saved
+// as secrets.
 //
 // Example 1: Configure DNS certificate content for Istiod
 // {
@@ -531,6 +542,29 @@ message ConfigSource {
 //     namespaces:
 //     - bookinfo
 // }
+//
+// Example 4: Provision certificate for a tenant or Istiod revision.
+// The Istiod instance will use a separate SA without RBAC permissions to
+// read the master CA. The istiod revision signing CSRs may run in an isolated,
+// more trusted cluster or in a separate namespace.
+//
+// {
+//     principal: istiod-service-account-staging.istio-system-staging
+//     dnsNames:
+//        - istiod-staging.istio-system-staging
+//        - istiod-staging.istio-staging.example.com
+// }
+//
+// Example 5: Provisioning certificate for an Istio Gateway dispatching
+// requests to multiple revisions or tenants:
+//
+// {
+//     principal: istio-ingressgateway-service-account
+//     dnsNames:
+//        - istiod.istio-system.svc
+//        - xds.example.com
+// }
+//
 message Certificate {
   // Deprecated. All requests will use the istiod signing interface.
   // Name of the secret the certificate and its key will be stored into.
@@ -545,16 +579,27 @@ message Certificate {
   // from the certificate, if present in the CSR.
   repeated string dns_names = 2;
 
-  // This is the identity of the caller of the signing service.
-  // For example it can be a SPIFFE identity, if Istio certificates
-  // are used for authentication, or the principal in a JWT token of
-  // OIDC is used. It can also be a k8s identity, if K8S tokens are used.
-  // Treated as an opaque string.
+  // This is the opaque identity of the CSR requester. Istiod must be
+  // configured to authenticate the caller using mTLS, JWT tokens or
+  // other plugins.
+  //
+  // The format matches the identity provided by the auth plugin - for
+  // example a SPIFFE URI for mTLS, or the principal in a JWT token of
+  // OIDC is used, or a k8s identity, if K8S tokens are used.
+  //
+  // Treated as an opaque string in matching a request.
   string principal = 3;
 
+  // List of URI SANs allowed in the CSR request.
+  // This is primarily used for explicit SPIFFE identities.
+  repeated string uri_SANs = 4;
+
   // List of allowed namespaces for issuing SPIFFE CSRs.
+  //
   // This allows a CI/CD or admin to provision certificates for
   // a set of managed namespaces, or a managed VM to use istio-agent
   // to retrieve certificates in its own namespace.
-  repeated string namespaces = 4;
+  repeated string namespaces = 5;
+
+
 }
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -512,13 +512,19 @@ message ConfigSource {
 
 
 
-// Certificate configures the provision of a certificate.
-// It allows the admin to specify the identity of the caller of the
-// CSR signing service and to specify what identities are allowed
-// in the CSR.
+// Certificate configures Istiod's certificate signing feature.
+//
+// Requests to Istiod are authenticated using mTLS, JWT or an authentication
+// plugin. The authenticated principal determines what is allowed in the
+// CSR request.
 //
 // Starting with release 1.6, Istio-generated certificates are no longer saved
-// as secrets.
+// as secrets, but served using SDS to Envoy.
+//
+// This is an internal config for Istiod when acting as a CA - it is not
+// a general API for certificate issuance nor a general purpose RBAC system.
+// The features are intentionally kept to the minimum set of features required
+// for Istio operation - external CAs should be used for more advanced features.
 //
 // Example 1: Configure DNS certificate content for Istiod
 // {
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -501,26 +501,60 @@ message ConfigSource {
 
 
 
-// Certificate configures the provision of a certificate and its key.
-// Example 1: key and cert stored in a secret
-// { secretName: galley-cert
-//   secretNamespace: istio-system
-//   dnsNames:
-//     - galley.istio-system.svc
-//     - galley.mydomain.com
+// Certificate configures the provision of a certificate.
+// It allows the admin to specify the identity of the caller of the
+// CSR signing service and to specify what identities are allowed
+// in the CSR.
+//
+// Starting with Istio 1.6 certificates are no longer saved as
+// secrets.
+//
+// Example 1: Configure DNS certificate content for Istiod
+// {
+//     principal: istiod-service-account.istio-system
+//     dnsNames:
+//     - istiod.istio-system.svc
+//     - istiod.mydomain.com
 // }
-// Example 2: key and cert stored in a directory
-// { dnsNames:
-//     - pilot.istio-system
-//     - pilot.istio-system.svc
-//     - pilot.mydomain.com
+//
+// Example 2: Certificate provisioning for a VM using a CI/CD
+// {
+//     principal: istio@github.com
+//     namespaces:
+//     - bookinfo
+//     - bookinfo-staging
+// }
+//
+// Example 3: Certificate provisioning for a VM using a managed SA.
+// {
+//     principal: aws-sa@amazon-iam
+//     namespaces:
+//     - bookinfo
 // }
 message Certificate {
+  // Deprecated. All requests will use the istiod signing interface.
   // Name of the secret the certificate and its key will be stored into.
   // If it is empty, it will not be stored into a secret.
   // Instead, the certificate and its key will be stored into a hard-coded directory.
   string secret_name = 1;
+
+
   // The DNS names for the certificate. A certificate may contain
-  // multiple DNS names.
+  // multiple DNS names as well as SPIFFE SANs.
+  // Wildcards are permitted. Names not in this list will be removed
+  // from the certificate, if present in the CSR.
   repeated string dns_names = 2;
+
+  // This is the identity of the caller of the signing service.
+  // For example it can be a SPIFFE identity, if Istio certificates
+  // are used for authentication, or the principal in a JWT token of
+  // OIDC is used. It can also be a k8s identity, if K8S tokens are used.
+  // Treated as an opaque string.
+  string principal = 3;
+
+  // List of allowed namespaces for issuing SPIFFE CSRs.
+  // This allows a CI/CD or admin to provision certificates for
+  // a set of managed namespaces, or a managed VM to use istio-agent
+  // to retrieve certificates in its own namespace.
+  repeated string namespaces = 4;
 }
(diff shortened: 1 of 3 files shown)


costinm:
@linsun  this should also better address the config for Istiod certificate. 

@rshriram I think this cover all use cases for VM cert provisioning. We can tweak it when we move
it to a proper API if needed. 

The configuration was initially intended to configure the certificate for the webhooks and internal
Istio components - but it is generic and can be used to configure the use cases we are interested in.

mesh/v1alpha1/config.proto
```diff
@@ -501,26 +501,60 @@ message ConfigSource {
...
-// Certificate configures the provision of a certificate and its key.
-// Example 1: key and cert stored in a secret
-// { secretName: galley-cert
-//   secretNamespace: istio-system
-//   dnsNames:
-//     - galley.istio-system.svc
-//     - galley.mydomain.com
+// Certificate configures the provision of a certificate.
+// It allows the admin to specify the identity of the caller of the
+// CSR signing service and to specify what identities are allowed
```
nrjpoddar, line 506: // It allows the admin to specify the identity of the CSR requester and
// the identities for which the certificate will be generated.

mesh/v1alpha1/config.proto
```diff
@@ -501,26 +501,60 @@ message ConfigSource {
...
+
+
   // The DNS names for the certificate. A certificate may contain
-  // multiple DNS names.
+  // multiple DNS names as well as SPIFFE SANs.
+  // Wildcards are permitted. Names not in this list will be removed
+  // from the certificate, if present in the CSR.
   repeated string dns_names = 2;
+
+  // This is the identity of the caller of the signing service.
```
nrjpoddar, line 548:
```suggestion
  // This is the identity of the CSR requester.
```

costinm:
Sven - there is no RBAC policy here. It simply configures how Istiod generates certificates. There is no 'role' and no 'access control'.  It is quite similar with the config supported by CertManager or Vault 
or K8S itself. Neither are based on RBAC.

costinm:
It is perfectly possible to augment this config with an Istio policy for example - and configure which roles or request properties are allowed to call the CSR signing method. It is also possible to add NetworkPolicies to further restrict access to the API. 

But an RBAC policy is not capable to look inside the CSR, at the SANs. 

The API exists because we had a use case to customize the SANs for internal Istio services. 
The case is still valid - Lin is using this for her MC changes ( with a 'feature' flag for now).

There is nothing in this API incompatible with having additional RBAC or Istio Policy.

mesh/v1alpha1/config.proto
```diff
@@ -501,26 +501,60 @@ message ConfigSource {
...
+  // Deprecated. All requests will use the istiod signing interface.
   // Name of the secret the certificate and its key will be stored into.
   // If it is empty, it will not be stored into a secret.
   // Instead, the certificate and its key will be stored into a hard-coded directory.
   string secret_name = 1;
+
+
   // The DNS names for the certificate. A certificate may contain
-  // multiple DNS names.
+  // multiple DNS names as well as SPIFFE SANs.
```
howardjohn, line 543: Can you put an example with SPIFFE? is it in this same field as dns_names?

mesh/v1alpha1/config.proto
```diff
@@ -501,26 +501,60 @@ message ConfigSource {
...
+//
+// Starting with Istio 1.6 certificates are no longer saved as
+// secrets.
+//
+// Example 1: Configure DNS certificate content for Istiod
+// {
+//     principal: istiod-service-account.istio-system
+//     dnsNames:
+//     - istiod.istio-system.svc
+//     - istiod.mydomain.com
```
linsun, line 517: @costinm is this same as istiodCustomHost I added but allowing user to configure from meshconfig?

mesh/v1alpha1/config.proto
```diff
@@ -501,26 +501,60 @@ message ConfigSource {
...
+// }
+//
+// Example 2: Certificate provisioning for a VM using a CI/CD
+// {
+//     principal: istio@github.com
+//     namespaces:
+//     - bookinfo
+//     - bookinfo-staging
```
linsun, line 525: why would namespace be relevant here when cert provisioning for a vm?

linsun: maybe reword to indicate the namespaces admin/operator can issue the CSR, assuming that is what you meant.

costinm: Each VM belongs to a namespace - same where the WorkloadEntry will be added. 

The CI/CD may manage multiple namespaces - for example a 'staging' - with non-production data for testing - and the prod namespace where it'll push after staging is passing. 

We certainly don't want the CI/CD to get certs for other namespaces.

mesh/v1alpha1/config.proto
costinm: Yes. But more flexible - and this can also be used if/when we expose Istiod using a Gateway. 

Another future use of this would be to specify 

principal: istio-ingress.istio-system
dnsNames:
   - istiod.example.com
   - bookinfo-test.example.com

This would allow the SDS and istio-agent implementation to get DNS certs signed by istiod
for testing or internal use. 

One big security problem we have is that the gateway is now holding extremely high RBAC permissions - can read the root CA private key and all the other secrets. Worse - the gateway is directly exposed to the internet, so should have the fewest privs. Specifying what certs istiod can return to a workload can also be used for this - I'm not sure of all the details, will follow up with a separate proposal.

mesh/v1alpha1/config.proto
costinm: I don't think users should directly manipulate SPIFFE - istio-agent or istioctl are responsible for mapping the service account + namespace to the SPIFFE URL. It is not the same fields as dns_names - we control which spiffee are allowed with namespaces.

costinm:
> This PR introduces a field that specifies which principals are allowed to perform a set of actions. That is access control (you're right it is not RBAC, but it is still access control).

So is 'exportTo' and Sidecar.egress if you look closely :-)

The other way to look at it is: access control defines who is allowed to call the API. 
How the API behaves - i.e. what kind of certs it issues - is business logic and deserve a first class config, so users don't get confused. 

Vault is not relying on anything like RBAC. 
CertManager also has its own high level API for issuing certificates.  It also uses RBAC - and 
we will also use RBAC if we want to restrict who can call this API, but not how it behaves.

> 
> For requests to mint a certificate with a particular k8s service account, we should check whether the calling principal has been granted rights to impersonate that service account.

Ah, that's a great point ! K8S does have a well defined API to allow one K8S user to directly impersonate another. And that's defined by RBAC - and in a pure K8S environment it can be used.

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation

This API is not defining any 'impersonation' - it controls what certificates will be signed by Istiod 
for a user. The CI/CD will then deploy the certificate. If the section related to 'hardware storage' is 
implemented - it is probably possible to bind the keys to the VM so CI/CD can't even use the certificate ( because it doesn't have the key ), so it can't impersonate the workload. ( will require 
some way to bind keys to device - similar with what Android uses ).


> 
> For non-k8s SA requests, why do we even have this feature? Users should use cert manager or something else to mint DNS certs, and not use Citadel for it. We don't even use Citadel for this anymore right?

We do use Citadel - K8S CA API is still in progress and missing few features (and not broadly 
deployed - since it's alpha and is missing features ). 

Yes, ideally VM users would not use Citadel/Istiod but a real CA - that's what I've been trying 
to convince people for a very long time. Apparently a lot of people still believe copying Secrets
from the VM is ok, and I'm very worried that people will start believing copying the CA private key
is fine too. That's why I want to fix this in 1.6.  

> 
> So I propose deprecating this api and controlling CSRs with k8s RBAC checks, and only supporting minting of spiffe certs with k8s service account SANs.

Why ? We allow istio-agent to use OIDC JWT tokens already, K8S supports them for auth - 
why remove this feature ? And why wouldn't we allow AWS/GCP tokens - it seems an important
use case.

We could also get user feedback - comparing the 'impersonation/RBAC' API with this.

> 
> Oh and another problem with this PR: Sticking this in mesh config means every time a service owner wants to provision a cert for a VM they need to go and get the mesh admin to change mesh config.

Yes, that's very intentional and we've been doing it for all 'experimental'/new APIs if there is 
not a super-strong reason to go straight to CRD/beta. 

The use of VM is not as common as pure k8s, the automated testing is still lacking - it would be 
very strange to move this directly to CRD. 

The other reason is multi-revision - we want users to test this on a canary CP, and possibly 
use different configs for the 'prod' CP and 'staging' CP. 

We do the same for the more frequent network config, and quite a few of other more used features.
After feedback - we do move them to proper APIs.

That's why I'm not sure why you're against merging this - if RBAC turns out to be indeed better, we'll
use it when it's ready ( hopefully after following the same process of first experimental/mesh config).
Most of the APIs had changes between 'experimental' to 'proper Istio API'. 

And so far it's not even clear if this will ever be a proper API - in 1.7 or 1.8 maybe K8S APIs will
be ready and we can just drop all signing features from Istiod.

mesh/v1alpha1/config.proto
linsun: I understand it is more flexible... if I don't want to change principal (use whatever default), can I just specify dnsNames?    expose Istiod using a Gateway is exactly my scenario and I don't want to change principal, just add additional dnsName(s).

I don't understand your other scenarios like adding bookinfo to dnsName of ingress gw server certificate.  I thought istiod doesn't manage outer certificates for the gateways, only inner certificates.

mesh/v1alpha1/config.proto
linsun: please clarify in the comment if the namespace is where the VM workload entry will refer to, or the namespace where the istioctl generate cert cmd runs, or both. :)

mesh/v1alpha1/config.proto
costinm: Multi tenant and multi-revision. We are trying to reduce the set of RBAC permissions and power that a revision (or a tenant) has. That's the discussion on having a 'master' istiod that has the high permission - and drop the Secret access for all other tenants. To make this work - we'll need this API to specify that the 'tenant' or 'remote' or 'low priv' istiod SA is allowed to get a cert, and for which DNS names ( including for 'real' DNS ).

Principal will be required in all cases.

The Certificate can also be extended with expiration and other attributes we want to include in
the cert. 

See also:
https://cert-manager.io/docs/concepts/certificate/

mesh/v1alpha1/config.proto
costinm: got it. 

I can use John suggestion and just use uriSan - see 
https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1alpha3.Certificate

My goal was to have this as a much simpler and Istio-focused variant of cert manager Certificate. But if you can suggest some wording that will make it more clear, happy to change, 
English is not my strongest.

mesh/v1alpha1/config.proto
costinm: Added more comments. 

You are not changing the principal - the principal is the service account running Istiod (for multi-tenant, multi-revision), or the service account running the Gateway.
Added example as well. 

Istiod doesn't currently manage certificates for the Gateway - instead Gateways run with RBAC permission allowing them to read all the Secrets in istio-system. While it works, it is not
an ideal situation - this API is a step towards fixing it.

mesh/v1alpha1/config.proto
costinm: PTAL.

costinm:
Sven, PTAL - I added more comments, I hope this is addressing your concerns.

mesh/v1alpha1/config.proto
```diff
@@ -501,26 +512,105 @@ message ConfigSource {
...
+// more trusted cluster or in a separate namespace.
+//
+// {
+//     principal: istiod-service-account-staging.istio-system-staging
+//     dnsNames:
+//        - istiod-staging.istio-system-staging
+//        - istiod-staging.istio-staging.example.com
+// }
+//
+// Example 5: Provisioning certificate for an Istio Gateway dispatching
```
howardjohn, line 564: I am concerned this will be turned into a general purpose ingress certificate thing, which we don't want (?). Why do we need these for gateway anyways, don't we do passthrough to the control plane?

costinm:
We do pass-through - in large part because it was very hard to set
certificates on the ingress.
I can remove the example - and leave just the istiod components, but I
think it is very
important to start improving how we handle the certs in Ingress.

In particular when K8S cert manager API moves out of alpha - or with Istiod
integrating with
other secret providers - it is possible to reduce the dangerous permissions
we now grant to
the gateway. I can start a design doc on this subject - I don't want to
confuse the discussion
on this API, I think the original use case ( of configuring the DNS certs
we use for control plane/istiod)
is good enough.

costinm:
I hope the new docs I added are a bit more clear: this config applies both
for Istiod acting
as a CA issuer but more importantly it applies to the original use case,
when Istiod delegates
to another CA ( K8S cert signing was the initial use case, with Vault also
discussed).

The problem with all those integrations is that JWT tokens that
Istiod receives from workload
can't be passed trough - it would violate the 'audience' field and is
considered a big security
issue. For example K8S signing is based on Istiod having the RBAC
permission to approve -
but it's Istiod that needs to make sure the issued certs meet the
expectations. If we integrate
with Vault or CertManager - Istiod will also use its own credentials and
permissions.

The large picture is complicated and will need multiple design docs and
phases - but the
changes in this PR and the implementation should make it much easier to
move forward
without big churn.

rshriram:
i would like to keep the scope of this work to be extremely constrained. Tenants, provisioning at gateways, etc. seems new - topics we have not discussed as a community yet. So lets remove that piece.

costinm:
Shriram - part of this has been discussed in context of Lin's changes for
multi-cluster ( how to
configure the DNS names associated with istiod ), and part on how to use
Gateways
to split traffic to injector or control plane revisions - in context of
multi-revision.

I don't mind removing it, but this particular API is result of looking at
multiple use cases.

rshriram:
@costinm I spoke with @smawson .. I think using the k8s "impersonate" API is going to eliminate the need for us to even create an API for this service account delegation (https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation). Users can create k8s rbac apis to decide who can impersonate who..

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: limited-impersonator
rules:
# Can impersonate the user "jane.doe@example.com"
- apiGroups: [""]
  resources: ["users"]
  verbs: ["impersonate"]
  resourceNames: ["jane.doe@example.com"]
```

Essentially, we just need istiod to support generating certs against a service account for any caller [which I think it already does, but correct me if I am wrong].

The DNS certs - why do we need a user accessible API for generating certs for our own internal components?

costinm:
Shriram, the proposal is not only for VM provisioning (your use case). You don't need K8S impersonation - if your VM provisioner has access to K8S you can create a service account with the 
desired name, get the token from K8S - and call Istiod directly. 

But there are other use cases - managed VMs ( in clusters where K8S can't authenticate), provisioning when you don't have K8S access,  support for services that need certs with DNS 
entries for legacy clients that don't support Spiffee.

rshriram:
>But there are other use cases - managed VMs ( in clusters where K8S can't authenticate), provisioning when you don't have K8S access, support for services that need certs with DNS
entries for legacy clients that don't support Spiffee.

By managed VMs, do you mean gcp/aws etc. ? If so, doesn't the vm identity proposal that @liminw had tackle this problem?

Can you clarify provisioning without k8s access? What is the deployment scenario here since the user or VM will need to have access to istiod running on some k8s cluster to obtain certs.

And the last piece, 
>support for services that need certs with DNS
entries for legacy clients that don't support Spiffee.

This is something I am not sure we should start supporting with global mesh config API. I am not even sure if we should be building some generic cert issuing system for non istio use cases, when there are other facilities like Cert Manager, etc. Atleast, this requires a much broader discussion about the ramifications to the project.

costinm:
> > But there are other use cases - managed VMs ( in clusters where K8S can't authenticate), provisioning when you don't have K8S access, support for services that need certs with DNS
> > entries for legacy clients that don't support Spiffee.
> 
> By managed VMs, do you mean gcp/aws etc. ? If so, doesn't the vm identity proposal that @liminw had tackle this problem?

I think this proposal is simpler and cleaner than the 3rd option in @liminw proposal.
For the first 2 options it seems we have agreed they're not viable.

The main difference is that in this PR, we have a mesh config that defines what certs 
Istiod will provision. In her proposal - it seems we'll take a dependency on K8S RBAC,
or make call to K8S (somehow that is not yet clear) to let K8S decide if Istiod can issue
the certs.

I generally like simpler solutions - we can start with what I'm proposing here for config and
implement the rest of the change - which is almost identical with Option3, i.e. istiod issuing 
the certs, and if/when an RBAC implementation is ready switch to it, mesh config is alpha.

> 
> Can you clarify provisioning without k8s access? What is the deployment scenario here since the user or VM will need to have access to istiod running on some k8s cluster to obtain certs.

Istiod access is required - it is the XDS server after all.

K8S access is not required. The CI/CD ( github for example, or VMs )  will not have access
to Apiserver in the recommended 'secure' mode, where apiserver access is limited. That's 
part of the reason for the other proposal on MCP-over-XDS - since our MC support has similar 
limitations when the remote K8S apiserver is not accessible. 

There is also the case of Istiod running without K8S - but probably less important. 


> 
> And the last piece,
> 
> > support for services that need certs with DNS
> > entries for legacy clients that don't support Spiffee.
> 
> This is something I am not sure we should start supporting with global mesh config API. I am not even sure if we should be building some generic cert issuing system for non istio use cases, when there are other facilities like Cert Manager, etc. Atleast, this requires a much broader discussion about the ramifications to the project.

I wrote a doc on how to use CertManager for this - https://docs.google.com/document/d/1QReA8XjWrWTnErQYPZUdPA5d2O6fc3VkLWr47hY0Nbg/edit#

They have a similar limitation as Istiod - lack of a config to specify who is allowed to generate
the cert. If they fix it - I agree, we shouldn't. 

Note that mesh config is not a 'generic support for cert provisioning'  - that would be the CertManager
API, with proper CRDs. This PR only defines a mesh config - limited to the mesh admin,
like the rest of the Istiod configuration - not a general CRD and mechanism for end users.

@liminw proposal is the one that would be accessible to end users (RBAC, etc) - and compete/duplicate CertManager/K8S/Vault cert provisioning APIs.

linsun:
@costinm I saw this PR has stalled, but we actually need this API for scenario 1, which is the most simple and understood case where we want to customize DNS host name for central istiod for 1.7.

Can we trim this PR or do a separate PR just for that user case first along with maybe 1-2 other cases that are critical for 1.7?  Would appreciate thoughts going forward with this.

cc @irisdingbj @GregHanson @rshriram @smawson

PR #1418, opened 2020-05-05
Change proxyConfig.concurrency to wrapper

howardjohn:
This allows distinguishing between 0 and undefined. See
https://github.com/istio/istio/issues/23470 for motivation


howardjohn:
> Why are there so many other changes here about ForwardClientCertDetails ?

Good question. I think its because I had to force commit to proto lock since this is changing the type - that might be why the build fails

proto.lock
```diff
@@ -36371,6 +36371,34 @@
               }
             ]
           },
+          {
+            "name": "Topology.ForwardClientCertDetails",
```
nrjpoddar, line 36375: This seems odd? Did you rebase on latest master?

howardjohn: Yes, I think the proto.lock must not always be updated or something?

jacob-delgado:
cc @howardjohn @nrjpoddar Should we get this into 1.6?

nrjpoddar:
Yes, we should.

howardjohn:
We don't need this in 1.6, in my opinion. Its a change to the API not a bug
fix

nrjpoddar:
> We don't need this in 1.6, in my opinion. Its a change to the API not a bug fix

Without this, all workloads will default to 0 (all cores) or not able to override proxy config correctly. If I'm not parsing it correctly, lmk.

howardjohn:
@nrjpoddar there was a bug in 1.6 causing this: #23472. The plan was to do a minimal fix (in that pr) for 1.6 to get avoid regression, then in master add this new option of undefined vs 0

nrjpoddar:
I see. So if I understand correctly:

1. Regression in 1.6 is fixed without API changes
1. In 1.6, a workload can override proxyconfig.concurrency using annotations to anything greater than 0 and it will work as expected. 
1. Setting this to `0` via workload annotations won't work. 

It would be good to get both 2 & 3 working in 1.6 if possible.

howardjohn:
@nrjpoddar I am pretty sure (3) works today - I will verify. The problems this is fixing is:

- Sidecar and gateway behave differently
- There is no way to specify if we want Envoy behavior of "all cpus" today or instead "use requests". It is not 100% clear if we need to support both options

PR #1420, opened 2020-05-05
Add preserveUnknownFields to CRDs

jasonwzm:
Starting Istio 1.6, the min version of k8s is 1.15(https://github.com/kubernetes/apiextensions-apiserver/blob/release-1.15/pkg/apis/apiextensions/v1beta1/types.go#L86) and we are able to support `preserveUnknownFields` field in CRDs. This enables `kubectl explain` as well as unknown fields validation.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -20,6 +20,7 @@ spec:
     listKind: HTTPAPISpecList
     plural: httpapispecs
     singular: httpapispec
+  preserveUnknownFields: false
   scope: Namespaced
   subresources:
     status: {}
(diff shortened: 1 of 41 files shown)


linsun:
Does this mean user can use `kubectl explain` cmd even when there are unknown fields in Istio resources?

jasonwzm:
> Does this mean user can use `kubectl explain` cmd even when there are unknown fields in Istio resources?

if there are unknown fields in the CR users put in, it will be rejected by apiserver. kubectl explain shows the definition of the CRD spec. To make it work, k8s requires `preserveUnknownFields` to be included in CRDs.

linsun:
Thanks @jasonwzm approved - look forward to try this!  Please create some doc for this!

PR #1423, opened 2020-05-06
Remove preserveUnknownFields from EnvoyFilter

jasonwzm:
As mentioned in https://github.com/istio/istio/pull/23559#issuecomment-624412699, the field is causing validation failure in EnvoyFilter. Will create a follow up PR to add such field for EnvoyFilter.

Also remove from `handler` and `instance` where google.protobuf.Struct exists in their fields.


jasonwzm:
Adding hold to test locally before it merges.

PR #1425, opened 2020-05-08
Add Domain to ProxyConfig

nmittler:
We don't currently have a common place to configure the domain suffix used by all system components/tools. This puts it into a common place.

Needed for https://github.com/istio/istio/issues/23652


mesh/v1alpha1/proxy.proto
```diff
@@ -382,6 +382,11 @@ message ProxyConfig {
   // to all the gateways in the cluster unless overriden via annotations of the
   // gateway workloads.
   Topology gateway_topology = 28;
+
+  // The domain suffix used for services within the mesh. K8s allows this to be customized,
+  // see https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/.
```
costinm, line 387: Can you find how it is set up for Kind/kubeadm ? A link would be great - the DNS is a bit different. 

And ideally we would use the same key as k8s uses. 

I generally don't like too broad terms - 'domain' means a lot of things, but if K8S also uses domain we can go with consistency.

costinm:
It seems kubelet is using cluster-domain - sounds a more qualified name than domain ?

nmittler:
@costinm 

> It seems kubelet is using cluster-domain - sounds a more qualified name than domain ?

Yeah I had considered that. In fact the operator's `ProxyConfig` uses that: https://github.com/istio/istio/blob/master/operator/pkg/apis/istio/v1alpha1/values_types.proto#L1346

I'm fine with that.

mesh/v1alpha1/proxy.proto
nmittler: K8s uses "cluster-domain", so I've switched to that.

nmittler:
@rshriram @nrjpoddar @louiscryan @smawson PTAL

nrjpoddar:
Hi @nmittler how does this field relate to this https://github.com/istio/api/blob/faa7ed49fd5af238db8728227995fcdd6f062983/mesh/v1alpha1/config.proto#L262 ?

Secondly, we set the proxy domain: https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml#L123 based on yet another field.

costinm:
The global.proxyDomain will be replaced by this new field - the plan was to
use mesh config and proxy
config ( which are proper APIs ) and move stuff from values.yaml.

In particular from global.yaml - which no longer makes sense ( there is
really one control plane component -
ingress doesn't share much ).

howardjohn:
trust domain is still orthogonal to this though, right?

nmittler:
@nrjpoddar 

> Hi @nmittler how does this field relate to this https://github.com/istio/api/blob/faa7ed49fd5af238db8728227995fcdd6f062983/mesh/v1alpha1/config.proto#L262 ?

IIUC, trust domain is slightly orthogonal to the DNS domain, although in practice they are typically the same. @myidpt should be able to expand on that.

> Secondly, we set the proxy domain: https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/files/injection-template.yaml#L123 based on yet another field. 

IIUC, this is using a proto that is private to istioctl: https://github.com/istio/istio/blob/0321da58ca86fc786fb03a68afd29d082477e4f2/operator/pkg/apis/istio/v1alpha1/values_types.proto#L1346

costinm:
And on the first q: we should probably add comment to both trustDomain and
the new field describing
the relation.

trustDomain defines the identity namespace in spiffee:// URLs - one
trustDomain may be shared
by multiple clusters and be different from proxyDomain, but the default
value is the same.

And probably both should keep the same value unless there is a strong
reason.

costinm:
Almost orthogonal.

In the 'ideal' deployment, the user will use a 'real' domain - like
prod.example.com, with proper DNS
or at least internal DNS ( or 'split DNS').
User would also run some OIDC app - which would be configured with the same
domain, so identities
can be asserted with OIDC JWT tokens under prod.example.com.  The identity
server may run
in cluster or be managed.

User may also use a managed identity server - in which case the trust
domain will be based on the
provider of identity services.

costinm:
If it needs to be passed to proxies - it should be in ProxyConfig.

MeshConfig includes ProxyConfig (or a 'default' proxy config to be exact) -
so istiod can use both.
Technically it can be in both - but people will have to remember to set it
2x.

costinm:
So the current practice is:
- anything that a proxy will use -> ProxyConfig
- anything that Istiod will use -> MeshConfig
- anything that is not used by Istiod but by installer -> values.yaml ( and
equivalent operator proto representation )
- temporary/experimental things - we still use env / annotations / node
metadata.

We still have a lot of CLI/env and values.yaml that need to be cleaned up,
reviewed and
moved to ProxyConfig and MeshConfig if they still make sense.

smawson:
Interesting, OK. Fine with it for now although I think we need to spend some time trying to rationalize this stuff in the near future, I think we may want to support pushing some mesh-wide configuration down to proxies without requiring an explicit field in proxy config. But that can wait till we have a proposal for how our config evolves.

costinm:
Well, we had a spaghetti of options passed to proxies - from global,
values, injection template and indirectly from proxy and mesh config.

The first fix was to pass entire mesh config - but it was too much info,
and the point and name of ProxyConfig was to have a clear representation of
the proxy config.

We can populate fields in proxy config based on mesh config.

nmittler:
@costinm 

> We can populate fields in proxy config based on mesh config.

Can you clarify what you mean by this? Are you suggesting to have this value both in 2 places: in ProxyConfig and at the top-level in MeshConfig? Or are you suggesting that the default ProxyConfig will be used to send to individual proxies?

costinm:
The second. We do this with everything else, default proxyconfig, we don't
duplicate.

nmittler:
@costinm I wasn't sure if we reached agreement.

@nrjpoddar @smawson mind approving?

PR #1427, opened 2020-05-09
authz: remove alpha RBAC API

yangminzhu:
The API will also be removed in code in https://github.com/istio/istio/pull/23668.

#23668


yangminzhu:
@jasonwzm : https://github.com/istio/api/blob/master/scripts/check-release-locks.md says that `the changes you are making are not breaking backwards compatibility with any of these releases`, this even includes istio 1.0, do you know why is this required?  If it's not allowed to make such API changes, do you know how to remove the CRD generation in customresourcedefinitions.gen.yaml? Thanks.

jasonwzm:
@yangminzhu CRD generation config is in https://github.com/istio/api/blob/master/cue.yaml, remove the rbac config there.

costinm:
Is the code in istio/istio already removed ? Will an upgrade break any
older version of Istio ( if purge is set or in any other scenario ) ?
How did we verify that ?

yangminzhu:
> Is the code in istio/istio already removed ? Will an upgrade break any older version of Istio ( if purge is set or in any other scenario ) ? How did we verify that ?

@costinm , I have a pending PR to remove the CRD in istio/istio: https://github.com/istio/istio/pull/23668. What do you mean of "purge"? The upgrade is to ask users to delete the CRDs manually with kubectl when upgrading to 1.6, and this PR will only be in 1.7.

costinm:
With multi-revision - it is not safe to delete the CRDs. User may have a
1.5 or even 1.2 revision running - and
deleting the CRDs will create a major security issue.

Helm or kubectl or operator may (or may not) auto remove the CRD if 1.7
doesn't include it.
Since this can have catastrophic security impact if older versions are
running concurrently - I don't think
it is safe unless we test all install scenarios and confirm in no case the
CRD will be deleted.

yangminzhu:
> With multi-revision - it is not safe to delete the CRDs. User may have a 1.5 or even 1.2 revision running - and deleting the CRDs will create a major security issue. 
> Helm or kubectl or operator may (or may not) auto remove the CRD if 1.7 doesn't include it. Since this can have catastrophic security impact if older versions are running concurrently - I don't think it is safe unless we test all install scenarios and confirm in no case the CRD will be deleted.

Why do we support multi-revision like this (mixing 1.2 and 1.5)? 1.2, 1.3 are not supported long time ago, 1.4 will not be supported on June 5th and I don't think we support skip upgrade, see https://istio.io/news/support/.

Also the I thought the multi-revision is only supported in 1.6 which this is the only version for the supported mixing control plane upgrade, and the mixing means 1.5 and 1.6. For 1.5 -> 1.6, the user must first convert the old RBAC resources (if any) to the new AuthorizationPolicy, delete the RBAC CRDs and then upgrade to 1.6. This is a safe upgrade because 1.5 supports both RBAC and AuthorizationPolicy.

costinm:
On Mon, May 11, 2020 at 12:03 PM Yangmin Zhu <notifications@github.com>
wrote:

> With multi-revision - it is not safe to delete the CRDs. User may have a
> 1.5 or even 1.2 revision running - and deleting the CRDs will create a
> major security issue.
> Helm or kubectl or operator may (or may not) auto remove the CRD if 1.7
> doesn't include it. Since this can have catastrophic security impact if
> older versions are running concurrently - I don't think it is safe unless
> we test all install scenarios and confirm in no case the CRD will be
> deleted.
>
> Why do we support multi-revision like this (mixing 1.2 and 1.5)? 1.2, 1.3
> are not supported long time ago, 1.4 will not be supported on June 5th and
> I don't think we support skip upgrade, see https://istio.io/news/support/.
>
Can a user of 1.5 use the old API ? AFAIK the support is still in 1.5 ?

We do support 1.5 running at the same time with 1.6.

And while istio doesn't support skip-upgrade - we do know plenty of users
and other vendors (including
google) may do skip-revision upgrades - supported by the vendor directly.


> Also the I thought the multi-revision is only supported in 1.6 which this
> is the only version for the supported mixing control plane upgrade, and the
> mixing means 1.5 and 1.6. For 1.5 -> 1.6, the user must first convert the
> old RBAC resources (if any) to the new AuthorizationPolicy, delete the RBAC
> CRDs and then upgrade to 1.6. This is a safe upgrade because 1.5 supports
> both RBAC and AuthorizationPolicy.
>

It is not a safe upgrade - user may keep 1.5 running with old Authorization
policy without doing the steps.
1.5 works fine with the old policy - there is no check that they did the
steps.

howardjohn:
If the CRD is removed its ok. We already removed the other ones like MeshPolicy. 

New users will install 1.6 and cannot use MeshPolicy

Old user will install MeshPolicy in 1.x, then upgrade to 1.6. The upgrade will not prune the CRD (with istioctl and helm this is true). Once they are done with 1.5 they can delete the CRD manually if they chose

yangminzhu:
> Can a user of 1.5 use the old API ? AFAIK the support is still in 1.5 ? We do support 1.5 running at the same time with 1.6. And while istio doesn't support skip-upgrade - we do know plenty of users and other vendors (including google) may do skip-revision upgrades - supported by the vendor directly.
> It is not a safe upgrade - user may keep 1.5 running with old Authorization policy without doing the steps. 1.5 works fine with the old policy - there is no check that they did the steps.

The support in 1.5 is for migration purpose only, we have said early in 1.4 that the old RBAC policy is deprecated and will be removed in 1.6. We keep both support in 1.4 and 1.5 only to give user a smooth upgrade. With that being said, the user should have migrated to the new AuthorizationPolicy in 1.5, and there shouldn't be any issues for 1.5 -> 1.6, and this PR further removes the CRD in 1.7.

We cannot keep the old support forever just because the user is still using it, otherwise what's the point of deprecation? And if Istio has said that the skip upgrade is not supported in the first place, we shouldn't worry about it either, otherwise it basically means we cannot make any non-compatible changes because a user could be using any old istio versions.

costinm:
You can remove support for it in 1.6 - in istio/istio, no problem with that.

The problem is if installing base - with any of the install options - will
remove or not the CRD.
I know we removed CRD in the past - like MeshPolicy. Most of the time is
harmless - very few
people use MeshPolicy ( besides the default ).

If we tested that installing 1.6 base with all used methods doesn't remove
the existing CRD - all good.
AFAIK "kubectl apply --prune" would delete it. I haven't tested helm3
behavior.

howardjohn:
helm and `--prune` will not remove it. Prune will only remove certain object types, CRD is not  part of that list

costinm:
If that has been tested ( including helm3 ) - no problem with removing the
CRDs.

yangminzhu:
> If we tested that installing 1.6 base with all used methods doesn't remove
the existing CRD - all good.

@costinm this PR goes to 1.7 and 1.6 still has the CRD.

> If that has been tested ( including helm3 ) - no problem with removing the CRDs.

Tested with kubectl prune, it doesn't delete the existing CRDs, helm3 is not supported according to the install page (https://istio.io/docs/setup/install/helm/#prerequisites):
```
$ kubectl apply -f istio16.yaml --prune --all --validate=false
clusterrole.rbac.authorization.k8s.io/prometheus-istio-system configured
clusterrolebinding.rbac.authorization.k8s.io/prometheus-istio-system configured
configmap/prometheus configured
deployment.apps/prometheus configured
service/prometheus configured
serviceaccount/prometheus configured
clusterrole.rbac.authorization.k8s.io/istiod-istio-system configured
clusterrole.rbac.authorization.k8s.io/istio-reader-istio-system configured
clusterrolebinding.rbac.authorization.k8s.io/istio-reader-istio-system configured
clusterrolebinding.rbac.authorization.k8s.io/istiod-pilot-istio-system unchanged
serviceaccount/istio-reader-service-account configured
serviceaccount/istiod-service-account configured
validatingwebhookconfiguration.admissionregistration.k8s.io/istiod-istio-system configured
customresourcedefinition.apiextensions.k8s.io/httpapispecs.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/httpapispecbindings.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/quotaspecs.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/quotaspecbindings.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/destinationrules.networking.istio.io configured
customresourcedefinition.apiextensions.k8s.io/envoyfilters.networking.istio.io configured
customresourcedefinition.apiextensions.k8s.io/gateways.networking.istio.io configured
customresourcedefinition.apiextensions.k8s.io/serviceentries.networking.istio.io configured
customresourcedefinition.apiextensions.k8s.io/sidecars.networking.istio.io configured
customresourcedefinition.apiextensions.k8s.io/virtualservices.networking.istio.io configured
customresourcedefinition.apiextensions.k8s.io/workloadentries.networking.istio.io unchanged
customresourcedefinition.apiextensions.k8s.io/attributemanifests.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/handlers.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/instances.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/rules.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/authorizationpolicies.security.istio.io configured
customresourcedefinition.apiextensions.k8s.io/peerauthentications.security.istio.io configured
customresourcedefinition.apiextensions.k8s.io/requestauthentications.security.istio.io configured
customresourcedefinition.apiextensions.k8s.io/adapters.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/templates.config.istio.io configured
customresourcedefinition.apiextensions.k8s.io/istiooperators.install.istio.io unchanged
horizontalpodautoscaler.autoscaling/istio-ingressgateway configured
deployment.apps/istio-ingressgateway configured
poddisruptionbudget.policy/istio-ingressgateway created
role.rbac.authorization.k8s.io/istio-ingressgateway-sds configured
rolebinding.rbac.authorization.k8s.io/istio-ingressgateway-sds configured
service/istio-ingressgateway configured
serviceaccount/istio-ingressgateway-service-account configured
horizontalpodautoscaler.autoscaling/istiod configured
configmap/istio configured
deployment.apps/istiod configured
configmap/istio-sidecar-injector configured
mutatingwebhookconfiguration.admissionregistration.k8s.io/istio-sidecar-injector configured
poddisruptionbudget.policy/istiod configured
service/istiod configured
envoyfilter.networking.istio.io/metadata-exchange-1.4 configured
envoyfilter.networking.istio.io/stats-filter-1.4 configured
envoyfilter.networking.istio.io/metadata-exchange-1.5 configured
envoyfilter.networking.istio.io/tcp-metadata-exchange-1.5 configured
envoyfilter.networking.istio.io/stats-filter-1.5 configured
envoyfilter.networking.istio.io/tcp-stats-filter-1.5 configured
envoyfilter.networking.istio.io/metadata-exchange-1.6 created
envoyfilter.networking.istio.io/tcp-metadata-exchange-1.6 created
envoyfilter.networking.istio.io/stats-filter-1.6 created
envoyfilter.networking.istio.io/tcp-stats-filter-1.6 created
envoyfilter.networking.istio.io/metadata-exchange-1.7 created
envoyfilter.networking.istio.io/tcp-metadata-exchange-1.7 created
envoyfilter.networking.istio.io/stats-filter-1.7 created
envoyfilter.networking.istio.io/tcp-stats-filter-1.7 created
configmap/pilot-envoy-config pruned
service/istio-pilot pruned
namespace/istio-system pruned
```

yangminzhu:
/hold

Probably need https://github.com/istio/api/issues/1435 to be fixed

yangminzhu:
@dcberg , @duderino , @linsun , @louiscryan , @rshriram , @smawson Could you take a look, this removes the v1alpha1 RBAC API completely, note this will only be in 1.7 and doesn't affect 1.6. Thanks.

PR #1437, opened 2020-05-16
Hide localhost TLS unimplemented APIs

nrjpoddar:


rshriram:
You might as well remove it. Someone will find out  from the protos and try to use it.

nrjpoddar:
> You might as well remove it. Someone will find out from the protos and try to use it.

@rshriram I don't have time today to do it, if you have please go for it. At the very least I wanted to hide it from docs.

PR #1443, opened 2020-05-23
Add the missing 'none'

costinm:
It is possible to set it via annotation - but not in mesh config, as default. 

To make things worse - the 'whitebox' mode in Sidecar is only possible with interception=none. 

We should probably backport it to 1.6.1 too - we only tested with the annotation.


brian-avery:
@costinm what ever happened to this PR? It looks like Sven approved it back in March 2020, but that the gencheck never got fixed.

PR #1446, opened 2020-05-26
Add selector to Ingress

howardjohn:
This is for https://github.com/istio/istio/issues/24055

This addresses a regression blocking some users from upgrading from 1.5 -> 1.6, so I would like to get this cherrypick to 1.6 as well

Implementation PR: https://github.com/istio/istio/pull/24111


howardjohn:
/test all

not sure why tests didn't start

mesh/v1alpha1/config.proto
```diff
@@ -130,6 +130,12 @@ message MeshConfig {
   // Defines whether to use Istio ingress controller for annotated or all ingress resources.
   IngressControllerMode ingress_controller_mode = 9;
 
+  // Defines which gateway to use as the Ingress controller. This field corresponds to
```
costinm, line 133: I don't get it - we already have IngressService - the name of the k8s service used by the ingress controller, which gets populated in the Status of the Ingress resource. 

Why would we use labels ?

howardjohn: Service != workload (although it often is). We need to get workload labels in order to create a Gateway in order to send the config to some specific Envoy workload. A Service may select part of a workload or multiple different workloads with distinct config.

The IngressService is just for the LB to update the IP address in status.

We could do an approximation by looking at the Service and then finding all pods selected by the service and trying to somehow get the labels to use for the Gateway, but it would be neither accurate nor simple

costinm: We must have a ServiceName - so we know what IP to put in the Ingress.status. 

If we have a ServiceName - it would be crazy to use a different set of labels, or to duplicate the labels. If anything needs the labels - we can get them from the Service directly to make sure they don't get out of sync.

howardjohn: Actually I think I may agree. I was thinking we would have to go from Service -> pods -> labels but we can just look at the `Service.spec.selector` and use the same for `Gateway.spec.selector`. Its still not 100% accurate but should be close enough for 99.9% of uses

costinm: Also - the bug is about using the Ingress config ( and I presume will apply to K8SGateway class ). 

There is a 1-1 mapping between Ingress class and the gateway.  CertManager for example is configured with the class - not with any istio-specific labels. 

The class determines which Istiod revision is used. It seems this part is broken due to master election, we can't seem to support multiple classes right now, but should be easy to fix and no need for an API change.

I don't see the practical benefit of this API - but quite a lot of opportunity for user to mess things up and extra complexity for setup.

howardjohn: > There is a 1-1 mapping between Ingress class and the gateway

The mapping is 1:1 between Ingress and Gateway. This is to allow SDS, not using `*` host, etc. We could use IngressClass to configure all of these options but its not available in K8s 1.17- so I don't think we can expect users to have this (yet).

I am not sure why you would have Class determine revision?

costinm: IngressClass is already available - has been available for a very long time. That's how CertManager selects Istio ingress instead of Nginx or others, and how an Istiod knows which Ingresses to support. 

Current code is still implemented assuming 1 Istiod managing 1 ingress class and handled by 1 ingress gateway. 

We do have the bug with the master election that currently prevents having multiple ingress classes - which is a regression, but easy to fix. 

In any case - the existing API should cover all the cases correctly. MeshConfig has a config to specify which ingress class to watch for, and which Gateway to associate with it. And that's all that a user needs.

costinm: The new K8S Gateway 'class' is the one that is not available - I am talking about the v1.Ingress CRD and  kubernetes.io/ingress.class: "istio".

howardjohn: > and which Gateway to associate with it

It does not have this -- that is what I am adding. It only has ingress service which is not the same thing. As mentioned above we can approximate "which Gateway to associate with it" with ingress service but its not actually correct.

I am not sure why/how it would make sense to have revision and class 1:1? If I understand this correctly, I think this would mean that:
* An ingress cannot apply to multiple revisions
* I need to update all my Ingress in order to update istio

Why not use the existing mechanism to control this (`istio.io/rev`) which already handles this?


Also IngressClass is related but not the same as GatewayClass, see https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class

costinm: I'm not sure why you want to add the label selection when VirtualService uses namespace/NAME, and the ingressService is also identified by name and has worked quite well. 

Can you describe the use case - when would a user customize the labels instead of using the gateway name.

For revisions - currently the code is written such that one Istiod deployment is watching only a single 'ingressClass' - and the config in meshConfig is also for a single class. We could change this so an Istiod revision can watch multiple classes - but up to 1.6 the config and implementations are based on this assumption.

On the other direction - it is probably possible to have N revisions watch the same class. In fact - that's how 1.6 seems to work, since there is a single master election. In the past I think we had a master election per ingress class - so you could have each multiple classes, with each revision handling 1 ingress class. That is currently broken, and should be fixed if we need to support multiple ingress classes. 

You certainly don't need to update Ingress objects to update Istio - all you need is to specify 
in the Ingress that you want to use Istio ( or have istio act as default ingress controller ). 
BTW - Istio as 'default' ingress controller is obviously singleton - and only on 'revision' should handle it. 

We could very well add support the istio.io/rev annotation on Ingress objects - that's perfectly fine feature. But even if this is done - it would only indicate which Istiod is managing that specific Ingress object. The actual Gateway IP would still be selected by name.

IngressClass/GatewayClass and the class annotations are related and not the same - but the annotation is what we have supported and what most other ingress controllers support (AFAIK). We'll migrate to the new ones as well.

costinm: TL;DR:  Gateway should be selected by name, like we do in VirtualService - and that's already supported.  Users should not care how we label the gateway workloads (or if any such gateway 'workload' even exists - it could as well be an out-of-cluster managed service).

howardjohn: In the ingress code we create a Gateway. This Gateway has selector (field https://istio.io/docs/reference/config/networking/gateway/#Gateway, not related to VS `ns/name`) which is hardcoded to `istio=ingressgateway`. Users with a custom deployment of ingress will not have a pod (or will not want to use) with `istio=ingressgateway` label. Instead, they want this gateway to get the selector set to `istio=my-custom-deployment` or whatever.

In previous code we had a Gateway created at install time (with a name similar to `istio-autogenerated-k8s-gateway`) which a user could (manually) modify to get this working. However, this had major limitations around hostnames/tls and conflicts with https://github.com/istio/istio/issues/22095 so it was removed as part of #22227. I think I recall discussing this very use case with you and you mentioned you didn't think anyone would use it and if users complain we can add it to the API which is what I am doing here.

serrien: Hello,
I'm facing the issue with cert manager http resolver than #24055. 
@howardjohn if I understand, a workaround may be to label the istio-ingressgateway-mine gateway intended for public traffic with istio: ingressgateway.
Am i right ?

howardjohn: @serrien yes that is what would be required today

serrien: @howardjohn I just tested and it works. Thanks for your explanations in comments.

costinm: The hardcoded part obviously needs to change in order to support multiple gateways. But it doesn't need to be an arbitrary label selection - it can still be based on the name of the gateway and match the name of the service.

We don't have any benefit in allowing arbitrary labels and having different names. There is still 1 IP /service that we need to update, and no reason to use multiple names that may or may not match. 

You create an ingress named "my-custom-deployment" - this should create a LB service named "my-custom-deployment", matching the ingressService in the config - and set as label in the synthetic Gateway ( istio=my-custom-deployment). And should also match the names in VirtualService - istio-system/my-custom-deployment. One name to rule them all.

costinm: On second part - I agree, we don't need to istio-autogenerated stuff. But still no need to go too far in 'arbitrary' labels and names. Just need to have the auto-generated VirtualService use 
'my-custom-deployment' as name.

costinm: And to be clear: reality in 1.6 is that we really support 1 K8S Ingress, and that must use the labels we use by default. 

Most of my comments are about 1.7+ - if we really want to enable old-style K8S Ingess with more than one class ( istio ). We may need to do this for the new K8S Gateway ( maybe !).

howardjohn: @costinm I would agree that users *should* use the same labels for the deployment and service, but do we really want to enforce that they *must*? Because there is absolutely no requirement that a service is 1:1 with a deployment or that they share the same set of labels.


> You create an ingress named "my-custom-deployment" - this should create a LB service named "my-custom-deployment", matching the ingressService in the config - and set as label in the synthetic Gateway ( istio=my-custom-deployment). And should also match the names in VirtualService - istio-system/my-custom-deployment. One name to rule them all.

This is not how this works today, see my above comments/PR. There has nothing to do with matching a VirtualService to a Gateway, this is about matching the synthetic Gateway to the deployment. This is about configuring the labels for the synthetic Gateway so we can do the "set as label in the synthetic Gateway" you mentioned

> And to be clear: reality in 1.6 is that we really support 1 K8S Ingress, and that must use the labels we use by default.

I am not sure what you mean by "1 K8s Ingress" but:
* Multiple ingressgateway deployments is both supported and widely used (by knative as well).
* Multiple `Ingress` objects is also widely used and clearly supported
* Multiple `ingressClass` is not but its pretty orthogonal to this conversation I think


It seems to me we have two options:
1) This PR
2) Lookup the selector on the ingressService and use that as the selector on the synthetic Gateway. This has the benefit of being 1 less config, with the downside that its less flexible/incorrect (service selector need not be the same as gateway selector) and that it is more complicated in implementation

Are you suggesting we go with (2) or something else?

serrien: @costinm "reality in 1.6 is that we really support 1 K8S Ingress, and that must use the labels we use by default." Yes, but there is nothing in the documentation about this. Nevertheless, I prefer the use of a label by default instead of the istio-autogenerated-k8s-gateway

costinm: Yes, we should have better docs or at least samples on how to use K8S Ingress - and in particular CertManager integration. 

You should never have to use istio-autogeneratd-blah - where did you use it ? It should be
an internal name only.

howardjohn: I have added both of these docs in 1.6 https://istio.io/docs/tasks/traffic-management/ingress/kubernetes-ingress/ and https://istio.io/docs/ops/integrations/certmanager/

The autogenerated thing was not an internal name prior to Istio 1.6 (meaning it was persisted as a real Gateway to the api-server). Now it actually is internal so it's not customizable for the selector so we need this pr to get back feature parity

costinm: > @costinm I would agree that users _should_ use the same labels for the deployment and service, but do we really want to enforce that they _must_? Because there is absolutely no requirement that a service is 1:1 with a deployment or that they share the same set of labels.

Users apply the templates we provide. We set the labels as needed. We should make no guarantee or promise that if you use a Gateway - some deployment with some labels will be created. It may or may not be, it's an implementation detail. 

In general the Service may use broader labels - for example if we add a 'canary' gateway, so
there is no requirement for 1:1 mapping - but that's our internal problem, should not be exposed to users. 


> 
> > You create an ingress named "my-custom-deployment" - this should create a LB service named "my-custom-deployment", matching the ingressService in the config - and set as label in the synthetic Gateway ( istio=my-custom-deployment). And should also match the names in VirtualService - istio-system/my-custom-deployment. One name to rule them all.
> 
> This is not how this works today, see my above comments/PR. There has nothing to do with matching a VirtualService to a Gateway, this is about matching the synthetic Gateway to the deployment. This is about configuring the labels for the synthetic Gateway so we can do the "set as label in the synthetic Gateway" you mentioned

Well - yes, VirtualService is referencing the synthetic Gateway by name, and the synthetic Gateway is using the labels. And the only tested/supported config is to match the default
gateway - at least in 1.6.

> 
> > And to be clear: reality in 1.6 is that we really support 1 K8S Ingress, and that must use the labels we use by default.
> 
> I am not sure what you mean by "1 K8s Ingress" but:
> 
> * Multiple ingressgateway deployments is both supported and widely used (by knative as well).
> * Multiple `Ingress` objects is also widely used and clearly supported
> * Multiple `ingressClass` is not but its pretty orthogonal to this conversation I think

Well, you are technically correct, if you install Istiod in a different namespace you can have 
multiple ingress classes. That's what knative is doing AFAIK. I don't think this is a config
we support for the 'general' users - but people who want to run completely separate namespace and run their own testing can do this.

Unless something changed - in istio-system we are now using a master election per namespace, so I highly doubt you could get 2 classes working without fixing that. Nor do I think it was ever tested in same ns. 

> 
> It seems to me we have two options:
> 
> 1. This PR

How is this PR fixing anything ? We need to fix the master election and remove the hardcoded labels - allowing users to go crazy in configuring arbitrary labels just add complexity and testing and is likely to break something.

If the feature is "allow multiple ingress classes" - we can do that without going from a 
hardcoded constant to arbitrary label selection. Using the Gateway/Service name is 
simpler. 

> 2. Lookup the selector on the ingressService and use that as the selector on the synthetic Gateway. This has the benefit of being 1 less config, with the downside that its less flexible/incorrect (service selector need not be the same as gateway selector) and that it is more complicated in implementation

Or 

3: use the same name for the gateway deployment ( values.gateways.foo ), className, Service. Easy to implement, easy to debug/troubleshoot, deterministic. Or if you don't want 
to use the same name - use some prefix, but keep it deterministic, i.e. based on the same identifier. 

> 
> Are you suggesting we go with (2) or something else?

To be honest - I'm suggesting neither of the above, just handle this for the new K8S Gateway 
and class - and document that old-style Ingress only supports a class ( per tenant/istio namespace ). It's not clear this is such a common feature request.

costinm:
If pre-Istio 1.6 we had an yaml file that defined the Gateway doesn't mean
we ever supported arbitrary labels or gateways.
We barely have tests for basic Ingress and install usage. Sure, everything
in the yaml can be changed by a user - it doesn't mean
it's supported.

The reason we used the yaml was to allow users to specify secrets in a more
flexible way. I'm still not entirely sure if we have
feature parity on that area - I did a small test using the 'default' cert
and it worked fine. That is something we should absolutely test
and fix, ability to use "spec.tls" and be able to use different certs for
different domains.

Let's start with requirements and use case. "I just want to arbitrarily set
labels because it was possible in 1.5" is not a use case.
Do they need multiple classes - in istio-system ? Then we need to fix
master election and probably revamp the config, make it a map or something.

Do they want to have multiple Service - so they get different IPs ?

costinm:
The TLS part seems to work fine, I tested with a tls.hosts=foo and it works
using SDS.

mesh/v1alpha1/config.proto
howardjohn: > And the only tested/supported config is to match the default gateway - at least in 1.6.

This is exactly the problem this PR is fixing. Prior to 1.6 users *were* matching a non-default gateway - in 1.6 this ability was removed, hence the original issue where this is no longer configurable.

This is not even users modifying arbitrary yaml, its a part of the helm "api" - see snippet below from Istio 1.5:

```yaml
kind: Gateway
metadata:
  name: istio-autogenerated-k8s-ingress
spec:
  selector:
    istio: {{ .Values.global.k8sIngress.gatewayName }}
```

> The reason we used the yaml was to allow users to specify secrets in a more flexible way. I'm still not entirely sure if we have feature parity on that area

We have tests for this, see

https://github.com/istio/istio/blob/9dd64b43f7fd7750eca86c940bd13c8255fb29ab/tests/integration/pilot/ingress/ingress_test.go#L222

> Let's start with requirements and use case. "I just want to arbitrarily set labels because it was possible in 1.5" is not a use case.

The use case is simple: to support custom gateway deployments. This is extremely common from what I have seen (knative does this for example). Currently if you have a custom gateway there is no way to mark this as the gateway to use for `Ingress`. This means, for example, you could not use an ILB for ingess without also changing the default ingressgateway deployment which you may not want.

I do not fully understand why you keep mentioning Ingress class, leader election, and revisions. This is (in my mind) completely orthogonal to what this change is about. This is not about running multiple different IngressClasses/Ingress controllers at the same time in the same cluster - this is just about modifying the single ingress class/controller to use an alternative gateway deployment.

costinm: > > And the only tested/supported config is to match the default gateway - at least in 1.6.
> 
> This is exactly the problem this PR is fixing. Prior to 1.6 users _were_ matching a non-default gateway - in 1.6 this ability was removed, hence the original issue where this is no longer configurable.

Sure - but that non-default gateway existed not because the intent was to support arbitrary
things, but to allow TLS configs. We never tested - or assumed - multiple K8S Ingress 
gateways ( except the advanced isolation like knative / early multi-tenant )


> 
> This is not even users modifying arbitrary yaml, its a part of the helm "api" - see snippet below from Istio 1.5:
> 
> ```yaml
> kind: Gateway
> metadata:
>   name: istio-autogenerated-k8s-ingress
> spec:
>   selector:
>     istio: {{ .Values.global.k8sIngress.gatewayName }}
> ```

Ok - to restore parity to this we just need exactly what I mentioned earlier, use the 
gatewayName :-) That's not an arbitrary label matcher.

And the same helm templates generate the Service with the same gatewayName. Simple.


> 
> > The reason we used the yaml was to allow users to specify secrets in a more flexible way. I'm still not entirely sure if we have feature parity on that area
> 
> We have tests for this, see
> 
> https://github.com/istio/istio/blob/9dd64b43f7fd7750eca86c940bd13c8255fb29ab/tests/integration/pilot/ingress/ingress_test.go#L222
> 
> > Let's start with requirements and use case. "I just want to arbitrarily set labels because it was possible in 1.5" is not a use case.
> 
> The use case is simple: to support custom gateway deployments. This is extremely common from what I have seen (knative does this for example). Currently if you have a custom gateway there is no way to mark this as the gateway to use for `Ingress`. This means, for example, you could not use an ILB for ingess without also changing the default ingressgateway deployment which you may not want.


Sounds good - and AFAIK this is already possible if you use a separate namespace (knative
still works), or with the simpler model of using the gatewayName == ingressService instead
of the hardcoded value. 

The use case doesn't require arbitrary label matching. 


> 
> I do not fully understand why you keep mentioning Ingress class, leader election, and revisions. This is (in my mind) completely orthogonal to what this change is about. This is not about running multiple different IngressClasses/Ingress controllers at the same time in the same cluster - this is just about modifying the single ingress class/controller to use an alternative gateway deployment.


I mention this because that seems to be what is preventing what you want - besides replacing the hardcoded name with the ingressService. 

We can deploy a second gateway - say 'internalGateway' with type ILB. That will create 
an internalGateway service, which can be used in an Istiod revision. If the new gateway is pointing to the second istiod revision - most will work. You can use both Ingress with class istio that is exposed on the internet ( default gateway ) and with class 'internalGateway' that 
will only be exposed by the second ingress (via second istiod revision). However if I read the 
code correctly, the 2 istiod revision will participate in the same master election - and a random one will be selected to update the Ingress.status. 

For 1.6 - running them in a separate namespace is probably the safest option.

costinm:
BTW - the reason I'm pushing back on this PR is that I don't think we should complicate the old-k8sIngress code. It's relatively stable, we need it for CertManager and few other things - but if people want to do advanced things they should use Istio APIs ( or the new K8S Gateway). 

Multi-revision in Gateway is quite tricky - since the external IP is allocated to the Service, so we'll probably need to do some fancy changes. After we figure that out - we may worry again about 
multiple legacy Ingress in same namespace.

mesh/v1alpha1/config.proto
howardjohn: > Ok - to restore parity to this we just need exactly what I mentioned earlier, use the
gatewayName :-) That's not an arbitrary label matcher. And the same helm templates generate the Service with the same gatewayName. 

The label is `ingressgateway`, the class is `istio`, and the Service is `istio-ingressgateway`. So we cannot really just use the same value for everything. I would be perfectly fine with just having exactly what we had earlier and use gatewayName. However, we need a place to configure it, which would be mesh config. That would basically mean changing this from `gatewaySelector` to `gatewayName`, and in the implementation instead of using `gatewaySelector` we use `map[string]string{"istio": meshConfig.gatewayName}` - I have no problem changing that.

> We can deploy a second gateway - say 'internalGateway' with type ILB. That will create
> an internalGateway service, which can be used in an Istiod revision. If the new gateway is pointing to the second istiod revision - most will work. You can use both Ingress with class istio that is exposed on the internet ( default gateway ) and with class 'internalGateway' that
> will only be exposed by the second ingress (via second istiod revision). However if I read the
> code correctly, the 2 istiod revision will participate in the same master election - and a random one will be selected to update the Ingress.status.

In my view this PR actually gets us closer to this model. Currently, aside from master election breaking things, you could not do this because there is no way to map the internalGateway class to the ILB.

serrien:
> "I just want to arbitrarily set
labels because it was possible in 1.5" is not a use case.
The use case is clear, as an administrator of my k8s cluster, when I configure multiple Istio ingress gateways, then I need to a way to configure which ingress gateway will be configured when a specific k8s ingress is deployed. This use case was possible prior to 1.6. From k8s administrator point of view, I choose the name/labels for my Istio ingress gateway so it seem legit that I can configure the name/label to use. The hardcoded label is a workaround, but I have no guarantee that he will not change in future releases.

costinm:
On Tue, May 26, 2020 at 11:04 PM John Howard <notifications@github.com>
wrote:

> *@howardjohn* commented on this pull request.
>
> <https://github.com/istio/api/pull/1446#discussion_r430845075>:
>
> > @@ -130,6 +130,12 @@ message MeshConfig {
>    // Defines whether to use Istio ingress controller for annotated or all ingress resources.
>    IngressControllerMode ingress_controller_mode = 9;
>
> +  // Defines which gateway to use as the Ingress controller. This field corresponds to
>
> Ok - to restore parity to this we just need exactly what I mentioned
> earlier, use the
> gatewayName :-) That's not an arbitrary label matcher. And the same helm
> templates generate the Service with the same gatewayName.
>
> The label is ingressgateway, the class is istio, and the Service is
> istio-ingressgateway. So we cannot really just use the same value for
> everything. I would be perfectly fine with just having exactly what we had
> earlier and use gatewayName. However, we need a place to configure it,
> which would be mesh config. That would basically mean changing this from
> gatewaySelector to gatewayName, and in the implementation instead of
> using gatewaySelector we use map[string]string{"istio":
> meshConfig.gatewayName} - I have no problem changing that.
>

We can use the same value for everything - except the default gateway. Just
like the default istiod doesen't have a revision, and does all kind of
things differently -
we can't go back in time and fix it.  Unfortunately the service name can't
be changed without upgrade issues for the default, we could use a new label
with value 'istio'.
BTW - if we go this route, it would be great if the SA would also have the
same name, without the -service-account suffix ( it has a Kind already ).

Now: if you find any good reason for the gatewayName to be different from
the service name - I can live with an API change to add the extra
gatewayName.
My primary concern was about using arbitrary labels - when VirtualService
uses name.  Maybe do an initial implementation using same name, and
if someone complains we can change the API later ?


We can deploy a second gateway - say 'internalGateway' with type ILB. That
> will create
> an internalGateway service, which can be used in an Istiod revision. If
> the new gateway is pointing to the second istiod revision - most will work.
> You can use both Ingress with class istio that is exposed on the internet (
> default gateway ) and with class 'internalGateway' that
> will only be exposed by the second ingress (via second istiod revision).
> However if I read the
> code correctly, the 2 istiod revision will participate in the same master
> election - and a random one will be selected to update the Ingress.status.
>
> In my view this PR actually gets us closer to this model. Currently, aside
> from master election breaking things, you could not do this because there
> is no way to map the internalGateway class to the ILB.
>

If you have a PR fixing master election - it shouldn't be hard to also
change the hardcoded 'istio:ingressgateway' to
if (istioService != "istio-ingressgateway" ) ... 'istio:' +
mesh.istioService.
And fix the templates for gateway to not add the "istio-" prefix to the
service.

If you want to have more consistency to current default gateway - you can
keep adding an istio- prefix - so for 'ilb' gateway we would have
label='ilb', class='ilb', Service: istio-ilb.

The 'class' used in K8S.Ingress objects would be the same with the gateway
name used in VirtualServices and the name used when
installing the gateway, which is good consistency. People keep complaining
about complicated install and debugging - think of all the
istioctl code to debug a problem if we need to map multiple names and
labels or explain to a user how to get logs or figure out
status.

costinm:
On Wed, May 27, 2020 at 2:51 AM Errien <notifications@github.com> wrote:

> "I just want to arbitrarily set
> labels because it was possible in 1.5" is not a use case.
> The use case is clear, as an administrator of my k8s cluster, when I
> configure multiple Istio ingress gateways, then I need to a way to
> configure which ingress gateway will be configured when a specific k8s
> ingress is deployed. This use case was possible prior to 1.6. From k8s
> administrator point of view, I choose the name/labels for my Istio ingress
> gateway so it seem legit that I can configure the name/label to use. The
> hardcoded label is a workaround, but I have no guarantee that he will not
> change in future releases.
>
>
For the first part: I agree, and AFAIK it is still possible in 1.6 if you
install the new gateway in a separate namespace. Which everyone should do
for security reasons -
the gateways are exposed to the internet - and run with permissions to read
all "Secret" objects in the namespace - including the Istiod CA root key.
So running each gateway in a separate namespace will reduce the blast
radius - and also isolate Gateways from each other.

Unfortunately we have a pair of bugs that prevents running multiple K8S
Ingress gateways in the same namespace in 1.6, it can be fixed.

For the second part: if you use the supported installer, I don't think you
should choose labels for installed objects - just the name of the gateway.
If you are a vendor implementing a custom installer - and
testing&supporting it - you can choose any labels as long as it works...

howardjohn:
> For the first part: I agree, and AFAIK it is still possible in 1.6 if you
> install the new gateway in a separate namespace.

How can you do this? The Gateway will get `istio: ingressgateway` selector and match the default gateway.

I have not heard of any user want to have multiple different gateways implementing Ingress - only users wanting a different gateway. So I don't think we should focus on this.

Yes it would be nice if everything lined up. The reality is that it doesn't, and I don't think its reasonable to ask people to rename everything. For example, if you rename the Service you will get a new LB IP which isn't feasible. 

I have updated the PR to just allow setting the `istio` value rather than arbitrary labels. Hopefully that can move forward. I also updated the implementation in https://github.com/istio/istio/pull/24111

costinm:
On Wed, May 27, 2020 at 9:33 AM John Howard <notifications@github.com>
wrote:

> For the first part: I agree, and AFAIK it is still possible in 1.6 if you
> install the new gateway in a separate namespace.
>
> How can you do this? The Gateway will get istio: ingressgateway selector
> and match the default gateway.
>

If it matches across namespaces - we have a bigger bug. That was the reason
we switched from VirtualService selecting gateway
by label to select by namespace and name.


I have not heard of any user want to have multiple different gateways
> implementing Ingress - only users wanting a different gateway. So I don't
> think we should focus on this.
>

Agreed.


> Yes it would be nice if everything lined up. The reality is that it
> doesn't, and I don't think its reasonable to ask people to rename
> everything. For example, if you rename the Service you will get a new LB IP
> which isn't feasible.
>
Agreed as well - but for new things we should be able to restore
consistency and simplify.

In reality the Service name is the hardest to upgrade, but AFAIK it is
already matching the gateway install name. So if we want
to support multiple K8S.Ingress classes - which is no broken in 1.6 - we
still can use the gateway install name as class.


> I have updated the PR to just allow setting the istio value rather than
> arbitrary labels. Hopefully that can move forward. I also updated the
> implementation in istio/istio#24111
> <https://github.com/istio/istio/pull/24111>
>
Can you merge the implementation fix using the 'class' initially - while we
move this discussion to the WG ? Seems we are getting closer, but
clearly it should get more eyes.
If we want to backport to 1.6 - API changes are usually not done in dot
releases. We could, but better avoid.

PR #1449, opened 2020-05-28
Add Status types to API

therealmitchconnors:
Add types for Status, including conditions and analysis messages.


jasonwzm:
So we would like the Status feature to be in beta but the API to be in alpha state?

therealmitchconnors:
Hmmm... no.  I just copied the Analysis protos from clay's PR. but I'll move them to beta...

analysis/v1beta1/message.proto
```diff
@@ -0,0 +1,144 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+// $title: Analysis Messages
```
rshriram, line 17: hide from docs?

Makefile.core.mk
```diff
@@ -396,6 +398,50 @@ generate-security: $(security_v1beta1_pb_gos) $(security_v1beta1_pb_docs) $(secu
...
 
+#####################
+# analysis/...
+#####################
+
+analysis_v1beta1_path := analysis/v1beta1
+analysis_v1beta1_protos := $(wildcard $(analysis_v1beta1_path)/*.proto)
+analysis_v1beta1_pb_gos := $(analysis_v1beta1_protos:.proto=.pb.go)
+analysis_v1beta1_pb_pythons := $(patsubst $(analysis_v1beta1_path)/%.proto,$(python_output_path)/$(analysis_v1beta1_path)/%_pb2.py,$(analysis_v1beta1_protos))
+analysis_v1beta1_pb_docs := $(analysis_v1beta1_protos:.proto=.pb.html)
```
rshriram, line 409: does this need docs at all? its for code level usage isnt it?

Makefile.core.mk
```diff
@@ -396,6 +398,50 @@ generate-security: $(security_v1beta1_pb_gos) $(security_v1beta1_pb_docs) $(secu
...
+analysis_v1beta1_openapi := $(analysis_v1beta1_protos:.proto=.gen.json)
```
rshriram, line 410: why do we need the open api thing? is this struct going to be persisted to crd by end user or is it going to be used only by istiod?

Makefile.core.mk
```diff
@@ -396,6 +398,50 @@ generate-security: $(security_v1beta1_pb_gos) $(security_v1beta1_pb_docs) $(secu
...
+# meta/...
+#####################
+
+meta_v1beta1_path := meta/v1beta1
+meta_v1beta1_protos := $(wildcard $(meta_v1beta1_path)/*.proto)
+meta_v1beta1_pb_gos := $(meta_v1beta1_protos:.proto=.pb.go)
+meta_v1beta1_pb_pythons := $(patsubst $(meta_v1beta1_path)/%.proto,$(python_output_path)/$(meta_v1beta1_path)/%_pb2.py,$(meta_v1beta1_protos))
+meta_v1beta1_pb_docs := $(meta_v1beta1_protos:.proto=.pb.html)
+meta_v1beta1_openapi := $(meta_v1beta1_protos:.proto=.gen.json)
+
```
rshriram, line 433: same questions as above

Makefile.core.mk
jasonwzm: I think we just followed the targets of other APIs. We have been providing the json schema as a file for end users. 
The json schema will also be used by CRD for field validation.

rshriram: Yes but we dont want to document our internal stuff in istio docs do we?

Makefile.core.mk
therealmitchconnors: This was documented in 1.6, and exposed using untyped objects.  This PR takes what already exists in 1.6 and formalizes its structure.

PR #1452, opened 2020-05-29
doc: correct the comment of Gateway

knight42:
The scope of label search is not only restricted to the configuration namespace in which the Gateway resource is present, the current comment may confuse users.

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -383,6 +383,7 @@ message Gateway {
   // label search is restricted to the configuration namespace in which the
   // the resource is present. In other words, the Gateway resource must
   // reside in the same namespace as the gateway workload instance.
+  // If selector is nil, the Gateway will be applied to all workloads.
   map<string, string> selector = 2 [(google.api.field_behavior) = REQUIRED];
 }
 
--- a/networking/v1alpha3/envoy_filter.pb.go
+++ b/networking/v1alpha3/envoy_filter.pb.go
@@ -27,7 +27,7 @@
 // sequentially in order of creation time.  The behavior is undefined
 // if multiple EnvoyFilter configurations conflict with each other.
 //
-// **NOTE 3**: *_To apply an EnvoyFilter resource to all workloads
+// **NOTE 3**: To apply an EnvoyFilter resource to all workloads
 // (sidecars and gateways) in the system, define the resource in the
 // config [root
 // namespace](https://istio.io/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig),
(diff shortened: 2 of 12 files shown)


knight42:
@rshriram if that is not true, how would you explain this example https://istio.io/docs/setup/getting-started/#ip in which `bookinfo-gateway` Gateway is created in `default` namespace while the `istio-ingressgateway` is in the `istio-system` namespace?

howardjohn:
It is just searching globally, its not looking for the root namespace

knight42:
@howardjohn Thanks for the explanation. After digging into the code, I found that  we can limit the gateway workload search scope by setting env var `PILOT_SCOPE_GATEWAY_TO_NAMESPACE`, should I mention that?

knight42:
@howardjohn @rshriram please check again!

networking/v1alpha3/gateway.proto
```diff
@@ -379,10 +379,13 @@ message Gateway {
...
-  // label search is restricted to the configuration namespace in which the
-  // the resource is present. In other words, the Gateway resource must
-  // reside in the same namespace as the gateway workload instance.
+  // on which this gateway configuration should be applied.
+  // Depending on the environment variable `PILOT_SCOPE_GATEWAY_TO_NAMESPACE` passed to pilot,
+  // if it is set to "false"(the default value), a gateway workload will select all gateway
+  // resources in all namespaces matching the given label selector,
+  // if it is set to "true", a gateway workload can only select gateway resources in the same namespace.
+  // In addition, if selector is nil, this gateway will be applied to all workloads
+  // asking for the gateway.
```
frankbu, line 388: This description seems pretty complicated. Do we need to document the `PILOT_SCOPE_GATEWAY_TO_NAMESPACE` env variable. There is no pilot anymore. Also, it's not clear what "workloads asking for the gateway" means. Maybe "all gateway workloads", but then again, do we want to document that behavior, since it's not really something we would recommend.

knight42: > There is no pilot anymore

yes, but istiod still respects this env var.

> do we want to document that behavior, since it's not really something we would recommend.

@howardjohn @rshriram any thoughts?

rshriram: So I was deliberately omitting the env var as we don't want people to be creating gateway objects that refer to workloads in other namespaces. At some point, we can be reasonably sure that everyone is doing the right thing. It then becomes easy for us to completely remove the option from istiod and always require gateway objects to be in same namespace as the workload.

knight42: Got it. So I decide to restore to the original comment and mention that if selector is nil, the gateway will be applied to all workloads in the same namespace.

knight42: Please take another look.

PR #1453, opened 2020-05-30
CORS Policy Example: Change allowOrigins block

dntosas:
### Description
In this PR, we change the block of CORS Policy definition from `allowOrigin` to `allowOrigins` so not to be misleading for people who take this example as guide to setup their meshes.

After the upgrading to Istio 1.6, the block with deprecated `allowOrigin` is not respected thus not working.

### Addresses
https://github.com/istio/istio/issues/24145

### DIFF
--- a/networking/v1beta1/virtual_service.proto
+++ b/networking/v1beta1/virtual_service.proto
@@ -1579,8 +1579,8 @@ message HTTPRetry {
 //         host: ratings.prod.svc.cluster.local
 //         subset: v1
 //     corsPolicy:
-//       allowOrigin:
-//       - example.com
+//       allowOrigins:
+//       - exact: https://example.com
 //       allowMethods:
 //       - POST
 //       - GET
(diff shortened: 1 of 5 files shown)


networking/v1beta1/virtual_service.proto
```diff
@@ -1579,8 +1579,8 @@ message HTTPRetry {
 //         host: ratings.prod.svc.cluster.local
 //         subset: v1
 //     corsPolicy:
-//       allowOrigin:
-//       - example.com
+//       allowOrigins:
+//       - exact: https://example.com
```
rshriram, line 1583: is this scheme needed here?

dntosas: hello @rshriram 

yes, it is actually needed. i made several tests with and without scheme and results showed that when we use `exact` as key, scheme is mandatory ^^

PR #1463, opened 2020-06-09
Move termination drain duration

howardjohn:
This was merged yesterday so no risk reverting. This moves from mesh
config to proxy config, where it should have originally been. This is a
proxy specific config.


howardjohn:
@smawson can you take a look?

PR #1466, opened 2020-06-10
deprecation(mixer): mark Mixer configuration options as deprecated

douglas-reid:
This PR marks Istio APIs that concern Mixer as deprecated.

Mixer, the component that backs the `istio-policy` and `istio-telemetry` services, was officially deprecated with the 1.5 release of Istio. The APIs, however, were not updated to reflect that status. This PR handles that update.

The annotation of fields (and RPC methods and proto files) as deprecated will allow `istioctl` to be updated to detect and warn users about their usage of the deprecated component.

Related issues:
- https://github.com/istio/istio/issues/24471
- https://github.com/istio/istio/issues/19355

Note: the tool that runs as a result of `make gen` does not support message-level deprecation annotation, so file-level deprecation options were added where appropriate.


esnible:
I'd like `istioctl analyze` to detect as many deprecated things as possible.

It only handles resources that Galley used to know about that have an entry in _pkg/config/schema/collections/collections.gen.go_.

- The PRs deprecated some fields in _IstioMeshV1Alpha1MeshConfig_.  I could warn about these, but does anyone still have a MeshConfig in their cluster?  I do not.
- Two collections mention the mixer, `IstioMixerV1ConfigClientQuotaspecbindings` and `IstioMixerV1ConfigClientQuotaspecs`.  Are these mixer configs that are going away or do they just have old names?
- There are many telemetry examples in _samples/bookinfo/telemetry_.  Stuff in config.istio.io/v1alpha2. 
 These are staying?
- There are many policy examples in _samples/bookinfo/policy_.  I see these are deprecated.  I will start on a PR for them.

PR #1469, opened 2020-06-12
add name to gateway servers

ramaraochavali:
Currently when a gateway server is defined on a port, we create a Http Connection Manager (for Http) and TCP Proxy (for TCP). We set listener name as stat prefix as (outbound_0.0.0.0_port). However when we define multiple servers on the same port (For example when we want to have different certs), we create two http connection managers for each server and set the same stat prefix(outbound_0.0.0.0_port). So it is difficult to distinguish the stats between these servers (http connection managers). 

This PR adds an api field to Server that allows user to specify a name(unique across all servers if set) which would be set to the stat prefix http connectoin mananger/tcp proxy created for that server.


ramaraochavali:
> Lets not do stat prefix as the scope of usage is niche. Instead, how about we add a name for the server, stating that the name must be unique across all servers if set. We can then use this name for variety of purposes including stats prefix, route names, filterchain names, etc.

I was exactly thinking the same today and want to change this PR. Fully agree. I will change it like that.

ramaraochavali:

networking/v1alpha3/gateway.proto
```diff
@@ -576,6 +576,11 @@ message Server {
   // NOT IMPLEMENTED.
   // $hide_from_docs
   string default_endpoint = 5;
+
+  // An optional name of the server, when set must be unique across all servers.
+  // This will be used for variety of purposes like prefixing stats generated with
```
louiscryan, line 581: This is a bit unclear. What other 'purposes' beyond stats would this be used for?

Typically we reserve 'name' for one API resource to refer to another. 

@rshriram 

If this only exists for situations where we allow duplicate ports because they have different certs then why don't we 'tag' the cert and not the server ?

rshriram: for posterity..

We already disambiguate certs across servers using port names, where we require all port names across servers in a Gateway to be unique. It feels a bit counter-intuitive to use port names across different objects. 

There are a few use cases of named servers other than stats: trace annotations, metrics, and in future, using the server names instead of port names to disambiguate across servers across gateways.

PR #1473, opened 2020-06-18
Update gateway selector to be consistent with other names

shamsher31:
To fix https://github.com/istio/istio.io/issues/7026

Keeping it consistent with  [ingressgateway](https://github.com/istio/istio/blob/master/manifests/charts/gateways/istio-ingress/values.yaml#L9) name.


shamsher31:
@smawson @rshriram  PTAL?

PR #1475, opened 2020-06-18
Introduce HoldApplicationUntilProxyStarts MeshConfig option

luksa:


mesh/v1alpha1/proxy.proto
```diff
@@ -389,6 +389,9 @@ message ProxyConfig {
   // sleeps for the termination_drain_duration and then kills any remaining active Envoy processes.
   // If not set, a default of 5s will be applied.
   google.protobuf.Duration  termination_drain_duration = 29;
+
+  // Ensures that the Proxy container becomes ready before the other containers are started
+  bool hold_application_until_proxy_starts = 30;
```
howardjohn, line 394: I think we probably only need one or the other (annotation or this field) because you can override any setting here as an annotation already (proxy.istio.io/config).

If we do keep it here we may want this to be a BoolValue as well so we can change the default later

luksa: Ah, ok. I'm not familiar with the ProxyConfig stuff at all. I'll remove the annotation then.

luksa: Hm, how does one obtain the correct proxyConfig in `webhook.go`/`inject.go` then? `MeshConfig.DefaultConfig` seems to be the global ProxyConfig, not one that is specific to the pod. Any existing examples?

luksa: Ah, I've found it. But it gets applied too far down, so I need to either move it up or pass the information within the `SidecarInjectionSpec`, as I did originally. I need to finish work right now, but I'll see if I can put in some work in the evening.

luksa: Thanks for the pointers!

nrjpoddar:
We should make a struct now like `LifeCycleHooks’ and have post start as field under it. Thoughts?

rshriram:
Why do we even need this as an option? Under what conditions does a user want the app container to start before the sidecar and do any action ? I feel like this should be a feature flag that should be turned on and then graduated, rather than a mesh toggle. Same goes for all the options around lifecycle events unless there is a very legitimate use case for not enabling these.

dgn:
> Why do we even need this as an option? Under what conditions does a user want the app container to start before the sidecar and do any action ? I feel like this should be a feature flag that should be turned on and then graduated, rather than a mesh toggle. Same goes for all the options around lifecycle events unless there is a very legitimate use case for not enabling these.

It's new functionality and I think we should just stick to the usual disabled-by-default, enabled-by-default, graduated progression to get some testing exposure. Putting this in ProxyConfig will allow users to test this on specific deployments, which is great I think

mandarjog:
Most users (perhaps all) encounter this as a bug, because they believe this to be true. No one has the expectation that the application starts before the sidecar.
I totally agree that it should be a feature flag which we can get rid of when everything works well and is fast enough.

This does lengthen the time to get ready which is a concern for some.

dgn:
> Most users (perhaps all) encounter this as a bug, because they believe this to be true. No one has the expectation that the application starts before the sidecar.

Not sure why people would assume expect that given that it's not standard behaviour for K8s sidecars

> I totally agree that it should be a feature flag which we can get rid of when everything works well and is fast enough.

If that's the consensus then I'll go with it

luksa:
OK, then I need someone to explain what is the correct way to add a feature flag that's used by both the sidecar injector webhook and `istioctl kube-inject`. I don't want to implement this again and then have to re-implement it yet again. It seemed like you all agreed to have this configurable in the proxy config.

tariq1890:
Any updates on this? We would really like this feature

tariq1890:
I would like to chime in here since I am a user blindsided by this. I would normally just expect the main containers to not run or be ready when the Istio proxy is starting up. So I agree with @costinm, @mandarjog and @rshriram here, but I don't see the need for this field. 

The code in the PR https://github.com/istio/istio/pull/24737 is honestly safe enough to not necessitate a feature gate.

I would vote for closing this API PR and progressing that Istio/Istio PR

costinm:
It seems most people agree with having this feature - but not as an API unless there is a use case for turning it off.

We may need a feature flag - to allow testing and emergency change if something goes wrong - but usually those go into env variables and mesh config, and go away when things are stable. 

I believe there is still work in K8S to have first-class support for container depenencies - not sure the status.

luksa:
Closing this, as no API change is necessary. See istio/istio#24737.

PR #1477, opened 2020-06-22
Add targetPort to ServiceEntry Port

rshriram:
We need targetPort to come to parity with K8S service objects, especially
when a ServiceEntry targets both pods and VMs, or when a K8S service (with targetPort)
targets pods and VMs. The lack of a fixed targetPort has caused a slew of inconsistencies in implementation 
of the VM support. In addition, our existing mechanism for a port map per endpoint seems ripe for deprecation
as people have not really been using this feature for per endpoint port customization. A single target Port
seems to fit the common needs that people have.


costinm:
It's a bit more complicated - K8S allows the endpoints to use different
ports ( matching on port name ). I think we handle this correctly now.

How about adding the port only to WorkloadEntry - each workload entry can
use a different port if needed.
The targetPort is not really a concern for Gateway or Sidecar - it's a
discovery/EDS-specific feature.

networking/v1beta1/gateway.proto
```diff
@@ -598,6 +598,10 @@ message Port {
 
   // Label assigned to the port.
   string name = 3 [(google.api.field_behavior) = REQUIRED];
+
+  // The port on the endpoint where the traffic will be
+  // received. Applicable only when used with ServiceEntries.
+  uint32 target_port = 4;
```
smawson, line 604: K8s uses a string here so it can be a named port, which then points at a named port on the pod or WorkloadEntry. This allows actual ports to vary as long as they are given a consistent name.

networking/v1alpha3/workload_entry.proto
```diff
@@ -263,10 +264,7 @@ message WorkloadEntry {
   // unix:///absolute/path/to/socket for Unix domain socket endpoints.
   string address = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // Set of ports associated with the endpoint. The ports must be
-  // associated with a port name that was declared as part of the
-  // service. Do not use for `unix://` addresses.
-  map<string, uint32> ports = 2;
+  map<string, uint32> ports = 2 [deprecated = true];
```
smawson, line 267: Why was ports deprecated as part of this?

This allows us to name ports (which can be used for name-targeted port mapping). Do we use pod container ports for anything right now? These would be used for similar purpose.

costinm:
Well, if we want to have parity - then Sven is right, we need to support
named ports.

But I think WorkloadEntry having a targetPort is equivalent. Keep in mind
in K8S the named port is based
on the Container definition - where the port name is mapped to actual port
number.

For VMs - we can skip some complexity by using the service port number -
which is now used pretty consistently,
and map service port number to endpoint port, without mixing in the port
name.

smawson:
I was thinking of cases where you have multiple ports, how would we know which service ports map to which target ports? WorkloadEntry shouldn't need to know anything about the services that are targeting it, it should just describe what ports are available (potentially giving them names), and then the service defines the mapping.

If we put these ports on WorkloadEntry now we've re-coupled WorkloadEntry and ServiceEntry, when a big part of the design here was to decouple them.

costinm:
You are right. WorkloadEntry already has a map<string,uint32> ( not sure
why 32 and not 16, but too late to change) - and doc says 'port
must be associated with the port name as part of service'.

So it is indeed based on the port name already - and ServiceEntry has the
port name.

Confused - what is the purpose of this again ? Target port is already in
WorkloadEntry.

Last I checked we did have a problem with 'whitebox' mode - where we can
only use an egress port that matches
the service port.

costinm:
Re.  whitebox mode issue - if 2 services are on the same port, we can't currently support TCP egress for both since the 'port' in egress section needs to match the Service port.  So we would need to specify a 'localPort' where we would bind locally.

networking/v1alpha3/workload_entry.proto
rshriram: I deprecated because I hadn't intended to use the named target port thing [how many people actually use it?]. Because if we use named target port, we will need to convert the targetPort into a oneOf with int or string (unless we decide to use just a string there and let people specify a numeric value or port name into the string value).

rshriram:
Service entry selects pods and VMs. VMs - the target port is identified per endpoint using the portMap that has a mapping from the service port name to workload entry port. What about pods? they have a bunch of named container ports (optional). The app on the pod may or may not be listening to the same port that is being used by the service entry (as the pods were initially being selected by a k8s service with a target port). The names on container ports may not match with the names of the service entry ports either. So how do you map a service port on Service entry to the pod? Today, we simply use the service port for the pod with no attention paid to the pod's container ports (as there is no guarantee of naming consistency or the user intent to keep the port names mapped properly across pods and VMs).

OTOH, an integer targetPort is a universally well understood concept. An app being migrated from VM to k8s is likely to retain the same port on the pod. So, being able to specify something like `port: 80, targetPort: 8080` makes it very simple for people to fork and lift their workload from vm to k8s. No need to muck around with port names, or ensuring they are consistent across two different platforms. It eliminates the common problem of the user forgeting to add specific container ports with same name as the service entry ports.

networking/v1alpha3/workload_entry.proto
smawson: I would leave this undeprecated, and comment it similarly to k8s container ports:

> ports : containerPortList of ports to expose from the container. Exposing a port here gives the system additional information about the network connections a container uses, but is primarily informational. Not specifying a port here DOES NOT prevent that port from being exposed. Any port which is listening on the default "0.0.0.0" address inside a container will be accessible from the network. Cannot be updated.

This was meant to be similar, and just informational and not required.

But I could also see us dropping it if we don't think it is worth it.

smawson:
The argument for using named ports is consistency with K8s. Would be good to understand if that is a commonly used feature or not.  I agree that most cases should be just port + targetPort on ServiceEntry.

I think we do need to keep the map on WorkloadEntry if we want to support using Service to select WorkloadEntry, and support named targetPorts there. Otherwise we'd have to explicitly say that we do not support named targetPort and targetPort must only be a number.

So my preference is keeping the port map in WorkloadEntry and making targetPort a string, and just matching k8s semantics. Less surprises for our users even if most people aren't using that feature.

costinm:
On Tue, Jun 23, 2020 at 11:34 AM Shriram Rajagopalan <

> *@rshriram* commented on this pull request.
>
> In networking/v1alpha3/workload_entry.proto
> <https://github.com/istio/api/pull/1477#discussion_r444427042>:
>
> > @@ -263,10 +264,7 @@ message WorkloadEntry {
>    // unix:///absolute/path/to/socket for Unix domain socket endpoints.
>    string address = 1 [(google.api.field_behavior) = REQUIRED];
>
> -  // Set of ports associated with the endpoint. The ports must be
> -  // associated with a port name that was declared as part of the
> -  // service. Do not use for `unix://` addresses.
> -  map<string, uint32> ports = 2;
> +  map<string, uint32> ports = 2 [deprecated = true];
>
> I deprecated because I hadn't intended to use the named target port thing
> [how many people actually use it?]. Because if we use named target port, we
> will need to convert the targetPort into a oneOf with int or string (unless
> we decide to use just a string there and let people specify a numeric value
> or port name into the string value).
>

The main issue is that at least for VMs, it is not guaranteed that all VMs
will use the same port.
If we do support the k8s named ports on k8s - we should be consistent.  I
know it has been a pain, and I wouldn't mind if we used 'servicePortNumber'
as
key - just as we use in cluster names.

But it is still worth allowing the endpoint target port to not be fixed -
and keep this logic/API in WorkloadEntry and EDS only.

costinm:
On Tue, Jun 23, 2020 at 11:44 AM Shriram Rajagopalan <

> Service entry selects pods and VMs. VMs - the target port is identified
> per endpoint using the portMap that has a mapping from the service port
> name to workload entry port. What about pods? they have a bunch of named
> container ports (optional). The app on the pod may or may not be listening
> to the same port that is being used by the service entry (as the pods were
> initially being selected by a k8s service with a target port). The names on
> container ports may not match with the names of the service entry ports
> either. So how do you map a service port on Service entry to the pod?
> Today, we simply use the service port for the pod with no attention paid to
> the pod's container ports (as there is no guarantee of naming consistency
> or the user intent to keep the port names mapped properly across pods and
> VMs).
>

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#endpointport-v1-core

In K8S we load Endpoints ( and the new EndpointSlice) - which has sets of
endpoints. The target port number doesn't have to be the same if named
ports are used.
An example use case is starting a new version of deployment that uses a
different target port, if port name is used.

AFAIK enpoints.go is (or was) doing the mapping, using port by name. Of
course the user should use the names consistently - just like they would
have to do in their pods. That's the point - and it's much easier on a VM
to use the port name, which is in config - than using the same targetPort.
Sometimes the VMs don't run as root, and multiple apps run on the same VM -
which may take over the port.

I would say the use case of not requiring a fixed target port is far more
important for VMs.



> OTOH, an integer targetPort is a universally well understood concept. An
> app being migrated from VM to k8s is likely to retain the same port on the
> pod. So, being able to specify something like port: 80, targetPort: 8080
> makes it very simple for people to fork and lift their workload from vm to
> k8s. No need to muck around with port names, or ensuring they are
> consistent across two different platforms. It eliminates the common problem
> of the user forgeting to add specific container ports with same name as the
> service entry ports.
>

I don't mind having a map of 'servicePort' to 'targetPort' - in
WorkloadEntry, where VMs can still use different targetPorts.

My concern is with having the targetPort in ServiceEntry/Gateway/etc.

smawson:
@costinm I don't think we want targetPort on WorkloadEntry. That couples the workload to the service. The workload should just list the ports that it exposes, it is up to the service to map to those ports. So having this on ServiceEntry is the right place.

rshriram:
alright. I removed the deprecation piece and changed the targetPort to a string.

networking/v1beta1/workload_entry.proto
```diff
@@ -198,6 +197,7 @@ import "google/api/field_behavior.proto";
 //   - number: 80
 //     name: http
 //     protocol: HTTP
+//     targetPort: 8080
```
howardjohn, line 200: This actually would need to be `     targetPort: "8080"` now right? which is ambiguous and not very ergonomic

costinm:
On Tue, Jun 23, 2020, 18:20 Sven Mawson <notifications@github.com> wrote:

> @costinm <https://github.com/costinm> I don't think we want targetPort on
> WorkloadEntry. That couples the workload to the service. The workload
> should just list the ports that it exposes, it is up to the service to map
> to those ports. So having this on ServiceEntry is the right place.
>

Is a Pod coupled ? This mirrors containerPort in pods, which include the
real targetPort where the app is listening and the named port where it
should listen.

You can't guarantee VMs will all get same port free.

costinm:
On Wed, Jun 24, 2020, 07:45 John Howard <notifications@github.com> wrote:

> *@howardjohn* commented on this pull request.
>
> In networking/v1beta1/workload_entry.proto
> <https://github.com/istio/api/pull/1477#discussion_r444948627>:
>
> > @@ -198,6 +197,7 @@ import "google/api/field_behavior.proto";
>  //   - number: 80
>  //     name: http
>  //     protocol: HTTP
> +//     targetPort: 8080
>
> This actually would need to be targetPort: "8080" now right? which is
> ambiguous and not very ergonomic
>

And we already have name, if we want a string matching k8s.

networking/v1beta1/workload_entry.proto
rshriram: oh yeah.. Oh well, I might as well make it a one of then.. I really hate the one-of for port selectors!

smawson:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#containerport-v1-core

Containers just have a port, no target port. The port can have a name and that name can be referred to by Services.

smawson:
Can we not just use a string? I thought yaml allows unquoted strings so you can still say targetPort: 8080.

Unless there is a good reason to be different than k8s I'd rather just match the API.

howardjohn:
unmarshaling `targetPort: 8080` into a string will fail, k8s gets around it by having an IntOrString type with custom marshaling logic (which we can probably also do)

rshriram:
I have updated the target port to one of int or string. PTAL

rshriram:
ping @smawson / @howardjohn

networking/v1alpha3/service_entry.proto
```diff
@@ -583,17 +582,13 @@ import "networking/v1alpha3/workload_entry.proto";
...
 //   resolution: DNS
 //   endpoints:
 //   - address: us.foo.bar.com
-//     ports:
-//       https: 8080
 //   - address: uk.foo.bar.com
-//     ports:
-//       https: 9080
 //   - address: in.foo.bar.com
-//     ports:
```
howardjohn, line 595: If this field is still going to exist we probably need to document why you should use it vs target port? Or recommend not using it

rshriram: I have documented it (see https://github.com/istio/api/pull/1477/files#diff-caf927451392d5943ac31afd98eb1012R271)

networking/v1alpha3/gateway.proto
```diff
@@ -596,6 +596,24 @@ message Port {
 
   // Label assigned to the port.
   string name = 3 [(google.api.field_behavior) = REQUIRED];
+
+  // The port or port name on the endpoint where the traffic will be
+  // received. Applicable only when used with ServiceEntries.
+  message TargetPort {
```
costinm, line 602: I don't get it. The port has a name already ( field 3) - 'label assigned to the port'. 

When would the 'targetPort' name (string) be different than the port name - and why ? 
In k8s - that's how they handle this, why would Istio users have to define 2 names for a port ?

networking/v1alpha3/service_entry.proto
```diff
@@ -555,17 +558,13 @@ import "networking/v1alpha3/workload_entry.proto";
 //   - number: 80
 //     name: http
 //     protocol: HTTP
+//     targetPort:
```
costinm, line 561: So original: using the port name, and allow multiple target ports.
Replacement: fixed target port ? 

And if user wants the first behavior:

- number: 80
  name: http
  targetPort:
    name: http 

How much redundant typing and complexity do we need ?

networking/v1alpha3/workload_entry.gen.json
```diff
@@ -19,7 +19,7 @@
             }
           },
           "ports": {
-            "description": "Set of ports associated with the endpoint. The ports must be associated with a port name that was declared as part of the service. Do not use for `unix://` addresses.",
+            "description": "Set of ports associated with the endpoint. If targetPort is not specified, the ports must be associated with a port name that was declared as part of the service. Do not use for `unix://` addresses. If targetPort is specified as a port name, the numeric target port is derived from the associated named port.",
             "type": "object",
```
costinm, line 23: can you explain here the use case for having a different name for the port and targetPort ? Or at least specify that 
"if targetPort is not present, the name of the port can still be used" ?

networking/v1alpha3/workload_entry.proto
```diff
@@ -103,6 +102,8 @@ import "google/api/field_behavior.proto";
 //   - number: 80
 //     name: http
 //     protocol: HTTP
+//     targetPort:
+//       number: 8080
```
costinm, line 106: Please add at least an example of using targetPort.name - and explain that it allows workloads on VMs to use DIFFERENT port numbers, for example for running multiple replicas of a workload on the same VM.

costinm:
How about:
- add just 'defaultTargetPort' - numeric
- comment will be "If specified, endpoints will run on the defaultTargetPort instead of the service port. Endpoints can use 
non-default ports by using the port name"

This eliminates the redundant/duplicated name - there is only one port name, allow a simpler numeric value, it's clear that
it's just a default but endpoints can also use custom ports if needed. I suspect the most common use case will be port < 1024
for running as non-root.

networking/v1alpha3/gateway.proto
rshriram: this structure is exact mirror of k8s. service port has a name, it has a target port field that can also be a name or string (and this target port name has no relation to the service port name). https://github.com/kubernetes/kubernetes/blob/faff3c5378c801eefa8f6bfcc4e988924979b5fb/staging/src/k8s.io/api/core/v1/generated.proto#L4633

rshriram:
there is so much confusion here. The spec here is a replica of k8s service sport spec https://github.com/kubernetes/kubernetes/blob/faff3c5378c801eefa8f6bfcc4e988924979b5fb/staging/src/k8s.io/api/core/v1/generated.proto#L4633 . Please read that spec and point out differences

networking/v1alpha3/gateway.proto
costinm: Port is not an exact mirror of K8S - and it is not clear why adding a second name us needed.

costinm:
How about an RFC - as it should be required before approving any change to a beta API, not only if there is apparent 
confusion ? 

I thought we now have processes in places.

I know how K8S service port works - including the problems we had with it. We can satisfy the use cases of endpoint port selected by name using the service name. What we can't do is provide both a simpler 'default numeric port' ( like 8080 or 8443), which is a very common use case to avoid running as root - as well as allow VMs to use a non-uniform, custom port based on name.

Istio is trying to improve some of the K8S APIs - this is not a case where a direct copy is best option.

networking/v1alpha3/gateway.proto
rshriram: It is absolutely an exact mirror. We have numeric port, port name, (we dont have nodePort but happy to add that), and then we have the targetPort. K8S has its custom proto IntOrString with custom marshalling to avoid using one-of in their structs. For us, its a standard oneof in our protos. Where is the discrepancy with k8s port here?

rshriram:
>I know how K8S service port works - including the problems we had with it. We can satisfy the use cases of endpoint port selected by name using the service name. What we can't do is provide both a simpler 'default numeric port' ( like 8080 or 8443), which is a very common use case to avoid running as root - as well as allow VMs to use a non-uniform, custom port based on name. Istio is trying to improve some of the K8S APIs - this is not a case where a direct copy is best option.

My first iteration of this PR had a simple numeric `targetPort` as part of the port message. It did not have this one-of. See https://github.com/istio/api/pull/1477/commits/4cf292d41579a38db2ac6a4712a8d4a35adf4333 . @smawson 's suggestion was to use the one-of so as to keep it in sync with k8s. I would "prefer" the numeric targetPort as well. I need some consensus here.

networking/v1alpha3/service_entry.proto
smawson: The service port name should be optional (only required if there are multiple ports), so it would just be:

- number: 80
  targetPort: http

if we had custom marshalling like k8s does.

I think the k8s behavior for a named port is to not try to match that by name to a container port, are we inconsistent there?

networking/v1alpha3/workload_entry.proto
```diff
@@ -263,9 +264,11 @@ message WorkloadEntry {
   // unix:///absolute/path/to/socket for Unix domain socket endpoints.
   string address = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // Set of ports associated with the endpoint. The ports must be
-  // associated with a port name that was declared as part of the
-  // service. Do not use for `unix://` addresses.
+  // Set of ports associated with the endpoint. If targetPort is not
+  // specified, the ports must be associated with a port name that was
+  // declared as part of the service. Do not use for `unix://`
+  // addresses. If targetPort is specified as a port name, the numeric
```
smawson, line 270: Remove the part here about targetPort as a port name?

networking/v1beta1/workload_entry.proto
```diff
@@ -262,9 +263,11 @@ message WorkloadEntry {
   // unix:///absolute/path/to/socket for Unix domain socket endpoints.
   string address = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // Set of ports associated with the endpoint. The ports must be
-  // associated with a port name that was declared as part of the
-  // service. Do not use for `unix://` addresses.
+  // Set of ports associated with the endpoint. If targetPort is not
+  // specified, the ports must be associated with a port name that was
+  // declared as part of the service. Do not use for `unix://`
+  // addresses. If targetPort is specified as a port name, the numeric
```
smawson, line 269: Same here.

I think on ServiceEntry port we should define how ports get mapped:
- Default is to map 1:1
- If targetPort is specified, we map to that.
- If port name is specified and it matches a port on the WorkloadEntry, we map to that?

So here we can just keep this simple but we should mention the name mapping.

networking/v1alpha3/workload_entry.proto
rshriram: oh left over docs.

networking/v1beta1/workload_entry.proto
rshriram: So the model is this:
1. you can specify targetPort - which takes precedence
2. if you dont specify targetPort, it is back to the old behavior, where the old (well known) behavior for service entry is
  1. if you dont specify any endpoint ports, then we assume 1-1 mapping
  2. if you specify endpoint port map (servicePortName: portNumber), then we will use that.

rshriram:
I have updated the documentation describing the fallback options (from targetPort, to endpoint's portmap to the 1-1 mapping if both are absent). PTAL

networking/v1alpha3/workload_entry.proto
```diff
@@ -37,12 +37,13 @@ import "google/api/field_behavior.proto";
...
-// The following example declares a workload entry representing a
-// VM for the `details.bookinfo.com` service. This VM has
-// sidecar installed and bootstrapped using the `details-legacy`
-// service account. The sidecar receives HTTP traffic on port 80
-// (wrapped in istio mutual TLS) and forwards it to the application on
-// the localhost on the same port.
+// The following example declares a workload entry representing a VM
+// for the `details.bookinfo.com` service. This VM has sidecar
+// installed and bootstrapped using the `details-legacy` service
+// account. The service is exposed by the sidecar on port 80 to
```
costinm, line 43: The service is exposed on port 80 ( without 'by the sidecar' - if you want you can say 'by the egress/client-side sidecar' ). 

It is confusing to talk about 'sidecar' without context - there are 2 of them.

networking/v1alpha3/workload_entry.proto
```diff
@@ -263,9 +264,22 @@ message WorkloadEntry {
   // unix:///absolute/path/to/socket for Unix domain socket endpoints.
   string address = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // Set of ports associated with the endpoint. The ports must be
-  // associated with a port name that was declared as part of the
-  // service. Do not use for `unix://` addresses.
+  // Set of ports associated with the endpoint. If omitted, and the
+  // targetPort is not specified as part of the service's port
+  // specification, traffic to a service port will be forwarded to one
+  // of the endpoints on the same port. If the targetPort is specified
```
costinm, line 270: This should be last. 

I.e: 
- if ports is included, it will take priority, using the Port.name as a key 
- if ports is not included, but targetPort is specified in SericeEntry - targetPort is used
- if neither ports nor targetPort are specified - the Port.number is used.

This allows a VM to use a random port, if the targetPort is in used by something else.

rshriram: The explanation here is correct as it deals with either targetPort or portMap. The note below has been fixed to show the precedence

rshriram:
I have fixed all the comments. PTAL

networking/v1alpha3/workload_entry.gen.json
```diff
@@ -19,7 +19,7 @@
             }
           },
           "ports": {
-            "description": "Set of ports associated with the endpoint. The ports must be associated with a port name that was declared as part of the service. Do not use for `unix://` addresses.",
+            "description": "Set of ports associated with the endpoint. If omitted, and the targetPort is not specified as part of the service's port specification, traffic to a service port will be forwarded to one of the endpoints on the same port. If the targetPort is specified as part of the service's port specification, traffic to the service port will be forwarded to one of the endpoints on the specified `targetPort`. If the port map is specified (and targetPort is omitted), it must be a map of servicePortName to this endpoint's port, such that traffic to the service port will be forwarded to the endpoint port that maps to the service's portName.",
```
costinm, line 22: If port map is specified - the target port doesn't matter ( i.e. remove 'and targetPort is omitted' ).

networking/v1alpha3/workload_entry.proto
costinm: Maybe I'm too tired to parse this: but it's still quite confusing. 

For example: 'if the targetPort is specified ..., traffic will be forwarded to one of the endpoints on the specified targetPort' - but only if the ports map is missing ! The text seem to suggest that if targetPort is specified - it will be used even if ports is present. 

But the NOTE 2 is correct, so I'll let native English speakers to parse the text above.

networking/v1alpha3/workload_entry.gen.json
rshriram: updated

networking/v1alpha3/workload_entry.proto
```diff
@@ -263,9 +264,20 @@ message WorkloadEntry {
...
+  // omitted, and the targetPort is specified as part of the service's
+  // port specification, traffic to the service port will be forwarded
+  // to one of the endpoints on the specified `targetPort`. If both
+  // the targetPort and endpoint's port map are not specified, traffic
+  // to a service port will be forwarded to one of the endpoints on
+  // the same port.
+  //
+  // **NOTE 1:** Do not use for `unix://` addresses.
+  //
+  // **NOTE 2:** endpoint port map takes precedence over targetPort.
```
smawson, line 280: I'm fine with this for now but I have a concern about coupling WorkloadEntry to service here. A big part of why WorkloadEntry exists is to decouple the workload from the service, so that services can be created and removed without having to make changes to the WorkloadEntry.

I think this is just a wording issue though, since the same coupling exists with k8s services and container ports. It is just the directionality that is different. There the service needs to be aware of the workload ports, and point at them (using named target port).

So maybe we can work on rewording this and ServiceEntry to indicate that the ServiceEntry port name is pointing at a WorkloadEntry or ContainerPort name, rather than the WorkloadEntry port name pointing at the ServiceEntry port name.

But those are just comment changes, it doesn't change the actual spec, so let's get this in and we can clean up comments after :)

PR #1483, opened 2020-06-24
Move operator to gogo proto

howardjohn:
This is probably the wrong direction long term. However, short term, we cannot update to the latest go-control-plane version without updating our protobuf version, and we cannot update our protobuf version because the operator is using a mix of golang proto and gogo proto, which is broken by the new protobuf update and will never be fixed because gogo proto is an abandoned project.

Long term we *probably* want to switch everything to golang proto - I will work on a design doc for this later, once mixer is removed. But for now, we will move it to fully gogo (aligned with all other istio APIs), then we can move all of them to golang/protobuf in one go if we choose to do so


howardjohn:
WIP until https://github.com/istio/istio/pull/24956 is green

PR #1488, opened 2020-06-29
add uid/gid and target port

TaichiHo:
Add a couple annotations for an upcoming istio/cni change.
This change will enable us to control cni behavior over kubernetes annotation.
* targetPort
* proxyUID
* proxyGID

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -267,6 +267,26 @@ annotations:
     hidden: false
     resources:
       - Pod
+  - name: traffic.sidecar.istio.io/targetPort
+    description: The port where envoy listens for all TCP traffic
+    deprecated: false
+    hidden: false
+    resources:
+      - Pod
+  - name: traffic.sidecar.istio.io/noRedirectUID
+    description: the UID of the user for which the redirection is not
+      applied. Typically, this is the UID of the proxy container
+    deprecated: false
+    hidden: false
+    resources:
+      - Pod
+  - name: traffic.sidecar.istio.io/noRedirectGID
+    description: the GID of the user for which the redirection is not applied.
+      Typically, this is the GID of the proxy container. Same default as UID
+    deprecated: false
+    hidden: false
+    resources:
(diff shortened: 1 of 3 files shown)


howardjohn:
Can you give a little more info about why we would want this? I don't see why this needs to be customizable

TaichiHo:
Thanks for the reply.

We are planning to use istio/cni without using istio. 

This change would allow us to redirect the traffic to the port our envoy is listening on. Same applies to uid and gid.

annotation/annotations.yaml
```diff
@@ -267,6 +267,26 @@ annotations:
     hidden: false
     resources:
       - Pod
+  - name: traffic.sidecar.istio.io/targetPort
+    description: The port where envoy listens for all TCP traffic
+    deprecated: false
+    hidden: false
+    resources:
+      - Pod
+  - name: traffic.sidecar.istio.io/noRedirectUID
```
costinm, line 276: How about 'sidecarGID'/'sidecarUID' ? Or 'noCaptureUID/noCaptureGID' ? 

Redirect happens to be what we use now - may change in future, and users may not understand it.

Also: do we still need both UID and GID ? I thought we found a way to use only GID ( which has the benefit of  allowing other UIDs to be excluded ).

annotation/annotations.yaml
```diff
@@ -267,6 +267,26 @@ annotations:
     hidden: false
     resources:
       - Pod
+  - name: traffic.sidecar.istio.io/targetPort
```
costinm, line 270: Except we now have 2 ports ? Generic words like 'target' should be avoided - this is not the only target in the world or on 'traffic'. 

captureInPort/captureOutPort ( or ingress/egeress ) maybe ?

TaichiHo:
Thanks @costinm 
Is there an existing RFC I can contribute to? I took a look at the team drive but didn't find one yet.

I might need to take some time to learn the history and understand the other use cases to come up with a proper RFC

PR #1489, opened 2020-06-29
Update enable_auto_mtls default to true

shamsher31:
To fix https://github.com/istio/istio.io/issues/7637

`EnableAutoMtls` [default is now set to `true`](https://github.com/istio/istio/pull/22687/files#diff-94cf4f94e1d1aac6fe3fd24fbb526bfbR157) but the comment was not updated to reflect in the docs.


costinm:
Would be great to take a quick look at all other defaults - I have a feeling this is not the only one.

PR #1493, opened 2020-06-30
Specify defaults for mesh and proxy configs

shamsher31:
Follow up PR for https://github.com/istio/api/pull/1489

Based on @costinm https://github.com/istio/api/pull/1489#issuecomment-651142386  , I have updated all other defaults as per the latest [DefaultMeshConfigs](https://github.com/istio/istio/blob/master/pkg/config/mesh/mesh.go#L80). 

Suggestions are welcome to improve default value descriptions as it will reflect on [istio.io docs](https://preliminary.istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig). 

Ref: https://github.com/istio/istio/pull/22687

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -272,11 +272,13 @@ message ProxyConfig {
 
   // The time in seconds that Envoy will drain connections during a hot
   // restart. MUST be >=1s (e.g., _1s/1m/1h_)
+  // Default drain duration is 45s.
   google.protobuf.Duration drain_duration = 4;
 
   // The time in seconds that Envoy will wait before shutting down the
   // parent process during a hot restart. MUST be >=1s (e.g., _1s/1m/1h_).
   // MUST BE greater than _drain_duration_ parameter.
+  // Default shutdown duration is 60s.
   google.protobuf.Duration parent_shutdown_duration = 5;
 
   // Address of the discovery service exposing xDS with mTLS connection.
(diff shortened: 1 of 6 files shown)


mesh/v1alpha1/proxy.proto
```diff
@@ -300,13 +302,15 @@ message ProxyConfig {
...
   // Port on which Envoy should listen for administrative commands.
+  // Default port is 15000.
   int32 proxy_admin_port = 11;
 
   // $hide_from_docs
   string availability_zone = 12 [deprecated=true];
 
   // Authentication policy defines the global switch to control authentication
   // for Envoy-to-Envoy communication for istio components Mixer and Pilot.
+  // Default is set to NONE which means do not encrypt Envoy to Envoy traffic.
```
howardjohn, line 313: This is misleading, because its about envoy -> istiod (not envoy to envoy as you mention) and is not really properly followed - we have some hack that if discoveryPort==15012 we use TLS anyways. The default is encrypted, even though its in a roundabout confusing way.

We probably want to fix the code but for now we can just leave this one maybe?

shamsher31: Agree, This was taken directly from [AuthPolicy enums](https://github.com/istio/api/blob/6e8e32f7acc2f6f603ffc4775e75568e1078dff9/mesh/v1alpha1/proxy.proto#L29). In that case enum also needs to be updated?

shamsher31: If default is encrypted, I think we also need to update [ControlPlaneAuthPolicy](https://github.com/istio/istio/blob/master/pkg/config/mesh/mesh.go#L57) to `MUTUAL_TLS` ?

PR #1499, opened 2020-07-07
add pending to operator status

richardwxn:
related: https://github.com/istio/istio/issues/23479


operator/v1alpha1/operator.proto
```diff
@@ -199,6 +199,9 @@ message InstallStatus {
         HEALTHY = 3;
         // Component is in an error state.
         ERROR = 4;
+        // Action is needed from user for reconciliation loop to proceed
+        // e.g. There are proxies still pointing to the control plane revision when try to remove IstioOperator CR.
+        PENDING = 5;
```
smawson, line 204: Also please update the comment below on the overall status field to describe when this state occurs if it is an overall state.

operator/v1alpha1/operator.proto
```diff
@@ -199,6 +199,9 @@ message InstallStatus {
         HEALTHY = 3;
         // Component is in an error state.
         ERROR = 4;
+        // Action is needed from user for reconciliation loop to proceed
```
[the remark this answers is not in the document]
richardwxn: done

operator/v1alpha1/operator.proto
richardwxn: done

operator/v1alpha1/operator.proto
```diff
@@ -185,37 +185,50 @@ message IstioOperatorSpec {
...
+    // Component is being updated to a different version.
+    UPDATING = 1;
+    // Controller has started but not yet completed reconciliation loop for the component.
+    RECONCILING = 2;
+    // Component is healthy.
+    HEALTHY = 3;
+    // Component is in an error state.
+    ERROR = 4;
+    // Overall status only and would not be set as a component status.
+    // Action is needed from user for reconciliation loop to proceed
```
smawson, line 201:
```suggestion
    // Action is needed from the user for reconciliation to proceed
```

operator/v1alpha1/operator.proto
```diff
@@ -185,37 +185,50 @@ message IstioOperatorSpec {
...
+    // e.g. There are proxies still pointing to the control plane revision when try to remove IstioOperator CR.
```
smawson, line 202:
```suggestion
    // e.g. There are proxies still pointing to the control plane revision when try to remove an IstioOperator CR.
```

operator/v1alpha1/operator.proto
```diff
@@ -185,37 +185,50 @@ message IstioOperatorSpec {
...
     // - If all components are HEALTHY, overall status is HEALTHY.
     // - If one or more components are RECONCILING and others are HEALTHY, overall status is RECONCILING.
     // - If one or more components are UPDATING and others are HEALTHY, overall status is UPDATING.
     // - If components are a mix of RECONCILING, UPDATING and HEALTHY, overall status is UPDATING.
     // - If any component is in ERROR state, overall status is ERROR.
+    // - If further action is needed for reconciliation to proceed, overall status is PENDING.
     Status status = 1;
-    // Individual status of each component controlled by the operator. The map key is the name of the component.
-    map<string, VersionStatus> component_status = 2;
+    // Optional message explains the reason and actions needed for the existing overall status.
```
smawson, line 230:
```suggestion
    // Optional message providing additional information about the overall status.
```

operator/v1alpha1/operator.proto
```diff
@@ -185,37 +185,50 @@ message IstioOperatorSpec {
...
-        ERROR = 4;
-    }
     // VersionStatus is the status and version of a component.
     message VersionStatus {
         string version = 1;
         Status status = 2;
         string error = 4;
     }
     // Overall status of all components controlled by the operator.
+    OverallStatus status = 1;
```
smawson, line 215: Is this change safe? This is not backwards compatible so want to make sure we can handle upgrades from previous version of Istio to new version. Also if anyone was relying on reading the status directly via API their code would break.

On the other hand this is alpha so maybe OK to make this change, just want to make sure we test upgrades and verify status doesn't break.

operator/v1alpha1/operator.proto
richardwxn: done

operator/v1alpha1/operator.proto
richardwxn: done

operator/v1alpha1/operator.proto
richardwxn: done

operator/v1alpha1/operator.proto
richardwxn: Yep this is not, so to make it more compatible, I think it might be better to just add an extra message field to InstallStatus and a pending to the enum status. Updated the change.

richardwxn:
@ostromart Can you also take a look?

operator/v1alpha1/operator.proto
ostromart: I think pending is too vague. Can we call it "ACTION_REQUIRED"?

richardwxn: ok done

PR #1503, opened 2020-07-09
Update auth policy enum for proxy config

shamsher31:
Based on the https://github.com/istio/api/pull/1493#discussion_r447717857, the `ControlPlaneAuthPolicy` applied to Envoy > Istiod.

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -24,7 +24,6 @@ option go_package="istio.io/api/mesh/v1alpha1";
 
 // AuthenticationPolicy defines how the proxy is authenticated when it connects to the control plane.
 // It can be set for two different scopes, mesh-wide or set on a per-pod basis using the ProxyConfig annotation.
-// and the most narrow scope with non-INHERIT value will be used.
 // Mesh policy cannot be INHERIT.
 enum AuthenticationPolicy {
   // Do not encrypt proxy to control plane traffic.
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -22,15 +22,15 @@ package istio.mesh.v1alpha1;
 
 option go_package="istio.io/api/mesh/v1alpha1";
 
-// AuthenticationPolicy defines authentication policy to be used by the proxy. It can be set for
-// different scopes (mesh, service …), and the most narrow scope with
+// AuthenticationPolicy defines how the proxy is authenticated when it connects to the control plane.
+// It can be set for different scopes (mesh, service …), and the most narrow scope with
 // non-INHERIT value will be used.
 // Mesh policy cannot be INHERIT.
 enum AuthenticationPolicy {
-  // Do not encrypt Envoy to Istiod traffic.
+  // Do not encrypt proxy to control plane traffic.
   NONE = 0;
 
-  // Envoy to Istiod traffic is wrapped into mutual TLS connections.
+  // Proxy to control plane traffic is wrapped into mutual TLS connections.
   MUTUAL_TLS = 1;
 
   // Use the policy defined by the parent scope. Should not be used for mesh
@@ -308,8 +308,7 @@ message ProxyConfig {
   // $hide_from_docs
   string availability_zone = 12 [deprecated=true];
 
-  // Authentication policy defines the global switch to control authentication
-  // for Envoy-to-Istiod communication.
+  // AuthenticationPolicy defines how the proxy is authenticated when it connects to the control plane.
   // Default is set to MUTUAL_TLS.
   AuthenticationPolicy control_plane_auth_policy = 13;
   // File path of custom proxy configuration, currently used by proxies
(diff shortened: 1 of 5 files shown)


mesh/v1alpha1/proxy.proto
```diff
@@ -22,15 +22,15 @@ package istio.mesh.v1alpha1;
 
 option go_package="istio.io/api/mesh/v1alpha1";
 
-// AuthenticationPolicy defines authentication policy. It can be set for
+// AuthenticationPolicy defines authentication policy to be used by the proxy. It can be set for
```
smawson, line 25: Can we reword this to say:

AuthenticationPolicy defines how the proxy is authenticated when it connects to the control plane.

mesh/v1alpha1/proxy.proto
```diff
@@ -22,15 +22,15 @@ package istio.mesh.v1alpha1;
...
 // different scopes (mesh, service …), and the most narrow scope with
```
smawson, line 26: I think this scope stuff is wrong? Where is it actually possible to set this? I only see it in ProxyConfig, so it can only be set at mesh level or at the level of an individual workload through an annotation? Or are there other ways to set ProxyConfig?

Sorry that's a lot of questions :)

mesh/v1alpha1/proxy.proto
```diff
@@ -22,15 +22,15 @@ package istio.mesh.v1alpha1;
...
 // non-INHERIT value will be used.
 // Mesh policy cannot be INHERIT.
 enum AuthenticationPolicy {
-  // Do not encrypt Envoy to Envoy traffic.
+  // Do not encrypt Envoy to Istiod traffic.
```
smawson, line 30: Let's generalize to proxy to control plane traffic and not explicitly say envoy or istiod.

mesh/v1alpha1/proxy.proto
```diff
@@ -309,7 +309,7 @@ message ProxyConfig {
   string availability_zone = 12 [deprecated=true];
 
   // Authentication policy defines the global switch to control authentication
```
smawson, line 311: Since this is in proxy config it isn't only global, we can reword this too.

mesh/v1alpha1/proxy.proto
shamsher31: Sound good. Done.

mesh/v1alpha1/proxy.proto
shamsher31: Perfect. Thanks.

mesh/v1alpha1/proxy.proto
shamsher31: @smawson  As far as I know, it's only been used in ProxyConfig. In that case, I think the scope related sentence can be completely removed?

https://github.com/istio/api/blob/22f2acfe5ab4008a4d14009b3975752ed8dffc1b/mesh/v1alpha1/proxy.proto#L26-L28

smawson: AFAICT this can only be mesh-wide or set on a per-pod basis using the ProxyConfig annotation, so that would be two scopes: mesh or pod.

I'm OK removing the scope discussion or fixing it to be accurate.

shamsher31: Done, Thanks.

mesh/v1alpha1/proxy.proto
```diff
@@ -22,15 +22,15 @@ package istio.mesh.v1alpha1;
 
 option go_package="istio.io/api/mesh/v1alpha1";
 
-// AuthenticationPolicy defines authentication policy. It can be set for
-// different scopes (mesh, service …), and the most narrow scope with
-// non-INHERIT value will be used.
+// AuthenticationPolicy defines how the proxy is authenticated when it connects to the control plane.
+// It can be set for two different scopes, mesh-wide or set on a per-pod basis using the ProxyConfig annotation.
+// and the most narrow scope with non-INHERIT value will be used.
```
smawson, line 27:
```suggestion
// The most narrow scope with non-INHERIT value will be used.
```

Or just delete this last line.

mesh/v1alpha1/proxy.proto
```diff
@@ -22,15 +22,15 @@ package istio.mesh.v1alpha1;
 
 option go_package="istio.io/api/mesh/v1alpha1";
 
-// AuthenticationPolicy defines authentication policy. It can be set for
-// different scopes (mesh, service …), and the most narrow scope with
-// non-INHERIT value will be used.
+// AuthenticationPolicy defines how the proxy is authenticated when it connects to the control plane.
+// It can be set for two different scopes, mesh-wide or set on a per-pod basis using the ProxyConfig annotation.
+// The most narrow scope with non-INHERIT value will be used.
```
costinm, line 27: I'm not sure what inherit does - the ProxyConfig is either the default from MeshConfig, added via injection or the one user specifies in the injection. Nothing gets merged AFAIK. 

It may be worth adding few words about the default behavior: Istiod will automatically used MTLS if the destination port is 15012 or 443 ( I think I got both right ).

shamsher31: > I'm not sure what inherit does

Agree. I haven't seen the use of `INHERIT` anywhere in repo. Only `NONE` and `MUTUAL_TLS` is widely used across test cases and examples.

PR #1504, opened 2020-07-09
add istiod-central component

irisdingbj:


linsun:
I think @irisdingbj is referring to the central component we discussed yesterday:  https://docs.google.com/document/d/1EymnzMqa7hdmKn8sEoaf1elajyp4gXEDSUavxzXB3jw/edit#heading=h.3qdwucgp1qe8

howardjohn:
I am a bit confused since that link the only real motivation I saw is that
I said it was a good idea - I don't remember saying that, we may have
gotten confused? I am not sure why we need an entirely new component just
to disable the webhook.

This feels a bit like creating an entirely new copy of the Istiod component
to turn off telemetry, which doesn't seem like a good idea to me

howardjohn:
I mean, we already have option to turn off the crd? I don't really get the
benefits

linsun:
Are you referring to the enableCRDTemplates flag?  I think that is only for helm stuff when I tried few weeks ago.

Also, what about validation webhook config?

howardjohn:
We need to make the webhook configurable anyways, there are users who don't
want to use validating webhook and currently we cannot turn it off

linsun:
yes it is pretty much a new component to allow us to build what we need out of base and discovery.  @irisdingbj said you may have some objections to modify existing codebase to allow these customizations to base and discovery in the past which is why she started with a new component to have less impact to existing components.

linsun:
@howardjohn can I just config validation webhook config on the remote cluster?  I don't understand why it is needed on the istiod control cluster other than we hardcoded it to the base today.

howardjohn:
Yes, we can add an option to skip the validating webhook. we used to have
one

irisdingbj:
@howardjohn @linsun Thanks for the comments. I will close this PR as we decided to use flags to control this other than introducing a new component.

PR #1508, opened 2020-07-10
Add Comment to CredentialName explaining use with Gateway Proxy only

nschhina:
Updates the comment for CredentialName in DestinationRule API to clearly specify use case is for originating traffic from gateway proxy type only.

Ref: https://github.com/istio/istio/issues/14039

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -2872,7 +2872,6 @@ spec:
     listKind: EnvoyFilterList
     plural: envoyfilters
     singular: envoyfilter
-  preserveUnknownFields: true
   scope: Namespaced
   versions:
   - name: v1alpha3
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -2872,6 +2872,7 @@ spec:
     listKind: EnvoyFilterList
     plural: envoyfilters
     singular: envoyfilter
+  preserveUnknownFields: true
   scope: Namespaced
   versions:
   - name: v1alpha3
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -929,9 +929,9 @@ message ClientTLSSettings {
   // ca.crt key for CA certificates is also supported.
   // Only one of client certificates and CA certificate
   // or credentialName can be specified.
-  // `This field is currently in alpha and is partially implemented.`
-  // Istio 1.7 implementation only supports it for Kubernetes
-  // secrets, used in Gateway.
+  //
+  // **NOTE:** This field is currently applicable only at gateways.
+  // Sidecars will continue to use the certificate paths.
   string credential_name = 7;
 
   // A list of alternate names to verify the subject identity in the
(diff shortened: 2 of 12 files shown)


kubernetes/customresourcedefinitions.gen.yaml
```diff
@@ -2872,7 +2872,6 @@ spec:
     listKind: EnvoyFilterList
     plural: envoyfilters
     singular: envoyfilter
-  preserveUnknownFields: true
```
nschhina, line 2875: Please confirm if this change is okay- was autogenerated by running make clean gen

nschhina:
cc: @costinm

networking/v1alpha3/destination_rule.proto
```diff
@@ -920,9 +920,10 @@ message ClientTLSSettings {
   string ca_certificates = 4;
 
   // The name of the secret that holds the TLS certs for the
-  // client including the CA certificates. Applicable
-  // only on Kubernetes. Secret must exist in the same
-  // namespace with the proxy using the certificates.
+  // client including the CA certificates. Can only be used
+  // for traffic originating from Gateway proxy and is
```
smawson, line 924:
```suggestion
  // for traffic originating from the Gateway proxy and is
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -920,9 +920,10 @@ message ClientTLSSettings {
...
+  // applicable only on Kubernetes. Secret must exist in the
```
smawson, line 925:
```suggestion
  // applicable only on Kubernetes. The secret must exist in the
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -920,9 +920,10 @@ message ClientTLSSettings {
...
+  // same namespace with the proxy using the certificates.
```
smawson, line 926:
```suggestion
  // same namespace as the proxy using the certificates.
```

nschhina:
> The comments are accurate for the current implementation.
> 
> I'm not sure if the API spec should document limitations of the implementation - maybe we should say ' istio 1.7 implementation only supports it for K8S Secrets, used in Gateway'. It's not the intent of the API that credentials names can only be stored as k8s secret or only be used in gateway.
> 
> However one way or another we need to communicate to the user that it's not fully implemented.
> 
> Can we add a 'this field is currently alpha and partially implemented' and describe the limitations ?

Done, specifying that it's limited for Istio 1.7 is better as we have future plans to expand it, the end user can also plan accordingly.

nschhina:
cc: @rshriram

networking/v1alpha3/destination_rule.proto
```diff
@@ -920,16 +920,18 @@ message ClientTLSSettings {
...
   // The secret (of type `generic`)should contain the
   // following keys and values: `key: <privateKey>`,
   // `cert: <serverCert>`, `cacert: <CACertificate>`.
   // Secret of type tls for client certificates along with
   // ca.crt key for CA certificates is also supported.
   // Only one of client certificates and CA certificate
   // or credentialName can be specified.
+  // `This field is currently in alpha and is partially implemented.`
+  // Istio 1.7 implementation only supports it for Kubernetes
+  // secrets, used in Gateway.
```
rshriram, line 934: rephrase..

```
//
// **NOTE:** This field is currently applicable only at gateways. Sidecars will continue to use the certificate paths.
```

PR #1509, opened 2020-07-10
Remove extraneous file.

ericvn:
It seems to have been added in a prior PR (duplicate of the same name in mesh/v1alpha1). The one in mesh/v1alpha1 has many updates while this one has not.

The way I ran into this was trying to update the istio.io reference docs which pulls files into the documentation. I found that two files had the same destination which was causing me issues with large deltas in the docs.


ericvn:
I did checkout the commit where the file was added and there is one change:
```
diff istio.mesh.v1alpha1.pb.html mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
4020c4020,4021
< <p>GOTYPE: map[string]interface&lbrace;}</p>
---
> <p>This is required because synthetic type definition has file rather than package scope.
> GOTYPE: map[string]interface&lbrace;}</p>
```

ericvn:
Also, the only change post creation of the file was in https://github.com/istio/api/pull/1408 and this made the change to both files, so nothing was changed in the root file that didn't have a matching change in the directory file.

PR #1510, opened 2020-07-10
Fix broken link for WorkloadSelector

psibi:
Currently various link for WorkloadSelector is broken in the istio
page. Example page where it's currently broken: https://istio.io/latest/docs/reference/config/security/peer_authentication/#PeerAuthentication


psibi:
Closing this as I don't think this is really correct!

PR #1513, opened 2020-07-10
Update virtual_service.proto

sudomann:
Existing documentation for `HTTPRewrite rewrite` can lead to wasted hours during debugging, trying to figure out why envoy sidecars are not logging the rewritten URI.
See https://stackoverflow.com/questions/56924639/istio-virtualservice-httprewrite-being-ignored-completely/56925344#56925344 and ensuing commentary.


sudomann:
> 
> 
> Please run `make gen` or `make clean gen`

I do not have a linux machine to run this on, unfortunately. Could you do this please?

PR #1514, opened 2020-07-13
Fix portNumber typo in EnvoyFilter example

zufardhiyaulhaq:
I spend 2 days figuring why EnvoyFilter is not working but the example is broken.

According to this issue: https://github.com/istio/istio/issues/20640
also, the upstream example still is broken: https://istio.io/latest/docs/reference/config/networking/envoy-filter/#EnvoyFilter

### DIFF
--- a/networking/v1alpha3/envoy_filter.proto
+++ b/networking/v1alpha3/envoy_filter.proto
@@ -74,8 +74,8 @@ import "networking/v1alpha3/sidecar.proto";
 //   - applyTo: NETWORK_FILTER
 //     match:
 //       context: SIDECAR_OUTBOUND # will match outbound listeners in all sidecars
+//       portNumber: 9307
 //       listener:
-//         portNumber: 9307
 //         filterChain:
 //           filter:
 //             name: "envoy.tcp_proxy"
@@ -125,8 +125,8 @@ import "networking/v1alpha3/sidecar.proto";
 //   - applyTo: HTTP_FILTER
 //     match:
 //       context: SIDECAR_INBOUND
+//       portNumber: 8080
 //       listener:
-//         portNumber: 8080
 //         filterChain:
 //           filter:
 //             name: "envoy.http_connection_manager"
--- a/networking/v1alpha3/envoy_filter.pb.go
+++ b/networking/v1alpha3/envoy_filter.pb.go
@@ -51,8 +51,8 @@
 //   - applyTo: NETWORK_FILTER
 //     match:
 //       context: SIDECAR_OUTBOUND # will match outbound listeners in all sidecars
+//       portNumber: 9307
 //       listener:
-//         portNumber: 9307
 //         filterChain:
 //           filter:
 //             name: "envoy.tcp_proxy"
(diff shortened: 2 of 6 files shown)


networking/v1alpha3/envoy_filter.proto
```diff
@@ -74,8 +74,8 @@ import "networking/v1alpha3/sidecar.proto";
 //   - applyTo: NETWORK_FILTER
 //     match:
 //       context: SIDECAR_OUTBOUND # will match outbound listeners in all sidecars
+//       portNumber: 9307
```
hzxuzhonghu, line 77: This is not right, portNumber is under `EnvoyFilter_ListenerMatch`

zufardhiyaulhaq: so the old docs are right?

zufardhiyaulhaq: I test that. it is not working when using `portNumber` under `listener`.
https://gist.github.com/zufardhiyaulhaq/fa151eb288bc17da89cb5df7775e0d59

zufardhiyaulhaq: may also need to change the proto for the table generated.

howardjohn: My guess is the change you made is actually applying it to ALL ports

JoeJasinski:
Thank you @zufardhiyaulhaq - you've been helping me on Slack to address this issue. As we've been discussing, in Istio 1.4.10, when applying a Lua filter to a sidecar inbound using a config similar to that in the [docs](https://archive.istio.io/v1.4/docs/reference/config/networking/envoy-filter/), the filter does not apply (the filter is visible in the output of `istioctl proxy-config listener my-pod-name -o json`, but print statements do not show up in the istio-proxy sidecar logs (even with Lua debugging turned to DEBUG) and the Lua filter does not seem to take effect. 

After moving the portNumber out of the listener section and into the match section, it does work. (I can see print statement output generated by my Lua filters and the Lua code does what it is supposed to do). In fact, removing the portNumber altogether, also works, so I'm not sure what impact the portNumber is having on the EnvoyFilter.  

Here is an example of the filter that is working for me:
```yaml
kind: EnvoyFilter
metadata:
  name: mytest
  ...
spec:
  configPatches:
  - applyTo: HTTP_FILTER
    match:
      context: SIDECAR_INBOUND
      portNumber: 18080
      listener:
        filterChain:
          filter:
            name: "envoy.http_connection_manager"
            subFilter:
              name: "envoy.router"
    patch:
      operation: INSERT_BEFORE
      value: # lua filter specification
       name: envoy.lua
       config:
        inlineCode: |
           ...
```

networking/v1alpha3/envoy_filter.proto
zufardhiyaulhaq: @hzxuzhonghu @howardjohn yes you are right. I realize that the `portNumber` required for matching the `listeners`. And when removing it, it should apply to all `ports`.

But the problem is when we specify the `portNumber`. the filter will not be working.

PR #1519, opened 2020-07-13
Add x-kubernetes-preserve-unknown-fields to instances and handlers

jasonwzm:
This is intended to fix mixer tests in https://github.com/istio/istio/pull/25263

These APIs will be removed after 1.7.


jasonwzm:
cc/ @rshriram

PR #1520, opened 2020-07-14
Make cert signing API public in istio/api.

myidpt:
https://github.com/istio/istio/issues/24553
This copies https://github.com/istio/istio/blob/master/security/proto/istioca.proto with ambiguous field "subject_id" removed.

It offers the public API for custom agents or CAs to integration with Istio. This proto is open to reasonable extensions for adoption of more CAs. All the extensions must be approved by the Istio security WG.

Will make changes on istio-agent and Citadel in istio/istio to support this API.


costinm:
We have the problem that now istio-agent has 2 similar but incompatible protos - Istiod and Google Mesh CA. 
This PR is neither - so now we have 3 protos we need to deal with, and a migration. 

I personally don't care if we adopt Google Mesh CA proto or Istiod - except the names they do the same thing. 
But please - pick one and make an _exact_ copy. Yes, the proto in istiod is alpha - but that's fine, new APIs should
not be beta and it's not even clear there is consensus on this as a long-term protocol ( vs. ACME and other options ).

Main difference is the package name - which unfortunately is significant in gRPC, it's part of the URL. One is
istio.v1.auth 
the other is
google.security.meshca.v1
and your PR has 
istio.security.v1beta1
There seem to also be proto numbers changed. Removing unused fields is fine. 

So if community is ok with having 'google' in the service name - I would use that and we'll have a single gRPC.
Otherwise - let's keep the EXACT proto, since there is no RFC proposing changes to what we have - and the migration
planning, testing, etc associated.

rshriram:
if we are going to make changes to what exists, then I would like to add the bootstrap node metadata here as well.

howardjohn:
I agree with everything costin said so I won't repeat it.

If we are going to invent a 3rd API, which I don't think we should we at the very least need a clear upgrade story which we don't have

myidpt:
I moved this to security/v1alpha1, since the compiler doesn't allow protos in different packages (ca.proto is using package istio.v1.auth, VS istio.security.v1beta1) to be in the same directory.
Now it's backward compatible.
Updated the comments as well.

linsun:
@hzxuzhonghu @irisdingbj do you recall the header you added for multicluster CSR?  I recall it is to help identify the caller (which cluster) who sends the CSR to istiod.

PR #1539, opened 2020-07-20
add TunnelType label

lambdai:
istio-auto-inject will set the label to "tunnel-h2-http"

pilot has individual flag to decide whether the tunnel feature is used.

Part of https://github.com/istio/istio/issues/24540

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -37,4 +37,8 @@ const (
 
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
+
+	// TunnelType is the name of the label to service instance to determine whether support tunnel.
+	// The valid values are tunnel-none, tunnel-h2-http, tunnel-h2-full.
+	TunnelType = "traffic.istio.io/tunnelType"
 )
--- a/proto.lock
+++ b/proto.lock
@@ -13049,13 +13049,7 @@
                         "value": "/v1/services/{service_name}/configs/{config_id}"
                       },
                       {
-                        "name": "additional_bindings",
-                        "aggregated": [
-                          {
-                            "name": "get",
-                            "value": "/v1/services/{service_name}/config"
-                          }
-                        ]
+                        "name": "additional_bindings"
                       }
                     ]
                   }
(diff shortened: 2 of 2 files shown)


howardjohn:
Will we have this label long term? or just as experimental

lambdai:
> Will we have this label long term? or just as experimental

I see 2 questions here:

1. the tunnel ability, from h2-http to h2-http-and-tcp
I once the envoy and pilot has the ability to tunnel all traffic, the tunnel label will assume the latter "http and tcp"

2. Whether the endpoint has the tunnel ability. It's possible the endpoint is not istio sidecar protected.
I will use this label to determine. Unless we have a better solution, the label will stay.

lambdai:
@howardjohn Should I run a script before commit?

lambdai:
Ooops https://github.com/istio/api#updating

lambdai:
@louiscryan

lambdai:
The label is used by the endpoint relay the tunnel support ability of the workload.

More specifically
1. Provide the hint if istio sidecar support http, or all, decided by the envoy version.  
2. Provide the the hint if the workload could support tunnel: a VM with correct setting could support channel as well. 
   The tunnel support could be achieve by envoy or even other http servers.
3. Istiod will filter the endpoints when push EDS cluster. 
    For short term: If a http service opt-in BTS, the endpoints should only includes the endpoints having this label.

label/label.go
```diff
@@ -37,4 +37,8 @@ const (
 
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
+
+	// TunnelType is the name of the label to service instance to determine whether support tunnel.
+	// The valid values are tunnel-none, tunnel-h2-http, tunnel-h2-full.
+	TunnelType = "traffic.istio.io/tunnelType"
```
rshriram, line 43: why do we need this per workload knob? and is this receiving side or sending?

lambdai: This label is used to mark ingress workload.
Egress sidecar need to know if ingress workload support Tunnel.

costinm:
I'm not so sure about this: jumping to API without even having a fully
working prototype doesn't seem right, nor
having this label this narrowly defined.

In reality the protocol used - TCP-over-H2-CONNECT and HTTP PROXY - are not
specific to Istio. Regular
workloads without a sidecar and not using Istio may use this protocol as
well - and we should make sure that
we don't repeat the mistakes with 'istio-mtls' where we locked ourselves
out of interoperating with non-istio
( except in 'plain text mode' ).


So I would suggest keeping this label private - the other labels in this
file also seem to have been rushed,
not clear how operator label for 'pilot' makes sense in Istiod-world, or
what will happen if users doesn't install
with operator.

And we should discuss in networking WG - and with K8S - if we can define a
broader-scoped label.

Also technically for HTTP we are not using a 'tunnel' - but a proxy, i.e.
the client is connecting to a HTTPS_PROXY
running on the same pod. I would like to make sure that BTS is as
compatible with HTTP proxy as possible,
including by testing with non-istio clients ( like a curl --proxy ).

costinm:
To answer Shriram q ( on why per workload label ): since we're using a
standard protocol, any workload including
non-istio could set this label, and we would be able to interoperate.  In
this case, it may be good to have the value
of the label a port:

traffic.istio.io/proxy: 15005

so we don't force non-istio workloads to use our port.

BTS is our best opportunity to better support gRPC and other 'native'
implementations of Istio and interoperability
with non-Istio applications, so I think this is super important to get
right.

lambdai:
OK, I am preparing a more official document than below and a demo.

In short, 

The tunnel announce the workload could serve http2 request with istio cert and route the request respecting istio security policy

1.  For non-h2-connect method, client should provide `X-istio-original-port` header
     The tunnel must respect L7 authorization policy and route the request directly to localhost:<value_of_X-istio-original-port>
2.  (Not implemented yet), For h2-connect method, client should provide `X-istio-tcp-attributes`
     The tunnel must respect L4 authorization policy using the attribute from `X-istio-tcp-attributes` and redirect to localhost:<value_of_X-istio-original-port>

costinm:
I think the big question is if it's ready for API, or can be done with an experimental label - until we have enough feedback and stability to move to proper API.

lambdai:
Got it. I am closing for now and raise it again after

PR #1541, opened 2020-07-21
Revert "Gateway CA certs can be stored in secret-cacert"

rshriram:
Reverts istio/api#1537

The updates have to be made in both versions.. Otherwise, we have postsubmit failures and inconsistencies.


rshriram:
cc @rvennam .. please send the PR again with both versions updated..

PR #1542, opened 2020-07-21
Gateway CA certs can be stored in secret-cacert. Take 2

rvennam:
Sending PR https://github.com/istio/api/pull/1537 again with both versions updated.


rvennam:
I'm not sure why there isn't a `gateway.pb.html` in `v1beta1` folder

PR #1546, opened 2020-07-21
pod label for network

stevenctl:
[Design doc](https://docs.google.com/document/d/1-VoG8qFmW0zhjaVGuKuztH4-xLu16zFZbE4_0g5z5sk/edit?usp=sharing)

Part of https://github.com/istio/istio/pull/25554

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -26,7 +26,7 @@ const (
 	// IstioCanonicalServiceRevision is the name of label for the Istio Canonical Service revision for a workload instance.
 	IstioCanonicalServiceRevision = "service.istio.io/canonical-revision"
 
-	// IstioRev is the Istio control plane revision associated with the resource; e.g. "canary"
+	// IoIstioRev is the Istio control plane revision associated with the resource; e.g. "canary"
 	IstioRev = "istio.io/rev"
 
 	// IstioOperatorComponent is the Istio operator component name of the resource, e.g. "Pilot"
@@ -38,9 +38,4 @@ const (
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
 
-	// IstioNetwork is the network name given to service instances used for providing cross-network EDS.
-	IstioNetwork = "topology.istio.io/network"
-
-	// IstioCluster is the cluster name where a service instance resides.
-	IstioCluster = "topology.istio.io/cluster"
 )
--- a/label/label.go
+++ b/label/label.go
@@ -40,7 +40,4 @@ const (
 
 	// IstioNetwork is the network name given to service instances used for providing cross-network EDS.
 	IstioNetwork = "topology.istio.io/network"
-
-	// IstioCluster is the cluster name where a service instance resides.
-	IstioCluster = "topology.istio.io/cluster"
 )
--- a/label/label.go
+++ b/label/label.go
@@ -38,6 +38,9 @@ const (
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
 
-	// IstioNetwork is the network name given to service instances used for providing cross-network EDS
-	IstioNetwork = "networking.istio.io/network"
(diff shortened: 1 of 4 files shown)


smawson:
Can you link the approved RFC or design doc into the description of this PR? We're trying to get better at verifying that API changes have been discussed and approved and tracking that in the PRs.

label/label.go
```diff
@@ -37,4 +37,10 @@ const (
 
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
+
+	// IstioNetwork is the network name given to service instances used for providing cross-network EDS.
+	IstioNetwork = "topology.istio.io/network"
```
rshriram, line 42: is this the same as the service entry's network field?

label/label.go
```diff
@@ -37,4 +37,10 @@ const (
...
+
+	// IstioCluster is the cluster name where a service instance resides.
+	IstioCluster = "topology.istio.io/cluster"
```
rshriram, line 45: where is this used?

label/label.go
stevenctl: Similar behavior, but for explicitly setting the network for a pod/namespace. Doc incoming

hzxuzhonghu: is  this same as `ISTIO_META_NETWORK`?

stevenctl: This will set ISTIO_META_NETWORK if the pod has the label pre-injection, otherwise the label will be injected the same way as meta network. 

Users setting this by hand is more of an advanced use-case/edge-case. the primary purpose of having this be a label is to decouple ServiceRegistry/Cluster from Network.

label/label.go
stevenctl: This will be injected along with ISTIO_META_CLUSTER - we can utilize this within k8s eds and cleanup a lot there. While technically it _could_ be overridden by a user, there is no real reason to.

costinm:
My point was: keep cluster and network separate.

Cluster can easily be set by the registry (as is today), and semantic of
user overriding it is less clear.
We can debate it in separate PR.

label/label.go
costinm: Yes, it's the 'formal'/promoted equivalent of ISTIO_META_NETWORK.

Regarding ServiceEntry - we should document what happens if both label and network field are set. Report as error ? 

The label can be set explicitly by user - but can also be set as part of injection, like auto-mtls.

label/label.go
costinm: I would leave this for a separate PR. The use case is far less clear.

stevenctl:
@linsun @rshriram @smawson 

Can we pull this in now that Cluster is removed?

label/label.go
```diff
@@ -37,4 +37,7 @@ const (
 
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
+
+	// IstioNetwork is the network name given to service instances used for providing cross-network EDS.
```
smawson, line 41: If this is user facing I wouldn't use "cross-network EDS" here and instead just say that this is used to support multi-network use cases.

smawson:
You could copy some of the comment from the "network" field on WorkloadEntry since this is the same thing right?

https://istio.io/latest/docs/reference/config/networking/workload-entry/#WorkloadEntry

label/label.go
```diff
@@ -37,4 +37,11 @@ const (
 
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
+
+	// IstioNetwork enables Istio to group endpoints resident in the same L3 domain/network.
+	// All endpoints in the same network are assumed to be directly reachable from one another.
+	// When endpoints in different networks cannot reach each other directly, an Istio Gateway can
+	// be used to establish connectivity (usually using the AUTO_PASSTHROUGH mode in a  Gateway Server).
```
smawson, line 44:
```suggestion
	// be used to establish connectivity (usually using the AUTO_PASSTHROUGH mode in a Gateway Server).
```

PR #1540, opened 2020-07-21
Fix invalid link for WorkloadSelector

shamsher31:
This was reported multiple times on istio.io docs. This PR fixes the workload selector to redirect it to a valid link. 

Related issues:
https://github.com/istio/api/issues/1511
https://github.com/istio/istio.io/issues/7203
https://github.com/istio/istio.io/issues/6877
https://github.com/istio/istio.io/issues/6921

PR #1550, opened 2020-07-22
Fix broken link for WorkloadSelector

shamsher31:
The original PR https://github.com/istio/api/pull/1540 appended `#WorkloadSelector` twice in the generated protobufs.

This PR removes it because `#WorkloadSelector` automatically gets appended in generated protobufs so no need to include it in `$location`.

Suggested by @brian-avery  https://github.com/istio/api/pull/1548#discussion_r458866256 
I think we can close cherry-pick https://github.com/istio/api/pull/1548 and the new one will be created once this is merged in master?

### DIFF
--- a/security/v1beta1/authorization.pb.html
+++ b/security/v1beta1/authorization.pb.html
@@ -181,7 +181,7 @@ spec:
 <tbody>
 <tr id="AuthorizationPolicy-selector">
 <td><code>selector</code></td>
-<td><code><a href="https://istio.io/latest/docs/reference/config/networking/sidecar/#WorkloadSelector#WorkloadSelector">WorkloadSelector</a></code></td>
+<td><code><a href="https://istio.io/latest/docs/reference/config/networking/sidecar/#WorkloadSelector">WorkloadSelector</a></code></td>
 <td>
 <p>Optional. Workload selector decides where to apply the authorization policy.
 If not set, the authorization policy will be applied to all workloads in the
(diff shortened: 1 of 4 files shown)


type/v1beta1/selector.proto
```diff
@@ -17,7 +17,7 @@ import "google/api/field_behavior.proto";
 
 // $title: Workload Selector
 // $description: Definition of a workload selector.
-// $location: https://istio.io/latest/docs/reference/config/networking/sidecar/#WorkloadSelector
+// $location: https://istio.io/latest/docs/reference/config/networking/sidecar/
```
ericvn, line 20: Do not put latest here as it will fail when linted in istio.io.

security/v1beta1/authorization.pb.html
```diff
@@ -181,7 +181,7 @@ <h2 id="AuthorizationPolicy">AuthorizationPolicy</h2>
 <tbody>
 <tr id="AuthorizationPolicy-selector">
 <td><code>selector</code></td>
-<td><code><a href="https://istio.io/latest/docs/reference/config/networking/sidecar/#WorkloadSelector#WorkloadSelector">WorkloadSelector</a></code></td>
+<td><code><a href="https://istio.io/latest/docs/reference/config/networking/sidecar/#WorkloadSelector">WorkloadSelector</a></code></td>
```
ericvn, line 184: Do not put latest here as it will fail when linted in istio.io.  See: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_istio.io/7764/lint_istio.io/7508/build-log.txt

frankbu:
@shamsher31 I think this is out of sync with both 1.6 and 1.7 branches, so you will need to manually cherrypick it to release-1.6 and release-1.7.

PR #1552, opened 2020-07-22
Extend authorization policy action with 'Audit' action

davidraskin:
We'd like to add an audit action to the Authorization Policy resource, which would be used to determine whether requests should be logged, and can be supported by Istio telemetry v2 plugins. When a policy matches the request, an attribute would be set through the envoy RBAC filter, which plugins can retrieve to check the decision.

See [istio/istio#25591](https://github.com/istio/istio/issues/25591)


davidraskin:
cc @liminw @mandarjog @louiscryan @smawson @yangminzhu

davidraskin:
Do you recommend we discuss specifically what the docs should say first, and get a PR in for those?

security/v1beta1/authorization.proto
```diff
@@ -34,9 +34,21 @@ import "type/v1beta1/selector.proto";
 // 3. If any of the ALLOW policies match the request, allow the request.
 // 4. Deny the request.
 //
-// For example, the following authorization policy sets the `action` to "ALLOW"
-// to create an allow policy. The default action is "ALLOW" but it is useful
-// to be explicit in the policy.
+// Istio Authorization Policy also supports the AUDIT action to decide whether to log requests. 
+// The decision can be read by telemetry plugins using the function getAuditPolicy
```
smawson, line 38: This is user facing doc, is this what we actually want to tell users? That they should go implement a telemetry filter to use this, and here is the documentation on how to do that?

This should instead be focused on when users should use this and how, and leave implementation to a separate developer-facing doc.

BTW, if the user does not have an audit-capable filter installed, and they add an audit policy, what happens? Do we have a way to detect and alert them?

davidraskin:
Audit policy [RFC](https://docs.google.com/document/d/1pcbgPXEYcSEVdYQbCL1vz9IBMCoRTdNUuyfQm0R2otA/edit#)

security/v1beta1/authorization.proto
```diff
@@ -34,9 +34,21 @@ import "type/v1beta1/selector.proto";
...
+// defined [here](https://github.com/istio/proxy/blob/master/extensions/common/context.h).
+// A call to getAuditPolicy will only return `true` in the following cases:
```
mandarjog, line 40: The context.h reference is fine in the attached design doc, but not in the API doc.
wWe should only specify the filter state key that this process is setting, and how a downstream filter like logging, or even metrics can use it. The setting of the filter state is the only externally observable action of `AUDIT`

davidraskin: Yeah, I'll be changing that. Is it fine to refer to the envoy specific dynamic metadata, or would it be better to just provide a list of access logging plugins that check this key?

smawson:
I would put anything envoy specific in a hide-from-docs section, that shouldn't show up in the public docs. I like the idea of public API docs listing what telemetry filters support this action.

security/v1beta1/authorization.proto
```diff
@@ -34,16 +34,16 @@ import "type/v1beta1/selector.proto";
...
-// The decision can be read by telemetry plugins using the function getAuditPolicy
-// defined [here](https://github.com/istio/proxy/blob/master/extensions/common/context.h).
-// A call to getAuditPolicy will only return `true` in the following cases:
+// Istio Authorization Policy also supports the AUDIT action to decide whether to log requests.
+// Telemetry v2 access logging plugins can be set to abide by the audit decision.
+// Currently this is supported by the [Stackdriver](https://istio.io/latest/docs/reference/config/proxy_extensions/stackdriver/) plugin.
+// A request will be audited in the following cases:
 //
 // - There exists an AUDIT policy on the workload that matches the request
 // - There are no AUDIT policies on the workload
```
smawson, line 43: I'm confused by this, the request will be audited if there are no AUDIT policies on the workload?

davidraskin: That does come off as weird. The point was to log everything if there are no policies. If we're making two separate streams for logs though, then it might be better to log only when an audit policy exists.

smawson: To log everything they could create a global empty audit policy right? So I'd just say that a request is audited if there is an audit rule that matches the request.

yangminzhu: this doesn't sound right to me, why would a request be audited even if there is no AUDIT policies on it? This is not the case described in the design doc.

security/v1beta1/authorization.proto
```diff
@@ -34,9 +34,21 @@ import "type/v1beta1/selector.proto";
...
-// to be explicit in the policy.
+// Istio Authorization Policy also supports the AUDIT action to decide whether to log requests.
+// Telemetry v2 access logging plugins can be set to abide by the audit decision.
+// Currently this is supported by the [Stackdriver](https://istio.io/latest/docs/reference/config/proxy_extensions/stackdriver/) plugin.
+// A request will be audited in the following cases:
+//
+// - There exists an AUDIT policy on the workload that matches the request
+// - There are no AUDIT policies on the workload
+// 
+// AUDIT policies do not affect whether requests are allowed to the workload.
```
yangminzhu, line 45: allowed or denied

yangminzhu: Also can we move this to the above where we first talked about the AUDIT action? (after line 37)

security/v1beta1/authorization.proto
```diff
@@ -34,9 +34,21 @@ import "type/v1beta1/selector.proto";
 // 3. If any of the ALLOW policies match the request, allow the request.
 // 4. Deny the request.
 //
-// For example, the following authorization policy sets the `action` to "ALLOW"
-// to create an allow policy. The default action is "ALLOW" but it is useful
-// to be explicit in the policy.
+// Istio Authorization Policy also supports the AUDIT action to decide whether to log requests.
+// Telemetry v2 access logging plugins can be set to abide by the audit decision.
```
yangminzhu, line 38: Can we make it super clear that the AUDIT action in AuthZ policy doesn't actually record or log anything anywhere, what it really does is to record the decision that the request should be audited, and that decision **must** be implemented by some other logging plugins to actually record and log the request, for example, the Stackdriver?

And more importantly, if there is no such logging plugin configured, **nothing** will be logged or audited at all? (To make sure the user doesn't misuse the feature and think this is all they need for auditing a request but later found out nothing is really audited by just using the authz policy).

Also is the `Telemetry v2 access logging` the same as the `Stackdriver`? What about the Envoy Access Log filter?

davidraskin: Telemtry v2 access logging is stackdriver, and in the furture will be all access log plugins. For envoy access log filter, a boolean needs to be added to meshconfig (which will trigger configuration of the metadata access log filter). I'll add this to the documentation if this is something we want to do as well.

security/v1beta1/authorization.proto
davidraskin: Yep you're right. I was thinking of it from the perspective of ALLOW policy where if there is no policy on the workload requests are allowed by default. But obviously this doesn't apply here.

security/v1beta1/authorization.proto
```diff
@@ -34,9 +34,19 @@ import "type/v1beta1/selector.proto";
...
 // 4. Deny the request.
 //
-// For example, the following authorization policy sets the `action` to "ALLOW"
-// to create an allow policy. The default action is "ALLOW" but it is useful
-// to be explicit in the policy.
+// Istio Authorization Policy also supports the AUDIT action to decide whether to log requests.
+// AUDIT policies do not affect whether requests are allowed or denied to the workload.
+// Requests will be allowed or denied based solely on ALLOW and DENY policies.
+//
+// Telemetry v2 access logging plugins can be set to abide by the audit decision.
```
yangminzhu, line 41: Can we change it to the following to make this part more clear (line 41-44), thanks:

A request will be internally marked that it should be audited if there is an AUDIT policy on the workload that matches the request. 

A separate plugin must be configured and enabled to actually fulfill the audit decision and complete the audit behavior. The request will not be audited if there are no such supporting plugins enabled.

Currently the only supported plugin is the [Telemetry v2 Stackdriver](https://istio.io/latest/docs/reference/config/proxy_extensions/stackdriver/) plugin.

PR #1554, opened 2020-07-22
WorkloadGroup CRD

justinwei2:
Added a WorkloadGroup CRD to be used with the sidecar VM commands.

Initial approved design doc:
https://docs.google.com/document/d/1PdS1a0lydRIhzNBjk9l-D1WkahfGqI9A6fU4kyJnk9k/edit#heading=h.xw1gqgyqs5b
Current design doc:
https://docs.google.com/document/d/1-RVOpCGbcDdr-yM4UrT0WgfHZglgQZlPL1d1aZbqEbk/edit?ts=5efb9259#heading=h.xw1gqgyqs5b


networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,89 @@
...
+
+import "google/api/field_behavior.proto";
+
+// $schema: istio.networking.v1alpha3.WorkloadGroup
+// $title: Workload Group
+// $description: Describes a collection VM workload instances.
+// $location: https://istio.io/docs/reference/config/networking/workload-group.html
+// $aliases: [/docs/reference/config/networking/v1alpha3/workload-group]
+
+// `WorkloadGroup` describes a collection of non-Kubernetes (VM) workload
```
smawson, line 25: FYI these don't actually have to be VMs, slowly trying to fix our docs to refer to these as non-containerized endpoints or non-pod endpoints or something. For example bare metal machines, or IoT devices, etc.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,89 @@
...
+import "google/api/field_behavior.proto";
+
+// $schema: istio.networking.v1alpha3.WorkloadGroup
+// $title: Workload Group
+// $description: Describes a collection VM workload instances.
+// $location: https://istio.io/docs/reference/config/networking/workload-group.html
+// $aliases: [/docs/reference/config/networking/v1alpha3/workload-group]
+
+// `WorkloadGroup` describes a collection of non-Kubernetes (VM) workload 
+// instances that will serve as identical applications. It provides a
```
smawson, line 26:
```suggestion
// instances. It provides a 
```

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,89 @@
...
+// $aliases: [/docs/reference/config/networking/v1alpha3/workload-group]
+
+// `WorkloadGroup` describes a collection of non-Kubernetes (VM) workload 
+// instances that will serve as identical applications. It provides a 
+// specification that the workload instances can use to bootstrap their 
+// proxies, including the metadata and identity.
+//
+// The following example declares a workload group representing a collection 
+// of VM workloads that will be registered under `details` in namespace
+// `bootstore`. The set of labels will be associated with each workload
```
smawson, line 32:
```suggestion
// `bookstore`. The set of labels will be associated with each workload 
```

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,89 @@
...
+// proxies, including the metadata and identity.
+//
+// The following example declares a workload group representing a collection 
+// of VM workloads that will be registered under `details` in namespace
+// `bootstore`. The set of labels will be associated with each workload 
+// instance during the VM bootstrap process, and the workloads will expose the
+// ports 3550 and 8080 to Istio. The workload group will be under the network
+// `remote` and service account `default`.
+//
+// {{<tabset category-name="example">}}
```
costinm, line 37: I would like some clear statement on what happens if Deployments and VMs running the same 
 version are mixed - it is a very common case.

Validation at least should make sure that the values don't conflict. Keep in mind endpoint selection and all other logic is based on labels - we do not want to have to track if a workload was created via Deployment, Pod, ReplicaSet - or WorkloadGroup.

Also ideally in 'lift and shift' or 'mixed K8S/VM' workloads we shouldn't require users to declare both the Deployment and WorkloadGroup - if a VM matches a Deployment ( label selector and service account ) - we should just use it.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,89 @@
...
+// {{<tab name="v1alpha3" category-value="v1alpha3">}}
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: WorkloadGroup
+// metadata:
+//   name: details
+//   namespace: bookstore
+// spec:
+//   labels:
```
costinm, line 46: Since WorkloadGroup is similar with Deployment ( and WorkloadEntry with Pod ) - why use a different syntax ? Deployment uses "selector.matchLabels" - and I suspect the same label matching will be used when VMs connect to Istiod.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,89 @@
...
+//     app.kubernetes.io/name: details-legacy
+//   network: remote
```
costinm, line 48: I don't think this makes sense - the pod or VM declares its network ( label as of last networking WG meeting, or based on the IP address and meshconfig ). Deployments don't declare the network, and we don't want inconsistency ( like requiring one WorkloadGroup per network )

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,89 @@
...
+//   ports:
+//     grpc: 3550
+//     http: 8080
```
costinm, line 51: Can you remind me why we need ports ? In pod they are optional AFAIK, I know in the past we used them for iptables - but locally. Service selection takes care of the port.

And I think we agreed that on VMs it is possible for each workload entry to use a different port ( because the port may be in use - VMs can run multiple apps, including multiple instances of a workload entry via docker ).

@howardjohn - do we use the container port in any other code in Istiod ? 

I would expect the VM to send the list of real port and labels - and for us to just copy it over in the WorkloadEntry, not force the same port. I think even in K8S it is possible for pods to use named ports and have different container port in a service.

howardjohn: We don't - but Kubernetes does. The only reason we don't is because we read Endpoints, the result of processing Service+Pod. For this case we do not have a pre-processed Service+WorkloadEntry so we do actually need the portsfor named ports.

WorkloadGroup (and deployment) describe a set of identical resources - if you want different ports you can make a new WorkloadGroup selected by the same service?

costinm:
We certaining do need them in WorkloadEntry - just like they exist in
EndpointSlice and Endpoint.

The question is why do we need them in WorkloadGroup.

You can have identical resources - running on VMs that just can't use the
port, because another
app on the VM is using it.

I thought we use the labels to find the services - which also have
targetPort as a default -
and we create inbound listeners based on that.

smawson:
@costinm I think the intent here is that this represents a template for WorkloadEntry. The flow is that the settings here will be copied into the ProxyConfig that goes to the VM, which is then used in auto-registration to create the WorkloadEntry.

So the labels are defining what labels will show up on WorkloadEntry, not selecting them.

costinm:
That makes sense. Maybe follow the Deployment pattern and just have a
'template:' inside ? Consistency and fewer mistakes...

hzxuzhonghu:
One question: how to deal with WLE in multi networks if there is only one network in WLG?

costinm:
Agreed, I had the same comment - at least network should be removed from
WLG.

I'm having more doubts about the assumptions: having a 'template' to
generate a config that is deployed to the VM which in
turns gets sent back to Istiod.

This assumes that all VMs will be identical - same network, same ports,
same labels. In practice - many VMs will
be able to have local labels ( metadata server on most clouds ), may be on
different networks - also detectable
from environment or the IP.

Can we start with the most minimal set of information that we are 100% sure
we won't need to deprecate or conflict with
other features ? Service Account and a set of default labels seems
sufficient for the initial work ( with labels clarified: additional
metadata should be possible to add, it's not the full set).

justinwei2:
> My overall comment is that there are a bunch of open discussions in the design doc specifically about the structure of the API, things like whether we should just nest a WorkloadEntry.
> 
> This PR looks like we just ignored those (and we haven't approved the design doc), please try to follow up on the comments.
> 
> Otherwise I can just copy all my comments on the doc to the PR but that doesn't seem particularly efficient...



> My overall comment is that there are a bunch of open discussions in the design doc specifically about the structure of the API, things like whether we should just nest a WorkloadEntry.
> 
> This PR looks like we just ignored those (and we haven't approved the design doc), please try to follow up on the comments.
> 
> Otherwise I can just copy all my comments on the doc to the PR but that doesn't seem particularly efficient...

Updated with your suggestions/fixes

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,96 @@
...
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message WorkloadGroup {
+  // Template to be used for the generation of WorkloadEntry resources that belong
+  // to this WorkloadGroup
+  WorkloadEntryTemplate template = 1;
+}
+
+message WorkloadEntryTemplate {
```
smawson, line 90: Need documentation on WorkloadEntryTemplate

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,96 @@
...
+//   limitations under the License.
+
+syntax = "proto3";
+
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "networking/v1alpha3/workload_entry.proto";
+
+// $schema: istio.networking.v1alpha3.WorkloadGroup
+// $title: Workload Group
+// $description: Describes a collection of workload instances.
```
linsun, line 22:
```suggestion
// $description: Describes a collection of non-Kubernetes workload instances.
```

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,96 @@
...
+// kind: WorkloadGroup
+// metadata:
+//   name: checkoutservice
+//   namespace: hipster
+// spec:
+//   template:
+//     metadata:
+//       labels:
+//         app.kubernetes.io/name: checkoutservice
+//         app.kubernetes.io/version: "1.3.4"
```
linsun, line 49: How the version works?  It was not mentioned in the desp earlier.

networking/v1alpha3/workload_group.proto
linsun: Agreed with Costin's comment.  What if I also have a deployment in k8s that is `checkoutservice` in namespace `hipster`?

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,96 @@
...
+//       serviceAccount: default
+// ```
+// {{</tab>}}
+// {{</tabset>}}
+//
+package istio.networking.v1alpha3;
+
+option go_package = "istio.io/api/networking/v1alpha3";
+
+// WorkloadGroup enables specifying the properties of a single non-Kubernetes workload for bootstrap.
```
linsun, line 63: A single non-k8s workload, if I have v1.3.5 of the `checkoutservice` in namespace`hipster`, do I create another workload group?

networking/v1alpha3/workload_group.proto
douglas-reid: these are labels on the WE resources that will be created. `version` is just an example.

networking/v1alpha3/workload_group.proto
douglas-reid: workload here is not necessarily `workload instance`, though I do think that `single` might be misleading. In k8s, if you wanted to change the version labels on your existing deployments, would you update the existing deployment or create a new one? I think the answer should be the same for WG.

networking/v1alpha3/workload_group.proto
douglas-reid: We definitely want to support the ability to distribute a workload across k8s and vm envs.

What is the state of validation/conflict today on `WorkloadEntry` and `Deployment` (in terms of labels and endpoint selection)? I don't believe that `WorkloadGroup` is changing that dynamic by referencing `WorkloadEntry`.

The ability to use a `Deployment` to control VM configuration and packaging is an interesting use case. I don't think the presence of `WorkloadGroup` rules that out -- and whatever tooling we develop around it could be expanded to support such a practice.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,97 @@
...
+  WorkloadEntryTemplate template = 1;
+}
+
+// `WorkloadEntryTemplate` describes the data a `WorkloadGroup` should have when created from a template.
+message WorkloadEntryTemplate {
+  // Metadata that will be used for all corresponding WorkloadEntries.
+  k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta metadata = 1;
+
+  // Describes the WorkloadEntry resources that will be created.
+  WorkloadEntry spec = 2;
```
rshriram, line 96: please dont use as is.. workload entry has address fields which shouldnt be used in the deployment. 
all you need is service account, port specs, locality, and network fields.

douglas-reid: We started there, and there were plenty of objections. Switching to WE as a template was directly encouraged as a way to avoid fights over individual fields and to be able to adapt as the WE schema evolves.

We can provide validation that prevents use of the address fields and/or document that those fields will be ignored.

Thoughts?

smawson: See the earlier discussion on this PR and the discussion on the design doc. We moved towards a template instead of inlining the fields so that we automatically keep in sync with workload entry.

We can make it an error to fill out certain fields (such as address).

douglas-reid:
@linsun @louiscryan @duderino @nrjpoddar et al., any chance for a second approval on this PR?

nrjpoddar:
@douglas-reid I started looking at it but since I was not involved in the design phase it's quite difficult to understand what pieces from design are yet to be implemented. I will look closely but someone like @linsun who was involved in design is a better candidate for review.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,97 @@
...
+import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+import "networking/v1alpha3/workload_entry.proto";
+
+// $schema: istio.networking.v1alpha3.WorkloadGroup
+// $title: Workload Group
+// $description: Describes a collection of workload instances.
+// $location: https://istio.io/docs/reference/config/networking/workload-group.html
+// $aliases: [/docs/reference/config/networking/v1alpha3/workload-group]
+
+// `WorkloadGroup` describes a collection of workload instances.
```
nrjpoddar, line 26: Can we add a line about workloads can be VMs/K8s pods or a mixture of two? It's quite unclear as to why a user should create this resource?

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,97 @@
...
+message WorkloadGroup {
+  // Template to be used for the generation of WorkloadEntry resources that belong
+  // to this WorkloadGroup
+  WorkloadEntryTemplate template = 1;
+}
+
+// `WorkloadEntryTemplate` describes the data a `WorkloadGroup` should have when created from a template.
+message WorkloadEntryTemplate {
+  // Metadata that will be used for all corresponding WorkloadEntries.
+  k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta metadata = 1;
```
nrjpoddar, line 93: If this CRD is created in K8s then there will be an outer namespace and namespace inside object meta, is that correct?

Do they need to be consistent?

networking/v1alpha3/workload_group.proto
linsun: Sorry - what I meant was if the version 1.3.4 is 1 workload group and I need to create a new workload group to present a different version of the checkout service in the hipster namespace, correct?

networking/v1alpha3/workload_group.proto
linsun: I agree the desp isn't clear.   I expect to see:
1) Can I use workload Group for instances on VM or a combination of instances on VM and pods in K8S?
2) If it is a combination, for a given workload group, are same version expected?
3) What is the relationship with other resources, e.g. Service Entry, Workload Entry.

networking/v1alpha3/workload_group.proto
douglas-reid: If you change any labels in the `WorkloadGroup`, there will not be any automated changes to the underlying already running VM instances. If you wanted to apply new labels for a group of VM instances, then you would likely want to create a new `WorkloadGroup` to use in bootstrapping the new instances. If you instead were deleting the already running instances and replacing them, you could optionally update the existing `WorkloadGroup`.

networking/v1alpha3/workload_group.proto
douglas-reid: `WorkloadGroup` is *only* intended to be used for non-k8s workloads. It is meant to be used to mimic the existing sidecar injection + deployment spec model used for k8s workloads to bootstrap istio-proxies.

For hybrid workload (VMs and pods), the current thinking is to use the existing `DeploymentSpec` for bootstrap purposes (in future iterations). (Reminder: this is alpha). 

`version` is just an example label. We should not focus too much on that as a thing in this API. Treat that the same way you would as a version label in a `Deployment`s `PodTemplateSpec`.

`WorkloadGroup` provides a template for `WorkloadEntry`, similar to the relationship to `Deployment` and `Pod`. It has no relation to `ServiceEntry`. It says nothing about addressing a service/host.

networking/v1alpha3/workload_group.proto
douglas-reid: the namespace for the templated `WorkloadEntry` will be ignored in favor of the namespace of `WorkloadGroup` (as it is for the Pod template inside of Deployment).

networking/v1alpha3/workload_group.proto
nrjpoddar: That's good to know, please add some of this content to user facing comments so they have more clarity on when/how to use the new API.

networking/v1alpha3/workload_group.proto
nrjpoddar: Make sense

justinwei2:
@nrjpoddar Please let me know what you think of the comments

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// $title: Workload Group
+// $description: Describes a collection of workload instances.
+// $location: https://istio.io/docs/reference/config/networking/workload-group.html
+// $aliases: [/docs/reference/config/networking/v1alpha3/workload-group]
+
+// `WorkloadGroup` describes a collection of workload instances.
+// It provides a specification that the workload instances can use to bootstrap 
+// their proxies, including the metadata and identity. It is only intended to 
+// be used with non-k8s workloads, and is meant to mimic the existing sidecar
+// injection and deployment specification model used for k8s workloads to
```
nrjpoddar, line 30:
```suggestion
// injection and deployment specification model used for Kubernetes workloads to 
```

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// $schema: istio.networking.v1alpha3.WorkloadGroup
+// $title: Workload Group
+// $description: Describes a collection of workload instances.
+// $location: https://istio.io/docs/reference/config/networking/workload-group.html
+// $aliases: [/docs/reference/config/networking/v1alpha3/workload-group]
+
+// `WorkloadGroup` describes a collection of workload instances.
+// It provides a specification that the workload instances can use to bootstrap 
+// their proxies, including the metadata and identity. It is only intended to 
+// be used with non-k8s workloads, and is meant to mimic the existing sidecar
```
nrjpoddar, line 29:
```suggestion
// be used with non-k8s workloads like Virtual Machines, and is meant to mimic the existing sidecar
```

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// injection and deployment specification model used for k8s workloads to 
+// bootstrap Istio proxies.
+//
+// For hybrid workloads with both VMs and k8s pods, use the existing `DeploymentSpec` for bootstrap purposes.
```
nrjpoddar, line 33:
```suggestion
// For hybrid workloads with both Virtual Machines and Kubernetes workloads, you should use the existing Kubernetes `Deployment` resource.
```

nrjpoddar, line 33: How should one go about using Deployment resource for VMs?

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+//
+// The following example declares a workload group representing a collection 
+// of workloads that will be registered under `checkoutservice` in namespace
+// `hipster`. The set of labels will be associated with each workload 
+// instance during the bootstrap process, and the workloads will expose the
+// ports 3550 and 8080 to Istio and use and service account `default`.
+// `app.kubernetes.io/version` is just an arbitrary example of a label.
+//
+// {{<tabset category-name="example">}}
```
nrjpoddar, line 42: Can we use bookinfo example which is used for all (most) examples on istio.io? Unless there's a good reason not to, I would suggest that we continue to use bookinfo.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// {{</tab>}}
+// {{</tabset>}}
+//
+package istio.networking.v1alpha3;
+
+option go_package = "istio.io/api/networking/v1alpha3";
+
+// `WorkloadGroup` enables specifying the properties of a single workload for bootstrap and
+// provides a template for `WorkloadEntry`, similar to the relationship between `Deployment` and `Pod`. 
+// `WorkloadGroup` has no relationship to `ServiceEntry` and says nothing about a service or host.
```
nrjpoddar, line 71:
```suggestion
// `WorkloadGroup` has no relationship to resources which control service registry like `ServiceEntry` and as such doesn't configure host name for these workloads.
```

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message WorkloadGroup {
+  // Template to be used for the generation of WorkloadEntry resources that belong
+  // to this WorkloadGroup
+  WorkloadEntryTemplate template = 1;
+}
+
+// `WorkloadEntryTemplate` describes the data a `WorkloadGroup` should have when created from a template.
```
nrjpoddar, line 98:
```suggestion
// `WorkloadEntryTemplate` is the underlying workload template that is associated with the workload group 
```

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+message WorkloadEntryTemplate {
+  // Metadata that will be used for all corresponding WorkloadEntries.
```
nrjpoddar, line 100: Can there be more than 1 WorkloadEntry?

nrjpoddar:
> @nrjpoddar Please let me know what you think of the comments

Thanks @justinwei2 for enhancing the comments. Creating user facing documentation always takes few iterations so we will continue to improve as we get more feedback from other TOC members.

networking/v1alpha3/workload_group.proto
justinwei2: Removed the line about deployment spec since it's something to be implemented later

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// their proxies, including the metadata and identity. It is only intended to 
+// be used with non-k8s workloads like Virtual Machines, and is meant to mimic 
+// the existing sidecar injection and deployment specification model used for 
+// Kubernetes workloads to bootstrap Istio proxies.
+//
+// The following example declares a workload group representing a collection 
+// of workloads that will be registered under `reviews` in namespace
+// `bookinfo`. The set of labels will be associated with each workload 
+// instance during the bootstrap process, and the workloads will expose the
+// ports 3550 and 8080 to Istio and use and service account `default`.
```
nrjpoddar, line 37: `and the workloads will expose the ports 3550 and 8080 to Istio` -> what does this mean exactly?

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// ```
+// {{</tab>}}
+// {{</tabset>}}
+//
+package istio.networking.v1alpha3;
+
+option go_package = "istio.io/api/networking/v1alpha3";
+
+// `WorkloadGroup` enables specifying the properties of a single workload for bootstrap and
+// provides a template for `WorkloadEntry`, similar to the relationship between `Deployment` and `Pod`.
```
nrjpoddar, line 68:
```suggestion
// provides a template for `WorkloadEntry`, similar to the how `Deployment` specifies properties of workloads via `Pod` templates. 
```

networking/v1alpha3/workload_group.proto
nrjpoddar: // Metadata that will be used for all corresponding WorkloadEntries. -> Can there be more than 1 WorkloadEntry? Why the use of `WorkloadEntries` here?

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+  // Template to be used for the generation of WorkloadEntry resources that belong
+  // to this WorkloadGroup
+  WorkloadEntryTemplate template = 1;
+}
+
+// `WorkloadEntryTemplate` is the underlying workload template that is associated with the workload group.
+// A `WorkloadGroup` can have more than one `WorkloadEntry`.
+message WorkloadEntryTemplate {
+  // Metadata that will be used for all corresponding WorkloadEntries.
+  k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta metadata = 1;
```
howardjohn, line 101: Can we avoid importing k8s apis? I don't think we do that in anything else. We only respect 2 fields anyways - namespace and name, so its a bit misleading to have all of them (I know k8s does it)

networking/v1alpha3/workload_group.proto
rshriram: Can we please please NOT import k8s stuff? Importing istio/api in other tools is already a super big pain. throw in the k8s deps and it gets very painful. I dont know who it was that pushed for this but let us please keep the API independent of k8s. Our docs need to be self sufficient for all the fields. We cannot push people to a non existent or a different api reference (that we dont control).

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+
+  // Describes the WorkloadEntry resources that will be created.
+  WorkloadEntry spec = 2;
```
rshriram, line 104: can you rename spec to template please? Otherwise, its confusing to read. spec.spec.

rshriram: Secondly, please mention explicitly that address and label fields SHOULD NOT be specified. Otherwise, it will be a trial and error for people to find out what is and what is not allowed.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,105 @@
...
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1alpha3
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message WorkloadGroup {
+  // Template to be used for the generation of WorkloadEntry resources that belong
+  // to this WorkloadGroup
+  WorkloadEntryTemplate template = 1;
```
rshriram, line 94: this can be `WorkloadEntry template = 1`. You dont need the other WorkloadEntryTemplate object once you remove the metadata thing. Make sure to mention that address and label fields are invalid.

smawson:
The Operator API has a copy of ObjectMeta, although it only has name and namespace and we want name, namespace, labels, annotations.  We could add labels and annotations to the copy in Operator or we could create a copy here as well. We actually don't really need name or namespace, just labels and annotations.

I do think we should keep the structure as it is since it is consistent with other templates, so I wouldn't change those (outer template, inner metadata and spec).

So my preference is to just add a new ObjectMeta into this file with the fields we support instead of importing k8s APIs, but otherwise leave the PR as-is.

networking/v1alpha3/workload_group.proto
justinwei2: They're just ports associated with the workload group, rephrased to make clearer

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,115 @@
...
+  ObjectMeta metadata = 1;
+
+  // Describes the WorkloadEntry resources that will be created.
+  WorkloadEntry template = 2;
+}
+
+// From k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta.
+message ObjectMeta {
+    string name = 1 [(google.api.field_behavior) = REQUIRED];
+    string namespace = 2 [(google.api.field_behavior) = REQUIRED];
```
rshriram, line 112: can we please kill this? It seems strange to have two names, where the name of workload group has no bearing on the name of the template. Secondly, there is nothing preventing the user from using a different namespace in the object metadata (which then will subvert all RBAC permissions!). In 99% of situations, this is going to be identical, contributing to more bloat. As I mentioned earlier, we simply need something like this

```
message WorkloadGroup {
 // WorkloadEntry specification. Please note that `address` and `labels` fields SHOULD NOT be set in the 
// template. The `serviceAccount` field MUST be specified. The workload identities (mTLS certificates) will be 
// bootstrapped using the specified service account's token. Workload entries in this group will be in the same 
// namespace as the workload group, and inherit the labels, and annotations from the group object.
 WorkloadEntry template = 1 [(google.api.field_behavior) = REQUIRED];
};
```

such that the workload entries inherit the name+(some random string), namespace, labels, annotations from the workload group. The template's doc

rshriram: Okay fine. As long as the metadata has just labels/annotations, we should be fine. Now that said, doesn't the workloadEntry already have labels? optics wise it may look nice in examples, but when viewing the entire API, the user would see labels in 3 places (in the crd object, in the spec.metadata in workloadGroup and then finally in the workloadEntry spec as well). If we wanted labels in just two places where the second one is an override, then why not re-use the labels field in the workloadEntry spec (and introduce annotations into the workloadEntry spec) ? Granted we wont have the Deployment doppelgänger but atleast it would be clean/consistent in our APIs
So something like 

```
message WorkloadGroup {
 // WorkloadEntry specification. Please note that `address` and `labels` fields SHOULD NOT be set in the 
// template. The `serviceAccount` field MUST be specified. The workload identities (mTLS certificates) will be 
// bootstrapped using the specified service account's token. Workload entries in this group will be in the same 
// namespace as the workload group, and inherit the labels, and annotations from the group object.
 WorkloadEntry template = 1 [(google.api.field_behavior) = REQUIRED];
};
```

where an example would be

```
apiVersion: networking.istio.io/v1beta1
kind: WorkloadGroup
metadata:
  name: foo
  namespace: bar
  labels:
    scooby: doo
spec:
  template:
    serviceAccount: wg-sa
    labels:
      scrappy: doo
    annotations:
      peter: pan
    locality: ..
    ...
```

rshriram: i.e we add annotations field to workload entry, tweak the workloadGroup as described in my comment.

justinwei2: Should we move `metadata` and `template` from `WorkloadEntryTemplate` to `WorkloadGroup` to reduce clutter?

rshriram: What I meant is you dont need metadata becuase labels and annotations are available in worklaod entry (if not, add annotations to workload entry). Then all you need in the workloadGroup is just the template field that is an instance of WorkloadEntry

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,116 @@
...
+  // specified service account's token. Workload entries in this group will be in the same namespace as the
+  // workload group, and inherit the labels, and annotations from the group object.
+  WorkloadEntry template = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+// From k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta.
+message ObjectMeta {
+    map<string, string> labels = 1;
+    map<string, string> annotations = 2;
+}
```
rshriram, line 116: new line.. and please document the fields properly. No pointers to other projects for message level docs :).

networking/v1alpha3/workload_group.proto
rshriram: @justinwei2 if we are going to retain metadata, we might as well move the fields from the entry template into workload group directly.

rshriram:
```
message WorkloadGroup {
  ObjectMeta metadata = 1;

  // Describes the WorkloadEntry resources that will be created.
  // Please note that `address` and `labels` fields should not be set in the template. The `serviceAccount`
  // field MUST be specified. The workload identities (mTLS certificates) will be bootstrapped using the 
  // specified service account's token. Workload entries in this group will be in the same namespace as the
  // workload group, and inherit the labels, and annotations from the group object.
  WorkloadEntry template = 2 [(google.api.field_behavior) = REQUIRED];
}
```

networking/v1alpha3/workload_group.proto
rshriram: ignore comment. I agree with your suggestion

networking/v1alpha3/workload_group.proto
smawson: Just chatted with Shriram and I'm good with this change, so rather than:

```
WorkloadGroup {
  WorkloadTemplate template {
    ObjectMeta meta
    WorkloadEntry spec
  }
}
```
we should instead have:
```
WorkloadGroup {
  ObjectMeta meta
  WorkloadEntry template
}
```

networking/v1alpha3/workload_group.proto
smawson: Yes there can be multiple, this is used for groups of VMs (for example autoscaling groups).

justinwei2:
@smawson @rshriram please look at my latest commit, I think it addresses most of the concerns here about condensing the structure and commentary

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,111 @@
...
+// +k8s:deepcopy-gen=true
+// -->
+message WorkloadGroup {
+  // Metadata that will be used for all corresponding `WorkloadEntries`.
+  // The labels and annotations in `metadata` can differ from those in `template`.
+  ObjectMeta metadata = 1;
+
+  // Template to be used for the generation of `WorkloadEntry` resources that belong to this `WorkloadGroup`.
+  // Please note that `address` and `labels` fields should not be set in the template. The `serviceAccount`
+  // field MUST be specified. The workload identities (mTLS certificates) will be bootstrapped using the
```
smawson, line 98: I thought we decided that serviceAccount could default to "default" so wouldn't be required?

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,111 @@
...
+message WorkloadGroup {
+  // Metadata that will be used for all corresponding `WorkloadEntries`.
+  // The labels and annotations in `metadata` can differ from those in `template`.
+  ObjectMeta metadata = 1;
+
+  // Template to be used for the generation of `WorkloadEntry` resources that belong to this `WorkloadGroup`.
+  // Please note that `address` and `labels` fields should not be set in the template. The `serviceAccount`
+  // field MUST be specified. The workload identities (mTLS certificates) will be bootstrapped using the 
+  // specified service account's token. Workload entries in this group will be in the same namespace as the
+  // workload group, and inherit the labels, and annotations from the group object.
```
smawson, line 100: They don't inherit labels or annotations from the group object, they inherit them from the above metadata field.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,111 @@
...
+//
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1alpha3
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message WorkloadGroup {
+  // Metadata that will be used for all corresponding `WorkloadEntries`.
+  // The labels and annotations in `metadata` can differ from those in `template`.
```
smawson, line 93: Template doesn't have annotations, and we are trying to have users not set labels in the temlate but instead set them here, so reword this.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,111 @@
...
+
+  // Template to be used for the generation of `WorkloadEntry` resources that belong to this `WorkloadGroup`.
+  // Please note that `address` and `labels` fields should not be set in the template. The `serviceAccount`
+  // field MUST be specified. The workload identities (mTLS certificates) will be bootstrapped using the 
+  // specified service account's token. Workload entries in this group will be in the same namespace as the
+  // workload group, and inherit the labels, and annotations from the group object.
+  WorkloadEntry template = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+// `ObjectMeta` describes metadata associated with a `WorkloadGroup`.
```
smawson, line 104: No, this is not metadata for the WorkloadGroup, this is metadata that will be attached to the WorkloadEntry.

The comment above where this is referenced is right.

Maybe drop the workload group specific wording here and just describe this as the subset of k8s metadata that is supported.

networking/v1alpha3/workload_group.proto
justinwei2: I understood this as the field shouldn't be empty - e.g. you'd have to specify `default` here rather than having istio/istio infer it as `default`. Could be rephrased to make clearer though

smawson: Why would we force them to write `default` everywhere? I'd rather make this field optional, and if unspecified it uses `default`

Is there a reason to make this required?

douglas-reid: I have a similar recollection as smawson. Having it populate as `default` automatically would be useful.

networking/v1alpha3/workload_group.proto
```diff
@@ -0,0 +1,112 @@
...
+  // Please note that `address` and `labels` fields should not be set in the template, and an empty `serviceAccount`
+  // should default to `default`. The workload identities (mTLS certificates) will be bootstrapped using the
+  // specified service account's token. Workload entries in this group will be in the same namespace as the
+  // workload group, and inherit the labels and annotations from the above `metadata` field.
+  WorkloadEntry template = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+// `ObjectMeta` describes metadata that will be attached to a `WorkloadEntry`.
+// It is a subset of the supported Kubernetes metadata.
+message ObjectMeta {
```
rshriram, line 106: nit: if you are going to make another change, please move this into the WorkloadGroup message for now.

rshriram:
Before you remove the hold, please copy the API to networking/v1beta1 as well. Both have to be identical. And since you have to make the change, can you please move the ObjectMeta message definition into the WorkloadGroup message, so that its localized?

smawson:
> Before you remove the hold, please copy the API to networking/v1beta1 as well. Both have to be identical. And since you have to make the change, can you please move the ObjectMeta message definition into the WorkloadGroup message, so that its localized?

That sounds broken, are you saying we can't actually launch alpha APIs anymore? We want this in alpha but not beta since it is alpha state...

douglas-reid:
Removed `hold` label per offline conversation.

PR #1559, opened 2020-07-27
Remove deprecated sds_uds_path proto

nschhina:
With: https://github.com/istio/istio/pull/25834 this deprecated field is no longer needed.


nschhina:
> But why not sds_refresh_delay 

good point, I didn't look into this field, but looks like it's not used anywhere in istio/istio anymore either, so I'll clean this up too

nschhina:
> But why not sds_refresh_delay ?
> Also: enable_sds_token_mount and sds_use_k8s_sa_jwt - most of this is now auto-detected,
> should not be exposed to user.

 Thanks! Good opportunity to clean up all of these.

nschhina:
> Also: enable_sds_token_mount 
This requires some cleanup in istio/istio, I'll open a quick PR for this too

PR #1563, opened 2020-07-29
tracing: add API for OpenCensus agent exporter

brianwolfe:
This is based heavily on the existing Stackdriver configuration.
Stackdriver and OpenCensusAgent exporter will both use the OpenCensus tracer
implementation. This will permit OpenCensus to export OpenCensus agent
formatted spans.

OpenCensus agent-formatted spans are handled by the OpenTelemetry collector,
providing a migration path from OpenCensus to OpenTelemetry when
OpenTelemetry is more complete and integrated with Envoy.

@howardjohn 

Change-Id: I566299c9291021d26ec1e839643871a380807bc3


brianwolfe:
Static to dynamic is similar. OpenCensus & Stackdriver using OpenCensus tracers have to have a fixed configuration for the entire proxy process because OpenCensus uses static variables for state and so can only handle a single version of the configuration, see: https://github.com/envoyproxy/envoy/pull/10519 for some context.

mesh/v1alpha1/proxy.proto
```diff
@@ -81,6 +81,29 @@ message Tracing {
     google.protobuf.Int64Value max_number_of_message_events = 4;
   }
 
+  // OpenCensus defines configuration for an OpenCensus tracer.
+  // See [OpenCensus trace config](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace_config.proto) for details.
+  message OpenCensus {
```
douglas-reid, line 86: Should we expand this to make it possible to support Stackdriver via this configuration (and deprecate the Stackdriver config)?

@mandarjog @bianpengyuan

bianpengyuan: Yes, make sense to me to replace Stackdriver configuration with this. We can add a backend enum in this Opencensus message.

mandarjog: We should support all oc tracer backends including including zipkin thru this.
Please add enum.

brianwolfe: I didn't think we would want to expose the tracing implementation to the user. The OpenCensus & Stackdriver are really the external interface that a user would deal with, they wouldn't need to necessarily know that it is all OpenCensus under the hood.

I can make a try at merging them if you disagree.

brianwolfe: sg. I'll update.

mandarjog: Actually please wait.
Let’s discuss backend vs implementation.

So far the names are all backends or formats. 
OC is confusing because it is an impl and also a format.
I think it does make sense to stick with backend names.
We just need a different name than opencensus.

douglas-reid: If we want to stick with backends, then perhaps this should be `OpenCensusAgent` ?

brianwolfe: The only names I could find for this format were:

* OpenCensus Agent protocol
* OpenCensus Service

tbh, neither of them seemed to add clarity.

One advantage of specifying the top-level OpenCensus format is that we can consolidate the logic for specifying which headers we read/write and allow a user to only specify those headers they actually want in their infrastructure.

To be more specific, I'd like to be able to only specify TRACE_CONTEXT to avoid extra bytes on the wire here: https://github.com/istio/istio/blob/master/tools/packaging/common/envoy_bootstrap.json#L627-L628 If we just make an OpenCensus top-level configuration with an enum for the backend, we can easily do that. The other trace implementations won't have the same header options.

Another advantage is that we can differentiate between Zipkin backend implementations and so migrate people to the OpenCensus implementation with some more flexible options for trace propagation headers.

douglas-reid: Should we have a short design doc on this change ?

brianwolfe: I can make one. How do I get write access to the drive though?

brianwolfe: I made a quick draft PTAL: https://docs.google.com/document/d/1rbVLU0G391eKincaAJQhpBYAdhcvaF2G1sqGkJgKLcQ/edit#heading=h.xw1gqgyqs5b

I'm still waiting on approval to actually have write access to the drive, I had a coworker copy it into the drive.

brianwolfe: @douglas-reid @bianpengyuan

brianwolfe:
@mandarjog @bianpengyuan, thanks for the feedback, I updated to match with https://docs.google.com/document/d/1rbVLU0G391eKincaAJQhpBYAdhcvaF2G1sqGkJgKLcQ/edit#

PTAL

mesh/v1alpha1/proxy.proto
```diff
@@ -81,6 +81,63 @@ message Tracing {
     google.protobuf.Int64Value max_number_of_message_events = 4;
   }
 
+  // OpenCensusAgent defines configuration for an OpenCensus tracer writing to
+  // an OpenCensus agent backend. See
+  // [OpenCensus trace config](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace_config.proto)
```
bianpengyuan, line 86: We probably should link to envoy's doc: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/opencensus.proto, which is more aligned with the proto defined here. Or link to both. we should update the comment of Stackdriver tracer to link to that as well.

mesh/v1alpha1/proxy.proto
```diff
@@ -81,6 +81,63 @@ message Tracing {
...
+  // for details.
+  message OpenCensusAgent {
+    // TraceContext selects the context propagation headers used for
+    // distributed tracing.
+    enum TraceContext {
+      // $hide_from_docs
+      // Unspecified context. Should not be used for now, but added to reserve
+      // the 0 enum value if TraceContext is used outside of a repeated field.
+      UNSPECIFIED = 0;
```
bianpengyuan, line 95: Why not make it NONE which follows envoy trace context: https://github.com/envoyproxy/envoy/blob/598e76a51b781bdd435d5d4f96d482fc283695b7/api/envoy/config/trace/v3/opencensus.proto#L30. I don't think it is necessary to hide this from docs, since it is not just a reserved field, but also a valid configuration for Envoy.

mesh/v1alpha1/proxy.proto
```diff
@@ -81,6 +81,63 @@ message Tracing {
...
+
+    // gRPC address for the OpenCensus agent (e.g. dns://authority/host:port or
+    // unix:path). See [gRPC naming
+    // docs](https://github.com/grpc/grpc/blob/master/doc/naming.md) for
+    // details.
+    string address = 1;
+
+    // Specifies the set of context propagation headers used for distributed
+    // tracing. Default is `["W3C_TRACE_CONTEXT"]`. If multiple values are specified,
+    // the proxy will attempt to read each header for each request and will
```
bianpengyuan, line 120: Do you want to income and outgoing context separated?

mesh/v1alpha1/proxy.proto
brianwolfe: I'm ok permitting NONE, but that seems like a detail that would only add confusion for users.

mesh/v1alpha1/proxy.proto
brianwolfe: I don't think it particularly makes sense to do so. I haven't seen a use-case where people really want to specify these separately for Istio. For Istio it adds to the confusion for users since it is incoming & outgoing w.r.t. envoy (really contexts-to-read and contexts-to-write).

I think this should not be separated unless we have a compelling use-case.

brianwolfe: maybe after we have per-listener configs it will make more sense to separate them out.

mesh/v1alpha1/proxy.proto
brianwolfe: I chose UNSPECIFIED since that is the closest I could find in the Google style guide for protobufs. https://developers.google.com/protocol-buffers/docs/style#enums and it seemed to be consistent within Istio. I assumed Istio would roughly follow the google style guide for protobuf instead of mix-and-match with envoy & google styles.

mesh/v1alpha1/proxy.proto
brianwolfe: ack. Will update after commit is reviewed internally.

mesh/v1alpha1/proxy.proto
```diff
@@ -81,6 +81,63 @@ message Tracing {
...
   }
 
+  // OpenCensusAgent defines configuration for an OpenCensus tracer writing to
+  // an OpenCensus agent backend. See
+  // [OpenCensus trace config](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace_config.proto)
+  // for details.
+  message OpenCensusAgent {
+    // TraceContext selects the context propagation headers used for
+    // distributed tracing.
+    enum TraceContext {
```
mandarjog, line 91: @douglas-reid should we hoist Tracecontext msg out of the OC proto? 
We can use it with SD as well, and is general enough.

brianwolfe: @mandarjog my main concern with hoisting it out is that it would also make sense for Lightstep implementation, but the set of TraceContexts supported are different, so we will end up needing to prefix them.

I'm ok with either option, just seems cleaner to keep the enums inside for each trace exporter.

douglas-reid: @mandarjog that sort of overlap was behind my original suggestion to deprecate stackdriver (and others) in favor of OpenCensus.

I think we need to decide what the scope of this change is.

I think adding `OpenCensusAgent` is a tactical change to support a needed use case -- so maybe we can avoid expanding TraceContext for now.

As you know, I feel strongly we should push on a Telemetry API that re-organizes some of this configuration.

brianwolfe: The main purpose I have for putting this change up is to get our Istio configurations to be more standard (right now we have weird configuration hacks to use the OpenCensus & custom metrics in a bootstrap configuration that make it hard to explain). When we have this change in and we upgrade to 1.7, I will be in a better place to meaningfully contribute to @douglas-reid's proposal since our forward migration will look like everyone else's.

brianwolfe: If you want to delay any push on this until the next telemetry API is out, I understand. I think we can technically continue running with our hacked-up bootstrap configuration for the next few versions.

douglas-reid: @brianwolfe i think this PR makes sense as-is, especially since it will remove hacked-up bootstraps in the short term.

mesh/v1alpha1/proxy.proto
brianwolfe: Updated.

mesh/v1alpha1/proxy.proto
brianwolfe: Thanks, I'll wait for next round of feedback from bianpengyuan, I'd like to be on the same page w.r.t. copying envoy TRACE_CONTEXT semantics.

brianwolfe:
@googlebot I signed it!

mandarjog:
@nrjpoddar can you approve the api change?

brianwolfe:
@nrjpoddar ping

nrjpoddar:
@brianwolfe currently Istio documentation only talks about propagating X-B3-* headers. Are you planning to add docs for how users can propagate other tracing headers if non Zipkin style headers are used? I’m not familiar with all the supported tracers in Envoy which looks like a rather long list.

brianwolfe:
@nrjpoddar Yeah, I can definitely add a longer section in here: https://github.com/istio/istio.io/blob/e9a89c879f6925593e2147b7778baeff85ac9bba/content/en/docs/tasks/observability/distributed-tracing/overview/index.md and https://github.com/istio/istio.io/blob/eff0d43a27d485e3997b38109768e1650cad70cf/content/en/faq/distributed-tracing/how-to-support-tracing.md to better document the options around context propagation for distributed tracing. I'll need help with the Chinese translation though.

brianwolfe:
@nrjpoddar what order should I update in? API, then main codebase, then site docs?

nrjpoddar:
> @nrjpoddar what order should I update in? API, then main codebase, then site docs?

If this is going in in release-1.7 branch then we need to make sure we get all 3 changes merged soon i.e. before end of next week.

If this is targeted for 1.8, then this is my preferred order:

1. API
2. Implementation
3. Docs

brianwolfe:
I don't think it makes sense to target release-1.7. I'm ok delaying this & related work until 1.8. Rushing it seems likely to mean we'll end up with something more confusing. We can backport the work internally for our use-case or bring it into a point release for 1.7 if people really want the feature.

mesh/v1alpha1/proxy.proto
```diff
@@ -81,6 +83,65 @@ message Tracing {
...
+
+    // Specifies the set of context propagation headers used for distributed
+    // tracing. Default is `["W3C_TRACE_CONTEXT"]`. If multiple values are specified,
+    // the proxy will attempt to read each header for each request and will
+    // write all headers.
+    repeated TraceContext context = 2;
+
+    // debug enables trace output to stdout
+    // $hide_from_docs
+    bool debug = 3;
```
nrjpoddar, line 130: Why are these hidden from docs but not the rest?

nrjpoddar:
LGTM as it's for 1.8 and we can refine it later. Please answer my question about hidden fields from docs before merging.

mesh/v1alpha1/proxy.proto
brianwolfe: We mostly use these for checking maximum performance impact & ensuring the tracer gets instantiated correctly. They are fairly detailed configuration options that I'd expect mostly istio developers who care about performance to examine. 

This appears to be the same reasoning as the existing parameters in Stackdriver that are also hidden from the docs.

wdyt?

nrjpoddar: We shouldn't even have an API for these options and see if ENV vars or other mechanism can be used for Istio devs. Having a hidden option in user facing API which is never going to be used to end-users seems odd to me.

brianwolfe: I see. That is fine for me. I can remove them from this tracer implementation. Do you see any reason to keep them for Stackdriver?

brianwolfe: Didn't realize this would automerge when I ran the test, sorry about that.

nrjpoddar: I'm not a stack driver expert so I would leave it up to @mandarjog or @douglas-reid to answer that question.

PR #1566, opened 2020-08-01
Extend VS.HTTPRoute with RegexRewrite

abhide:
Tickets: GH-22290


networking/v1beta1/virtual_service.proto
```diff
@@ -594,10 +594,14 @@ message HTTPRoute {
...
 
-  // Rewrite HTTP URIs and Authority headers. Rewrite cannot be used with
-  // Redirect primitive. Rewrite will be performed before forwarding.
-  HTTPRewrite rewrite = 4;
-
+  // Only one of rewrite or regex_rewrite can be specified
+  oneof rewrites {
+    // Rewrite HTTP URIs and Authority headers. Rewrite cannot be used with
+    // Redirect primitive. Rewrite will be performed before forwarding.
+    HTTPRewrite rewrite = 4;
```
howardjohn, line 601:
```
rewrite:
  authority:
  uri:
regexRewrite:
  regex:
```

Seems like these are the possible options, which raises the question to me - what is being rewritten by `regex`, the uri or the authority? And can I do a authority prefix rewrite along with a regex rewrite of uri?

abhide: Only the URI part.

https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route_components.proto#envoy-api-field-route-routeaction-regex-rewrite

howardjohn: I don't think this is clear to users or consistent with the existing field. we should either change `regex` to uri or nest it under rewrite maybe?

nrjpoddar:
@abhide can you add examples in the doc on how to use regex rewrite?

abhide:
> @abhide can you add examples in the doc on how to use regex rewrite? This is a great functionality to add but will be difficult for users to follow without concrete examples.

@nrjpoddar Do you want me to add envoy regex examples? Essentially, we would be passing the values from VS to envoy config, correct?

networking/v1beta1/virtual_service.proto
abhide: I see.. So I can nest `rewriteRegex` under `rewrite`. Will do that change.

abhide: @howardjohn given that we can either have rewrite or regexRewrite have it nested under `rewrite` doesn't make sense.

howardjohn: Sorry, I am not sure why it doesn't make sense?
```
rewrite:
  authority:
  uri:
regexRewrite:
  regex:
```

is clearly inconsistent to me.

 ---

```
rewrite:
  authority:
  uri:
    regix: ...
    prefix: ...
```

is ideal. However, it is not backwards compatible.

 ---

Therefor it seems that:

```
rewrite:
  authority:
  uri: ...prefix...
  uriRegix: ...regex..
```

is the most consistent

howardjohn:
> Do you want me to add envoy regex examples? Essentially, we would be passing the values from VS to envoy config, correct?

Yes please. An example of a simple regex (we don't need a complex one, we want to show how it can be used not how regex works), and the input URL and what it will be rewritten to.

networking/v1beta1/virtual_service.proto
abhide: @howardjohn yes.. makes sense. fixed in latest changes.

abhide:
@howardjohn @nrjpoddar Added a WIP patch that uses these API changes: https://github.com/istio/istio/pull/28098

Looking for feedback/comments so that we can work through naming params and documentation.

networking/v1alpha3/virtual_service.proto
```diff
@@ -1470,6 +1470,16 @@ message HTTPRewrite {
...
   // rewrite the Authority/Host header with this value.
   string authority = 2;
+
+  // uri_regex can be used for rewriting portions of path that match the
+  // pattern during forwarding the request
+  // RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
+  UriRegexRewrite uri_regex = 3;
+}
+
+message UriRegexRewrite {
```
abhide, line 1480: Using same parameter names as envoy: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route_components.proto#envoy-api-field-route-routeaction-regex-rewrite

howardjohn (approved):

ramaraochavali (approved):

networking/v1alpha3/virtual_service.proto
```diff
@@ -1470,6 +1470,16 @@ message HTTPRewrite {
 
   // rewrite the Authority/Host header with this value.
   string authority = 2;
+
+  // uri_regex can be used for rewriting portions of path that match the
+  // pattern during forwarding the request
```
smawson, line 1475:
```suggestion
  // pattern when forwarding the request.
```

networking/v1beta1/virtual_service.proto
```diff
@@ -1464,6 +1464,16 @@ message HTTPRewrite {
 
   // rewrite the Authority/Host header with this value.
   string authority = 2;
+
+  // uri_regex can be used for rewriting portions of path that match the
+  // pattern during forwarding the request
```
smawson, line 1469:
```suggestion
  // pattern when forwarding the request.
```

guillevalin:
Hi istio team,

I've been following this issue and PR since september past year, we want to migrate to istio but we need regex rewrite rules to replace nginx-ingress.

Is there any chance this PR could be merged soon?

Thank you in advance.

costinm:
I am a bit concerned with this - can you add a discussion on networking WG ? 
AFAIK it is possible to use an EnvoyFilter to achieve this result - so advanced users who need this are not blocked. 

I don't think Istio API must mirror all the features in envoy - right now the consensus in the WG is to adopt the upstream
K8S networking APIs, which in turn allow multiple implementation besides Istio.  They do allow some 'extension points'
as well - but any extension point will be vendor-specific. 

I don't think RE2 based rewrite a URL is a common enough feature that would be supported cross-vendor, and it puts
unreasonable requirements on non-envoy based implementations of Istio ( proxyless ).

costinm:
At least we should bounce the idea with the K8S Services WG - and the other projects planning to support the K8S APIs. 
I assume since this didn't make it to 1.9 we can afford few weeks to discuss with the relevant WGs.

abhide:
I am not actively working on this GH issue. Will un-assign myself so that someone can pick it up.

SpecialYang:
I really need this awesome feature and want to support it. Can I pick it up?

PR #1567, opened 2020-08-01
Fix quantity and lock down others to a more specific type

liamawhite:
As demonstrated by https://github.com/istio/istio/pull/25538 the `Quantity` type is currently (has always been?) broken in the install API. This PR changes them all to `IntOrString` and adds code move the type inline with the actual YAML spec.

The correct fix for this would be to deep dive into the `gogo` code and work out why this is happening but given that we are moving away from gogo soonish and I don't have the time right now this alternate fix will have to do.

It also changes a couple of the other types to be their correct type as specified in the Kubernetes API (`targetAverageUtilization`) and changes some of the `TypeInterface` that should actually be `IntOrString` (e.g. ports).

Signed-off-by: Liam White <liam@tetrate.io>


howardjohn:
1.7 cherrypick seems reasonable given this is a bug fix and low risk

PR #1570, opened 2020-08-03
define EnvoyFilter REPLACE operation

mandarjog:
Use cases
1. merge semantics are not adequate, since deletions cannot be performed.
2. A request-classification filter is inserted in the correct spot by the config root namespace with `use_if_overridden `. 
    Every workload deploys its own request-classification filters based on open-api-spec, but does not need to worry about filter ordering.

    Not every canonical service will have a request classification filter, so the place holder should be elided in the XDS output for that workload.


networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,6 +533,18 @@ message EnvoyFilter {
...
+      // Replace should be used when you do not need to or cannot use
+      // json merge semantics.
+      // A typical use case for REPLACE is when an EnvoyFilter in the
+      // config root namespace specifies the order of filters but you want
+      // a completely different configuration for a specific filter.
+      // Since JSON merge cannot remove elements, REPLACE operation should
+      // be used. REPLACE operation is only valid for HTTP_FILTER and
+      // NETWORK_FILTER. If the named filter is not found, this operation
+      // has no effect.
+      REPLACE = 7;
```
howardjohn, line 547: is replace equivalent to REMOVE+ADD?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -541,6 +553,15 @@ message EnvoyFilter {
...
     google.protobuf.Struct value = 2;
+
+    // A `should_override` patch is used to provide filter ordering with
+    // optional default values. A later MERGE or REPLACE operation
+    // should override the contents of the configuration or the
+    // operation is a NOOP. This allows for an empty or a partial
+    // configuration to be specified in the config root namespace which
+    // is elided from the final output unless overridden by a
+    // subsequent operation.
+    bool should_override = 3;
```
howardjohn, line 564: This seems really confusing to me in terms of how this is meant to be used

networking/v1alpha3/envoy_filter.proto
mandarjog: It is not.
ADD requires you take care of ordering. REPLACE does not.

networking/v1alpha3/envoy_filter.proto
mandarjog: I have added details in the PR description.

networking/v1alpha3/envoy_filter.proto
rshriram: why is it valid only for filters and not for the rest?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,6 +533,18 @@ message EnvoyFilter {
       // This is specifically useful when you want your filter first in the
       // list based on a match condition specified in Match clause.
       INSERT_FIRST = 6;
+
+      // Replace contents of a named filter object with new contents.
+      // Replace should be used when you do not need to or cannot use
+      // json merge semantics.
+      // A typical use case for REPLACE is when an EnvoyFilter in the
+      // config root namespace specifies the order of filters but you want
+      // a completely different configuration for a specific filter.
```
rshriram, line 542: This line does not make sense.. Why does a completely different config for a filter have a bearing on the ordering of filters? It would help to clarify this properly. In fact, I would say add an example of replace operation

networking/v1alpha3/envoy_filter.proto
rshriram: Why do we need this? Why can't we detect that at runtime? In fact merge operates on content created by a previous envoy filter. 

I think this is simply describing the lowest level operation of what this flag does but doesn't tell the user how to actually use it. If viewed that way, this flag may not be the ultimate thing we express in the API to accomplish the intent. Can you describe a use case and I can try to see if there is a better way of describing the function?

networking/v1alpha3/envoy_filter.proto
mandarjog: Keep the API as narrow as possible, If we have more use cases, we can add it later.

networking/v1alpha3/envoy_filter.proto
mandarjog: Use case:

A request-classification filter is inserted in the correct spot by the config root namespace with `should_override=true`.
Every workload/service deploys its own request-classification filters based on open-api-spec, but does not need to worry about filter ordering.

Not every workload will have a request classification filter / openapi-spec, so the place holder should be elided in the XDS output for that workload.

mandarjog:
@rshriram 
The following is in the config root namespace
```yaml
  - applyTo: HTTP_FILTER
    match:
      context: SIDECAR_INBOUND
      listener:
        filterChain:
          filter:
            name: envoy.http_connection_manager
            subFilter:
              name: envoy.router
      proxy:
        proxyVersion: ^1\.8.*
    patch:
      operation: INSERT_BEFORE
      value:
        name: istio.stats
        typed_config:
          '@type': type.googleapis.com/udpa.type.v1.TypedStruct
          type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
          value:
            config:
              configuration:
                '@type': type.googleapis.com/google.protobuf.StringValue
                value: "{}"
              root_id: stats_inbound
              vm_config:
                code:
                  local:
                    filename: /etc/istio/extensions/stats-filter.wasm
                runtime: envoy.wasm.runtime.v8
                vm_id: stats_inbound
```
and this is in the workload namespace
```yaml
  - applyTo: HTTP_FILTER
    match:
      context: SIDECAR_INBOUND
      listener:
        filterChain:
          filter:
            name: envoy.http_connection_manager
            subFilter:
              name: envoy.router
      proxy:
        proxyVersion: ^1\.8.*
    patch:
      operation: REPLACE
      value:
        name: istio.stats
        typed_config:
          '@type': type.googleapis.com/udpa.type.v1.TypedStruct
          type_url: type.googleapis.com/envoy.extensions.filters.http.wasm.v3.Wasm
          value:
            config:
              configuration:
                '@type': type.googleapis.com/google.protobuf.StringValue
                value: "{}"
              root_id: stats_inbound
              vm_config:
                code:
                  local:
                    inline_string: envoy.wasm.stats
                runtime: envoy.wasm.runtime.null
                vm_id: stats_inbound
```

note `vm_config.code.local` can either have a `inline_string` or `filename` since it is a `oneof`.
Using a replace operation I can just specify the name.

networking/v1alpha3/envoy_filter.proto
kyessenov: This is indeed very confusing. Maybe make it malformed by default and allow elision if it's malformed?

kyessenov: IMHO if you could drop proto, you could just use "null" as the placeholder value and implement PATCH directly. I think that's how GCP APIs implement PATCH. The problem is that we're constrained by the proto syntax.

mandarjog: not sure I understand fully. `should_override` is like an abstract class, if you instantiate it, you must provide an impl for the pure virtual fn. Are you suggesting we use pull GCP PATCH here?

kyessenov: What you are trying to do makes sense, but it doesn't come across. I think it's because you are trying to express a meta-patch (patch over patches). My suggestion re: "null" is to allow null as value in YAMLs. That won't be parsable by standard protobuf tooling, but it'll allow removing fields.

networking/v1alpha3/envoy_filter.proto
kyessenov: Speaking of which: what restricts a second patch from altering the order (e.g. specific filter EnvoyFilter changing root namespace filter order). There's no visibility restriction on what can be changed with EnvoyFilters.

nrjpoddar:
Does this solve the merge semantics in array fields where currently you cannot revolve repack entire contents of an array?

networking/v1alpha3/envoy_filter.proto
mandarjog: It is possible for a patch to change order. If you are changing order your understand order and know how to change it. 
Using “should_override” you can opt out of knowing about order and dependencies and just change the inner config of a filter.

mandarjog:
@nrjpoddar You will replace contents of a filter configuration with the new one provided with the “replace” directive, so yes it sidesteps / resolves the issue.

networking/v1alpha3/envoy_filter.proto
ramaraochavali: fwiw, One of the use cases we needed replace is for virtual host domains - for a given flow and under some conditions we had to allow wildcard domain, we used `MERGE` to add  '*' that means it has other domains as well as wildcard '*' - in this case other domains are redundant and we had to take care of the case where the generated domain list already has a wildcard (otherwise Envoy complains with an error). `REPLACE` would have been easier in that case. 

May be we should change the doc to indicate currently it is implemented for filters only so that later it can expanded to others as needed?

networking/v1alpha3/envoy_filter.proto
ramaraochavali: Would it be less confusing if we tie this back to operation like `MERGE_IF_OVERRIDDEN`  or some thing like that? Then user is forced to specify should override otherwise it is confusing with two distinct fields?

networking/v1alpha3/envoy_filter.proto
hzxuzhonghu: `Replace` is a good ops, should be generic, not only limited to filters.

xref http://jsonpatch.com/

mandarjog: I am trying to keep the api narrowly defined. 
The api as defined allows you to replace the full contents of the filter. This is easy to understand and trivial to implement. 

a generic replace will have us use or implement jsonpatch which is more complex and slow, just like merge. I really think we should not make it more complex.

networking/v1alpha3/envoy_filter.proto
mandarjog: How about USE_IF_Overridden ?

ramaraochavali: I am fine with that as well

ramaraochavali: Sorry. I misunderstood - are you referring to changing "should_override" to "use_if_overridden" ? or adding a new operation name?

mandarjog: Changing the field.

mandarjog:
@rshriram @kyessenov @ramaraochavali I have changed names per discussion and added an example in the doc. PTAL.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,6 +593,18 @@ message EnvoyFilter {
...
+      // Since JSON merge cannot remove elements, REPLACE operation should
```
kyessenov, line 603: nit: proto merge. JSON merge can remove elements.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -203,6 +203,66 @@ import "networking/v1alpha3/sidecar.proto";
...
+//           filter:
+//             name: "envoy.http_connection_manager" # can be elided
+//             subFilter:
+//               name: "istio.openapi" # replacing previous placeholder by name.
+//     patch:
+//       operation: REPLACE
+//       value:
+//         name: istio.openapi
+//         # Full configuration must be specified here.
+// ```
```
rshriram, line 264: can you please describe the intent here? you have described a how, but i am struggling to understand the why/what of this example. Since we are in the istio expert land here wrt envoy filters, i would also like to get a feel of how others familiar with xds/istiod comprehend this example [may be i am being very dense].
@hzxuzhonghu / @ramaraochavali / @howardjohn - what do you understand from the example above?

howardjohn: I *think* the benefit is you only need to be "order aware" in the single root namespace config, then the overrides just replace it without bothering about ordering?

ramaraochavali: I see two benefits
- Order is guaranteed in the root namespace
- Only applied to namespaces that actually provide real config

ramaraochavali: @mandarjog since this is only used during replace, how about calling this `use_if_replaced` or `use_if_replace_exists` - `overridden` seems little OO and technical?

mandarjog: It applies to replace and merge equally and both are way to override.

mandarjog:
Closing in favor of #1588

PR #1571, opened 2020-08-04
Add ProxyOutboundListenerPortRange to MeshConfig

liuxu623:
https://github.com/istio/istio/issues/26101


liuxu623:
> Do you have evidence this improves performance?
Hi, I use ab test different port ranges, you can see https://github.com/istio/istio/issues/26101#issue-672478829.

lambdai:
I agree this make sense in the edge cases where either 
1. 600 outbound short reqs are established per second, or
2. client maintains more than 60K connections
or something in between.

However, this should not be the mesh config applying to all pod.
Instead this should be per pod config since port range accross 10 port
is carried by envoy listener config which repeating the same listener 
config 10 times.

liuxu623:
> I agree this make sense in the edge cases where either
> 
> 1. 600 outbound short reqs are established per second, or
> 2. client maintains more than 60K connections
>    or something in between.
> 
> However, this should not be the mesh config applying to all pod.
> Instead this should be per pod config since port range accross 10 port
> is carried by envoy listener config which repeating the same listener
> config 10 times.

How about add a pod annotation like `traffic.sidecar.istio.io/outboundRedirectPortRange` ?

PR #1574, opened 2020-08-04
VerifyCertificateAtClient for DestinationRule

nschhina:
Ref: https://github.com/istio/istio/issues/25652

Old versions: opt into fallback with env var
New versions: opt out of fallback with InsecureSkipVerify in DR API
Nothing set -> verify with system certs
caCerts set -> verify with caCerts
InsecureSkipVerify -> no verify
Global override is temporary for transition, not permanent api


networking/v1beta1/destination_rule.proto
```diff
@@ -900,6 +900,12 @@ message ClientTLSSettings {
   // using TLS. The value of this field determines how TLS is enforced.
   TLSmode mode = 1 [(google.api.field_behavior) = REQUIRED];
 
+  // If set to true, client will not verify server-side TLS certificates or SAN
+  // during TLS handshake when the TLS mode is set to SIMPLE or MUTUAL and
+  // CaCertificates or SubjectAltNames is empty. By default client will fallback
+  // to use system CA certificates to validate server identity.
```
nschhina, line 906: What's the proposed fallback for SubjectAltNames?  cc: @nrjpoddar @costinm

rshriram:
Btw, if you feel really strongly about this, please talk to the envoy community and ask them to change their API semantics. Ours is a mirror of their API. If we find that this practice is insecure, then the envoy community should be onboard with this as well. If you manage to update the envoy implementation to auto verify if CAcertificates is not set, then all newly updated Istio proxies will automatically start verifying CA certificates using system built in. It does not make sense to introduce redundant options in Istio configs for this purpose.

rshriram:
Alternatively, if you and the rest of the Google folks (@louiscryan / @smawson ) as well as the IBM folks (@linsun ) are willing to accept responsibility for this huge breaking change that could affect all istio installations, I am happy to take my hands off this.

networking/v1alpha3/destination_rule.proto
```diff
@@ -902,6 +902,12 @@ message ClientTLSSettings {
   // using TLS. The value of this field determines how TLS is enforced.
   TLSmode mode = 1 [(google.api.field_behavior) = REQUIRED];
 
+  // If set to true, client will not verify server-side TLS certificates or SAN
+  // during TLS handshake when the TLS mode is set to SIMPLE or MUTUAL and
+  // CaCertificates or SubjectAltNames is empty. By default client will fallback
+  // to use system CA certificates to validate server identity.
```
nrjpoddar, line 908:
```suggestion
  // If set to true, client-side proxy will not verify that the presented peer TLS certificate is
  // signed by a trusted CA or verify that the requested host is in the SAN field of the certificate
  // during TLS handshake.
  // By default client-side proxy will fallback to use system CA certificates to verify the certificate and validate the SAN if either of caCertifcates or SubjectAltNames fields are empty
```

networking/v1beta1/destination_rule.proto
nrjpoddar: I would by default verify that the requested host matches one of the SAN entries obeying wildcards.

ramaraochavali: are we defaulting to false? Wouldn't this break existing customers? Also we discussed having a global mesh config option to drive this behaviour (defaulted to current)?

nrjpoddar: We will/should have a mesh config flag to enable transitions

myidpt:
@rshriram Louis and Sven have already approved this approach.
The issue is that the existing behavior doesn't match the customer's expectation. When "ISTIO_MUTUAL or "SIMPLE_TLS" is indicated in the DR, the user is expecting a secure connection. We rated this close to a CVE (@fpesce ). For the breaking change concern, I think < 1% of use cases will break, in which the users 1) forgets the set the caCertificate field AND 2) the external service doesn't run on a WebPKI-based cert. Plus this break is a "good break" that indicates there's a security hole in the cluster. Also, we will give release notes about this change.

cc @lizan We could discuss with Envoy team about supporting auto-fallback to system root, but I doubt about it since their general use cases may be different and there's no way for Envoy to easily feature out where to load the system root ca file (it depends on the OS).

One more thing as Louis has pointed out, we don't check the the identity in the server's certificate by default. A certificate check should include matching the service name with the identity presented in certificate by default. That's another Istio behavior that doesn't match user expectation. We will need to figure out how to fix that as well.

networking/v1beta1/destination_rule.proto
nschhina: My understanding is that the transition env flag is for older istio releases only. Default behavior from 1.7 onwards is fallback unless InsecureSkipVerify: true with no global override. This is what was proposed during the networking wg meeting cc: @costinm @howardjohn

nschhina:
> @rshriram Louis and Sven have already approved this approach.
> The issue is that the existing behavior doesn't match the customer's expectation. When "ISTIO_MUTUAL or "SIMPLE_TLS" is indicated in the DR, the user is expecting a secure connection. We rated this close to a CVE (@fpesce ). For the breaking change concern, I think < 1% of use cases will break, in which the users 1) forgets the set the caCertificate field AND 2) the external service doesn't run on a WebPKI-based cert. Plus this break is a "good break" that indicates there's a security hole in the cluster. Also, we will give release notes about this change.
> 
> cc @lizan We could discuss with Envoy team about supporting auto-fallback to system root, but I doubt about it since their general use cases may be different and there's no way for Envoy to easily feature out where to load the system root ca file (it depends on the OS).
> 
> One more thing as Louis has pointed out, we don't check the the identity in the server's certificate by default. A certificate check should include matching the service name with the identity presented in certificate by default. That's another Istio behavior that doesn't match user expectation. We will need to figure out how to fix that as well.

I think like @nrjpoddar proposed by default we should just match the host specified in DestinationRule in certificate SAN

networking/v1beta1/destination_rule.proto
ramaraochavali: IMO, we can not ship 1.7 with a behaviour change.

nrjpoddar: @nschhina you need a mesh wide setting for 1.7, so users can upgrade and preserve old behavior. Without a mesh global setting users would have to go and change all their existing DRs.

howardjohn: +1 strongly agree. I think we should actually ship 1.7 with it off. We need to give users time to migrate. Changing behavior of an API cannot be taken lightly

nschhina: Okay sounds good. I will replace env var with meshconfig option in 1.7

nrjpoddar: You can keep ENV var but toggle it based on mesh config option via templating.

lizan:
> cc @lizan We could discuss with Envoy team about supporting auto-fallback to system root, but I doubt about it since their general use cases may be different and there's no way for Envoy to easily feature out where to load the system root ca file (it depends on the OS).

It depends on how complicate to determine where it is. Can you file an issue in envoy side?

networking/v1beta1/destination_rule.proto
```diff
@@ -900,6 +900,48 @@ message ClientTLSSettings {
...
+  // If both CaCertificates and SubjectAltNames fields are empty in DestinationRule, client-side proxy will
+  // verify CA signature with default System CA bundle and SAN matches the host in DestinationRule.
+  //
+  // If both CaCertificates and SubjectAltNames are non empty in DestinationRule, client-side proxy will verify
+  // CA signature with CaCertificates and SubjectAltNames with SAN in presented server certificate during TLS
+  // handshake.
+  //
+  // If CaCertificates is non empty and SubjectAltNames is empty in DestinationRule, client-side proxy will
+  // verify CA signature with CaCertificates but will verify the SAN in the presented server certificate
+  // matches the host in DestinationRule during TLS handshake.
```
rshriram, line 935: so there is no way to not do host based san verification, i.e. we are bringing the same feature/limitation as the java clients into istio universally? Are you absolutely sure that every certificate being used will have an exact SAN hostname match and that no one in the planet (hint: aws s3) will have wildcard SANs?

networking/v1alpha3/destination_rule.proto
```diff
@@ -902,6 +902,55 @@ message ClientTLSSettings {
...
+    // proxy will verify CA signature with CaCertificates but will verify the SAN in the presented
+    // server certificate matches the host in DestinationRule during TLS handshake.
+    //
+    // If CaCertificates is empty and SubjectAltNames is non empty in DestinationRule, client-side
+    // proxy will verify CA signature ith default System CA bundle but will verify SubjectAltNames
+    // matches SAN in the presented server certificate during TLS handshake.
+    DISABLED = 2;
+  }
+
+  // InsecureSkipVerify sets the mesh global default for peer certificate validation
```
howardjohn, line 947: this comment is on DR but seems like its meant for mesh config

networking/v1alpha3/destination_rule.proto
```diff
@@ -902,6 +902,49 @@ message ClientTLSSettings {
   // using TLS. The value of this field determines how TLS is enforced.
   TLSmode mode = 1 [(google.api.field_behavior) = REQUIRED];
 
+  // ValidateServerCertificate decide how the presented peer certificate during TLS handshake
```
smawson, line 905: Same comment here as on the other PR, this documentation is too long and convoluted and repetitive. You don't need to describe all possible combinations of ValidateServerCertificate, CaCertificates and SubjectAltNames since CaCertificates and SubjectAltNames are independent. Just describe when CaCertificates will be checked and when they won't, and same with SubjectAltNames.

networking/v1alpha3/destination_rule.proto
```diff
@@ -902,6 +902,49 @@ message ClientTLSSettings {
...
+  // certificate during TLS handshake.
+  //
+  // If CaCertificates is non empty and SubjectAltNames is empty in DestinationRule, client-side
+  // proxy will verify CA signature with CaCertificates but will verify the SAN in the presented
+  // server certificate matches the host in DestinationRule during TLS handshake.
+  //
+  // If CaCertificates is empty and SubjectAltNames is non empty in DestinationRule, client-side
+  // proxy will verify CA signature ith default System CA bundle but will verify SubjectAltNames
+  // matches SAN in the presented server certificate during TLS handshake.
+  google.protobuf.BoolValue validate_server_certificate = 8;
```
rshriram, line 946: option name mismatch

nrjpoddar:
@nschhina @howardjohn what’s the status of this PR? We want this resolved in 1.8 so I’m not sure if another PR has been merged and this PR is stale or this PR has just not moved forward.

PR #1585, opened 2020-08-07
Add VerifyCertificateAtClient to MeshConfig

nschhina:
Please review https://github.com/istio/api/pull/1574 in parallel


nschhina:
Will need a cherrypick for 1.6 but default value will be true

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,14 @@ message MeshConfig {
   reserved 53;
   reserved "termination_drain_duration";
 
+  // If DisableSystemCaFallbackDestinationRule is set to true and either CaCertificates
+  // or SubjectAltNames fields in DestinationRule are empty, client-side proxy will not
+  // verify presented peer TLS certificates during TLS handshake.
+  // Default value is false.
+  bool disable_system_ca_fallback_destinationrule = 54;
```
nrjpoddar, line 455: This should include both SAN and CA cert functionality, why not call it the same name you have in the DR API?

nschhina: Good call, I'll update this

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,14 @@ message MeshConfig {
   reserved 53;
   reserved "termination_drain_duration";
 
+  // If EnableInsecureSkipVerifyDestinationRule is set to true and either CaCertificates
+  // or SubjectAltNames fields in DestinationRule are empty, client-side proxy will not
+  // verify presented peer TLS certificates during TLS handshake.
+  // Default value is false.
```
nrjpoddar, line 454: There's a bit of truth table logic which needs to be explained well. Add to my table here what's the expected behavior:

InsecureFlag   CaCertificates   SubjectAlternateNames          What's Verified
true                   empty                empty                                       No peer attribute is verified
false                  <dir>                 xyz.com                                    CA signature and Host SAN in cert is xyz.com

Can you fill in the rest of table for me?

Suggestion for name `EnableInsecureSkipVerifyDestinationRule` -> `EnableInsecureClientSkipVerify`

nschhina: @nrjpoddar  I added all possible cases(I think?) please verify. For DestionationRule level, `InsecureSkipVerify` will override global behavior in same way.

nrjpoddar: This looks really good, please add this in docs so that it's very clear how these 3 config knobs interact with each other.

@linsun wdyt?

nschhina:
Please feel free to leave suggestions for the user facing comment that'll show up on the istio docs

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,50 @@ message MeshConfig {
   reserved 53;
   reserved "termination_drain_duration";
 
+  enum InsecureCertificateValidationAtClient {
+    UNKNOWN_CLIENT_VERIFY = 0;
+    ENABLED = 1;
+    DISABLED = 2;
+  }
+
+  // If InsecureSkipClientVerify is set to `ENABLED`:
```
nrjpoddar, line 457: // InsecureSkipClientVerify sets the mesh global default for peer certificate validation at the client-side proxy when TLS or mutual TLS (non ISTIO_MUTUAL) origination modes are used. This setting can be overridden at the host level via DestinationRule API.

// By default, InsecureSkipClientVerify is DISABLED.

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,50 @@ message MeshConfig {
   reserved 53;
   reserved "termination_drain_duration";
 
+  enum InsecureCertificateValidationAtClient {
+    UNKNOWN_CLIENT_VERIFY = 0;
```
nrjpoddar, line 452: // UNKNOWN_CLIENT_VERIFY field is not set
UNKNOWN_CLIENT_VERIFY

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,50 @@ message MeshConfig {
...
+    ENABLED = 1;
```
nrjpoddar, line 453: // ENABLED skips certificate and SAN validation at the client-side proxy unless overridden by DestinationRule

// For a particular host the following logic is used when DestinationRule are configured.
// If both CaCertificates and SubjectAltNames fields are empty in DestinationRule, client-side proxy will not verify either peer attribute during TLS handshake.
 //
 // If both CaCertificates and SubjectAltNames are non empty in DestinationRule, client-side proxy will verify
 // CA signature with CaCertificates and SubjectAltNames with SAN in presented server certificate during TLS
 // handshake.
 //
 // If CaCertificates is non empty and SubjectAltNames is empty in DestinationRule, client-side proxy will
 // verify CA signature with CaCertificates but will not verify any SAN in the presented server certificate
 // during TLS handshake.
  //
 // If CaCertificates is empty and SubjectAltNames is non empty in DestinationRule, client-side proxy will
 // not verify CA signature but will verify SubjectAltNames matches SAN in the presented server certificate
 // during TLS handshake.

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,50 @@ message MeshConfig {
...
+    DISABLED = 2;
```
nrjpoddar, line 454: // DISABLED implies client-side proxy will verify CA signature with default System CA bundle and ensure SAN matches the host in DestinationRule.

// For a particular host the following logic is used when DestinationRule are configured.

// If both CaCertificates and SubjectAltNames fields are empty in DestinationRule, client-side proxy will
  // verify CA signature with default System CA bundle and SAN matches the host in DestinationRule.
  //
  // If both CaCertificates and SubjectAltNames are non empty in DestinationRule, client-side proxy will verify
  // CA signature with CaCertificates and SubjectAltNames with SAN in presented server certificate during TLS
  // handshake.
  //
  // If CaCertificates is non empty and SubjectAltNames is empty in DestinationRule, client-side proxy will
  // verify CA signature with CaCertificates but will verify the SAN in the presented server certificate
  // matches the host in DestinationRule during TLS handshake.
  //
  // If CaCertificates is empty and SubjectAltNames is non empty in DestinationRule, client-side proxy will
  // verify CA signature ith default System CA bundle  but will verify SubjectAltNames matches SAN in the
  // presented server certificate during TLS handshake.

mesh/v1alpha1/config.proto
nrjpoddar: Add hide_from_docs for UNKNOWN

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,60 @@ message MeshConfig {
   reserved 53;
   reserved "termination_drain_duration";
 
+  enum InsecureCertificateValidationAtClient {
```
ramaraochavali, line 451: Do we need to carry the "Insecure" wording to enum name? 

How about calling it  "ClientTLSValidation" 
- STRICT/SECURE
- INSECURE

nrjpoddar: I like your suggested @ramaraochavali but the issue is `ClientTLSValidation` can be misconstrued as client certificate validation at server for mutual TLS. However, I do think that we can change `InsecureCertificateValidationAtClient` -> `CertificateValidationAtClient` and use "SECURE"/"INSECURE" as enum values. WDYT?

ramaraochavali: Sure. That makes sense.  and use one of the enum values
`INSECURE_SKIP_VERIFY` / `INSECURE`
`SECURE_WITH_DEFAULTS`/'SECURE`

ramaraochavali: Couple of other names that come to mind `OutboundCertificateValidationPolicy`/`OutboundCertificateValidation`/`EgressCertificateValidationPolicy` to indicate it is at client. I am good with any one you choose :-)

rshriram:
the double negative naming and api setting is very confusing. We dont have to copy the naming mistakes that other tools have made verbatim. I would suggest a simple boolean wrapper like this at destination rule

```
google.protobuf.BoolValue validate_server_certificate
```
where if not set, global defaults (if set) apply, and if set, then user overrides.

and a mesh global bool value

```
google.protobuf.BoolValue verify_external_server_certs_using_system_ca
```

And I suggest you take a second look at the way you are trying to infer default values if any for SAN validation. I see one too many pitfalls there that could break in production.

nrjpoddar:
> the double negative naming and api setting is very confusing. We dont have to copy the naming mistakes that other tools have made verbatim. I would suggest a simple boolean wrapper like this at destination rule
> 
> ```
> google.protobuf.BoolValue validate_server_certificate
> ```
> 
> where if not set, global defaults (if set) apply, and if set, then user overrides.
> 
> and a mesh global bool value
> 
> ```
> google.protobuf.BoolValue verify_external_server_certs_using_system_ca
> ```
> 
> And I suggest you take a second look at the way you are trying to infer default values if any for SAN validation. I see one too many pitfalls there that could break in production.

@rshriram I’m assuming that Boolean wrapper takes care of “default” issue with protobuf and it’s possible to override in DRs. If that’s possible, then Boolean wrapper does make sense. For SAN validation you have to do wildcard validation like I commented before as Microsoft, Amazon and many issue wildcard domains in SAN.

nschhina:
> I’m assuming that Boolean wrapper takes care of “default” issue with protobuf and it’s possible to override in DRs. If that’s possible, then Boolean wrapper does make sense.

I think BoolValue wrapper is nullable by default. So the "default" case is that it's null which is sufficient for us to differentiate in code. I've updated the PR with this change

nschhina:
`VerifyExternalServerCertsUsingSystemCa` isn't perfect either imo because We are not only verifying external server cert using system CA but also doing host SAN matching

nrjpoddar:
> > I’m assuming that Boolean wrapper takes care of “default” issue with protobuf and it’s possible to override in DRs. If that’s possible, then Boolean wrapper does make sense.
> 
> I think BoolValue wrapper is nullable by default. So the "default" case is that it's null which is sufficient for us to differentiate in code. I've updated the PR with this change

@nschhina let's wait for some consensus to reduce churn on PR. I don't think `verify_external_server_certs_using_system_ca` is appropriate if we are also acting on SAN which we should do by default. Use of "external" here is misleading as it can be for internal purposes too. I liked `CertificateValidationAtClient` to not get into external/internal nuances. `OutboundCertificateValidation` seems to be fine too. Note than you will have to handle wildcard host matching in SAN by default as many well known sites have wildcard hostnames in SAN.

nschhina:
@nrjpoddar gotchu. Any way we can escalate the priority on this? We are 4 days away from 1.7 release and it's important that we ship this fix out

nrjpoddar:
> @nrjpoddar gotchu. Any way we can escalate the priority on this? We are 4 days away from 1.7 release and it's important that we ship this fix out. Also since istio/istio also requires the security fix that depends on this API change

Sure, you can start the implementation as API naming might take some time. I like Shriram's suggestion of not using double negatives. You can use boolean wrapper and call it `VerifyOutboundCertificate` or `VerifyCertificateAtClient` with default as "true" in 1.7. The caveat with boolean wrapper is if in future for mesh wide we need to give an option for verifying signature using system CA but not SAN then we are stuck with adding another boolean. If you make it enum you can start with two fields and add more enum types later.

rshriram:
Whatever option you decide, please keep the option name exactly the same in both dest rule and mesh global. There is no room for language comprehension by inverting client/server in the field name depending whether its used in dest rule vs mesh config.

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,53 @@ message MeshConfig {
   reserved 53;
   reserved "termination_drain_duration";
 
+  // VerifyCertificateAtClient sets the mesh global default for peer certificate validation
+  // at the client-side proxy when `SIMPLE` TLS or `MUTUAL` TLS (non ISTIO_MUTUAL) origination
+  // modes are used. This setting can be overridden at the host level via DestinationRule API.
+  // By default, VerifyExternalServerCertsUsingSystemCa is true.
```
nrjpoddar, line 454:
```suggestion
  // By default, VerifyCertificateAtClient is true.
```

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,53 @@ message MeshConfig {
...
+  //
+  // When VerifyCertificateAtClient is set to false, certificate and SAN
+  // validation at the client-side proxy is skipped unless overridden by DestinationRule.
+  //
+  // For a particular host the following logic is used when DestinationRule are configured.
```
nrjpoddar, line 459:
```suggestion
  // For a particular host the following logic is used when DestinationRule are configured to control TLS settings in conjunction with global setting.
```

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,64 @@ message MeshConfig {
...
+  //
+  // If CaCertificates is empty and SubjectAltNames is non empty in DestinationRule, client-side
+  // proxy will not verify CA signature but will verify SubjectAltNames matches SAN in the
+  //presented server certificate during TLS handshake.
+
+  // When VerifyCertificateAtClient is set to true, client-side proxy will verify CA signature
+  // with default System CA bundle and ensure SAN matches the host in DestinationRule.
+  //
+  // For SAN, client-side proxy will exact match host in DestinationRule as well as one level
+  // wildcard for the specified host. For example, if host in DestinationRule is x.y.com, client-side
```
nrjpoddar, line 481:
```suggestion
  // wildcard if the specified host in DestinationRule doesn't contain a wildcard. For example, if the host in DestinationRule is x.y.com, client-side
```

nrjpoddar:
Pinging @smawson @linsun

nschhina:
Can we even ship this with 1.7? I was told by release managers that they need at least 48 hours to run perf tests so this might not make it in?

nrjpoddar:
> Can we even ship this with 1.7? I was told by release managers that they need at least 48 hours to run perf tests so this might not make it in?

I doubt it now given it's close to the release date.

mesh/v1alpha1/config.proto
```diff
@@ -448,8 +448,65 @@ message MeshConfig {
...
 
+  // VerifyCertificateAtClient sets the mesh global default for peer certificate validation
+  // at the client-side proxy when `SIMPLE` TLS or `MUTUAL` TLS (non ISTIO_MUTUAL) origination
+  // modes are used. This setting can be overridden at the host level via DestinationRule API.
+  // By default, VerifyCertificateAtClient is true.
+  //
+  // When VerifyCertificateAtClient is set to false, certificate and SAN
+  // validation at the client-side proxy is skipped unless overridden by DestinationRule.
+  //
+  // For a particular host the following logic is used when DestinationRule are configured
```
smawson, line 459: This documentation is way too long and convoluted, seems like it can be cleaned up and made much shorter and less repetitive.

CaCertificates: If set, proxy verifies CA signature based on given CaCertificates. If unset, and VerifyCertificateAtClient is true, proxy uses default System CA bundle. If unset and VerifyCertificateAtClient is false, proxy will not verify the CA.

SubjectAltNames: If set, proxy verifies subject alt names are present in the SAN. If unset, and VerifyCertificateAtClient is true, proxy uses host in destination rule to verify the SANs. If unset, and VerifyCertificateAtClient is false, proxy does not verify SANs.

That's basically the whole thing right? Other than the discussion of wildcard matchin

nschhina: You are right the documentation was probably too long. I'm happy to use the suggested change. cc: @nrjpoddar

hsmade:
As the DR pr is still open, is there no way currently to override the global setting? I have a destination with a self signed certificate, and istio now refuses to route there because of this change, I think.

mesh/v1alpha1/config.proto
aryan-ibm: @nrjpoddar  For mode:Simple also its ignoring SAN from destination rule.

PR #1588, opened 2020-08-10
Define FilterClasses with REPLACE

mandarjog:
1. Define FilterClass enum which defines well know insertion points with respect to filters inserted by the controlplane.
2. Define REPLACE operation, which is the only operation that should rely on filter names. REPLACE is used when proto merge semantics cannot be used, as it does not support field removal.  

Goals:
1. Discourage use of name relative operations with unstable names.
2. Support named replace for filters, when name is defined by the user.


mandarjog:
@rshriram @kyessenov This PR is a redux of #1570 , but with the addition of the enum.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,14 +612,56 @@ message EnvoyFilter {
...
     Operation operation = 1;
 
     // The JSON config of the object being patched. This will be merged using
-    // json merge semantics with the existing proto in the path.
+    // proto merge semantics with the existing proto in the path.
     google.protobuf.Struct value = 2;
+
+    // FILTER_CLASS determines the filter insertion point in the filter chain
+    // relative to the filters implicitly inserted by the control plane.
+    // It is used in conjuction with the ADD operation.
```
howardjohn, line 634: is it valid only with ADD?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,14 +612,56 @@ message EnvoyFilter {
...
+    // filter names.
+    // Filter ordering is important if your filter depends on or affects the
+    // functioning of a another filter in the filter chain.
+    // Within a filter class, filters are inserted in the order of processing.
+    enum FilterClass {
+      // For backwards compatiblity, control plane decides where to insert.
+      UNSPECIFIED = 0;
+
+      // Insert filter before authentication filters inserted by the controlplane.
+      PRE_AUTHN = 1;
```
howardjohn, line 646: is authn -> authz -> stats the ordering? I am wondering about the case of a filter that needs to be after stats AND after authz. Not sure if this is a valid use case, or if we can/want to keep the ordering as an implementation detail?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,14 +612,56 @@ message EnvoyFilter {
       // This is specifically useful when you want your filter first in the
       // list based on a match condition specified in Match clause.
       INSERT_FIRST = 6;
+
+      // Replace contents of a named filter object with new contents.
+      // Replace should be used when you do not need to or cannot use
```
howardjohn, line 617: just wondering, is this a hypothetical or is there use cases we have in mind? I think it probably makes sense to include for completeness, but hopefully people don't end up replacing the whole HCM

networking/v1alpha3/envoy_filter.proto
mandarjog: As of now yes. other operations are name driven.

networking/v1alpha3/envoy_filter.proto
mandarjog: Canarying or partial deployment is a major use case.

There is a default filter that sets up some config, but you want to replace it. There are plenty of examples where you want to unset fields, that proto merge does not support. Replace operation is also *much* more efficient than merge even when both are acceptable.

networking/v1alpha3/envoy_filter.proto
mandarjog: Perhaps full ordering is an implementation detail. We can make it into a `repeated` field, but that would be much.
The option is INSERT_FIRST / INSERT_LAST can also be used to place things at the end.
As part of the impl I am going to implement a bunch of http defaulting, so that specifying `hcm`, `envoy.router` is not necessary, so that makes INSERT_FIRST, INSERT_LAST very straightforward.

howardjohn: I think it would be good to clarify if this is an ordering or not in the doc. Otherwise it will likely be taken as an ordering incorrectly. I don't think we actually need repeated for complexity, just wan to make sure we don't get people depending on things that are not guaranteed

networking/v1alpha3/envoy_filter.proto
```diff
@@ -203,6 +199,89 @@ import "networking/v1alpha3/sidecar.proto";
...
+// fields, REPLACE is preferred over MERGE.
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: EnvoyFilter
+// metadata:
+//   name: myns-ext-authz
+//   namespace: myns
+// spec:
+//   configPatches:
```
rshriram, line 269: where is the workload selector to qualify this to that specific workload?

mandarjog: Missed it, will update.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -127,15 +127,11 @@ import "networking/v1alpha3/sidecar.proto";
...
-//             name: "envoy.http_connection_manager"
-//             subFilter:
-//               name: "envoy.router"
 //     patch:
-//       operation: INSERT_BEFORE
+//       operation: ADD
 //       value: # lua filter specification
-//         name: envoy.lua
+//         name: acme.lua.myfilter # The free form name is used to identify a filter instance.
+//                                 # typed_config."@type" identifies it as a lua filter.
```
rshriram, line 134: missing filter class

networking/v1alpha3/envoy_filter.proto
```diff
@@ -203,6 +199,89 @@ import "networking/v1alpha3/sidecar.proto";
...
+//   configPatches:
+//   - applyTo: HTTP_FILTER
+//     match:
+//       context: SIDECAR_INBOUND
+//     patch:
+//       operation: ADD
+//       filterClass: PRE_STATS
+//       value:
+//         name: istio.openapi
+//         # Full configuration must be specified here.
```
rshriram, line 226: it would be better to have full config here. If not, let us kill this blurb to motivate replace. Replace can stand on its own with a simpler example, like here is how you replace the istio fault injection with a custom fault injection filter implementation with some custom config.

networking/v1alpha3/envoy_filter.proto
rshriram: I would skip the "should be used when you" .. I dont think very many care about when/when not proto merge works :).

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,14 +612,56 @@ message EnvoyFilter {
...
 
     // Determines how the patch should be applied.
     Operation operation = 1;
 
     // The JSON config of the object being patched. This will be merged using
-    // json merge semantics with the existing proto in the path.
+    // proto merge semantics with the existing proto in the path.
     google.protobuf.Struct value = 2;
+
+    // FILTER_CLASS determines the filter insertion point in the filter chain
```
rshriram, line 632: FilterClass (not caps

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,14 +612,56 @@ message EnvoyFilter {
...
+    // relative to the filters implicitly inserted by the control plane.
+    // It is used in conjuction with the ADD operation.
+    // This is the preferred insertion mechanism for adding filters over
+    // the INSERT_* operations since those operations rely on potentially unstable
+    // filter names.
+    // Filter ordering is important if your filter depends on or affects the
+    // functioning of a another filter in the filter chain.
+    // Within a filter class, filters are inserted in the order of processing.
```
rshriram, line 640: May be add a blurb about how well known filters will be automatically ordered within the filter class?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -533,14 +612,56 @@ message EnvoyFilter {
...
+      POST_AUTHN = 2;
+
+      // Insert filter before authorization filters inserted by the controlplane.
+      PRE_AUTHZ = 3;
+
+      // Insert filter after authorization filters inserted by the controlplane.
+      POST_AUTHZ = 4;
+
+      // Insert filter before stats filters inserted by the controlplane.
+      PRE_STATS = 5;
```
kyessenov, line 658: I wonder if we are assuming that stats is a filter too early. The right extension point for stats is an access log extension. If we make the transition, then we're stuck with this API.

networking/v1alpha3/envoy_filter.proto
kyessenov: Where would we put a transcoding filter?

mandarjog:
@rshriram the issue with not having pre-post is that how do you specify a filter in relation to what is automatically inserted, especially in a supported way.

networking/v1alpha3/envoy_filter.proto
mandarjog: Istio does not have a transcoding filter, so it does not matter.
But optically post_authz.

networking/v1alpha3/envoy_filter.proto
mandarjog: I see your point, but pre_stats means that put me before stats so that I can affects its output.
When/if we make the transition pre_stats will continue to work as is.
Post_stats filters themselves will break so it will go thru standard deprecation.

networking/v1alpha3/envoy_filter.proto
kyessenov: I'm pretty sure I saw json transcoder enabled via EnvoyFilter. So it is in fact supported, as any other envoy filters that we supply in our build via EnvoyFilter. I guess the question is, if we have to permit class-less filters, how would they interact with classified ones?

networking/v1alpha3/envoy_filter.proto
kyessenov: Maybe we shouldn't add POST_STATS then since it's bound for deprecation?

networking/v1alpha3/envoy_filter.proto
mandarjog: @kyessenov I meant that the control plane does not implicitly enable any transcoding filter using Istio API.
The user can add any filter they want, and they would place it an appropriate spot, when a user adds a filter, they supply a `name`, which **is** dependable.
Now if anyone else needs to add a filter wrt this filter, they can use positioning by `name` using the existing API.

networking/v1alpha3/envoy_filter.proto
mandarjog: Ok, I can remove POST_STATS.

rshriram:
>@rshriram the issue with not having pre-post is that how do you specify a filter in relation to what is automatically inserted, especially in a supported way.

The point is that most well known filters have a well defined insertion location in their class. We can figure out the location internally. Doing a pre/post will imply user knowledge of what our internal filters are and do. But the filter class here indicates what the user's filter does and we decide to insert it in the right place. If the user does not like that, she can set the filter class to UNSET and use the insert_* to insert the filter at the right location.

mandarjog:
I am trying to make it such that a user  uses name-relative operations against names that they control.
When you are inserting ext authz filter you either care about if it happens before or after the built in authz or you don’t. If you don’t, you can leave it unspecified. You only need to order things if there is a dependency. Stats and authz filters do their work in different callbacks so the order does not matter. ‘Unspecified’ will work fine in that case.

I think pre-post let’s us collect constraints when we need them, and not when they are not needed.

hzxuzhonghu:
From a pure user view, I think this is very confusing. 

Actually, they do not know what does this mean, then how can they use it? 

```
          "UNSPECIFIED",
          "PRE_AUTHN",
          "POST_AUTHN",
          "PRE_AUTHZ",
          "POST_AUTHZ",
          "PRE_STATS",
          "POST_STATS"
```

As i know `Patch` is not only used for filters, but also for route. cluster, etc. How to interpret what does it mean for these configurations?

ramaraochavali:
On the first read, "FilterClass" is very confusing - when we are determining where to insert the filter i.e. its position. Are we calling it "FilterClass" because it is a class of Filter that needs to run before/after those well known insertion points?

rshriram:
There is no use case for having ext authz before our authz. People just want ext authz to be triggered (if our authz rejects something, then it doesn't matter whether it is before or after our authz). The assumption here is that all the filters within a class are non competing, and that the classes are the ones that compete with each other. [I would be interested in seeing a common case example that contradicts this assumption]. The order of classes themselves (decoder, authn, authz, stats) is unlikely to change over time but if a user wants to control the order, we can, in future, introduce a global option to specify the order of classes. The pre-post resembles IPtables very much - which while nice has its own downsides, starting with the amount of confusion people have with regard to when to do PRE vs POST.  The second benefit of the fixed classes is that we can, in future, introduce a class replace option, i.e. user can completely replace istio's authn/stats filters with their own sequence of filters. We can introduce pre-sets for each class (i.e. set of filters to be injected) - e.g., stackdriver stats preset or prometheus stats preset, or lua authz preset with minimal configs (e.g., just give the script), so on. User just needs to specify the preset name and we can do all the insertion stuff inside istiod. PRE-POST makes it unintuitive to accomplish this.

My proposal of fixed classes is to cater to the 80% common case while the pre-post caters to the 20% case that wants to control the specific insertion point.

mandarjog:
If your filter is independent of other filters (and do most of their work in different call backs), then 80% usecase is that ordering does not matter.
I think we are only talking about cases where filters do depend on each other and want the ability to affect or be affected by another filter. I expect “unspecified” to be used often, which happens when just leave it out.

mandarjog:
The other problem with “decoder,authn,authz,stats” is that it mixes callback (decoder) with functions that it performs. Authz and authn are both run in decodeheader stage.
I am planning to add “remove all” filter and then set what you want option, but i have not come across a usecase for it yet.

@hzxuzhonghu I understand your point, however there are other fields which are specialized for specific resources. Filters are used often and have potential ordering issue which clusters and others do not care about. The alternative was to create a whole new api that just deals with filters .

rshriram:
I expect a lot of users to use authz class (for lua/ext authz). That is the 90% use case for this whole thing. Our internal use case is well defined and we can make room for it in the code [we know where we are going to insert].

mandarjog:
I see that even though PRE_ , POST_ are more precise they may be extra overhead for common use cases.
Specifically, POST_AUTH, POST_AUTHN and PRE_STATS cover almost all cases.
We can start with just having {UNSPECIFIED, AUTH, AUTHN, STATS| which are implicit aliases for the above, and introduce the full range of PRE_POST as needed.
This leaves out transcoding filters (like @kyessenov mentioned), but I think they can be left `UNSPECIFIED` since they work on a different call back and will be independent of other filters.

kyessenov:
It sounds fair. It's just very complicated to be usable, and being explicit helps with understanding (that is AUTHN means POST_AUTHN which is somehow different from PRE_AUTHZ, etc).

ramaraochavali:
> We can start with just having {UNSPECIFIED, AUTH, AUTHN, STATS}

+1.  This eliminates lot of confusion and complication. I am more comfortable with it. And this also explains the `FilterClass` well.

Just for my understanding where does STATS filter comes in the order? And are people aware of STATS filter outside of our telemetry and have use cases to configure STATS filter class now?

mandarjog:
a stats filter will be inserted before istio.stats filter.
The order wrt to authz and authn does not normally matter so we should leave it unspecified.

kyessenov:
I think there is still a surprise factor AUTHZ means POST_AUTHZ. If one expect it to be pre-, and inserts some filter that injects data for AUTHZ, it is unexpected. So the problematic use cases are attempts to feed data into the default filters, or work around them (e.g. convert from gRPC to HTTP).

mandarjog:
It is a compromise, my first choice is pre/post. The more appropriate field name for it is filter_position or similar. Position is not so much a property of the filter as it is the relationship of the filter with others inserted by the control plane.

Filterclass with names that appear to be filter properties is inexact but caters to the 80% use case. Given that people generally familiar with the EnvoyFilter api found the pre/post confusing tells me that it will be a hard sell to our users. We will add extra enums if we need it.

linsun:
can we develop some examples of filters (current and w/ the new filter class) so that we can compare if it makes sense?

howardjohn:
One thing that may be useful would a list of common filters for each slot - for example in `Authz: [ext_authz, ...]`

Then users have a few examples of placement

networking/v1alpha3/envoy_filter.proto
mandarjog: updated.

networking/v1alpha3/envoy_filter.proto
mandarjog: done

networking/v1alpha3/envoy_filter.proto
mandarjog: I don;t think we need to specify it now. The existing filter classes address common filters.

mandarjog:
I have updated the lua example and added ext_authz example. Both are typical and are examples of the preferred use of API instead of name relative API.

networking/v1alpha3/envoy_filter.proto
zhaohuabing: I have a use case which can only be addressed by "REPLACE".  I need to replace a TCP proxy with a Redis Proxy. The current Redis configuration in Istio is pretty naive, which doesn't support Redis Cluster, that's why we need to use Envoyfilter to replace it.

PR #1590, opened 2020-08-12
Mesh-wide minimum tls version

liamawhite:
As discussed on slack. It's obviously too late for 1.7.0 but would anyone be against this being cherry-picked into 1.7.1?

Signed-off-by: Liam White <liam@tetrate.io>


mesh/v1alpha1/proxy.proto
```diff
@@ -461,6 +461,31 @@ message ProxyConfig {
   // All control planes running in the same service mesh should specify the same mesh ID.
   // Mesh ID is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.
   string mesh_id = 30;
+
+  // TLS protocol versions.
```
liamawhite, line 465: I copied and pasted from Gateways here because I couldn't see an obvious way to structure that one for re-use.

mesh/v1alpha1/proxy.proto
```diff
@@ -461,6 +461,31 @@ message ProxyConfig {
...
+    // TLS version 1.2
+    TLSV1_2 = 3;
+
+    // TLS version 1.3
+    TLSV1_3 = 4;
+  }
+
+  // Default minimum TLS protocol version. This is overridden by a minimum protocol version set
+  // in a more specific resource (e.g. Gateways).
+  TLSProtocol min_protocol_version = 31;
```
nrjpoddar, line 485: Keeping it inside ProxyConfig has the following tradeoffs:

It can be overridden at the workload level via annotations (proxy.config.io) which may be a security concern. Secondly, if you add it inside ProxyConfig for this config change to take effect you will need to restart all the pods injected with sidecars.

nrjpoddar:
@liamawhite is this supposed to work for Gateways & sidecars? For Gateways can you explain and add comments on how it interacts with the Gateway resource specification? I'm guessing it's an override.

mesh/v1alpha1/proxy.proto
```diff
@@ -461,6 +461,31 @@ message ProxyConfig {
...
+    TLSV1_1 = 2;
+
+    // TLS version 1.2
+    TLSV1_2 = 3;
+
+    // TLS version 1.3
+    TLSV1_3 = 4;
+  }
+
+  // Default minimum TLS protocol version. This is overridden by a minimum protocol version set
```
liamawhite, line 483: @nrjpoddar were you looking for something more detailed than this?

nrjpoddar: This is good, I missed it earlier.

mesh/v1alpha1/proxy.proto
howardjohn, line 485: Is it for outbound, inbound, or both?

liamawhite: This is a good question. The user requirement is only for server as client-side auto already prevents 1.0 but I will defer to whichever one people want to do.

rshriram:
Why does this ever need to be an API? What business does the end user have to dictate the connectivity between two sidecars that Istio supplies? As long as it does not break any app compatibility, the choice of version should be completely upto us (and we might as well pick 1.3 and be done with it). With better transport security, this will become all the more pointless.

mesh/v1alpha1/proxy.proto
smawson: My .02 is that putting it in proxy config is reasonable, this is actually configuring the proxy. If we want to come up with a way to prevent override of proxy settings we should do that holistically and not by putting things in mesh config instead of proxy config.

nrjpoddar:
> Why does this ever need to be an API? What business does the end user have to dictate the connectivity between two sidecars that Istio supplies? As long as it does not break any app compatibility, the choice of version should be completely upto us (and we might as well pick 1.3 and be done with it). With better transport security, this will become all the more pointless.

I agree but as an operator I want to ensure when I have multiple revisions of proxies that in my entire cluster no TLS traffic uses a version below x. I also think that when you have communication between peers where one side doesn't have a sidecar proxy this ensures you never select a TLS version below the required one.

mesh/v1alpha1/proxy.proto
nrjpoddar: Fair point but almost all options in [mesh config](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig) affects proxy so we shouldn't really have this demarcation. IIRC, the idea was to have "ProxyConfig" be the place for only bootstrap config and have all other runtime config be in mesh config.

smawson: We probably want to have a documented process for what goes in MeshConfig vs. ProxyConfig rather than making it up as we go along, we can push on environments WG to take that on I think. My guess is bootstrap vs. non-bootstrap isn't the right model given that we also have per control plane vs. per-proxy as a separate dimension.

But anyway for this specific setting does it actually make sense to change at runtime or is OK being bootstrap config for the proxy?

rshriram:
>I agree but as an operator I want to ensure when I have multiple revisions of proxies that in my entire cluster no TLS traffic uses a version below x. I also think that when you have communication between peers where one side doesn't have a sidecar proxy this ensures you never select a TLS version below the required one.

This is a completely separate topic. You can add TLS stuff to destination rules if you wish to have this feature, like gateways. But for sidecar to sidecar, there is no need to have a global API to select TLS 1.0 of all things, or do TLS auto. We already have an API at the gateway to make sure proxies select the right TLS version per server. At the sidecars, we either do mTLS or logically plaintext (i.e. treat https,tcp etc as all the same).

hzxuzhonghu:
I agree with @rshriram, sidecar--> sidecar communication is transparent to users.

liamawhite:
Ok. @smawson, @nrjpoddar any objections to me just opening a PR to set a default minimum to move this along?

nrjpoddar:
> Ok. @smawson, @nrjpoddar any objections to me just opening a PR to set a default minimum to move this along?

I don't quite follow @rshriram argument. How will a DR help if you want to ensure minimum TLS version where server has sidecar and client doesn't? I understand that use case of Ingress gateways is handled with the current APIs but mixed mode (only one peer in the mesh) needs configurability.

I agree that we don't need to support all the options like TLS 1.0 and a reasonable minimum default makes sense but for legacy use case a user still might want to augment the default?

rshriram:
two different things here. sidecar to sidecar using istio mTLS does not need any form of configuration or api change.. its our protocol and we choose the TLS version, SNI, ports (in case of BTS, etc.), name of the secret to use, the SAN to use, etc.. Just like the way we have the label based auto mtls. 

The second thing that you are talking about is a totally different issue. Sidecar talking to external service or even some k8s service without sidecar (i.e. user is using dest rule to say tls.mode = simple) - this is one case where there is a potential for specifying TLS version. We cannot do anything of that sort in the inbound path for non-mesh TLS traffic, as the inbound listener is doing a plain tcp forwarding for such non-mesh TLS traffic.

ramaraochavali:
FWIW, we do have similar requirement from security to enforce `TLSv2` for all internal traffic. The current Envoy default for server is `TLSv1_0` and where as for client is `TLSv1_2` . So we currently configures with those defaults. So we are taken care on the client side but on server(inbound) we need some mechanism to enforce this.

There is also a requirement from security to enforce/allow only certain cipher suites.

So if we want to provide an API for this, should we default the inbound listeners TLS to `TLSv2` as well for sidecars? So that sidecar -> sidecar will always be `TLSv1_2`

ramaraochavali:
There seems to be other users also asking for this https://github.com/istio/istio/issues/7707.

@rshriram  What do you think? If creating an API for this is not needed, can we change the default of server to TLSv1_2? We are also looking for cipher suites. Should we also default them (by removing the weak ones). BTW, I tried to create an envoy filter for this and it is not actually merging the exiting TLS context but creating a new one. Looks like the only way is to replace the entire TLS context which ugly.

```
kind: EnvoyFilter
metadata:
  name: tls-parameters
  namespace: service-mesh
spec:
  configPatches:
  - applyTo: LISTENER
    match:
      context: SIDECAR_INBOUND
      listener:
        name: virtualInbound
        filterChain:
          name: virtualInbound
    patch:
      operation: MERGE
      value:
        filter_chains:
        - transport_socket:
            name: "envoy.transport_sockets.tls"
            typed_config:
              "@type": "type.googleapis.com/envoy.api.v2.auth.DownstreamTlsContext"
              common_tls_context:
                tls_params: { tls_minimum_protocol_version: TLSv1_2 }
```

@liamawhite what version are you planning to default?

liamawhite:
User requirement was for > 1.0

ramaraochavali:
Ok. I have pushed a PR to set tls version and cipher suites

infa-ddeore:
when will it be available and which version?

ramaraochavali:
https://github.com/istio/istio/pull/27500 - Fixed it and it is available in 1.8 by default and you can enable it in 1.7 by setting PILOT_SIDECAR_ENABLE_INBOUND_TLS_V2 to true

infa-ddeore:
> [istio/istio#27500](https://github.com/istio/istio/pull/27500) - Fixed it and it is available in 1.8 by default and you can enable it in 1.7 by setting PILOT_SIDECAR_ENABLE_INBOUND_TLS_V2 to true

thx for the quick reply, `1.7` [doc](https://istio.io/latest/docs/reference/commands/pilot-agent/#envvars) aren't updated yet looks like, do you also know how to restrict the cipher suites as well?

ramaraochavali:
> thx for the quick reply, 1.7 doc aren't updated yet with this variable

Yes. I have a TODO to update the docs. Sorry for that. Will do next week.

> do you also know how to restrict the cipher suites as well

So that variable also restricts the cipher suites. Please see the PR for more details

infa-ddeore:
> So that variable also restricts the cipher suites. Please see the PR for more details

looks like cipherSuites are predefined, what i meant that the option to set [cipherSuites](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings) similar to gateway

liamawhite:
Thanks @ramaraochavali, will close this PR.

PR #1596, opened 2020-08-14
Added MeshConfig for Custom External CA

hovonako:
Hi Istio team,
This pull request is contains options for Feature Signing Workload Certificate via custom Certificate Authorities
It's based on the discussion of Pull Request: [Signing Workload Certificate via custom Certificate Authorities #26316](https://github.com/istio/istio/pull/26316#issuecomment-673486549)
Please help me to approve it,

Thanks!


mesh/v1alpha1/config.proto
```diff
@@ -447,9 +447,36 @@ message MeshConfig {
...
   reserved "termination_drain_duration";
-
+  
+  // $hide_from_docs
+  // ExternalCA describes information about the external certificate authority that signs the CSRs by workloads inside the
+  // mesh. Istiod will forward authenticated CSR requests from the workloads to the external CA using the istio CA api defined 
+  // in security/v1alpha1/ca.proto
+  message ExternalCA {
+    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+    // Can be IP address or a fully qualified DNS name.
```
costinm, line 457: And port ?

mesh/v1alpha1/config.proto
```diff
@@ -447,9 +447,36 @@ message MeshConfig {
...
+    string address = 1;
+    // Use the tls_settings to specify the tls mode to use. If the CA certificates 
+    // (cert bundle to verify the CA server's certificate) is omitted, Istiod will 
+    // use the system root certs to verify the CA server's certificate. 
+    istio.networking.v1alpha3.ClientTLSSettings tls_settings = 2;
+    // Specifies the Kubernetes secret containing JWT tokens (if ANY) that istiod should present to the CA server
+    // for authorization.
+    string credentials_secret = 3;
```
costinm, line 465: That doesn't work. JWT tokens usually have audience/expiration, and need to be rotated periodically. 

We do support STS - RFC8694 - in agent and envoy ( with google_grpc ). 

Note that we may use an envoy sidecar to handle the connections between Istiod and CA ( or other services we may call ) - and it would be far better to allow the use of Istio own APIs instead of duplicating them in MeshConfig. 

In other words - I don't know why 'connect timeout' is here for a gRPC connection, but we have request timeouts. 

Can we just define the address, and use a DestinationRule or VirtualService to get the other settings ? I think the defaults - https with platform certs - would work for the most common case, including 'central istiod'.

mesh/v1alpha1/config.proto
```diff
@@ -447,9 +447,36 @@ message MeshConfig {
 
   reserved 53;
   reserved "termination_drain_duration";
-
+  
+  // $hide_from_docs
+  // ExternalCA describes information about the external certificate authority that signs the CSRs by workloads inside the
+  // mesh. Istiod will forward authenticated CSR requests from the workloads to the external CA using the istio CA api defined 
+  // in security/v1alpha1/ca.proto
+  message ExternalCA {
```
costinm, line 455: I don't think "External CA" is the correct term here - it is the CA.

I would instead call it message CA {} - and we can add any fields related to setting the CA configuration for Istiod. It wouldn't be very clean if K8S signer - which is also an external CA - would be in a whole different section.

mesh/v1alpha1/config.proto
rshriram: It will not. The most common case here is inhouse CA where the system cert bundles that we have in our debian image does not capture the inhouse CA certs. And typically, for these externally hosted ones, people have API keys or long lived ones. No token rotation infrastructure like what you guys have at google :). So, having an explicit secret here helps.

I agree the connect timeout is kind of useless. But let us not please drag destination rule or virtual service because the caller is istiod for the near future.

mesh/v1alpha1/config.proto
rshriram: I disagree. If we call it as CA, then we have to remove and move all the self signed CA related options into this message. And the semantics get very murky - if address is specified, then istiod becomes a RA but if address is omitted, then istiod becomes a self signed CA. Its not the cleanest way to go about this. Even if we have a CA message, we would end up having a nested ExternalCA message within it to clarify the user intent. We do not want to have a generic catch all term that users will have to interpret based on value of various fields.

rshriram:
>For now - let's get the implementation clear, and decide if we should add a sidecar if Istiod needs to make outbound connections - so we don't have to reinvent the API and metrics and all the features that Istio provides.

Yes but the basic implementation needs these values. We should not shove these into environment variables, which are then forgotten and no one bothers to do the APIs (as is the case today with several citadel related values!). 

Adding a sidecar or not back to istiod is a bigger discussion independent of this CA stuff. We have precedence for TLS settings embedded in mesh config (access log, metrics, config sources, etc.). This object follows the same pattern (with the exception of an additional secret for storing the API keys if any to talk to the CA, in case the user is doing one-way TLS).

mesh/v1alpha1/config.proto
```diff
@@ -470,6 +470,28 @@ message MeshConfig {
...
+    // use the system root certs to verify the CA server's certificate. 
+    istio.networking.v1alpha3.ClientTLSSettings tls_settings = 2;
+    // timeout for forward CSR requests from Istiod to External CA
+    // Default: 10s
+    google.protobuf.Duration request_timeout = 3;
+  }
+  // $hide_from_docs
+  // If specified, Istiod will authorize and forward the CSRs from the workloads to the specified external CA server using the
+  // Istio CA gRPC API.
+  CA ca = 55;
```
rshriram, line 493: please rename this back to externalCA .. as I noted in my comments, CA is very very generic, confusing and misleading.

costinm:
Don't call it 'JWT token' if it's a shared secret. And using a CA without a
token rotation is a bit scary - I understand in some cases
it may be needed, but I would not consider it the 'default' or
'recommended' or optimize for this.
Maybe we can reuse the format/keywords from kubeconfig ( which also
supports a shared secret ).

AFAIK JWT and OAuth2 specs recommend a max 1 hour life for bearer tokens. I
think 1-hour JWTs and mTLS should
be 'best practice' - and API should be optimized for this, while still
supporting a password-equivalent shared secret.

costinm:
I think we disagree on 'ExternalCA' - let's discuss in on EnvWG or
SecurityWG.

This needs to be made consistent with the K8S-as-CA and the settings for
internal CA we already have.

costinm:
And as usual - a design doc listing the API goals, relation with the
existing model.

We already support external CAs directly from agent - and we want to move
the existing configs to MeshCA - I don't think
we can just add this new external CA and keep the existing support as env
variables, as Shriram correctly pointed
env are intended for experiments and temporary things.  Now that we decided
to add CA settings to MeshConfig - we can't
do hacks like we did in values.yaml and global.yaml, where things
were added without any consideration for consistency
and API design.

costinm:
On Mon, Aug 17, 2020 at 6:45 AM Shriram Rajagopalan <

> For now - let's get the implementation clear, and decide if we should add
> a sidecar if Istiod needs to make outbound connections - so we don't have
> to reinvent the API and metrics and all the features that Istio provides.
>
> Yes but the basic implementation needs these values. We should not shove
> these into environment variables, which are then forgotten and no one
> bothers to do the APIs (as is the case today with several citadel related
> values!).
>
> Adding a sidecar or not back to istiod is a bigger discussion independent
> of this CA stuff. We have precedence for TLS settings embedded in mesh
> config (access log, metrics, config sources, etc.). This object follows the
> same pattern (with the exception of an additional secret for storing the
> API keys if any to talk to the CA, in case the user is doing one-way TLS).
>
We do have config sources, access logs metrics - I don't mind if the CA is
consistent with the config sources for example, but this is not something
to decide in the PR review, either EnvWG or SecurityWG and a proper doc are
needed.

Why would a long lived secret be defined only for CA - and not for access
logs/metrics/configsources which have far less security implications
compared
with a CA ? Or even as part of DestinationRule  - I can't imagine the use
case for adding a long-lived token is specific for CA ( or even if we should
support it - unexpiring bearer tokens may be ok for metrics, but not for
signing arbitrary certs).

mesh/v1alpha1/config.proto
costinm: To be clear: our APIs should be able to express 'best practices' - not based on insecure practices. It is an API and model we'll need to support for a long time - I would go as far moving the 'a never expiring bearer token' to 'env variable only' - and not part of the API.  

Let's keep the API based on mTLS and STS - with proper use of OAuth2 and JWT, it's a CA API.

costinm:
On Mon, Aug 17, 2020 at 6:40 AM Shriram Rajagopalan <

> *@rshriram* commented on this pull request.
>
> <https://github.com/istio/api/pull/1596#discussion_r471485838>:
>
> > @@ -447,9 +447,36 @@ message MeshConfig {
>
>    reserved 53;
>    reserved "termination_drain_duration";
> -
> +
> +  // $hide_from_docs
> +  // ExternalCA describes information about the external certificate authority that signs the CSRs by workloads inside the
> +  // mesh. Istiod will forward authenticated CSR requests from the workloads to the external CA using the istio CA api defined
> +  // in security/v1alpha1/ca.proto
> +  message ExternalCA {
>
> I disagree. If we call it as CA, then we have to remove and move all the
> self signed CA related options into this message. And the semantics get
> very murky - if address is specified, then istiod becomes a RA but if
> address is omitted, then istiod becomes a self signed CA. Its not the
> cleanest way to go about this. Even if we have a CA message, we would end
> up having a nested ExternalCA message within it to clarify the user intent.
> We do not want to have a generic catch all term that users will have to
> interpret based on value of various fields.
>

I don't agree we disagree :-) I actually agree with almost all you are
saying !

We don't want MeshConfig to get 'murky' and have confusion - given that we
must get the existing
citadel and 'external CA from agent ( CA_ADDR )' into mesh config as well.

The 'externalCA' may be set either on agent or istiod, with best practice
being in agent.
The config must cover the case of Istiod as a RA, and be consistent with
any setting we move
from env/global for the current stable/beta config and behavior.

It would be crazy to have MeshConfig define settings for the most
experimental and
least supported option we are just adding - and leave the supported cases
in env.

That's why I insist on a proper document and design that covers how we
configure CA
( and SDS )

costinm:
Started https://docs.google.com/document/d/11iTpLQqaqmvGS0uvqktuQwKQSClQHgx7FAuPdIqt3Gg/edit#

rshriram:
The PR as is does not have any shared secrets. It just has the standard TLS settings and nothing else. We can discuss the secret/api keys, etc. in design doc. As far as the name ExternalCA, I dont think anyone is using the eccentric CA_ADDR env var in the agent (not direct end user. May be some vendor distros do but vendors are not the focus here). People use istiod built-in CA or people plug in CA using CA certs. So, may be a clearer way to go about this would be

```
message CertificateAuthority {
 message WorkloadCertificates {
   oneof {
     // we can add the built-in CA message once we start moving configs
     ExternalCA external
   }
  message ExternalCA { ... }
 }
 
 WorkloadCertificates workload_certificates = 1;
}

CertificateAuthority certificate_authority = 1;
```

costinm:
The CA can be called from agent or from istiod. Today we only support the
first, and that is using CA_ADDR - like it or not, there are customers
using it, and the plan is to move it to ProxyConfig - possibly next to
DiscoveryAddress ( which it defaults to).

Istiod has support for signing using k8s - which is also for external ca
integrations, or this new feature.

I think adding a new API for the brand new feature without considering
current features and consistency doesn't make sense.

costinm:
And John is working on secret integration - also similar pattern, agent,
istiod or external. Consistency is critical.

rshriram:
@hovonako fix build issues

hovonako:
@rshriram I fixed it,

mesh/v1alpha1/config.proto
```diff
@@ -470,8 +470,26 @@ message MeshConfig {
...
   google.protobuf.BoolValue verify_certificate_at_client = 54;
 
+  message CA {
+    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+    // Can be IP address or a fully qualified DNS name with port
+    // Eg: custom-ca.default.svc.cluster.local:8932, 192.168.23.2:9000
+    string address = 1;
+    // Use the tls_settings to specify the tls mode to use. If the CA certificates 
+    // (cert bundle to verify the CA server's certificate) is omitted, Istiod will 
+    // use the system root certs to verify the CA server's certificate.
```
hzxuzhonghu, line 480: Why not skip verify, system root certs are diferent in different OS

hovonako: @hzxuzhonghu 
- TLS setting is have a mode TLS MODE = DISABLE. However, I used only on dev environment. Because Istio Certificate Signing API require External CA have to use mTLS.
- Verifying Certificates is only enabled when users intentionally set TLS MODE = MUTAL.

costinm:
Thanks.

Regarding tls_settings:
- DISABLE is legitimate for the case Istiod is making the request via an
Envoy sidecar.
- setting the client TLS certificates should probably be on by default,
using Istiod's own workload identity
unless an explicit override is set. It depends on server capabilities - if
server asks, Istiod should authenticate
itself.
- I think it would be great if we also document that by default, unless an
explicit root ca is set, Istiod
should default to platform roots for TLS connections ( as we are doing for
egress ). In no case should it
be 'skipVerify' ( DISABLE can also be used for testing ).

We'll probably need to add some boolean to indicate it is a 'Istiod-side'
integration - long term we
hope CAs will have direct native support from agent, but we can have the
default to be via Istiod.

+1 on the PR - but we'll need to add some of the other settings listed in
the document, so probably more
adjustments will be made before 1.8 is done.

mesh/v1alpha1/config.proto
hzxuzhonghu: Can we add these to the docs as well otherwise it is a little confusing?

hovonako: I have added @hzxuzhonghu

hovonako:
@costinm I updated code and document, Can you approve this pull request?
Thank you!

hzxuzhonghu (approved):

PR #1599, opened 2020-08-16
Add envoy stats inclusion matcher to proxy config

bianpengyuan:
Add envoy stats inclusion matcher to proxy config, which makes it possible to set global default to expose extra envoy stats and also provides a consistent way to configure this at gateway and sidecar proxy.

ref: https://github.com/istio/istio/issues/26546


mesh/v1alpha1/proxy.proto
```diff
@@ -461,6 +461,24 @@ message ProxyConfig {
   // All control planes running in the same service mesh should specify the same mesh ID.
   // Mesh ID is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.
   string mesh_id = 30;
+
+  // Prefix matcher for Envoy stats creation. To reduce memory and CPU
+  // overhead from envoy stats system, istio-proxy by default only create and
+  // expose a subset of envoy stats. This option is to include extra envoy stats
+  // with prefix match.
+  repeated string stats_inclusion_prefixes = 31;
```
howardjohn, line 469: We may consider a more nested field like:
```
envoy_stats: {
  prefixes: ...
  suffixes: ...
}
```

No strong preference from me though

bianpengyuan:
@howardjohn is there a way to notify user about deprecation of an annotation, other than release notes and proxy log? I'd like to only allow this new way to configure stats inclusion, but don't want to make it a surprise. This will for sure replace annotation at istio.io docs.

howardjohn:
I think we can warn in istioctl analyze? we may already have similar. We should keep it around for a while though, it doesn't add too much cost to support both for some time I think

mesh/v1alpha1/proxy.proto
bianpengyuan: Yeah, good point. Done.

bianpengyuan:
@howardjohn ok, Seems like it is not very reliable to use `istioctl analyze` to convey deprecation message. User need to proactively run it to learn the annotation is no longer supported. I guess we can just keep both of them for now.

mesh/v1alpha1/proxy.proto
```diff
@@ -461,6 +461,24 @@ message ProxyConfig {
...
+
+    // Regexp matcher for Envoy stats creation.
+    repeated string stats_inclusion_regexps = 3;
+  }
+
+  // Stats name inclusion matcher for Envoy stats creation. To reduce memory and
+  // CPU overhead from envoy stats system, istio-proxy by default only create
+  // and expose a subset of envoy stats. This option is to include extra envoy
+  // stats with prefix, regex, and regexps match.
+  EnvoyStatsInclusion envoy_stats_inclusion = 31;
```
ramaraochavali, line 481: Should we call this `ProxyStatsMatcher` so that we might be more generic about inclusion/exclusions in future if we want to?

mesh/v1alpha1/proxy.proto
```diff
@@ -461,6 +461,24 @@ message ProxyConfig {
   // All control planes running in the same service mesh should specify the same mesh ID.
   // Mesh ID is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.
   string mesh_id = 30;
+
+  // Envoy stats name matchers for stats inclusion.
+  message EnvoyStatsInclusion {
+    // Prefix matcher for Envoy stats creation.
+    repeated string stats_inclusion_prefixes = 1;
```
ramaraochavali, line 468: if we agree to my other comment - can we rename this to `inclusion_prefixes`?

And if we retain the message name as `EnvoyStatsInclusion`- we can just call it "prefixes" because the StatsInclusion already comes from the message context

mesh/v1alpha1/proxy.proto
bianpengyuan: Yeah, make sense to me. thanks!

ramaraochavali (approved):
Thank you

mesh/v1alpha1/proxy.proto
```diff
@@ -461,6 +461,24 @@ message ProxyConfig {
   // All control planes running in the same service mesh should specify the same mesh ID.
   // Mesh ID is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.
   string mesh_id = 30;
+
+  // Proxy stats name matchers for stats creation.
+  message ProxyStatsMatcher {
```
mandarjog, line 466: I think extra metrics probably will not be an should be same for ingress and a sidecar.
For example enabling upstream metrics may be ok for sidecar,  but would be too much for gateway.
As a byproduct of annotations we get this configurability.
1. Clearly define the precedence here
2. Perhaps better support for gateways directly? We have some precedent there.

bianpengyuan: This is part of proxy configuration, so if global configuration does not work for both gateway and sidecars, a proxy config override annotation could be used at gateway to remove undesired inclusions. So I don't think we need special config here for gateway.

I'd expect that user won't have mix usage of ProxyConfig annotation and the existing stats inclusion annotation, but anyhow the precedence I think we should take is:
1) existing stats inclusion annotation/env var
2) proxy config annotation
3) global proxy config

douglas-reid:
Can we discuss in P&T WG this week? I'd like to understand our tactical vs strategic plan for adding more telemetry config to proxy config better.

bianpengyuan:
The decision from work group meeting is to hold this PR and wait for the new telemetry APi design, since this could be a good use case of it.

bianpengyuan:
Revive this. @douglas-reid Since we've decided that envoy stats is not in telemetry API scope, can we add these stats inclusion options? thanks!  cc @mandarjog @kyessenov

mandarjog (approved):
lgtm

kyessenov:
LGTM

bianpengyuan:
@douglas-reid @rshriram @nrjpoddar ptal, thanks!

howardjohn (approved):
Do we need any notes about defaults? Is there still "required" stats that are always included we should comment?

bianpengyuan:
@howardjohn Added some comment about default stats inclusion. Thanks! I did not list the specific stats though, since it is easy to get out of sync with real behavior.

douglas-reid (approved):

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,28 @@ message ProxyConfig {
   // kubernetes readiness probe configuration both in schema and logic.
   // Only one health check method of 3 can be set at a time.
   istio.networking.v1alpha3.ReadinessProbe readiness_probe = 31;
+
+
+  // Proxy stats name matchers for stats creation.
```
nrjpoddar, line 456: We should add more description about what this field is and functionality it provides. I would even suggest adding an example match string as a common use case and explain what it does.

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,28 @@ message ProxyConfig {
...
+  message ProxyStatsMatcher {
+    // Proxy stats name prefix matcher for inclusion. Note this is in addition
+    // to the minimum Envoy stats that Istio generates by default.
```
nrjpoddar, line 459: Can we move “Note that..” above all the options on the outer strict message as all fields have the same message and can reduce some repetition.

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,28 @@ message ProxyConfig {
...
+
+    // Proxy stats name regexps matcher for inclusion. Note this is in addition
+    // to the minimum Envoy stats that Istio generates by default.
+    repeated string inclusion_regexps = 3;
+  }
+
+  // Proxy stats name matcher for stats creation. To reduce memory and
+  // CPU overhead from envoy stats system, istio proxy by default creates
+  // and exposes a subset of envoy stats. This option is to control creation of
+  // additional envoy stats with prefix, regex, and regexps match on stats name.
```
nrjpoddar, line 474: Should we link to what Envoy stats are available? How will users know?

nrjpoddar (commented):
Few minor comments. You should add how the existing annotation which will be deprecated behaves and interacts with new options, and a note that this config replaces them.

mesh/v1alpha1/proxy.proto
nrjpoddar: This can be added with comments around line 475 where you have more context.

bianpengyuan: Done.

mesh/v1alpha1/proxy.proto
bianpengyuan: These options will be mentioned here: https://istio.io/latest/docs/ops/configuration/telemetry/envoy-stats/ and the default stats is at the same place so I think user could know them easily.

mesh/v1alpha1/proxy.proto
bianpengyuan: Done.

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,39 @@ message ProxyConfig {
...
+    repeated string inclusion_prefixes = 1;
+
+    // Proxy stats name suffix matcher for inclusion.
+    repeated string inclusion_suffixes = 2;
+
+    // Proxy stats name regexps matcher for inclusion.
+    repeated string inclusion_regexps = 3;
+  }
+
+  // Proxy stats name matcher for stats creation. To reduce memory and
```
nrjpoddar, line 468:
```suggestion
  // Proxy stats matcher defines configuration for reporting custom Envoy stats. To reduce, memory and 
```

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,39 @@ message ProxyConfig {
...
+  // CPU overhead from envoy stats system, istio proxy by default creates
```
nrjpoddar, line 469:
```suggestion
  // CPU overhead from envoy stats system, Istio proxies by default create
```

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,39 @@ message ProxyConfig {
...
+  // and exposes a subset of envoy stats. This option is to control creation of
```
nrjpoddar, line 470:
```suggestion
  // and expose only a subset of Envoy stats. This option is to control creation of
```

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,39 @@ message ProxyConfig {
...
+  // additional envoy stats with prefix, regex, and regexps match on stats name.
```
nrjpoddar, line 471:
```suggestion
  // additional Envoy stats with prefix, suffix, and regex expressions match on the name of the stats.
```

nrjpoddar (commented):
Added minor doc suggestions.

Should we add a note about cardinality increase and potential increase in Prometheus resource consumption if users are not careful about dropping labels?

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,39 @@ message ProxyConfig {
...
+
+    // Proxy stats name regexps matcher for inclusion.
+    repeated string inclusion_regexps = 3;
+  }
+
+  // Proxy stats name matcher for stats creation. To reduce memory and
+  // CPU overhead from envoy stats system, istio proxy by default creates
+  // and exposes a subset of envoy stats. This option is to control creation of
+  // additional Envoy stats with prefix, suffix, and regex expressions match on the name of the stats.
+  // This aims to replace the stats inclusion annotations
```
nrjpoddar, line 472:
```suggestion
  // This replaces the stats inclusion annotations
```

bianpengyuan:
@nrjpoddar thanks for the review! I added a note about prometheus concern.

nrjpoddar (approved):

PR #1600, opened 2020-08-17
update envoy filter docs

ramaraochavali:
Update Envoy Filter docs to use the canonical names and v3 types

Refer https://github.com/istio/istio/pull/25966

### DIFF
--- a/networking/v1alpha3/envoy_filter.proto
+++ b/networking/v1alpha3/envoy_filter.proto
@@ -78,7 +78,7 @@ import "networking/v1alpha3/sidecar.proto";
 //         portNumber: 9307
 //         filterChain:
 //           filter:
-//             name: "envoy.tcp_proxy"
+//             name: "envoy.filters.network.tcp_proxy"
 //     patch:
 //       operation: INSERT_BEFORE
 //       value:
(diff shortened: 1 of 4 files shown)


networking/v1alpha3/envoy_filter.proto
```diff
@@ -131,10 +131,9 @@ import "networking/v1alpha3/sidecar.proto";
 //       operation: ADD
 //       filterClass: AUTHZ # This filter will run *after* the Istio authz filter.
 //       value: # lua filter specification
-//         name: acme.lua.myfilter # The free form name is used to identify a filter instance.
-//                                 # typed_config."@type" identifies it as a lua filter.
+//         name: envoy.filters.http.lua
```
mandarjog, line 134: If you have two lua filters, and use the same name both times, you can order it.
Can you have two lua filters?

Sam question with ext_authz

ramaraochavali: Yes. We can have two lua filters with the same name.

mandarjog: My point is if you have 2 lua filters with the same name then you can’t use name relative ordering. One lua filter could be doing authz and One doing telemetry and if you name both of them the same, you can’t really have valid ordering.

ramaraochavali: Ok. I see. Right now, we can name the filter whatever we want - but Envoy throws a deprecated warning as mentioned here  https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.14.0#deprecated and some point may reject the config. But If we want name "relative" ordering - I do not know how this will work in future.

Do you have any suggestions? Or should I revert this specific doc comment for now?

mandarjog: We should fix that in envoy then, using free form names should not cause warnings. Using *old* names should. 
Do you have an example of this behavior?

ramaraochavali: https://github.com/istio/istio/pull/25966#issuecomment-666905221 - This is one example. Refer to few comments above that has the context

PR #1603, opened 2020-08-18
Explicitly document the root namespace behavior

m-czernek:
In this PR, I wanted to provide a small additional bit of documentation.

The concept of root namespace is quite important for the authz policies yet it is very easily missed, mentioned only once in the whole text.

This PR adds two small paras explaining why readers should care about authz policies. 

WDYT?

### DIFF
--- a/security/v1beta1/authorization.proto
+++ b/security/v1beta1/authorization.proto
@@ -39,13 +39,13 @@ import "type/v1beta1/selector.proto";
 // Requests will be allowed or denied based solely on ALLOW and DENY policies.
 //
 // A request will be internally marked that it should be audited if there is an AUDIT policy on the workload that matches the request.
-// A separate plugin must be configured and enabled to actually fulfill the audit decision and complete the audit behavior. 
+// A separate plugin must be configured and enabled to actually fulfill the audit decision and complete the audit behavior.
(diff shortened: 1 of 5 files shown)


m-czernek:
CI passing; thoughts anyone?

security/v1beta1/authorization.proto
```diff
@@ -181,6 +181,15 @@ import "type/v1beta1/selector.proto";
 //    matchLabels:
 //      version: v1
 // ```
+//
+// When designing authorization policies, you can use the
+// `rootNamespace` option for default network behavior.
+// If you set `rootNamespace` in your `MeshConfig` resource, Istio defaults to
+// using authorization policies set in the root namespace when you create no
+// other policy in your namespace.
```
hzxuzhonghu, line 189: Basically it it right, but I think you need to rework on the words

m-czernek: Can you be more specific please? What exactly do you object to?

hzxuzhonghu: I mean this is not clear for a newbie. 

Both the workload namespace and rootnamespace authoriz policy take effect, which is different from config like DestinationRule

PR #1611, opened 2020-08-27
Remove policy and telemetry component from operator API

bianpengyuan:


bianpengyuan:
@nrjpoddar Can you help approve this? This is purely a clean up PR to remove policy and telemetry from operator, so that user won't be confused.

PR #1612, opened 2020-08-30
Update mesh id type to string

bianpengyuan:
Unfortunately https://github.com/istio/api/pull/1575 was not cherry-picked into 1.7. Mesh ID was already string on master, this is to make it string in 1.7 as well. Seems like this has not been wired up with any configuration options or implementation, so this change is unlikely going to cause problem to user.


bianpengyuan:
@douglas-reid to double check, we don't really have any implementation in 1.7 that depends on this field right?

douglas-reid:
@bianpengyuan i believe that the changes to populate the value happened in master after the 1.7 branch cut.

PR #1614, opened 2020-08-31
VM Health Checking: Add HealthStatus to WorkloadEntry

adiprerepa:
Adds an enum to `WorkloadEntry` with associated Virtual Machine Health status.

Design Doc (@howardjohn): https://docs.google.com/document/d/1G4s_PrmvkwCKoH_kXVeSUAszsaj3ZfpIlraSdeTSPKU/


networking/v1beta1/workload_entry.proto
```diff
@@ -319,5 +319,22 @@ message WorkloadEntry {
   // in the same namespace as the configuration ( WorkloadEntry or a
   // ServiceEntry)
   string service_account = 7;
+
+  // Associated Health of the Workload.
+  // HealthStatus is only sent when the associated WorkloadEntry
+  // is registered with istiod. If HealthStatus is UNHEALTHY,
+  // the Workload is excluded from the EDS response.
+  HealthStatus health_status = 8;
```
smawson, line 327: Seems like this should be in Status rather than in spec?

howardjohn: Its not in endpoint or EndpointSlice, so it may be a bit inconsistent. We also don't really pass the status around everywhere internally but that doesn't need to drive the API

smawson: Yeah depends if we're modelling after Pod or Endpoint. If we move towards auto-registration of WorkloadEntry it is more like EndpointSlice/Endpoint. Also maybe it works better for ServiceEntry to have it on WorkloadEntry too.

OK I'll buy it.

smawson:
LGTM but will give others time to review.

nrjpoddar (commented):
@adiprerepa is this health check "local" done by the agent/proxy on the same VM or a health check from outside?

Should we specify it in comments?

nrjpoddar:
I don't want to rehash the design if it was approved but having a status field in the spec seems odd. The health checks will be performed by agent locally for VMs, report the status via XDS and then WorkloadEntry spec will be updated by istiod.

How will this work with GitOps if folks are keeping WorkloadEntry resources in source control repo?

howardjohn:
> How will this work with GitOps if folks are keeping WorkloadEntry resources in source control repo?

This is a good point. The original idea here was to match Endpoint/slice, and I think this works well IF we are using auto registration. However, its probably viable to have WorkloadEntry manually created but also use health checking. So perhaps status is the logical place to put it.

The bad part is we don't have any current API doing this so we need to infra in API as well as in istio/istio to handle this, but that shouldn't make us pick a suboptimal API

nrjpoddar:
This aim of this field seems to be to inform the user of the running health state and not giving users knobs to configure something so "status" makes more sense sense

We currently update status in other CRDs for analyzer messages so can we do something similar here?

adiprerepa:
closing this pr in favor of using Istio Conditions:

https://github.com/istio/api/blob/master/meta/v1alpha1/status.proto#L45

PR #1620, opened 2020-09-03
Add HealthCheck Config to ProxyStatus

adiprerepa:
For VM Health Checking, adds HealthCheck fields.

Design doc (@howardjohn): https://docs.google.com/document/d/1G4s_PrmvkwCKoH_kXVeSUAszsaj3ZfpIlraSdeTSPKU/edit#


mesh/v1alpha1/proxy.proto
```diff
@@ -461,3 +466,62 @@ message RemoteService {
...
+message TCPHealthCheckConfig {
+  // ip of endpoint
+  string host = 1;
+  // port of endpoint
+  uint32 port = 2;
+}
+
+message ExecHealthCheckConfig {
+  // name of executable to run
+  string executable_name = 1;
```
howardjohn, line 524: We should mirror the kubernetes api exactly, so users can copy+paste there configuration and get the same behavior. This seems a bit different (`command` vs `executable_name`)

adiprerepa: should be exactly the same now.

mesh/v1alpha1/proxy.proto
```diff
@@ -445,6 +445,11 @@ message ProxyConfig {
   // All control planes running in the same service mesh should specify the same mesh ID.
   // Mesh ID is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.
   string mesh_id = 30;
+
+  // VM Health Checking message. This health check config exactly mirrors the kubernetes
+  // readiness probe configuration both in schema and logic.
+  // Only one health check method of 3 can be set at a time.
+  HealthCheckConfig health_check_config = 31;
```
howardjohn, line 452: just brainstorming some alternative name ideas:

* `readiness_probe` - matches k8s, makes it clear its not a liveness probe
* `health_check`


health_check_config works too, I don't have a strong preference

mesh/v1alpha1/proxy.proto
```diff
@@ -461,3 +466,61 @@ message RemoteService {
   // If set then set SO_KEEPALIVE on the socket to enable TCP Keepalives.
   istio.networking.v1alpha3.ConnectionPoolSettings.TCPSettings.TcpKeepalive tcp_keepalive = 3;
 }
+
+message HealthCheckConfig {
+
+  // Number of seconds after the container has started before liveness probes are initiated.
```
howardjohn, line 472:
```suggestion
  // Number of seconds after the container has started before probes are initiated.
```

Not sure why k8s puts "liveness"

mesh/v1alpha1/proxy.proto
```diff
@@ -461,3 +466,61 @@ message RemoteService {
...
+
+message HTTPHealthCheckConfig {
+  // path on localhost to perform an http get to.
+  string path = 1;
+  // port on localhost on which the endpoint lives.
+  uint32 port = 2;
+  // http or https
+  string scheme = 3;
+  // headers the proxy will pass on to make the request.
+  map<string, string> http_headers = 4;
```
howardjohn, line 513: In kubernetes this is a list of `{name, value}`

mesh/v1alpha1/proxy.proto
```diff
@@ -461,3 +466,61 @@ message RemoteService {
...
+    // and the status/able to connect determines health.
+    HTTPHealthCheckConfig http_get = 7;
+    // health is determined by if the proxy is able to connect.
+    TCPHealthCheckConfig tcp_socket = 8;
+    // health is determined by how the command that is executed exited.
+    ExecHealthCheckConfig exec = 9;
+  }
+}
+
+message HTTPHealthCheckConfig {
```
howardjohn, line 505: I think we are missing host. Can we copy the comments as well?

howardjohn (commented):
Looks good!

mesh/v1alpha1/proxy.proto
adiprerepa: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes thought host was in the httpHeaders but will add.

mesh/v1alpha1/proxy.proto
adiprerepa: fixed

mesh/v1alpha1/proxy.proto
adiprerepa: `readiness_probe` was the better choice yeah

adiprerepa:
@howardjohn i think this addresses everything

howardjohn (approved):

mesh/v1alpha1/proxy.proto
```diff
@@ -445,6 +445,11 @@ message ProxyConfig {
   // All control planes running in the same service mesh should specify the same mesh ID.
   // Mesh ID is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.
   string mesh_id = 30;
+
+  // VM Health Checking readiness probe. This health check config exactly mirrors the
+  // kubernetes readiness probe configuration both in schema and logic.
+  // Only one health check method of 3 can be set at a time.
+  ReadinessProbe readiness_probe = 31;
```
hzxuzhonghu, line 452: How to set and make use of it?

adiprerepa:
I believe it is set when the user applies their health check config.

@howardjohn probably has more knowledge here.

rshriram (commented):
is this also part of workload entry or group? users should not be setting proxy config directly

adiprerepa:
@rshriram I do not *think* that the user edits ProxyConfig directly, they apply their readinessProbe config and that is set to proxy config. I may be completely wrong.

@howardjohn would know better.

howardjohn:
The flow will be:
1) User writes the readiness logic in WorkloadGroup (similar to writing deployment spec)
2) Tooling that already translate WorkloadGroup to VM config (which includes the proxy config) will write to the proxy config file to be copied to the VM
3) Proxy will read the proxy config and do the appropriate health checks

Alternative workflow for users not using workload group (this will not be documented or recommended) is directly modifying the proxy config through various means of doing so.

adiprerepa:
the workloadgroup changes will be in another PR. this is just for quick iteration/testing of the implementation of healthchecking

mesh/v1alpha1/proxy.proto
```diff
@@ -461,3 +466,77 @@ message RemoteService {
   // If set then set SO_KEEPALIVE on the socket to enable TCP Keepalives.
   istio.networking.v1alpha3.ConnectionPoolSettings.TCPSettings.TcpKeepalive tcp_keepalive = 3;
 }
+
+message ReadinessProbe {
```
rshriram, line 470: do you want to do this the other way around? move the messages to workload group and import here as those apis have visible documentation on our docs site.

networking/v1alpha3/workload_group.proto
```diff
@@ -109,4 +110,6 @@ message WorkloadGroup {
       // Annotations to attach
       map<string, string> annotations = 2;
   }
+
+  istio.mesh.v1alpha1.ReadinessProbe probe = 3;
```
rshriram, line 114: missing comments. these go on our docs site. so needs to be explanatory

networking/v1alpha3/workload_group.proto
```diff
@@ -16,6 +16,7 @@ syntax = "proto3";
 
 import "google/api/field_behavior.proto";
 import "networking/v1alpha3/workload_entry.proto";
+import "mesh/v1alpha1/proxy.proto";
 
 // $schema: istio.networking.v1alpha3.WorkloadGroup
 // $title: Workload Group
```
rshriram, line 22: missing an example workload group with the health check example

adiprerepa: does this example go in the comments?

mesh/v1alpha1/proxy.proto
adiprerepa: ok that sounds good

adiprerepa:
@rshriram I think I addressed everything

mesh/v1alpha1/proxy.proto
adiprerepa: @hzxuzhonghu this has been moved to `workload_group.proto` and that is where the user will set it, and that message is copied into proxyconfig

PR #1288, opened 2020-02-13
fix authz API documentation

yangminzhu:

PR #1623, opened 2020-09-04
fix authz API documentation

hangpark:
This is a cherry-pick of #1288 for v1.4 since v1.4 is also facing the same issue.


hangpark:
@googlebot I signed it!

PR #1625, opened 2020-09-04
clarify INSERT support for HTTP_ROUTE

GregHanson:
docs currently have `MERGE` as the only supported action for HTTP_ROUTE in EnvoyFilters.  But there is already support for `ADD` and `REMOVE` in the code.  Pending PR to enabled `INSERT_XXXX` actions as well

Pending PR for INSERT operations: https://github.com/istio/istio/pull/27084 

related issue: https://github.com/istio/istio/issues/26692


networking/v1alpha3/envoy_filter.proto
```diff
@@ -664,33 +663,33 @@ message EnvoyFilter {
...
       // order of the element in the array does not matter. Insert
       // before the selected filter or sub filter. If no filter is
       // selected, the specified filter will be inserted at the front
       // of the list.
       INSERT_BEFORE = 4;
 
       // Insert operation on an array of named objects. This operation
-      // is typically useful only in the context of filters, where the
-      // order of filters matter. For clusters and virtual hosts,
+      // is typically useful only in the context of filters or routes,
```
nrjpoddar, line 682: AFAICT `routes` have match criterion and not evaluated purely on order.

GregHanson: yes, routes have a match criteria but order still matters.  The default `/` route that is created by istio for default routing is a catch-all.  Any routes appended after it will never get used because envoy uses the first match

nrjpoddar: I see, should we explain that nuance in the comment?

GregHanson: @nrjpoddar how does this sound?
```
This operation is typically useful only in the context of filters, where order matters, but also for routes since the first matching route is used.  For cluster and virtual hosts . . . 
```

nrjpoddar: Much better, I would just recommend breaking up the comment but the direction you’re heading makes sense.

linsun (approved):
LGTM

GregHanson:
@nrjpoddar can you give this one last look?

nrjpoddar (approved):

PR #1627, opened 2020-09-07
add use client protocol option for destination rules

ramaraochavali:
As discussed in https://github.com/istio/istio/pull/27035#issuecomment-686924765 (and below comments), adding an option for backends to choose if they want to preserve client protocol.


networking/v1alpha3/destination_rule.proto
```diff
@@ -635,6 +635,10 @@ message ConnectionPoolSettings {
     };
     // Specify if http1.1 connection should be upgraded to http2 for the associated destination.
     H2UpgradePolicy h2_upgrade_policy = 6;
+
+    // This is applicable for gateways and when set to true, gateway will initiate connection to 
+    // backend with the protocol of the client when protocol sniffing is enabled at the backend.
```
mandarjog, line 640: I don’t think we should mention protocol sniffing here.
All we are saying is that the backend *can handle* h1.1 and h2 and the gateway should just preserve the protocol.
Backend could be without a sidecar and this would still work, right?

networking/v1alpha3/destination_rule.proto
```diff
@@ -635,6 +635,10 @@ message ConnectionPoolSettings {
     };
     // Specify if http1.1 connection should be upgraded to http2 for the associated destination.
     H2UpgradePolicy h2_upgrade_policy = 6;
+
+    // This is applicable for gateways and when set to true, gateway will initiate connection to
```
howardjohn, line 639: Why only for gateway?

networking/v1alpha3/destination_rule.proto
ramaraochavali: Correct. That makes sense. I will remove the protocol sniffing part.

networking/v1alpha3/destination_rule.proto
ramaraochavali: https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/cluster.go#L1051 - We already do this for sidecars. Gateways we excluded for this https://github.com/istio/istio/issues/21230#issuecomment-588007873

networking/v1alpha3/destination_rule.proto
```diff
@@ -635,6 +635,10 @@ message ConnectionPoolSettings {
     };
     // Specify if http1.1 connection should be upgraded to http2 for the associated destination.
     H2UpgradePolicy h2_upgrade_policy = 6;
+
+    // This is applicable for gateways and when set to true, gateway will initiate connection to 
+    // backend with the protocol of the client when protocol sniffing is enabled at the backend.
+    bool use_client_protocol = 7;
```
rshriram, line 641: is the intent to allow same conn type as downstream (internet) client to gateway? if so, putting this in dest rule feels a bit strange as the decision is on the gateway owner and not the service owner to decide what inbound protocol they want to use

ramaraochavali: The intent is to use the downstream client protocol (internet or application client) when initiating a connection with the backend service (upstream service)

rshriram: right.. but my point is dest rule is authored by service owner. so if someone creates gateway, then by default, its h2/h1 based on port name. but if people want to use same downstream protocol, then they have to go to the dest rule of that service to tinker with this setting, just for gateway? that seems downright weird.

ramaraochavali: The way I am looking at it is  - it is still service owners decision, he is saying whenever my service is accessed via gateway preserve the client protocol. So he sets this flag in DR and it is not changed when creating gateway.

If restricting this to just gateway is causing confusion - I think I can remove that part and apply this uniformly across sidecars and gateways.

ramaraochavali:
@howardjohn @mandarjog @rshriram removed gateway reference to avoid any confusion. PTAL.

networking/v1beta1/destination_rule.proto
```diff
@@ -633,6 +633,9 @@ message ConnectionPoolSettings {
     };
     // Specify if http1.1 connection should be upgraded to http2 for the associated destination.
     H2UpgradePolicy h2_upgrade_policy = 6;
+
+    // If set to true, client protocol will be preserved while initiating connection to backend.
+    bool use_client_protocol = 7;
```
mandarjog, line 638: Setting use-client-protocol makes h2-upgrade-policy ineffective.

networking/v1alpha3/destination_rule.proto
mandarjog: This needs to change as well, right?

ramaraochavali:
Yeah, v1alpha3 also needs to change. thanks for the catch. Updated now

networking/v1beta1/destination_rule.proto
ramaraochavali: Yeah, in the implementation of `use_client_protocol` we will set h2 options and use_downstream_protocol together to support both h1 and h2 clients and yes, upgrade will not happen.  

client protocol will be preserved - indicates upgrade wont happen? Or should I explicitly add that in the docs?

mandarjog: I think we should add it. Even though number of users who will use both these options is vanishingly small, perhaps only you :-)

mandarjog (approved):
Lgtm
With clarification on interaction.

networking/v1beta1/destination_rule.proto
ramaraochavali: makes sense. :-) added in the docs

ramaraochavali:
@rshriram @howardjohn can you PTAL?

mandarjog (approved):
Lgtm

networking/v1beta1/destination_rule.proto
```diff
@@ -633,6 +633,11 @@ message ConnectionPoolSettings {
     };
     // Specify if http1.1 connection should be upgraded to http2 for the associated destination.
     H2UpgradePolicy h2_upgrade_policy = 6;
+
+    // If set to true, client protocol will be preserved while initiating connection to backend.
+    // Note that when this is set to true, h2_upgrade_policy will be ineffective i.e. the client
```
howardjohn, line 638: will we add validation to block setting both of them?

networking/v1beta1/destination_rule.proto
```diff
@@ -633,6 +633,11 @@ message ConnectionPoolSettings {
     };
     // Specify if http1.1 connection should be upgraded to http2 for the associated destination.
     H2UpgradePolicy h2_upgrade_policy = 6;
+
+    // If set to true, client protocol will be preserved while initiating connection to backend.
```
howardjohn, line 637: Should we clarify "protocol" here? Its just h1 vs h2 right? Or does it go beyond that?

Just worried people may think this somehow applies to TCP or even tls

howardjohn (approved):

networking/v1alpha3/destination_rule.proto
rshriram: what is the goal here? to preserve H2 from internet to the sidecar? or to preserve H1 from internet to sidecar?
And in a world where we do better transport security (where all traffic ends up tunneling over H2 to specific envoy port), how will this change?

I understand the intent you provided @ramaraochavali and I buy the rationale. Question is whether to have separate settings for traffic from gateway vs sidecar or just have one set of settings for traffic from another envoy, irrespective  of its role. And if we do the latter, the question is, what are we gaining by forcing H1 all the way, when envoy can downgrade an incoming connection in the listener to H1 in the inboundCluster. We have quite a bit of if-else-if-else thing going on for this H1/H2 thing. so it would be good to have some clarity.

ramaraochavali: Our use case is we want to preserve client protocol at gateway i.e. client connects to gateway with h1 we want h1 to backend and if he connects to gateway with h2 we want h2 to backend. That is what how it works already if protocol sniffing is enabled at sidecar if port is not named. Since it causing confusion, I made it uniform so latter  part of what you described and by user choice.

Regarding what we are are gaining by forcing H1 all the way - there are two main motivations
- There are lot of legacy application client that can only speak h1. While we do this migration of clients from h1 -> h2 and finally land at h2 for every one, our platform team wants to know the client protocol (that is the reason for preserving it) so that they can try protocol specific things.
- If client is on h1, we do an upgrade at gateway on the request path and downgrade it back to h1 on the response path (which is not necessary) and might haver some performance impact (encode/reencode, compress/decompress etc..) which we want to avoid for h1 clients.  We do not have numbers on how much it is, but definitely non zero.

networking/v1beta1/destination_rule.proto
ramaraochavali: Yes. I do plan to add validation during the implementation.

networking/v1beta1/destination_rule.proto
ramaraochavali: Since this is under http settings I thought it is clear that it is h1 vs h2. but happy to clarify if you think it would be useful.

howardjohn: oh didn't realize it was in http - probably self explanatory then

rshriram (approved):

PR #1642, opened 2020-09-18
Fix inconsistent rendering of enum constant and code examples

shamsher31:
To fix https://github.com/istio/api/issues/1643

Found during https://github.com/istio/api/pull/1641

Related:
https://github.com/istio/api/pull/1610


nrjpoddar (approved):
Thanks for fixing this up!

smawson (approved):

PR #1646, opened 2020-09-22
Fix link and bullet point rendering for IstioOperatorSpec

shamsher31:
Ref: https://preliminary.istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#IstioOperatorSpec

### DIFF
--- a/operator/v1alpha1/operator.proto
+++ b/operator/v1alpha1/operator.proto
@@ -214,15 +214,15 @@ message InstallStatus {
         string error = 4;
     }
     // Overall status of all components controlled by the operator.
-    // ```
-    // - If all components have status NONE, overall status is NONE.
-    // - If all components are HEALTHY, overall status is HEALTHY.
-    // - If one or more components are RECONCILING and others are HEALTHY, overall status is RECONCILING.
-    // - If one or more components are UPDATING and others are HEALTHY, overall status is UPDATING.
-    // - If components are a mix of RECONCILING, UPDATING and HEALTHY, overall status is UPDATING.
-    // - If any component is in ERROR state, overall status is ERROR.
-    // - If further action is needed for reconciliation to proceed, overall status is ACTION_REQUIRED.
-    // ```
+    //
+    // * If all components have status `NONE`, overall status is `NONE`.
+    // * If all components are `HEALTHY`, overall status is `HEALTHY`.
+    // * If one or more components are `RECONCILING` and others are `HEALTHY`, overall status is `RECONCILING`.
+    // * If one or more components are `UPDATING` and others are `HEALTHY`, overall status is `UPDATING`.
+    // * If components are a mix of `RECONCILING`, `UPDATING` and `HEALTHY`, overall status is `UPDATING`.
+    // * If any component is in `ERROR` state, overall status is `ERROR`.
+    // * If further action is needed for reconciliation to proceed, overall status is `ACTION_REQUIRED`.
+    //
     Status status = 1;
     // Optional message providing additional information about the existing overall status.
     string message = 3;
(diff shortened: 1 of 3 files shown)


operator/v1alpha1/operator.proto
```diff
@@ -141,24 +141,28 @@ EXAMPLES
 // component values.
 message IstioOperatorSpec {
     // Path or name for the profile e.g.
-    //     - minimal (looks in profiles dir for a file called minimal.yaml)
-    //     - /tmp/istio/install/values/custom/custom-install.yaml (local file path)
+    // ```
```
howardjohn, line 144: I agree the current one is wrong but this also feels wrong. It shouldn't be a code block, it should be a list right? Is there any way to do that?

shamsher31: List item symbol ( - ) works fine for `.md` files. The `.html` does not understand it and hence render it as a continues line.

If we use code blocks using (```), the `.html` converts it to "pre" tag hence we can use ( - ) in whatever way we want and it renders it as following.
 
```
- minimal (looks in profiles dir for a file called minimal.yaml)
- /tmp/istio/install/values/custom/custom-install.yaml (local file path)
```

> Is there any way to do that?

Not sure if there any better alternative.

howardjohn:
blackfriday.Run([]byte(text),
blackfriday.WithExtensions(blackfriday.FencedCode|blackfriday.AutoHeadingIDs))

is what we run. Seems like that ought to handle it - surprised it doesn't..
From cmd/protoc-gen-docs in tools repo

shamsher31:
> I agree the current one is wrong but this also feels wrong. It shouldn't be a code block, it should be a list right? Is there any way to do that?


@howardjohn Found a solution. Instead of ( - ), we should use ( * ) for list items. 😛

https://github.com/istio/api/blob/d9fef91fd35122fa0a3c79a4fac029edec8bb84a/networking/v1alpha3/envoy_filter.proto#L522-L525

operator/v1alpha1/operator.proto
```diff
@@ -210,13 +214,15 @@ message InstallStatus {
         string error = 4;
     }
     // Overall status of all components controlled by the operator.
+    // ```
     // - If all components have status NONE, overall status is NONE.
```
howardjohn, line 218: should this be `*` instead of code ticks

shamsher31: Yes, it needs to be `*`. Thanks.

shamsher31:
@howardjohn Thanks for the review. I have created an issue https://github.com/istio/api/issues/1650 to track these enum constant related problems in docs. Also fixed some of the issues in other PR.

nrjpoddar (approved):

shamsher31:
@linsun @smawson ping.

smawson (approved):

ericvn:
@shamsher31 Note that the link here: 
```
<p>Overrides for default <code>values.yaml</code>. This is a validated pass-through to Helm templates.
 See the <a href="/v1.5/docs/reference/config/installation-options/">Helm installation options</a> for schema details.
 Anything that is available in <code>IstioOperatorSpec</code> should be set above rather than using the passthrough. This
 includes Kubernetes resource settings for components in <code>KubernetesResourcesSpec</code>.</p>
```
isn't getting rendered correctly in the doc update and the resulting link is failing in https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio.io/8238/lint_istio.io/8629.

shamsher31:
@ericvn strange because code generation in `.html` looks ok.

```
<p>Overrides for default <code>values.yaml</code>. This is a validated pass-through to Helm templates.
See the <a href="https://istio.io/v1.5/docs/reference/config/installation-options/">Helm installation options</a> for schema details.
Anything that is available in <code>IstioOperatorSpec</code> should be set above rather than using the passthrough. This
includes Kubernetes resource settings for components in <code>KubernetesResourcesSpec</code>.</p>
```

Looks like `https://istio.io/` is removed during the Automator update.

PR #1648, opened 2020-09-22
clarify http retry attempts

linsun:
our default attempts is 2.  We didn't say it earlier because it depends on the value of http timeout.  Now timeout is disabled by default, the default retry attempts should be clear to our users.

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -1536,10 +1536,11 @@ message StringMatch {
 // {{</tabset>}}
 //
 message HTTPRetry {
-  // Number of retries for a given request. The interval
+  // Number of retries for a given request, default is 2. The interval
   // between retries will be determined automatically (25ms+). Actual
   // number of retries attempted depends on the request `timeout` of the
-  // [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute).
+  // [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute)
+  // which is disabled by default.
   int32 attempts = 1 [(google.api.field_behavior) = REQUIRED];
 
   // Timeout per retry attempt for a given request. format: 1h/1m/1s/1ms. MUST BE >=1ms.
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -1540,7 +1540,7 @@ message HTTPRetry {
   // between retries will be determined automatically (25ms+). When request 
   // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute) 
   // is configured, the actual number of retries attempted also depends on
-  // the specified request `timeout` value.
+  // the specified request `timeout` and `per_try_timeout` values.
   int32 attempts = 1 [(google.api.field_behavior) = REQUIRED];
 
   // Timeout per retry attempt for a given request. format: 1h/1m/1s/1ms. MUST BE >=1ms.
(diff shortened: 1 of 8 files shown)


networking/v1alpha3/virtual_service.proto
```diff
@@ -1536,10 +1536,11 @@ message StringMatch {
...
 message HTTPRetry {
-  // Number of retries for a given request. The interval
-  // between retries will be determined automatically (25ms+). Actual
-  // number of retries attempted depends on the request `timeout` of the
-  // [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute).
+  // Number of retries for a given request, default is 2. The interval
+  // between retries will be determined automatically (25ms+). When request 
+  // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute) 
+  // is configured, the actual number of retries attempted also depends on
+  // the specified request `timeout` value.
```
ramaraochavali, line 1543: and per-try-timeout also right?

hzxuzhonghu (approved):

networking/v1alpha3/virtual_service.proto
linsun: Updated, thanks @ramaraochavali

linsun:
@nrjpoddar @rshriram @louiscryan could one of you approve this minor PR?

networking/v1alpha3/virtual_service.proto
```diff
@@ -1536,10 +1536,11 @@ message StringMatch {
...
 message HTTPRetry {
-  // Number of retries for a given request. The interval
-  // between retries will be determined automatically (25ms+). Actual
-  // number of retries attempted depends on the request `timeout` of the
-  // [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute).
+  // Number of retries for a given request, default is 2. The interval
+  // between retries will be determined automatically (25ms+). When request 
+  // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute) 
+  // is configured, the actual number of retries attempted also depends on
+  // the specified request `timeout` and `per_try_timeout` values.
```
nrjpoddar, line 1543: Is there a default value of `per_try_timeout`?

the actual number of retries attempted also depends on the specified request `timeout` and `per_try_timeout` values. ->  Should we explain how the number of retries are calculated?

linsun: default value is no timeout from [envoy doc](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-field-config-route-v3-hedgepolicy-hedge-on-per-try-timeout).  

```
If left unspecified, Envoy will use the global route timeout for the request. Consequently, when using a 5xx based retry policy, a request that times out will not be retried as the total timeout budget would have been exhausted.
```

linsun:
@hzxuzhonghu @ramaraochavali @nrjpoddar @howardjohn  I need some help in determining the default num of retry attempts now.  It is not simple!  Given our http req timeout default is no timeout, and per_try_timeout is configured same as http req timeout when not specified (which istio doesn't specify by default)... I'm inclined to say we don't do actual retries... Is this conclusion crazy?

ramaraochavali:
@linsun I think it is not correct to conclude there wont be any retries. If client is not able to connect to server i.e. Service Unavailable cases , retries do happen. If client is not able to connect, connect_timeout happens first (even before request_timeout) and request will be retried.  If it is able to connect and request is sent and under process then since we do not have a sensible/safe default for request timeouts - that request is not retried.

ramaraochavali (approved):

networking/v1alpha3/virtual_service.proto
linsun: Regarding how retries are calculated, it is complicated. I think it is best to refer to envoy docs vs istio provides explanation here.

linsun:
@ramaraochavali thanks for chime in.  I agree with your assessment.

PR #1652, opened 2020-09-23
envoyfilter: add ecds option

kyessenov:
Adding more dishes to the kitchen sink.


mandarjog (commented):
We should add more details about how this will be used.
Either in the proto doc or at least in the PR description.

kyessenov:
@mandarjog This is completely opaque from istio standpoint. The goal is to expose an arbitrary proto config through ExtensionConfig discovery service (https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/extension#discovery-service).

howardjohn:
I am a bit confused by your opaque comment. Are you saying Istio code will not need to change to implement this? Or its opaque to users? Or something else

kyessenov:
By opaque I mean uninterpreted by istio.

howardjohn:
Isn't the whole point of Istio APIs to be interpreted by istio? I am not sure how this is used

mandarjog:
Istio `EnvoyFilter` api does not interpret the functional protos though. And this is adding another resource.

kyessenov:
If you look at a line above, it's doing exactly the same thing for ECDS as it does for CDS already. So the precedent is there. We just want more xDS types covered.

mandarjog:
@kyessenov I agree there is precedent, but an example in PR or proto doc would be very useful here.

howardjohn:
The existing ones are not opaque though. For example, CDS: https://github.com/istio/istio/blob/c727a58e60c2c6a5f9bcaf03ee19eb8aca26b70a/pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go#L37

kyessenov:
It is functionally opaque. We'll use UDPA struct to wrap unknown protos (https://github.com/cncf/udpa/blob/master/udpa/type/v1/typed_struct.proto).

howardjohn:
@kyessenov I don't see how UDPA structs impacts this at all. We still need to know what we want the patch to apply to?

kyessenov:
@howardjohn Maybe it's not clear, but here is an example:
- a user wants to deploy a bespoke extension (Wasm or not); istio is not aware of its protobuf config;
- a user applies a regular EnvoyFilter to patch HTTP filter, but instead of using typed_config, we point to config_discovery back to istio, with some name `X`;
- additionally, a user applies the new ECDS EnvoyFilter with the name `X` containing UDPA struct wrapping the filter config.

The delegation is useful in case config for X is repeated in many places (Envoy will have a single instance of X instead of many). It's _necessary_ for Wasm since reloading Wasm is very expensive (v8 needs to boot, compile, etc). So having a single instance of Wasm config X optimizes this path.

howardjohn:
I think that makes sense but is missing step (4) after `additionally, a user applies the new ECDS EnvoyFilter with the name X containing UDPA struct wrapping the filter config.`. Then what? How does Istio apply the ECDS EnvoyFilter?

kyessenov:
Istio needs to implement ECDS and serve only EnvoyFilters there.

howardjohn:
Thanks. A few questions:

* Do we only support ADD then? If not, how do the others work?

* Do we have any access control concerns or will we allow anyone to access any configs cross-cluster?

* What happens if I have conflicting resource names in different namespaces? Are they namespace scoped?

kyessenov:
@howardjohn None of the concerns are specific to ECDS, so we follow what has been accepted in EnvoyFilter. EnvoyFilters compose, so you can PATCH something ADDed.

mandarjog:
@kyessenov  What is the order of ECDS in ADS wrt other resource?

I mostly expect this to be "ADD", but following normal envoyfilter a namespace can "REPLACE" something that is added in the root namespace.

howardjohn:
> EnvoyFilters compose, so you can PATCH something ADDed.

This does not match the implementation. Is it specified in the spec? If so we should reconcile things.

I think these are concerns with other ones by more specifically with ECDS. WASM may have secrets bundled in them, which means that access control *may* be relevant. I am not trying to block this with a bunch of requirements, just want to make sure we explicitly state what we do and do not want or need implemented.

For the conflicting names - I had thought I saw some conversations about having root namespace filters that are overriden by namespaces which is why its particularly relevant, but not a common use case for other XDS types, despite the fact that it *could*

kyessenov:
We definitely don't want secret in Wasm files. There's no confidentiality in either xDS or Istio APIs so secrets should not be there.

kyessenov:
@mandarjog I don't follow the question about the order. ADS is multiplexed, order is not deterministic. I'd expect Istio to implement standalone ECDS as it is simpler.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -366,6 +366,9 @@ message EnvoyFilter {
 
     // Applies the patch to a cluster in a CDS output. Also used to add new clusters.
     CLUSTER = 8;
+
+    // Applies the patch to or adds an extension config in ECDS output.
```
rshriram, line 370: may be mention that the extension filter should be configured first

rshriram (commented):
i am okay with this as long as you add an example describing how to configure a wasm extension via ecds . Otherwise, its hard for others to reuse this and someone else will come along adding a similar functionality for internal use

networking/v1alpha3/envoy_filter.proto
kyessenov: Mentioned in the example.

kyessenov:
Added an example how to use this.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -303,6 +303,58 @@ import "networking/v1alpha3/sidecar.proto";
 //             envoy_grpc:
 //               cluster_name: acme-ext-authz-alt
 // ```
+//
+// The following example deploys a Wasm extension for all inbound sidecar HTTP requests.
```
rshriram, line 307: so there will be no wasm for tcp?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -303,6 +303,58 @@ import "networking/v1alpha3/sidecar.proto";
...
+// metadata:
+//   name: wasm-example
+//   namespace: myns
+// spec:
+//   configPatches:
+//   # The first patch defines a named Wasm extension and provides a URL to fetch Wasm binary from,
+//   # and the binary configuration. It should come before the next patch that applies it.
+//   - applyTo: EXTENSION_CONFIG
+//     match:
+//       context: SIDECAR_INBOUND
```
rshriram, line 321: question: can we match on port/listeners? and if wasm for tcp is possible, how do we distinguish between http vs tcp here?

rshriram: never mind, i understand the flow. I think it will be instructive to indicate that extension config can match only on the context and nothing else. Its just a named object. and this needs to be emphasized in the enum comment as well.

rshriram: In fact, validation check would also help

kyessenov: It's identical to CDS. Match doesn't work that way for CDS, it doesn't matter it's TCP or HTTP.

networking/v1alpha3/envoy_filter.proto
kyessenov: ECDS is only implemented for HTTP, so not yet.

networking/v1alpha3/envoy_filter.proto
mandarjog: We should also elaborate on the named config object. Are we going to make ECDS resource be namespace scoped?
The simple thing is for them to be free form, which introduces override ambiguity. Lets specify an ordering here, leaving future room for namespaced access enforcement.

networking/v1alpha3/envoy_filter.proto
rshriram: can you make a note in the enum doc please? other than that, its lgtm

networking/v1alpha3/envoy_filter.proto
```diff
@@ -303,6 +303,58 @@ import "networking/v1alpha3/sidecar.proto";
...
+//               {}
+//   # The second patch instructs to apply the above Wasm filter to the listener/http connection manager.
+//   - applyTo: HTTP_FILTER
+//     match:
+//       context: SIDECAR_INBOUND
+//     patch:
+//       operation: ADD
+//       filterClass: AUTHZ # This filter will run *after* the Istio authz filter.
+//       value:
+//         name: my-wasm-extension # This must match the name above
```
rshriram, line 347: what happens if people define these two things separately? is that possible or disallowedif possible, then what is format of reference? ns/name - (which we mutate under the hood to remove /).

in fact, i would suggest the following model
1. ns/name/extname format
2. ./name/extname imports from current namespace 
3. */name/extname imports from current or root namespace

import possible only if applicable to current workload by label selection or absence of thereof.
this way its congruent with all other resource reference schemes.

mandarjog: I was suggesting that we start with just matching on free form name, and introduce namespaced names later. We could do it now though.

rshriram: we dont have to implement it but its better to document semantics now. the split config model is most likely to happen very soon. One defines all the available extensions and another decides which ones to consume.

kyessenov: The name is free form. I think it makes sense to qualify by namespace in the example, but any logic to fallback names contradicts EnvoyFilter usages for other named xDS resources. There is no fallback of CDS/RDS/filters by name.

networking/v1alpha3/envoy_filter.proto
kyessenov: Updated.

networking/v1alpha3/envoy_filter.proto
```diff
@@ -303,6 +303,57 @@ import "networking/v1alpha3/sidecar.proto";
...
+// kind: EnvoyFilter
+// metadata:
+//   name: wasm-example
+//   namespace: myns
+// spec:
+//   configPatches:
+//   # The first patch defines a named Wasm extension and provides a URL to fetch Wasm binary from,
+//   # and the binary configuration. It should come before the next patch that applies it.
+//   # This resource is visible to all proxies in the namespace "myns".
+//   - applyTo: EXTENSION_CONFIG
```
hzxuzhonghu, line 320: Trying to understand, does this mean this config will be delivered via ECDS and in reality it can be used by any http filters?  The second patch is one of that.

mandarjog: Yes. One patch points to a "named" configuration, and the other patch defines the named configuration.

It is similar to SDS or RDS in this regard.

kyessenov:
Any more comments @rshriram ?

networking/v1alpha3/envoy_filter.proto
rshriram: true but we have a reference in the ecds. while cds/rds etc. dont have the reference thing. (well except lds but we are not expecting people to define full listeners). And its this reference that is creating an issue in my mind. Freeform names are fine but we have no way to guarantee uniqueness, or resolve ambiguities when two namespaces have same freeform name like "lua".

kyessenov: Ability to provide two versions in two namespaces for the same name seems like a good feature. Are you worried about ADDing things twice?

rshriram: Yes. Or if people have two different configs (one with the filter and another with the thing referring to that)

kyessenov: Ok, added a blurb saying that name should be qualified to avoid collisions. Or not qualified if default in the root is used.

rshriram (approved):

kyessenov:
@linsun @nrjpoddar can one you give second approval?

howardjohn (approved):

PR #1662, opened 2020-09-29
Distinguish upstream errors from local error for outlier detection

kidiyoor:
Subject: Allow to eject hosts when ONLY local origin errors are seen.
Description:
  Currently exposed configuration doesn't allow to only consider local connect error, timeout error etc for outlier detection. This feature is needed if application workload is like a proxy and we want to failover only when the application is irresponsive and not the target. Envoy seen to support this feature already - https://github.com/envoyproxy/envoy/pull/4822.

k8s[ envoy(ingress) -> application ] -> target


networking/v1beta1/destination_rule.proto
```diff
@@ -743,6 +743,40 @@ message OutlierDetection {
   // no effect.
   google.protobuf.UInt32Value consecutive_5xx_errors = 7;
 
+  // The % chance that a host will be actually ejected when an outlier status is
```
mandarjog, line 746: @rshriram this is adding more configuration elements from envoy directly here. They will be passed along almost verbatim. Do we need some different for DR that is half way between and EnvoyFilter and full schematization?

PR #1664, opened 2020-09-30
Amending mesh CA config to include k8s CA config

shankgan:
* Amending mesh CA config to include k8s CA config
Refer https://docs.google.com/document/d/1KAw8-0FivdYQAcWMVTMl-JYvVCn2fpNTbu6Ya3y9d1E/edit#

* Ties with Issue https://github.com/istio/istio/issues/27606 in istio/istio Repository

* Design Doc explaining the configuration requirements - https://docs.google.com/document/d/1X_BSNv7EztNAvND1JJs9YKFzWzn65cYXMeTkPh9ncr4/edit#heading=h.se6wjxq9mtpk


mesh/v1alpha1/config.proto
```diff
@@ -474,9 +474,10 @@ message MeshConfig {
   google.protobuf.BoolValue verify_certificate_at_client = 54;
 
   message CA {
-    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API or CA server integrated with k8s CSR API. 
     // Can be IP address or a fully qualified DNS name with port
     // Eg: custom-ca.default.svc.cluster.local:8932, 192.168.23.2:9000
+    // Or in case of CA integrated with Kubernates, address = kubernates
```
howardjohn, line 480: this feels like a hacked on field here. Can we reconfigure this to make it more natural? Note that we never shipped this API yet, so it can be backwards incompatible

mesh/v1alpha1/config.proto
```diff
@@ -492,10 +493,21 @@ message MeshConfig {
     // Use istiod_side to specify CA Server integrate to Istiod side or Agent side
     // Default: true
     bool istiod_side = 4;
```
howardjohn, line 495: While you are here can you rename this. I don't think `istiod_side` is a very good API field name

mesh/v1alpha1/config.proto
```diff
@@ -492,10 +493,21 @@ message MeshConfig {
...
+
+    message K8sCAConfig {
+      // k8s external signer Name associated with external CA
+      // Defaults to kubernetes.io/legacy-unknown if not specified
+      string signer_name = 1;
+
+      // k8s external CA Root Certificate Location for Istiod/Istio-Agent
+      string ca_root_path = 2;
```
howardjohn, line 503: What is the path? A file in istiod? How does it get there?

mesh/v1alpha1/config.proto
```diff
@@ -492,10 +493,21 @@ message MeshConfig {
     // Use istiod_side to specify CA Server integrate to Istiod side or Agent side
     // Default: true
     bool istiod_side = 4;
+
+    message K8sCAConfig {
+      // k8s external signer Name associated with external CA
+      // Defaults to kubernetes.io/legacy-unknown if not specified
```
howardjohn, line 499: We should not encode this as part of the API IMO, given it is already deprecated

shankgan: Makes sense. will update it to "kubernetes.io/kube-apiserver-client"

mesh/v1alpha1/config.proto
shankgan: This field would used by the helm to configure a VolumeMount for a secret containing the CA root cert. The other option is that we directly pass the CARootCert as part of the config.

mesh/v1alpha1/config.proto
shankgan: istio_ra ? This would imply that istiod would act as an RA for all the CSR's before forwarding it to the external CA

mesh/v1alpha1/config.proto
howardjohn: This is not correct either. This is intended for authentication with api-server only, not mesh traffic. SEe https://github.com/istio/istio/issues/22161

shankgan: Noted. From the thread, I gather that the best decision is to mandate the signername in the mesh config. Given that the workloads in one cluster are potentially going to communicate with workloads in another cluster, I think it would be fair to expect the customer to set this all the time, especially if they intend to integrate with custom K8s CA's

mesh/v1alpha1/config.proto
shankgan: Have resolved this by adding an enum to indicate the integration type of the ExternalCA

mesh/v1alpha1/config.proto
```diff
@@ -473,30 +473,60 @@ message MeshConfig {
...
     // timeout for forward CSR requests from Istiod to External CA
     // Default: 10s
     google.protobuf.Duration request_timeout = 3;
-    // Use istiod_side to specify CA Server integrate to Istiod side or Agent side
-    // Default: true
-    bool istiod_side = 4;
+
+    // whether CSR requests will be authenticated and authorized by Istiod before being forwarded to the external CA. A false implies direct integration of external CA with 
+    // with the Istio Agent/Gateways
+    bool istiod_ra = 4;
```
linsun, line 510: what does ra mean?  istiod_external_ca

mesh/v1alpha1/config.proto
```diff
@@ -473,30 +473,60 @@ message MeshConfig {
...
+    // REQUIRED: CA Integration Type
+    CAIntType type = 6;
   }
-  // $hide_from_docs
+
   // If specified, Istiod will authorize and forward the CSRs from the workloads to the specified external CA
-  // using the Istio CA gRPC API.
-  CA ca = 55;
+  // using the Istio CA gRPC API or using k8s CSR API.
+  CustomCA custom_ca = 55;
```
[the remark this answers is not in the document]
shankgan: Will Unify it. There is no difference.

mesh/v1alpha1/config.proto
```diff
@@ -473,30 +473,60 @@ message MeshConfig {
...
   google.protobuf.BoolValue verify_certificate_at_client = 54;
 
-  message CA {
-    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+  // Configuration of CustomCA integrated with the Istio Ecosystem
+  message CustomCA {
+    // Different Integration types for CA
+    enum CAIntType {
+      // Default Type. Self Signed Certificate
+      ISTIOD = 0;
```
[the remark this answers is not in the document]
shankgan: Thought it would simplify code handling in the application logic. I can remove this if needed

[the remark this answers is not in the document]
costinm: I would remove this completely. There is no such thing as 'istiod integration', and istiod does not implies self-signed cert.

We have a pattern and APIs to specify the certificate for a service ( and default to system roots )

mesh/v1alpha1/config.proto
```diff
@@ -473,30 +473,60 @@ message MeshConfig {
...
+      // Integration with external CA’s plugged into K8s system
+      KUBERNATES = 1;
```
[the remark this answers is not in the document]
costinm: There are 2 configs for the ca - one for the agent, and one in Istiod. 

No point to mix them. How we configure Istiod is slighly different - k8s is an experimental/alpha feature ( we haven't solved some of the critical problems ).

mesh/v1alpha1/config.proto
```diff
@@ -473,30 +473,60 @@ message MeshConfig {
...
+      // Integration with external CA’s supporting the Istio CA gRPC API
+      GRPC = 2;
+      // Support migration of legacy config to newer configuration
+      GOOGLE_MESH = 3;
```
costinm, line 487: I wouldn't add a first class API for a legacy config.

Right now google mesh is auto-detected from the URL - so user doesn't have to configure anything (there are slight changes in the proto - that probably should be handled on server side )

mesh/v1alpha1/config.proto
```diff
@@ -473,30 +473,60 @@ message MeshConfig {
   // ``.x.y.com` suffix.
   google.protobuf.BoolValue verify_certificate_at_client = 54;
 
-  message CA {
-    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+  // Configuration of CustomCA integrated with the Istio Ecosystem
```
costinm, line 476: IMO - this should be removed completely. 

In proxy we only need the CaAddr and the CaCerts.

mesh/v1alpha1/config.proto
costinm: We are mixing 2 different things. Istiod will always authenticate and authorize any CSR it receives, no matter what. It might sign the requests, or delegate to a different CA. The second part is already added to the API. Istio is a RA if no external CA is configured.

mesh/v1alpha1/config.proto
costinm, line 529: Renaming a proto field is a big 'never'. Also not clear what 'custom' means - it is a CA, nothing custom about it.

costinm (commented):
I suggest only adding CAAddr to ProxyConfig - following the existing pattern we use for tracing and other services that envoy talks with directly ( i.e. CaAddr + indication of where to find the cert - maybe keeping it close to the env variables we use ).
And don't mix the server-side implementation with how the client finds the certs.

mesh/v1alpha1/config.proto
howardjohn: We never shipped this. We have changed api fields in unreleased versions many times in the past

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
shankgan: @costinm My argument: There are a lot of shared configs when integrating with an external CA. Doesn't it make sense to have all of them in one place to reduce confusion for the end user? For instance - there are several community contributions that attempt to integrate with custom CA using Istio security grpc API. For this, we need the "istio.networking.v1alpha3.ClientTLSSettings tls_settings" config in order to connect securely to the external CA - both istiod and Istio proxy will need this depending on istio_ra.  The per component configuration (in manifest/charts) can be handled separately  (by the feature owners) by reading specific parameters from the meshConfig.customCA for Istiod/Gateway/Istio-Agent. What do you think?

mesh/v1alpha1/config.proto
shankgan: @costinm - However, there are implementations where the CSR request **never** gets forwarded to Istiod from the workload at all. Correct me if I am wrong, but GoogleMeshCA appears to be one such implementation. Also, as per the design doc mentioned here - https://docs.google.com/document/d/1KAw8-0FivdYQAcWMVTMl-JYvVCn2fpNTbu6Ya3y9d1E/edit#heading=h.se6wjxq9mtpk, this does look like a valid option going forward. I just wanted this variable to indicate whether the implementation wishes to bypass istiod completely. Does **_istiod_bypass_** sound good?

mesh/v1alpha1/config.proto
shankgan: I did run a "make clean gen" and forced the protolock to change the name of the proto field. I can change back to the original names if needed. However, as @howardjohn mentioned, we have never shipped with these changes. What is the protocol here (if there is one)? Should we never change the protobuf message names once a PR has been approved and merged into the repo?

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
shankgan: Resolving this discussion. This config is purely for Istiod. Istio-agent relevant configs will go into proxy-config

mesh/v1alpha1/config.proto
shankgan: This configuration is for Istiod only

mesh/v1alpha1/config.proto
shankgan: Resolving this. This configuration will only be for Istiod. Removing this variable altogether.

mesh/v1alpha1/config.proto
```diff
@@ -473,29 +473,51 @@ message MeshConfig {
   // ``.x.y.com` suffix.
   google.protobuf.BoolValue verify_certificate_at_client = 54;
 
+  // Configuration of CustomCA integrated with the Istiod
   message CA {
-    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+    // Different Integration types for CA
+    enum CAIntType {
```
hzxuzhonghu, line 479: maybe don't need the CA duplicate prefix

mesh/v1alpha1/config.proto
```diff
@@ -473,29 +473,51 @@ message MeshConfig {
...
     // timeout for forward CSR requests from Istiod to External CA
     // Default: 10s
     google.protobuf.Duration request_timeout = 3;
-    // Use istiod_side to specify CA Server integrate to Istiod side or Agent side
-    // Default: true
-    bool istiod_side = 4;
+
+    message K8sCAConfig {
+      // k8s external signer name associated with external CA plugged into k8s system
+      string signer_name = 1;
```
hzxuzhonghu, line 506: What is the signer name, not clear just from the comment

mesh/v1alpha1/config.proto
shankgan: Will rename this to IntegrationType

mesh/v1alpha1/config.proto
```diff
@@ -473,29 +473,50 @@ message MeshConfig {
...
 
+  // Configuration of CustomCA integrated with the Istiod
   message CA {
-    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+    // Different Integration types for CA
+    enum IntegrationType {
+     // Integration with external CA’s plugged into K8s system
+      KUBERNETES = 0;
+      // Integration with external CA’s supporting the Istio CA gRPC API
+      GRPC = 1;
```
myidpt, line 483: Can we use this as the single source of truth for all external CA integrations? The IntegrationType can have:
ISTIOD_RA_KUBERNETES_API -> Istiod as RA, using K8s CSR API to talk to external CA
ISTIOD_RA_ISTIO_API -> Istiod as RA, using Istio CSR API to talk to external CA
ISTIOD_CA_ISTIO_API -> istiod as CA, using Istio CSR API to provision its own key and cert from external CA
AGENT_ISTIO_API -> Istio agent using Istio CSR API to talk to external CA

And you can define a CAConfig with `oneof` type that can take one specific config.

shankgan: I believe we are deliberately keeping the CA configurations for the Istio-agent/Gateways separate from the configurations for Istiod. There is prior detailed discussion on the PR regarding this with @costinm. Will change  KUBERNETES to "ISTIOD_RA_KUBERNETES_API"  and GRPC to ISTIOD_CA_ISTIO_API

mesh/v1alpha1/config.proto
```diff
@@ -479,29 +479,50 @@ message MeshConfig {
   // ``.x.y.com` suffix.
   google.protobuf.BoolValue verify_certificate_at_client = 54;
 
+  // Configuration of Istiod Registration Authority to integrate with custom CA
   message CA {
-    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API. 
+    // Different Integration types for CA
+    enum IntegrationType {
```
costinm, line 485: Redundant ? Presence of the message will indicate if it's k8s or not.

'integration type' is also not a very common term - either in security or istio. Provider maybe.

mesh/v1alpha1/config.proto
```diff
@@ -479,29 +479,50 @@ message MeshConfig {
...
+      // Istiod as RA, using K8s CSR API to talk to external CA
+      ISTIOD_RA_KUBERNETES_API = 0;
+      // Istiod as RA, using Istio CSR API to talk to external CA
+      ISTIOD_RA_ISTIO_API = 1;
+    }
+
+    // Address of the CA server implementing the Istio CA gRPC API 
     // Can be IP address or a fully qualified DNS name with port
     // Eg: custom-ca.default.svc.cluster.local:8932, 192.168.23.2:9000
     string address = 1;
```
costinm, line 495: The pattern we use in MeshConfig/ProxyConfig is to use a RemoteService.

mesh/v1alpha1/config.proto
```diff
@@ -479,29 +479,50 @@ message MeshConfig {
...
+
+    message KubernetesCAConfig {
+      // Refer to https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/
+      // for details regarding how signer_name is used by external CAs plugged into Kubernetes system
+      string signer_name = 1;
+    }
+
+    // Configs for integrating with Kubernetes External CA
+    // Required if type is Kubernetes
+    KubernetesCAConfig kubernetes_config = 5;
```
costinm, line 518: CA is redundant ( it's in a struct called CA )
Config is redundant - every field in MeshConfig is a config for something.

So how about just k8s or kubernetes ?

costinm (commented):
I think for 1.8 it is sufficient to use env variables, it's still pretty experimental.

PR #1665, opened 2020-09-30
Idletimeout clarification

linsun:
Updated:
Turned out doc is correct, our code comment is wrong, I'll open a PR to correct it.   A minor update on drain, thanks to @ramaraochavali for helping me clarify things.

--------
We are having it disabled by default:

https://github.com/istio/istio/blob/master/pilot/pkg/model/context.go#L480

Also, from my envoy sidecar default config:

```
            "stream_idle_timeout": "0s",
```

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -617,9 +617,12 @@ message ConnectionPoolSettings {
     // cluster at a given time. Defaults to 2^32-1.
     int32 max_retries = 4;
 
-    // The idle timeout for upstream connection pool connections. The idle timeout is defined as the period in which there are no active requests.
-    // If not set, the default is 1 hour. When the idle timeout is reached the connection will be closed.
-    // Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections.
+    // The idle timeout for upstream connection pool connections. The idle timeout 
+    // is defined as the period in which there are no active requests.
+    // The default is disabled. When the idle timeout is set and the timeout value 
+    // is reached, the connection will be closed. 
+    // Note that request based timeouts mean that HTTP/2 PINGs will not 
+    // keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections. 
     google.protobuf.Duration idle_timeout = 5;
 
     // Policy for upgrading http1.1 connections to http2.
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -619,8 +619,8 @@ message ConnectionPoolSettings {
 
     // The idle timeout for upstream connection pool connections. The idle timeout 
     // is defined as the period in which there are no active requests.
-    // The default is disabled which means no timeout. When the idle timeout is set 
-    // and the timeout value is reached, the connection will be closed. If the connection is an HTTP/2 
+    // If not set, the default is 1 hour. When the idle timeout is reached, 
+    // the connection will be closed. If the connection is an HTTP/2 
     // connection a drain sequence will occur prior to closing the connection. 
     // Note that request based timeouts mean that HTTP/2 PINGs will not 
     // keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections. 
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -619,8 +619,9 @@ message ConnectionPoolSettings {
 
     // The idle timeout for upstream connection pool connections. The idle timeout 
     // is defined as the period in which there are no active requests.
-    // The default is disabled. When the idle timeout is set and the timeout value 
-    // is reached, the connection will be closed. 
+    // The default is disabled which means no timeout. When the idle timeout is set 
+    // and the timeout value is reached, the connection will be closed. If the connection is an HTTP/2 
+    // connection a drain sequence will occur prior to closing the connection. 
     // Note that request based timeouts mean that HTTP/2 PINGs will not 
     // keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections. 
     google.protobuf.Duration idle_timeout = 5;
--- a/networking/v1alpha3/destination_rule.gen.json
+++ b/networking/v1alpha3/destination_rule.gen.json
@@ -470,7 +470,7 @@
             "format": "int32"
           },
           "idleTimeout": {
-            "description": "The idle timeout for upstream connection pool connections. The idle timeout is defined as the period in which there are no active requests. If not set, the default is 1 hour. When the idle timeout is reached the connection will be closed. Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections.",
+            "description": "The idle timeout for upstream connection pool connections. The idle timeout is defined as the period in which there are no active requests. The default is disabled. When the idle timeout is set and the timeout value is reached, the connection will be closed. Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections.",
             "type": "string"
           },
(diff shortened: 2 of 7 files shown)


networking/v1alpha3/destination_rule.proto
```diff
@@ -617,9 +617,12 @@ message ConnectionPoolSettings {
     // cluster at a given time. Defaults to 2^32-1.
     int32 max_retries = 4;
 
-    // The idle timeout for upstream connection pool connections. The idle timeout is defined as the period in which there are no active requests.
-    // If not set, the default is 1 hour. When the idle timeout is reached the connection will be closed.
-    // Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections.
+    // The idle timeout for upstream connection pool connections. The idle timeout 
+    // is defined as the period in which there are no active requests.
+    // The default is disabled. When the idle timeout is set and the timeout value
```
nrjpoddar, line 622: What does disabled mean? I'm guessing no idle time out? What happens in that case? Does the connection remains open for ever?

networking/v1alpha3/destination_rule.proto
```diff
@@ -617,9 +617,12 @@ message ConnectionPoolSettings {
...
     int32 max_retries = 4;
 
-    // The idle timeout for upstream connection pool connections. The idle timeout is defined as the period in which there are no active requests.
-    // If not set, the default is 1 hour. When the idle timeout is reached the connection will be closed.
-    // Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive. Applies to both HTTP1.1 and HTTP2 connections.
+    // The idle timeout for upstream connection pool connections. The idle timeout 
+    // is defined as the period in which there are no active requests.
+    // The default is disabled. When the idle timeout is set and the timeout value 
+    // is reached, the connection will be closed. 
+    // Note that request based timeouts mean that HTTP/2 PINGs will not
```
nrjpoddar, line 624: Is this accurate? HTTP/2 PING frames are sent over the same connection and used to determine if a connection is still active or not.

Is Envoy ensuring that any PING HTTP/2 frames doesn't reset TCP level idle timer?

networking/v1alpha3/destination_rule.proto
linsun: yes disabled means no timeout.  It would remain open forever.

nrjpoddar: Do you think adding that in doc will be good? I always get confused by what disabled means in different contexts.

linsun:
After a big more digging, I believe we are using the default idle timeout which is 1 hour, so the doc is correct.  I got confused from the code comment https://github.com/istio/istio/blob/master/pilot/pkg/model/context.go#L480 and the value of stream_idle_timeout which we don't expose users to configure it today.

@howardjohn let me know if I'm wrong here.

linsun:
cc @GregHanson since you are looking at some of these as well

networking/v1alpha3/destination_rule.proto
linsun: There is no evidence this is no longer accurate. :)  Given this is an http level configuration, pings are not http so would not apply here.

linsun: I assume here it means it is different request, so if you are running HTTP/2 pings on the side, it can keep the pings request alive but not your other request.

networking/v1alpha3/destination_rule.proto
linsun: probably if you are confused. :)

ramaraochavali: Are you sure if not specified it is disabled i.e. no idle timeout? If we do not specify any value the default is 1 hour in Envoy https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/protocol.proto.html?highlight=idle_timeout#core-httpprotocoloptions

The code comment is https://github.com/istio/istio/blob/master/pilot/pkg/model/context.go#L480 is incorrect. It was the original behaviour in Envoy i.e. when not set no idle timeout. Later it has been changed to 1 hour.

stream_idle_timeout is a different timeout from idle_timeout. idle_timeout applies at the connection level where as stream_idle_timeout is for individual streams carried by the http connection. https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts.html?highlight=timeouts#how-do-i-configure-timeouts

I think changing this is not ocrrect.

ramaraochavali (changes requested):
It seems the current document is correct unless I am missing some thing. Can you please review my comments?

networking/v1alpha3/destination_rule.proto
linsun: @ramaraochavali I think you are correct.  I came to similar conclusion earlier too for idle_timeout.

https://github.com/istio/api/pull/1665#issuecomment-701744044

Thanks for the clarification on stream_idle_timeout, which we don't allow users to set today.

ramaraochavali (approved):

networking/v1alpha3/destination_rule.proto
nrjpoddar: I don't think that's how it will work. PING should be send on the same underlying TCP connection as HTTP/2 by default only creates a single connection.

@PiotrSikora am I not misunderstanding something here? A TCP level timer should be reset if any Layer 7 packets are send.

PiotrSikora: But this is not a TCP-level timer, this is timeout for a HTTP-level timer, and each HTTP request gets it's own.

HTTP/2 PINGs are connection-level events that are not associated with any particular HTTP request, so this descriptions sounds about right and it matches Envoy's documentation.

nrjpoddar: Damn I got confused by 2 of @linsun PRs and thought this one was for TCP setting. Sorry about the churn.

linsun: thanks @PiotrSikora for the clarification!  isn't HTTP/2 pings ping based on http2? :)  If it is TCP only, yep, it won't impact idle_timeout which is for http(2) protocol.

nrjpoddar (approved):

networking/v1alpha3/destination_rule.proto
PiotrSikora: HTTP/2 PING is HTTP/2 frame, yes. But it's associated with the whole HTTP/2 connection, which carries many independent HTTP/2 streams (request/response pairs), and the timeouts are set on a per-stream basis.

nrjpoddar: That's what I thought. This idle timeout talks about connections and not per HTTP/2 stream timeout afaict.

This is the current comment:

// Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive.

Why won't a PING frame keep the underlying connection alive? I must be missing something basic here.

linsun: There is also stream_idle_timeout, which HTTP/2 Pings won't impact.   If idle_timeout is not associated with any particular HTTP request as the doc has indicated, sounds like HTTP/2 Pings can keep the connection from hitting idle_timeout.

@PiotrSikora @ramaraochavali thoughts?

ramaraochavali: The way idle timeout is implemented in Envoy is the timer is started on new connection and is disabled a new request starts and enabled when the request ends. So it is always at the request level. If there are no requests on the connection - it will be treated as idle.

linsun: @ramaraochavali based on what you said, iiuc, continuous http/2 pings will be the active request to keep the connection from idle.   Safe to delete this?

```
Note that request based timeouts mean that HTTP/2 PINGs will not keep the connection alive.
```

ramaraochavali: Http2 ping is not considered a request and hence the above comment is correct.

linsun: I see.  That was what I was not sure earlier given it is not regular ping, it is Http2 ping.   I think we are good now!

PR #1667, opened 2020-10-01
Add example for how to use IstioOperatorSpec

shamsher31:
Attempt to fix https://github.com/istio/istio/issues/24739

Ref: https://preliminary.istio.io/latest/docs/reference/config/istio.operator.v1alpha1/#IstioOperatorSpec

### DIFF
--- a/operator/v1alpha1/operator.proto
+++ b/operator/v1alpha1/operator.proto
@@ -44,23 +44,21 @@ option (gogoproto.sizer_all) = false;
 // Because the spec is a customization API, specifying an empty IstioOperatorSpec results in a default Istio
 // component values.
 //
-// ```
-// // Apply a default Istio installation
-// istioctl install
-//
-// // Use a different profile for installation
-// istioctl install --set profile=demo --skip-confirmation
-// istioctl install --set profile=minimal --set values.global.istioNamespace=istio-dev
-//
-// // Set custom HUB and TAG
-// istioctl install --set hub=gcr.io/istio-testing --set tag=latest
-//
-// // Enable Tracing
-// istioctl install --set meshConfig.enableTracing=true
-//
-// // Install multiple control plane revisions
-// istioctl install --set revision=canary
-// istioctl install --set revision=1-7-3
+// ```yaml
+// apiVersion: install.istio.io/v1alpha1
+// kind: IstioOperator
+// spec:
+//   profile: default
+//   hub: gcr.io/istio-testing
+//   tag: latest
+//   revision: 1-8-0
+//   meshConfig:
+//     accessLogFile: /dev/stdout
+//     enableTracing: true
+//   components:
+//     egressGateways:
+//     - name: istio-egressgateway
+//       enabled: true
 // ```
 //
 message IstioOperatorSpec {
--- a/operator/v1alpha1/operator.proto
+++ b/operator/v1alpha1/operator.proto
@@ -43,6 +43,26 @@ option (gogoproto.sizer_all) = false;
 // The spec is a used to define a customization of the default profile values that are supplied with each Istio release.
 // Because the spec is a customization API, specifying an empty IstioOperatorSpec results in a default Istio
 // component values.
+//
+// ```
+// // Apply a default Istio installation
+// istioctl install
+//
+// // Use a different profile for installation
+// istioctl install --set profile=demo --skip-confirmation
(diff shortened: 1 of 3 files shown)


linsun (approved):
Thanks!

operator/v1alpha1/operator.proto
```diff
@@ -43,6 +43,26 @@ option (gogoproto.sizer_all) = false;
 // The spec is a used to define a customization of the default profile values that are supplied with each Istio release.
 // Because the spec is a customization API, specifying an empty IstioOperatorSpec results in a default Istio
 // component values.
+//
+// ```
+// // Apply a default Istio installation
+// istioctl install
+//
+// // Use the different profile for installation
```
smawson, line 51:
```suggestion
// // Use a different profile for installation
```

operator/v1alpha1/operator.proto
```diff
@@ -43,6 +43,26 @@ option (gogoproto.sizer_all) = false;
...
+// istioctl install --set profile=demo --skip-confirmation
+// istioctl install --set profile=minimal --set values.global.istioNamespace=istio-dev
+//
+// // Set custom HUB and TAG
+// istioctl install --set hub=gcr.io/istio-testing --set tag=latest
+//
+// // Enable Tracing
+// istioctl install --set meshConfig.enableTracing=true
+//
+// // Install multiple control plan revision
```
smawson, line 61:
```suggestion
// // Install multiple control plane revisions
```

smawson (commented):
A couple typo fixes but otherwise LGTM

shamsher31:
`make gen` generated a lot of files that are not related to this PR. I also tried `make clean gen` but it still shows the same issue.

operator/v1alpha1/operator.proto
```diff
@@ -43,6 +43,26 @@ option (gogoproto.sizer_all) = false;
 // The spec is a used to define a customization of the default profile values that are supplied with each Istio release.
 // Because the spec is a customization API, specifying an empty IstioOperatorSpec results in a default Istio
 // component values.
+//
+// ```
+// // Apply a default Istio installation
```
smawson, line 48: This is really confusing to me, this is in the documentation of the spec but these examples don't use a spec, they use --set commands.

I expected to see example spec yaml files.

smawson: Sorry I didn't notice this earlier.

shamsher31: > This is really confusing to me, this is in the documentation of the spec but these examples don't use a spec, they use --set commands.

@smawson As per the issue https://github.com/istio/istio/issues/24739, the example using `--set` command is requested.

> I expected to see example spec yaml files.

Maybe we can point to spec YAML somewhere else because I think this example already occupies lot of area in the doc.

ostromart: The issue is not really asking for --set examples, it just says the relationship between --set and -f is confusing. 
How about this: give the first example as a YAML file, show the same thing with --set and explain that the remaining examples are with --set for compactness; but -f is strongly recommended for production use.

ostromart: related: https://github.com/istio/istio.io/pull/8291

shamsher31: Thanks will work on it.

smawson: IMO this is documenting the yaml file, so only yaml examples belong here. If we want to document --set that goes somewhere else (istioctl install documentation).

shamsher31: @smawson @ostromart I have updated using YAML examples.
Thanks.

operator/v1alpha1/operator.proto
```diff
@@ -43,6 +43,24 @@ option (gogoproto.sizer_all) = false;
 // The spec is a used to define a customization of the default profile values that are supplied with each Istio release.
```
smawson, line 43:
```suggestion
// The spec is used to define a customization of the default profile values that are supplied with each Istio release.
```

smawson: While you're in this file can you fix this typo?  :)

smawson (approved):

operator/v1alpha1/operator.proto
shamsher31: I will take care of that in some other PR as this PR got merged.

PR #1670, opened 2020-10-02
Adding HoldApplicationUntilProxyStarts to proxyconfig

stewartbutler:
Adding API entry for HoldApplicationUntilProxyStarts to proxyconfig


mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,11 @@ message ProxyConfig {
   // kubernetes readiness probe configuration both in schema and logic.
   // Only one health check method of 3 can be set at a time.
   istio.networking.v1alpha3.ReadinessProbe readiness_probe = 31;
+
+  // Boolean flag for enabling/disabling the holdApplicationUntilProxyStarts behavior.
+  // This feature adds hooks to delay application startup until the pod proxy
+  // is ready to accept traffic, mitigating some startup race conditions.
+  bool hold_application_until_proxy_starts = 32;
```
howardjohn, line 458: maybe we should make it a BoolVal wrapper so we can make it default if we want in the future?

proto.lock
```diff
@@ -13049,13 +13049,7 @@
...
                       },
                       {
-                        "name": "additional_bindings",
-                        "aggregated": [
-                          {
-                            "name": "get",
-                            "value": "/v1/services/{service_name}/config"
-                          }
-                        ]
+                        "name": "additional_bindings"
```
howardjohn, line 13052: I don't know how this happened but this file shouldn't change

howardjohn (approved):

mesh/v1alpha1/proxy.proto
```diff
@@ -451,6 +451,11 @@ message ProxyConfig {
   // kubernetes readiness probe configuration both in schema and logic.
   // Only one health check method of 3 can be set at a time.
   istio.networking.v1alpha3.ReadinessProbe readiness_probe = 31;
+
+  // Boolean flag for enabling/disabling the holdApplicationUntilProxyStarts behavior.
```
linsun, line 455: Add a default value (false) here?

stewartbutler:
@howardjohn @linsun Can I get approval on this?

mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
```diff
@@ -362,6 +362,11 @@
           },
           "proxyStatsMatcher": {
             "$ref": "#/components/schemas/istio.mesh.v1alpha1.ProxyConfig.ProxyStatsMatcher"
+          },
+          "holdApplicationUntilProxyStarts": {
+            "description": "Boolean flag for enabling/disabling the holdApplicationUntilProxyStarts behavior. This feature adds hooks to delay application startup until the pod proxy is ready to accept traffic, mitigating some startup race conditions. Default value is 'false'. If users set the 'values.global' setting of the same name to 'true' this value will be ignored.",
```
linsun, line 367: Is the last sentence correct? I'd assume user could set it to false to disable this even if it is globally set to true.

linsun:
@stewartbutler just a minor comment, can you fix the conflict and gen, I'll approve it.

linsun:
@stewartbutler a friendly ping on this.

mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
stewartbutler: Well, with the way I'd implemented the template it would be true. This value would only be checked if the global setting was not set. I can swap that, so the global setting is only used if a proxyconfig setting isn't set.

linsun: I like this better `the global setting is only used if a proxyconfig setting isn't set.` - I think this is consistent with how we handle other annotations like sidecar injector.

@howardjohn thoughts?

howardjohn:
```
Boolean flag for enabling/disabling the holdApplicationUntilProxyStarts behavior. This feature adds hooks to delay application startup until the pod proxy is ready to accept traffic, mitigating some startup race conditions. Default value is 'false'.
```

Is my recommendation. I would not put `values.global` option here, as this is at a higher level than the api here and we will add a warning to istioctl to switch to direct mesh config api if user uses the old one.

> only used if a proxyconfig setting isn't set.

is a bit odd since this *is* proxy config. The only config will be proxy config. This may be modified by values.yaml, and annotation, or others, but ultimately proxy config is the source of truth. And we should not document how to configure proxy config here, but rather at the top level of proxy config

stewartbutler:
OK, I think it is resolved. Added release-notes-none -- the release note will be added in the istio/istio PR

linsun (approved):
Thank you @stewartbutler and @howardjohn! The updated wording LG!

PR #1674, opened 2020-10-03
add external action to authorization policy

yangminzhu:
There are some pending discussions in the design, send out a draft PR to help speed up for making the decision.

cc @incfly @liminw @mandarjog @kyessenov @rshriram, 

[latest commit 3](https://github.com/istio/api/pull/1674/commits/7602328beb86948983b1ba5a403e760891f30efd): this is the option of using Istio's own abstraction. The API is not passthrough but it's mostly exactly the same as the Envoy ext_authz API with a lot simplification to make it easier to use from end users.

[commit 2](https://github.com/istio/api/pull/1674/commits/099b62329c4aba9b33799b3fc8c54877d843c28e): this is the option of simplify not include the `config` field in first version, we can always add later for those most used fields based on actual user feedback so that users don't need EnvoyFilter in most cases. Even this is very simple but in reality it will most likely always need a separate EnvoyFilter to customize some commonly used fields (e.g. control headers in the request).

[commit 1](https://github.com/istio/api/pull/1674/commits/d7e8162567e18265f72d3a6e98c88e6c58ea567b): this is the option of using forked (passthrough) Envoy ext_authz API to allow fine-turning the configuration for better user experience and API stability. It actually looks quite cumbersome and seems a maintenance problem to mirror all the fields, especially some are very complex and has a lot dependency.

Design doc: https://docs.google.com/document/d/1V4mCQCw7mlGp0zSQQXYoBdbKMDnkPOjeyUb85U07iSI/edit#
Issue: https://github.com/istio/istio/issues/27790


security/v1beta1/authorization.proto
```diff
@@ -181,6 +182,59 @@ import "type/v1beta1/selector.proto";
...
+//  namespace: istio-system
+// spec:
+//  selector:
+//    matchLabels:
+//      app: istio-ingressgateway
+//  action: EXTERNAL
+//  external:
+//    http:
+//      server: "http://ext-authz.foo.svc.cluster.local:8000"
+//      config:
```
rshriram, line 227: I recall from the design doc, the consensus was to not do this model (exposing ext authz config pieces, etc.) and instead do the former example?

security/v1beta1/authorization.proto
```diff
@@ -258,10 +312,58 @@ message AuthorizationPolicy {
...
+    // "http://localhost:8001/" for HTTP server along with the sidecar (inside the pod).
+    //
+    // At least one of the server and config must be set.
+    string server = 1;
+
+    // Supply a custom external authorization configuration if more fine-tuning is needed. If not specified, a default
+    // external authorization configuration will be used.
+    //
+    // At least one of the server and config must be set.
+    istio.security.v1beta1.ExtAuthZ config = 2;
```
rshriram, line 365: same as above. IMO we can definitely skip this.

security/v1beta1/authorization.proto
mandarjog: The additional pass thru config is needed here, whether schematized or not, right?

security/v1beta1/authorization.proto
mandarjog: Skip and replace by struct or just skip?

security/v1beta1/ext_authz.proto
```diff
@@ -0,0 +1,120 @@
...
+// $title: External Authorization
+// $description: Mirrored version of Envoy ext_authz filter v3 API.
+// $location: https://istio.io/docs/reference/config/security/ext-authz.html
+
+package istio.security.v1beta1;
+
+option go_package="istio.io/api/security/v1beta1";
+
+// This is a mirrored version of the subset of the Envoy ext_authz v3 API
+// (https://github.com/envoyproxy/envoy/blob/c318156496edc46c844822a3d1f107ee496fa449/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto#L27).
```
mandarjog, line 28: We must ensure that this proto matches precisely with the proto that istio-proxy ships with.

security/v1beta1/authorization.proto
kyessenov: I don't think we can skip this. We need to support this directly or via EnvoyFilter. Maybe EnvoyFilter needs better support for augmenting Istio APIs? This can be done by placing some hook here that's recognizable from EnvoyFilter match condition.

mandarjog: I don’t think we want to get deeper into EnvoyFilter api, this is a chance to make a clean break.

security/v1beta1/authorization.proto
yangminzhu: I think in the doc we're actually tending to use a forked API? It's one of the options and I hope PR make it more clear about what it looks like.

security/v1beta1/ext_authz.proto
yangminzhu: we can add tests in istio/istio to covert the forked ExtAuthz to the real ExtAuthz imported in `github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_authz/v3`, the test can make sure the istio/api version of ExtAuthz can always be converted to the corresponding Envoy ExtAuthz used in the istio repo.

security/v1beta1/authorization.proto
yangminzhu: @rshriram This field is optional. If the user prefers to use EnvoyFilter for the fine-tuning, they can still use EnvoyFilter but isn't all our goal to improve the user experience by not using EnvoyFilter in most cases?

just to make this clear, the handling of these fields will be:
0. Reject the config that neither `server` nor `config` field is set;
1. Create an ext_authz filter config from the `config` field, it doesn't matter if the `config` field is not set;
2. If the `server` field is specified, pilot uses it to figure out the correct endpoint configuration that should be added to the ext_authz filter config from step 2. This includes resolving the cluster name automatically, configuring the gRPC authority field override to workaround the `|` invalid character issue or any other necessary issues.
3. Some final validation of the ext_authz filter config, for example, make sure all required fields are populated.

security/v1beta1/authorization_policy.proto
```diff
@@ -181,6 +182,59 @@ import "type/v1beta1/selector.proto";
...
+//      server: "http://ext-authz.foo.svc.cluster.local:8000"
+//  rules:
+//  - to:
+//    - operation:
+//        # Conditionally trigger the check only when the path has prefix "/data/".
+//        paths: ["/data/*"]
+// ```
+//
+// The following authorization policy applies to ingress gateway to enable the external authorization for all HTTP
+// requests and fine-tune to set the check request timeout to 2 seconds and also append the X-foo header to it.
```
incfly, line 211: might worth documenting what's the default if we don't specify the fine tune?

security/v1beta1/authorization_policy.proto
```diff
@@ -258,10 +312,58 @@ message AuthorizationPolicy {
...
+  // At least one of http and tcp must be set.
+  Config tcp = 2;
+
+  // Config supplies detail configurations about where and how to talk to the external server.
+  message Config {
+    // Supply the full URL of the external server. If specified, Istio will use it to smartly populate the external
+    // authorization configuration about where and how to talk to the external server.
+    //
+    // Example values:
+    // "http://ext-authz.foo.svc.cluster.local:8000/" for HTTP server in the mesh.
```
incfly, line 353: whether that's mTLS or not is just as same as a normal request to the ext-authz.foo svc within the mesh, right?

security/v1beta1/authorization_policy.proto
```diff
@@ -258,10 +312,58 @@ message AuthorizationPolicy {
...
+    // "https://ext-authz.company.com/" for HTTP server out of mesh.
```
incfly, line 354: I think the http and https choice is independent of whether it's within or outside the mesh right? The example makes me think internal people should use "http"; outside use "https".

security/v1beta1/ext_authz.proto
incfly: What's the desired process to add/modify this mirror? When and how?

security/v1beta1/authorization_policy.proto
yangminzhu: yes, this should work out-of-box for either plaintext or mTLS

security/v1beta1/authorization_policy.proto
yangminzhu: technically it's independent but I think it's probably better to use http + mTLS (configured separately) within mesh and https out of mesh.

security/v1beta1/ext_authz.proto
yangminzhu: updated to not include config in the first version.

security/v1beta1/authorization.proto
yangminzhu: updated not to include config in the first version.

security/v1beta1/authorization.proto
yangminzhu: updated not to include config in the first version.

security/v1beta1/authorization_policy.proto
yangminzhu: I think it should be fine, we can link to the Envoy doc if needed. We will only set the endpoint and all other will be the default values.

rshriram (approved):
Thanks for incorporating all the feedback Yangmin. This is good from my side.

security/v1beta1/authorization_policy.proto
```diff
@@ -258,10 +311,110 @@ message AuthorizationPolicy {
...
+
+  // If specified, the external authorization will be applied to the TCP filter chain, otherwise the TCP filter chain
+  // will have no external authorization.
+  //
+  // At least one of http and tcp must be set.
+  TCPConfig tcp = 2;
+
+  // HTTPConfig supplies detail configurations of the HTTP filter chain.
+  message HTTPConfig {
+    // Supply the full URL of the external server that implements the Envoy ext_authz filter check request API.
```
incfly, line 350: Supplies? Same as below tcp one.

security/v1beta1/authorization_policy.proto
```diff
@@ -258,10 +311,110 @@ message AuthorizationPolicy {
     // $hide_from_docs
     // Audit policy decisions can be read by telemetry plugins using the function getAuditPolicy
     // defined [here](https://github.com/istio/proxy/blob/master/extensions/common/context.h).
+
+    // External delegates the authorization decision to an external server out of Envoy. An extra authorization request
+    // will be sent to the external server and the response will decide whether or not the original request should be
+    // allowed or denied.
+    //
+    // The EXTERNAL action is evaluated before the ALLOW and DENY action, implemented by the Envoy ext_authz filter, and
+    // is currently an experimental feature in Istio.
```
incfly, line 320: Crete new line from "and". and capitalize or use `**` to emphaisze the experimental?

security/v1beta1/authorization_policy.proto
```diff
@@ -258,10 +311,110 @@ message AuthorizationPolicy {
...
   Action action = 3;
+
+  // action_detail specifies details of the corresponding action.
+  oneof action_detail {
+    // Configures how to talk to the external server. Must be set only with EXTERNAL action.
+    External external = 4;
+  }
+}
+
+// External supplies detail configurations of the EXTERNAL action.
```
incfly, line 334: detailed

incfly, line 334: an

mandarjog (approved):
@louiscryan 
This partially schematizes the ext_auth proto, and ensures that the schematized part includes the most often used features. If a customer needs things beyond that, they *will* use envoyfilter API.
lgtm

mandarjog:
@yangminzhu in order to support EnvoyFilter API correctly, we need to make sure that the `name` used to insert this filter is predictable, otherwise we cannot "merge" with the correct filter. Also keep in mind that we can have multiple ext_authz filters.

PR #1680, opened 2020-10-07
Move WorkloadGroup/WorkloadEntry to core API group

williamaronli:
draft change for the ref https://github.com/istio/istio/issues/27682


howardjohn (changes requested):
Per my comment on https://github.com/istio/istio/issues/27682#issuecomment-702892718, I am strongly against this change as I think it is harmful to our users and developers

liminw:
John, please see my response to your comments on the same issue. We want to have API group accurately reflect the purpose of the CRD resources, so that customers will not be confused when they create ACL for different admins.

williamaronli:
@liminw @justinwei2 @smawson @liminw could we come to a consensus? whether currently the workloadgroup and workloadentry which is designed for the VM should be moved under a new folder **core**

PR #1682, opened 2020-10-08
Remove unneeded open API schema

howardjohn:
This was added a while back, I assume as an initial step to support CRD
schema. However, the CRD schema does not actually use these files.
Therefor, I think there are no users of these.


kyessenov:
I think it would be nice to use OpenAPI as the source of truth. Istio proxy extensions do not use protobuf as the schema, so having OpenAPI would allow alternative schema formats.

kyessenov:
Example where relying on protobuf-JSON bites users https://github.com/istio/istio/issues/27851#issuecomment-706298381
Protobuf uses a particular dialect of JSON that is hard to replicate with other parsers.

PR #1688, opened 2020-10-14
use Destination for external authz

yangminzhu:
This changes the `string server` field to `Destination service` for specifying the address of the external server.

Also changed to use `http`/`grpc` to specify the protocol (API) used to talk to the external authorization server. Technically this could be independent from the original user request protocol (HTTP or TCP) in the future. Today the existing ext_authz filter only supports gRPC API for TCP traffic. In other words the HTTP API is only supported with HTTP traffic.

However I cannot use the `istio.networking.v1beta1.Destination` directly in the authorization_policy.proto due to an error in the open API scheme generation (see more details in the comment in authorization_policy.proto).


security/v1beta1/authorization_policy.proto
```diff
@@ -345,19 +356,58 @@ message External {
   // At least one of http and tcp must be set.
   TCPConfig tcp = 2;
 
+  // $hide_from_docs
+  // The following Destination and PortSelector is copied from the virtual_service.proto, this is to workaround the error
```
yangminzhu, line 360: This error is returned by the `cue-gen`:
```
$ cue-gen -paths=common-protos -f=./cue.yaml
v1beta1 redeclared as imported package name
	previous declaration at /home/ymzhu/go/src/istio.io/api/security/v1beta1/authorization_policy_proto_gen.cue:233:2:
    /home/ymzhu/go/src/istio.io/api/security/v1beta1/authorization_policy_proto_gen.cue:234:2
PeerAuthentication.selector: reference "v1beta1" not found:
    /home/ymzhu/go/src/istio.io/api/security/v1beta1/peer_authentication_proto_gen.cue:109:13
RequestAuthentication.selector: reference "v1beta1" not found:
    /home/ymzhu/go/src/istio.io/api/security/v1beta1/request_authentication_proto_gen.cue:132:13
RequestAuthentication.jwtRules: reference "JWTRule" not found:
    /home/ymzhu/go/src/istio.io/api/security/v1beta1/request_authentication_proto_gen.cue:140:17
genoapi.go:337: Instance failed
```

@jasonwzm could you take a look? This seems a bug as the proboful compile without problems. Thanks

yangminzhu:
cc @louiscryan @rshriram

security/v1beta1/authorization_policy.proto
```diff
@@ -26,18 +26,20 @@ import "type/v1beta1/selector.proto";
...
-// evaluated first. The evaluation is determined by the following rules:
+// Authorization policy supports ALLOW, DENY and EXTERNAL actions for access control. When multiple policies are used on
+// the same workload at the same time, the evaluation order is external -> deny -> allow. The evaluation is determined
+// by the following rules:
 //
-// 1. If there are any DENY policies that match the request, deny the request.
-// 2. If there are no ALLOW policies for the workload, allow the request.
-// 3. If any of the ALLOW policies match the request, allow the request.
-// 4. Deny the request.
+// 1. If there are any EXTERNAL policies that match the request, trigger the external authorization request. The request
```
howardjohn, line 33: can it only be denied? or can it be explicitly allowed and exit early with "allow"?

security/v1beta1/authorization_policy.proto
```diff
@@ -199,9 +201,15 @@ import "type/v1beta1/selector.proto";
 //  action: EXTERNAL
 //  external:
 //    http:
-//      server: "grpc://ext-authz.foo.svc.cluster.local:9000"
+//      service:
```
howardjohn, line 204: "service" implies k8s `Service` to me which I don't think is accurate here?

security/v1beta1/authorization_policy.proto
```diff
@@ -345,19 +356,58 @@ message External {
...
-    //
-    // Examples:
-    // - "http://ext-authz.foo.svc.cluster.local:8000/" for HTTP server on port 8000 in the mesh.
-    // - "https://ext-authz.company.com/" for HTTPS server on port 80 out of mesh.
-    // - "grpc://ext-authz.foo.svc.cluster.local:9000" for gRPC server on port 9000 in the mesh.
-    // - "http://127.0.0.1:8001/" for local HTTP server on port 8001 along with the Envoy sidecar.
-    string server = 1;
+    // Specifies the external authorization service that implements the Envoy ext_authz filter check request API.
+    // The port will be used to decide the protocol (HTTP or gRPC) for talking to the external authorization service.
+    Destination service = 1;
```
howardjohn, line 410: should we clarify this must be in the service registry? So you cannot just put google.com without a ServiceEntry?

mandarjog: Previously  based on the config here, the controlplane would create the appropriate cluster definition and use it in configuration.
The new approach is more explicit. So either something is defined thru service entry or exists otherwise in the mesh. It is more consistent with other APIs but one extra step.

security/v1beta1/authorization_policy.proto
```diff
@@ -345,19 +356,58 @@ message External {
...
   TCPConfig tcp = 2;
 
+  // $hide_from_docs
+  // The following Destination and PortSelector is copied from the virtual_service.proto, this is to workaround the error
+  // "v1beta1 redeclared as imported package name" during the open API scheme generation when using "istio.networking.v1beta1.Destination".
+  // It seems the underlying toolchain is not happy when two imported packages ("istio.networking.v1beta1.Destination" and "type/v1beta1/selector.proto")
+  // ends with the same name ("v1beta1").
+
+  // Destination indicates the network addressable service to which the
+  // request/connection will be sent after processing a routing rule. The
```
mandarjog, line 366: This is copied verbatim from source, but does not make sense here.
There is no routing going on, it’s just used to match an existing cluster.

security/v1beta1/authorization_policy.proto
yangminzhu: yes, I think the use of `Destination` should make it more clear and explicit that the service should be in the service registry (see the comments inside the Destination).

security/v1beta1/authorization_policy.proto
yangminzhu: it can only be denied, but since we're using the same matching rule, if the user needs to allow and exit early, they can configure another ALLOW policy with the exact same matching rule as the EXTERNAL policy.

security/v1beta1/authorization_policy.proto
yangminzhu: thanks, updated the comment to adjust for the ext_authz use case here.

security/v1beta1/authorization_policy.proto
yangminzhu: hmm, it is supposed to mean service in the Istio service registry, maybe we have better name for this concept? I'm fine with any suggestions.

security/v1beta1/authorization_policy.proto
smawson: What benefit does the user get from us requiring them to create a ServiceEntry?

security/v1beta1/authorization_policy.proto
```diff
@@ -333,31 +344,68 @@ message AuthorizationPolicy {
...
+  // PortSelector specifies the number of a port to be used for sending the authorization check request.
+  message PortSelector {
+    // Valid port number
+    uint32 number = 1;
+  }
+
+  // HTTPConfig supplies detail configurations of the HTTP traffic.
+  message HTTPConfig {
+    // Specifies the external authorization service that implements the Envoy ext_authz filter check request API.
+    // The port will be used to decide the protocol (HTTP or gRPC) for talking to the external authorization service.
```
smawson, line 407: So the requirement is that they define the port on a service entry and indicate on the port which protocol it is? This seems overly convoluted to me. Why are we making these changes? It seems like we're just making the API worse for the user without any clear benefit.

yangminzhu: the requirement is actually almost the same, both the URL string and the Destination is just used to construct the cluster-name. By using the Destination, we're making it clear that the service is from the Istio service registry which should be populated from other sources like k8s service, explicit ServiceEntry, etc.

I make this change based on feedback from @louiscryan, the main benefit I believe is to be consistent in the API for selecting the target of a request. The VirtualService use this Destination to select where to send the request to after the routing, here we just borrow the same concept except we do not have routing.

The other benefit is that it allows to use the subset, and since the service is in the service registry ahead of time, it actually simplifies the implementation so that we do not need to create ad-hoc cluster, and making it easier to use DestinationRule and PeerAuthentication to enable mTLS if the external server is deployed in the mesh.

security/v1beta1/authorization_policy.proto
yangminzhu: In most cases they don't, a simple k8s service should be good enough. And the biggest benefit I believe is the consistency in the API, see my replies to your other comment.

security/v1beta1/authorization_policy.proto
smawson: I don't think this is quite equivalent to virtual service though, virtual service is for routing within the mesh, this is for configuring an external plugin. It is fine if we decide that all plugins must be configured in the service registry but we should make that clear to users and be consistent, and I'm not entirely convinced it is worth it.

To be clear I actually like the API (using Destination), I'm just pushing back on whether the user is required to create an entry in the service registry for any external authz server, or whether we can dynamically create those if needed. And to be honest since this is a platform admin thing and won't happen often it is probably OK for us to require them to add it to the registry, but we should get feedback from users and see if this ever becomes an issue. It seems easy enough to fix later anyway.

mandarjog: Ext authz inside mesh may not be a common case. In absence of this config how will we configure mtls?

security/v1beta1/authorization_policy.proto
lambdai: Can you clarify what is creating establishing the http request, maybe in envoy's term?

security/v1beta1/authorization_policy.proto
```diff
@@ -399,21 +450,13 @@ message External {
...
-    // Only gRPC protocol is supported on TCP filter chain.
-    //
-    // Examples:
-    // - "grpc://ext-authz.foo.svc.cluster.local:9000" for gRPC server on port 9000 in the mesh.
-    // - "grpc://127.0.0.1:8001" for local gRPC server on port 8001 along with the Envoy sidecar.
-    string server = 1;
+  // GRPCConfig supplies detail configurations of the gRPC authorization check request.
+  message GRPCConfig {
+    // Specifies the external authorization service that implements the Envoy ext_authz filter gRPC check request API.
+    Destination service = 1;
```
lambdai, line 456: I suppose the subset is actually selecting the envoy cluster without specifying the name. But a `envoy cluster` is not compatible with  GoogleGrpc. Can you confirm?

kyessenov: This is a good point. ext_authz clusters cannot be dynamic, so they must be configured using google gRPC which is unaware of subsets and virtual services.

security/v1beta1/authorization_policy.proto
kyessenov: It cannot use mTLS. It has to be TLS due to Google gRPC.

security/v1beta1/authorization_policy.proto
```diff
@@ -333,31 +340,75 @@ message AuthorizationPolicy {
...
-  // If specified, the external authorization will be applied to the TCP filter chain, otherwise the TCP filter chain
-  // will have no external authorization.
+  // Specifies to use the gRPC authorization check API to talk to the external authorization server.
+  // Note, the gRPC authorization check API can be used with both HTTP and TCP (non-HTTP) traffic.
   //
-  // At least one of http and tcp must be set.
-  TCPConfig tcp = 2;
+  // At least one of http and grpc must be set.
+  // If only grpc field is set, both HTTP and TCP (non-HTTP) traffic will be enforced with the gRPC API.
+  // Normally only one of http and grpc field is good enough but if both fields are set, the HTTP traffic will be enforced
```
hzxuzhonghu, line 356: To make it easy, we can disallow setting both of them

security/v1beta1/authorization_policy.proto
```diff
@@ -333,31 +340,75 @@ message AuthorizationPolicy {
...
+  // with the HTTP API and the TCP (non-HTTP) traffic will be enforced with the gRPC API.
+  GRPCConfig grpc = 2;
+
+  // Specifies if the peer certificate is included in the authorization check request.
+  bool include_peer_certificate = 3;
```
hzxuzhonghu, line 361: What does this mean? Is there an example?

security/v1beta1/authorization_policy.proto
hzxuzhonghu: I could not understand why would we enforce the destination must be in the service registry?

yangminzhu: @lambdai , I'm not sure I understand your question, the following is a snippet of the ext_authz config generated from my local work code, we just use construct the name of the `cluster` field and put it in the ext_authz config:

```json
             {
              "name": "envoy.filters.http.ext_authz",
              "typed_config": {
               "@type": "type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz",
               "http_service": {
                "server_uri": {
                 "uri": "http://ext-authz-server.foo.svc.cluster.local",
                 "cluster": "outbound|8000||ext-authz-server.foo.svc.cluster.local",
                 "timeout": "1s"
                }
               },
```

@hzxuzhonghu that's the existing definition of the `Destination`: https://istio.io/latest/docs/reference/config/networking/virtual-service/#Destination

lambdai: @yangminzhu Thank you for the reference!

I can answer my question now. Write here in case any one has the same question.

As per the current envoy impl, the ext request is owned by cluster. The analogy is the the http request which carries the xds request.
These envoy cluster is a direct product of a destination rule. Nice reuse.

security/v1beta1/authorization_policy.proto
lambdai: > ext_authz clusters cannot be dynamic

Why ext_authz clusters cannot be dynamic?

yangminzhu: no, the subset is defined in DestinationRule, for example, the following AuthorizationPolicy config
```
  external:
    grpc:
      service:
        host: "ext-authz-server.foo.svc.cluster.local"
        subset: "v2"
        port:
          number: 9000
```
will construct the cluster `outbound|9000|v2|ext-authz-server.foo.svc.cluster.local`, note the subset is `v2` in this case, we're not making the ext_authz cluster dynamic, we just configure it to use different cluster that is already generated by Istio.

Also I don't understand what do you mean `a envoy cluster is not compatible with GoogleGrpc`? How is that related to the question here?

lambdai: > Also I don't understand what do you mean a envoy cluster is not compatible with GoogleGrpc? How is that related to the question here?

IIUC regardless how the config transformed, you end up with create a `GoogleGrpc` along with an Envoy `cluster`(most likely an EDS cluster). Normally the EDS cluster contains multiple addresses (10.X.Y.a:9000, 10.X.Y.b:9000, 10.X.Y.c:9000...), as well as the transport socket(tls or raw), etc.

As far as I know, in the current envoy code base, you cannot create a GoogleGrpc request/client with a destination as Envoy cluster. 
You can only pass GoogleGrpc a single URI(ext-authz-server.foo.svc.cluster.local:9000) or ip:port(10.X.Y.Z:9000) as the L4 address.

security/v1beta1/authorization_policy.proto
yangminzhu: @mandarjog  @kyessenov  I think it's actually not that uncommon to deploy ext_authz server in the mesh, there are many benefits of this, e.g. better performance, easier management/security and all the benefits from Istio itself (like the mTLS).

It can use mTLS. The user experience of enabling mTLS is the same as any other services, you can rely on autoMTLS or configure the DestinationRule and PeerAuthentication to enable it explicitly. 

The following is a snippet of the ext_authz config generated from my local work code that uses (Envoy) gRPC for the check request, and mTLS is working just as expected:
```json
             {
              "name": "envoy.filters.http.ext_authz",
              "typed_config": {
               "@type": "type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz",
               "grpc_service": {
                "envoy_grpc": {
                 "cluster_name": "outbound|9000||ext-authz-server.foo.svc.cluster.local",
                 "authority": "outbound-9000--ext-authz-server.foo.svc.cluster.local"
                },
                "timeout": "1s"
               },
```
The `"outbound|9000||ext-authz-server.foo.svc.cluster.local"` cluster config is generated completely independent from AuthorizationPolicy:
```json
{
     "version_info": "2020-10-15T04:03:55Z/6",
     "cluster": {
      "@type": "type.googleapis.com/envoy.config.cluster.v3.Cluster",
      "name": "outbound|9000||ext-authz-server.foo.svc.cluster.local",
      "type": "EDS",
      "eds_cluster_config": {
       "eds_config": {
        "ads": {},
        "resource_api_version": "V3"
       },
       "service_name": "outbound|9000||ext-authz-server.foo.svc.cluster.local"
      },
      "connect_timeout": "10s",
      "circuit_breakers": {
       "thresholds": [
        {
         "max_connections": 4294967295,
         "max_pending_requests": 4294967295,
         "max_requests": 4294967295,
         "max_retries": 4294967295
        }
       ]
      },
      "http2_protocol_options": {
       "max_concurrent_streams": 1073741824
      },
      "filters": [
       {
        "name": "istio.metadata_exchange",
        "typed_config": {
         "@type": "type.googleapis.com/udpa.type.v1.TypedStruct",
         "type_url": "type.googleapis.com/envoy.tcp.metadataexchange.config.MetadataExchange",
         "value": {
          "protocol": "istio-peer-exchange"
         }
        }
       }
      ],
      "transport_socket_matches": [
       {
        "name": "tlsMode-istio",
        "match": {
         "tlsMode": "istio"
        },
        "transport_socket": {
         "name": "envoy.transport_sockets.tls",
         "typed_config": {
          "@type": "type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext",
          "common_tls_context": {
           "alpn_protocols": [
            "istio-peer-exchange",
            "istio",
            "h2"
           ],
           "tls_certificate_sds_secret_configs": [
            {
             "name": "default",
             "sds_config": {
              "api_config_source": {
               "api_type": "GRPC",
               "grpc_services": [
                {
                 "envoy_grpc": {
                  "cluster_name": "sds-grpc"
                 }
                }
               ],
               "transport_api_version": "V3"
              },
              "initial_fetch_timeout": "0s",
              "resource_api_version": "V3"
             }
            }
           ],
           "combined_validation_context": {
            "default_validation_context": {
             "match_subject_alt_names": [
              {
               "exact": "spiffe://cluster.local/ns/foo/sa/default"
              }
             ]
            },
            "validation_context_sds_secret_config": {
             "name": "ROOTCA",
             "sds_config": {
              "api_config_source": {
               "api_type": "GRPC",
               "grpc_services": [
                {
                 "envoy_grpc": {
                  "cluster_name": "sds-grpc"
                 }
                }
               ],
               "transport_api_version": "V3"
              },
              "initial_fetch_timeout": "0s",
              "resource_api_version": "V3"
             }
            }
           }
          },
          "sni": "outbound_.9000_._.ext-authz-server.foo.svc.cluster.local"
         }
        }
       },
       {
        "name": "tlsMode-disabled",
        "match": {},
        "transport_socket": {
         "name": "envoy.transport_sockets.raw_buffer"
        }
       }
      ]
     },
     "last_updated": "2020-10-15T04:07:13.179Z"
    }
   ]
  }
```

@smawson  


> To be clear I actually like the API (using Destination), I'm just pushing back on whether the user is required to create an entry in the service registry for any external authz server, or whether we can dynamically create those if needed.

I see, thanks for the detail explanation. I think we could create the entry if it's not found it in the service registry for both API style, the API style should be a blocking issue.

security/v1beta1/authorization_policy.proto
yangminzhu: it means the x509 certificate is included in the authorization check request from ext_authz to the external server. In the external server, user can then use the x509 certificate to do the necessary authN/Z, see https://github.com/envoyproxy/envoy/blob/8fb3cb86082b17144a80402f5367ae65f06083bd/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto#L124

security/v1beta1/authorization_policy.proto
yangminzhu: we could use oneof to enforce this but actually intentionally give the flexibility to allow setting both `http` and `grpc`.

security/v1beta1/authorization_policy.proto
yangminzhu: Thanks for the info, but I'm still not sure why this is related, the design proposal is to use envoy gRPC client, are there any reasons to use google gRPC client? In my test code the envoy gRPC is working perfectly with k8s service and Istio ServiceEntry.

security/v1beta1/authorization_policy.proto
hzxuzhonghu: yeah, but seems no such use case

security/v1beta1/authorization_policy.proto
```diff
@@ -198,18 +200,20 @@ import "type/v1beta1/selector.proto";
...
 //  action: EXTERNAL
 //  external:
-//    http:
-//      server: "grpc://ext-authz.foo.svc.cluster.local:9000"
-//    tcp:
-//      server: "grpc://ext-authz.foo.svc.cluster.local:9000"
+//    grpc:
+//      service:
+//        host: "ext-authz.foo.svc.cluster.local"
+//        port:
```
costinm, line 206: I am concerned about consistency with the rest of the Istio APIs. 
We specify discoveryAddress and caAddress using gRPC style - host:port
I think integrations with external loggers and tracers use the same. 
MCP sources are URL.

I don't think we have any other API that specifies and external address using host and port number/name.
We use that in our APIs because we need to represent multiple settings for each port.

security/v1beta1/authorization_policy.proto
```diff
@@ -224,7 +228,10 @@ import "type/v1beta1/selector.proto";
 //  action: EXTERNAL
 //  external:
 //    http:
-//      server: "http://ext-authz.foo.svc.cluster.local:8000"
+//      service:
+//        host: "ext-authz.foo.svc.cluster.local"
+//        port:
+//          number: 8080
 //      timeout: 2s
```
costinm, line 235: Shouldn't timeout be specified in the DestinationRule ( with the new default scheme, etc) ?

security/v1beta1/authorization_policy.proto
```diff
@@ -333,31 +340,75 @@ message AuthorizationPolicy {
...
+    // instead of "reviews.default.svc.cluster.local"), Istio will interpret
+    // the short name based on the namespace of the rule, not the service. A
+    // rule in the "default" namespace containing a host "reviews will be
+    // interpreted as "reviews.default.svc.cluster.local", irrespective of
+    // the actual namespace associated with the reviews service. To avoid
+    // potential misconfiguration, it is recommended to always use fully
+    // qualified domain names over short names.
+    string host = 1 [(google.api.field_behavior) = REQUIRED];
+
+    // The name of a subset within the service. Applicable only to services
```
costinm, line 391: Don't we have VirtualService API for this ? Why duplicate it here ? 

Can you describe the UX for using subsets ? Will they modify 1000 Authz rules whenever they roll a new version of the authz server ? Can we remove this from the API until there is a clear requirement and UX.

security/v1beta1/authorization_policy.proto
```diff
@@ -368,8 +419,8 @@ message External {
     // Settings used for controlling authorization response metadata.
     AuthorizationResponse authorization_response = 4;
 
-    // Specifies if the peer certificate is included in the external authorization request.
-    bool include_peer_certificate = 5;
+    // Sets a prefix to the value of authorization request header *Path*.
+    string path_prefix = 5;
 
     // AuthorizationRequest allows to customize the authorization request.
     message AuthorizationRequest {
```
costinm, line 426: I don't get this name: it seems to indicate the list of headers to forward to the authz server. 
Why is it called 'AuthorizationRequest' ? HeaderFilter or IncludeHeaders would make sense.

costinm: I.e. if something allows to customize the authorization request - the name is probably not AuthorizationRequest, but based on the customization. For example timeout is also a customization of the request - but it's called timeout :-)

security/v1beta1/authorization_policy.proto
```diff
@@ -368,8 +419,8 @@ message External {
     // Settings used for controlling authorization response metadata.
     AuthorizationResponse authorization_response = 4;
 
-    // Specifies if the peer certificate is included in the external authorization request.
-    bool include_peer_certificate = 5;
+    // Sets a prefix to the value of authorization request header *Path*.
+    string path_prefix = 5;
```
costinm, line 423: I don't get this. How would the path prefix be used ? Change the path of the actual gRPC/http call the the server ? The forwared path ?

costinm (commented):
Please add this to the networking WG agenda for the DestinationRule implications.
Also Env WG needs to review this - it changes the pattern used for integrating with remote services. I suspect UX will need to approve this too.

costinm (commented):
Also, critical: this makes changes to a beta API. It is not clear yet how to  indicate that the added fields are 'experimental' - and not long-term supported ( since we just added the implementation and API, and I suspect the criteria for beta is not met)

security/v1beta1/authorization_policy.proto
costinm: Same for the response actually - and I suspect the use case for forwarding response to upstream has nothing to do with authorization, but it's actually authentication.

Also have a feeling part of the request is really authn - not authz. Since Istio separates the 2 - it would be good to use clear names and have plenty of comments when mixing them.

costinm (changes requested):
@linsun - how does this fit with the separation of 'mesh operator' vs 'app developer' ? 
Wouldn't the address of the authz server be part of the mesh operator role ? 

It seems super dangerous to allow users with namespace permissions to forward requests to arbitrary external addresses - including cookies and other headers.

smawson:
Very good point Costin, we're doing something similar for IDPs for VMs (which IDPs are allowed is in mesh config, using an IDP is on an application resource).

So yeah we should probably separate out defining authz endpoints and setting up rules that use them.

howardjohn:
> It seems super dangerous to allow users with namespace permissions to forward requests to arbitrary external addresses - including cookies and other headers.

Can't they already do this with virtual service mirror and others?

security/v1beta1/authorization_policy.proto
lambdai: > Are there any reasons to use google gRPC client?

That's a good question. I don't have the answer :) Just surfacing it.

lambdai: @kyessenov emphasizes GoogleGrpc is a MUST in this thread.

kyessenov: GoogleGrpc is a must for most external service integrations in envoy (tracing, xDS, etc). The lifecycle is just too complicated to depend on envoy cluster warming. What happens to requests if the Envoy cluster does not warm?
GoogleGrpc does not support EDS so pretty much all Istio functions are not applicable, but it does support extended authentication schemes like token exchange.

security/v1beta1/authorization_policy.proto
```diff
@@ -224,7 +228,10 @@ import "type/v1beta1/selector.proto";
 //  action: EXTERNAL
 //  external:
 //    http:
```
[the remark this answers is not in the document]
yangminzhu: > How common is it to have multiple external authz servers ?
If a user has 1000 rules - does he have to repeat the address 1000 times, and if the server changes to replace those as well ? What if he has a different server for test and prod ? Seems like a lot of work for users - when the most common case will be a single external authorization server.

No, the `external` field is specified only once, it's defined out of the `rules` field, so it's the same for 1 or 1000 rules. If they want to change to use a test server, they just change one line.


> A much simpler solution is to specify the authz server address in MeshConfig ( like we do for all external integrations ),
so Istio configs can be used in a CI/CD system and deployed in multiple meshes.
It would EXTREMELY hard to use Istio with kustomize if the user has to patch his own external address on each rule, and very dangerous too, if a new rule is added and the kustomization is not updated - app will be broken.

This doesn't make sense, are we suddenly pushing all APIs to MeshConfig now? This is a feature controlled by Istio API which is tested perfectly in e2e tests. And why does this have anything to do with kustomize? please read the API first, the `external` field is out of `rules`, the user does NOT need to change every rules in the first place.


> I don't know any other Istio API where the config embeds deployment-specific URL.

This is the service address, how is it deployment specific? Isn't the same `Destination` used in VirtualService.

security/v1beta1/authorization_policy.proto
yangminzhu: > Don't we have VirtualService API for this ? Why duplicate it here ?

The `Destination` message is duplicated as we cannot refer to it directly by `istio.networking.v1beta1.Destination` due to some toolchain issues, see comment at line 364.


> Can you describe the UX for using subsets ? Will they modify 1000 Authz rules whenever they roll a new version of the authz server ? Can we remove this from the API until there is a clear requirement and UX.

If you have a new version (e.g. v2) of the external server, you can set the `subset` to v2 to test it without changing the host field. Again, you don't modify 1000 Authz rules, this part is completely out of the `rules` field.

security/v1beta1/authorization_policy.proto
yangminzhu: this is a by-product of using `Destination`, it changes the path of the call to the external authorization server.

security/v1beta1/authorization_policy.proto
yangminzhu: >  it seems to indicate the list of headers to forward to the authz server.
> Why is it called 'AuthorizationRequest' ? HeaderFilter or IncludeHeaders would make sense.

Yes, it is the list of headers forwarded to the authz server. It's called `AuthorizationRequest` to be consistent with the [envoy ext_authz API](https://github.com/envoyproxy/envoy/blob/master/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto#L212). All Istio users today are using EnvoyFilter to configure ext_authz filter, choose the same name makes it much easier for them to use the new API and convert their existing config, also the comment is pretty clear of this field.

> Same for the response actually - and I suspect the use case for forwarding response to upstream has nothing to do with authorization, but it's actually authentication.
> Also have a feeling part of the request is really authn - not authz. Since Istio separates the 2 - it would be good to use clear names and have plenty of comments when mixing them.

It's up to the user to decide how do they want to use the API, `forward_to_upstream` is just a simple field we expose to user to make their life much easier. I'm pretty sure that users are just going to use EnvoyFilter to configure it all the time If we don't.  The same argument goes to Envoy, the filter is called external authorization but just allows you to customize the headers.

security/v1beta1/authorization_policy.proto
yangminzhu: I don't think GoogleGrpc is a must, @kyessenov please let me know why you think so, thanks.

security/v1beta1/authorization_policy.proto
yangminzhu: there are some important differences that make the choice different:

`discoveryAddress` and `caAddress` is configuring Istio itself in the bootstrap stage and it makes sense to configure it with the raw address instead of using Istio API because you don't have Istio at this moment. The traffic to `discoveryAddress` and `caAddress` is control-plane traffic abd the traffic to the external server is data-plane traffic.

The external address is configuring the user deployed external service, which happens after Istio is installed. And you can use VirtualService/DestinationRule/PeerAuthentication/etc. to mange the traffic to the external server, you don't use the these API to control the traffic to pilot or CA.

And how is this so different from a simple URL (e.g. `grpc://ext-authz.foo.svc.cluster.local:9000`)? It's essentially a different way to represent the address and I think it is actually more consistent with other APIs, the way we refer to the service in VirtualService.

security/v1beta1/authorization_policy.proto
yangminzhu: You can still do that in DestinationRule, this is a bit special here because ext_authz filter itself also allows to configure the timeout because it has its own gRPC/HTTP client.

yangminzhu:
> Also, critical: this makes changes to a beta API. It is not clear yet how to indicate that the added fields are 'experimental' - and not long-term supported ( since we just added the implementation and API, and I suspect the criteria for beta is not met)

this is marked as experimental in API and later clearly in all the documentation (task, blog, concept, etc.). It is supported in long-term and the "experimental" is because it may have some breaking changes in the API later.

smawson:
@mandarjog Mandar do we have an istio-wide plan for how to configure external plugins? We should be consistent and have a design for this that gets wide agreement. In particular we need to cover:
- Where available plugin endpoints are defined
- How plugin endpoints are defined
- Authn/authz model for calling into plugin endpoints

I imagine we'll use this for things like rate limiting servers and other external filters. There should probably be a doc just on external filters.

security/v1beta1/authorization_policy.proto
kyessenov: They are two different transports with their own benefits. Google gRPC works better for Google services. Envoy gRPC is lighter, but has complex enough life cycle issues that Envoy has assertions disallowing its use for control plane calls. Why take away the choice?

security/v1beta1/authorization_policy.proto
kyessenov: How does this work with git ops? Shouldn't we wait until external service is up before accepting traffic? It does make sense to make it part of install then (just like Mixer).

yangminzhu:
> Very good point Costin, we're doing something similar for IDPs for VMs (which IDPs are allowed is in mesh config, using an IDP is on an application resource).
> 
> So yeah we should probably separate out defining authz endpoints and setting up rules that use them.

@costinm  @smawson  I think there are some critical use case differences that make the design choice different. The IDPs for VMs is considered part of the Istio Mesh bootstrap and control plane configuration, you basically cannot use the security API without it (no identity).  That's why you need it in MeshConfig, the name explains its use case.

However, the external authorization is a different use case, it can but doesn't have to exchange new token for the workload. The external server is deployed after Istio is fully installed, different workloads, namespaces and teams can have their own (different) external server and configuratio, and this can change frequently (not like the IDP for VM where it is not changed too often after bootstrap and usually not configured by users themselves). Putting it in MeshConfig or somewhere else would be a very hard user experience to fit these use cases.

I think the other example is RequestAuthenticaion for JWT validation, we just put the issuer URL in the API instead of in MeshConfig or any other resources.

security/v1beta1/authorization_policy.proto
kyessenov: @costinm The async client is not the same one as on data path. It exposes only a few options https://github.com/envoyproxy/envoy/blob/master/include/envoy/http/async_client.h#L182 (timeout, tracing, etc were motivated by Mixer calls).

security/v1beta1/authorization_policy.proto
yangminzhu: If the external service is down, the request will just be denied (fail-close).

Mixer is part of Istio, external server is not. Different user/team could have different external server for their workloads/namespaces, and the server itself could change pretty often, and all this happen after Istio is fully installed.

Do you have examples why this not work with git ops? The user configure and commit the yaml the same way like they do for any other APIs.

security/v1beta1/authorization_policy.proto
```diff
@@ -333,31 +340,75 @@ message AuthorizationPolicy {
...
-  // If specified, the external authorization will be applied to the HTTP filter chain, otherwise the HTTP filter chain
-  // will have no external authorization.
+  // Specifies to use the HTTP authorization check API to talk to the external authorization server.
+  // Note, the HTTP authorization check API can only be used with HTTP traffic.
   //
-  // At least one of http and tcp must be set.
+  // At least one of http and grpc must be set.
+  // If only http field is set, the HTTP traffic will be enforced with HTTP API and there will be no external authorization
+  // enforced for TCP traffic.
   HTTPConfig http = 1;
```
liminw, line 349: @yangminzhu It looks like HTTPConfig is a superset of GRPCConfig (i.e., GRPCConfig plus a few other fields), can we use a single config and mark some fields optional (or only applicable to http protocol)? This way, customers do not have to explicitly specify "http" vs "gRPC", but our code can find out based on the actual fields set in the policy.

security/v1beta1/authorization_policy.proto
```diff
@@ -26,18 +26,20 @@ import "type/v1beta1/selector.proto";
 
 // Istio Authorization Policy enables access control on workloads in the mesh.
 //
-// Authorization policy supports both allow and deny policies. When allow and
-// deny policies are used for a workload at the same time, the deny policies are
-// evaluated first. The evaluation is determined by the following rules:
+// Authorization policy supports ALLOW, DENY and EXTERNAL actions for access control. When multiple policies are used on
+// the same workload at the same time, the evaluation order is external -> deny -> allow. The evaluation is determined
```
liminw, line 30: Is there a use case that we want to enforce "deny", "allow" before sending out request to external authz server?

security/v1beta1/authorization_policy.proto
yangminzhu: Thanks Kuat that's very useful information.

> What happens to requests if the Envoy cluster does not warm?

It's fail-close by default.

> Envoy gRPC is lighter, but has complex enough life cycle issues that Envoy has assertions disallowing its use for control plane calls. Why take away the choice?

There is an important difference, the external server is out of Istio control plane. You apply the API to enforce the external action after the pilot/envoy is full up to service, how would the life cycle issues affect this use case? The Envoy gRPC is used in the data plane request, from the life cycle perspective, isn't it the same for the external authorization server integration?

Also reading the envoy doc (https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_protocols/grpc.html#grpc-services) it says `It is recommended to use the Envoy gRPC client in most cases, where the advanced features in the Google C++ gRPC client are not required.`.

And many users are using the Envoy gRPC with ext_authz (via EnvoyFilter) just fine, they seem not having issues with that.

kyessenov: I think you would hit some issues if you delete the upstream cluster, for example, after applying ext_authz. EnvoyFilter is not a supported API, but this will be supported forever, so it needs to be more robust than that.

security/v1beta1/authorization_policy.proto
kyessenov: The readiness for istio installation needs to take into account availability of ext_authz. We cannot serve traffic and then randomly deny requests.

yangminzhu:
> @linsun - how does this fit with the separation of 'mesh operator' vs 'app developer' ?
> Wouldn't the address of the authz server be part of the mesh operator role ?
> 

This should completely be the 'app developer' responsibility to configure, the external authorization is no different from JWT validation and normal authorization control. Each team/user may have their own external server and address to use. 

> It seems super dangerous to allow users with namespace permissions to forward requests to arbitrary external addresses - including cookies and other headers.

This doesn't make sense, it's the same dangerous to configure a wrong authZ policy to accidentally open some access to the workload. If the user has namespace permission, they're the owner of the namespace already, they can already use VirtualService or even EnvoyFilter to configure whatever they want.

security/v1beta1/authorization_policy.proto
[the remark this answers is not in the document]
yangminzhu: I don't think so, we already provide the matching rules, if the user needs to bypass the external action, they can configure the matching rule to do it.

security/v1beta1/authorization_policy.proto
yangminzhu: it's superset today but the GRPCConfig may be extended to include many GRPC-only fields in the future, the other reason we have both `http` and `grpc` is that the `http` can only be used with HTTP traffic while `grpc` can be used with both HTTP and TCP traffic, there may be a case we want to use both, but if not, we could make the `http` and `grpc` in a oneof field.

security/v1beta1/authorization_policy.proto
yangminzhu: I don't think so, if they need this they can just configure the matching rule to achieve the same result.

security/v1beta1/authorization_policy.proto
yangminzhu: Do you mean delete by user? (like removing the k8s service), this is just fail-close or do you mean it could cause issues other than fail-close? the user should remove their ext_authz from the AuthorizationPolicy.

My point is that if the Envoy gRPC is good enough for data plane usage, it should be robust enough for the usage here.

security/v1beta1/authorization_policy.proto
liminw: Do you see a use case that need both http and grpc? AuthorizationPolicy so far does not require uses to specify protocols. We have some fields applicable to http only.

yangminzhu: it's the protocol between Envoy ext_authz and the external server, the user must specify this, either here or via some other fields. We need to tell ext_authz what protocol (and thus different API and request/response message) to use when talking to the external server.

It's different from the user request protocol. For user request of HTTP, the user can choose between `http` and `grpc` when talking to the ext_authz server, but for user request of TCP, the user can only use the `grpc`.

smawson:
> > Very good point Costin, we're doing something similar for IDPs for VMs (which IDPs are allowed is in mesh config, using an IDP is on an application resource).
> > So yeah we should probably separate out defining authz endpoints and setting up rules that use them.
> 
> @costinm @smawson I think there are some critical use case differences that make the design choice different. The IDPs for VMs is considered part of the Istio Mesh bootstrap and control plane configuration, you basically cannot use the security API without it (no identity). That's why you need it in MeshConfig, the name explains its use case.
> 
> However, the external authorization is a different use case, it can but doesn't have to exchange new token for the workload. The external server is deployed after Istio is fully installed, different workloads, namespaces and teams can have their own (different) external server and configuratio, and this can change frequently (not like the IDP for VM where it is not changed too often after bootstrap and usually not configured by users themselves). Putting it in MeshConfig or somewhere else would be a very hard user experience to fit these use cases.


That's not really true, we had actually considered putting it in the namespace (Limin's initial design had it that way) and we explicitly moved it out of the namespace because of security concerns. So it is a very real example of us deciding to separate the roles. We did the separation for both security reasons and for usability reasons. I think for external authz there are also both security and usability concerns with having the namespace user responsible for the entire configuration.


> I think the other example is RequestAuthenticaion for JWT validation, we just put the issuer URL in the API instead of in MeshConfig or any other resources.

We considered putting that in MeshConfig or a global resource as well, but yeah we decided not to separate it and allow namespace owners to control issuers. But I also remember us discussing in depth whether we needed to separate the rule part from the issuer part through indirection, to allow for this type of separation of concern. I don't quite recall why we decided we didn't need separation, I think part of it was that this is specifying authentication and not authorization, so its OK to have extra things apply.

The same logic does not apply to authorization, it is not OK to make authorization policy deny more things than desired.

So I think you're going to need to convince TOC that:
1) There are no security concerns with namespace admins controlling the destination of external authz requests (I think the argument that they could do this themselves anyway is probably valid).
2) There are no usability concerns with each namespace admin repeating the full configuration if for example you needed it for half of your namespaces.

security/v1beta1/authorization_policy.proto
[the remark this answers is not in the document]
liminw: Another question. If I have a mesh wide AuthorizationPolicy (defined in root namespace) and a namespace level AuthorizationPolicy, both define "EXTERNAL" action, but based on different matching rules. What will be the behavior? Will the external authz server be called twice? (But I probably want to have a global policy and have namespace override instead).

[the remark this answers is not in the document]
yangminzhu: both EXTERNAL actions will be evaluated and enforced independently from each other. it's the OR-ed semantic just like the DENY or JWT policy. The request is denied if any of the EXTERNAL/DENY or JWT policy returns deny.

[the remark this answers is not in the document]
liminw: The other way to phrase it is, a request is allowed if all the EXTERNAL actions allow it. If we interpret it this way, it is AND-ed. We cannot have two JWT policies on the same workload, one will override the other. It is the same case here.

yangminzhu:
Closing in favor of https://github.com/istio/api/pull/1698.

PR #1692, opened 2020-10-15
Add observedGeneration to status

therealmitchconnors:
This makes the Istio status field much more usable...


howardjohn:
Copying discussion from offline chat:
Concerns:
* Status may be written by many people. observedGeneration is a "global" - what does it mean when this is set and there are multiple parts of the styatus
* This means we need to update status on every write to the spec. We currently will do this for propogation anyways, but not for other status types
* There is *some* prior art, but not ubiqitous. See https://github.com/jetstack/cert-manager/issues/1551 for example.

smawson:
Is there a link to a doc or issue or something for this? Any changes to an API must include a link to something describing why we're making the change.

therealmitchconnors:
Marking this WIP, until user-reported issue can be linked.

therealmitchconnors:
This PR is the first step in fixing https://github.com/istio/istio/issues/28003

smawson (commented):
I think for the first, if the meaning of this field is the generation that the status refers to, then controllers could guarantee they only ever increase it, never decrease it. Or we could put it in IstioCondition instead of in IstioStatus and it would indicate the generation of the resource that led to that condition, but that doesn't match with how observedGeneration is used elsewhere.

On the second it is probably fine to update the status after each write to the spec (which is when generation changes). If the spec changes we need to propagate that change so it seems right.

For prior art, I went and looked at the kubernetes api reference, there are 15-ish resources with observedGeneration, and they are all used by the controller operating on that resource to indicate the generation that the controller has completed operating for.

If we think of Istio as a controller operating on config resources then this matches and we can use the same model (what is in this PR). We need to fix the comment though (looks like someone copy-pasted from replication controller).

So yeah I'm OK with this I think although we should probably get a few more eyes on it since it is pervasive. But it does match how k8s uses generation and observedGeneration so seems OK.

howardjohn:
Just to clarify on my concerns - I am not opposed to this, just a bit undecided and wanted to make sure we thought about them.

meta/v1alpha1/status.proto
```diff
@@ -40,6 +40,10 @@ message IstioStatus {
   // +patchMergeKey=type
   // +patchStrategy=merge
   repeated analysis.v1alpha1.AnalysisMessageBase validation_messages = 2;
+
+  // Resource Generation to which this status refers.
```
smawson, line 44: Can we put in something about how if this is the same as generation in metadata then status is up to date, if not it means status is still being calculated for the latest generation?

smawson:
What's the next step to address your concerns? I think I'm OK with this as-is, modulo having a better comment, but we'll need to make sure the design gets appropriate reviews to make sure it covers your concerns. Is that good enough or do we need that design before submitting the API change?

> Just to clarify on my concerns - I am not opposed to this, just a bit undecided and wanted to make sure we thought about them.

howardjohn:
Thinking a bit more, the same problems all apply to other usages. Seems they get around it by this comment:
```
   observedGeneration   <integer>
     The generation observed by the deployment controller.
```

So if we similarly scope this to the "config distribution", we are in the clear. The issue that may arise then is if we need similar field for the other sources of status...

smawson (commented):
So maybe update comment to refer to which controller is filling this out?

I think if for some reason multiple controllers needed to update this status we could add a different field for the other controller. That seems theoretical enough that I'm comfortable considering config distribution to be the default controller for config resources.

smawson (approved):

meta/v1alpha1/status.proto
```diff
@@ -40,6 +40,12 @@ message IstioStatus {
   // +patchMergeKey=type
   // +patchStrategy=merge
   repeated analysis.v1alpha1.AnalysisMessageBase validation_messages = 2;
+
+  // Resource Generation to which the Reconciled Condition refers.
+  // When this value is not equal to the object's metadata generation, reconciled condition  calculation for the current
+  // generation is still in progress.
+  // +optional
```
nrjpoddar, line 47: I might be missing context here but this description is not very useful for an end user. I don't know what Reconciled Condition means or who is reconciling and why.

therealmitchconnors: The Reconciled Condition is a part of our api, and is documented here: https://istio.io/latest/docs/reference/config/config-status/

nrjpoddar: I see, I understand the field only because I know this feature well. Can you add more description or link out for folks to understand what this field means and how they should use it?

Naming bike shedding:

Is reconciled_generation a better name for this field?

therealmitchconnors: I'm adding a docs reference now.  The name is consistent with many other implementations, and I think we can get to the point where the field applies for the whole status object.

meta/v1alpha1/status.proto
```diff
@@ -43,7 +43,7 @@ message IstioStatus {
 
   // Resource Generation to which the Reconciled Condition refers.
   // When this value is not equal to the object's metadata generation, reconciled condition  calculation for the current
-  // generation is still in progress.
+  // generation is still in progress.  See https://istio.io/latest/docs/reference/config/config-status/ for more info.
```
nrjpoddar, line 46: I think you need to use relative paths for docs referenced from istio.io

nrjpoddar (approved):
Nit about reference link format but everything else LGTM

PR #1696, opened 2020-10-16
Fix comment/doc of proxy stats matcher

bianpengyuan:


mandarjog (approved):
changing key naming conventions only in doc.

smawson (approved):

PR #1698, opened 2020-10-16
update CUSTOM action in AuthorizationPolicy

yangminzhu:
Updated to move the config to MeshConfig based on discussions from the TOC meeting.

Design doc: https://docs.google.com/document/d/1V4mCQCw7mlGp0zSQQXYoBdbKMDnkPOjeyUb85U07iSI/edit#  (join [istio-team-drive](https://groups.google.com/forum/#!forum/istio-team-drive-access) to gain access)


security/v1beta1/authorization_policy.proto
```diff
@@ -312,109 +280,19 @@ message AuthorizationPolicy {
...
     // defined [here](https://github.com/istio/proxy/blob/master/extensions/common/context.h).
 
+    // $hide_from_docs
     // External delegates the authorization decision to an external server out of Envoy. An extra authorization request
     // will be sent to the external server and the response will decide whether or not the original request should be
     // allowed or denied.
     //
     // The EXTERNAL action is evaluated before the ALLOW and DENY action, implemented by the Envoy ext_authz filter
     // and is currently an **experimental feature** in Istio.
+    // The configuration of the external server is specified in the `external_auth` field in the MeshConfig.
```
louiscryan, line 290: @smawson @costinm @nrjpoddar @howardjohn 

To John's point in the TOC meeting this a.m I wonder if we want to make a strong statement about logical de-coupling from implementation here. In theory this can be CUSTOM/EXTERNAL and defined to be a binding to a default or named authz system declared in mesh-config without implying anything else in the contract. This would enable it to be generic and accommodate use-cases like WASM. Then the 'ext_authz' filter is just an example rather than a concrete part of this contract.

mandarjog: I think from the perspective of user journey we need to state "what" it is by pointing to mesh config.
This is identical to tracer. Here we are saying when to trigger tracing and in some other place we say *how* to trace.
We need to specify where that other place is.

With that in mind the mesh config specified here should change. It *should* concrete, and some enum should say that it is `ext_authz` and not just text.

yangminzhu: @mandarjog I wonder should we do that when we actually need that flexibility? That means add this when a different place other than MeshConfig and a different implementation other than ext_authz is actually used. 
The API can be added in a later version without breaking existing user by having proper default values. Before that we are not sure what the concrete API would look like, we could try to make the API as generic as possible but it still seems would be diverged from the real user case arises, and it's much harder to fix that if we added the generic API too early.

mandarjog: Since we are actively doing this indirection to support “more than one” now would be good.

But sure we can introduce enum later and deprecate the current field in mesh config.

yangminzhu: In MeshConfig, I pushed a commit to make the API more generic. It is a list of external server identified by a unique name (only 1 is allowed for now) and the detail configuration is in its own struct organised under the oneof (can easily add other type of external server).

My only concern is that will this be too early to say that the AuthzPolicy is the integration point for things like WASM?

In AuthorizationPolicy, I'm not sure do we want to just make the comment more generic, or add the fields to allow configuring the name of the external server (i.e. state "what" it is), and one more field to state "where" it is (currently only MeshConfig)? Like the following:
```yaml
kind: AuthorizationPolicy
metadata:
  name: ext-auth
  namespace: istio-system
spec:
  selector:
    matchLabels:
      app: istio-ingressgateway
  action: EXTERNAL
  external:
    server: "default" # the unique name of the server (defined in MeshConfig)
    source: MESH_CONFIG # the place of the definition, this is an enum and can be omitted for now as the default is MESH_CONFIG
```

mesh/v1alpha1/config.proto
```diff
@@ -499,7 +500,74 @@ message MeshConfig {
...
+        // Sets a prefix to the value of authorization request header *Path*.
+        string path_prefix = 1;
+
+        // List of headers that should be included in the authorization check request to the external auth server.
+        // Note that in addition to the user's supplied headers:
+        // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically included to the list.
+        // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+        repeated string include_headers = 2;
+
+        // List of headers from the external server that should be added to the original request and forwarded to the
```
liminw, line 533: Is it "add" only? I think external_authz filter spec also allows you to remove/replace headers in addition to "add".

mesh/v1alpha1/config.proto
```diff
@@ -499,7 +500,74 @@ message MeshConfig {
...
+        // If not specified, the original request will not be modified and forwarded to backend as-is.
+        // Note, any existing headers will be overridden.
+        repeated string forward_to_upstream = 3;
+
+        // List of headers from the external server that should be forwarded to downstream when the authorization check result
+        // is not allowed (HTTP code other than 200).
+        // If not specified, all the authorization response headers, except *Authority (Host)* will be in the response to
+        // the downstream. When a header is included in this list, *Path*, *Status*, *Content-Length*, *WWWAuthenticate*
+        // and *Location* are automatically added.
+        repeated string forward_to_downstream = 4;
```
liminw, line 544: How about call it "headers_on_deny"?

mesh/v1alpha1/config.proto
```diff
@@ -499,7 +500,74 @@ message MeshConfig {
...
+        // Note that in addition to the user's supplied headers:
+        // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically included to the list.
+        // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+        repeated string include_headers = 2;
+
+        // List of headers from the external server that should be added to the original request and forwarded to the
+        // upstream when the authorization check result is allowed (HTTP code 200).
+        // If not specified, the original request will not be modified and forwarded to backend as-is.
+        // Note, any existing headers will be overridden.
+        repeated string forward_to_upstream = 3;
```
liminw, line 537: Do we want to call it "headers_on_allow"? I think it is better to call out this is "headers".

mesh/v1alpha1/config.proto
```diff
@@ -499,7 +500,74 @@ message MeshConfig {
...
+      message GRPC {}
+
+      // REQUIRED. Specifies the protocol used to talk to the external authorization server.
+      oneof external_authz_service_type {
+        // Send the authorization check request to the external server using HTTP protocol.
+        HTTP http = 3;
+
+        // Send the authorization check request to the external server using gRPC protocol as defined by
+        // [Envoy ext_authz check request API](https://github.com/envoyproxy/envoy/blob/9dce187eca9aa10e98c0ae81368c078384528801/api/envoy/service/auth/v3/external_auth.proto#L29-L34).
+        GRPC grpc = 4;
```
liminw, line 557: Do you still need users to configure "http" vs "grpc"? I think the feedback was that this should be implementation details.

mesh/v1alpha1/config.proto
```diff
@@ -499,7 +500,74 @@ message MeshConfig {
...
 
   // $hide_from_docs
-  // Next available field number: 56
+  message ExternalServer {
+    // REQUIRED. A unique name identifying the external server.
+    string name = 1;
+
+    // REQUIRED. Specifies detail configurations of the external server.
+    oneof config {
+      Authz authz = 2;
```
liminw, line 509: Calling it authz is too limited. In the use cases of RC token, group resolver, do we still use "authz" type? Also, in API, it is better to use the full name like "Authorization" instead of "authz".

mesh/v1alpha1/config.proto
smawson: I think it is OK, this is the platform admin configuring what type of external authz service this is. I like having that explicit instead of implicit based on ports or whatever.

mesh/v1alpha1/config.proto
```diff
@@ -499,7 +500,74 @@ message MeshConfig {
...
+      }
+    }
+  }
+
+  // $hide_from_docs
+  // Defines a list of external servers that provide extra functionality to the mesh. The external servers are usually
+  // deployed and managed by the mesh administrator to extend Istio's functionality, for example, an external authorization
+  // server is used with the AuthorizationPolicy to integrate custom authorization system into Istio.
+  // Note, at most 1 external server is allowed for now.
+  repeated ExternalServer external_servers = 56;
```
smawson, line 567: We may want to pick a name other than "external" for this, because it is external to the envoy proxy, not external to the mesh. Or maybe we should make that clear in the comments that these can be inside the mesh and don't need to be external to it.

I'd actually prefer something like "extension_servers" and call the message "ExtensionServer." Actually maybe service is better: ExtensionService and extension_services.

Inside authz proto we can still call it external because it is an external authz call, which goes to one of the extension servers.

security/v1beta1/authorization_policy.proto
smawson: I think this should just be a named external service, so we should include a field to indicate which external service to call (there may be more than one defined).

Then we can just require unique names if we expand the set of types, or we can introduce a type field here to distinguish later.

mesh/v1alpha1/config.proto
yangminzhu: I think this is a bit different, the http/grpc here specifies the API used to talk to the external server

1. the feedback is that we shouldn't distinguish the user traffic (tcp v.s. http) which is orthogonal to the API used to talk to the external server, this means each of the following combination should be possible:
user traffic HTTP + external server HTTP API
user traffic HTTP + external server gRPC API
user traffic TCP + external server HTTP API (not supported today)
user traffic TCP + external server gRPC API

2. the configuration for HTTP and gRPC API is different, the `forward_to_upstream`, `forward_to_downstream` are only applicable to HTTP API, in the future, there may be more differences in the two API

mesh/v1alpha1/config.proto
yangminzhu: for HTTP API (the protocol used to talk to external server, not user traffic), I don't see it supports remove, doc here: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto.html#extensions-filters-http-ext-authz-v3-authorizationresponse.

It will override (replace) if the header already exists.

mesh/v1alpha1/config.proto
yangminzhu: Or in other words, we will always need this `http`/`grpc` field to supply the detail configuration of the check request/response.

mesh/v1alpha1/config.proto
```diff
@@ -499,7 +500,74 @@ message MeshConfig {
...
+      // Note, the `service.port` must always be configured explicitly even if the service exposes only a single port.
+      istio.networking.v1alpha3.Destination service = 1;
+
+      // Sets the maximum duration in milliseconds for connection to the external server (default is 1s).
+      google.protobuf.Duration timeout = 2;
+
+      // $hide_from_docs
+      message HTTP {
+        // Sets a prefix to the value of authorization request header *Path*.
+        string path_prefix = 1;
```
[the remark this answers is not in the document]
yangminzhu: Updated comments, this is to append an extra path prefix when sending the authorization check request to the external server.

mesh/v1alpha1/config.proto
yangminzhu: updated to `headers_to_downstream_on_deny`

mesh/v1alpha1/config.proto
yangminzhu: updated to `headers_to_upstream_on_allow`

mesh/v1alpha1/config.proto
yangminzhu: Updated to use the full name, I think we can just use the `authorization` type for RC token or group resolver, this is the specific extension service implementation but not limited to the high-level use case.

mesh/v1alpha1/config.proto
yangminzhu: Thanks, updated to use ExtensionService and extension_services.

security/v1beta1/authorization_policy.proto
yangminzhu: Thanks Sven, updated and the example now looks like the following:
```yaml
kind: AuthorizationPolicy
metadata:
  name: ext-auth
  namespace: istio-system
spec:
  selector:
    matchLabels:
      app: istio-ingressgateway
  action: EXTERNAL
  external:
    service: "default" # the unique name of the external service to call (defined in MeshConfig)
```

yangminzhu:
@liminw @smawson @mandarjog @louiscryan @costinm @howardjohn @nrjpoddar I updated the PR to address the comments so far, can we take another look? thanks.

security/v1beta1/authorization_policy.proto
louiscryan: @smawson 

How about 'provider' instead of 'service'? 

Also do we need to nest ?

action: EXTERNAL
provider: "reference to named thing somewhere"

Do we expect a need to parameterize these 'calls' beyond what resides in some centralized config?

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
   CA ca = 55;
 
   // $hide_from_docs
-  // Next available field number: 57
+  message ExtensionService {
```
louiscryan, line 509: Better name? 'ExternalAuthorizer' ?

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+    // REQUIRED. A unique name identifying the extension service.
+    string name = 1;
+
+    // REQUIRED. Specifies detail configurations of the extension service.
```
louiscryan, line 513: Specifies detailed configuration for the external authorizer

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+    oneof config {
+      Authorization authorization = 2;
+    }
+
+    // $hide_from_docs
+    // Authorization supplies the detail configuration of the external authorization server.
+    message Authorization {
```
louiscryan, line 520: Better name? This proto represents an 'external' authorizer implementation based on ext_authz filter in Envoy. Do we consider this feature a standard or a part of Envoy or something else?

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+    // REQUIRED. A unique name identifying the extension service.
+    string name = 1;
+
+    // REQUIRED. Specifies detail configurations of the extension service.
+    oneof config {
+      Authorization authorization = 2;
+    }
+
+    // $hide_from_docs
+    // Authorization supplies the detail configuration of the external authorization server.
```
louiscryan, line 519: Almost duplicate with 513 above. If this is 'oneof' the possible authorizers then you should document it concretely rather than abstractly. E.g.

'Configures an Envoy ext_authz filter as an external authorizer"

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+    message Authorization {
+      // REQUIRED. Specifies the network addressable service to which the authorization check request should be sent.
+      // Note, the `service.port` must always be configured explicitly even if the service exposes only a single port.
```
louiscryan, line 522: Why? To avoid hijacking? Is this a security concern.

What is the service has a HTTP port and a GRPC one. Which would we pick?

security/v1beta1/authorization_policy.proto
smawson: Provider is fine. I think we can add things like parameterization later if we leave the external block in there.

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+      google.protobuf.Duration timeout = 2;
+
+      // $hide_from_docs
+      message HTTP {
+        // Sets a prefix to the value of authorization request header *Path*.
+        // For example, setting this to "/check" for an original user request at path "/admin" will cause the
+        // authorization check request to be sent to the external server at the path "/check/admin" instead of "/admin".
+        string path_prefix = 1;
+
+        // List of headers that should be included in the authorization check request to the external auth server.
```
louiscryan, line 535: s/authorization check request to the external auth server./authorization request sent to the external authorizer.

mesh/v1alpha1/config.proto
smawson: ExtensionProvider or ExternalProvider?

I think Yangmin's goal here was to make this a holder for multiple types of external providers (so we could add things like rate limiting providers or whatever here too). But we could also make it specialized and remove the oneof.

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+        // Note that in addition to the headers supplied by users:
+        // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically included to the list.
```
louiscryan, line 537: s/are automatically included to the list./are automatically sent

mesh/v1alpha1/config.proto
smawson: If specialized to type it could be:

message ExternalAuthorizer {
  string name = 1;
  Destination service = 2;
  Duration timeout = 3;
  oneof type (grpc | http)
}

That's probably the simplest approach.

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+        // List of headers from the external server that should be added to the original request and forwarded to the
+        // upstream when the authorization check result is allowed (HTTP code 200).
+        // If not specified, the original request will not be modified and forwarded to backend as-is.
+        // Note, any existing headers will be overridden.
+        repeated string headers_to_upstream_on_allow = 3;
+
+        // List of headers from the external server that should be forwarded to downstream when the authorization check result
+        // is not allowed (HTTP code other than 200).
+        // If not specified, all the authorization response headers, except *Authority (Host)* will be in the response to
+        // the downstream. When a header is included in this list, *Path*, *Status*, *Content-Length*, *WWWAuthenticate*
```
louiscryan, line 550: Why would Path be in this list? What about a response payload ?

mesh/v1alpha1/config.proto
louiscryan: The more I read below of the Envoy quirks the more I think we should explicitly tie this to Envoy and also follow Envoy's versioning. 

@smawson AFAICT the only reason to use the Envoy type here in 'Any' like fashion is because we want to use Destination

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
smawson: So the API for this is to make an HTTP request with /prefix/<actual path>?

That is a pretty bizarre API, but OK that's what is in envoy authz api then that's what we're exposing.

smawson (commented):
I think I agree with Louis that we should make this more specific, and not try to build in abstraction/indirection through a wrapper type.

So just have a list of ExternalAuthorization objects in mesh config with names, and refer to them by name from AuthorizationPolicy. We can later add other types of extensions by just adding more lists into mesh config.

mesh/v1alpha1/config.proto
louiscryan: I agree there's not a lot of value in trying to make a wrapper message around a bunch of oneof's for different use-cases generic. OTOH we suspect that we would need more than one for each use-case which is why I didn't propose flattening entirely

yangminzhu: @louiscryan , this was to make it a bit more generic but I guess it is probably a bit over-designed at this moment. I will change it to just ExternalAuthorizer, and keep the oneof but renaming it to be entirely tied with Envoy (`envoy_ext_authz`) as your other comment suggests.

smawson: Well name, service, and timeout are all pretty generic, we just need to decide where to put the specific stuff. Here are a few different possible models:

1) Generic Service

    message ExternalService {
      name, service, timeout
      oneof type (GrpcExternalAuthorizationService | HttpExternalAuthorizationService)
    }

Extension is to just add to the oneof to introduce new types.

2) Specific Services

    message GrpcExternalAuthorizationService {
      name, service, timeout
    }

    message HttpExternalAuthorizationService {
      name, service, timeout, prefix, headers, ...
    }

Extension is to add new types.

3) Generic Named Thing

    message ExternalThing {
      string name
      oneof (gRPCExternalAuthorizationService | HttpExternalAuthorizationService)
    }

Extension is to just add to the oneof

4) Nesting Genericness

    message ExternalThing {
      string name
      oneof (ExternalAuthorizationService)
    }

    message ExternalAuthorizationService {
      destination, timeout
      oneof (gRPCExternalAuthorizationService | HttpExternalAuthorizationService)
    }

Given that copy-paste is cheap I would suggest we go with the fully specific one and just copy the name, timeout, duration into each object. Actually we should pull out name only since everything needs a name. Wrap the whole thing with a oneof, so something like:

    repeated ExternalService external_services = x;

    message ExternalService {
      string name = 1;
      oneof (ExternalHttpAuthorizationService, ExternalGrpcAuthorizationService)
    }

    message ExternalHttpAuthorizationService {
      timeout, duration, path, headers
    }

    message ExternalGrpcAuthorizationService {
      timeout, duration
    }


That's close to what is there now but we can just flatten the Authorization message down into the grpc and http oneof to simplify.

louiscryan:
+1 to Sven's suggested layout

mesh/v1alpha1/config.proto
yangminzhu: Will rename to make it more specific that this is ext_authz filter in Envoy.

I agree it is probably very unlikely that a different external authorization system would have an API compatible with the Envoy ext_authz filter, and it's hard (and probably impossible) to define an API generic enough that would be implemented by every external authorization system, so it's best to call out the current one is just the Envoy ext_authz API, others could be added in the future to the oneof.

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
yangminzhu: yes, that's the existing API in the Envoy ext_authz.

mesh/v1alpha1/config.proto
costinm, line 509: We already use RemoteService - any reason to invent a new proto, just for this case - where we use RemoteService for the other ?

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,76 @@ message MeshConfig {
...
+      }
+    }
+  }
+
+  // $hide_from_docs
+  // Defines a list of extension services that provide extra functionality to the mesh. The extension services are usually
+  // deployed and managed by the mesh administrator to extend Istio's functionality, for example, an external authorization
+  // service is used with the AuthorizationPolicy to integrate custom authorization system into Istio.
+  // Note, at most 1 extension service is allowed for now.
+  repeated ExtensionService extension_services = 57;
```
costinm, line 575: Please follow the same pattern use use for RemoteService - I don't think all 'extension services' in Istio will be of this type.

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,51 @@ message AuthorizationPolicy {
...
+    // kind: AuthorizationPolicy
+    // metadata:
+    //  name: ext-auth
+    //  namespace: istio-system
+    // spec:
+    //  selector:
+    //    matchLabels:
+    //      app: istio-ingressgateway
+    //  action: EXTERNAL
+    //  external:
```
costinm, line 285: Do we expect that all 'external services' we'll use for authz will use this particular interface and nothing else ? 

Should we use instead:

ext_authz: 
   service: "default" 

so it is clear what 'external service' we are talking about?  In Istio we also have other kinds of external services, like logging, tracing, etc. 

Unfortunately security is using some very broad and generic words without taking into account other uses.

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,51 @@ message AuthorizationPolicy {
...
+    // apiVersion: security.istio.io/v1beta1
+    // kind: AuthorizationPolicy
+    // metadata:
+    //  name: ext-auth
+    //  namespace: istio-system
+    // spec:
+    //  selector:
+    //    matchLabels:
+    //      app: istio-ingressgateway
+    //  action: EXTERNAL
```
costinm, line 284: Do we need both 'action:EXTERNAL' and the 'external: ' section ? Isn't it implicit that if you have external: {} we will have an external action ? I don't think user typing the same thing 2x just to confirm he means it will help.

mesh/v1alpha1/config.proto
yangminzhu: not for hijacking, previously I was just trying to make the configuration more consistent for service with single port or multiple ports and easier for us to generate the Envoy configuration in istiod so that in the authz code path we don't need to look up in the service registry to find the port number and can just use what the user provides.

In case there are multiple ports (HTTP and GRPC), the user will need to explicitly specify it here (same as the current behavior of the `Destination`).

mesh/v1alpha1/config.proto
yangminzhu: That is the current Envoy ext_authz behavior (doc: https://github.com/envoyproxy/envoy/blob/415af040e09a1f6993c15ffc022793d39ecf5e8e/api/envoy/extensions/filters/http/ext_authz/v3/ext_authz.proto#L253), I'm not sure about the reason for that. 
The response body (from the external authorizer) will be sent back to downstream, this is not clearly documented but I found it in the code.

mesh/v1alpha1/config.proto
yangminzhu: There are a couple of reasons we cannot reuse it, the RemoteService uses a `string address` that doesn't really fit the use case here, the service should be in Istio service registry, and can be controlled with DestinationRule for things like subset, which means we should not have the duplicate `ClientTLSSettings` because it should be configured by DestinationRule. also the `tcp_keepalive` doesn't really apply here (not supported by Envoy ext_authz)

Also different provider may have it's own requirement for specifying the address format and that is not bound to the ExternalProvider here.

mesh/v1alpha1/config.proto
yangminzhu: see latest PR and my replies [here ](https://github.com/istio/api/pull/1698#discussion_r508141631), I don't think we can or should use the same RemoteService for all different providers.

security/v1beta1/authorization_policy.proto
yangminzhu: the `action: EXTERNAL` is enum and we cannot change it without breaking backward compatibility.

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,74 @@ message MeshConfig {
   CA ca = 55;
 
   // $hide_from_docs
-  // Next available field number: 57
+  message ExternalProvider {
+    // REQUIRED. A unique name identifying the external provider.
+    string name = 1;
+
+    // REQUIRED. Specifies detailed configuration for the external provider.
+    oneof config {
```
smawson, line 514: nit: provider?

security/v1beta1/authorization_policy.proto
yangminzhu: there may be external authz interface other than the envoy ext_authz one. The concrete type is defined in the ExternalProvider in MeshConfig now, the comments also refer the user to check the MeshConfig to decide the type of the external provider.

smawson (commented):
LGTM, one small nit but otherwise I think it looks pretty clean. Happy to approve once Louis signs off.

mesh/v1alpha1/config.proto
yangminzhu: @smawson thanks for the very detailed suggestion, updated to use `ExternalProvider`, `ExternalHttpAuthorizationProvider` and `ExternalGrpcAuthorizationProvider`. I also added the service to the both HTTP and GRPC message and updated the name with "Envoy" prefix to make it super clear it's the concrete implementation of Envoy ext_authz filter.

mesh/v1alpha1/config.proto
yangminzhu: thanks, fixed.

yangminzhu:
@louiscryan mind to take a look for another round? I have updated PR to address all comments so far, thanks.

mesh/v1alpha1/config.proto
costinm: We can't define a new proto just because one field is optional/not used, for something that is 99% identical.

costinm (changes requested):
I would like to see more opinions on non reusing RemoteService... 

I don't think introducing a new pattern - just for his particular service - is a good idea.

security/v1beta1/authorization_policy.proto
costinm: I don't see the problem. 
If 'external' is present, the 'action' field can be just ignored ( the default value is 'allow'). 
I don't think any user will be confused that he doesn't have to add an action for 'external' services, quite 
the opposite, they may appreciate not having to type 2x external.

Or maybe use the action as a default, in case the external service is unreachable - I noticed the fail open/fail close is also
missing in the API. I remember for mixer it was a big thing.

security/v1beta1/authorization_policy.proto
costinm: Sounds good.

costinm:
Another thing I noticed - if we use this pattern of using the 'name', and finding the actual resource definition in istio-system, it may
make sense to create a regular ServiceEntry in istio-system with that name, and use the DestinationRule and standard first class Istio APIs. 

Is there anything missing in Istio APIs that would prevent us describing all that is needed to connect ? If yes, are they general enough things we should add the the main API ? I don't see how the external authz service would be any different than any other mesh service we support.

We discussed this in the past, when we reviewed RemoteService - which is my second choice, for consistency. Same thing - if we need to add something to RemoteService, we can. 

But defining a 3rd way to define the address and few properties of  service ( if we don't count the K8S Service too with all its variants) seems extreme.

mesh/v1alpha1/config.proto
liminw: Ok. Maybe just document it as "add or replace"?

mesh/v1alpha1/config.proto
yangminzhu: There is a key difference in the two use cases that I don't think simplify reusing RemoteService is a good idea.

The RemoteService is for services not really part of the mesh (Istio service registry) and configured in bootstrap stage. I guess that's why it needs `tls_settings` and `tcp_keepalive` instead of using the DestinationRule. 

However the ext_authz service is a real service in the Istio service registry (thus using EDS) and configured by the DestinationRule, that's why we don't need (and shouldn't) have any a separate `tls_settings` or `tcp_keepaplive`, not to mention that we need the `subset` to make it fully work with DestinationRule.

Unless we decide that we don't want the benefits of DestinationRule to configure the cluster for ext_authz but rather prefer a pretty static cluster generated from RemoteService, we can't just use RemoteService. It seems the RemoteService should use `Destination` in the fist place but it cannot because of some limitation in the bootstrap.

security/v1beta1/authorization_policy.proto
yangminzhu: I think ignoring the `action` field is not a very user friendly change because it changes the existing semantics. It could fail existing analyze tools because the tool doesn't know the `external` field. And the explicit `EXTERNAL` action provides a very good protection from applying the policy in old Istio version that doesn't support EXTERNAL: a new field is ignored silently but a new enum value will be caught by the webhook validation.

Last, I think the proper long term change is add other fields (`allow,`deny` and `audit`) to the actionDetails and then deprecate the current enum, that's better but can do in later version.

yangminzhu:
> Another thing I noticed - if we use this pattern of using the 'name', and finding the actual resource definition in istio-system, it may
> make sense to create a regular ServiceEntry in istio-system with that name, and use the DestinationRule and standard first class Istio APIs.
> 
> Is there anything missing in Istio APIs that would prevent us describing all that is needed to connect ? If yes, are they general enough things we should add the the main API ? I don't see how the external authz service would be any different than any other mesh service we support.
> 

See https://github.com/istio/api/pull/1698#discussion_r508243869 for my other reply. 

What's the benefit of using it if we already have the service abstraction well defined in `Destination` which is also used in VirtualService? we already have the TLS setting (either auto mTLS or manual setting in DR), subset for testing/migrating between external service and probably some other stuff without needing to add any new fields (not like RemoteService).

The more subtle difference is do we consider the external service as a static config that needs to be configured in bootstrap and out of Istio's xDS dynamic update? Is that the main reason we have this RemoteService instead of using Destination in the fist place?

> We discussed this in the past, when we reviewed RemoteService - which is my second choice, for consistency. Same thing - if we need to add something to RemoteService, we can.
> 
> But defining a 3rd way to define the address and few properties of service ( if we don't count the K8S Service too with all its variants) seems extreme.

We are using the existing `istio.networking.v1alpha3.Destination` to refer to existing service, not defining anything. Is the RemoteService going to replace the `istio.networking.v1alpha3.Destination` in the VirtualService?

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,83 @@ message MeshConfig {
...
+      // Note that in addition to the headers supplied by users:
+      // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
+      // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      repeated string include_headers_in_check = 3;
+
+      // List of headers from the external server that should be added to the original request and forwarded to the
+      // upstream when the authorization check result is allowed (HTTP code 200).
+      // If not specified, the original request will not be modified and forwarded to backend as-is.
+      // Note, any existing headers will be overridden.
+      repeated string headers_to_upstream_on_allow = 4;
```
costinm, line 558: Very strange API - so it's a different list of headers that are sent depending on the status code ? Any real use case where you would send different set of headers ? Can we have a single list ?  

I'm not even sure why a list is needed - if the server decides to send back headers, all could be sent ( and simplify the API).

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,83 @@ message MeshConfig {
   CA ca = 55;
 
   // $hide_from_docs
-  // Next available field number: 57
+  message ExternalProvider {
```
costinm, line 509: Too generic name. There are many other things beyond security that can use 'external providers'. Like Config providers.

Our APIs are too hard too read and write - too many layers of redundant text, too deep ( and error prone ).

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,83 @@ message MeshConfig {
...
+    // REQUIRED. A unique name identifying the external provider.
+    string name = 1;
+
+    // REQUIRED. Specifies detailed configuration for the external provider.
+    oneof provider {
+      // Configures an external authorizer that implements the Envoy ext_authz filter authorization check service using the HTTP API.
+      EnvoyExternalHttpAuthorizationProvider envoy_ext_authz_http = 2;
```
costinm, line 516: I'm very uncomfortable with this pattern - a bool 'useHTTP' or reusing the protocol ( that we have in other protos) would be far more intuitive for a user.

Why don't you use map, with name as a key ?

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +506,83 @@ message MeshConfig {
...
+    // EnvoyExternalAuthorizationServer supplies common configuration for Envoy ext_authz filter authorization check service of HTTP and gRPC API.
+    message EnvoyExternalAuthorizationServer {
+      // REQUIRED. Specifies the network addressable service of the external authorizer that implements the Envoy ext_authz
+      // authorization check service.
+      // Note, the `service.host` must be fully qualified domain name (e.g. "ext-authz.istio-system.svc.cluster.local") and
+      // not short name.
+      istio.networking.v1alpha3.Destination service = 1;
+
+      // Sets the maximum duration in milliseconds for connection to the external authorizer (default is 1s).
+      google.protobuf.Duration timeout = 2;
```
costinm, line 531: Given all the work we've done to resolve the timeout mess and how to set the defaults - this seems to make it worse. 

Why 1s ? Is this the default that envoy recommends ? What did Mixer Policy use ?

security/v1beta1/authorization_policy.proto
costinm: action has a default value - it is not required. Tools that don't know about external proto will fail either way.

Audit is actually a reasonable addition - it fits with 'allow' and 'deny', it represents a typical outcome of the rule. 
'external' is an implementation detail - use the ext_authz. What will be the evolution ? Action will be a list of implementation options ( 'wasm', 'foo' ) and outcomes ?

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,51 @@ message AuthorizationPolicy {
...
-      //
-      // Note that when this list is *not* set, all the authorization response headers, except *Authority (Host)* will be in
-      // the response to the client. When a header is included in this list, *Path*, *Status*, *Content-Length*, *WWWAuthenticate*
-      // and *Location* are automatically added.
-      repeated string forward_to_downstream = 2;
-    }
+  // $hide_from_docs
+  message External {
+    // Specifies the name of the external provider. The list of available providers are defined in the MeshConfig.
+    string provider = 1;
```
costinm, line 301: Maybe use ExtAuthzService ? Please don't take over generic words and use them in very narrow scope. It prevents or makes it very difficult to use them in other parts of the API.

costinm:
Where does the RemoteService proto does it specify that it must be used in
bootstrap ? Or that it must not be part
of the mesh - many of them are.

ExtAuthz also doesn't have to be part of the mesh AFAIK, and with
ServiceEntry every service used can be in or out.
I think it would be extremely bad to require that ExtAuthz runs in-cluster
and prevent out-of-cluster - given that many
clusters can't be properly secured, and ExtAuthz is extremely critical I
would suspect running it in an isolated or
managed env would be a far better practice ( just like CA )

Creating new protos just because we find some differences doesn't make
sense - users would end up with a
different way to configure every single part, with each developer using a
different set of keywords. It's usually called NIH.

mesh/v1alpha1/config.proto
smawson: Do you have an alternative? This is meant to capture any named plugins to Envoy.

mesh/v1alpha1/config.proto
smawson: The name is about what type of extension this is, not just about what protocol it uses. This is specifically an ExternalAuthz filter service that uses the HTTP external authz protocol defined by Envoy.

A map could work, we'd just move the name field up to the map. Only real benefit is preventing duplicates right? Is the yaml better or worse with map vs. list of named items?

mesh/v1alpha1/config.proto
smawson: Can timeout be specified in DestinationRule? If so maybe we don't need this here?

mesh/v1alpha1/config.proto
smawson: See https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/http/ext_authz/v2/ext_authz.proto

costinm:
On Mon, Oct 19, 2020 at 11:52 PM Yangmin Zhu <notifications@github.com>
wrote:

> Another thing I noticed - if we use this pattern of using the 'name', and
> finding the actual resource definition in istio-system, it may
> make sense to create a regular ServiceEntry in istio-system with that
> name, and use the DestinationRule and standard first class Istio APIs.
>
> Is there anything missing in Istio APIs that would prevent us describing
> all that is needed to connect ? If yes, are they general enough things we
> should add the the main API ? I don't see how the external authz service
> would be any different than any other mesh service we support.
>
> See #1698 (comment)
> <https://github.com/istio/api/pull/1698#discussion_r508243869> for my
> other reply.
>
> What's the benefit of using it if we already have the service abstraction
> well defined in Destination which is also used in VirtualService? we
> already have the TLS setting (either auto mTLS or manual setting in DR),
> subset for testing/migrating between external service and probably some
> other stuff without needing to add any new fields (not like RemoteService).
>
> The more subtle difference is do we consider the external service as a
> static config that needs to be configured in bootstrap and out of Istio's
> xDS dynamic update? Is that the main reason we have this RemoteService
> instead of using Destination in the fist place?
>

Bootstrap is an implementation detail of envoy. It is not relevant to the
Istio API. I think Envoy is already moving things out of bootstrap, and
bootstrap
itself is generated dynamically and we can reload envoy if we want.

IMO main reason for RemoteService - just like ConfigSource and many other
things - is that API reviews didn't sufficiently consider API consistency
and developer don't look at the big picture but focus on a specific feature
in isolation.

Now that we have it - we're stuck, we should at least not repeat the
mistakes.


> We discussed this in the past, when we reviewed RemoteService - which is
> my second choice, for consistency. Same thing - if we need to add something
> to RemoteService, we can.
>
> But defining a 3rd way to define the address and few properties of service
> ( if we don't count the K8S Service too with all its variants) seems
> extreme.
>
> We are using the existing istio.networking.v1alpha3.Destination to refer
> to existing service, not defining anything. Is the RemoteService going to
> replace the istio.networking.v1alpha3.Destination in the VirtualService?
>

I don't know - there have been discussions to replace DestinationRule, and
discussions to use DR instead of RemoteService.
Let's not make this even more complicated by adding a 4rd way.

security/v1beta1/authorization_policy.proto
smawson: I would much rather this be explicit and slightly repetitive than rely on a magic behavior of using whatever is in action_detail if action is empty.

For anything that fits in ExternalProvider external will work. So wasm authz could fit in this model without needing a new action type. So this is more about the type of action (call an external service) and not how to implement that action (use envoy's ext_authz api with a grpc call to service x).

So I think the abstraction is the right one here.

security/v1beta1/authorization_policy.proto
smawson: This is the indirection, it is not specifically ExtAuthzService, it is one of the named ExternalProviders (and this is External/provider).

smawson:
Using RemoteService is an interesting idea but it doesn't quite fit here as-is. The problem is that we need to specify additional configuration options beyond just the address. It would maybe work for the grpc api but not the http api.

I think we should look at refactoring to merge RemoteService and ExternalProvider though since they are serving similar purposes. We basically need an any to define additional parameters to the remote service, and we also need to think about how those services are configured (using destination vs. what remote service does). There is definitely an opportunity for convergence.

So I think the question is whether we get this in as-is and then iterate to converge or we try to figure out a convergence story up front.

mesh/v1alpha1/config.proto
yangminzhu: that's what in the Envoy ext_authz filter and gets used by people today, whether you think it's good or not is a different thing. The MeshConfig just includes a list of different concrete providers, and for now we only support Envoy ext_authz filter. If we ever need a different concrete provider, it can be added and named accordingly.

mesh/v1alpha1/config.proto
yangminzhu: The benefit is probably a slighter benefit in preventing duplicate, note it's different than a CRD as this is in MeshConfig which means it cannot be rejected in webhook validation anyway. We can change to use map, it doesn't affect the yaml output too much.

mesh/v1alpha1/config.proto
yangminzhu: We could remove the timeout here and rely on the one in DR.

yangminzhu:
> ExtAuthz also doesn't have to be part of the mesh AFAIK, and with
ServiceEntry every service used can be in or out.
I think it would be extremely bad to require that ExtAuthz runs in-cluster
and prevent out-of-cluster

No one says the ExtAuthz has to be in the mesh, you can deploy it anywhere as needed, with mesh or out of mesh, with or out of sidecar. How is that even related to this PR?

> I don't know - there have been discussions to replace DestinationRule, and
discussions to use DR instead of RemoteService.
Let's not make this even more complicated by adding a 4rd way.

Not adding a 3rd way, it's either RemoteService or the Destination. This PR just choose to use Destination.

I think there are 2 primary issues for this PR:
1. In MeshConfig when configuring how to talk to the external server, do we want the API more generic (ExtensionProvider) or more specific (ExternalAuthorizer)? 
2. Also in MeshConfig when configuring where to talk to the external server, do we use the RemoteService or the Destination? This PR

The PR currently picks the more generic way (ExtensionProvider) and the Destination based on previous comments and suggestions.  @louiscryan do you have any suggestions? thanks.

costinm:
On Tue, Oct 20, 2020 at 8:20 AM Sven Mawson <notifications@github.com>
wrote:

> Using RemoteService is an interesting idea but it doesn't quite fit here
> as-is. The problem is that we need to specify additional configuration
> options beyond just the address. It would maybe work for the grpc api but
> not the http api.
>

Common problem for all services - the fact that a service has some
additional setting doesn't mean they need to be crammed into RemoteService
( just like in K8S, Service is mostly used for configuring properties of
the service ).

It appears for grpc API we actually don't even need RemoteService - if the
Policy external address is using the actual service, which
makes the grpc case super clean and simple.



I think we should look at refactoring to merge RemoteService and
> ExternalProvider though since they are serving similar purposes. We
> basically need an any to define additional parameters to the remote
> service, and we also need to think about how those services are configured
> (using destination vs. what remote service does). There is definitely an
> opportunity for convergence.
>
I would go one step further - we need a mechanism to define additional
parameters for any 'feature' or 'service'.
I think Nate already added something a few releases back for a similar use
case, for DestinationRule ?

If we do this - in most cases we'll not need either RemoteService nor
ExternalProvider, instead we can use the 'natural'
name of the service ( authz.istio-system:9080), as we do in the common
Istio APIs.



> So I think the question is whether we get this in as-is and then iterate
> to converge or we try to figure out a convergence story up front.
>
We had an offline chat - it appears we are in pretty strong agreement that
adding the Policy EXTERNAL is useful, and the gRPC case
is the simplest and most useful.

We can just add gRPC support in 1.8, and have more time for 1.9 to decide
how ( and if ) we support http.
Most of our protocols are gRPC based, so was the original mixer Policy that
this API replaces. We also need
to reconsider the caching and performance/overhead.

I think this will unblock the most important use case, and we are not even
tied to the specifics of ext_auth current
behavior.



>

yangminzhu:
@louiscryan @smawson @mandarjog @costinm @justinpettit @nrjpoddar @linsun @liminw I have updated the PR based the feedback and discussions, can you take a look at again, thanks.

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +505,97 @@ message MeshConfig {
...
+      // Configures an external authorizer that implements the Envoy ext_authz filter authorization check service using the HTTP API.
+      EnvoyExternalAuthorizationHttpProvider envoy_ext_authz_http = 2;
+      // Configures an external authorizer that implements the Envoy ext_authz filter authorization check service using the gRPC API.
+      EnvoyExternalAuthorizationGrpcProvider envoy_ext_authz_grpc = 3;
+    }
+
+    // $hide_from_docs
+    message EnvoyExternalAuthorizationHttpProvider {
+      // REQUIRED. Specifies the service that implements the Envoy ext_authz HTTP authorization service.
+      // The format is "[<Namespace>/]<Service>". The <Namespace> will have the default value "istio-system" if it is
```
louiscryan, line 523: @costinm I don't think we need to explicitly state that the default value of Namespace=istio-system here. Can just state that namespace can be used to more explicitly qualify which service to use if the service name is ambiguous.

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +505,97 @@ message MeshConfig {
...
+      // omitted. The <Service> is the name of the service object (k8s service or ServiceEntry).
+      // Example: "foo/my-ext-authz" or "my-ext-authz".
+      string service = 1;
+
+      // Specifies the port of the service. Can be omitted if it can be decided unambiguously for the service (e.g. the
```
louiscryan, line 528: s/decided unambiguously for/derived unambiguously from

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +505,97 @@ message MeshConfig {
...
+
+      // Sets a prefix to the value of authorization request header *Path*.
+      // For example, setting this to "/check" for an original user request at path "/admin" will cause the
+      // authorization check request to be sent to the authorization service at the path "/check/admin" instead of "/admin".
+      string path_prefix = 3;
+
+      // If true, the user request will be allowed even if the communication with the authorization service has failed,
+      // or if the authorization service has returned a HTTP 5xx error.
+      // Default is false and the request will be rejected with "Forbidden" response.
+      bool fail_open = 4;
```
louiscryan, line 540: @costinm 
@smawson 

should this setting be hoisted a level? Isn't this generic for all possible providers

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +505,97 @@ message MeshConfig {
...
+      // automatically added.
+      // Note, the body from the authorization service is always included in the response to downstream.
+      repeated string headers_to_downstream_on_deny = 8;
+    }
+
+    // $hide_from_docs
+    message EnvoyExternalAuthorizationGrpcProvider {
+      // REQUIRED. Specifies the service that implements the Envoy ext_authz gRPC authorization service.
+      // The format is "[<Namespace>/]<Service>", the <Namespace> will have the default value "istio-system" if it is
+      // omitted. The <Service> is the name of the service object (k8s service or ServiceEntry).
```
louiscryan, line 572: same comment as above

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +505,97 @@ message MeshConfig {
...
+      // The default status is "403" (HTTP Forbidden).
+      string status_on_error = 4;
+    }
+  }
+
+  // $hide_from_docs
+  // Defines a list of extension providers that extend Istio's functionality. For example, the AuthorizationPolicy
+  // can be used with an extension provider to delegate the authorization decision to a custom authorization system.
+  // Note, currently at most 1 extension provider is allowed.
+  repeated ExtensionProvider extension_providers = 57;
```
louiscryan, line 595: I think this was updated in docs to be 'extended_authorizers' ?

@smawson

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,55 @@ message AuthorizationPolicy {
     // Audit policy decisions can be read by telemetry plugins using the function getAuditPolicy
     // defined [here](https://github.com/istio/proxy/blob/master/extensions/common/context.h).
 
-    // External delegates the authorization decision to an external server out of Envoy. An extra authorization request
-    // will be sent to the external server and the response will decide whether or not the original request should be
-    // allowed or denied.
+    // $hide_from_docs
+    // The EXTENDED action lets an extension to handle the user request if the  matching rules evaluate to true.
```
louiscryan, line 264: s/lets/allows

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,55 @@ message AuthorizationPolicy {
...
+    // The extension is evaluated independently and before the native ALLOW and DENY actions. When used together, A request
+    // is allowed if and only if all the actions return allow, in other words, the extension cannot bypass the
+    // authorization decision made by ALLOW and DENY action.
+    // The detail extension is referred to by its provider name defined in the MeshConfig.
```
louiscryan, line 268: reword?

Extension behavior is defined by the named providers declared in MeshConfig

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,55 @@ message AuthorizationPolicy {
...
+    // One example use case of the extension is to integrate with a custom external authorization system to delegate
+    // the authorization decision to it.
     //
-    // The EXTERNAL action is evaluated before the ALLOW and DENY action, implemented by the Envoy ext_authz filter
-    // and is currently an **experimental feature** in Istio.
-    EXTERNAL = 3;
+    // Note: The EXTENDED action is currently an **experimental feature** and is subject to breaking changes in later versions.
+    // We're actively working to improve it based on user feedbacks.
```
louiscryan, line 273: s/feedbacks/feedback

louiscryan: actually, this is true for all features. Just remove, saying it's experimental is enough

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,55 @@ message AuthorizationPolicy {
...
+    //
+    // The following authorization policy applies to ingress gateway and delegates the authorization check to the extension
```
louiscryan, line 275: The following authorization policy applies to an ingress gateway and delegates the authorization check to a named extension

louiscryan:
Generally LGTM. Some doc and naming nits and one question about where to model fail-open in the hierarchy that I could be persuaded either way on.

mesh/v1alpha1/config.proto
smawson: I think we should just require namespace to be specified. Service names can start unambiguous and become ambiguous later, so we wouldn't want to have a search.

mesh/v1alpha1/config.proto
smawson: I'm not sure, would a telemetry provider have fail open vs. fail closed?

mesh/v1alpha1/config.proto
smawson: No I think we had it as extension_providers since this list can be reused for other types of providers.

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +505,97 @@ message MeshConfig {
...
+    // $hide_from_docs
+    message EnvoyExternalAuthorizationHttpProvider {
+      // REQUIRED. Specifies the service that implements the Envoy ext_authz HTTP authorization service.
+      // The format is "[<Namespace>/]<Service>". The <Namespace> will have the default value "istio-system" if it is
+      // omitted. The <Service> is the name of the service object (k8s service or ServiceEntry).
+      // Example: "foo/my-ext-authz" or "my-ext-authz".
+      string service = 1;
+
+      // Specifies the port of the service. Can be omitted if it can be decided unambiguously for the service (e.g. the
+      // service has defined only 1 port).
```
smawson, line 529: So if this is not specified, and the user adds a second port to the service, authz will break? Making it explicit is probably less dangerous.

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,55 @@ message AuthorizationPolicy {
...
   // Optional. The action to take if the request is matched with the rules.
   Action action = 3;
 
-  // action_detail specifies details of the corresponding action.
-  oneof action_detail {
-    // Configures how to talk to the external server. Must be set only with EXTERNAL action.
-    External external = 4;
+  // $hide_from_docs
+  message ExtensionProvider {
+    // Specifies the name of the extension provider. The list of available providers are defined in the MeshConfig.
```
smawson, line 304:
```suggestion
    // Specifies the name of the extension provider. The list of available providers is defined in the MeshConfig.
```

smawson (commented):
Looking good, a few last items to resolve.

mesh/v1alpha1/config.proto
smawson: Louis and I chatted, I'm OK with local resolution meaning "in the same namespace as mesh-config resides". So if namespace is not specified, it is relative to the namespace the config is in. That should be safe.

So reword to say:

If <Namespace> is omitted then the <Service> is resolved within the same namespace as this configuration resource.

Or something like that (don't love the wording).

mesh/v1alpha1/config.proto
smawson: Chatted with Louis and since it doesn't actually save our users any typing we can leave this down here on the hope that extension providers is useful beyond authorization.

nrjpoddar:
@yangminzhu is this API hidden from docs because it’s not implemented or we think that in future a newer better first class API will be created for this feature? If it’s the second one then are we going to advertise this feature to users and given them a migration path?

mesh/v1alpha1/config.proto
```diff
@@ -505,7 +505,97 @@ message MeshConfig {
...
+      string path_prefix = 3;
+
+      // If true, the user request will be allowed even if the communication with the authorization service has failed,
+      // or if the authorization service has returned a HTTP 5xx error.
+      // Default is false and the request will be rejected with "Forbidden" response.
+      bool fail_open = 4;
+
+      // Sets the HTTP status that is returned to the client when there is a network error to the authorization service.
+      // The default status is "403" (HTTP Forbidden).
+      string status_on_error = 5;
```
hzxuzhonghu, line 544: This only covers http protocol, what's it for the others?

yangminzhu: other protocol you mean TCP? I think the ext_authz filter doesn't provide this feature for TCP. Note this is the user traffic and it should be either HTTP or TCP based.

yangminzhu:
> @yangminzhu is this API hidden from docs because it’s not implemented or we think that in future a newer better first class API will be created for this feature? If it’s the second one then are we going to advertise this feature to users and given them a migration path?

it's not implemented for now, the API will be made public once implemented (will also have new task on istio.io). There might be further API changes and the first class API (like a new CRD) is more likely for the changes in MeshConfig and not for changes in AuthZ, but it's too early to decide anything now and we need some more feedback later.

Later if we have a different API, there will be a way to migrate the core use case even the current API is marked as experimental and subject to breaking changes.

security/v1beta1/authorization_policy.proto
yangminzhu: removed.

mesh/v1alpha1/config.proto
yangminzhu: yes, I think we had it as extension provider later from the discussion. The example in the design doc is little bit out-of-date (just updated).

mesh/v1alpha1/config.proto
yangminzhu: updated.

mesh/v1alpha1/config.proto
yangminzhu: +1 to make this default to the namespace of the mesh-config, updated.

mesh/v1alpha1/config.proto
yangminzhu: yeah, I also prefer to leave it to the specific type as some other extension provider may doesn't need or have a different interpretation or default value for the `fail_open` concept.

security/v1beta1/authorization_policy.proto
```diff
@@ -313,108 +260,55 @@ message AuthorizationPolicy {
...
+    //      app: istio-ingressgateway
+    //  action: EXTENDED
+    //  provider:
+    //    name: "my-custom-authz"
+    //  rules:
+    //  - to:
+    //    - operation:
+    //        paths: ["/admin/*"]
+    // ```
+    EXTENDED = 3;
```
nrjpoddar, line 296: EXTENDED name here for an action seems off and probably a bit confusing. Other suggestion:

1. DELEGATE

mesh/v1alpha1/config.proto
yangminzhu: That's a very good point. Previously this was to follow the convention of the `Destination` and also try to make it a bit simpler to configure. Note it shouldn't be more dangerous as we will always fall back to a deny-all config if we encounter any errors like this.

I think it's probably expected that something might break if you change the name or modify the existing port of a service, but probably a bit surprise that something could break just because you add a new port to a service? 
I now prefer to make the port required to avoid such surprises, @louiscryan  do you have any suggestions?

security/v1beta1/authorization_policy.proto
yangminzhu: I think we're trying to be a little bit more generic here but I get your point why the name EXTENDED seems a bit off when comparing to ALLOW/DENY. 

But I'm a little worried that the name DELEGATE is a bit too specific for a generic API that integrates different extensions to Istio, many extensions might be doing things different than the "delegate" concept. Note this is not saying the Authz policy will become such generic API but I think it's the name and convention we used here should be generic. This also aligns with the naming in MeshConfig where we had the `extension_providers`.

 cc @louiscryan @smawson

mesh/v1alpha1/config.proto
louiscryan: It hasn't been much of an issue with Destination so far so I don't think we need to be overly concerned. In terms of blast radius a user could just as easily remove the port or improperly declare its supported protocol.

That said either is fine with me.

mesh/v1alpha1/config.proto
louiscryan: For gRPC it should derive the gRPC status from it's canonical HTTP mapping.

mesh/v1alpha1/config.proto
louiscryan: ok, sgtm.

security/v1beta1/authorization_policy.proto
louiscryan: TBH I'm not a fan of DELEGATE as it suggest 'as an alternative to' DENY/ALLOW.

I prefer EXTENDED but would be comfortable with CUSTOM. I believe EXTERNAL was discussed but it's implied 'out of process' semantics wasn't liked.

nrjpoddar: I see both your points. I don't mind "CUSTOM" as it fits the extension mechanism here and leaves room for doing improvements in future.

yangminzhu: yes, `EXTERNAL` was discussed but not used because it implies that an extension is always `out of process`. 

I still kind more prefer the `EXTENDED` because it aligns with the terms we used in the MeshConfig (`extendion_provider`), and `CUSTOM` doesn't seem any different from the `EXTENDED` but it lost the connection with the related fields in MeshConfig.

mesh/v1alpha1/config.proto
yangminzhu: thanks, changed the document to mark the port as required.

security/v1beta1/authorization_policy.proto
nrjpoddar: I understand the linkage but an "EXTENDED" action in this API doesn't sound right to me. I would advocate for "CUSTOM".

PS: I know naming and driving consensus is hard so thanks for being patient as all of us give you this feedback.

nrjpoddar:
@yangminzhu I see that this PR is marked for "release-1.8". I don't think it's a good idea to change APIs and add features after the branch has been created for stability. I would strongly recommend against adding this feature to 1.8.

smawson (approved):
Good enough for me :)

yangminzhu:
> @yangminzhu I see that this PR is marked for "release-1.8". I don't think it's a good idea to change APIs and add features after the branch has been created for stability. I would strongly recommend against adding this feature to 1.8.

There is a bit more context here. The current API (i.e. the EXTERNAL action and other fields) was merged before 1.8 release and it becomes part of the authz CRD that will be installed in user's cluster. As you can see this PR updates the API a lot and we really should cherry-pick this to the 1.8 to make sure the CRD in user cluster is the same to avoid upgrade issues later.

We only need to cheery-pick the API to 1.8 to fix the CRD and will NOT cherry-pick the implementation. For 1.8, the API is hidden in the doc and there is no implementation for it as well. @nrjpoddar let me know if this sounds good to you, thanks.

security/v1beta1/authorization_policy.proto
yangminzhu: I see, then let's make it "CUSTOM", thanks for the review Neeraj!

costinm (approved):

nrjpoddar:
Yes it does, thanks!

nrjpoddar:
@yangminzhu have we discussed about "default" extension provider selection works for your mesh in cases when you have multiple ones defined or there's no concept of "default", and you only select/attach per workload?

yangminzhu:
> @yangminzhu have we discussed about "default" extension provider selection works for your mesh in cases when you have multiple ones defined or there's no concept of "default", and you only select/attach per workload?

currently there is no default extension provider and you have to select it per Authz policy (it can apply to multiple workloads), the initial implementation will also restrict to allow at most 1 extension provider, this means effectively all authz policy can only use the same extension provider (for the initial implementation)

PR #1699, opened 2020-10-18
Fix kubectl explain

jasonwzm:
Fixes https://github.com/istio/istio/issues/28020

Tested in k8s 1.19.0 with latest dev build.

As before, envoyfilter does not work with kubectl explain as the cue-gen tool to generate v1beta1 did not support specifying preserving unknown field at the field level.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -35,6 +35,7 @@ spec:
     shortNames:
     - dr
     singular: destinationrule
+  preserveUnknownFields: false
   scope: Namespaced
   subresources:
     status: {}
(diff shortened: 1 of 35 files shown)


howardjohn (approved):

howardjohn:
LGTM, I manually tested envoyfilter pruning since i was confused by the PR initially since i didn't realize we also have the preserve fields at the field level - all good (istio/istio tests would catch it anyways). Thanks!

smawson (approved):
Fine by me, but can we figure out a way to add a test for this so we don't regress in the future? A golden-file test that runs kubectl explain and verifies the output would be a little noisy since we'd need to update it regularly, but it would catch this problem...

howardjohn:
we could even skip the need for a golden file, just check *something* is
output under .spec I think is sufficient

smawson:
Yeah I was thinking the same thing, but an actual golden file may be nice, depends on noise level. It would catch someone unintentionally changing the explain output. If those files are auto-updated just like the rest of our files we could at least look at them in a review but it wouldn't add too much noise (sort of like the schema files that are generated).

PR #1701, opened 2020-10-20
deprecate addon field from the operator

richardwxn:
in 1.7 we deprecated the telemetry addons and in 1.8 we deprecate istiocoredns as well so we can mark this whole field as deprecated.


howardjohn (approved):
what is the user result of this? just a doc change?

richardwxn:
@howardjohn it is doc change, and for telemetry addon specifically, we would fail hard during runtime of installation

richardwxn:
@linsun @smawson Can you help review?

linsun:
@richardwxn can you make sure we have release notes indicate the deprecation?

Otherwise LGTM.

operator/v1alpha1/operator.proto
```diff
@@ -99,7 +99,7 @@ message IstioOperatorSpec {
     IstioComponentSetSpec components = 50;
 
     // Extra addon components which are not explicitly specified above.
-    map<string, ExternalComponentSpec> addon_components = 51;
+    map<string, ExternalComponentSpec> addon_components = 51 [deprecated = true];
```
smawson, line 102: Can you add a comment here indicating that Istio will no longer manage external components and that users should move to managing these themselves?

richardwxn: added

smawson (approved):

PR #1702, opened 2020-10-20
Cleanup mesh config docs

howardjohn:
Original intent was to document proxy config annotation and unhide the
proxy metadata section, which is used by many users now. On the way I
removed+reserved some dead fields and cleaned up a few comments. I can
revert the clean up if its controversial


smawson (approved):
Generally I would favor doing these separately just in case we need to do a rollback of one part without affecting the other, but it seems like all of the changes are straightforward so I'm OK with it all bundled in one.

ericvn (approved):

PR #1715, opened 2020-10-27
Promote pod label for locality to istio/api

nmittler:
In istio/istio this label is currently called "istio-locality", but is not currently documented anywhere. Changing it to "topology.istio.io" to be consistent with other labels.

### DIFF
--- a/label/label.go
+++ b/label/label.go
@@ -38,8 +38,16 @@ const (
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
 
-	// IstioSubZone identifies the locality subzone of a workload.
+	// IstioSubZone is a node label that allows a cluster admin to set the locality
+	// sub-zone of workload instances assigned to that node.
 	IstioSubZone = "topology.istio.io/subzone"
+
+	// IstioLocality is a pod label that allows the service owner to override the locality
+	// node labels (set by Kubernetes or a cluster admin) for a pod. The string is a
+	// dot-separated string (since k8s labels do not support `/`) of the form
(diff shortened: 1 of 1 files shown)


label/label.go
```diff
@@ -38,9 +38,14 @@ const (
 	// IstioOperatorVersion is the Istio operator version that installed the resource, e.g. "1.6.0"
 	IstioOperatorVersion = "operator.istio.io/version"
 
-	// IstioSubZone identifies the locality subzone of a workload.
+	// IstioSubZone is a node label that identifies the locality subzone of a workload instance.
 	IstioSubZone = "topology.istio.io/subzone"
 
+	// IstioLocality is a pod label that allows the service owner to override the locality
+	// of a workload instance as defined by the cloud vendor. The string is a dot-separate string
+	// (since k8s labels do not support `\`) of the form region.zone.subzone.
```
hzxuzhonghu, line 46: s/`\`/`/`

label/label.go
```diff
@@ -38,9 +38,14 @@ const (
...
+	IstioLocality = "topology.istio.io/locality"
```
howardjohn, line 47: Are we sure we want to actually allow this? part of the reason its not in api is because there was no real use cases (at the time, at least) to set this

nmittler: That's really part of the motivation for this PR ... do we want this?

I'm actually in the process of converting the locality LB page into an actual task. In order to set locality, I'm using this ... but it's awkward because it's not actually documented anywhere. If I make this task, it will be ... which raises the API question.

If I don't use this in the task, are there other options? I don't suspect cloud vendors will let you manually specify the region and zone labels.

nmittler: @smawson @louiscryan @linsun any thoughts?

nmittler:
Decided to abandon this change for now.

PR #1720, opened 2020-10-30
typo in virtual-service docs

rvennam:
It also romoves the foo response header -> It also removes the foo response header


hzxuzhonghu (approved):

hzxuzhonghu:
ping @linsun to approve

linsun (approved):

PR #1723, opened 2020-10-30
Specify feature status for annotations

nmittler:
I'd like to establish the feature level for each annotation. They default to `Alpha`. Please go leave a comment in the yaml next to annotations you're familiar with.


nmittler:
No idea why gen is failing ... seems like it's still using the old version of the annotation generator

ericvn:
Common files should be updated by the automated PRs.

common/scripts/setup_env.sh
```diff
@@ -59,7 +59,7 @@ fi
 
 # Build image to use
 if [[ "${IMAGE_VERSION:-}" == "" ]]; then
-  export IMAGE_VERSION=master-2020-10-30T19-21-15
+  export IMAGE_VERSION=master-2020-10-30T22-07-33
```
ericvn, line 62: This change is a down level version at this point.

ericvn (dismissed):

annotation/annotations.yaml
```diff
@@ -254,58 +285,67 @@ annotations:
...
     resources:
       - Any
   - name: install.operator.istio.io/version
+    featureStatus: Alpha
     description: Represents the Istio version associated with the resource
     deprecated: false
     hidden: false
     resources:
       - Any
   - name: galley.istio.io/analyze-suppress
```
louiscryan, line 347: @esnible 
@therealmitchconnors 
 I don't suppose we can rename this or deprecate? We shouldn't be naming things after components.

@nmittler 

Do we have any framework abstraction around reading annotations that could be used to help with deprecations etc? For example if we declared here that this was a deprecated 'alias' of a supported annotation could the framework take advantage of that?

config.istio.io/....  seems more appropriate

annotation/annotations.yaml
```diff
@@ -22,6 +23,7 @@ annotations:
     resources:
       - Pod
   - name: alpha.istio.io/kubernetes-serviceaccounts
```
louiscryan, line 25: Is this a supported naming convention, i.e has some framework support? Judging by the other names it would seem not.

louiscryan:
@smawson 

Thanks for sending this out Nate, very helpful. Ill go through the names in detail and add commentary but I have a higher level question about ensuring conformance.

I'm wondering if a framework abstraction around annotation reading in the code would be a useful thing or not? A framework would enable us to track usage, or the lack thereof, of given features. We could also use linting in theory to ban reading of annotations via anything other than the framework to ensure process is followed.

@nrjpoddar 
@howardjohn

annotation/annotations.yaml
nmittler: I'm not sure we ever reached agreement on how to handle feature status in annotations/labels. Kubernetes played around with something similar, but I believe they've moved away from this as well. In the TOC on Friday, I suggested not using this, but instead just documenting the feature status of each label (which is what I'm doing here).

annotation/annotations.yaml
nmittler: AFAIK there is no processing of this data within Istio today. However all the data is here for such a framework, if it were built. For now, the purpose of this is for generating html that appears on istio.io.

nmittler:
@ericvn I think there is something wrong with the version of the build-tools that is being used. It's still somehow using a slightly older version. I suspect there may be a caching issue?

annotation/annotations.yaml
smawson: Looks like we should take a pass over the existing annotations and work out a plan for unifying our naming and usage, since it is kind of all over the place right now.

Anyone want to start a doc with a list of the existing ones and proposals for what to do with them? If not I can do it.

annotation/annotations.yaml
```diff
@@ -40,13 +43,15 @@ annotations:
...
       - Service
   - name: alpha.istio.io/identity
+    featureStatus: Alpha
     description: Identity for the workload.
       NOTE This API is Alpha and has no stability guarantees.
     deprecated: false
     hidden: true
     resources:
       - Pod
   - name: networking.istio.io/exportTo
```
smawson, line 53: Should exportTo be considered Beta? Thoughts from networking folks?

annotation/annotations.yaml
smawson: Our naming of annotations is pretty chaotic, we should probably get a better handle on those and try to align our api groups in general across annotations, labels, and CRDs.

ericvn:
@nmittler I think you'll just need to rebase and pick up the common-file changes (which haven't been going through in an automated fashion) Not sure if the lint changes in the last failed merge should have an affect at this point.

ericvn:
Last gencheck test did used the latest build images, and showed conflicts on the two mentioned files:
```
Auto-merging annotation/annotations.pb.html
CONFLICT (content): Merge conflict in annotation/annotations.pb.html
Auto-merging annotation/annotations.gen.go
CONFLICT (content): Merge conflict in annotation/annotations.gen.go
```

louiscryan:
@therealmitchconnors helpfully pointed to

https://github.com/istio/pkg/blob/master/env/var.go

as an example of a framework doing some validation for ENV variable. We may look at creating a similar manifest for those to track status, usage, defaults etc. There might even be a linter validating that the framework is used.

Mitch also suggested adding a /ctrlz page to see usage status

annotation/annotations.yaml
```diff
@@ -64,13 +70,15 @@ annotations:
     resources:
       - Pod
   - name: sidecar.istio.io/inject
+    featureStatus: Alpha
```
nrjpoddar, line 73: I'm assuming this is beta or higher.

annotation/annotations.yaml
louiscryan: @howardjohn 

I believe so, it's part of ServiceEntry

annotation/annotations.yaml
```diff
@@ -55,6 +60,7 @@ annotations:
     resources:
       - Service
   - name: security.istio.io/tlsMode
+    featureStatus: Alpha
```
louiscryan, line 63: @incfly

annotation/annotations.yaml
```diff
@@ -64,13 +70,15 @@ annotations:
...
   - name: sidecar.istio.io/inject
+    featureStatus: Alpha
     description: Specifies whether or not an Envoy sidecar should be automatically
       injected into the workload.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: sidecar.istio.io/status
+    featureStatus: Alpha
```
nrjpoddar, line 81: If this is not a user provided configuration I think we should separate it out in a different table or section.

annotation/annotations.yaml
louiscryan, line 73: @costinm this should be Beta I believe?

annotation/annotations.yaml
```diff
@@ -79,13 +87,15 @@ annotations:
...
   - name: sidecar.istio.io/rewriteAppHTTPProbers
+    featureStatus: Alpha
     description: Rewrite HTTP readiness and liveness probes to be redirected to
       the Envoy sidecar.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: sidecar.istio.io/controlPlaneAuthPolicy
+    featureStatus: Alpha
```
nrjpoddar, line 98: Do we still support this annotation? AFAIK control plane auth policy is enabled by default and the only supported model.

annotation/annotations.yaml
```diff
@@ -94,31 +104,36 @@ annotations:
...
   - name: sidecar.istio.io/discoveryAddress
+    featureStatus: Alpha
     description: Specifies the XDS discovery address to be used by the Envoy
       sidecar.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: sidecar.istio.io/proxyImage
+    featureStatus: Alpha
```
nrjpoddar, line 115: Beta or higher

annotation/annotations.yaml
```diff
@@ -231,6 +260,7 @@ annotations:
     resources:
       - Pod
   - name: traffic.sidecar.istio.io/includeOutboundIPRanges
+    featureStatus: Alpha
```
nrjpoddar, line 263: Beta or higher

annotation/annotations.yaml
```diff
@@ -239,6 +269,7 @@ annotations:
     resources:
       - Pod
   - name: traffic.sidecar.istio.io/excludeOutboundIPRanges
+    featureStatus: Alpha
```
nrjpoddar, line 272: Beta or higher

annotation/annotations.yaml
```diff
@@ -131,40 +146,46 @@ annotations:
     resources:
       - Pod
   - name: sidecar.istio.io/interceptionMode
+    featureStatus: Alpha
```
louiscryan, line 149: @costinm is this even Alpha?

annotation/annotations.yaml
```diff
@@ -317,6 +357,7 @@ annotations:
     resources:
       - Any
   - name: proxy.istio.io/config
+    featureStatus: Alpha
```
nrjpoddar, line 360: Beta or higher @howardjohn ?

annotation/annotations.yaml
```diff
@@ -131,40 +146,46 @@ annotations:
...
       - Pod
   - name: sidecar.istio.io/bootstrapOverride
+    featureStatus: Alpha
     description: Specifies an alternative Envoy bootstrap configuration file.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: sidecar.istio.io/statsInclusionPrefixes
+    featureStatus: Alpha
```
louiscryan, line 164: @douglas-reid @mandarjog PTAL

nrjpoddar:
@nmittler thanks for taking a crack at this. Looks like my & Loui's review overlapped so you might end up getting few duplicate comments.

annotation/annotations.yaml
```diff
@@ -14,6 +14,7 @@
 
 annotations:
   - name: prometheus.istio.io/merge-metrics
+    featureStatus: Alpha
```
douglas-reid, line 17: this is probably fine, but with the behavior on by default now, i'm wondering what would be needed to promote to Beta. @howardjohn thoughts?

annotation/annotations.yaml
```diff
@@ -94,31 +104,36 @@ annotations:
     resources:
       - Pod
   - name: sidecar.istio.io/discoveryAddress
+    featureStatus: Alpha
```
nrjpoddar, line 107: @howardjohn I think these settings now come from the /config i.e. ProxyConfig fields?

annotation/annotations.yaml
```diff
@@ -94,31 +104,36 @@ annotations:
...
       - Pod
   - name: sidecar.istio.io/proxyImage
+    featureStatus: Alpha
     description: Specifies the Docker image to be used by the Envoy sidecar.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: sidecar.istio.io/proxyCPU
+    featureStatus: Alpha
```
nrjpoddar, line 122: We can start deprecating this and use ProxyConfig or `sidecar/istio.io/config` instead.

annotation/annotations.yaml
howardjohn: yes, would call deprecated

annotation/annotations.yaml
howardjohn: Not this one, its not in proxy config. At least not yet

annotation/annotations.yaml
howardjohn: Beta SGTM

annotation/annotations.yaml
howardjohn: That is questionable to me. I don't think this is the proper API. At the very least seems like it could be in proxy config

annotation/annotations.yaml
howardjohn: cannot even find a mention in istio/istio

annotation/annotations.yaml
nrjpoddar: I was confusing this with concurrency, you're correct.

annotation/annotations.yaml
douglas-reid: This has been around for awhile now but with recent API changes adding this control into proxy config, this was basically considered deprecated (though still supported).  Is there a `Deprecated` status? I guess `Alpha` could suffice, but that almost implies a velocity that isn't there.

Same for `suffixes` and `regexps` below.

annotation/annotations.yaml
howardjohn: Its one of those things that has no tests and doesn't work but has users. There are some efforts to add TPROXY tests ongoing

nrjpoddar: I don't think so. TPROXY is often broken and I'm pretty sure not tested or maintained.

annotation/annotations.yaml
howardjohn: this is more internal. Also isn't it a label?

annotation/annotations.yaml
```diff
@@ -254,58 +285,67 @@ annotations:
     resources:
       - Pod
   - name: traffic.sidecar.istio.io/excludeInboundPorts
+    featureStatus: Alpha
```
nrjpoddar, line 288: Beta or higher as this is the only mechanism to bypass sidecars per port afaik.

louiscryan:
Sounds like a lot of things are deprecated. Can we add a 'deprecatedBy' field to the spec pointing to the reference material for what if anything replaced it. This would really help the generated docs be useful.

nmittler:
@louiscryan 

> Sounds like a lot of things are deprecated. Can we add a 'deprecatedBy' field to the spec pointing to the reference material for what if anything replaced it. This would really help the generated docs be useful.

https://github.com/istio/tools/pull/1293

annotation/annotations.yaml
nmittler: We could separate input and output annotations into separate files. However, do we want feature levels for output annotations? Seems like we would, no?

louiscryan:
I think yes. Can we add a field to indicate if the annotation is 'runtime'
(generated by us) vs 'config'

They are still likely to need versioning and deprecation tracking over time
too.

nmittler:
@louiscryan I've updated https://github.com/istio/tools/pull/1293 to add `DeprecatedBy` and `Source`.

annotation/annotations.yaml
```diff
@@ -79,13 +87,15 @@ annotations:
     resources:
       - Pod
   - name: sidecar.istio.io/rewriteAppHTTPProbers
+    featureStatus: Alpha
```
nrjpoddar, line 90: Probably beta or higher cc @howardjohn

annotation/annotations.yaml
```diff
@@ -131,40 +146,46 @@ annotations:
...
   - name: sidecar.istio.io/interceptionMode
+    featureStatus: Alpha
     description: Specifies the mode used to redirect inbound connections to Envoy
       (REDIRECT or TPROXY).
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: sidecar.istio.io/bootstrapOverride
+    featureStatus: Alpha
```
nrjpoddar, line 157: Is that actually supporter?

annotation/annotations.yaml
```diff
@@ -131,40 +146,46 @@ annotations:
...
   - name: sidecar.istio.io/statsInclusionRegexps
+    featureStatus: Alpha
     description: Specifies the comma separated list of regexes the stats should match
       to be emitted by Envoy.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: sidecar.istio.io/userVolume
+    featureStatus: Alpha
```
nrjpoddar, line 188: Is this still supported? I know we have had it from the beginning so we should either deprecate or promote this.

annotation/annotations.yaml
```diff
@@ -179,50 +200,58 @@ annotations:
...
       - Pod
   - name: readiness.status.sidecar.istio.io/failureThreshold
+    featureStatus: Alpha
     description: Specifies the failure threshold for the Envoy sidecar readiness probe.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: readiness.status.sidecar.istio.io/applicationPorts
+    featureStatus: Alpha
```
nrjpoddar, line 254: I’m guessing rewriteProbe feature makes this deprecated

annotation/annotations.yaml
```diff
@@ -254,58 +285,67 @@ annotations:
...
   - name: traffic.sidecar.istio.io/excludeOutboundPorts
+    featureStatus: Alpha
     description: A comma separated list of outbound ports to be excluded from redirection
       to Envoy.
     deprecated: false
     hidden: false
     resources:
       - Pod
   - name: traffic.sidecar.istio.io/kubevirtInterfaces
+    featureStatus: Alpha
```
nrjpoddar, line 312: Do we actively test this annotation?

annotation/annotations.yaml
nrjpoddar: We have both proxy.istio.io and sidecar.istio.io annotations. I’m assuming proxy annotations apply to both gateways and sidecars?

annotation/annotations.yaml
hzxuzhonghu: TPROXY is not working now

annotation/annotations.yaml
therealmitchconnors: By definition, Alpha-level features cannot be enabled by default, and must be opted into by users.  I suggest we classify this as beta.

douglas-reid: There is a difference between the feature itself (metrics merge) and the support for an annotation to override it though, right? We don't require annotations for all features, etc.

annotation/annotations.yaml
linsun: Is this still useful? :) Seems no need to use it unless you want to disable it.

annotation/annotations.yaml
linsun: agreed - seems no need to have it now.

annotation/annotations.yaml
linsun: or remove it? :)

annotation/annotations.yaml
```diff
@@ -14,6 +14,7 @@
 
 annotations:
```
linsun, line 15: are all of our annotations at pod level?

linsun:
Wow i had no idea we have this many annotations!!  would be nice if we can categorize them, not only with feature status, but grouping them to help users to understand (scenarios) when each should be used.

annotation/annotations.yaml
nmittler: um ... so I just looked through the istio repo and AFAICT this annotation is not tested.

@howardjohn please tell me I missed something :)

annotation/annotations.yaml
nmittler: I don't see any tests that use this outside of istioctl.

@howardjohn

annotation/annotations.yaml
nmittler: Created https://github.com/istio/api/issues/1737 to track

annotation/annotations.yaml
nmittler: @howardjohn does your comment on `traffic.sidecar.istio.io/excludeOutboundIPRanges` (below) apply here as well?

nmittler:
@louiscryan @douglas-reid @smawson @nrjpoddar @linsun @howardjohn I've gone through your comments and promoted those for which I could find tests in istio/istio to Beta and deprecated those suggested.

I've also opened tracking bugs for enhancements to the annotation generation tooling:

- https://github.com/istio/api/issues/1737: Separate input/output annotations
- https://github.com/istio/api/issues/1738: Remove unused annotations

PTAL

annotation/annotations.yaml
esnible: > are all of our annotations at pod level?

The `resources` field explains what the annotation can be legally be applied to.  (An istioctl analyzer checks this.). Istio has annotations for Pod, Service, Ingress, and "Any".

annotation/annotations.yaml
esnible: Is the idea that when we "promote" an annotation to beta it has two names for a period of time, the "alpha" name is deprecated?  Then eventually the "alpha" name goes away?
There is a TODO in the istioctl analyzer for annotations, to report the use of deprecated annotations, should we ever deprecate one.

annotation/annotations.yaml
```diff
@@ -79,46 +84,44 @@ annotations:
...
-    deprecated: false
-    hidden: false
-    resources:
-      - Pod
   - name: sidecar.istio.io/discoveryAddress
+    featureStatus: Alpha
     description: Specifies the XDS discovery address to be used by the Envoy
       sidecar.
-    deprecated: false
+    deprecated: true
```
esnible, line 98: This is the first annotation that we have deprecated.  There is a TODO in the analyzer code to report to the user the use of deprecated annotations.  Shall I create an issue and start work on reporting once this merges?

annotation/annotations.yaml
esnible: How would this be used?  Should `istioctl analyze` report, perhaps at INFO level, the use of alpha annotations?  Would the user want to see that?

annotation/annotations.yaml
nmittler: Yeah that sounds good! Thanks!

annotation/annotations.yaml
nmittler: The fact that we have annotations that start with `alpha.` is unfortunate and we don't want to continue that going forward. I think the hope is that we can simply get rid of the ones that are in that category today (not used?) and avoid that pattern in the future. The feature status will instead be a documentation-level artifact rather than being embedded in the name.

annotation/annotations.yaml
```diff
@@ -79,46 +84,44 @@ annotations:
...
       - Pod
   - name: sidecar.istio.io/rewriteAppHTTPProbers
+    featureStatus: Alpha
     description: Rewrite HTTP readiness and liveness probes to be redirected to
       the Envoy sidecar.
     deprecated: false
     hidden: false
     resources:
       - Pod
-  - name: sidecar.istio.io/controlPlaneAuthPolicy
```
louiscryan, line 88: Please don't delete things - instead let's add a featureStatus to capture that they are RETIRED. That way analyzers can still detect their use and warn folks. 

@esnible 
@therealmitchconnors same comment for ENV

annotation/annotations.yaml
louiscryan: If we can retire these outliers immediately that would be great. If we have to support them then a framework supported mechanism for aliasing them into a better name as @esnible suggests seems like the best choice

louiscryan: On a related note we could add some naming guidance to the header of this file. Want to take a stab @nmittler or do in another PR?

annotation/annotations.yaml
```diff
@@ -31,6 +33,7 @@ annotations:
     resources:
       - Service
   - name: alpha.istio.io/canonical-serviceaccounts
+    featureStatus: Alpha
```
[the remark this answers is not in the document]
louiscryan: @mandarjog isn't this for visualization, doesn't Kiali rely on this?

annotation/annotations.yaml
```diff
@@ -40,13 +43,15 @@ annotations:
     resources:
       - Service
   - name: alpha.istio.io/identity
+    featureStatus: Alpha
```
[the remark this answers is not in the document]
louiscryan: @myidpt any ideas?

annotation/annotations.yaml
louiscryan: Does this belong in the proxyConfig bucket? If so then keep at Alpha?

annotation/annotations.yaml
louiscryan: @nmittler @therealmitchconnors 
thinking about analyzer use-cases what do we think would be useful?

- Warn a user they have an annotation that is RETIRED in the current version
- Warn a user they are using an annotation that is deprecated in the current version
- Inform/Warn a user they are using an annotation that is Alpha

Where 'current version' can be substituted to mean the version they plan to upgrade to.

annotation/annotations.yaml
louiscryan: Proxy config has granularity issues though, how do we feel about the usability of it for use-cases like this vs annotations? Could we distill this evaluation into guidance?

annotation/annotations.yaml
nmittler: Let's add a `README.md` in a follow-on. I suspect the content to grow and evolve as we make progress here.

annotation/annotations.yaml
nmittler: sgtm

annotation/annotations.yaml
nrjpoddar: Interesting that there are no tests for this. I would keep it in alpha and possibly provide alternate ProxyConfig options if they are available.

annotation/annotations.yaml
nrjpoddar: @louiscryan can you clarify your comment about granularity issues? ProxyConfig can be overridden per workload so I must be missing something.

annotation/annotations.yaml
nrjpoddar: That’s a good question. Useful for disabling and can’t think why a user would want that?

annotation/annotations.yaml
nrjpoddar: I’m assuming deprecation would also indicate the new available supported option.

Another istioctl use case can be to take an input pod spec and mutate all deprecated annotations to their newer versions.

nrjpoddar:
@nmittler comments on this PR are growing but creating a good space for much needed conversation. How do you want to proceed? There are still few unresolved comments where we need inputs from WG leads to make an informed decision.

annotation/annotations.yaml
nrjpoddar: +1

annotation/annotations.yaml
incfly: Yes it's label & internal. We don't have any documentation expose this as an option AFAIK.

annotation/annotations.yaml
nmittler: Created https://github.com/istio/api/issues/1745 to address this.

annotation/annotations.yaml
nmittler: Yeah I saw at least one or two labels here. I'm planning on autogenerating the labels in the same way that we're doing here for annotations. After that I'll migrate this over.

annotation/annotations.yaml
incfly: Not sure if you refer to the same thing, but `rewriteAppHTTPProbe` is for application's own readiness check probe. This one is to ensure envoy sidecar get the minimum config from Istiod before marking itself ready.

cc @howardjohn @nmittler

annotation/annotations.yaml
nmittler: Created a tracking bug for adding `RETIRED`: https://github.com/istio/api/issues/1746

nmittler:
@nrjpoddar @sven @louiscryan @linsun 

I've created several tracking bugs from this, which were beyond the scope of this PR. I think I've done about as much as I can do here ... can I get approval?

smawson:
I defer to Louis and Lin, they should stamp it :)

smawson:
And Neeraj. Basically take me off your email list (hah).

nrjpoddar (approved):
@nmittler this is a good first pass for me so I will approve. Here are other updates that I think would be good to have as follow-on PRs:

* Many annotations are alpha and hidden. I don’t see a reason to keep them and we should mark them deprecated.
* Deprecate annotations which are supported via ProxyConfig
* File issues for maturing alpha annotations that we know are heavily used by following our feature enhancement process.

nmittler:
@nrjpoddar 

> Many annotations are alpha and hidden. I don’t see a reason to keep them and we should mark them deprecated.

I've updated the PR accordingly

nmittler:
@louiscryan @linsun mind rubber stamping so we can get this in?

annotation/annotations.yaml
louiscryan: How should this be presented then? Is this annotation deprecated/RETIRED?

annotation/annotations.yaml
louiscryan: I was worried about the granularity of the proxy-config API and that it entails a full-copy whereas annotations are like a field-level override of the default proxy-config. In general I think I agree with John that this can be moved into proxy config, that said I think the current usage likely makes this annotation in Beta so migration to proxy-config will take a long time. So Beta + deprecated ?

louiscryan (approved):
Lets convert the open comments to issues and add them as YAML comments here to unblock.

annotation/annotations.yaml
nmittler: That's right, I think the issue is that the term `featureStatus` is incorrect here ... it's really the status of the API. I think we should switch to just calling it `status`.

nmittler: Raised https://github.com/istio/api/issues/1794

annotation/annotations.yaml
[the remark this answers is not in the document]
nmittler: Raised https://github.com/istio/api/issues/1795

annotation/annotations.yaml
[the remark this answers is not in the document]
nmittler: Raised https://github.com/istio/api/issues/1796

annotation/annotations.yaml
nmittler: Raised https://github.com/istio/api/issues/1797

annotation/annotations.yaml
nmittler: Raised https://github.com/istio/api/issues/1798

annotation/annotations.yaml
nmittler: Raised https://github.com/istio/api/issues/1799

annotation/annotations.yaml
nmittler: Raised https://github.com/istio/api/issues/1800

annotation/annotations.yaml
nmittler: Raised https://github.com/istio/api/issues/1801

annotation/annotations.yaml
nmittler: Raised https://github.com/istio/api/issues/1802

annotation/annotations.yaml
nmittler: Raised https://github.com/istio/api/issues/1803

annotation/annotations.yaml
louiscryan: Agreed, 'featureStatus' is a loaded term here, 'status' is sufficient as long as we doc it properly

PR #1726, opened 2020-11-02
Manual make update-common and make gen

ericvn:
The automatic update #1725 failed due to make gen failing. This adds that step.


annotation/annotations.gen.go
```diff
@@ -301,6 +345,7 @@ var (
 		Name: "sidecar.istio.io/inject",
 		Description: "Specifies whether or not an Envoy sidecar should be "+
                         "automatically injected into the workload.",
+		FeatureStatus: Alpha,
```
howardjohn, line 348: why is this alpha @nmittler

howardjohn (approved):
approve since its running `make gen`, but this doesn't look right with marking everything alpha?

annotation/annotations.gen.go
nmittler: Everything defaults to alpha for now. I have a PR asking others to help fill in feature status of all annotations: https://github.com/istio/api/pull/1723

linsun (approved):

PR #1754, opened 2020-11-13
Mirror Destinationrule connectionPool to Sidecar

howardjohn:
Design
https://docs.google.com/document/d/1bIuwv0E98kyn8VZ_CDVGcZziIHyr3-ACiTmPqGgp5uY


networking/v1alpha3/sidecar.proto
```diff
@@ -448,6 +448,9 @@ message Sidecar {
   // detected defaults from the namespace-wide or the global default Sidecar.
   repeated IstioEgressListener egress = 3;
 
+  // Settings controlling the volume of connections to an upstream service
+  istio.networking.v1alpha3.ConnectionPoolSettings connection_pool = 7;
```
howardjohn, line 452: I think the open question here that I didn't notice in the original design is where exactly we put this. In destination rule, we have the top level setting and perPortSettings. However for Sidecar we don't have this. If we put it here, users cannot configure it per-port. If we put in in Ingress, then users need to opt into explicitly configuring ALL ports just to add connection pool settings

networking/v1alpha3/sidecar.proto
```diff
@@ -448,6 +448,9 @@ message Sidecar {
   // detected defaults from the namespace-wide or the global default Sidecar.
   repeated IstioEgressListener egress = 3;
 
+  // Settings controlling the volume of connections to an upstream service
```
howardjohn, line 451: I will update comment later

howardjohn:
@louiscryan @costinm can you take a look at this?

networking/v1alpha3/sidecar.proto
nrjpoddar: You need to add it in both places I'm guessing like how DR is structured.

PR #1755, opened 2020-11-13
Add config to enable application logs formatted for stackdriver

gargnupur:


mesh/v1alpha1/config.proto
```diff
@@ -168,6 +168,10 @@ message MeshConfig {
   // Default value is `false`.
   bool disable_envoy_listener_log = 56;
 
+  // This flag enables outputting application logs om a format compatible with
+  // stackdriver. See [Stacdriver Application Log](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/application_logging#stackdriver-logging-with-gke)
+  bool enable_stackdriver_formatted_app_logs = 58;
```
howardjohn, line 173: Does this apply only to Envoy logs? What about istiod/agent logs?

mesh/v1alpha1/config.proto
```diff
@@ -168,6 +168,10 @@ message MeshConfig {
   // Default value is `false`.
   bool disable_envoy_listener_log = 56;
 
+  // This flag enables outputting application logs om a format compatible with
+  // stackdriver. See [Stacdriver Application Log](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/application_logging#stackdriver-logging-with-gke)
```
howardjohn, line 172: Does SD support reading json logs rather than some magic `'%L%m%d %T.%e %t envoy] [%t][%n]%v'` format? Is there a reference onb the formats it can parse?

howardjohn:
```
  parsers.conf: |-
    [PARSER]
        Name        docker
        Format      json
        Time_Key    time
        Time_Format %Y-%m-%dT%H:%M:%S.%L%z

    [PARSER]
        Name        containerd
        Format      regex
        Regex       ^(?<time>.+) (?<stream>stdout|stderr) [^ ]* (?<log>.*)$
        Time_Key    time
        Time_Format %Y-%m-%dT%H:%M:%S.%L%z

    [PARSER]
        Name        json
        Format      json

    [PARSER]
        Name        syslog
        Format      regex
        Regex       ^\<(?<pri>[0-9]+)\>(?<time>[^ ]* {1,2}[^ ]* [^ ]*) (?<host>[^ ]*) (?<ident>[a-zA-Z0-9_\/\.\-]*)(?:\[(?<pid>[0-9]+)\])?(?:[^\:]*\:)? *(?<message>.*)$
        Time_Key    time
        Time_Format %b %d %H:%M:%S

    [PARSER]
        Name        glog
        Format      regex
        Regex       ^(?<severity>\w)(?<time>\d{4} [^\s]*)\s+(?<pid>\d+)\s+(?<source_file>[^ \]]+)\:(?<source_line>\d+)\]\s(?<message>.*)$
        Time_Key    time
        Time_Format %m%d %H:%M:%S.%L

    [PARSER]
        Name        network-log
        Format      json
        Time_Key    timestamp
        Time_Format %Y-%m-%dT%H:%M:%S.%L%z
```

Can we just use json logging?

mesh/v1alpha1/config.proto
gargnupur: I am hoping to apply it to istiod/agent logs too...

howardjohn: I would rather recommend users to use JSON, or convince the GKE team to add parsing for `zap` logs (our format) than adding another flag and complexity into mesh config to workaround the limitations of a single platform+logging ingestion pipeline.

howardjohn (changes requested):
I don't think SD-on-GKE logging limitations should impact our APIs like this

gargnupur:
Closing as we would be instead using log-as-json option.

douglas-reid:
@gargnupur can you validate that log-as-json will address the support concerns that were addressed by the glog format?  i'm concerned that we might lose some of the built-in bits around ease of connection tracing and envoy labelling that came with the recommended glog format string. if there are some defaults that we need to suggest/require, I think we could reconsider next steps here.

gargnupur:
@douglas-reid : Using log-as-json does solve problems of logs getting split into multiple lines and log not getting marked correctly as error if it is an error. Ex: https://screenshot.googleplex.com/AdyyxCxtAhDoGrM.

PR #1756, opened 2020-11-16
Add short name to EnvoyFilter

micnncim:
Added a short name to EnvoyFilter as well as other CRDs.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -1200,6 +1200,8 @@ spec:
     kind: EnvoyFilter
     listKind: EnvoyFilterList
     plural: envoyfilters
+    shortNames:
+    - ef
     singular: envoyfilter
   preserveUnknownFields: true
   scope: Namespaced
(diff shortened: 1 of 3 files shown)


howardjohn (commented):
lgtm

hzxuzhonghu (approved):

costinm:
Should we avoid taking over 2-letter names for alpha APIs ? I'm not sure how k8s reacts if different products register short names, 
and we may already have a problem with the k8s Gateways if they start registering short name. 

At least the 'short name' should have a failsafe setting, I suspect install would fail if there is a conflict ( I'm more concerned of what will happen with gateways for this @howardjohn )

Not blocking this, just thinking out loud.

howardjohn:
If there is a conflict k8s picks one of them to "win" when you type
`kubectl get ef`. I am not sure how it picks, it seems `kubectl get
gateway` always picks the k8s one. I opened an issue in k8s to make it give
a warning on conflict.

micnncim:
Thank you for point out the issue.

I've checked the related issues and PRs (https://github.com/kubernetes/kubernetes/issues/94860, https://github.com/kubernetes-sigs/service-apis/pull/426).

The best solution would be that k8s takes care of this issue like warning, right? Because we can't expect a conflict with other CRDs completely.

However, since it's not certain k8s will take care of it, would it be better to take more than 2 letter word like `evf`, `efl`?

smawson:
I wonder if we should have a policy against adding short names to alpha apis? Seems like it could be done as part of moving an api to beta. Thoughts?

hzxuzhonghu:
All our APIs are alpha now, there isn't any harm even conflict with other CRs, is there?

PR #1760, opened 2020-11-19
Base component only allows overlays field

shamsher31:
To fix https://github.com/istio/istio/issues/27240

Based on https://github.com/istio/istio/issues/27240#issuecomment-692059502,  https://github.com/istio/istio/issues/27240#issuecomment-692634227 and [this](https://github.com/istio/istio/blob/4e9e07c3fb322d947e896f88987fa6bf1db0d7d1/operator/cmd/mesh/testdata/manifest-generate/input/pilot_override_kubernetes.yaml#L13).

I'm not sure but I run `protolock commit --force` before I running `make gen`.


operator/v1alpha1/operator.proto
```diff
@@ -173,8 +173,14 @@ message BaseComponentSpec {
     // Selects whether this component is installed.
     TypeBoolValueForPB enabled = 1;
 
-    // Kubernetes resource spec.
-    KubernetesResourcesSpec k8s = 50;
+    // Overlays for k8s resources.
```
costinm, line 176: Is it backward compatible ? Upgrade story ?

costinm (changes requested):
Any change should include info about upgrade and migration.

Also since the operator API is likely to need major changes, to remove unused and make consistent with helm - we should evaluate carefully what we want to keep.

operator/v1alpha1/operator.proto
shamsher31: `BaseComponentSpec` only supported `overlay` and not the complete resource spec. Not sure if backward compatibility matters here?

shamsher31:
Please see the issue https://github.com/istio/istio/issues/27240 and let me know if this PR needs any updates. Thanks.

richardwxn (changes requested):
This change doesn't seem to solve the problem, it would only allow overlays field. I am not sure what went wrong but the original API definition looks correct and aligns with other component, so probably it is not an API issue but somewhere in the operator code.

shamsher31:
Closing as this is not a valid solution.

PR #1762, opened 2020-11-21
api: add `discovery_service` field to `ProxyConfig`

yskopets:
Signed-off-by: Yaroslav Skopets <yaroslav@tetrate.io>

## Summary

* this PR adds `discovery_service` field to `ProxyConfig`
* part of https://github.com/istio/istio/issues/29108


costinm (changes requested):
The discoveryAddress is broadly used, no reason to change to RemoteService while we are also looking to deprecated or replace RemoteService.

In addition, there are well-established practices for distributing the roots and configuring the mesh roots of trust - with subtle upgrade and migration implication.

IMO for 'external istiod' the default should be 'public certificates' ( ACME, etc) - we have a well established mechanism to distribute roots ( well, has a regression but it's getting fixed ), via the config map. I don't think injecting the roots of trusts via ProxyConfig is ideal - and it would require an RFC approved by both Security and Env WG ( and probably networking ).

costinm:
I would suggest instead starting with a doc ( RFC, etc) describing the current distribution mechanism and options for roots of trust, 
and if we want to make change, have a clear explanation of migration and how the new mechanism should look like. It's too critical change to make directly in the API.

PR #1770, opened 2020-12-04
Mark port as required field for health check

howardjohn:
Matches Kubernetes


mandarjog:
Nice you added shriram ;-)

hzxuzhonghu (approved):

hzxuzhonghu:
We have many other apis that does not set as required

costinm (commented):
1. Upgrade impact ?
2. Why required ? Most microservices and apps have a single http port - and we can default to it.
We should make the simple/common case easy, and advanced cases possible.

BTW - for VMs in particular using port number for endpoint is far more risky than for pods, due to collisions. K8S supports named ports - not sure what is the use case in k8s, but for VM it is more likely to be needed. 

Unless the port is the 'service port' - which is mapped to the actual endpoint port (may be different) on each workload. But the doc is 'endpoint' - which implies all workloads have the same
endpoint port and the API doesn't support named ports.

costinm:
BTW - can we create a bot that asks "What is the upgrade impact" on each PR that doesn't have a section "Upgrade impact" ?

howardjohn:
There is no upgrade impact, this is just changing the documentation and the feature is new in 1.9.

The api mirrors kubernetes exactly currently making this not required would be our first difference. I am not strictly against it but I also don't know how we could automatically determine the correct port.

This api is on the ""deployment "" level - like k8s all instances in the group are expected to be uniform. Kubernetes does not have a way to automatically detect a port to use even with named port (the named port is just referencing a mapping in the same pod spec, it's a convenience not an indirection)

costinm:
Got it, mirroring k8s exactly is good.

I think if a field is marked as 'required' in the doc the expectation is
that validators will reject CRDs with the
field not set - if that's not happening now it probably should. Which mean
adding a new 'required' field to an existing
 proto is not a  safe operation for upgrades.

But LGTM

howardjohn:
> 
> I think if a field is marked as 'required' in the doc the expectation is
> that validators will reject CRDs with the
> field not set - if that's not happening now it probably should. Which mean
> adding a new 'required' field to an existing
>  proto is not a  safe operation for upgrades.


Agree in general. In this case, this field is new in 1.9. I added the validation when I opened this PR, which is when I noticed the docs do not match the validator. So this just makesthe docs match

smawson:
What's the status on this PR? It LGTM but don't want to approve if it is out of date...

howardjohn:
Still desired, it is just making docs match the code.

howardjohn:
@smawson mind taking another look?

smawson (approved):

PR #1774, opened 2020-12-07
add destination port support for envoyfilter

zhaohuabing:
I'm working on [Aeraki](https://github.com/aeraki-framework/aeraki) to extend Istio's traffic management to other layer-7 protocols. When trying to build EnvoyFilters for inbound listeners, I find that we need a destination port match condition to distinguish multiple service ports in the virtual inbound listener.

Signed-off-by: zhaohuabing <huabingzhao@tencent.com>


zhaohuabing:
I got the following error when running "make gen"

```
Writing to output file operator/v1alpha1/operator.pb.go
sed: 1: "operator/v1alpha1/istio ...": invalid command code o
make: *** [operator/v1alpha1/operator_deepcopy.gen.go] Error 1
```

hzxuzhonghu:
We have Port match in EnvoyFilter_ListenerMatch, however, this does not work well with virtualinbound listener.

zhaohuabing:
> We have Port match in EnvoyFilter_ListenerMatch, however, this does not work well with virtualinbound listener.

We need a port match in the filter chain match since multiple ports are in different filter chains of the single virtual inbound listener.

howardjohn (changes requested):
We need to consider how this will work when we switch to single outbound listener.

Either keep the matching tied to the real implementation details, and *every* listener match EnvoyFilter needs to switch to filter chain match, or we have some indirection.

I don't like either option really - probably needs some more thought.

cc @stevenctl

zhaohuabing:
The biggest concern for me is that the structural and name changes of listeners and other xds resources would affect existing EnvoyFilters. Maybe we could have a way to abstract the structure of EnvoyFilter and standardize the resource names.

zhaohuabing:
Can we add the destination port？ It's a showstopper for me to continue working on Aeraki.
We can make it back compatible after we switch to a single outbound listener.
If there's a destination_port in the FilterChainMatch
     use the destination_port to match filterchain
else if there's a port_number in the ListenerMatch
    use the port_number to match the filterchain

howardjohn (approved):

howardjohn:
My original thought was this is not stable against the underlying generated Envoy config - which applies to ALL of envoyfilter, so I do think this makes sense. I would change the comment about it only applying to SIDECAR_INBOUND though - that is a coincidence of the fact that *today* we don't set destination_port in FCM. But that should not be a concern of the API.

zhaohuabing:
> My original thought was this is not stable against the underlying generated Envoy config - which applies to ALL of envoyfilter, so I do think this makes sense. I would change the comment about it only applying to SIDECAR_INBOUND though - that is a coincidence of the fact that _today_ we don't set destination_port in FCM. But that should not be a concern of the API.

You're correct. I remove that part of comment.

hzxuzhonghu:
This should also be added to v1alpha version

hzxuzhonghu (approved):
We have no beta version envoyfilter, so lgtm

zhaohuabing:
Hi guys, how could I proceed with this PR?

linsun (approved):
Thanks!  Look forward to seeing this work land

PR #1775, opened 2020-12-09
update sampling doc

linsun:
Since it is already in our user [doc](https://istio.io/latest/docs/tasks/observability/distributed-tracing/configurability/#customizing-trace-sampling), it makes sense to have this reflected in our generated API doc. 

Also, the default was wrong.

### DIFF
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
@@ -819,7 +819,7 @@
             "type": "integer"
           },
           "sampling": {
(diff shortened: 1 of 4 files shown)


linsun:
cc @rvennam who asked about this

hzxuzhonghu (approved):

louiscryan:
@douglas-reid any comment?

louiscryan (approved):

douglas-reid:
We were hiding this option initially in anticipation of a first-class API for controlling that value. Ditto with the custom tags in that doc. But I think it is OK to expose. It would be nice to be able to annotate it as experimental / deprecated in the near future.

PR #1779, opened 2020-12-16
Test add message field into AnalysisMessage

shamsher31:
Testing https://github.com/istio/api/pull/1776


shamsher31:
Works fine in https://github.com/istio/api/pull/1776. Closing it.

PR #1781, opened 2020-12-16
Cleanup generated files that we no longer generate

howardjohn:
These files are no longer used/generated, but still lingering since our cleaning logic isn't great (yet!). Removing these now to make future PR to improve the gen logic smaller


howardjohn:
@istio/technical-oversight-committee can you take a look? This is just code cleanup

smawson (approved):

hzxuzhonghu (approved):

PR #1782, opened 2020-12-16
WIP: move proto generation to prototool

howardjohn:
The idea here is to have consistency between all of our folders, and not have the same complex protoc commands in makefile. Instead we just have a simple yaml (more or less...)

built on https://github.com/istio/api/pull/1781, will be smaller if we merge that first

A side benefit of this is it will mean moving to golang protobuf from gogo protobuf is easier

Depends on some unmerged tweaks to istio/tools


howardjohn:
@kyessenov @jasonwzm  interested in early feedback if you think this is useful? prototool is technically deprecated in favor of https://github.com/bufbuild/buf, but `buf` is strongly opinionated in opinions that don't match our repo structure, and we are already using prototool for other things.

howardjohn:
> @kyessenov @jasonwzm interested in early feedback if you think this is useful? prototool is technically deprecated in favor of https://github.com/bufbuild/buf, but `buf` is strongly opinionated in opinions that don't match our repo structure, and we are already using prototool for other things.

any thoughts?

kyessenov:
I don't have a strong opinion since I didn't use either tool. IMHO we don't need more build complexity, bazel or makefile suffice.

jasonwzm:
I am okay with using a single tool and it seems prototool keeps protoc version almost in sync, so it should be fine. 

The only thing would be compatibility with golang proto compatiblity, since the protoc tools we have are mostly gogo. So I guess it would be change protoc tools to use golang -> change to rely on single prototool?

bufdev:
Just chiming in as the primary [author](https://github.com/uber/prototool/graphs/contributors) of [both](https://github.com/bufbuild/buf/graphs/contributors) - I'd *highly* recommend buf over prototool, a brief overview is in [buf's docs](https://docs.buf.build/migration-prototool/) but suffice to say, buf is actively maintained.

howardjohn:
@bufdev The main issue I had with `buf` was that it is opinionated in ways that don't match our opinions. Primarily, the folder layout requirement - I think it didn't like that we have multiple roots, and its explicitly mentioned in the doc this is intended. I can give it another shot with potentially constructing the desired folder layout on demand or via symlinks though.

bufdev:
You can use multiple roots with no issue: https://docs.buf.build/build-configuration#roots
You can also just disable the linter that does the file/package matching: https://docs.buf.build/lint-configuration#except
We added symlink support in v0.32.0 https://github.com/bufbuild/buf/releases/tag/v0.32.0

Happy to provide any other help.

howardjohn:
@bufdev thanks! I will give it another shot

bufdev:
No problem, let me know if you get stuck and I can take a pass (especially easy if you have a branch I can check out :-) )

howardjohn:
@jasonwzm 
> So I guess it would be change protoc tools to use golang -> change to rely on single prototool?

The plan is to migrate to the new tool (prototool in the original PR, I think it will be replaced with `buf` per ^ in a bit once I finish it). Then, later, we will move from gogo -> golang. The migration will be a single line here, and 1000s of lines in istio/istio

howardjohn:
Closing in favor of https://github.com/istio/api/pull/1833

PR #1785, opened 2020-12-18
Adding a global configuration field to control MCS serviceexport creation behavior

josephpeacock:
Added the `bool enable_mcs_service_export = 58;` field to support work on https://github.com/istio/istio/issues/29385

As per the readme, this commit contains the repo state after the field addition, and running `make clean gen` then `make proto-commit`.


howardjohn (commented):
Is this a temporary feature flag for is this expected to be the permanent API?

josephpeacock:
My understanding is that this flag will be in place until we move to phase 5 as per https://docs.google.com/document/d/1K8hvQ83UcJ9a7U8oqXIefwr6pFJn-VBEi40Ak-fwQtk/edit#heading=h.xw1gqgyqs5b

josephpeacock:
@nmittler may be able to shed a bit more light on what the final config state is expected to look like, I'm working on just the ServiceExport part of that RFC.

nmittler:
@howardjohn 

> Is this a temporary feature flag for is this expected to be the permanent API?

I suspect we're always going to need these levers to control the MCS integration level in Istio. Does that change where the config should go? My understanding was that meshconfig was an appropriate place for experimental features. If not, what is the current approach for new APIs?

smawson:
My slight preference is to leave this a bit more extensible by having something like a ServiceExportPolicy rather than having this as a flag in mesh config. This is something we had brought up with the MCS authors as a possible extension point, to control default behavior.

Also having this in mesh config is a bit weird if there are multiple revisions in a cluster, since you really want this as a global setting, not config for an individual control plane.

So I'd push for defining an alpha API called ServiceExportPolicy that allows for global and per-namespace control over export behavior.

nmittler:
@smawson @josephpeacock I've taken a crack at a proposed API: https://github.com/istio/api/pull/1870

nmittler:
@smawson I suspect the process for getting approval for a proper API will take some time. Can we just approve this for now to unblock the MCS effort?

howardjohn:
Why not just an environment variable if we just want a temporary feature
flag that we know is not the long term API?

nmittler:
@josephpeacock can you change your implementation to use an environment variable as @howardjohn suggested?

josephpeacock:
> @josephpeacock can you change your implementation to use an environment variable as @howardjohn suggested?

Yup, that should be a pretty minimal change. I'll work in it in the next few days

nmittler:
@josephpeacock should we abandon this change?

josephpeacock:
Yeah, I think it's made obsolete by the more formal API that will come out

PR #1792, opened 2021-01-05
Add configuration options for Workload generated certificates

jacob-delgado:
Issue: Clients are able to configure workload TTL and encryption type
only through environmental variables. The user should be able to
configure this with a supported API.

Solution: Support this ability in meshConfig.defaultConfig as it is the
responsibility of pilot-agent to request a certificate.


jacob-delgado:
## Pending approval of RFC https://docs.google.com/document/d/1KUQlAGnpajDJC9G6Em-OGGCn_xEftpw-J6TXcMB_MNE

PR #1806, opened 2021-01-08
Add default revision to operator

Monkeyanator:
Introduces `bool default_revision` as a top level field in the operator. This change (pending corresponding changes in `istio/istio`) will let us:

1. Fix validation for revisions (specify `istioctl install --revision canary --default-revision` and install will create `istiod-default` service from base pointing to revision)
2. Fix mutating webhook `istio-injection` targeting (we'll only create webhook targeting `istio-injection` if there is no revision specified or the revision is the default revision)

Not sure a top level field in the operator is the correct place for this, to me it makes sense keeping it consistent with where `revision` is defined but would appreciate suggestions.

Per environments WG discussion around revision tags @zerobfd @howardjohn


operator/v1alpha1/operator.proto
```diff
@@ -91,6 +91,10 @@ message IstioOperatorSpec {
     // This option is currently experimental.
     string revision = 16;
 
+    // Identify whether this revision is the default revision for the cluster
+    // This option is currently experimental.
+    bool default_revision = 17;
```
howardjohn, line 96: Do we want a `defaultRevision` field? Or a `tags` field, which can contain a special `default` value?

Monkeyanator: As I understand it, `tags` (as proposed, anyway) correspond directly to `MutatingWebhooks` pointing to revisions (purely concerned with injection). This is not quite what the `defaultRevision` here tries to address.

`defaultRevision` would just be there to give reasonable semantics around (1) where the `ValidationWebhook` points, and (2) which revision `istio-injection` labels will point to by default. It will do this by creating a new default `Service` rather than a `MutatingWebhookConfiguration`.

I can see where having `defaultRevision` and eventually `revisionTags` both in the operator might be confusing for users. At the same time, I'm a fan of keeping `revisionTags` relatively dumb and having them correspond 1-to-1 with mutating webhooks (rather than having `default` value that does something different). Also not sure how helm installations would be able to handle a special value in `revisionTags` (might be possible, not sure)

howardjohn: I agree they are two different concepts, the idea was just having one field (tags) with a special value (default), or having two fields (default_revision and tags). I don't have a strong opinion either way.

I don't think we need to let the implementation details around Service and webhooks drive the API though, we should find the best UX and make the implementation work for that (within reason :slightly_smiling_face: )

Monkeyanator: 100% agree we shouldn't let the implementation details drive the API here. From UX standpoint I think I'm still leaning towards having two fields (default_revision and tags), although both approaches make sense I think. @zerobfd any thoughts on this?

zerobfd: I think we should have two different fields: I can imagine someone wanting to do something like this:

0) mesh exists with prod and canary tags
1) start a canary deploy by installing new canary revision
2) change the default revision to point to "canary" to test things like validatingwebhook etc. that use the default revision
3) finally, tag the canary revision with "prod" and clean up

If all we have is a special name, I feel like we lose some flexibility.

operator/v1alpha1/operator.proto
```diff
@@ -91,6 +91,14 @@ message IstioOperatorSpec {
...
+
+    // Revision tags associated with this control plane installation
+    // This option is currently experimental.
+    repeated string revision_tags = 18;
```
zerobfd, line 100: I thought that the decision was to hold off on putting the tags in the Operator for the first iteration? Is this field necessary for the basic implementation?

zerobfd (commented):
Just because I'm not very familiar with all of the options: is IstioOperator the right place for the default_revision bool? Maybe there's no other higher-level resource for a singleton accessed by all control planes, but I'm worried that two revisions might get `default_revision` set to `true` at the same time.

operator/v1alpha1/operator.proto
howardjohn: How revision works (and so should this) is that it maps to to the helm values.revision_tags. So its not strictly required, as they can always use the helm passthrough API.

Monkeyanator: You're right, we can add this in later

smawson (changes requested):
I don't like this field showing up in operator spec because what happens if two specs both claim to be the default?

I'm fine with istioctl install to take a flag specifying that a particular revision is the default because it is imperative, and so last one wins, but that doesn't work well for declarative config.

I think we need a separate resource to indicate the default (and general tag mapping). Either a clsuter-scoped config map or a cluster-scoped singleton CRD to specify the mapping.

howardjohn:
@smawson I disagree and agree. I think we have effectively already done what you are proposing. Trying to work backwards here - adding new resources opens a lot of new problems. What really matters, at the end of the day, is the webhook configs. So we need *something* to actuate the configuration (whether its IstioOperator with istioctl, values.yaml with helm, or I suppose some custom controller with configmap).

I do not consider a custom controller/tool to do this viable, so that leaves IstioOperator or values.yaml.

I don't think making it only IstioOperator makes sense, as we want to support helm, so that leaves values.yaml (Optionally, IstioOperator can wrap around values.yaml, as proposed here).

If we want a global option, configured through helm values.yaml, that would mean we need dedicated chart for this that is shared between revisions and configures these global objects (webhooks). We actually already have exactly that - the `base` chart, which is what will be modified by the `default_revision` option.

For Helm, this works great, because the charts are decoupled. For IstioOperator, they are all bundled so its less great - but its *already* broken if they re-apply an old base (consider upgrading from 1.7 to 1.8, then tweaking the 1.7 IstioOperator - the `base` will drop the CRD versions from 1.8 to 1.7). So in some ways, this just adds to an existing issue, that `base` is a singleton but it isn't treated as such in the API. If we are going to make some sort of singleton, I think we may need to do a bigger change than just this single `default_revision` field.

I feel like I went in circles a bit, let me know if that made sense...

Monkeyanator:
@smawson agreed that ideally this would fit in better with the declarative API but IMO the problems this solves in the short-to-mid term offsets that. 

Having an `istiod-default` service in base that gets overwritten based on which installation claims to be `default` makes a lot of sense to in the mid-to-short term.  We could handle this solely in `istioctl` with `--default-revision` but this would leave users using operator and helm in the dust for revisioned upgrades for the near future.

Introducing custom resources for this mapping solves the declarative problem but adds quite a few others, and as John mentioned, I think adding a CRD that specifies all common resources between revisions would have to be a whole separate design (including but not limited to revision tags and default revision).

Revision-based upgrades, our recommended upgrade path, is currently [just broken](https://github.com/istio/istio/issues/27501) with validation and this seems to be the most pragmatic approach to solve that. The next best approach IMO would be revisioning the validationwebhook ([like here](https://github.com/istio/istio/pull/27578)) which would redirect validation from a stable install to a new revision

Looping in @costinm who I think has looked into this before

smawson:
Makes sense @howardjohn thanks. Given base overrides is already broken in OperatorSpec I'm OK with this as a temporary solution. I'd like to make sure we track extracting out the base parts from IstioOperatorSpec and include the revision map as part of that. Can someone file an issue tracking this and someone in environments take on fixing this problem?

At least this is marked experimental but I hate adding fields we know are in the wrong place...so lets see if we can fix it the right way sooner rather than later.

smawson (approved):
Approving, please start putting together the plan for getting base config out of operator spec.

zerobfd:
Opened #1834 to track removing the base config from operator spec.

operator/v1alpha1/operator.proto
linsun: Is it whoever specify the default_revision newest wins?  It seems possible for user to specify an IOP default revision while the cluster had a default revision already.

releasenotes/notes/default-revision.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: release-notes/v2
+kind: feature
+area: installation
+issue:
+  - 27501
+
+releaseNotes:
+  - |
+    **Added** `defaultRevision` field to the `IstioOperator` to allow users to specify which control plane revision
+    should handle validation and default injection.
```
linsun, line 10: Is this intended for server side operator as well?

howardjohn (approved):

joberdick:
@Monkeyanator Was this tested against CNI? :) It seems many times we can't use features when they are released because we use CNI injection for security reasons.

PR #1813, opened 2021-01-12
docs: Update sidecar docs to explain outbound traffic restrictions

howardjohn:
This clears up some very common misunderstandings around these fields

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -152,11 +152,18 @@ message MeshConfig {
 
   message OutboundTrafficPolicy {
     enum Mode {
-      // outbound traffic will be restricted to services defined in the
-      // service registry as well as those defined through ServiceEntries
+      // Outbound traffic will be restricted to services defined in the
+      // service registry as well as those defined through `ServiceEntry` configurations.
+      //
+      // Note: This configuration is not a security policy, but instead an explicit declaration of dependencies,
+      // which configures how the sidecar proxy responds to unknown destinations.
+      // This is helpful in scenarios where failing fast is preferred for misconfigurations or for preventing
+      // unintended dependencies.
+      // Applications can still bypass these restrictions, so if strict enforcement is
+      // required alternative mechanisms like Kubernetes `NetworkPolicy` should be configured.
       REGISTRY_ONLY = 0;
-      // outbound traffic to unknown destinations will be allowed, in case
-      // there are no services or ServiceEntries for the destination port
+      // Outbound traffic to any endpoint or service will be allowed, in case
+      // there are no services or `ServiceEntry` resources for the destination port.
       ALLOW_ANY = 1;
 
       reserved 2;
--- a/networking/v1alpha3/sidecar.proto
+++ b/networking/v1alpha3/sidecar.proto
@@ -586,6 +586,11 @@ message OutboundTrafficPolicy {
   enum Mode {
     // Outbound traffic will be restricted to services defined in the
     // service registry as well as those defined through `ServiceEntry` configurations.
+    //
+    // Note: These restrictions are not a security boundary, and instead just refer to the response
+    // to unknown resources. This is helpful to fail-fast to misconfigurations or avoid accidental calls
+    // to unknown destinations. Applications may still bypass these restrictions, so if enforcement is
+    // required alternative mechanisms are required.
     REGISTRY_ONLY = 0;
     // Outbound traffic to unknown destinations will be allowed, in case
     // there are no services or `ServiceEntry` configurations for the destination port.
--- a/networking/v1beta1/sidecar.proto
+++ b/networking/v1beta1/sidecar.proto
@@ -586,6 +586,11 @@ message OutboundTrafficPolicy {
   enum Mode {
     // Outbound traffic will be restricted to services defined in the
     // service registry as well as those defined through `ServiceEntry` configurations.
+    //
+    // Note: These restrictions are not a security boundary, and instead just refer to the response
+    // to unknown resources. This is helpful to fail-fast to misconfigurations or avoid accidental calls
+    // to unknown destinations. Applications may still bypass these restrictions, so if enforcement is
+    // required alternative mechanisms are required.
     REGISTRY_ONLY = 0;
     // Outbound traffic to unknown destinations will be allowed, in case
     // there are no services or `ServiceEntry` configurations for the destination port.
--- a/mesh/v1alpha1/config.pb.go
+++ b/mesh/v1alpha1/config.pb.go
@@ -180,11 +180,18 @@ func (MeshConfig_H2UpgradePolicy) EnumDescriptor() ([]byte, []int) {
 type MeshConfig_OutboundTrafficPolicy_Mode int32
 
 const (
-	// outbound traffic will be restricted to services defined in the
-	// service registry as well as those defined through ServiceEntries
+	// Outbound traffic will be restricted to services defined in the
+	// service registry as well as those defined through `ServiceEntry` configurations.
+	//
+	// Note: This configuration is not a security policy, but instead an explicit declaration of dependencies,
+	// which configures how the sidecar proxy responds to unknown destinations.
+	// This is helpful in scenarios where failing fast is preferred for misconfigurations or for preventing
+	// unintended dependencies.
+	// Applications can still bypass these restrictions, so if strict enforcement is
+	// required alternative mechanisms like Kubernetes `NetworkPolicy` should be configured.
 	MeshConfig_OutboundTrafficPolicy_REGISTRY_ONLY MeshConfig_OutboundTrafficPolicy_Mode = 0
-	// outbound traffic to unknown destinations will be allowed, in case
-	// there are no services or ServiceEntries for the destination port
+	// Outbound traffic to any endpoint or service will be allowed, in case
+	// there are no services or `ServiceEntry` resources for the destination port.
 	MeshConfig_OutboundTrafficPolicy_ALLOW_ANY MeshConfig_OutboundTrafficPolicy_Mode = 1
 )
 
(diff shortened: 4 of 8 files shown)


mesh/v1alpha1/config.proto
```diff
@@ -150,11 +150,16 @@ message MeshConfig {
 
   message OutboundTrafficPolicy {
     enum Mode {
-      // outbound traffic will be restricted to services defined in the
-      // service registry as well as those defined through ServiceEntries
+      // Outbound traffic will be restricted to services defined in the
+      // service registry as well as those defined through `ServiceEntry` configurations.
+      //
+      // Note: These restrictions are not a security boundary, and instead just refer to the response
+      // to unknown resources. This is helpful to fail-fast to misconfigurations or avoid accidental calls
```
mandarjog, line 157: It encourages explicit declaration of dependencies.

mesh/v1alpha1/config.proto
```diff
@@ -150,11 +150,16 @@ message MeshConfig {
...
+      // to unknown destinations. Applications may still bypass these restrictions, so if enforcement is
+      // required alternative mechanisms are required.
```
nrjpoddar, line 159:
```suggestion
      // Note: This configuration is not a security policy, but instead an explicit declaration of dependencies
      // which configures how the sidecar proxy responds to unknown destinations.
      // This is helpful in scenarios where failing fast is preferred for  misconfigurations or preventing
      // accidental requests to unknown destinations.
      // Applications can still bypass these restrictions, so if strict enforcement is
      // required alternative mechanisms like Kubernetes network policies should be configured.
```

mesh/v1alpha1/config.proto
```diff
@@ -150,11 +150,16 @@ message MeshConfig {
...
       REGISTRY_ONLY = 0;
-      // outbound traffic to unknown destinations will be allowed, in case
-      // there are no services or ServiceEntries for the destination port
+      // Outbound traffic to unknown destinations will be allowed, in case
+      // there are no services or `ServiceEntry` configurations for the destination port.
```
nrjpoddar, line 162:
```suggestion
      // there are no services or `ServiceEntry` resources for the destination port.
```

networking/v1beta1/sidecar.proto
```diff
@@ -546,14 +546,10 @@ message IstioEgressListener {
...
-  // be able to reach them otherwise. For example, add host
-  // `istio-system/istio-telemetry.istio-system.svc.cluster.local` if telemetry
-  // is enabled, `istio-system/istio-policy.istio-system.svc.cluster.local` if
-  // policy is enabled, or add `istio-system/*` to allow all services in the
-  // `istio-system` namespace. This requirement is temporary and will be removed
-  // in a future Istio release.
+  // **WARNING:** This list does not act as a firewall; destinations not listed may still be reached
+  // when `outboundTrafficPolicy` is set to `ALLOW_ANY`, or by bypassing the sidecar through various means.
+  // Instead, this acts only to restrict the set of configuration sent to the proxy. To secure control
+  // traffic, alternatives such as `NetworkPolicy` must be used.
```
nrjpoddar, line 552:
```suggestion
  // **WARNING:** This list is not a firewall or security policy. Destinations which are not listed here
  // can still be reached if the `outboundTrafficPolicy` is set to `ALLOW_ANY`, or by bypassing the
  // sidecar proxy traffic interception.
  // Instead, this allows you to restrict the set of configuration sent to the proxy and should only used
  // for this purpose. To securely control traffic, alternatives such as Kubernetes `NetworkPolicy` should be used.
```

nrjpoddar (changes requested):

networking/v1beta1/sidecar.proto
nrjpoddar: How does NetworkPolicy help here?

howardjohn: Only allow traffic to the egress gateway, then it cannot be bypassed: https://istio.io/latest/docs/tasks/traffic-management/egress/egress-gateway/#apply-kubernetes-network-policies

mesh/v1alpha1/config.proto
louiscryan: I would suggest a phrasing which talks about 'preventing unintended dependencies' rather than 'unknown desintations'

It's also probably appropriate to say that while this is not a hard security boundary it can be a useful part of a layered security approach as it allows for auditing and other security related controls

mesh/v1alpha1/config.proto
```diff
@@ -150,11 +150,16 @@ message MeshConfig {
...
+      // service registry as well as those defined through `ServiceEntry` configurations.
+      //
+      // Note: These restrictions are not a security boundary, and instead just refer to the response
+      // to unknown resources. This is helpful to fail-fast to misconfigurations or avoid accidental calls
+      // to unknown destinations. Applications may still bypass these restrictions, so if enforcement is
+      // required alternative mechanisms are required.
       REGISTRY_ONLY = 0;
-      // outbound traffic to unknown destinations will be allowed, in case
-      // there are no services or ServiceEntries for the destination port
+      // Outbound traffic to unknown destinations will be allowed, in case
```
louiscryan, line 161: s/unknown destination/any endpoint or service ?

networking/v1alpha3/sidecar.proto
```diff
@@ -547,14 +547,10 @@ message IstioEgressListener {
...
   //
-  // **WARNING:** The list of egress hosts in a `Sidecar` must also include
-  // the Mixer control plane services if they are enabled. Envoy will not
-  // be able to reach them otherwise. For example, add host
-  // `istio-system/istio-telemetry.istio-system.svc.cluster.local` if telemetry
-  // is enabled, `istio-system/istio-policy.istio-system.svc.cluster.local` if
-  // policy is enabled, or add `istio-system/*` to allow all services in the
-  // `istio-system` namespace. This requirement is temporary and will be removed
-  // in a future Istio release.
+  // **WARNING:** This list does not act as a firewall; destinations not listed may still be reached
```
louiscryan, line 550: @justinpettit You probably want to review too

howardjohn:
@mandarjog @nrjpoddar @louiscryan I updated the PR, thanks for the suggestions.

networking/v1alpha3/sidecar.proto
nrjpoddar: Can we not use the word firewall but instead similar wording as used in MeshConfig comments?

networking/v1alpha3/sidecar.proto
```diff
@@ -547,14 +547,10 @@ message IstioEgressListener {
...
+  // when `outboundTrafficPolicy` is set to `ALLOW_ANY`, or by bypassing the sidecar through various means.
+  // Instead, this acts only to restrict the set of configuration sent to the proxy. To secure control
```
nrjpoddar, line 552:
```suggestion
  // Instead, this acts only to restrict the set of configuration sent to the proxy. To securely control
```

networking/v1alpha3/sidecar.proto
```diff
@@ -547,14 +547,10 @@ message IstioEgressListener {
...
+  // traffic, alternatives such as `NetworkPolicy` must be used.
```
nrjpoddar, line 553:
```suggestion
  // traffic, alternatives such as Kubernetes `NetworkPolicy` must be used.
```

howardjohn:
@nrjpoddar PTAL, I think i have addressed all comments

mesh/v1alpha1/config.proto
```diff
@@ -152,11 +152,18 @@ message MeshConfig {
 
   message OutboundTrafficPolicy {
     enum Mode {
-      // outbound traffic will be restricted to services defined in the
-      // service registry as well as those defined through ServiceEntries
+      // Outbound traffic will be restricted to services defined in the
```
linsun, line 155: when registry_only is enabled, correct?

howardjohn: this comment is for registry only field

mesh/v1alpha1/config.proto
```diff
@@ -152,11 +152,18 @@ message MeshConfig {
...
+      // service registry as well as those defined through `ServiceEntry` configurations.
+      //
+      // Note: This configuration is not a security policy, but instead an explicit declaration of dependencies,
+      // which configures how the sidecar proxy responds to unknown destinations.
+      // This is helpful in scenarios where failing fast is preferred for misconfigurations or for preventing
+      // unintended dependencies.
+      // Applications can still bypass these restrictions, so if strict enforcement is
+      // required alternative mechanisms like Kubernetes `NetworkPolicy` should be configured.
```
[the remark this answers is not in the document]
howardjohn: authz policy cannot block egress traffic since they can just entirely bypass the sidecar

mesh/v1alpha1/config.proto
linsun, line 155:
```suggestion
      // When `REGISTRY_ONLY` is configured, outbound traffic will be restricted to services defined in the
```

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
linsun: I see, is this for your change in sidecar.proto also or just for external traffic?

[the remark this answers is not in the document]
howardjohn: That is just how authz works, its only on the inbound side.

mesh/v1alpha1/config.proto
howardjohn: We don't follow this form in other APIs, for example https://istio.io/latest/docs/reference/config/networking/destination-rule/#ClientTLSSettings-TLSmode

nrjpoddar (approved):

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
linsun: Thanks!  Makes total sense now

linsun (approved):

PR #1815, opened 2021-01-12
Update common-protos

howardjohn:
Pulls in update from common files. A bunch of files we don't use were
removed. For some reason the locking mechanism is asserting stuff in
common-protos when it probably shouldn't. I can look into this, but for
now just updated the locks.


sdake (approved):
LGTM. Please make sure to update the API makefile as well. I have a few PRs in this area you are free to borrow from. The makefile is the only thing that really needs changing for the API repo (and then a regen).

Cheers,
-steve

sdake:
> LGTM. Please make sure to update the API makefile as well. I have a few PRs in this area you are free to borrow from. The makefile is the only thing that really needs changing for the API repo (and then a regen).
> 
> Cheers,
> -steve

My suggested update should be a different PR.

smawson (approved):

PR #1817, opened 2021-01-13
Define Istio Stability Levels

jasonwzm:
This is based on current discussions and design from https://docs.google.com/document/d/1ZcTIdHYNnLpwgLJa8tWg0YwsyHAmc_4lXguLmMQvAj0/edit?resourcekey=0-6OWzJ5cnlh_oYl4gFMbOuA#heading=h.lrto46ego5qb

Comments are warmly welcome!

/cc @nmittler @therealmitchconnors @douglas-reid @brian-avery

### DIFF
--- a/feature/stability.proto
+++ b/feature/stability.proto
@@ -1,0 +1,27 @@
+// Copyright 2020 Istio Authors
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.feature;
+
+option go_package="istio.io/api/feature";
+
+// Level defines the stability level naming Istio uses.
+enum Level {
+    EXPERIMENTAL = 0;
+    ALPHA = 1;
+    BETA = 2;
+    STABLE = 3;
+}
--- a/feature/stability.proto
+++ b/feature/stability.proto
@@ -1,4 +1,4 @@
-// Copyright 2020 Istio Authors
+// Copyright 2021 Istio Authors
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
--- a/stability/level.proto
+++ b/stability/level.proto
@@ -1,4 +1,4 @@
-// Copyright 2021 Istio Authors
+// Copyright Istio Authors
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
@@ -14,11 +14,11 @@
 
 syntax = "proto3";
 
-package istio.feature;
+package istio.stability;
 
-option go_package="istio.io/api/feature";
+option go_package="istio.io/api/stability";
 
-// Level defines the stability level naming Istio uses.
+// Level defines the stability levels Istio uses.
 enum Level {
     EXPERIMENTAL = 0;
     ALPHA = 1;


feature/stability.proto
```diff
@@ -0,0 +1,27 @@
+// Copyright 2021 Istio Authors
```
mandarjog, line 1: nit: no year per new style.

feature/stability.proto
```diff
@@ -0,0 +1,27 @@
...
+
+option go_package="istio.io/api/feature";
+
+// Level defines the stability level naming Istio uses.
+enum Level {
+    EXPERIMENTAL = 0;
+    ALPHA = 1;
+    BETA = 2;
+    STABLE = 3;
+}
```
mandarjog, line 27: is `deprecated` and `retired` left out on purpose?

Should the doc have definitions of what these mean? I think this should be authoritative source.

feature/stability.proto
```diff
@@ -0,0 +1,27 @@
...
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.feature;
+
+option go_package="istio.io/api/feature";
+
+// Level defines the stability level naming Istio uses.
+enum Level {
```
mandarjog, line 22: We need tooling (apart from what is described in the doc) to ensure that a fields do not make backwards progress.
So a stable field cannot go experimental....

feature/stability.proto
esnible: Should we have 
UNKNOWN=0, so that the default isn't EXPERIMENTAL if forgotten?

feature/stability.proto
```diff
@@ -0,0 +1,27 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+syntax = "proto3";
+
+package istio.feature;
```
nmittler, line 17: Do we want this enumeration to be used for both features and API fields? If so, I think we should hold off on this PR until we discuss this distinction in the TOC meeting.

feature/stability.proto
nmittler: @mandarjog deprecated and retired are missing intentionally. The idea is that you don't want to lose the last stability level, even after deprecation/retirement. Also, deprecation and retirement may have additional information that we'd like to tag along, such as the date that it was deprecated/retired, links for more information, etc.

feature/stability.proto
jasonwzm: Yea, I think so and that is my current thinking. So feature may not be a good name here. Will wait till the TOC meeting and make changes if necessary.

feature/stability.proto
jasonwzm: Yes, that is the plan.

feature/stability.proto
jasonwzm: @esnible Unknown sounds to me will give to much flexibility to the statuses. All APIs should have one of the four stabilities.

nmittler: > @esnible Unknown sounds to me will give to much flexibility to the statuses. All APIs should have one of the four stabilities.

I tend to agree with @jasonwzm here. Defaulting to Experimental seems like a logical choice. Regardless of level, the team should re-evaluate feature status periodically (at release time?) anyway. Using "Unknown" as a signal doesn't really add much in that case.

howardjohn: I don't think we should default to Experimental. We can either default to unknown, or we can make it a required to specify an option (which, due to lack of distinguishing case `0`, mostly means we need unknown afaik?).

It does not make sense to mark some core feature "experimental" just because someone forgot to label it.

nmittler: To be clear, the intent here isn't really `feature status`, which has been part of the issue when discussing these things. We're really talking about versioning individual fields (e.g. annotations, proto fields). By marking them all as experimental, we're (hopefully) forcing the team to look through them and assign (more) accurate values. I'm not sure how marking things as `Unknown` would have any benefit over simply assuming the worst and marking them `Experimental`. Since the team is generally trying move in the direction of promotion, `Experimental` should be an indicator that we need to make a decision to either work on promotion or deprecate.

howardjohn: What is a user going to think if they see "VirtualService.match is experimental"? That would not be great.

We should either
* REALLY force all fields to be set, ie this PR is not even merged until its filling out every field.
* allow some period of time when things are not "experimental", but "unknown"

otherwise I worry we are telling users the wrong thing. A field didn't become experimental just because our team is too busy to deal with another process of labeling things. I worry this will become just like the "features label", which has had near-zero usage/adoption, and now all of our fields will be "experimental" and be super confusing for users.

FWIW, not having an UNKNOWN field in enum has caused us pain over and over in the past

howardjohn: Put another way, I would rather tell the user nothing than tell them the wrong thing.

nmittler: Understand, but I still don't see how `Unknown` helps. It's the same problem if a field is listed as `Alpha` but it's really `Stable`. Ultimately the Istio team has to continually go through our API and make sure the values are correct for each release.

Regarding user visibility, there is no plan on making any of this user-visible until we're confident that it won't generate a ton of spam and FUD. Any tooling will build will be opt-in initially.

douglas-reid: fwiw, UNKNOWN would provide an escape hatch, particularly for currently deprecated and retired fields. It would allow labeling of fields that never went through any sort of formal level assignment and to which assigning any meaningful level would be inappropriate. I don't know how many cases we have like that, however. But, this would be on top of allowing incremental assessment and updating.

jasonwzm: > It would allow labeling of fields that never went through any sort of formal level assignment and to which assigning any meaningful level would be inappropriate.

@douglas-reid Do you have an example for this? That would be super helpful in determining what we can do in the short term and plan in the long term. My understanding is that we are able to label all fields with a stability.

douglas-reid: Maybe: https://github.com/istio/api/blob/master/mesh/v1alpha1/proxy.proto#L334-L338 (and similar) ?

Are those fields even still supported? Is there an appropriate level for them? I don't know how to answer some of these questions. For some, there were probably tests + other bits that merited a Beta / Stable status, but those are likely long deleted or refactored. Would we still feel comfortable assigning a firm status to them?

jasonwzm: In the case where the API is *already* deprecated, I think what we will do is that make best effort (through WG determination) for mark the stability level and the API also needs to be marked deprecated or retired. At that point, stability level is not that important to users when the field is already deprecated, or even retired.

smawson (approved):

nmittler (approved):

PR #1818, opened 2021-01-13
Remove auth/v1alpha auth policy

jacob-delgado:
These were deprecated and hidden from the docs, but the functionality was removed in 1.6.


howardjohn (approved):
Nice! probably the makefile needs an update as well unless it was removed from there?

smawson (approved):

nrjpoddar (approved):
Thanks @jacob-delgado!

PR #1822, opened 2021-01-13
Generate protos

jacob-delgado:
The latest build image uses gogo/protobuf 1.3.2. 

Only did the following.

1. make clean gen
2. make proto-commit

### DIFF
--- a/proto.lock
+++ b/proto.lock
@@ -36039,28 +36039,6 @@
                 "type": "string"
               },
               {
-                "id": 10,
-                "name": "auth_policy",
-                "type": "AuthPolicy",
-                "options": [
-                  {
-                    "name": "deprecated",
-                    "value": "true"
-                  }
-                ]
-              },
-              {
-                "id": 11,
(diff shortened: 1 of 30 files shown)


proto.lock
```diff
@@ -36038,28 +36038,6 @@
                 "name": "ingress_selector",
                 "type": "string"
               },
-              {
```
howardjohn, line 36041: if this fails in CI, try running `export LC_ALL=C` before you do it. I had a similar issue that I didn't debug, but I think their may be inconsistent results even in build container

howardjohn (approved):

proto.lock
jacob-delgado: Ok. Thanks. This may be an issue I was seeing in my other PR. Very frustrating, but that is a good fix. If I need that I'll add the instruction to the README.md

smawson (approved):

nrjpoddar (commented):
Thanks @jacob-delgado for this work!

PR #1830, opened 2021-01-15
add discoverySelectors to meshConfig

harveyxia:


mesh/v1alpha1/config.proto
```diff
@@ -533,8 +534,13 @@ message MeshConfig {
   // can be used with an extension provider to delegate the authorization decision to a custom authorization system.
   repeated ExtensionProvider extension_providers = 57;
 
+  // A list of kubernetes selectors that specify the set of namespaces that istio processes.
+  // Elements in the list are disjunctive, i.e. a namespace will be included if it matches any selector.
```
howardjohn, line 538: Maybe include an example? And a link to k8s docs on how the selector works could help.

howardjohn (approved):

mesh/v1alpha1/config.proto
```diff
@@ -533,8 +534,13 @@ message MeshConfig {
...
+  // If omitted, istio will use the default behavior of processing all namespaces.
+  repeated k8s.io.apimachinery.pkg.apis.meta.v1.LabelSelector discovery_selectors = 58;
```
hzxuzhonghu, line 540: Why `repeated` since metav1.LabelSelector already can represent multi matching labels

howardjohn: You can not represent all possibilities with a single one. For example, `foo=bar OR baz=abc` cannot be expressed

hzxuzhonghu: Got it from the proposal, wouldn't this make it so complex? What's the real use case then

howardjohn:
@louiscryan can you take a look?

christian-posta:
@hzxuzhonghu @linsun @louiscryan @nrjpoddar can someone please move this along?

hzxuzhonghu:
IMO, https://github.com/istio/api/pull/1830#discussion_r559280106 this is my only concern

mesh/v1alpha1/config.proto
```diff
@@ -530,8 +531,29 @@ message MeshConfig {
   // can be used with an extension provider to delegate the authorization decision to a custom authorization system.
   repeated ExtensionProvider extension_providers = 57;
 
+  // A list of kubernetes selectors that specify the set of namespaces that istio processes.
```
nrjpoddar, line 534: What does processes mean here? Can you add more details?

mesh/v1alpha1/config.proto
```diff
@@ -530,8 +531,29 @@ message MeshConfig {
...
+  // Elements in the list are disjunctive, i.e. a namespace will be included if it matches any selector.
```
nrjpoddar, line 535:
```suggestion
  // Elements in the list are disjunctive (OR semantics), i.e. a namespace will be included if it matches any selector.
```

releasenotes/notes/mesh-config-discovery-selectors.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+issue:
+  - https://github.com/istio/istio/issues/26679
+
+releaseNotes:
+  - |
+    **Added** Add discovery selectors to mesh config, a disjunctive list of k8s selectors used to restrict the namespaces that istiod processes.
```
nrjpoddar, line 9: This is quite hard for any user to parse, suggest simplifying the description. let's not use K8s here but expanded Kubernetes for release notes.

mesh/v1alpha1/config.proto
harveyxia: I've revised this add additional detail (though avoiding the word "process" is difficult without getting into implementation details). Let me know how the revised copy sounds, I'm happy to take any suggestions you have

releasenotes/notes/mesh-config-discovery-selectors.yaml
harveyxia: I've revised it, let me know how if it reads better

mesh/v1alpha1/config.proto
harveyxia: @hzxuzhonghu We discussed this extensively with the networking WG. Like John said, a disjunctive _list_ is needed to allow for selecting all possible subsets of namespaces. One use case for this is if different automation tooling or different organizational entities assign different labels to different namespaces.

mesh/v1alpha1/config.proto
```diff
@@ -530,8 +531,31 @@ message MeshConfig {
   // can be used with an extension provider to delegate the authorization decision to a custom authorization system.
   repeated ExtensionProvider extension_providers = 57;
 
+  // A list of Kubernetes selectors that specify the set of namespaces that Istio considers when
+  // computing configuration updates for sidecars. This can be used to reduce Istio's computational load
+  // by limiting the number of entities (including services, pods, and endpoints) that are watched and processed.
+  // If omitted, Istio will use the default behavior of processing all namespaces.
```
nrjpoddar, line 537:
```suggestion
  // If omitted, Istio will use the default behavior of processing all namespaces in the cluster.
```

nrjpoddar (approved):

PR #1831, opened 2021-01-16
ext-authz: add a few more header manipulation capability and also unhide the doc

yangminzhu:
This PR allows some more control of the header between ext-authz filter and the external server. At least the prefix matching will be super useful in many cases and it has already been used by users, e.g. https://discuss.istio.io/t/cannot-get-envoyfilter-ext-authz-to-work/7095.

cc @howardjohn @mandarjog @costinm


yangminzhu:
@louiscryan @smawson @linsun @nrjpoddar 
Could anyone take a look at the change? I'm hoping to get it for the 1.9 release. The change is small and natural extension of the current semantics, no major changes. Thanks.

mesh/v1alpha1/config.proto
```diff
@@ -486,25 +485,44 @@ message MeshConfig {
       // Note that in addition to the headers supplied by users:
       // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
       // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      // Prefix, suffix and substring matching is supported by using "*" in the format of "prefix*", "*suffix" and "*substring*".
```
costinm, line 488: *substring* seems to be a new pattern that no other Istio APIs supports AFAIK. Policy doesn't seem to have it, risky to diverge.

mesh/v1alpha1/config.proto
```diff
@@ -486,25 +485,44 @@ message MeshConfig {
...
       // Note, any existing headers will be overridden.
+      // Prefix, suffix and substring matching is supported by using "*" in the format of "prefix*", "*suffix" and "*substring*".
       repeated string headers_to_upstream_on_allow = 7;
 
+      // List of headers from the authorization service that should be added or appended in the original request and
+      // forwarded to the upstream when the authorization check result is allowed (HTTP code 200).
+      // If not specified, the original request will not be modified and forwarded to backend as-is.
+      // Note, any existing headers will be appended.
+      // Prefix, suffix and substring matching is supported by using "*" in the format of "prefix*", "*suffix" and "*substring*".
+      repeated string append_headers_to_upstream_on_allow = 10;
```
costinm, line 513: The naming is very confusing - I doubt users will understand what 'allow' and 'check' mean - that's an implementation detail of how the API is implemented ( making an API call named 'check' I assume ).

mesh/v1alpha1/config.proto
yangminzhu: good point, removed to only support prefix/suffix.

mesh/v1alpha1/config.proto
```diff
@@ -486,25 +485,44 @@ message MeshConfig {
       // Note that in addition to the headers supplied by users:
       // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
       // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      // Prefix, suffix and substring matching is supported by using "*" in the format of "prefix*", "*suffix" and "*substring*".
       repeated string include_headers_in_check = 6;
 
+      // HeaderValue defines the header name and value pair.
+      message HeaderValue {
```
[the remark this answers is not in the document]
yangminzhu: changed to use map<string, string> to match the VS pattern.

mesh/v1alpha1/config.proto
yangminzhu: the naming is to be consistent with the other field `headers_to_upstream_on_allow`, the only difference is this field will append instead of override when the same header is found.

Issue #1678, opened 2020-10-06
`make gen` is not deterministic


howardjohn:
It is an extremely common issue where `make gen` is not matching CI. I think its some makefile madness where it detects files are not changed so it skips them, but the makefile does not properly declare all dependencies (the version of the tools, for example) so they are skipped.

We need to make it so that `make gen` ALWAYS matches the output of CI, or the dev ex is horrible.

howardjohn:
Example:

I ran `make gen` after a tooling update to the json generated output. CI shows diff:
```
	modified:   analysis/v1alpha1/message_json.gen.go
	modified:   authentication/v1alpha1/policy_json.gen.go
	modified:   meta/v1alpha1/status_json.gen.go
	modified:   security/v1alpha1/ca_json.gen.go
	modified:   type/v1beta1/selector_json.gen.go
```

Then I deleted all of the json.gen.go (`find | rg json.gen.go | xargs rm)`, and reran. 

Now, I get a bunch of deleted files!:
````
        deleted:    analysis/v1alpha1/message_json.gen.go
        modified:   authentication/v1alpha1/policy_json.gen.go
        deleted:    meta/v1alpha1/status_json.gen.go
        deleted:    operator/v1alpha1/operator_json.gen.go
        deleted:    security/v1alpha1/ca_json.gen.go
        deleted:    security/v1beta1/authorization_policy_json.gen.go
        deleted:    security/v1beta1/jwt_json.gen.go
        deleted:    security/v1beta1/peer_authentication_json.gen.go
        deleted:    security/v1beta1/request_authentication_json.gen.go
        deleted:    type/v1beta1/selector_json.gen.go
```

Based on this, I think the generation code for security, operator, and analysis has some issues.

sdake:
The idea of `make gen` is that `make clean` is always run prior to `make gen`. This will result in deterministic results. It is not ideal, although this is a limitation of the build system. One option is to enforce a dependency on clean prior to `make gen`. In CI and developer environments.

jasonwzm:
+1 to Steven's comment. Running `make clean` before `make gen` is always suggested.

kyessenov:
Sorry for bringing this is up, but this is a solved problem with bazel.

howardjohn:
https://github.com/istio/api/pull/1833

PR #1833, opened 2021-01-20
Move protobuf generation/configuration to buf

howardjohn:
https://docs.buf.build/

Buf is the successor to https://github.com/uber/prototool which we
already use for linting.

This dramatically simplifies our Makefiles, which are both extremely
complicated and have led to numerous bugs historically, such as
https://github.com/istio/api/issues/1678.

This will make changes to the generation much simpler as well. For
example, to migrate to gogo protobuf, we will just need to change `gogo`
-> `go` in one location, rather than trying to wrangle 500 lines of
Makefiles. Additionally, its quite a bit faster - the whole proto stuff
is done in <1s now.


gen.sh
```diff
@@ -0,0 +1,58 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+set -eu
+
+# TEMPORARY! DO NOT MERGE. SHOULD ADD TO BUILD CONTAINER
```
howardjohn, line 19: I put a hold on this. If we want to merge this, I will add buf to the build container and remove this code and remove my hold.

networking/v1alpha3/sidecar.pb.go
```diff
@@ -949,50 +949,50 @@ func init() { proto.RegisterFile("networking/v1alpha3/sidecar.proto", fileDescri
...
-	0x92, 0x8c, 0xa0, 0x13, 0x08, 0xfe, 0x71, 0xae, 0xaa, 0xd0, 0x17, 0xb6, 0xe4, 0xf0, 0x7a, 0x75,
-	0xd9, 0x7a, 0xc2, 0xed, 0xc7, 0xd4, 0x5a, 0x1d, 0x6a, 0xaa, 0x35, 0x6b, 0x70, 0xd5, 0x6e, 0xbf,
-	0xee, 0x0c, 0x86, 0xf6, 0xa9, 0xd3, 0xeb, 0x5a, 0xa7, 0xc6, 0x12, 0x5a, 0x85, 0xe5, 0x96, 0x65,
-	0xf5, 0x4e, 0x9c, 0x56, 0xf7, 0xd4, 0x00, 0xbb, 0x4d, 0xa8, 0xdf, 0x68, 0x2c, 0xd2, 0x61, 0xf1,
-	0xb0, 0x7d, 0xd4, 0x7a, 0x6b, 0x0d, 0x8d, 0x25, 0xb4, 0x02, 0x4b, 0x9d, 0xfe, 0xb0, 0x75, 0x60,
-	0xb5, 0x07, 0x06, 0x40, 0x25, 0xa8, 0x75, 0x7b, 0xdd, 0xb6, 0x91, 0x3b, 0x30, 0xbf, 0x5c, 0x6e,
-	0x83, 0xaf, 0x97, 0xdb, 0xe0, 0xe2, 0x72, 0x1b, 0xbc, 0xaf, 0x26, 0xf5, 0x51, 0xae, 0xb6, 0x78,
-	0xc6, 0x86, 0x1d, 0x15, 0xd4, 0x46, 0xdd, 0xfb, 0x19, 0x00, 0x00, 0xff, 0xff, 0xaa, 0x40, 0x86,
-	0xbf, 0x2c, 0x06, 0x00, 0x00,
+	0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x02, 0xff, 0xbc, 0x54, 0x4d, 0x6f, 0xda, 0x4a,
```
howardjohn, line 952: buf has a slightly different compiler that generates a slightly different descriptor. This does not impact Istio at all, its just adding some additional metadata

bufdev: Just some context for others - we integration test this against thousands of files continuously for equivalence, and haven't had an actual difference (as represented by Google's comparison library) in over a year - getting to actual byte equivalence is virtually impossible for a myriad of reasons, although we tried

bufdev:
Let me know if we can be of any assistance here, and I'll take a glance through in a bit!

buf.gen-noncrd.yaml
```diff
@@ -0,0 +1,10 @@
+version: v1beta1
+plugins:
+- name: gogofast
+  out: .
+  opt: plugins=grpc,paths=source_relative,Mgogoproto/gogo.proto=github.com/gogo/protobuf/gogoproto,Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/descriptor.proto=github.com/gogo/protobuf/protoc-gen-gogo/descriptor,Mgoogle/protobuf/duration.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/struct.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/wrappers.proto=github.com/gogo/protobuf/types,Mgoogle/rpc/status.proto=istio.io/gogo-genproto/googleapis/google/rpc,Mgoogle/rpc/code.proto=istio.io/gogo-genproto/googleapis/google/rpc,Mgoogle/rpc/error_details.proto=istio.io/gogo-genproto/googleapis/google/rpc,Mgoogle/api/field_behavior.proto=istio.io/gogo-genproto/googleapis/google/api
```
bufdev, line 5: Just a note: you can put these on separate lines to make this more readable, `buf generate` accepts both string and list<string> for opt:

```yaml
opt:
  - plugins=grpc
  - paths=source_relative
  - Mfoo=bar
  - ...
```

This results in equivalent behavior, and does have precedent with `protoc` - the following three commands are equivalent:

```
protoc -I . --foo_out=. --foo_opt=one --foo_opt=two,three $(find . -name '*.proto')
protoc -I . --foo_out=. --foo_opt=one,two,three $(find . -name '*.proto')
protoc -I . --foo_out=one,two,three:. $(find . -name '*.proto')
```

buf.yaml
```diff
@@ -0,0 +1,4 @@
+version: v1beta1
+build:
```
bufdev, line 2: No need for this technically (this is the default)

ericvn (approved):
Discussed and approved in T&R WG meeting

bufdev:
Just a quick note as I don't think it is a part of this PR (and should probably be done in a follow-up): https://github.com/istio/api#backwards-incompatible-changes references protolock, which can be replaced with buf as well, here's a summary of the pros/cons, and the migration overall: https://docs.buf.build/migration-protolock

louiscryan (commented):
LGTM from me. 

@howardjohn 
Ping when ready to submit

howardjohn:
> Just a quick note as I don't think it is a part of this PR (and should probably be done in a follow-up): https://github.com/istio/api#backwards-incompatible-changes references protolock, which can be replaced with buf as well, here's a summary of the pros/cons, and the migration overall: https://docs.buf.build/migration-protolock

I was planning to look into this, and maybe linting, in a followup. Thanks!

> LGTM from me.

@louiscryan I was waiting on a dependency, should be good to go now. I removed the hold

louiscryan (approved):

PR #1842, opened 2021-01-22
HTTP local rate limit

kyessenov:
Expose per-route local rate limiting in HTTP.
Simplified version of Envoy APIs.


kyessenov:
/assign @bianpengyuan 
/assign @douglas-reid 
/assign @mandarjog 

We can use stage 0 for local rate limit (this PR) and stage 1 for global rate limit (follow up).

networking/v1beta1/virtual_service.proto
```diff
@@ -640,8 +640,89 @@ message HTTPRoute {
...
 
+  // Rate limits
+  RateLimit rate_limit = 21;
+
   // $hide_from_docs
-  // Next available field number: 21
+  // Next available field number: 22
+}
+
+// RateLimit configuration.
```
mandarjog, line 650: We need some good examples here.

kyessenov: Dome.

networking/v1alpha3/virtual_service.proto
```diff
@@ -643,10 +643,185 @@ message HTTPRoute {
...
+//           fill_interval: 1s
+//         descriptors:
+//         - entries:
+//           - key: method
+//             value: POST
+//           token_bucket:
+//             max_tokens: 100
+//             tokens_per_fill: 100
+//             fill_interval: 1s
+// ```
```
mandarjog, line 748: Thanks ! 
1. either description or example when max tokens and tokens per fill are different? We can point to the envoy documentation or another place as well, but would be good to have basics here.
2. Add an example for sidecar to sidecar communication where we want to limit easy source canonical service to a certain amount.

kyessenov: 1. It's hard to put in words a custom token bucket (it's a tradeoff between the timer overhead and "smoothness" of request bursts).
2. I don't think that's possible with the networking APIs since we need this on the server-side.

mandarjog: That is a bummer, also a deal breaker. Ability to ratelimit inter service traffic is one of the main customer asks.

PR #1844, opened 2021-01-25
Add CRD labels to WorkloadGroup

jasonwzm:
Fix of api for https://github.com/istio/istio/issues/30312


tmshort:
Are the manifest files automatically updated?

howardjohn (approved):
I think the labels here are probably not great (app=istio-pilot is not used anywhere else), but its consistent with other CRDs which is more important. LGTM

smawson (approved):

jasonwzm:
@tmshort After this is merged, we will manually update the istio/istio repo.

howardjohn:
It will auto update (if it works right)

PR #1848, opened 2021-01-26
Add root configuration API.

myidpt:
Resolves https://github.com/istio/istio/issues/30629.

This API is for configuring roots of trust for the workloads.
The Mesh Config supports adding the roots fetched from the SPIFFE endpoints or root PEMs defined directly in the config.
The Proxy Config only contains the PEM of the root certs. The Istiod is responsible for fetching all the roots and construct the PEM in the ProxyConfig. The Istiod sends the ProxyConfig to the workloads (istio-agents) via the XDS side channel.


mesh/v1alpha1/config.proto
```diff
@@ -210,6 +210,26 @@ message MeshConfig {
   // or `td3/ns/foo/sa/a-service-account` will be treated the same in the Istio mesh.
   repeated string trust_domain_aliases = 46;
 
+  message RootConfig {
+    enum Mode {
+      // The Citadel root that is available as the secret istio-ca-secret or cacerts in the control plane namespace.
+      // The config field is ignored in this mode.
+      CITADEL = 0;
```
howardjohn, line 217: I don't think we want to expose "citadel" to users anymore. Maybe this should be "self signed" or similar?

shankgan: Is there a possibility that we might want to combine these different approaches to allow more flexibility? For instance, we use a combination of SPIFFE_BUNDLE_ENDPOINT in addition to statically configuring PEM_CONTENT?

howardjohn: It is a repeated field, is that what you mean or something else?

shankgan: my bad. Missed that

myidpt: John: Something like "root-secret"? "self-sgined" isn't always true because when users plug in certs, it's in the cacerts secret.
Shankar: I used "repeated" so admin can define multiple RootConfigs.

shankgan: If this is intended to be a repeated field, then perhaps we need not have this explicit configuration option at all? Implicitly, we can always assume that if Istiod is signing workload certificates (whether self-signed or otherwise), it will always add them to the root of trust..Is there any scenario where we will not need to distribute the root of trust when Istiod acts as the CA?

costinm: I also don't like 'mode' - that's implementation detail. 
We can add a repeated cert-authority-url and maybe add a repeated cert-authority-data ( using the term used in kube .config file ). 

RootConfig is super confusing - it's enough security config took over the word 'policy', there are many uses for 'root config' behind certificate authority config.

costinm: Also add comments explaining how this interact with the current config:
- if cacerts Secret is present ( user bring their own cert ) - that root will be added to the roots.
- if istio-ca-secret is present - it will also be added automatically ( self signed certificate )
- if 'platform' integrations are enabled ( like MeshCA ) - their roots can be added automatically. 

The list defines additional roots.

mesh/v1alpha1/config.proto
```diff
@@ -210,6 +210,26 @@ message MeshConfig {
...
+      // https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE_Trust_Domain_and_Bundle.md#the-spiffe-trust-domain-and-bundle
+      // The config field specifies the SPIFFE Bundle Endpoint URL.
+      SPIFFE_BUNDLE_ENDPOINT = 1;
+      // The config specifies the PEM content of the root certificates.
+      PEM_CONTENT = 2;
+    }
+    // Mode specifies how the root certificate is loaded.
+    Mode mode = 1;
+    // Config specifies the details of the root certificate or its fetching method, depending on the mode.
+    string config = 2;
```
costinm, line 228: Also the term 'config' is too broad. Everything in istio is a config - why would the certificate authority root be called 'config' ?

mesh/v1alpha1/proxy.proto
```diff
@@ -493,6 +493,8 @@ message ProxyConfig {
   // Default value is 'false'.
   google.protobuf.BoolValue hold_application_until_proxy_starts = 33;
 
+  // The PEM content of the root certificates used in workload-to-workload communication.
+  repeated string root_cert_pem = 34;
```
costinm, line 497: I would use the same name in both proxy config and mesh config.

We use the term 'ca_certificates' in DestinationRule - I would use either this or whatever K8S Services API is using, no need to define 5 keys for the same thing ( root_cert_pem, config, etc)

mesh/v1alpha1/config.proto
costinm: Revising my comment - we use ca_certificates in DestinationRule, let's keep that ( and ca_certificates_url for the download case)

costinm: Not aware of any such case. This config is for extra roots - there is no point for users to have to copy the pub key of the self-signed cert into mesh config.

costinm: We will allow users to stop using the self-signed cert - for example by removing the secret, after they confirm all workloads are using the new root. And we'll probably need a way to not create a new one automatically - for example if mesh config includes an explicit root configuration, istiod could skip auto-creating the self-signed cert ( like it does if user has explicit cacert - brings his own certs ). Also Istiod should skip creating it if an external CA is configured.

smawson:
To get us in the habit, can you link to the RFC or design doc for this as part of the description? Thanks!

mesh/v1alpha1/config.proto
lizan: +1, and in general API style, prefer oneof instead of enum + opaque config like this to be more descriptive.

lizan (commented):
Discussed with @myidpt earlier today, can we make this easier to be extended for multiple trust bundles, which is coming in near future (in 2021 roadmap). see https://github.com/istio/istio/issues/30454

mesh/v1alpha1/config.proto
costinm: I prefer consistency with the rest of K8S and Istio APIs. 
Adding a mode with internal component names ( citadel ) doesn't make things more descriptive or readable, and the key is descriptive enough:
- mode: pem
  config: PEM data

vs 

root-cert-pem: PEM_data

mesh/v1alpha1/proxy.proto
myidpt: I'm fine with using ca_certificates. It's not very precise, but it's not a big issue.

mesh/v1alpha1/config.proto
myidpt: As we discussed, the Citadel root secrets will be implicitly added to the roots by the control plane. Also changed the field names to be clearer.

myidpt:
@lizan as we discussed separately, we'll keep the root config for local mesh separate from the config for other meshes, as the later can include other fields such as multiple trust domains and will need to have another design.

myidpt:
@smawson Yeah, I'll attach a short RFC for this.

mesh/v1alpha1/config.proto
```diff
@@ -210,6 +210,14 @@ message MeshConfig {
   // or `td3/ns/foo/sa/a-service-account` will be treated the same in the Istio mesh.
   repeated string trust_domain_aliases = 46;
 
+  // The PEM data of the root certificates for workload-to-workload communication.
+  repeated string ca_certificates = 58;
```
costinm, line 214: caCertificates in DestinationRule is a file path

I would use caCertificatesPEM

mesh/v1alpha1/proxy.proto
```diff
@@ -493,6 +493,8 @@ message ProxyConfig {
   // Default value is 'false'.
   google.protobuf.BoolValue hold_application_until_proxy_starts = 33;
 
+  // The PEM data of the root certificates for workload-to-workload communication.
```
costinm, line 496: Add to the comment: "this includes the certificates defined in MeshConfig and any other certificates that Istiod uses. The "cacert" certificates, self-signed, external CAs  certificates are added automatically by Istiod."

mesh/v1alpha1/config.proto
myidpt: Changed to ca_certificate_pems.

mesh/v1alpha1/proxy.proto
myidpt: Done.

costinm:
Looks good - but a native English speaker should confirm if the plurals sound right ( pems ?). I don't remember how K8S handles repeated fields and plurals - any UX around ?

mesh/v1alpha1/config.proto
```diff
@@ -210,6 +210,15 @@ message MeshConfig {
   // or `td3/ns/foo/sa/a-service-account` will be treated the same in the Istio mesh.
   repeated string trust_domain_aliases = 46;
 
+  // The PEM data of the root certificates for workload-to-workload communication.
+  repeated string ca_certificate_pems = 58;
```
howardjohn, line 214: bikeshed: It may be more clear the connection between these by changing it to be:
```protobuf
repeated CertificateReference ca_certificates;
message CertificateReference {
  oneof { data string; url string }
}
```

howardjohn:
Naming bikeshedding aside, LGTM

mesh/v1alpha1/config.proto
myidpt: I like this idea. It's more extensible to support other approaches to fetch the root.
For example, if the SPIFFE URL service is using a SPIFFE cert signed by a custom CA, we need to define a more complicated way to fetch the root. The config will include {<root cert>, <expected SPIFFE ID in the server cert>, <url>}. Using the "oneof" can better normalize the different configs.
@costinm WDYT?

costinm: ca_certificates:
  - url: ....
  - pem: ...

I can live with that - as long as it doesn't have a 'mode: citadel' or something like that.

mesh/v1alpha1/config.proto
```diff
@@ -210,6 +210,22 @@ message MeshConfig {
   // or `td3/ns/foo/sa/a-service-account` will be treated the same in the Istio mesh.
   repeated string trust_domain_aliases = 46;
 
+  message CertificateReference {
```
costinm, line 213: "Reference" may not be the right name - it's the data.

costinm (approved):

mesh/v1alpha1/config.proto
myidpt: Sounds good. I'll change that to be more precise.

mesh/v1alpha1/config.proto
```diff
@@ -210,6 +210,22 @@ message MeshConfig {
...
+      // https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE_Trust_Domain_and_Bundle.md#the-spiffe-trust-domain-and-bundle
+      // The endpoint should support authentication based on Web PKI:
+      // https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE_Trust_Domain_and_Bundle.md#521-web-pki
+      // The certificate is retrieved from the endpoint.
+      string spiffe_bundle_url = 2;
+    }
+  }
+
+  // The root certificates for workload-to-workload communication.
+  repeated CertificateData ca_certificates = 58;
```
howardjohn, line 227: Should we comment about the implicit root (self signed/ ca-certs)? Otherwise it sounds like this is required for all users.

mesh/v1alpha1/proxy.proto
howardjohn: Can you add on the config.proto the same? so users do not think they need to duplicate it

howardjohn (approved):

mesh/v1alpha1/config.proto
smawson: +1 to describing the built-in ca certs.

mesh/v1alpha1/proxy.proto
```diff
@@ -493,6 +493,11 @@ message ProxyConfig {
...
+  // This includes the certificates defined in MeshConfig and any other certificates that Istiod uses as CA.
+  // The plugin certificates (the 'cacerts' secret), self-signed certificates (the 'istio-ca-secret' secret)
```
smawson, line 498:
```suggestion
  // The plugin certificates (the 'cacerts' secret) and self-signed certificates (the 'istio-ca-secret' secret)
```

mesh/v1alpha1/config.proto
myidpt: Yeah, we can do that.

smawson (commented):
Some minor comments, LGTM and will approve once those are fixed.

mesh/v1alpha1/proxy.proto
myidpt: Got what you mean. Sure.

mesh/v1alpha1/proxy.proto
myidpt: I'll use "or" here because when using plugin certs, self-signed certs won't exist/used.

smawson (approved):

PR #1849, opened 2021-01-26
networking extensions

kyessenov:
Generalization of https://github.com/istio/api/pull/1842 to support forward-thinking, proprietary extensions in the HTTP rules.

This roughly matches [HTTPRouteFilter](https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteFilter). 

/assign @howardjohn 
/assign @mandarjog 
/assign @douglas-reid 


networking/v1alpha3/virtual_service.proto
```diff
@@ -643,10 +644,26 @@ message HTTPRoute {
...
+
   // $hide_from_docs
-  // Next available field number: 21
+  // Next available field number: 22
 }
 
+// HTTP route filter for custom extensions.
+message HTTPRouteFilter {
+  // Extension category determining the type of a filter extension.
+  string category = 1;
```
mandarjog, line 657: ENUM ? or a list in the doc?

networking/v1alpha3/virtual_service.proto
```diff
@@ -643,10 +644,26 @@ message HTTPRoute {
...
+
+  oneof ref_type {
+    // External reference to the extension configuration by name.
+    string name = 2;
+
+    // Explicit extension configuration.
+    google.protobuf.Struct value = 3;
```
mandarjog, line 664: UpdaStruct, so Istio can verify it against **some** proto.

mandarjog (commented):
I think this is fine, but it only deal with client side extensions, which are only really used at gateways. We need something equivalent on the server side.

networking/v1alpha3/virtual_service.proto
kyessenov: I want to leave it open-ended just like in k8s.

networking/v1alpha3/virtual_service.proto
kyessenov: Istio API are only using proto for a schema, and I imagine the extension may chose to use k8s schema instead. UDPA struct is pure proto.

kyessenov:
@mandarjog Yes, I think for server-side extensions we can generalize Wasm extensions API to cover native extensions. This would be a brand new API. cc @dgn

networking/v1alpha3/virtual_service.proto
dgn: what kind of categories would these be? if we're referencing a generic Extension resource here, that resource will define its type itself, right?

kyessenov: Yes, name is sufficient. This is trying to align GVK from k8s Gateway with extensions, so it could help to have an explicit "type" of an extension here.

kyessenov:
@dgn I think the back-reference from the Extension resource to the VirtualService route is also a viable approach. I do want to make sure that k8s Gateway can be translated to a set of Extension and VirtualService resources. I think there are a few requirements for this to be possible:
1. k8s allows arbitrary order of filtesr, which is inconsistent with "phases" approach.
2. We need an individual route match in the match context in Extension.
3. Wasm filters can be applied per route via a generic typed_per_filter_config mechanism or something like that (cc @PiotrSikora )

douglas-reid:
@kyessenov do we have a doc or more info on usage / alignment with other proposed APIs ?

kyessenov:
Daniel's proposal deals with server-side extensions. That is more restricted and simpler, so it should take priority. Ingress/client side can follow.

PR #1853, opened 2021-01-27
Updating initial stability of subzone and network labels

nmittler:
subzone: has been around for a while. It's tested and I believe some folks are using it in production. Probably at least beta.

network: fundamental label used by multicluster in particular for configuring multi-network scenarios. Since multicluster as a feature is beta, it makes sense that the stability here matches.


smawson (approved):

nmittler:
@howardjohn mind approving as well?

howardjohn (approved):

PR #1855, opened 2021-01-27
Fix MeshNetworks example documentation

cainelli:
The example for MeshNetworks in the [documentation page](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshNetworks) doesn't seem to be correct. The generated pb.go file looks like this:

https://github.com/istio/api/blob/11df83b4d252c9eee246fc26aaa5d4fc2e0acd78/mesh/v1alpha1/network.pb.go#L327-L331

### DIFF
--- a/mesh/v1alpha1/network.proto
+++ b/mesh/v1alpha1/network.proto
@@ -102,7 +102,7 @@ message Network {
 // ```yaml
 // networks:
 //   network1:
-//   - endpoints:
+//     endpoints:
 //     - fromRegistry: registry1 #must match kubeconfig name in Kubernetes secret
 //     - fromCidr: 192.168.100.0/22 #a VM network for example
 //     gateways:
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
@@ -1715,7 +1715,7 @@ inside a mesh and how to route to endpoints in each network. For example</p>
 
 <pre><code class="language-yaml">networks:
   network1:
-  - endpoints:
+    endpoints:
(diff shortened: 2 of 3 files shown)


mesh/v1alpha1/network.proto
```diff
@@ -102,7 +102,7 @@ message Network {
 // ```yaml
 // networks:
 //   network1:
-//   - endpoints:
+//     endpoints:
```
howardjohn, line 105:
```
type Network struct {
	// The list of endpoints in the network (obtained through the
	// constituent service registries or from CIDR ranges). All endpoints in
	// the network are directly accessible to one another.
	Endpoints []*Network_NetworkEndpoints `protobuf:"bytes,2,rep,name=endpoints,proto3" json:"endpoints,omitempty"`
```

seems like endpoints is a repeated fiield?

cainelli: Yes, this seems to be allowed (at least in the configuration parsing) in the configuration:

```yaml
networks:
   network1:
     endpoints:
     - fromRegistry: registry1 #must match kubeconfig name in Kubernetes secret
     - fromCidr: 192.168.100.0/22 #a VM network for example
```

but not this:

```yaml
networks:
   network1:
   - endpoints:
     - fromRegistry: registry1 #must match kubeconfig name in Kubernetes secret
     - fromCidr: 192.168.100.0/22 #a VM network for example
```

howardjohn: ah right.. yaml tricks me every time :slightly_smiling_face:

smawson (approved):
YAML is hard!

hzxuzhonghu (approved):

PR #1861, opened 2021-01-29
Workaround golang bug causing make gen to fail

howardjohn:
Workaround https://github.com/golang/go/issues/43994

Run tidy after download


ericvn (approved):

jacob-delgado (approved):
Trying to get 1.16 into Istio 1.9. LGTM

PR #1867, opened 2021-02-01
rename document_url to documentUrl

adiprerepa:
Didn't expect it to require an api change.

for [istio/istio#30514](https://github.com/istio/istio/issues/30514)

cc @howardjohn


howardjohn (changes requested):
All proto fields are snake_case, the generated json/yaml will still be camel case though

adiprerepa:
ah, right. forgot.

PR #1870, opened 2021-02-02
[WIP] Introducing alpha API for ServiceExportPolicy

nmittler:
This attempts to provide finer-grained control over how a services are exported across the mesh.

Allows setting both mesh-wide and namespace defaults as well as overriding defaults with policies for individual services.


nmittler:
FYI @jeremyOT

smawson:
My initial gut reaction is that this feels too complicated. I was imagining something a bit simpler, just controlling generation of ServiceExport resources. Basically just a service selector to indicate which services in a namespace should be exported, and control over whether they are exported or not. Then if that is in the system namespace it would set default for the cluster.

```
message ServiceExportPolicy {
  ServiceSelector selector = 1;
  bool auto_export = 2;
}
```

The simple match-current-istio behavior would be no selector, auto_export true, placed in the mesh system namespace.

networking/v1alpha3/service_export_policy.proto
```diff
@@ -0,0 +1,170 @@
...
+
+// ServiceExportPolicy configures the exporting of services across the mesh.
+//
+// We use the term "exporting" to refer to the process where a service's endpoints are
+// made accessible to one or more workloads. If no endpoint for a service is exported
+// to a workload, the service will be effectively unreachable from the workload.
+//
+// ServiceExportPolicy supports a variety of use cases that require partitioning the mesh
+// and providing customized views of service endpoints to each workload. For example,
+// it may be desirable that workloads in a given cluster use only the service endpoints
```
frankbu, line 32: Isn't this already handled quite simply in Istio? When/why would/should one use ServiceExportPolicy vs. locality LB, to restrict calls to the same cluster?

I think I agree with @smawson's comment, this seems like it might be overkill. Do we really need all the flexibility?

Also, I assume we would want to use ServiceExportPolicy to control the generating of MCS ServiceExports. If that's correct, it seems like the "to" part can't be honoured, since MCS doesn't support exporting to a subset of the workloads, or even subset of clusters.

nmittler: > Isn't this already handled quite simply in Istio? When/why would/should one use ServiceExportPolicy vs. locality LB, to restrict calls to the same cluster?

Locality LB can only restrict calls to region, zone, subzone. It doesn't address cluster at all. The only API for cluster-local today is in mesh config, which isn't really accessible for service owners.

> Also, I assume we would want to use ServiceExportPolicy to control the generating of MCS ServiceExports. If that's correct, it seems like the "to" part can't be honoured, since MCS doesn't support exporting to a subset of the workloads, or even subset of clusters.

I was more focused on thinking about what we think the Istio API should be, rather than being strictly restricted to the current set of capabilities afforded by the MCS API. If we decide we'd like more control, we could help drive additional requirements for MCS.

nmittler:
> My initial gut reaction is that this feels too complicated.

@smawson I don't entirely disagree. There were a few things that I was trying to fit into this, however:

1. The ability to invert the Istio default, making service export explicit in order to match the behavior of MCS.
2. Some way to integrate cluster-local into all of this, since it's effectively the same thing ... killing 2 birds with one stone.
3. Unblock the MCS integration work

The only way to do 2 AFIAK is with workload selectors on both the service endpoints and client workloads.

The way I chose to do 1 was to allow `ServiceExportPolicy` to be defined in the `istio-system` namespace to change the global configuration and it seemed natural to allow a per-namespace and per-service override. I suppose we could put the global default in mesh config, but it seems a bit awkward to have 2 different APIs. A middle-ground might be to just embed the CR in mesh config, but I'm not sure if that's an improvement or not.

nrjpoddar:
@nmittler is there a RFP for this API change and how does this interact with `networking.istio.io/exportTo` annotation that can be added on any Kubernetes service?

smawson:
We should discuss (2), I don't see why it requires the workload selectors on to and from. We don't have that today for cluster-local right? With MCS model cluster-local is the default and services need to be explicitly supported, so I think we're OK with just a boolean on the mesh level, namespace level and service level, with inheritance.

howardjohn:
One thing we are looking into for the service-apis implementation is how to handle service/config import/export. Its possible we would want to define a separate policy in Istio or the k8s api, which seems plausible to have significant overlap with this. I could see us defining a policy that deals with exporting to other namespaces and exporting to other clusters.

This discussion is still pretty early though, but we may want to sync up on this when we understand the problem more (should be soon)

nmittler:
@smawson 

> We should discuss (2), I don't see why it requires the workload selectors on to and from. We don't have that today for cluster-local right? With MCS model cluster-local is the default and services need to be explicitly supported, so I think we're OK with just a boolean on the mesh level, namespace level and service level, with inheritance.

I should have also added this as a goal:

4. Support the existing `exportTo` capabilities for exporting to selected namespaces.

A boolean is enough to capture cluster-local vs global, but it's not sufficient to cover exporting to selected namespaces or a subset of services. Selecting a destination host (with wildcards) gets you the ability to select the services and namespaces to export to. But then you start to run up against the question "what does export really mean"? If we limit exporting to cluster-local vs mesh-wide, allowing namespace/service selection starts to break this down quickly.  I believe the general solution to all of this is exporting selected service endpoints to selected workloads. This allows us fine-grained control of service exporting across the mesh, and frees us from having to think about cluster entirely (cluster is just a label we can select on). It also has implications WRT potentially simplifying some of the service exposure-related issues with mesh federation.

nmittler:
@howardjohn 

> One thing we are looking into for the service-apis implementation is how to handle service/config import/export. Its possible we would want to define a separate policy in Istio or the k8s api, which seems plausible to have significant overlap with this. I could see us defining a policy that deals with exporting to other namespaces and exporting to other clusters.

Yeah sounds very similar. Let's sync when you're at a good point.

networking/v1alpha3/service_export_policy.proto
```diff
@@ -0,0 +1,170 @@
...
+// residing within the same cluster.
+//
+// By default (if no ServiceExportPolicy is specified), all services are exported to all
```
linsun, line 35: would prefer to remove by default... it is possible vendor may choose to have a service export policy in place as the default.

networking/v1alpha3/service_export_policy.proto
```diff
@@ -0,0 +1,170 @@
...
+// workloads in the mesh. This means that any workload can call any service endpoint.
+//
+// The mesh-wide default can be overridden by applying a ServiceExportPolicy to the
+// Istio system namespace, using the service wildcard. For example, the following will
```
linsun, line 39: or a root namespace

networking/v1alpha3/service_export_policy.proto
```diff
@@ -0,0 +1,170 @@
...
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: ServiceExportPolicy
+// metadata:
+//   name: auto-export-off
+//   namespace: istio-system # Using the Istio system namespace applies mesh-wide
+// spec:
+//   from:
+//     service: * # Using wildcard applies to all services.
+//   # No "to" is specified, so services will not export by default.
```
linsun, line 54: I think this example is somewhat confusing.  what if I remove this?  Does it also mean everything is not exported?

```
//   from:
//     service: * # Using wildcard applies to all services.
```

Do you think it would be useful to have a config in the API to indicate it is globally auto exported vs not?  like our mtls strict or none api...

networking/v1alpha3/service_export_policy.proto
```diff
@@ -0,0 +1,170 @@
...
+// {{<tabset category-name="example">}}
+// {{<tab name="v1alpha3" category-value="v1alpha3">}}
+// ```yaml
+// apiVersion: networking.istio.io/v1alpha3
+// kind: ServiceExportPolicy
+// metadata:
+//   name: auto-export-off
+//   namespace: example-ns
+// spec:
+//   from:
```
linsun, line 71: I think the from and to in the export service policy is very confusing.  I had envisioned this to be service export policy at a cluster level.  Within the cluster, we already have sidecar API where you could specify ingress and egress.  How is this different?

linsun:
Hey @nmittler - I am going to echo some of the points raised by others earlier.  I would see a former proposal on this to introduce the problems we are trying to solve on top of what we already have (exportTo & sidecar API) and also understand how this relates to the k8s MCS API.

networking/v1alpha3/service_export_policy.proto
```diff
@@ -0,0 +1,170 @@
...
+//     service: MyService
+//     selector:
+//       "topology.istio.io/cluster": cluster1
+//   to:
+//     hosts:
+//     - *
+//     selector:
+//       "topology.istio.io/cluster": cluster1
```
frankbu, line 101: This seems like a strange example. Isn't this just configuring what would be the default if there was no export at all?

smawson:
> @smawson
> 
> > We should discuss (2), I don't see why it requires the workload selectors on to and from. We don't have that today for cluster-local right? With MCS model cluster-local is the default and services need to be explicitly supported, so I think we're OK with just a boolean on the mesh level, namespace level and service level, with inheritance.
> 
> I should have also added this as a goal:
> 
> 1. Support the existing `exportTo` capabilities for exporting to selected namespaces.
> 
> A boolean is enough to capture cluster-local vs global, but it's not sufficient to cover exporting to selected namespaces or a subset of services. Selecting a destination host (with wildcards) gets you the ability to select the services and namespaces to export to. But then you start to run up against the question "what does export really mean"? If we limit exporting to cluster-local vs mesh-wide, allowing namespace/service selection starts to break this down quickly. I believe the general solution to all of this is exporting selected service endpoints to selected workloads. This allows us fine-grained control of service exporting across the mesh, and frees us from having to think about cluster entirely (cluster is just a label we can select on). It also has implications WRT potentially simplifying some of the service exposure-related issues with mesh federation.

I don't know if I agree that supporting existing exportTo capabilities should be goal. To me this should be focused solely on controlling ServiceExport resources and nothing more. This controls whether a set of endpoints for a service are exported to the mesh-wide service or not.

Otherwise if we try to bring in the exportTo stuff it will get really complicated really quickly. What if different clusters say different things about exportTo? Do those only count in that cluster? Or do they affect all consumers of that service?

You're bringing in new requirements that make solving this correctly much harder, because the number of cases explodes. If we limit ourselves instead to just the question of when to create a ServiceExport and when not to, the design is much simpler and straightforward. It also gives us a nice composable piece we can build things like exportTo on top of.

PR #1871, opened 2021-02-03
Change `http2MaxRequests` setting to `maxRequests`

kailun-qin:
The combination of H1 and H2 conn_pool implementations makes `max_requests` to
work for both H1 and H2 in Envoy proxy
(https://github.com/envoyproxy/envoy/issues/9215).

In accordance with this change, this patch updated the field `http2MaxRequests`
of `ConnectionPoolSettings.HTTPSettings` in DestinationRule API to `maxRequests`
to resolve the outdated H1 only implications/misunderstandings for the field.

Addresses https://github.com/istio/istio/issues/27473

Signed-off-by: Kailun Qin <kailun.qin@intel.com>


networking/v1alpha3/destination_rule.proto
```diff
@@ -607,7 +607,7 @@ message ConnectionPoolSettings {
     int32 http1_max_pending_requests = 1;
 
     // Maximum number of requests to a backend. Default 2^32-1.
-    int32 http2_max_requests = 2;
+    int32 max_requests = 2;
```
ramaraochavali, line 610: We can not change these names now as it will break existing deployments. We have to follow the deprecation process

howardjohn: strong +1 here

howardjohn (changes requested):
Adding a new field is probably a good idea, and hiding the old one - but we must support both indefinitely

networking/v1alpha3/destination_rule.proto
costinm: I assume this and http1_max_pending_requests will also need to be deprecated, new field added - and some tooling in istioctl before we even consider removing the old field and behavior. 

An alternative is to just change the documentation - "Istio may upgrade HTTP/1.1 requests to HTTP/2 internally, this setting will cover both HTTP/1.1 and HTTP/2 requests". That's what we plan to do with BTS AFAIK, everything including TCP will be over H2. 

If we add a new field - we should also consider BTS, and move it up, since TCP will also be treated as a long-lived H2.

ramaraochavali: I was thinking about update the doc as @costinm mentioned. IMO, that is better - it is not worth the churn to add a new field and retain the old field forever.

kailun-qin: [http1_max_pending_requests](https://github.com/envoyproxy/envoy/blame/main/docs/root/intro/arch_overview/upstream/circuit_breaking.rst#L26) might be slightly different since it involves extra settings to cover behaviors of H2.

Anyway, marking BTS related issues here: https://github.com/istio/istio/issues/24540, https://github.com/istio/istio/issues/25713.

PR #1873, opened 2021-02-04
[#30611] Fix perTryTimeout documentation to reflect the code

etifontaine:
This PR fix the issue #30611 on istio/istio
https://github.com/istio/istio/issues/30611

### DIFF
--- a/networking/v1beta1/virtual_service.proto
+++ b/networking/v1beta1/virtual_service.proto
@@ -1539,7 +1539,7 @@ message HTTPRetry {
   // the specified request `timeout` and `per_try_timeout` values.
   int32 attempts = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // Timeout per retry attempt for a given request. format: 1h/1m/1s/1ms. MUST BE >=1ms.
+  // Timeout per attempt for a given request, including the initial call and any retries. Format: 1h/1m/1s/1ms. MUST BE >=1ms.
   // Default is same value as request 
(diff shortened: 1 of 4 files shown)


howardjohn (approved):

networking/v1beta1/virtual_service.proto
```diff
@@ -1539,7 +1539,7 @@ message HTTPRetry {
   // the specified request `timeout` and `per_try_timeout` values.
   int32 attempts = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // Timeout per retry attempt for a given request. format: 1h/1m/1s/1ms. MUST BE >=1ms.
+  // Timeout per try attempt for a given request, including the initial call. format: 1h/1m/1s/1ms. MUST BE >=1ms.
```
smawson, line 1542:
```suggestion
  // Timeout per attempt for a given request, including the initial call and any retries. format: 1h/1m/1s/1ms. MUST BE >=1ms.
```

smawson (approved):

hzxuzhonghu (approved):
/ok-to-test

etifontaine:
@smawson How does this will be roll out on the website now ?

smawson:
IIRC because this is in master it won't go out until 1.10 release, since 1.9 was already cut. You probably want to cherry pick this into 1.9 branch so we can get it into those docs. But @howardjohn would know better than I.

etifontaine:
If it can be also put in place for older versions it would be great. At least the supported versions.

PR #1879, opened 2021-02-10
Fix for some text that went missing

ericvn:
Noted in slack that some text was missing on istio.io. The file is copied from here.


howardjohn (approved):

smawson (approved):
Any idea why the generated docs disappeared?

howardjohn:
The issue is the newline between the `package` and the comment.

smawson:
Whitespace is semantically important, wonderful :)

PR #1863, opened 2021-01-29
Use buf for protobuf linting

bufdev:
Following up on #1833, this uses `buf` for linting instead of `prototool`.

A couple changes are made:

- `prototool` lint configuration is deleted, as it is not used anymore, and lint configuration remaining in `prototool.yaml` could lead to confusion as to where the SOT is for Protobuf lint configuration. It looked like the ignores were out of date as well.
- `common-protos` is added to `excludes` in `buf.yaml`. This is more a bug fix - since `common-protos` is symlinked into the base of the repository, we want `buf` to ignore `common-protos` altogether for `buf` to work out of the box. This makes `buf` file discovery ignore `common-protos`, and all `buf` commands work now assuming the symlinks are set up (which `make` does). As a fun example:

```
$ buf build --exclude-imports --exclude-source-info -o -#format=json | jq '.file[] | .package' | sort | uniq
"istio.analysis.v1alpha1"
"istio.authentication.v1alpha1"
"istio.envoy.config.filter.http.alpn.v2alpha1"
"istio.envoy.config.filter.http.authn.v2alpha1"
"istio.envoy.config.filter.http.jwt_auth.v2alpha1"
"istio.envoy.config.filter.network.tcp_cluster_rewrite.v2alpha1"
"istio.mcp.v1alpha1"
"istio.mesh.v1alpha1"
"istio.meta.v1alpha1"
"istio.networking.v1alpha3"
"istio.networking.v1beta1"
"istio.operator.v1alpha1"
"istio.security.v1beta1"
"istio.type.v1beta1"
"istio.v1.auth"
```

- Lint configuration is added to `buf.yaml` that seems to align with what is wanted in this repository. See https://docs.buf.build/lint-configuration for configuration options, and https://docs.buf.build/lint-rules for available rules. With this configuration, the configured rules are:

```
$ buf config ls-lint-rules
ID                                CATEGORIES                                  PURPOSE
DIRECTORY_SAME_PACKAGE            MINIMAL, BASIC, DEFAULT, FILE_LAYOUT        Checks that all files in a given directory are in the same package.
PACKAGE_SAME_DIRECTORY            MINIMAL, BASIC, DEFAULT, FILE_LAYOUT        Checks that all files with a given package are in the same directory.
PACKAGE_SAME_CSHARP_NAMESPACE     MINIMAL, BASIC, DEFAULT, PACKAGE_AFFINITY   Checks that all files with a given package have the same value for the csharp_namespace option.
[...]

PR #1886, opened 2021-02-17
Update to use local-lint-protos

ericvn:
#1863 updated the `lint-protos` take in the common directory. Of course a common-files change will overwrite the change in that PR and that cause a lint failure. See #1872.

This PR makes the lint-protos change local to istio/api so a later common-files update regressing the file here will have no effect.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -105,7 +105,11 @@ release-lock-status:
 # Misc
 #####################
 
-lint: lint-all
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

ericvn:
Looking for @istio/technical-oversight-committee approval to get this in so we can update the common files.

smawson (approved):

PR #1890, opened 2021-02-17
Add root configuration API

myidpt:
Manual cherrypick from https://github.com/istio/api/pull/1848.

Today, Istio workloads doesn't support multiple roots. This is a big limitation, meaning CA certificate rotation, or the migration from one CA to another isn't feasible for production systems (due to downtime). We consider enabling multiple roots of trust as a high-priority fix and it is needed in 1.9 as well.


shankgan (approved):

costinm (approved):

shamsher31 (approved):

howardjohn:
A bit confused here. The implementation is not merged and 2k+ complex LOC. Are we planning to backport all of this in a patch release?

myidpt:
I think the answer is we need to talk to our customer. If our customers have good justification that the implementation needs to be merged, we probably should do it. If the API is sufficient (and they patch the implementation on their side) then I think we are good.

PR #1891, opened 2021-02-18
add support for zone/subzone level failover settings

masquee:
Current failover settings only support failover from one region to another region, but sometimes we prefer to failover from one zone to another zone, even from one subzone to another subzone.

This pull request adds support for zone/subzone failover settings. For example, failover settings can set as
`
{ 
  From: "region1/zone1/subzone1", 
  To: "region2/zone2/subzone2" 
}
`

For pepole who only concerns about region-level failover, they can still use failover setting as follow, which compatible with before.
`
{ 
  From: "region1", 
  To: "region2" 
}
`

**This pull request just modify failover releated comments, no proto struct changes.**

Related implementation changes resides in this pull request: https://github.com/istio/istio/pull/30781

### DIFF
--- a/networking/v1beta1/destination_rule.proto
+++ b/networking/v1beta1/destination_rule.proto
@@ -984,18 +984,18 @@ message ClientTLSSettings {
 // operational requirements an operator can set a 'failover' policy instead of
 // a 'distribute' policy.
 //
-// The following example sets up a locality failover policy for regions.
-// Assume a service resides in zones within us-east, us-west & eu-west
-// this example specifies that when endpoints within us-east become unhealthy
-// traffic should failover to endpoints in any zone or sub-zone within eu-west
-// and similarly us-west should failover to us-east.
+// The following example sets up a locality failover policy.
+// Assume a service resides in "us-east/zone1/subzone1" and "eu-west/zone2/subzone2",
+// this example specifies that when endpoints within "us-east/zone1/subzone1" become unhealthy
+// traffic should failover to endpoints in "eu-west/zone2/subzone2"
+// and similarly "eu-west/zone2/subzone2" should failover to "us-east/zone1/subzone1".
 //
 // ```yaml
 //  failover:
-//    - from: us-east
-//      to: eu-west
-//    - from: us-west
-//      to: us-east
+//    - from: us-east/zone1/subzone1
+//      to: eu-west/zone2/subzone2
+//    - from: us-west/zone2/subzone2
+//      to: us-east/zone1/subzone1
 // ```
 // Locality load balancing settings.
 message LocalityLoadBalancerSetting{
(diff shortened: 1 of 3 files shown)


networking/v1beta1/destination_rule.proto
```diff
@@ -984,18 +984,18 @@ message ClientTLSSettings {
...
+// traffic should failover to endpoints in "eu-west/zone2/subzone2"
+// and similarly "eu-west/zone2/subzone2" should failover to "us-east/zone1/subzone1".
 //
 // ```yaml
 //  failover:
-//    - from: us-east
-//      to: eu-west
-//    - from: us-west
-//      to: us-east
+//    - from: us-east/zone1/subzone1
```
howardjohn, line 995: With this configuration, if I have pods in us-east/zone1/subzone1, us-east/zone1/subzone2, and eu-west/zone2/subzone2.

Client pod is in us-east/zone1/subzone1.

If us-east/zone1/subzone1 becomes unhealthy, where do requests go to? us-east/zone1/subzone2 or eu-west/zone2/subzone2?

`eu-west/zone2/subzone2` seems to be consistent with the API described here. However, it seems inconsistent with the expectation that we send traffic to the closest region/zone/subzone first.

In the current API, `failover` basically just adds a 4th level to the hierarchy. My understanding is this intends to add N levels, which seems reasonable. I could see a user wanting to spillover in explicit order like `a/b/c1 -> a/b/c2 -> a/b/c3 -> a/b/<any> -> a/<any>/<any>`.

hzxuzhonghu: Yeah, same doubt what's the use case for forwarding the traffic to remote zone instead of same zone first?

linsun: Same concerns.  Please give examples where there is also us-east/zone1/subzone1 & us-east/zone1/subzone2 &  us-east/zone2/subzone1 & us-east/zone2/subzone1

masquee: @howardjohn Yes it's more flexible and allows completely re-ordering the hierarchy, just like "distribute" settings.

@hzxuzhonghu @linsun I agree that it's not a good idea to failover to remote zone instead of same zone first. But in some cases,  we need failover to a specified subzone first, which may be a logic part of a zone.

I think locality failover should support following settings:
-  failover between different subzones which all in same zone
-  failover  between different zones which all in same region
-  failover between regions


Yes the failover settings in this pull request is unreasonable. How about this:

`
   failover:
       - from: region1/*/*
       - to: region2/*/*
`

`
   failover:
       - from: region1/zone1/*
       - to: region1/zone2/*
`

`
   failover:
       - from: region1/zone1/sunzone1
       - to: region1/zone1/sunzone2
`

hzxuzhonghu:
@masquee See a more general way, supporting failover by topology keys

PR #1902, opened 2021-03-04
Update VirtualService docs with defaults for retry behaviour, remove required attribute on attempts

Niksko:
This updates the documentation for the VirtualService resource to
include the default values for the `onRetry`, `attempts` and `retryRemoteLocalities` fields.
It also updates the main `HTTPRoute` docs to include a note that retries are enabled by default, and that the policy defaults to the default values for each field.


hzxuzhonghu (commented):
lg, can you add default attempts doc. It is 2 by default.

Niksko:
> lg, can you add default attempts doc. It is 2 by default.

I thought attempts was a required field?

Niksko:
Updated to document `attempts` and `retryRemoteLocalities`, as well as adding a note that indicates that retries are enabled by default

Niksko:
I've only updated the proto for `v1alpha3`. Should I be updating the `v1beta1` proto as well/instead?

ericvn:
Added the cherrypick-1.9 label so this would be cherrypicked back to the release-1.9 branch and changes picked up on istio.io (changes pulled at each patch release so not effective there on merge (preliminary.istio.io should pick up changes in `master` after a daily job is run))

Niksko:
Can somebody please add the `release-notes-none` label (I don't think this needs release notes)

nrjpoddar (approved):

networking/v1alpha3/virtual_service.proto
```diff
@@ -1542,6 +1544,7 @@ message HTTPRetry {
   // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute) 
   // or `per_try_timeout` is configured, the actual number of retries attempted also depends on
   // the specified request `timeout` and `per_try_timeout` values.
+  // Defaults to `2`.
```
nrjpoddar, line 1547: is this field required @howardjohn?

hzxuzhonghu: Yes, if the HTTPRetry not specified the default value is 2. So we may add more comments

nrjpoddar: If it has a default, then this field is not required. Is that what you're saying?

hzxuzhonghu: I mean when user specifies the upper level httpRetry, then this field must be set too. If httpRetry is empty, then we apply a default retry policy with attempts = 2

hzxuzhonghu: I am thinking maybe we can change our validation to let attempts unset within HTTPRetry, in this case we apply the default `2`.

ramaraochavali: I agree. It is unnecessary to set again if default attempts are fine.

Niksko: Happy to make this change

nrjpoddar: Makes sense to me, @howardjohn wdyt?

linsun:
Can you make similar change to virtual service in v1beta1 as well?  We need to keep both in sync.

Niksko:
> Can you make similar change to virtual service in v1beta1 as well? We need to keep both in sync.

@linsun I've updated the docs for v1beta1. Do I also need to remove the 'required' property of the `attempts` field in v1alpha3?

PR #1906, opened 2021-03-09
Add releaselocks for 1.7-1.9

jacob-delgado:
Run make proto-commit
run make presubmit and fix all issues for each release


jacob-delgado:
It's not clear to me that this is working. The `relaselocks` haven't been updated since 1.6 and `make proto-commit` is not a part of the check and as a result `proto.lock` is out of sync. We should improve that and have it be part of `presubmit`.

How far back should we support for releaselocks, if at all?

jacob-delgado:
After doing some digging it looks like buf provides a similar mechanism. See https://docs.buf.build/migration-protolock/

PR #1740, opened 2020-11-10
Add initial Telemetry API with Tracing support for sampling and custom tags

douglas-reid:
This PR adds support for new `Telemetry` resource. In particular, it adds `Tracing` rules to enable workload-specific customization of tracing behavior. This is based on design discussions from [Proposal: Telemetry API](https://docs.google.com/document/d/1pTFVzGii7sB4uALMsrUN4ALFc0J7vI2UyBnJpP_bsgU/edit#) document (and the RFCs that preceded it).

The purpose of the new `Telemetry` resource is to enable a number of use cases around telemetry generation customization through a stable, supportable API that is much more operator-friendly than the current `EnvoyFilter`-based approaches in use today. Fuller justification of the PR (and the issues that is attempting to address) can be found in the proposal.

The `Telemetry` resource will hold provider-agnostic telemetry override configuration for workloads. It is meant to be used by Application Operators to customize telemetry generation for their workloads. In this PR, the `Telemetry` resource only configures tracing behavior to limit the overall scope and discussion while still providing insight into how the two additions will work together. Logging and metrics support will be quick-follows upon approval of this PR.

As the new resources mature, and full support is added to the control plane, subsequent PRs will be required to un-hide these changes in the docs and to label the various bits of `MeshConfig` and `ProxyConfig` that are being replaced as deprecated.

PR #1908, opened 2021-03-10
Add a flag to disable tracing at proxy level

nacx:
Adds a flag to completely disable tracing configuration at the proxy configuration.

Sidecar proxies do not mount the `istio` configmap and do not see the `enableTracing` flag. The only way to modify the tracing configuration is by editing the `tracing` object of the proxy config directly, but that does not work for disabling it. When overriding it in the `proxy.istio.io/config` annotation, setting it to `null` does not work, as it is interpreted as an omitted value and the default tracing config (Zipkin) is not modified.

This PR adds a flag that can be used in the config annotation to fully disable tracing like:
```
proxy.istio.io/config: "enableTracing: false"
```

Without this, even when tracing is globally disabled, the sidecar proxies contain tracing configuration.
More details in the implementation pull request: https://github.com/istio/istio/pull/31379


howardjohn:
@douglas-reid 

I appreciate the goal, but have a few comments:
* Does https://github.com/istio/istio/issues/29898#issuecomment-756290975 work to disable it?
* In 1.10(?) we are planning a first class tracing api that may make this obsolete: https://github.com/istio/api/pull/1740

douglas-reid:
@nacx i'm currently working on changes to the tracing API (and behaviors). i'd prefer we work on pushing that across the line to adding more configuration into `ProxyConfig`. Have you had a chance to look at that proposal?

nacx:
Totally. I'll track that., thx! :) I jumped into looking at the code before looking at the current open issues/PRs.

PR #1909, opened 2021-03-11
add Apache SkyWalking tracer support

Shikugawa:
Add API level Apache SkyWalking tracer support. 
It already implements on Envoy after Envoy 1.17, and we can introduce next istio release.

cc @wu-sheng


nrjpoddar:
Thanks @Shikugawa for this enhancement but I’m wondering if we should wait for the new API that @douglas-reid is working on and support SkyWalking there instead?

Shikugawa:
> Thanks @Shikugawa for this enhancement but I’m wondering if we should wait for the new API that @douglas-reid is working on and support SkyWalking there instead?

@nrjpoddar I didn't recent movement related to istio tracing. What objective is the new API and behavior change? Does it cover injecting SkyWalking config into Istio proxy?

douglas-reid:
@nrjpoddar we _really_ need to decide on where "providers" fit. this doesn't seem to conflict with the current proposal for the first pass Telemetry API, as it is purely about "provider" config.

wu-sheng:
@douglas-reid @nrjpoddar Do you have any update about this?

nrjpoddar:
We are meeting today to decide on the new API and how to phase out/continue development in old API. I will comment here after the meeting.

Shikugawa:
@douglas-reid How's it going?

douglas-reid:
> @douglas-reid How's it going?

OK-ish. We are still trying to get this in for 1.10.

wu-sheng:
@douglas-reid @nrjpoddar Could you consider merging this?

nrjpoddar:
> @douglas-reid @nrjpoddar Could you consider merging this?

The Telemetry API PR got merged yesterday. Can you take a look and see if you can use the new APIs?

Shikugawa:
> > @douglas-reid @nrjpoddar Could you consider merging this?
> 
> The Telemetry API PR got merged yesterday. Can you take a look and see if you can use the new APIs?

@nrjpoddar We don't have any specific changes to follow new API exclude a bit conflict. It was fixed already.

Shikugawa:
> > @douglas-reid @nrjpoddar Could you consider merging this?
> 
> The Telemetry API PR got merged yesterday. Can you take a look and see if you can use the new APIs?

@nrjpoddar @douglas-reid Fixed. Could you take a look. I don't think we should support SkyWalking tracer for old telemetry API.

mesh/v1alpha1/config.proto
```diff
@@ -612,6 +614,27 @@ message MeshConfig {
       uint32 max_tag_length = 3;
     }
 
+    // Defines configuration for a SkyWalking tracer.
+    message SkyWalkingTracingProvider {
+      // REQUIRED. Specifies the service for the SkyWalking OAP.
```
nrjpoddar, line 619: What’s an OAP. Can you spell it out here?

mesh/v1alpha1/config.proto
```diff
@@ -612,6 +614,27 @@ message MeshConfig {
...
+      // The format is "[<Namespace>/]<Hostname>". The specification of <Namespace> is required only when it is insufficient
```
nrjpoddar, line 620: Why do we need to support NS/ semantics rather than sticking with FQDN host format?

mesh/v1alpha1/config.proto
```diff
@@ -612,6 +614,27 @@ message MeshConfig {
...
+      // Example: "skywalking.default.svc.cluster.local" or "bar/skywalking.example.com".
+      string service = 1;
+
+      // REQUIRED. Specifies the port of the service.
+      uint32 port = 2;
+
+      // Optional. The SkyWalking OAP access token.
+      string access_token = 3;
+
+      // Optional. Controls the overall path length allowed in a reported span.
```
nrjpoddar, line 633: Is this a copy paste comment or this is how Skywalk works? Just making sure it’s actually supported like this.

nrjpoddar (commented):
Thanks for being patient with us and working through your changes as we were building the new API. I added minor comments. PTAL.

mesh/v1alpha1/config.proto
Shikugawa: It is because SkyWalking receiver is needed to be deployed on k8s cluster.

mesh/v1alpha1/config.proto
Shikugawa: Sorry. this feature is not supported on skywalking agent

mesh/v1alpha1/config.proto
Shikugawa: Should we adopt oneof semantics to assure we can accept FQDN and service name? (service will be rejected when validation)

nrjpoddar: Yes, let's only support FQDN notation with no short names

mesh/v1alpha1/config.proto
nrjpoddar: Can we remove this field since it's not supported?

mesh/v1alpha1/config.proto
```diff
@@ -616,23 +616,24 @@ message MeshConfig {
...
-      uint32 port = 2;
+      uint32 port = 3;
 
       // Optional. The SkyWalking OAP access token.
-      string access_token = 3;
-
-      // Optional. Controls the overall path length allowed in a reported span.
-      // NOTE: currently only controls max length of the path tag.
-      uint32 max_tag_length = 4;
+      string access_token = 4;
```
nrjpoddar, line 636: Should be 3..

Shikugawa: 3 is already used.

lizan:
This should be good to go. @nrjpoddar @douglas-reid any blocker?

mesh/v1alpha1/config.proto
```diff
@@ -683,6 +685,28 @@ message MeshConfig {
...
+      oneof receiver {
+        // The format is "[<Namespace>/]<Hostname>". The specification of <Namespace> is required only when it is insufficient
+        // to unambiguously resolve a service in the service registry. The <Hostname> is a fully qualified host name of a
+        // service defined by the Kubernetes service or ServiceEntry.
+        //
+        // Example: "skywalking.default.svc.cluster.local" or "bar/skywalking.example.com".
+        string service = 1;
+        
+        // Address of SkyWalking receiver.
+        string address = 2;
```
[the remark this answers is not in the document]
Shikugawa: @howardjohn I think it is needed because we should assure compatibility with old temetery API. The API accepts FQDN as the trace aggregator. We should dispose to include this chages into old API, so we need this here and accepts old-style FQDN and new-style service name.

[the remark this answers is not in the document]
howardjohn: I am not sure what you mean by old API - isn't here only one new Telemetry API? And its brand new so we shouldn't have legacy concerns?

Also given `service: skywalking.example.com` is valid, its not clear when I would even need `address` at all?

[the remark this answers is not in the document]
Shikugawa: Agreed. I have a concern about the difference of address (described such as [here](https://github.com/istio/api/blob/master/mesh/v1alpha1/proxy.proto#L47)). Is it correct that we don't have any difference between `address` on old API and `service` on new API?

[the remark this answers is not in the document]
howardjohn: There are differences but I don't think that mean we need to/can keep both around. All use cases provided by `address` can be satisfied with `service`.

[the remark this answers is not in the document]
Shikugawa: Sure. fixed.

Shikugawa:
@douglas-reid Could we merge this?

nrjpoddar (approved):

douglas-reid:
@Shikugawa thanks for adding. will you be owning the impl PR as well?

Shikugawa:
@douglas-reid Yes. I can do that.

Shikugawa:
@douglas-reid friendly ping

douglas-reid:
> @douglas-reid friendly ping

Unfortunately, I do not have approval powers here. We'll need a second approval from a blessed individual.

douglas-reid (approved):

douglas-reid:
Well, nevermind. I guess my powers were active :).

nrjpoddar:
> Well, nevermind. I guess my powers were active :).

Sorry to burst your bubble Doug but I had approved it before 😄

PR #1911, opened 2021-03-11
Fixing docs for cluster_local defaults.

nmittler:


smawson (approved):

howardjohn (approved):

mesh/v1alpha1/config.proto
```diff
@@ -386,8 +386,8 @@ message MeshConfig {
       //     this service-by-service (e.g. mysvc.myns.svc.cluster.local) or as a group
       //     (e.g. *.myns.svc.cluster.local).
       //
-      // By default, Istio will consider all services in the kube-system namespace to be cluster-local,
-      // unless explicitly overridden here.
+      // By default Istio will consider kubernetes.default.svc (i.e. the API Server) as well as all
+      // services in the kube-system namespace to be cluster-local, unless explicitly overridden here.
```
nrjpoddar, line 390: How can you override using this API? AFAICT API only supports making any service or wildcard host cluster local, not the other way around.

nmittler: You would set one of the default services/namespaces cluster-local value to false.

PR #1912, opened 2021-03-11
Support configuring the cipher suites for inbound TLS connections to workloads

lei-tang:
The API in this PR is for: https://github.com/istio/istio/issues/31380
The PR that depends on this API: https://github.com/istio/istio/pull/31381


mesh/v1alpha1/config.proto
```diff
@@ -573,8 +573,23 @@ message MeshConfig {
   // for additional detail on selector semantics.
   repeated k8s.io.apimachinery.pkg.apis.meta.v1.LabelSelector discovery_selectors = 59;
 
+
+  // The cipher suites for inbound TLS connections to workloads, delimited by comma.
+  // For the ciphers, please refer to the Envoy BoringSSL FIPS cipher list on
+	// https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto.html.
+	// The order matters: first one on the server list that's also supported
+	// by the client is used for the connection.
+  string tls_inbound_workload_cipher_suites = 60;
```
howardjohn, line 582: I think we would want to have this mirror the Gateway API. Maybe not the exact same proto since it has some fields not relevant, but a subset of it: https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings

Eventually/now we will also want to configure this in DestinationRule for outbound traffic, which should also use the same exact settings

ramaraochavali: +1 for mimicking gateway api. How about adding both to Sidecar like InboundTLSSettings and OutboundTLSSettings? Then would it be easier to configure both at a single place than going to mesh config for inbound and DR for outbound?

howardjohn: Sidecar doesn't allow setting it globally since we don't have merge semantics for Sidecar.

I am also not sure DR outbound should be global, since it will depend on each external svc?

ramaraochavali: > DR outbound should be global

For Mesh DRs it could be same and for external svc those DRs can override?

linsun: Is this configuring only the sidecar but not the gateway?  

Also, wouldn't it make sense to configure this gradually per workload so user could test it first?  Seems too forceful to configure this in mesh config.

hzxuzhonghu: This should be an array type. And definitely should be consistent with gateway settings

mesh/v1alpha1/config.proto
```diff
@@ -573,8 +573,23 @@ message MeshConfig {
...
+
+  // The default profile for the cipher suites of inbound TLS connections to workloads,
+  // which is used when tls_inbound_workload_cipher_suites is empty.
+  // When configuring the value of the default profile:
+  // "default": means that the default profile will be used.
+  // "restricted": means that the restricted profile will be used for higher security.
+  string default_profile_for_tls_inbound_workload_cipher_suites = 61;
```
hzxuzhonghu, line 589: This is not neccessary, by default we can provide a default cipher suites. This does not mean we should add another filed for this.

PR #1914, opened 2021-03-13
unhide bind in gateway api

ramaraochavali:
There are usecases for "bind" field in gateway so that gateway can communicate to Mesh clusters and listen on local bind (127.0.0.1) if specified here so that external clients of gateway can not access those services.

This is being discussed here https://docs.google.com/document/d/166XC7L8xNLMoGVRzDpl6_r-o9J9a--eA_DY1JAKlA7k/edit?ts=6045b7cd#heading=h.xw1gqgyqs5b

Draft PR for using this https://github.com/istio/istio/pull/31419

I can not think of any reason why this should not be supported except for it is not implemented.


releasenotes/notes/gateway-bind.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: networking
+issue:
+  - 31419
+
+releaseNotes:
+  - |
+    **Added** Support for specifying bind in gateway server definition.
```
linsun, line 9:
```suggestion
    **Added** Support for specifying bind to IP or Unix domain socket in gateway server definition.
```

linsun:
LGTM, thanks @ramaraochavali

networking/v1alpha3/gateway.proto
```diff
@@ -531,7 +531,6 @@ message Server {
   // connections.
   Port port = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // $hide_from_docs
```
howardjohn, line 534: can we look through history to get some background on why this was hide_from_docs in the first place? Was it ever implemented?

ramaraochavali: I looked at the code - it was never implemented.

howardjohn: Looks like it was added when we added Sidecar for some reason, I guess for consinstency

ramaraochavali:
@nrjpoddar Do you have any concerns unhiding it and implementing this? If not can you please approve this?

nrjpoddar (commented):
I don't have concerns but it would be better to get another Networking WG @howardjohn approve this.  I would like that you add validation where the config is rejected if port is specified with UDS bind address.

Additionally, the comment in docs should explain the use case, I don't quite get that from current comments. Why and when should a user configure this field?

ramaraochavali:
I can add comments. @howardjohn @hzxuzhonghu can you PTAL?

ramaraochavali:
@nrjpoddar just rereading your comment - are you suggesting to not support UDS at gateway?

nrjpoddar:
> @nrjpoddar just rereading your comment - are you suggesting to not support UDS at gateway?

Nope, I meant that if UDS is specified as bind address then port doesn’t make sense. TBT UDS itself doesn’t make much sense to me for Gateways but there might be some odd use cases.

ramaraochavali:
Yup. I will add that validation

ramaraochavali:
@nrjpoddar updated comments. PTAL when you get chance

networking/v1alpha3/gateway.proto
```diff
@@ -531,11 +531,14 @@ message Server {
...
   // The ip or the Unix domain socket to which the listener should be bound
   // to. Format: `x.x.x.x` or `unix:///path/to/uds` or `unix://@foobar`
   // (Linux abstract namespace). When using Unix domain sockets, the port
   // number should be 0.
+  // This is used to restrict the avaialability of this server to the gateway only. 
+  // This is typically used when a gateway needs to communicate to another mesh cluster 
+  // to publish metrics for example. In such case, the server created with the
+  // specified bind will not be available for gateway clients.
```
nrjpoddar, line 541:
```suggestion
  // This is used to restrict the reachability of this server to be gateway internal only. 
  // This is typically used when a gateway needs to communicate to another mesh service 
  // e.g. publishing metrics. In such case, the server created with the
  // specified bind will not be available to external gateway clients.
```

networking/v1beta1/gateway.proto
```diff
@@ -531,11 +531,14 @@ message Server {
   // connections.
   Port port = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // $hide_from_docs
   // The ip or the Unix domain socket to which the listener should be bound
   // to. Format: `x.x.x.x` or `unix:///path/to/uds` or `unix://@foobar`
   // (Linux abstract namespace). When using Unix domain sockets, the port
   // number should be 0.
+  // This is used to restrict the reachability of this server to be gateway internal only.
```
howardjohn, line 538: nit: I don't think this is the only use case. One other example is using hostNetwork and binding to a wireguard interface instead of all interfaces for example to only expose over VPN. So maybe the wording should be *can be used* not *is used*

howardjohn (approved):

ramaraochavali:
@nrjpoddar can you approve this when you get time?

ramaraochavali:
@nrjpoddar @hzxuzhonghu can you please approve this?

ramaraochavali:
@nrjpoddar @linsun @hzxuzhonghu ping. Can you please approve this?

nrjpoddar (approved):

PR #1917, opened 2021-03-16
Support configuring deny for AuthorizationPolicy

wzshiming:
xref https://github.com/istio/istio/issues/31452


howardjohn:
I am not sure how common this is that it warrants the API here. Is this an option exposed by other softwares offering authz?

If it is, is this the right place to put it? This is not a concern that applies only to authz - also to the authn rejections, and even virtual service direct response are highly related.

yangminzhu (changes requested):
Thanks for helping on improving this!

I think this change actually has a lot implications:
- API design, do we want to put this in authz policy or in meshconfig? This affects multiple things: who should have the permission to configure it. Should it be per-workload or per-mesh config?
- Implementation, how will this support the CUSTOM action which uses ext-authz filter for now?
- Implementation, for ALLOW/DENY, do we plan to use the `local_reply` in HCM or add first-class support in RBAC filter?

I suggest let's put this into a doc (could be a short one-page doc) for better discussion in the community (may also present it in the security WG) and get it approved before changing the API directly.

wzshiming:
@yangminzhu @howardjohn 
Thanks for reminding, I noticed the newly opened https://github.com/istio/enhancements repository,  open an issues to add documentation?

yangminzhu:
> @yangminzhu @howardjohn
> Thanks for reminding, I noticed the newly opened https://github.com/istio/enhancements repository, open an issues to add documentation?

I think you could start a short doc with some more details about the API change, share it to the community for comments and if possible bring it up in the security WG for some discussions. See the guidance in https://github.com/istio/enhancements#when-to-create-a-new-enhancement-issue

PR #1918, opened 2021-03-17
Revision is no longer experimental

shamsher31:
Not sure what is the plan for the `revision` flag but the `revision` is pretty stable across `install`, `upgrade`, and `operators`. It was added as an experimental year ago in 1.5.

Ref: https://github.com/istio/api/pull/1260


smawson:
Is this field linked to a feature that has been promoted, or should be promoted? I think we're still figuring out the process of promoting fields...

linsun:
Do we have some automated test that tests this?  Last time when we brought this field to graduate out of experimental, testing was a concern.

shamsher31:
> Is this field linked to a feature that has been promoted, or should be promoted? I think we're still figuring out the process of promoting fields...

There are a couple of PR (https://github.com/istio/istio/pull/23547 and https://github.com/istio/istio/pull/25716) that adds the ability to handle revision. But does not talk about whether it was experimental. 

There is also a graduation doc but it does not mention gradutaiton of `revision` for any subcommand.
https://docs.google.com/document/d/1sHpqT8mX8BN7DaoGd8rWk0cSjqdsTkxJy5zX2Cyfmp0/

shamsher31:
> Do we have some automated test that tests this? Last time when we brought this field to graduate out of experimental, testing was a concern.

@linsun Recently `istioctl x revision` has been added https://github.com/istio/istio/pull/29713 which do have an integration test https://github.com/istio/istio/pull/30711, but it is different from `--revision` flag in `IstioOperatorSpec`. So I think it's safe to merge this PR.

nrjpoddar:
@shamsher31 I think we are a bit confused as to how this revision field is used. Can you elaborate?

shamsher31:
@nrjpoddar Sure.

Following are the ways currently we use `revision` flag.

```
$ istioctl install --revision 1-9-1
$ istioctl install --set revision=canary
$ istioctl x uninstall --revision 1-6-5
```

```
$ istioctl operator init --revision 1-9-1
```

And this PR just removed the `This option is currently experimental` comment.

nrjpoddar:
I see. But the feature is still experimental afaict, so why remove this comment?

shamsher31:
> I see. But the feature is still experimental afaict, so why remove this comment?

@nrjpoddar can you please point me to the right docs to verify which features are still experimental?

nrjpoddar:
> > I see. But the feature is still experimental afaict, so why remove this comment?
> 
> @nrjpoddar can you please point me to the right docs to verify which features are still experimental?

I looked but I can't find. I actually don't know if someone has actually sat down and figured out what's the correct maturity level of "revision" based upgrade is today. @ostromart @howardjohn do you know?

I do know that https://istio.io/latest/docs/setup/upgrade/gateways/ is marked experimental.

linsun:
the revision semantics changed in 1.8, I am very sure it was experimental for 1.8.   I think @brian-avery is trying to promote it via our promotion process.

PR #1919, opened 2021-03-17
Fix broken links for locations

shamsher31:
Similar to https://github.com/istio/api/pull/1907

Other `kubectl explain kind` are also have a broken links.

```
$ k explain Gateway.spec
KIND:     Gateway
VERSION:  networking.istio.io/v1beta1

RESOURCE: spec <Object>

DESCRIPTION:
     Configuration affecting edge load balancer. See more details at:
     https://istio.io/docs/reference/config/networking/gateway.html
```

```
$ k explain WorkloadGroup.spec
KIND:     WorkloadGroup
VERSION:  networking.istio.io/v1alpha3

RESOURCE: spec <Object>

DESCRIPTION:
     Describes a collection of workload instances. See more details at:
     https://istio.io/docs/reference/config/networking/workload-group.html
```

### DIFF
--- a/security/v1beta1/authorization_policy.proto
+++ b/security/v1beta1/authorization_policy.proto
@@ -19,7 +19,7 @@ import "type/v1beta1/selector.proto";
 // $schema: istio.security.v1beta1.AuthorizationPolicy
 // $title: Authorization Policy
 // $description: Configuration for access control on workloads.
-// $location: https://istio.io/latest/docs/reference/config/security/authorization-policy.html
+// $location: https://istio.io/docs/reference/config/security/authorization-policy.html
 // $weight: 20
 // $aliases: [/docs/reference/config/authorization/authorization-policy]
 
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -5423,7 +5423,7 @@ spec:
         properties:
           spec:
             description: 'Configuration for access control on workloads. See more
-              details at: https://istio.io/latest/docs/reference/config/security/authorization-policy.html'
+              details at: https://istio.io/docs/reference/config/security/authorization-policy.html'
             oneOf:
             - not:
                 anyOf:
--- a/security/v1beta1/authorization_policy.pb.html
+++ b/security/v1beta1/authorization_policy.pb.html
@@ -1,7 +1,7 @@
 ---
 title: Authorization Policy
 description: Configuration for access control on workloads.
-location: https://istio.io/latest/docs/reference/config/security/authorization-policy.html
+location: https://istio.io/docs/reference/config/security/authorization-policy.html
 layout: protoc-gen-docs
 generator: protoc-gen-docs
 schema: istio.security.v1beta1.AuthorizationPolicy


howardjohn (approved):
we should add a redirect if possible as well

shamsher31:
> we should add a redirect if possible as well

I found later today that If we remove the `.html` extension, the redirect happens automatically (even though `/latest` was not there in the URL). But both achieve the same thing so I'm keeping `/latest` in the URL for now.

ericvn:
Note my comment in another thread that latest caused an issue in the docs where we can't consume the change and I will look at that at some point.

ericvn:
Looking at the scrapping code, it only pulls files whose location contains `https://istio.io/docs`. The logic basically converts the location to be a directory within the istio repo, and that repo does not have `latest` within it's directory structure so no reason to try copying the file. We would need to update the code which copies the files to also strip `latest` from the location the file is placed.

Earlier, I had made some changes so links within the docs could be updated to include `latest` and we strip that portion out of the URLs after copying the file so links point to actual locations (there is no redirection).

ericvn:
@shamsher31 Based on the above, I think if you removed `latest`, the istio.io script copying the file would work correctly, and removing the `.html` will make the location: value be a link that someone could put in a browser and it would be found. So something like:
```
https://istio.io/docs/reference/config/istio.analysis.v1alpha1
```
seems to make sense. I haven't verified that the removal of the .html won't break something else.

ericvn:
As a quick test, I created a branch with a revert of the earlier PR (so I can get the linter to pass) plus one of the changes in this PR: https://github.com/ericvn/api/commit/ebc3381af07df6a4cabd53e684dbaf1e4d666501

I then ran the istio.io code against that branch, which copied the files. I ran a `make lint` which passed as well as a `make serve` and tested the pages.

Based on that, it seems that simply removing the `.html` will work.

If it doesn't make sense to remove the `.html`, maybe @davidhauck can help with the redirection, or I will need to do some change in the scrapping script to also pull pages with a `location:` containing `latest` and remove that during the copy.

davidhauck:
The logic for the redirect adds a redirect for every page by removing "/latest" from the path and redirecting it to the path.

E.g. there is a redirect for:
docs/reference/config/networking/gateway -> latest/docs/reference/config/networking/gateway

There is not a redirect for:
docs/reference/config/networking/gateway.html -> latest/docs/reference/config/networking/gateway

If adding this redirect will make things easier, I could add it. We should try to ensure page names should be formatted like `docs/reference/config/networking/gateway` and NOT like `docs/reference/config/networking/gateway.html`

ericvn:
@davidhauck We have code today to strip the `.html` from the end of the location if it exists when copying the file.

I created https://github.com/istio/istio.io/pull/9287 which will allow the location to have `latest` or not, the file will still be scraped. Once merged, we should allow locations with to without `latest` and if there is a `.html` at the end, it is removed from the destination file location.

`location: https://istio.io/latest/docs/reference/config/security/authorization-policy.html` -> `docs/reference/config/security/authorization-policy`

`location: https://istio.io/docs/reference/config/security/authorization-policy.html` -> `docs/reference/config/security/authorization-policy`

`location: https://istio.io/docs/reference/config/security/authorization-policy.` -> `docs/reference/config/security/authorization-policy`

ericvn:
We will want to revert the other PR to remove `latest` (maybe just do here).

shamsher31:
> We will want to revert the other PR to remove `latest` (maybe just do here).

Yup, I will also update https://github.com/istio/api/pull/1907 here.

shamsher31:
> @shamsher31 So, once that PR is done, I think the locations can basically stay the same (without latest and with .HTML).

I will keep this PR open for now. Once redirect is done. We can either close this or merge.

ericvn:
@shamsher31 So the redirection PR is in for istio.io release-1.9 and the original locations like: `https://istio.io/docs/reference/config/istio.analysis.v1alpha1.html` no longer get a 404, but actually end up on pages.

shamsher31:
@ericvn Thanks for verifying.

security/v1beta1/authorization_policy.proto
```diff
@@ -19,7 +19,7 @@ import "type/v1beta1/selector.proto";
 // $schema: istio.security.v1beta1.AuthorizationPolicy
 // $title: Authorization Policy
 // $description: Configuration for access control on workloads.
-// $location: https://istio.io/latest/docs/reference/config/security/authorization-policy.html
+// $location: https://istio.io/latest/docs/reference/config/security/authorization-policy
```
ericvn, line 22: @shamsher31 Remove the `latest` here.

ericvn (changes requested):

ericvn:
As for adding back the `.html` location, it doesn't need to be removed anymore.

security/v1beta1/authorization_policy.proto
shamsher31: Done.

shamsher31:
> As for adding back the `.html` location, it doesn't need to be removed anymore.

@ericvn Updated the PR to only remove `latest` and keeping `.html` as it is (will be handle by redirect rule). Thanks.

ericvn (approved):
LGTM. Basically this is a revert of #1907. The fix for the original issue was adding istio.io redirects.

shamsher31:
redirect rule PR: https://github.com/istio/istio.io/pull/9288

ericvn:
@istio/technical-oversight-committee Looking for approval for this one to remove the blockage of reference doc updates into istio.io. As noted, this is a revert of #1907 and the original issue was fixed instead by istio/istio.io#9288.

nrjpoddar (approved):

PR #1920, opened 2021-03-17
Fix all wrong locations

zxh326:
Replace all kind locations
`https://istio.io/docs` to `https://istio.io/latest/docs/`


zxh326:
May be duplicate with https://github.com/istio/api/pull/1919,
/close

PR #1923, opened 2021-03-23
Add SPIFFE independent multiple trust domain API.

mathetake:
This PR adds API for https://github.com/istio/istio/issues/30454

References:
- [Design doc](https://docs.google.com/document/d/1_WcOYZo8MkWMOsLMH3PyPkRW1tUYdQJkqasbCemzr08/edit)
- https://github.com/envoyproxy/envoy/issues/14614

Signed-off-by: Takeshi Yoneda <takeshi@tetrate.io>


mathetake:
cc @lizan

lizan:
@myidpt @howardjohn PTAL

mesh/v1alpha1/config.proto
```diff
@@ -222,6 +222,13 @@ message MeshConfig {
       // The certificate is retrieved from the endpoint.
       string spiffe_bundle_url = 2;
     }
+
+    // Optional. Specify the trust domain to which this certificate data belongs.
```
howardjohn, line 226: what happens if its not set?

mesh/v1alpha1/config.proto
```diff
@@ -230,6 +237,25 @@ message MeshConfig {
...
+  // For example, consider the case where a cluster A manages the trust domain "aaa.com", cluster B manages "bbb.com".
+  // When services in cluster A talk to services in cluster B, services in cluster B have to validate client certificates
+  // based on the root certificates of cluster A. However, if we merge root certificates from both cluster A and cluster B,
+  // then we cannot trust that presented client certificates are actually signed by the corresponding cluster. That is because
+  // cluster A might sign client certificates with "spiffe://bbb.com/**" URI SAN in order to bypass
+  // cluster B's authorization based on workload identity. In that case, we say these trust domains are NOT isolated from each other.
+  // This is why we need to select root certificates based on the trust domain before verify client certificates.
+  //
+  // Please note that this setting requires `STRICT` mTLS mode to take into effect.
+  bool enabled_spiffe_multiple_independent_trust_domain_authentication = 60;
```
howardjohn, line 257: Do we need this option? Or should putting `trust domain` field in `CertificateData` sufficient to enable this functionality?

mathetake: good point. I think for now it's enough to just enable this feature when `CertificateData.trust_donain` is given. But what if  other features want to use `trust_domain` fields while keeping the original normal mTLS cert validation scheme? Then enabling automatically this feature with `trust_domain` field might be problematic, though I'm not sure if there is such case in the future. WDYT?

mesh/v1alpha1/config.proto
```diff
@@ -222,6 +222,13 @@ message MeshConfig {
       // The certificate is retrieved from the endpoint.
       string spiffe_bundle_url = 2;
     }
+
+    // Optional. Specify the trust domain to which this certificate data belongs.
+    // If `enabled_spiffe_multiple_independent_trust_domain_authentication` enabled, this must be set for
+    // distinguishing different trust domains from each other. Certificates in a same trust domain will be merged
```
costinm, line 228: I don't know where enabled_long_option is set, but I don't think it is or should be part of the API.

If trust_domain is set, it'll be used for this root CA. If not set - the root CA will be used for the default trust domain and its aliases.

mesh/v1alpha1/config.proto
costinm: Please remove this. Not only it is not needed - presence of trust_domain in a cert is sufficient, but the name is pretty horrible ( and I'm bad with names ). 

Also please don't include too much SVID/SPIFFE in the API docs - the concept of 'trust domain' is not tied to Spiffe
( and the API in general should not be tied to either Spiffe or mTLS )

security/v1beta1/authorization_policy.proto
```diff
@@ -410,6 +410,10 @@ message Source {
 
   // Optional. A list of negative match of remote IP blocks.
   repeated string not_remote_ip_blocks = 10;
+
+  // Optional. A list of trust domains of client certificates.
+  // This field requires mTLS enabled.
```
costinm, line 415: Does it ? Isn't trust domain logically equivalent with the OIDC issuer ? 

I don't want to have options tied with transport options or protocols.

costinm: Also, 'if not specified, the default trust domain and its aliases will be used'

mesh/v1alpha1/config.proto
```diff
@@ -222,6 +222,13 @@ message MeshConfig {
...
+    // and used together to validate the trust domain.
+    // If `enabled_spiffe_multiple_independent_trust_domain_authentication` disabled, it is not necessary to set.
+    string trust_domain = 3;
```
howardjohn, line 231: can you have multiple entries with the same trust_domain? Preferably answer this in the form of a comment in the proto :slightly_smiling_face:

mathetake: actually it is possible to assign multiple trust domains to a ca cert in Envoy side, and it makes sense to have multiple ones here. I made  the change accordingly.

security/v1beta1/authorization_policy.proto
mathetake: Thanks, done!

mesh/v1alpha1/config.proto
howardjohn: Agreed, it makes sense to allow both. Thanks for clarifying

mathetake: oops, I misunderstood your comment, let me add comments.

mesh/v1alpha1/config.proto
mathetake: done. Thanks!

mesh/v1alpha1/config.proto
mathetake: https://github.com/istio/api/pull/1923/commits/eeb68137c44bb186caa44ec221373c95c5eea704 done :+1:

mesh/v1alpha1/config.proto
```diff
@@ -222,6 +222,13 @@ message MeshConfig {
       // The certificate is retrieved from the endpoint.
       string spiffe_bundle_url = 2;
     }
+
+    // Optional. Specify the list of trust domains to which this certificate data belongs.
+    // If set, they are used for this root CA. Otherwise, this root CA is used for default trust domain
+    // and its aliases.
+    // Note that we can have multiple certificate data for a same trust_domain.
+    // In that case, certificates with a same trust domain will be merged and used together to verify peer certificates.
+    repeated string trust_domains = 3;
```
nrjpoddar, line 231: How does this relate to trust domain defined above?

nrjpoddar: I see. May be add a comment related to it? I found it confusing so asked.

linsun: Can we clarify this?  Even in the names, such as cluster_trust_domains or mesh_trust_domains?

mathetake: I don't think this is that confusing since this is a field of `CertificateData` which is not in the top-level MeshConfig. But I agree that this is a bit confusing. @costinm @howardjohn thoughts?

hzxuzhonghu, line 231: why is the trust_domains plural, does that means the certificate_data can belong to multi trust domains?

mathetake: yes. See discussion her: https://github.com/istio/api/pull/1923#discussion_r603709658 for example, it can be used for aliasing trust domains.

mathetake:
@costinm @howardjohn could you take another look? Thanks!

security/v1beta1/authorization_policy.proto
```diff
@@ -410,6 +410,10 @@ message Source {
 
   // Optional. A list of negative match of remote IP blocks.
   repeated string not_remote_ip_blocks = 10;
+
+  // Optional. A list of trust domains of peer certificates.
```
howardjohn, line 414: If I have mesh config:
```
trustDomain: foo
trustDomainAliases: [bar, baz]
```


and I set trustDomains=foo, do `bar` and `baz` apply?
what if I set trustDomains=baz?

Basically are aliases applied here?

howardjohn (commented):
LGTM, one small clarification

security/v1beta1/authorization_policy.proto
mathetake: Thanks for putting another insight. I think we should implicitly apply `foo` and `bar` as well when only setting `trustDomains=baz` in line with the doc of trustDomainAliases saying "aliases are treated as the same trust domain". And this is how we can treat foo, bar and baz as the same trust domain. 

so maybe should i put the comment about this behavior here?

mesh/v1alpha1/config.proto
shankgan: Is there a corresponding PR in the Istio Repo yet to handle Certificates for multiple TrustDomains? As of now, it is assumed that all these trustAnchors are intended for the configured mesh TrustDomain

costinm (commented):
I am a bit concerned with the 3 names - trust_domain, trust_domain_aliasses and trust_domains. Note that trust_domain is used to indicate what we'll put in the certs we generate, and the aliases was introduced mainly for migrations - and doesn't make sense once
we have the more explicit trust bundles.

From the Spiffee spec: "implementations MUST securely associate bundle endpoints with their respective trust domain, nominally through explicit configuration of a <trust_domain_name, endpoint> tuple" - no mention of aliases, which can be represented by having more tuples ( except that it is discouraged to have same keys associated with different names !)

https://tools.ietf.org/html/rfc5280#section-4.2.1.10 has the concept of 'name constraints' and 
'permittedSubtrees' - which can apply to both Spiffee URL SANs but also domain SANs and even IP certs. However those are built into the cert PEM - so wouldn't need any configuration on our side. 

Same for OIDC - the 'trust domain' is built into the discovery URL.

In other words: I'm worried we're again reinventing the wheel and confusing people, like we did with few other things in security.

I am also worried about the 'mesh wide' / mesh operator applicability of the config - instead of using ProxyConfig which allows more specific targeting. Since DestinationRule, Policy, etc are driven by the mesh user - I don't see why the trust endpoints would be mesh-wide and controlled by operator. 

So my alternate proposals:
- add this to ProxyConfig instead ( or maybe even DestinationRule - combined with the proposal we just heard in Env/networking for improving verification of SANs )
-  either use a 'trustDomain' ( without aliases - see spiffee spec/recommendation ), or use the RFC5280 naming, mentioning that the constraints can also be loaded from the cert itself. Since this kind of constraining of a root cert is in no way specific to spiffee URL SANs - I would very much prefer the 'constraints' approach. 
- If the URL form is specified I would just document that for Istio we recommend using the OIDC format for the URL, which includes the 'trust domain' in the URL in a well-known form, and validate that.

costinm:
I would also request having this reviewed by Env WG ( MeshConfig vs ProxyConfig ) and Networking ( vs DestinationRule, and how it can fit/interact with the new proposal for validation in DR ). In particular we have a pretty big disconnect between networking - defining SAN validations in DR and client SAN validation in Gateway - and mesh config terms that are only used in Spiffee ( like the fact we validate ones by default and others not, ownership of configs, etc )

security/v1beta1/authorization_policy.proto
myidpt: Yes, please. In this case, we should always treat foo, bar, baz as identical.
Normally, users should not specify the trust_domains when they apply the authz policy to the local trust domain. But if they did, trust domain aliasing also applies here.

mathetake:
@costinm  Thank you very much for putting your thoughts in detail.

> add this to ProxyConfig instead ( or maybe even DestinationRule - combined with the proposal we just heard in Env/networking for improving verification of SANs )

Could you add reviews from these WG to this PR? And I would appreciate it if you could put the link to the proposal.

> either use a 'trustDomain' ( without aliases - see spiffee spec/recommendation ), or use the RFC5280 naming, mentioning that the constraints can also be loaded from the cert itself. Since this kind of constraining of a root cert is in no way specific to spiffee URL SANs - I would very much prefer the 'constraints' approach.

Agreed and now I would love to go with 'naming constraints' and I'm even thinking of removing 'trust_domain(s)' field, which seems much clearer and standardized. Does it make sense to you to remove the field and only support loading domains from naming constrains?

> If the URL form is specified I would just document that for Istio we recommend using the OIDC format for the URL, which includes the 'trust domain' in the URL in a well-known form, and validate that.

sg but if a user wants to load a "inline" cert (via "pem" field or equivalent in the future), then this cannot be available. So I guess this would introduce another confusion on users. I would prefer we stick to either "naming constrians" approach or "trust_domain" (without "s") approach as you proposed above. WDYT?

PR #1812, opened 2021-01-12
add volume and volumeMount to k8s settings.

morvencao:
add volume and volumeMount to k8s settings API for : https://github.com/istio/istio/issues/29973

PR #1925, opened 2021-03-25
fix volume marshal/unmarshal issue for operator api.

morvencao:
try to resolve: https://github.com/istio/istio/issues/31630
follow up PR for https://github.com/istio/api/pull/1812


operator/v1alpha1/operator.proto
```diff
@@ -294,7 +294,8 @@ message KubernetesResourcesSpec {
     // k8s volume
     // [https://kubernetes.io/docs/concepts/storage/volumes/](https://kubernetes.io/docs/concepts/storage/volumes/)
     // Volumes defines the collection of Volume to inject into the pod.
-    repeated k8s.io.api.core.v1.Volume volumes = 17;
+    // Duplicate the volume definition here to resolve the marshal/unmarshal issue caused by the inline field in the k8s api.
```
shamsher31, line 297: Better remove this comment as it will get appear on istio.io docs.
```suggestion
 
```

morvencao: thanks, done.

howardjohn (commented):
I don't really understand why we need this. Kubernetes types already have all the marshaling logic required. Why can we not just utilize that? AFAIK even a simple `json.Unmarshal` works

morvencao:
@howardjohn Let me explain this. We found that Kubernetes types has some inline field, for example `Handler` field in `Probe`(https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2098) and `VolumeSource` field in `Volume`(https://github.com/kubernetes/api/blob/master/core/v1/types.go#L46)
These inline fields will not exist in the serialized objects(yaml/json), but do exist in golang type.
If we reuse the Kubernetes types, which means we import Kubernetes' proto in istio's proto and generate the istio's golang types, then the inline metadata will be lost, the marshal/unmarshal will not work as expected, just as what you see in  https://github.com/istio/istio/issues/31630

howardjohn:
> If we reuse the Kubernetes types, which means we import Kubernetes' proto in istio's proto and generate the istio's golang types, then the inline metadata will be lost, the marshal/unmarshal will not work as expected

Why? If we use k8s protos, then we will also use k8s go types, which have the `inline` tag, right?

I don't really feel great about attempting to mirror their massive API...

ostromart:
@howardjohn this is partially lost in the mists of time but i think we had to go down this path because we use proto differently from k8s. we generate our structs from proto3 and use those, while k8s start with go structs and generate proto2 from those. this forces us to use different unmarshal logic, which doesn't integrate with the k8s json code. our unmarshal won't end up calling the k8s unmarshal for a given type. but i don't recall everything we tried so maybe it's worth trying this again.  if we could eliminate duplicating k8s protos it would be wonderful.

ostromart:
@litong01 fyi

PR #1926, opened 2021-03-25
update the ext-authz extension provider and promote to alpha

yangminzhu:
This PR makes necessary change to promote the ext-authz feature to alpha:

1. Added `with_request_body` to support including body in the check request: This addresses user feedback: https://github.com/istio/istio/issues/31182#issue-820305857 and https://github.com/istio/istio/issues/27790#issuecomment-805800853
2. Added `prefix` and `suffix` matching when including headers in the check request: This matches the syntax in the authorization policy, this is needed to simplify the configuration when there are large number of headers with same prefix (e.g. `x-forwarded-access-token`, `x-forwarded-user`, `x-forwarded-email`, ...) need to be included
3. Added `add_headers_in_check` to support including extra headers in the check request: This is just another way to include more headers.
4. Added `timeout` to support configuring the timeout to the external service: This is needed as we have found that the timeout in DestinationRule does not cover the ext-authz filter usage because the ext-authz filer has its own timeout constraints. The current default value 600s is chosen to match the current workaround, we could change it to a lower value (e.g. 200ms) if needed.
5. Updated the authorization_policy.proto documentation


yangminzhu:
@louiscryan @smawson @nrjpoddar @linsun please take a look at these changes for promoting the ext-authz feature to alpha in 1.10, let me know if you have any questions, thank you.

mesh/v1alpha1/config.proto
```diff
@@ -487,6 +502,9 @@ message MeshConfig {
       // REQUIRED. Specifies the port of the service.
       uint32 port = 2;
 
+      // Sets the maximum duration in milliseconds that a response can take to arrive upon request (default timeout: 600s).
```
smawson, line 505: Reword to say that this is the maximum duration that the proxy will wait for a response from the provider.

Also this is not in milliseconds, this is a google.protobuf.Duration, not a string.

mesh/v1alpha1/config.proto
```diff
@@ -475,6 +475,21 @@ message MeshConfig {
       EnvoyExternalAuthorizationGrpcProvider envoy_ext_authz_grpc = 3;
     }
 
+    message EnvoyExternalAuthorizationRequestBody {
+      // Sets the maximum size of a message body that the filter will hold in memory. Envoy will return HTTP 413 and
```
smawson, line 479: The interaction between this and the following field needs to be more clearly described.

You want to say that if max_request_bytes is reached, and allow_partial_message is false, Envoy will return a 413. Otherwise the request will be sent to the provider with a partial message.

mesh/v1alpha1/config.proto
```diff
@@ -475,6 +475,21 @@ message MeshConfig {
...
+      // will not initiate the authorization process when buffer reaches the number set in this field. Note that this
+      // setting will have precedence over failure_mode_allow.
+      uint32 max_request_bytes = 1;
+
+      // When this field is true, Envoy will buffer the message until max_request_bytes is reached.
+      // The authorization request will be dispatched and no 413 HTTP error will be returned by the filter.
+      bool allow_partial_message = 2;
+
+      // If true, the body sent to the external authorization service is set with raw bytes, it sets the raw_body field
```
smawson, line 488: Clarify this a bit too. By default the request body will contain what? A JSON object with a field set to the original body? Something else? Be specific about how this works.

mesh/v1alpha1/config.proto
```diff
@@ -505,12 +523,26 @@ message MeshConfig {
       // Note that in addition to the headers supplied by users:
       // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
       // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      //
+      // Exact, prefix and suffix presence match is supported:
```
smawson, line 527:
```suggestion
      // Exact, prefix and suffix matches are supported:
```

mesh/v1alpha1/config.proto
```diff
@@ -505,12 +523,26 @@ message MeshConfig {
...
+      // - Exact match: "abc" will match on value "abc".
+      // - Prefix match: "abc*" will match on value "abc" and "abcd".
+      // - Suffix match: "*abc" will match on value "abc" and "xabc".
       repeated string include_headers_in_check = 6;
 
+      // Set of headers that will be included in the authorization request sent to the authorization service.
+      // Note that client request of the same key will be overridden.
```
smawson, line 534: This needs to be clarified, using reference to the include_headers_in_check field. Basically any headers added here will override headers from include_headers_in_check or headers that are automatically sent.

Also we should consider something that validates the user doesn't specify the same header in include_headers_in_check and in additional headers.

mesh/v1alpha1/config.proto
```diff
@@ -505,12 +523,26 @@ message MeshConfig {
...
+      map<string, string> add_headers_in_check = 10;
```
smawson, line 535: I don't love this name, maybe additional_headers_in_check?

mesh/v1alpha1/config.proto
smawson: Same fix everywhere you added prefix and suffix matching.

mesh/v1alpha1/config.proto
```diff
@@ -520,7 +552,17 @@ message MeshConfig {
...
       // Note, the body from the authorization service is always included in the response to downstream.
+      //
+      // Exact, prefix and suffix presence match is supported:
+      // - Exact match: "abc" will match on value "abc".
+      // - Prefix match: "abc*" will match on value "abc" and "abcd".
+      // - Suffix match: "*abc" will match on value "abc" and "xabc".
       repeated string headers_to_downstream_on_deny = 8;
+
+      // Enables the ext-authz filter to buffer the client request body and send it within the authorization request.
+      // A "x-envoy-auth-partial-body: false|true" metadata header will be added to the authorization request message
```
smawson, line 563: Comment about partial header should go with the field allowing partial bodies.

mesh/v1alpha1/config.proto
```diff
@@ -520,7 +552,17 @@ message MeshConfig {
...
+      // indicating if the body data is partial.
+      EnvoyExternalAuthorizationRequestBody with_request_body = 11;
```
smawson, line 565: Maybe include_request_body to match naming?

include_request_body_in_check would exactly mirror include_headers_in_check.

mesh/v1alpha1/config.proto
```diff
@@ -543,6 +588,11 @@ message MeshConfig {
       // Sets the HTTP status that is returned to the client when there is a network error to the authorization service.
       // The default status is "403" (HTTP Forbidden).
       string status_on_error = 4;
+
+      // Enables the ext-authz filter to buffer the client request body and send it within the authorization request.
+      // A "x-envoy-auth-partial-body: false|true" metadata header will be added to the authorization request message
```
smawson, line 593: Same comment on this going on the partial field.

mesh/v1alpha1/config.proto
```diff
@@ -543,6 +588,11 @@ message MeshConfig {
...
+      // indicating if the body data is partial.
+      EnvoyExternalAuthorizationRequestBody with_request_body = 5;
```
smawson, line 595: include_request_body or include_request_body_in_check

security/v1beta1/authorization_policy.proto
```diff
@@ -25,18 +25,19 @@ import "type/v1beta1/selector.proto";
 
 // Istio Authorization Policy enables access control on workloads in the mesh.
 //
-// Authorization policy supports both allow and deny policies. When allow and
-// deny policies are used for a workload at the same time, the deny policies are
-// evaluated first. The evaluation is determined by the following rules:
+// Authorization policy supports CUSTOM, DENY and ALLOW action for access control. When CUSTOM, DENY and ALLOW actions
```
smawson, line 28:
```suggestion
// Authorization policy supports CUSTOM, DENY and ALLOW actions for access control. When CUSTOM, DENY and ALLOW actions
```

security/v1beta1/authorization_policy.proto
```diff
@@ -25,18 +25,19 @@ import "type/v1beta1/selector.proto";
...
+// are used for a workload at the same time, the CUSTOM action is evaluated first, then the DENY action, and the last is ALLOW action.
```
smawson, line 29:
```suggestion
// are used for a workload at the same time, the CUSTOM action is evaluated first, then the DENY action, and finally the ALLOW action.
```

mesh/v1alpha1/config.proto
yangminzhu: Changed to `include_request_body_in_check`

mesh/v1alpha1/config.proto
yangminzhu: oh, yes, it's better to be consistent with the other header field, renamed to `include_request_body_in_check`

mesh/v1alpha1/config.proto
yangminzhu: renamed.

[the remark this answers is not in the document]
yangminzhu: Key is the header name and value is the header value.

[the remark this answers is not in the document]
howardjohn: I see. Headers are a `map[string][]string` not a `map[string]string` so I was not sure this is the best type, but looks like that is the API in VS already. I guess you only need  `map[string][]string` for parsing not writing since you can flatten them

mesh/v1alpha1/config.proto
yangminzhu: updated, this only works with the gRPC provider and the `body` or `raw_body` field is used to store the body of the request: https://github.com/envoyproxy/envoy/blame/cffb095d59d7935abda12b9509bcd136808367bb/api/envoy/service/auth/v3/attribute_context.proto#L147-L153

mesh/v1alpha1/config.proto
yangminzhu: updated, and yes we can check this in the validation.

mesh/v1alpha1/config.proto
```diff
@@ -475,6 +475,21 @@ message MeshConfig {
...
+      // will not initiate the authorization process when buffer reaches the number set in this field. Note that this
+      // setting will have precedence over failure_mode_allow.
+      uint32 max_request_bytes = 1;
+
+      // When this field is true, Envoy will buffer the message until max_request_bytes is reached.
+      // The authorization request will be dispatched and no 413 HTTP error will be returned by the filter.
+      bool allow_partial_message = 2;
+
+      // If true, the body sent to the external authorization service is set with raw bytes, it sets the raw_body field
+      // of HTTP request attribute context. Otherwise, it will be filled with UTF-8 string request body.
```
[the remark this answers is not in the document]
yangminzhu: updated the comment, this refers to the `body` and `raw_body` field in the check request: https://github.com/envoyproxy/envoy/blame/cffb095d59d7935abda12b9509bcd136808367bb/api/envoy/service/auth/v3/attribute_context.proto#L147-L153

mesh/v1alpha1/config.proto
```diff
@@ -487,6 +502,9 @@ message MeshConfig {
       // REQUIRED. Specifies the port of the service.
       uint32 port = 2;
 
+      // Sets the maximum duration in milliseconds that a response can take to arrive upon request (default timeout: 600s).
+      google.protobuf.Duration timeout = 9;
```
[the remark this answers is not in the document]
yangminzhu: yeah I think this is probably the best we can do for the short term.
It will require quite some work to support VS in the ext-authz filter in Envoy. Maybe we could add the timeout also in DR for default value when not specified in VS as the long term solution?

mesh/v1alpha1/config.proto
```diff
@@ -505,12 +523,26 @@ message MeshConfig {
       // Note that in addition to the headers supplied by users:
       // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
       // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      //
+      // Exact, prefix and suffix presence match is supported:
+      // - Exact match: "abc" will match on value "abc".
+      // - Prefix match: "abc*" will match on value "abc" and "abcd".
+      // - Suffix match: "*abc" will match on value "abc" and "xabc".
       repeated string include_headers_in_check = 6;
```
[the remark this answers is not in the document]
yangminzhu: yes, sometimes there are unknown headers with common prefix or suffix (e.g. `x-forwarded-*`, `*-debug`). `HTTPMatchRequest` only supports exact match on header name but I guess it can later be extended to support prefix/suffix if needed?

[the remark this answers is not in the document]
howardjohn: VS does support it: https://istio.io/latest/docs/reference/config/networking/virtual-service/#StringMatch

mesh/v1alpha1/config.proto
```diff
@@ -475,6 +475,21 @@ message MeshConfig {
       EnvoyExternalAuthorizationGrpcProvider envoy_ext_authz_grpc = 3;
     }
 
+    message EnvoyExternalAuthorizationRequestBody {
```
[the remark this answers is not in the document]
yangminzhu: The `Envoy` is only in the type so the user won't have it in the config, and we actually already have the it in the `EnvoyExternalAuthorizationHttpProvider envoy_ext_authz_http` and `EnvoyExternalAuthorizationGrpcProvider envoy_ext_authz_grpc`.

We're treating the Envoy ext-authz similar to tracking providers: `zipkin`, `lightstep`, `datadog` and `opencensus`. Each tracer represents a specific tracing API with its own configuration. The `Envoy ext-authz` is also considered a specific type of external authz with its own configuration. I think the Istio abstraction over Envoy is the corresponding Istio Telemetry API or the AuthorizationPolicy (CUSTOM action) in this case.

For other data plane, maybe basic networking functionality is supported but I believe a lot of critical security features are not there, e.g., JWT, Authz, SDS. etc.

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
yangminzhu: it's for header value but not header keys right?

mesh/v1alpha1/config.proto
yangminzhu: updated.

mesh/v1alpha1/config.proto
```diff
@@ -514,13 +537,33 @@ message MeshConfig {
       // List of headers that should be included in the authorization request sent to the authorization service.
       // Note that in addition to the headers supplied by users:
       // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
-      // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      // 2. *Content-Length* will be set to 0 and the request will not have a message body. However, the authorization
+      // request can include the buffered client request body (controlled by include_request_body_in_check setting),
+      // consequently the value of Content-Length of the authorization request reflects the size of its payload size.
+      //
+      // Exact, prefix and suffix matches are supported:
```
nrjpoddar, line 544: This is confusing. I would make this consistent with other APIs like VS where we have string match semantics.

mesh/v1alpha1/config.proto
```diff
@@ -514,13 +537,33 @@ message MeshConfig {
...
+      // - Exact match: "abc" will match on value "abc".
+      // - Prefix match: "abc*" will match on value "abc" and "abcd".
+      // - Suffix match: "*abc" will match on value "abc" and "xabc".
       repeated string include_headers_in_check = 6;
 
+      // If set, the client request body will be included in the authorization request sent to the authorization service.
+      EnvoyExternalAuthorizationRequestBody include_request_body_in_check = 10;
```
nrjpoddar, line 551: Why are these higher numbered proto messages in middle? I’m guessing you want to group them but seeing out of order fields is bit of maintenance  issue.

mesh/v1alpha1/config.proto
```diff
@@ -545,6 +593,9 @@ message MeshConfig {
       // REQUIRED. Specifies the port of the service.
       uint32 port = 2;
 
+      // The timeout for the gRPC request. This is the timeout for a specific request (default timeout: 600s).
```
nrjpoddar, line 596: Can you define the behavior what happens after timeout fires?

mesh/v1alpha1/config.proto
```diff
@@ -497,6 +517,9 @@ message MeshConfig {
       // REQUIRED. Specifies the port of the service.
       uint32 port = 2;
 
+      // The maximum duration that the proxy will wait for a response from the provider (default timeout: 600s).
```
nrjpoddar, line 520: Define what happens when timeout fires.

nrjpoddar:
@yangminzhu we need to look at the string match API here as we are creating a divergence from other Networking APIs.

mesh/v1alpha1/config.proto
```diff
@@ -593,12 +598,13 @@ message MeshConfig {
       // REQUIRED. Specifies the port of the service.
       uint32 port = 2;
 
-      // The timeout for the gRPC request. This is the timeout for a specific request (default timeout: 600s).
+      // The maximum duration that the proxy will wait for a response from the provider, this is the timeout for a specific request (default timeout: 600s).
+      // On timeout, the communication with the authorization service is considered failed and the client connection will be closed.
```
nrjpoddar, line 602: I meant what does the user get as response in this case? 4xx or 2xx?

yangminzhu: I was just still updating the PR, it is the same as connection to the external authz service has failed (`fail_open`). For HTTP it's 403, for TCP the connection is just closed.

mesh/v1alpha1/config.proto
yangminzhu: Yes, I'm grouping the similar fields together. I added a comment for the next available ID for improving the maintenance issue.

mesh/v1alpha1/config.proto
yangminzhu: I'm trying to align with the AuthorizationPolicy API as this is currently only used with AuthZ API instead of VirtualService. In Authz we are using the `*` for the prefix/suffix matching.

yangminzhu:
> @yangminzhu we need to look at the string match API here as we are creating a divergence from other Networking APIs.

@nrjpoddar I was trying to align with the AuthZ API but I do not have a strong preference here. I think either one does the job and it is probably more convenient to be consistent with the AuthZ as this provider is only used with AuthZ API.

nrjpoddar:
> > @yangminzhu we need to look at the string match API here as we are creating a divergence from other Networking APIs.
> 
> @nrjpoddar I was trying to align with the AuthZ API but I do not have a strong preference here. I think either one does the job and it is probably more convenient to be consistent with the AuthZ as this provider is only used with AuthZ API.

Yeah, I see it now. Can we link to https://istio.io/latest/docs/reference/config/security/authorization-policy/#Rule so that users know that API semantics are same between this and AuthZ policies?

mesh/v1alpha1/config.proto
```diff
@@ -485,6 +485,26 @@ message MeshConfig {
...
+      // Sets the maximum size of a message body that the ext-authz filter will hold in memory.
+      // If max_request_bytes is reached, and allow_partial_message is false, Envoy will return a 413 (Payload Too Large).
+      // Otherwise the request will be sent to the provider with a partial message.
+      // Note that this setting will have precedence over the fail_open field.
+      uint32 max_request_bytes = 1;
+
+      // When this field is true, ext-authz filter will buffer the message until max_request_bytes is reached.
+      // The authorization request will be dispatched and no 413 HTTP error will be returned by the filter.
+      // A "x-envoy-auth-partial-body: false|true" metadata header will be added to the authorization request message
+      // indicating if the body data is partial.
```
nrjpoddar, line 498: I'm a bit confused on this behavior. So if `allow_partial_message` is set to true, ext AuthZ will act on partial body and allow/deny the request? What happens more more data arrives for the body?

Can this behavior lead to escapes for malicious clients? I understand that this is to prevent DOS attack vector.

mesh/v1alpha1/config.proto
```diff
@@ -485,6 +485,26 @@ message MeshConfig {
...
+      bool allow_partial_message = 2;
+
+      // If true, the body sent to the external authorization service in the gRPC authorization request is set with raw bytes
+      // in the raw_body field (https://github.com/envoyproxy/envoy/blame/cffb095d59d7935abda12b9509bcd136808367bb/api/envoy/service/auth/v3/attribute_context.proto#L153).
+      // Otherwise, it will be filled with UTF-8 string in the body field (https://github.com/envoyproxy/envoy/blame/cffb095d59d7935abda12b9509bcd136808367bb/api/envoy/service/auth/v3/attribute_context.proto#L147).
+      // This field only works with the envoy_ext_authz_grpc provider and has no effect for the envoy_ext_authz_http provider.
```
nrjpoddar, line 504: What's the benefit of setting this field i.e. why/when should a user set this boolean?

mesh/v1alpha1/config.proto
```diff
@@ -497,6 +517,11 @@ message MeshConfig {
       // REQUIRED. Specifies the port of the service.
       uint32 port = 2;
 
+      // The maximum duration that the proxy will wait for a response from the provider (default timeout: 600s).
+      // On timeout, the communication with the authorization service is considered failed, see fail_open for the response
+      // in this case.
```
nrjpoddar, line 522:
```suggestion
      // When this timeout condition is met, the proxy marks the communication to the authorization service as failure. In this situation, the response sent back to the client will depend on the configured `fail_open` policy field. 
```

mesh/v1alpha1/config.proto
```diff
@@ -514,13 +539,33 @@ message MeshConfig {
       // List of headers that should be included in the authorization request sent to the authorization service.
       // Note that in addition to the headers supplied by users:
       // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
-      // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      // 2. *Content-Length* will be set to 0 and the request will not have a message body. However, the authorization
```
nrjpoddar, line 542: // List of headers that should be included in the authorization request sent to the authorization service.
// Note that in addition to the headers supplied by users following headers are included by default:

mesh/v1alpha1/config.proto
```diff
@@ -514,13 +539,33 @@ message MeshConfig {
...
+      // - Suffix match: "*abc" will match on value "abc" and "xabc".
       repeated string include_headers_in_check = 6;
 
+      // If set, the client request body will be included in the authorization request sent to the authorization service.
+      EnvoyExternalAuthorizationRequestBody include_request_body_in_check = 10;
+
+      // Set of additional headers that will be included in the authorization request sent to the authorization service.
+      // Key is the header name and value is the header value.
+      // Note that client request of the same key or headers specified in include_headers_in_check will be overridden.
+      map<string, string> additional_headers_in_check = 11;
```
nrjpoddar, line 558: This field and `include_headers_in_check` look very similar. This is a synthetic and override header use case where as `include_headers_in_check` only takes headers from incoming request. Can we make the naming more clear?

`include_headers_in_check` -> `include_request_headers_in_check` 
`additional_headers_in_check` -> `add_or override_headers_in_check`

mesh/v1alpha1/config.proto
```diff
@@ -514,13 +539,33 @@ message MeshConfig {
...
       // 1. *Host*, *Method*, *Path* and *Content-Length* are automatically sent.
-      // 2. *Content-Length* will be set to 0 and the request will not have a message body.
+      // 2. *Content-Length* will be set to 0 and the request will not have a message body. However, the authorization
+      // request can include the buffered client request body (controlled by include_request_body_in_check setting),
+      // consequently the value of Content-Length of the authorization request reflects the size of its payload size.
+      //
+      // Exact, prefix and suffix matches are supported:
+      // - Exact match: "abc" will match on value "abc".
+      // - Prefix match: "abc*" will match on value "abc" and "abcd".
+      // - Suffix match: "*abc" will match on value "abc" and "xabc".
```
nrjpoddar, line 549: What's the behavior in the following scenarios:

* If the headers are not present in the incoming request?
* If I'm using a prefix match and multiple headers match what's the behavior?
* If incoming request has multiple headers of the same key, what's passed to the AuthZ server?

mesh/v1alpha1/config.proto
nrjpoddar: Since `include_headers_in_check` is already in API we might have less flexibility to change it now.

mesh/v1alpha1/config.proto
yangminzhu: it's needed when gRPC requests have invalid UTF-8 data (not compatible with the `body` field), see more details in https://github.com/envoyproxy/envoy/issues/9822.

mesh/v1alpha1/config.proto
yangminzhu: I can add an extra `include_request_headers_in_check` and deprecate the old one, this should be fine as the API is still in experimental/alpha stage. (we can check in validation and warn on the old field, and prefer to use the new field if both are set)

Regarding the new field, I originally added it as `add_headers_in_check` but @smawson does not like it and suggested the new name. Maybe we can change it to `include_additional_headers_in_check`?

mesh/v1alpha1/config.proto
yangminzhu: This only controls the check request to the external authorization service. Data are buffered until it reaches the `max_request_bytes` limit, in other words, at most `max_request_bytes` are sent to the external authorization service.

The client request will not hit the backend unless the external authorization service allows it, the backend is not affected by this flag and will see all data when allowed. When denied the backend saw nothing.

yangminzhu:
@nrjpoddar @howardjohn @smawson I have updated the PR to address comments so far. PTAL and let me know if you still have more comments, thanks.

smawson:
My concerns were met so deferring to @howardjohn and @nrjpoddar

howardjohn (approved):

nrjpoddar:
@yangminzhu I missed submitting my review yesterday. Updates look good. I have 2 minor comments:

1) For one field you have a comment saying: “Note that this setting will have precedence over the fail_open field.” but don’t explain the relationship?
2) Secondly, yesterday when I had looked your comment said that there was a bug for duplicated header keys. I think we need to document the expected behavior instead of expecting users to go through Envoy docs and PRs. If indeed, there’s a discrepancy for duplicated headers it can lead to CVEs so we should be careful about merging this without understanding what’s happening.

nrjpoddar:
I’m adding do-not-merge until I understand the duplicate headers behavior.

nrjpoddar:
Talked offline with @yangminzhu. Removing the don’t merge label. Thanks for incorporating the suggestion!

nrjpoddar (approved):

PR #1928, opened 2021-03-26
Add info about `portLevelMtls` requirement of workload selector with PeerAuthentication

rytswd:
Adding validation would be better, but in either case, the documentation should make that requirement clear.

Ref:
https://istio.slack.com/archives/C37A4KAAD/p1615592450286300

### DIFF
--- a/security/v1beta1/peer_authentication.proto
+++ b/security/v1beta1/peer_authentication.proto
@@ -154,6 +154,7 @@ message PeerAuthentication {
   // Mutual TLS settings for workload. If not defined, inherit from parent.
   MutualTLS mtls = 2;
 
-  // Port specific mutual TLS settings.
+  // Port specific mutual TLS settings. This only works when workload selector
+  // is specified.
   map<uint32, MutualTLS> port_level_mtls = 3;
 }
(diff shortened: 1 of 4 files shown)


security/v1beta1/peer_authentication.proto
```diff
@@ -154,6 +154,7 @@ message PeerAuthentication {
   // Mutual TLS settings for workload. If not defined, inherit from parent.
   MutualTLS mtls = 2;
 
-  // Port specific mutual TLS settings.
+  // Port specific mutual TLS settings. This only works when workload selector
```
smawson, line 157:
```suggestion
  // Port specific mutual TLS settings. These only apply when a workload selector
```

howardjohn (approved):
LGTM. You mentioned we should add validation, but looks like we do have it?? Does it not work?
```go
		if emptySelector && len(in.PortLevelMtls) != 0 {
			errs = appendErrors(errs,
				fmt.Errorf("mesh/namespace peer authentication cannot have port level mTLS"))
		}
```

smawson (approved):

rytswd:
> LGTM. You mentioned we should add validation, but looks like we do have it?? Does it not work?

Ah sorry I missed that, just confirmed that we are well covered with `istioctl validate`. I was hoping validating webhook catches this sort of error - I think I was testing with v1.8.x, so it may have been that I was testing with an old version.

PR #1933, opened 2021-03-31
add dry-run annotation

yangminzhu:
Allows customers to dry-run an authorization policy to test the effect using real traffic without enforcing the policy, reducing the risk of creating or changing the authorization policy. 

Only supports AuthorizationPolicy for now, the admission controller will reject other resources with the dry-run annotation.

Design doc: https://docs.google.com/document/d/1xQdZsEgJ3Ld2qebfT3EJkg2COTtCR1TqBVojmnvI78g (approved by Security WG)

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -392,3 +392,11 @@ annotations:
     hidden: false
     resources:
       - Pod
+
+  - name: istio.io/dry-run
+    featureStatus: Alpha
+    description: Specifies whether or not the given resource is in dry-run mode.
+    deprecated: false
+    hidden: true
+    resources:
+      - AuthorizationPolicy
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -28,6 +28,7 @@ type ResourceTypes int
 const (
 	Unknown ResourceTypes = iota
     Any
+    AuthorizationPolicy
     Ingress
     Pod
     Service
@@ -38,10 +39,12 @@ func (r ResourceTypes) String() string {
 	case 1:
 		return "Any"
 	case 2:
-		return "Ingress"
+		return "AuthorizationPolicy"
 	case 3:
-		return "Pod"
+		return "Ingress"
 	case 4:
+		return "Pod"
+	case 5:
 		return "Service"
 	}
 	return "Unknown"
@@ -155,6 +158,18 @@ var (
 		Deprecated:    false,
 		Resources: []ResourceTypes{
 			Any,
+		},
+	}
+
+	IoIstioDryRun = Instance {
+		Name:          "istio.io/dry-run",
+		Description:   "Specifies whether or not the given resource is in dry-run "+
+                        "mode.",
+		FeatureStatus: Alpha,
+		Hidden:        true,
+		Deprecated:    false,
+		Resources: []ResourceTypes{
+			AuthorizationPolicy,
 		},
 	}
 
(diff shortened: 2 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -392,3 +392,11 @@ annotations:
     hidden: false
     resources:
       - Pod
+
+  - name: istio.io/dry-run
```
howardjohn, line 396: copying my comment from https://github.com/istio/istio/pull/31569#discussion_r605163665

yangminzhu: > After reading the doc, I am not sure why an annotation rather than a field in authz policy is prefered. It seems the primary benefit is that we can avoid full API review, which I don't think we should do anyways.?
> The other benefit listed is about using it for other Istio API types. It seems unlikely the any other API can support this in the near future, but even if it did I would expect it to follow the same pattern as exportTo - a field shared by many types, but defined in each API explicitly.

it's never to avoid full API review, the API change requirements are the same for annotation and first-class field, they both have the feature stage.

1) we want to achieve the user experience that customers do not need to change the policy spec when switching between dry-run and enforcement mode. This is to minimize the risk of introducing unexpected changes accidentally. The annotation also gives a clear separation and boundary between features that focus on improving day-2 operation and the spec for specifying access control rules .

2) (future) the annotation supports other APIs without touching CRD definitions. Not planned for now but this turns out to be useful and needed by customers, we could easily backport the feature to older versions. Note the requirements are the same for annotation and first-class field but the implementation would be much simpler with the annotation. We can always promote the annotation to first-class field at the later stage of the feature (e.g. alpha -> beta) but not vice versa, it's hard to remove a field in the CRD.

3) It may not support other APIs in the near future but I think this is not the reason to make the API specific to one resource. The API should be in a more general form to give us more flexibility in the future. That's why it's changed from `authorization.istio.io/dry-run` to `istio.io/dry-run` as suggested by @smawson

annotation/annotations.yaml
```diff
@@ -392,3 +392,11 @@ annotations:
...
+    featureStatus: Alpha
```
yangminzhu, line 397: I wanted to set the featureStatus to experimental but the tool currently only supports alpha, beta and stable. I will look into the tool later to add the support and update the PR later when exposing this in the doc (currently this is hidden).

yangminzhu:
@istio/technical-oversight-committee Hi, could anyone from TOC take a quick look at this, the design is already approved by security WG, the implementation is blocked on this and I'm trying to get this into the 1.10 release (code cut in 1 week).

smawson (approved):
I'm comfortable adding this annotation, it seems similar in spirit to some of our other annotations/labels that are meta-controls and don't really make sense as fields (for example istio.io/rev).

howardjohn (approved):

nrjpoddar:
@yangminzhu since this annotation can be used in other resources in future. Can we describe what the annotation does for AuthZ policy. Where and how do the users see the policy dry-run output?

yangminzhu:
> @yangminzhu since this annotation can be used in other resources in future. Can we describe what the annotation does for AuthZ policy. Where and how do the users see the policy dry-run output?

yes, we will add documentation on istio.io for using the annotation. The output can be found in Envoy (log), Prometheus (metric) and Jaeger (tracking).

nrjpoddar:
That’s great. Thanks!

PR #1934, opened 2021-04-02
Use wrapper types for Tracing fields

douglas-reid:
Given the shallow-merge semantic adopted for the Telemetry API, we need to use wrapper types in order to avoid sentinel-based detection of presence (thanks proto3!). This came up in the code review of the initial implementation PR.


howardjohn (approved):

mandarjog (approved):
lgtm per follow up from review.

Needed to provide correct override/merge semantics for top level fields.

nrjpoddar (approved):
Good catch!

PR #1937, opened 2021-04-06
Add topologySpreadConstraints field to operator k8s opts

Niksko:
This adds the topologySpreadConstraints field to the k8s section of the
operator CRD.

This field is an alternative to common anti-affinity configurations that attempt
to configure pods to spread evenly across nodes and availability zones.
This is important for running highly available services, and is
something that people will want to do with their Istio resources.

Lots of people are going to want to make use of this feature, since EKS is now on 1.19 (when this graduated to stable)

Part of implementing https://github.com/istio/istio/issues/30719


linsun (approved):
Thanks, LGTM

PR #1939, opened 2021-04-06
Fix BoolValueForPB.GetValue allow nil

kebe7jun:
For: https://github.com/istio/api/issues/1938

### DIFF
--- a/operator/v1alpha1/common.go
+++ b/operator/v1alpha1/common.go
@@ -61,6 +61,14 @@ type BoolValueForPB struct {
 	protobuf.BoolValue
 }
 
+// GetValue returns true if self is not nil and Value is true, or false otherwise.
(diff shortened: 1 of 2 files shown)


howardjohn (changes requested):
You don't need this, protobuf.BoolValue already defines it

kebe7jun:
> You don't need this, protobuf.BoolValue already defines it

See

https://github.com/istio/api/blob/a9c5294cac41c9089557ef1464f8353c5ef2b346/operator/v1alpha1/common.go#L60-L62

BoolValueForPB use struct `protobuf.BoolValue` not `*protobuf.BoolValue`, which makes nil(*BoolValueForPB).GetValue raise a nil pointer exception.

howardjohn (approved):

linsun (approved):

PR #1952, opened 2021-04-20
fix incorrect type for maxUnavailable/minAvailable of PDB.

morvencao:
resolve: https://github.com/istio/istio/issues/32307

This change is supposed to backward compatible because new type `IntOrString` support old type, so the old CR should still be working with new API.


hzxuzhonghu (approved):

howardjohn (approved):

ostromart (approved):

linsun (approved):
Thanks @morvencao !

PR #1956, opened 2021-04-21
Document annotations used for injection

howardjohn:
These are not new annotations, but backfilling ones used in injection
that were not previously in the API repo.

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -408,3 +408,20 @@ annotations:
     hidden: true
     resources:
       - AuthorizationPolicy
+
+  - name: proxy.istio.io/overrides
+    featureStatus: Alpha
+    description: Used internally to indicate user-specified overrides in the proxy container of the pod during injection.
+    deprecated: false
+    hidden: true
+    resources:
+      - Pod
+
+  - name: inject.istio.io/templates
+    featureStatus: Alpha
+    description: The name of the inject template(s) to use, as a comma separate list. See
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -408,3 +408,20 @@ annotations:
     hidden: true
     resources:
       - AuthorizationPolicy
+
+  - name: proxy.istio.io/overrides
```
nrjpoddar, line 412: I had no idea these existed. Do they work and are they tested?

howardjohn: Yes it works and is tested. Its an internal annotation, not intended to be used by users (hence why its hidden). Its injected to retain state of the state of the pod prior to injection.

howardjohn: And the other one is documented and tested: https://preliminary.istio.io/latest/docs/setup/additional-setup/sidecar-injection/#custom-templates-experimental

linsun (approved):
LGTM

annotation/annotations.yaml
```diff
@@ -408,3 +408,20 @@ annotations:
...
+    featureStatus: Alpha
+    description: Used internally to indicate user-specified overrides in the proxy container of the pod during injection.
+    deprecated: false
+    hidden: true
+    resources:
+      - Pod
+
+  - name: inject.istio.io/templates
+    featureStatus: Alpha
+    description: The name of the inject template(s) to use, as a comma separate list. See
```
Monkeyanator, line 422:
```suggestion
    description: The name of the injection template(s) to use, as a comma separate list. See
```

Monkeyanator (approved):

PR #1957, opened 2021-04-22
add http header casing option to mesh config

su225:
Part of https://github.com/istio/istio/issues/32008

Signed-off-by: su225 <suchithjn22@gmail.com>


howardjohn:
I think there are two questions we need to answer here - should this be configurable and should this be default on? This has a lot of similarities to https://istio.io/latest/blog/2021/upcoming-networking-changes/

* For new users, it seems better it all cases to preserve the case. I don't see a need to allow lowercasing it always
* For existing users, they *might* have come to rely on the lowercasing. It seems a bit odd, as most apps probably assume title casing if anything, so hopefully when they adopt Istio they fix to being case insensitive instead of assuming lowercase, but I am sure in practice it may break users

su225:
I think it is important to have this configurable. Unfortunately, something like `istioctl precheck` as mentioned in the blogpost is not applicable here as apps are blackboxes from Istio perspective. If we are turning it on by default, then it is important to notify existing users. 

Otherwise, not necessary. This way, if someone is migrating their legacy apps and rely on header casing behavior then they can turn this on. Apps which don't care, don't break anyways.

nrjpoddar:
How does this impact Authorization and VirtualService routing policies? Will it break existing user config if casing of header keys is preserved?

su225:
Apparently, envoy can route if casing is not correct. Here is a small experiment I did to test that.

docker-compose.yaml
```yaml
version: '3'
services:
  envoy:
    image: envoyproxy/envoy-dev:ae780e21824bcb947299f5a048797fbf3cac21ba
    ports:
      - "10000:10000"
    volumes:
      - ./envoy.yaml:/etc/envoy/envoy.yaml
  app1:
    image: gcr.io/google-samples/hello-app:1.0
    ports:
      - "8080:8080"
  app2:
    image: gcr.io/google-samples/hello-app:2.0
    ports:
      - "8090:8080"
```

envoy.yaml
```yaml
static_resources:
  listeners:
    - address:
        socket_address:
          address: 0.0.0.0
          port_value: 10000
      filter_chains:
        - filters: 
          - name: envoy.filters.http_connection_manager
            typed_config:
              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
              stat_prefix: ingress_http

              # HTTP header casing options
              http_protocol_options:
                header_key_format:
                  stateful_formatter:
                    name: preserve_case
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.http.header_formatters.preserve_case.v3.PreserveCaseFormatterConfig

              http_filters:
                - name: envoy.filters.http.router
              route_config:
                virtual_hosts:
                  - name: default
                    domains: ["*"]
                    routes:
                      - match:
                          prefix: "/"
                          headers:
                            name: X-App-Version
                            exact_match: v2
                        route: { cluster: app_v2 }
                      - match: { prefix: "/" }
                        route: { cluster: app_v1 }
  clusters:
    - name: app_v1
      type: STRICT_DNS
      connect_timeout: 1s
      
      # HTTP header casing options
      typed_extension_protocol_options:
        envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
          "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
          explicit_http_config:
            http_protocol_options:
              header_key_format:
                stateful_formatter:
                  name: preserve_case
                  typed_config:
                    "@type": type.googleapis.com/envoy.extensions.http.header_formatters.preserve_case.v3.PreserveCaseFormatterConfig
      
      load_assignment:
        cluster_name: app_v1
        endpoints:
          - lb_endpoints:
            - endpoint:
                address:
                  socket_address:
                    address: app1
                    port_value: 8080
    - name: app_v2
      type: STRICT_DNS
      connect_timeout: 1s
      
      # HTTP header casing options
      typed_extension_protocol_options:
        envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
          "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
          explicit_http_config:
            http_protocol_options:
              header_key_format:
                stateful_formatter:
                  name: preserve_case
                  typed_config:
                    "@type": type.googleapis.com/envoy.extensions.http.header_formatters.preserve_case.v3.PreserveCaseFormatterConfig
      
      load_assignment:
        cluster_name: app_v2
        endpoints:
          - lb_endpoints:
            - endpoint:
                address:
                  socket_address:
                    address: app2
                    port_value: 8080
```

At least the routing part is fine.

ramaraochavali:
I agree with @howardjohn here. I think we should preserve casing for http1.1. by default. I think we may not need an API at mesh config. If at all some customers need other behaviour - we can probably do it vin env var which can be easily removed later.

PR #1964, opened 2021-04-27
Change snake_case to camelCase for yaml documentation

t0m:
Since these variables are supposed to refer to the yaml config, I believe they should be the camel cased names, not the pb struct names.

```release-note
NONE
```


howardjohn (approved):
thanks!

hzxuzhonghu (approved):

t0m:
Hi folks, I need a hand with the release-notes-none label, as I don't think I'm able to add it myself. The PR is a couple line fix for the docs, so should not need a release note (IMO).

smawson (approved):

PR #1966, opened 2021-05-03
Clarify subjectAltNames end in wildcards

kailun-qin:
Signed-off-by: Kailun Qin <kailun.qin@intel.com>


howardjohn (changes requested):
Putting a hold to ensure this is only merged *after* https://github.com/istio/istio/pull/32582 is

PR #1968, opened 2021-05-04
Use RE2 instead of ECMAscript regex-based match

shamsher31:
To fix https://github.com/istio/istio/issues/32598

### DIFF
--- a/networking/v1alpha3/virtual_service.pb.go
+++ b/networking/v1alpha3/virtual_service.pb.go
@@ -1501,7 +1501,7 @@ type HTTPMatchRequest struct {
 	//
 	// - `prefix: "value"` for prefix-based match
 	//
-	// - `regex: "value"` for ECMAscript style regex-based match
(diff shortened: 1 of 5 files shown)


howardjohn (approved):

shamsher31:
@linsun PTAL.

shamsher31:
@nrjpoddar PTAL.

shamsher31:
cc: @linsun @nrjpoddar

nrjpoddar (commented):
What's the impact of this change on existing user resources? Is there any guidance to help them migrate from ECMA to RE2?

howardjohn:
> What's the impact of this change on existing user resources? Is there any guidance to help them migrate from ECMA to RE2?

It was always RE2 I believe. If not, it was changed long ago in 1.4 to RE2 (when they moved to safe_regex). So this is just a docs update, not a behavioral change

nrjpoddar (approved):

PR #1971, opened 2021-05-06
Update Service Entry subjectAltNames documentation.

Kmoneal:
Address https://github.com/istio/istio/issues/32539

* Specify that SE SAN is not applied with SIMPLE TLS mode.
* Remove that DR SAN overrides SE SAN.

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -943,8 +943,6 @@ message ClientTLSSettings {
   // A list of alternate names to verify the subject identity in the
   // certificate. If specified, the proxy will verify that the server
   // certificate's subject alt name matches one of the specified values.
-  // If specified, this list overrides the value of subject_alt_names
-  // from the ServiceEntry.
   repeated string subject_alt_names = 5;
 
   // SNI string to present to the server during TLS handshake.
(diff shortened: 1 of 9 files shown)


networking/v1alpha3/destination_rule.proto
```diff
@@ -943,8 +943,6 @@ message ClientTLSSettings {
   // A list of alternate names to verify the subject identity in the
   // certificate. If specified, the proxy will verify that the server
   // certificate's subject alt name matches one of the specified values.
-  // If specified, this list overrides the value of subject_alt_names
-  // from the ServiceEntry.
```
jacob-delgado, line 947: Does it do this when the TLS mode is not SIMPLE? This may be another case where SIMPLE is different than MUTUAL or ISTIO_MUTUAL.

Kmoneal:
@ramaraochavali I would appreciate any feedback or insight you have!

networking/v1alpha3/destination_rule.proto
ramaraochavali: Yes. For simple we do not add validation context https://github.com/istio/istio/pull/31391

ramaraochavali:
@nrjpoddar I completely misread the question and my answer was wrong. https://github.com/istio/istio/issues/32539 This is a bug we need to fix. I will take a look

PR #1972, opened 2021-05-06
Cherrypick fields set in x-forwarded-client-cert

arkodg:
Introduce `set_current_client_cert_details`
to configure the fields in the client certificate
that are added to the `x-forwarded-client-cert` header
similar to
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-setcurrentclientcertdetails

This cannot be achieved using a EnvoyFilter due to
https://github.com/istio/istio/issues/23178 / https://github.com/istio/istio/issues/18169

Defaults to `true` to maintain backward compatibility in istio.

Signed-off-by: Arko Dasgupta <arko@tetrate.io>


arkodg:
@googlebot I signed it!

PR #1973, opened 2021-05-07
Mesh-wide CA path

Kmoneal:
Address https://github.com/istio/istio/issues/32539

* Specify that SE SAN is not applied with SIMPLE TLS mode.
* Remove that DR SAN overrides SE SAN.


mesh/v1alpha1/proxy.proto
```diff
@@ -498,6 +498,32 @@ message ProxyConfig {
...
+  // certificate path at the client-side proxy when `SIMPLE` TLS or `MUTUAL`
+  // TLS (non `ISTIO_MUTUAL`) origination modes are used. This setting can be
+  // overridden at the host level via DestinationRule API.
+  //
+  // If `VerifyCertificateAtClient` is `false`,
+  //`DefaultCertificateAuthorityPath` will default to empty.
+  //
+  // If `VerifyCertificateAtClient` is `true`,
+  // `DefaultCertificateAuthorityPath` will default to the OS CA bundle for the proxy.
+  string default_certificate_authority_path = 36;
```
howardjohn, line 526: I don't think we should have a path. Default should always be system certificates, which is context specific for each proxy (depending on OS)

mesh/v1alpha1/proxy.proto
```diff
@@ -498,6 +498,32 @@ message ProxyConfig {
   // The plugin certificates (the 'cacerts' secret), self-signed certificates (the 'istio-ca-secret' secret)
   // are added automatically by Istiod.
   repeated string ca_certificates_pem = 34;
+
+  // `VerifyCertificateAtClient` sets the mesh global default for peer certificate validation
```
howardjohn, line 502: this is not a mesh global, its a per proxy configuration

networking/v1alpha3/destination_rule.proto
```diff
@@ -923,6 +923,10 @@ message ClientTLSSettings {
   // certificates to use in verifying a presented server certificate. If
   // omitted, the proxy will not verify the server's certificate.
   // Should be empty if mode is `ISTIO_MUTUAL`.
+  //
+  // If `CaCertificates`is set, proxy verifies CA signature based on given CaCertificates.
+  // This setting take priority over ProxyConfig `DefaultCertificateAuthorityPath` and
```
howardjohn, line 928: How can I "unset" VerifyCertificateAtClient=true to not verify?

networking/v1alpha3/destination_rule.proto
```diff
@@ -945,10 +949,28 @@ message ClientTLSSettings {
   // certificate's subject alt name matches one of the specified values.
   // If specified, this list overrides the value of subject_alt_names
   // from the ServiceEntry.
+  //
+  // If unset, and `VerifyCertificateAtClient` is true, proxy uses host in
+  // destination rule to verify the SANs. If unset, and
+  // `VerifyCertificateAtClient` is false, proxy does not verify SANs.
+  //
+  // Client-side proxy will exact match host as well as one level wildcard if
```
howardjohn, line 957: is this the current semantics today?

Kmoneal: This is brought over from `VerifyCertificateAtClient` in MeshConfigs. It seemed more suitable here

mesh/v1alpha1/proxy.proto
Kmoneal: Thanks! Didn't catch this when I was updating the doc

PR #1974, opened 2021-05-07
envoyfilter: add bootstrap patch

kyessenov:
Issue https://github.com/istio/istio/issues/26830
Ref https://github.com/istio/istio/pull/32496


networking/v1alpha3/envoy_filter.proto
```diff
@@ -370,6 +370,38 @@ import "networking/v1alpha3/sidecar.proto";
 //                   cluster_name: xds-grpc
 //           type_urls: ["envoy.extensions.filters.http.wasm.v3.Wasm"]
 // ```
+//
+// The following example adds a Wasm service extension for all proxies using a locally available Wasm file.
```
louiscryan, line 374: Could you contextualize the example a bit better? Provide some situations where this would be useful

louiscryan:
Will this be covered in the dedicated WASM API proposal?

networking/v1alpha3/envoy_filter.proto
kyessenov: Added two examples. Rate limit and a user database.

kyessenov:
@louiscryan No, Wasm singletons are not in Wasm API, since we made a decision to leave it out of scope to get some MVP first. EnvoyFilter is an escape hatch.

louiscryan:
> @louiscryan No, Wasm singletons are not in Wasm API, since we made a decision to leave it out of scope to get some MVP first. EnvoyFilter is an escape hatch.

Presumably though we would add support eventually and the escape hatch would disappear

louiscryan (approved):
LGTM

kyessenov:
@louiscryan Correct, we do not have enough evidence to warrant a concrete API for this, but it's better experience with EnvoyFilter than without it for it.

mandarjog:
And it is not just the configuration API.  In one type of usage we may be able to replace configuration driven singleton pattern with a runtime singleton.

mandarjog:
@smawson @nrjpoddar ?
We need two approvals from toc .

smawson (approved):

PR #1978, opened 2021-05-08
Add network load balancing settings

hzxuzhonghu:
Refer to this [design](https://docs.google.com/document/d/1kgrMhC4wwQbOllcycW9ICqUt6l8EoUkOWGY-QyEJTTg/edit?usp=sharing)


hzxuzhonghu:
Another option is to add one bool flag to tell whether to enable network aware lb, it looks simpler to users, but may lack extensibility.

hzxuzhonghu:
@costinm @howardjohn @nmittler ptal

networking/v1alpha3/destination_rule.proto
```diff
@@ -1051,3 +1055,47 @@ message LocalityLoadBalancerSetting{
...
+// When it is specified simultaneously with locality load balancing, the priority order is network > locality.
+// e.g. if both service A and B resides in two networks{`network1`, `network2`} and two az{`az-1`, `az-2`}
+// for service A instance in `network1/us-east`, the expected endpoints priority will be:
+// ```
+//   `network1/az-1`: 0
+//   `network1/az-2`: 1
+//   `network2/az-1`: 2
+//   `network2/az-2`: 3
+// ```
+message NetworkLoadBalancerSetting{
```
nrjpoddar, line 1082: This seems like a copy of `LocalityLoadBalancerSetting.Failover` and I think we can just that struct instead of creating a copy. Why do you need to create a parallel struct here?

hzxuzhonghu: From LocalityLoadBalancerSetting it is clearly related to locality based lb, and make network-aware lb there seems not good.

hzxuzhonghu:
in favor of #2002

PR #1979, opened 2021-05-11
Add configuration of proxy inbound listen port

luckyxiaoqiang:
Add configuration of proxy inbound listen port.

Used in: https://github.com/istio/istio/pull/32772


luckyxiaoqiang:
> can you describe the use case for customizing this?

It is used in https://github.com/istio/istio/pull/32772.

luckyxiaoqiang:
> Can you explain more about _why_ its inconvenient? Just because a field exists for outbound doesn't mean it needs to exist for inbound if there is no good use case for it.

Listen port is not recommended in `net.ipv4.ip_local_port_range` which will be automatically allocated to local port of tcp connection by system. In some scenarios,  15006 may be in range of `net.ipv4.ip_local_port_range` (eg. 10000~65535), so it is highly recommended to support configuration of listen port.

hzxuzhonghu:
Make sense to me, it is similar to allow setting outbound listener port.

PR #1982, opened 2021-05-11
Security fix on URL normalization (GoldPanda).

myidpt:


howardjohn (approved):

jacob-delgado (approved):

mesh/v1alpha1/config.proto
```diff
@@ -813,8 +813,37 @@ message MeshConfig {
...
+
+      // In addition to normalization in `MERGE_SLASHES`, slash characters are UTF-8 decoded (case insensitive) prior to merging.
+      // This means `%2F`, `%2f`, `%5C`, and `%5c` sequences in the request path will be rewritten to `/` or `\`.
+      // For example, `/a%2f/b` normalizes to `a/b`.
+      DECODE_AND_MERGE_SLASHES = 4;
+    }
+    NormalizationType normalization = 1;
+  }
+
+  // ProxyPathNormalization configures how URL paths in incoming and outgoing HTTP requests are normalized by the proxy.
```
linsun, line 841: Does this apply to gateway as well?  I assume yes.  Would be good to add it as proxy is commonly interpreted as sidecar proxy in Istio.

releasenotes/notes/url-normalization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+releaseNotes:
+  - |
+    **Added** workload request URL path normalization configurations
```
linsun, line 6: add for sidecar proxy and gateways?  
```suggestion
    **Added** workload request URL path normalization configurations for sidecar proxy and gateways
```

mesh/v1alpha1/config.proto
linsun: Also, can you clarify default value in the comment as well?

myidpt: Clarified. Should be clear now.

nrjpoddar (approved):

PR #1989, opened 2021-05-13
add local origin error configuration to destination rules

ramaraochavali:
For https://github.com/istio/istio/issues/32645


ramaraochavali:
@nrjpoddar @howardjohn @hzxuzhonghu Can you PTAL?

releasenotes/notes/32645.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: release-notes/v2
+kind: feature
+area: networking
```
brian-avery, line 3:
```suggestion
area: traffic-management
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -721,6 +721,19 @@ message OutlierDetection {
   // $hide_from_docs
   int32 consecutive_errors = 1 [deprecated=true];
 
+  // Determines whether to distinguish local origin failures from external errors. If set to true
+  // consecutive_local_origin_failure is taken into account for outlier detection calculations. 
+  // This should be used when we want to derive the outlier detection status based on the errors 
+  // seen locally such as failure to connect, timeout while connecting etc. rather than the status code
```
nrjpoddar, line 727:
```suggestion
  // This should be used when you want to derive the outlier detection status based on the errors 
  // seen locally such as failure to connect, timeout while connecting etc. rather than the status code
```

networking/v1beta1/destination_rule.proto
```diff
@@ -721,6 +721,19 @@ message OutlierDetection {
   // $hide_from_docs
   int32 consecutive_errors = 1 [deprecated=true];
 
+  // Determines whether to distinguish local origin failures from external errors. If set to true
+  // consecutive_local_origin_failure is taken into account for outlier detection calculations. 
+  // This should be used when we want to derive the outlier detection status based on the errors
```
nrjpoddar, line 726:
```suggestion
  // This should be used when you want to derive the outlier detection status based on the errors 
```

nrjpoddar (approved):

nrjpoddar (commented):
@ramaraochavali it might be helpful to add example use case like you mentioned in the issue. At first glance it's not obvious why a user should enable this.

ramaraochavali:
@nrjpoddar Thanks. Will add usecases

ramaraochavali:

networking/v1alpha3/destination_rule.proto
```diff
@@ -721,6 +721,21 @@ message OutlierDetection {
...
 
+  // Determines whether to distinguish local origin failures from external errors. If set to true
+  // consecutive_local_origin_failure is taken into account for outlier detection calculations. 
+  // This should be used when you want to derive the outlier detection status based on the errors 
+  // seen locally such as failure to connect, timeout while connecting etc. rather than the status code
+  // retuned by upstream service. This is especially useful when the upstream service explicitly returns
+  // a 5xx for some requests and you want to ignore those responses from upstream service while determining
+  // the outlier detection status of a host.
+  // Defaults to false.
+  bool split_external_local_origin_errors = 8;
```
howardjohn, line 732: Why do we need this field? is presence of `consecutive_local_origin_failure` not enough to determine this?

ramaraochavali: consecutive_local_origin_failure - right now it can probably. But it helps if we add more local origin detection mechanisms in future - so that we do not have to check for the presence of multiple fields.

hzxuzhonghu (approved):

networking/v1alpha3/destination_rule.proto
ramaraochavali: I think it is slightly different because this is doing some thing special compared to regular outlier detection so user explicitly setting it to true to indicate only local_xxx related detections will be enforced and rest of them are ignored. It is used to classify outlier detectors and determines which type of detections should be enforced.  So if we introduce another local outlier detector setting that flag to true means both will be honoured (with defaults also). 

The advantage for user in this case is with this boolean he choose set of detectors either local only or the regular irrespective what other values are. 

But if you think we should not add this bool for now - but if users gives consecutive_local_origin_failure - we ignore others - I am fine with that though I think explicit choice may be better here.

PR #1998, opened 2021-05-19
Document how to configure proxy config

howardjohn:
How it looks rendered:

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -308,6 +308,11 @@ message Topology {
 //   proxy.istio.io/config: |
 //     discoveryAddress: istiod:15012
 // ```
+//
+// If both are configured, the two are merged with per field semantics; the field set in annotation will fully replace the field from mesh config defaults.
+// This is different than a deep merge provided by protobuf.
+// For example, `"tracing": { "sampling": 5 }` would completely override a setting configuring a tracing provider
+// such as `"tracing": { "zipkin": { "address": "..." } }`.
 message ProxyConfig {
   // Path to the generated configuration file directory.
   // Proxy agent generates the actual configuration and stores it in this directory.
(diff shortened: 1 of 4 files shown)


louiscryan (commented):
LGTM

louiscryan (approved):

kyessenov:
1. Why is file and environment variable method not mentioned? Is the plan to deprecate those sideloaded APIs?
2. Need explanation of how merging is done with both mesh and workload defined together.

howardjohn:
> 1. Why is file and environment variable method not mentioned? Is the plan to deprecate those sideloaded APIs?
These are really, IMO, internal implementation details. There is no reason I know of that a user should ever touch them.

> 2. Need explanation of how merging is done with both mesh and workload defined together.
Agreed

kyessenov:
1. That's fine, but we should work towards removing those features. It's weird to have a hidden way to bypass the API semantics defined here. It would be equally strange if some file overrides a virtual service.

howardjohn:
>  It would be equally strange if some file overrides a virtual service.

You can configure this :slightly_smiling_face:

mesh/v1alpha1/proxy.proto
```diff
@@ -291,7 +291,26 @@ message Topology {
...
+// This can also be configured on a per-workload basis by configuring the `proxy.istio.io/config` annotation on the pod. For example:
+//
+// ```
+// annotations:
+//   proxy.istio.io/config: |
+//     discoveryAddress: istiod:15012
+// ```
+//
+// If both are configured, the two are merged with per field semantics; and field set will fully replace that field.
+// This is different than a deep merge provided by protobuf or similar.
```
mandarjog, line 313: Clarify: annotation fields replace the defaultProxyConfig

brian-avery:
This is awesome. Thanks @howardjohn. Spent too long trying to figure this out when working on https://istio.io/latest/docs/ops/configuration/telemetry/envoy-stats/

brian-avery (approved):

nrjpoddar (commented):
@howardjohn should we mention that users need to restart workloads if they change at the mesh level?

PR #1999, opened 2021-05-20
all: regenerate using tools with cuelang.org/go@v0.4.0

myitcv:
This change is the result of regenerating using the version of
cuelang.org/go implied by istio/tools#1539.

### DIFF
--- a/analysis/v1alpha1/message.gen.json
+++ b/analysis/v1alpha1/message.gen.json
@@ -18,24 +18,7 @@
           },
           "documentationUrl": {
             "description": "A url pointing to the Istio documentation for this specific error type. Should be of the form `^http(s)?://(preliminary\\.)?istio.io/docs/reference/config/analysis/` Required.",
-            "type": "string",
-            "format": "string"
-          }
-        }
-      },
-      "istio.analysis.v1alpha1.AnalysisMessageBase.Type": {
-        "description": "A unique identifier for the type of message. Name is intended to be human-readable, code is intended to be machine readable. There should be a one-to-one mapping between name and code. (i.e. do not re-use names or codes between message types.)",
-        "type": "object",
-        "properties": {
-          "name": {
-            "description": "A human-readable name for the message type. e.g. \"InternalError\", \"PodMissingProxy\". This should be the same for all messages of the same type. Required.",
-            "type": "string",
-            "format": "string"
-          },
-          "code": {
-            "description": "A 7 character code matching `^IST[0-9]{4}$` intended to uniquely identify the message type. (e.g. \"IST0001\" is mapped to the \"InternalError\" message type.) 0000-0100 are reserved. Required.",
-            "type": "string",
-            "format": "string"
+            "type": "string"
           }
         }
       },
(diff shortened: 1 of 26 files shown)


myitcv:
Firstly, my apologies, I'm not entirely clear on the sequence of how you merge changes to generated files in this repository, given that I understand the generated contents are a function of a Docker image built from https://github.com/istio/tools:

https://github.com/istio/api/blob/6f705c42b0340db2c63edd420d07d99f6ef18a40/common/scripts/setup_env.sh#L66

However, I created this PR in any case as a place to discuss the diff that results from https://github.com/istio/tools/pull/1539, because it requires some commentary.

The diff in this PR looks incredibly scary, but normalised (by unmarshaling and then re-marshaling the JSON) the differences look clearer:

https://gist.github.com/myitcv/b3e889140bca9b1848b42c18cb0f7bc4

The diffs breaks down into the following categories:

* `"format": "string",` on a `"type": "string"` field is superfluous, hence this is elided in later CUE versions
* The `""description"` field for schemas now consistently and correctly reflects the doc comment for the proto type, not a field of that type. For example, currently the documentation for the `google.rpc.Status` schema is as follows (in one place):

https://github.com/istio/api/blob/6f705c42b0340db2c63edd420d07d99f6ef18a40/mcp/v1alpha1/istio.mcp.v1alpha1.gen.json#L278-L279

However, you can see this documentation comes from the field doc comment:

https://github.com/istio/api/blob/6f705c42b0340db2c63edd420d07d99f6ef18a40/mcp/v1alpha1/mcp.proto#L131-L134

Whereas it should come from:

https://github.com/istio/api/blob/6f705c42b0340db2c63edd420d07d99f6ef18a40/common-protos/google/rpc/status.proto#L27-L80

This results in some schema descriptions apparently being dropped where the schema has no doc comment, but a field of that type did have a doc comment.

* The `"format"` value for a timestamp field is now correctly set as `"date-time"` and not `"dateTime"`, per https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times
* The definition of the `istio.networking.v1alpha3.ReadinessProbe` schema within the generated file `mesh/v1alpha1/istio.mesh.v1alpha1.gen.json` is currently incorrect, and inconsistent with `networking/v1alpha3/workload_group.gen.json`. This change corrects the former, such that both instances are consistent.

cc @mpvl

howardjohn:
Regarding the merge order - unfortunately we don't have a great mechanism for this. The best we can do is, once we agree this should be merged, merge the tools one and then when the image is updated here the PR can merge. Other PRs may be broken in the window, but it should be only a few minutes

howardjohn:
It looks like only one part of the cue gen was ran here, I am having problems with the other:
```
$ cue-gen -paths=common-protos -f=./cue.yaml --crd=true -snake=jwksUri,apiKeys,apiSpecs,includedPaths,jwtHeaders,triggerRules,excludedPaths,mirrorPercent
Building OpenAPIs for istio.io/api/networking/v1beta1...
DestinationRule.trafficPolicy.portLevelSettings.*.loadBalancer.consistentHash: openapi: #LoadBalancerSettings.consistentHash: undefined field #ConsistentHashLB
genoapi.go:432: Error generating OpenAPI schema
```

(from gen.sh)

myitcv:
Thanks, @howardjohn. Let me know if you have any questions/concerns regarding this PR, https://github.com/istio/tools/pull/1539 or  the analysis in https://github.com/istio/api/pull/1999#issuecomment-845079684. 

> It looks like only one part of the cue gen was ran here, I am having problems with the other:

Ah, good catch, thank you. I tried to actually build the Docker image for tools to do a proper check but my hard disk ran out of space 😄 

I'll fix that up now.

howardjohn:
> Ah, good catch, thank you. I tried to actually build the Docker image for tools to do a proper check but my hard disk ran out of space

I believe the only things using cue are the cue command you ran and the one I mentioned above, so you should be able to test it locally with just the binary.

> Let me know if you have any questions/concerns regarding this PR

It LGTM so far, but I want to analyze the output CRDs to make sure its backwards compatible for Kubernetes. Mainly the readiness probe thing

howardjohn (approved):
Put a hold on since we need tools in first

myitcv:
This PR is now ready from my perspective, per https://github.com/istio/tools/pull/1539#issuecomment-847924890

howardjohn (approved):

myitcv:
This PR is now obsolete; an automated PR that picks up the tool dependency change in https://github.com/istio/tools/pull/1539 has already been merged in aadf1ba23.

PR #2002, opened 2021-05-24
Added topology aware loadbalancing

hzxuzhonghu:
Refer to this [design](https://docs.google.com/document/d/1kgrMhC4wwQbOllcycW9ICqUt6l8EoUkOWGY-QyEJTTg/edit?usp=sharing)

To provide a more general way to replace #1978,   support "NETWORK", "REGION", "ZONE", "SUBZONE" at first, and it can be extended to support more topology keys.


hzxuzhonghu:
@howardjohn Hope this align with the network WG meeting agreements

mesh/v1alpha1/config.proto
```diff
@@ -283,6 +283,22 @@ message MeshConfig {
...
+  //   3. endpoints in same ["NETWORK"] but different REGION with the client proxy, the priority is 2.
+  //   4. all other endpoints, the priority is 3.
+  // ```yaml
+  // topologyKeys:
+  // - keys: ["NETWORK", "REGION", "ZONE"]
+  // - keys: ["NETWORK", "REGION"]
+  // - keys: ["NETWORK"]
+  // ```
+  // Note: topologyKeys can not be specified simultaneously with localityLoadBalancerSetting
+  repeated istio.networking.v1alpha3.TopologyKeys topologyKeys = 62;
```
howardjohn, line 300: We already have a way to group endpoints - subsets, which allows grouping by networking/region/zone already. I wonder if there is some way we can align the two concepts.

I could imagine something like 
```
  // topologyKeys:
  // - subsets: [my-network]
  // - subsets: [my-region]
```

But I do have some concerns that it is hard to make it a global setting, we don't have a way to say "same namespace", and it may be overly verbose

mesh/v1alpha1/config.proto
```diff
@@ -283,6 +283,22 @@ message MeshConfig {
...
+  // The below config specifies the priority levels:
+  //   1. for endpoints in same ["NETWORK", "REGION", "ZONE"] with the client proxy, the priority is 0.
+  //   2. endpoints in same ["NETWORK", "REGION"] but different ZONE with the client proxy, the priority is 1.
+  //   3. endpoints in same ["NETWORK"] but different REGION with the client proxy, the priority is 2.
+  //   4. all other endpoints, the priority is 3.
+  // ```yaml
+  // topologyKeys:
+  // - keys: ["NETWORK", "REGION", "ZONE"]
+  // - keys: ["NETWORK", "REGION"]
+  // - keys: ["NETWORK"]
```
howardjohn, line 297: I think we also want "cluster" and "node"?

hzxuzhonghu: I donot want to make it complex to implement.  We have no node info currently. 
But of course, these can be done as a second step

mesh/v1alpha1/config.proto
hzxuzhonghu: >We already have a way to group endpoints - subsets, which allows grouping by networking/region/zone already. I wonder if there is some way we can align the two concepts.

Can you elaborate a little bit more?

howardjohn: I mean - what we are adding is basically a way to say "first use this set of endpoints, then this other set, then this set".

So basically we need a way to define a set of endpoints, then make an ordered list of the sets.

Subset is already a set of endpoints

mandarjog: What does “same namespace” mean in this context ?
I don’t think it means anything. I think Normal locality constraints 
Host/node, rack, dc, zone, region make more sense to me from a latency perspective.
You can add to it single hop and two hops — cluster .
we should provide guidance on the order.
Comparing two regional clusters with presence in same zones. We want it to prefer cluster ahead of zone so as to avoid 2 hops.

hzxuzhonghu: @mandarjog Totally agree with `Comparing two regional clusters with presence in same zones. We want it to prefer cluster ahead of zone so as to avoid 2 hops.`, this is the intention i wanted to do from beginning.

@howardjohn the subsets we have now is static endpoints, they can not vary according to the calling proxy‘s metadata. So it can not solve the case we face.

mesh/v1alpha1/config.proto
ramaraochavali: Couple of nits:
- I think we do not have to document the priority number here - We can just say network, region and zone will have highest priority etc...
- The structure to express priority seems very cumbersome with keys repeating multiple times. Do we want to support more than 3 priority levels or is that static?

hzxuzhonghu: We have to combine different dimension attributes(network and locality here). 

> Do we want to support more than 3 priority levels or is that static?

It depends on users, we just provide this schema to allow them setting any number of priority levels

linsun: hey @hzxuzhonghu could you provide some examples/scenarios on how to use this?  I'm having a hard time to understand how this is going to be configured by users. I think your priority levels make sense but don't understand what are the extra things users would need to configure or customize.

hzxuzhonghu: @linsun The valid keys include `"NETWORK", "REGION", "ZONE", "SUBZONE"` .These are the metadata both proxy and endpoint have, acquired automatically.  The ideal status, user only need to configure outlier detection which is required by envoy.

ramaraochavali: These are labels on Pod right? Can we use the labels directly instead of coming up with our convention here ?

hzxuzhonghu: The labels can be on pods (`istio-locality`) or from nodes, like `failure-domain.beta.kubernetes.io/zone` `failure-domain.beta.kubernetes.io/region`  `topology.kubernetes.io/zone` `topology.kubernetes.io/region` or even from `topology.istio.io/network`

hzxuzhonghu: Added comments

networking/v1alpha3/destination_rule.proto
```diff
@@ -1066,3 +1083,26 @@ message LocalityLoadBalancerSetting{
...
+    // it can be acquired from kubernetes node label `topology.kubernetes.io/region` or the deprecated label `failure-domain.beta.kubernetes.io/region`.
+    REGION= 1;
+    // ZONE is used to match the region metadata of the endpoint,
+    // it can be acquired from kubernetes node label `topology.kubernetes.io/zone` or the deprecated label `failure-domain.beta.kubernetes.io/zone`.
+    ZONE= 2;
+    // SUBZONE is used to match the subzone metadata of the endpoint,
+    // it can be acquired from k8s node label `topology.istio.io/subzone`.
+    // Note: the label is not applied by upstream kubernetes.
+    SUBZONE= 3;
+  }
```
hzxuzhonghu, line 1104: @linsun Added comments about the topology metadata from, ptal

rcernich:
Is this a WIP?  The related design has not been approved by the working group.  Hopefully, the status will be updated in the next WG call.

mesh/v1alpha1/config.proto
hzxuzhonghu: I donot remeber we record the node info of the proxy, let me double check

hzxuzhonghu: I am sure there is no node name there.

hzxuzhonghu:
@nrjpoddar and @rcernich commented, and it looks quite confusing for me, not to say normal users.

hzxuzhonghu:
@istio/technical-oversight-committee Would you give some comments? The most urgent use case is to support [cluster local loadbalancing ](https://istio.slack.com/archives/C01AT2S8B7B/p1622470081033400)

networking/v1alpha3/destination_rule.proto
```diff
@@ -1066,3 +1083,26 @@ message LocalityLoadBalancerSetting{
...
+    NETWORK= 0;
+    // REGION is used to match the region metadata of the endpoint,
+    // it can be acquired from kubernetes node label `topology.kubernetes.io/region` or the deprecated label `failure-domain.beta.kubernetes.io/region`.
+    REGION= 1;
+    // ZONE is used to match the region metadata of the endpoint,
+    // it can be acquired from kubernetes node label `topology.kubernetes.io/zone` or the deprecated label `failure-domain.beta.kubernetes.io/zone`.
+    ZONE= 2;
+    // SUBZONE is used to match the subzone metadata of the endpoint,
+    // it can be acquired from k8s node label `topology.istio.io/subzone`.
+    // Note: the label is not applied by upstream kubernetes.
```
nrjpoddar, line 1102:
```suggestion
    // Note that this label is not added by default by most Kubernetes distributions and needs to be manually added.
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1066,3 +1083,26 @@ message LocalityLoadBalancerSetting{
...
+    SUBZONE= 3;
+  }
+  // keys contains the topology key that used to select endpoints
```
nrjpoddar, line 1105:
```suggestion
  // Keys contains the list of topology keys used to select endpoints for prioritized load balancing.
```

nrjpoddar (commented):
I think this API is quite simple to understand so I am going to approve it after my comments are addressed.

rcernich:
I'm just wondering if this has been approved by the networking wg.  The associated RFC is marked as WIP.  Does it require approval by the WG before being accepted?  What's the process?

hzxuzhonghu:
@ramaraochavali @stevenctl WDYT

networking/v1alpha3/destination_rule.proto
```diff
@@ -1066,3 +1083,26 @@ message LocalityLoadBalancerSetting{
     // e.g. true means that turn on locality load balancing for this DestinationRule no matter what mesh wide settings is.
     google.protobuf.BoolValue enabled = 3;
 }
+
+// TopologyKeys is the topology key list that used to select a group of endpoints.
+message TopologyKeys {
+  // Supported topology key
+  enum TopologyKey {
+    // NETWORK is used to match the network metadata of the endpoint,
+    // it can be acquired from istio label `topology.istio.io/network`.
```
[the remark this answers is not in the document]
hzxuzhonghu: label，  could also be from namespace label or `meshnetworks`

ramaraochavali:
Since this was agreed in networking WG to move towards generic topology aware load balancing - I am OK with high level approach for this. I will give any detailed feedback on api in a little while

howardjohn (changes requested):
I would very much like to hold off on this until we get consensus. This is going directly into a beta API and is an extremely important feature with a lot of options, so I think it's important to get right. Kubernetes had the ~same feature and just reverted it, which to me is proof we need to think really hard before commiting to this.

In the networking WG we had discussed wanting a more generic approach using arbitrary label groupings. I do still think that is worth investigating, and Rob has mentioned they are working on a proposal.

If we do *not* do a more generic approach, then this feels like the wrong API to me. How often do we think that users would actually need to customize where network falls in the hierarchy? I know it may be bigger or smaller than a region, but in practice does that happen? Should we just select a spot in the hierarchy to insert it?

nrjpoddar:
@howardjohn you beat me to it. I was also going to comment that K8s reverted the topology keys in K8s service APIs. However, I don't know the history or why it was done. Can someone expain that?

howardjohn:
https://opensource.googleblog.com/2020/11/kubernetes-efficient-multi-zone.html
discusses it a bit. But it should be noted that k8s implementation was a
lot worse than what we can do since they are only operating in kube-proxy
which has a very limited view compared to us.

hzxuzhonghu:
@howardjohn @rcernich Not to mention we have already had subset under DestinationRule. Can you give me a deadline when the proposal you are working on can be done? We could not wait for undetermined stuff indefinitely.

hzxuzhonghu:
Let me clarify the current model we have:
image

hzxuzhonghu:
@nrjpoddar @howardjohn https://groups.google.com/g/kubernetes-sig-network/c/wXd1D_fKjqU/m/EmU3mCrBBgAJ this is the discussion about service topology in k8s.

nrjpoddar:
I agree that K8s case is different and they are trying to do "auto" routing so that an API is not required. Additionally, the implementation via iptables was quite constrained which is not the case with Istio. I do think we need a generic API whether we implement it via this mechanism or use subsets, we can compare once we have both APIs in front of us.

hzxuzhonghu:
>Lastly, I think it's worthwhile to discuss normalizing topology keys.
I've always said that it's arbitrary, but in truth there's been low
demand for other keys and other systems have standardized on 2 or 3
level hierarchies. Maybe we should standardize
on 2 or 3 levels? We already define region and zone. xDS defines
sub-zone (https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-locality).

howardjohn:
I think this approach may be too generalized to be easy to use and understand (it is really hard to wrap your head around!) but also not generalized enough to be more useful than a simpler API. For example, there is no use to ever have region, zone, subzone in a different order. So all we really get here is the ability to decide where in the hierarchy network lands. Is it worth the complexity when we don't solve ~half of the use cases for prioritized load balancing (see the design doc)? I don't think it is.

I would like to wait to see Rob's proposal - and any other proposals as well! It's well worth the wait, this is a tricky but extremely important API to get right, and because it's being added to a beta API the bar should be incredibly high. So I understand your concerns about the time this is taking, but I think we need to be cautious here to make sure we do the right thing instead of rush a suboptimal API. Thanks for your work on this

hzxuzhonghu:
>For example, there is no use to ever have region, zone, subzone in a different order. 

I think you must be saying the below keys, it is not ordered sensitive, it is just a list of keys. We want to satisfy general use cases. BTW, this is like k8s `ClusterRole`, which includes many string lists.  

 
```
   topologyKeys:
   - keys: ["NETWORK", "REGION", "ZONE"]
```

```
type PolicyRule struct {

	Verbs []string

	APIGroups []string

	Resources []string

	ResourceNames []string

	NonResourceURLs []string
}

```

>Is it worth the complexity when we don't solve ~half of the use cases for prioritized load balancing (see the design doc)? I don't think it is.

I am not sure what exactly is the use case for `prioritized load balancing` you mean?

>I would like to wait to see Rob's proposal - and any other proposals as well! It's well worth the wait, this is a tricky but extremely important API to get right, and because it's being added to a beta API the bar should be incredibly high. So I understand your concerns about the time this is taking, but I think we need to be cautious here to make sure we do the right thing instead of rush a suboptimal API. Thanks for your work on this

I was almost completely puzzled at the previous design from Rob, can i say i don't agree with that one? We donot want to wait indefinitely, neither do most users

rcernich:
@howardjohn, I think @hzxuzhonghu's solution applies to lb subset config, which selects a subset of the available endpoints based on whether or not the specified labels are a match between the endpoint and the proxy.  It's a list of a set of label keys, where the first set of keys that matches is used.  This works independently of any other endpoint prioritization (i.e. it layers on top of locality load balancing).

@hzxuzhonghu, as I mentioned in my comments on your doc, I was reusing the concept of subsets to define endpoint sets that would be applied to load balancing configuration, i.e. the subset is used to select workloads based on labels, and those subsets were referenced in load balancing configuration, independent of the usual application of DestinationRule subset.  Because of the complications around using subset (is a subset a collection of workloads for load balancing, or is it intended to be use for routing, or both?), I was asked to refine the proposal.

That said, the draft of the RFC is complete and I'm waiting for any comments from our internal team before sharing it out.  It should be ready for discussion in this week's networking wg meeting.

PR #2004, opened 2021-05-25
add option for preserving header case in mesh config

ramaraochavali:


mesh/v1alpha1/config.proto
```diff
@@ -847,8 +847,14 @@ message MeshConfig {
   // If not set, the NormalizationType.DEFAULT configuration will be used.
   ProxyPathNormalization path_normalization = 61;
 
+  // Currently When handling HTTP/1.1, Istio will normalize the header keys to be all lowercase as 
+  // per the HTTP/1.1 specification. This can potentially result in issues when migrating existing systems 
+  // that might rely on specific header casing. Set this true to preserve header casing.
+  // Defaults to false.
```
hzxuzhonghu, line 853: For bool type, this is redudant.

ramaraochavali: Yeah. Just added to be extra careful. But I can remove

ramaraochavali:
@hzxuzhonghu @howardjohn @nrjpoddar PTAL

mesh/v1alpha1/config.proto
```diff
@@ -847,8 +847,13 @@ message MeshConfig {
   // If not set, the NormalizationType.DEFAULT configuration will be used.
   ProxyPathNormalization path_normalization = 61;
 
+  // Currently When handling HTTP/1.1, Istio will normalize the header keys to be all lowercase as 
+  // per the HTTP/1.1 specification. This can potentially result in issues when migrating existing systems 
+  // that might rely on specific header casing. Set this true to preserve header casing.
+  bool preserve_header_key_casing = 62;
```
nrjpoddar, line 853: Should this be a wrapper type?

mesh/v1alpha1/config.proto
```diff
@@ -847,8 +847,13 @@ message MeshConfig {
   // If not set, the NormalizationType.DEFAULT configuration will be used.
   ProxyPathNormalization path_normalization = 61;
 
+  // Currently When handling HTTP/1.1, Istio will normalize the header keys to be all lowercase as 
+  // per the HTTP/1.1 specification. This can potentially result in issues when migrating existing systems 
+  // that might rely on specific header casing. Set this true to preserve header casing.
```
nrjpoddar, line 852:
```suggestion
  // Istio proxies by default normalize HTTP/1.1 header keys to lowercase as per the HTTP/1.1 specification.
  // This can potentially lead to issues when migrating existing systems that rely on header casing being preserved.
  // Set this true to preserve HTTP/1.1 header casing.
```

mesh/v1alpha1/config.proto
ramaraochavali: I do not think it needs wrapper type. Any reason why you think it should be?

hzxuzhonghu (approved):

mesh/v1alpha1/config.proto
costinm: Are there any other header options or filters related to headers ? If yes or we expect others, we can have 
headers: 
  preserve_case: true
  max_len: 1000
  ...

or even:

header_options: 
  - preserve_case
  - filter_x_headers
  - ???

ramaraochavali: I am fine with adding top level header_options if we anticipate fields in this category in future.

mesh/v1alpha1/config.proto
```diff
@@ -847,8 +847,13 @@ message MeshConfig {
   // If not set, the NormalizationType.DEFAULT configuration will be used.
   ProxyPathNormalization path_normalization = 61;
 
+  // Istio proxies by default normalize HTTP/1.1 header keys to lowercase as per the HTTP/1.1 specification.
+  // This can potentially lead to issues when migrating existing systems that rely on header casing being preserved.
+  // Set this true to preserve HTTP/1.1 header casing.
```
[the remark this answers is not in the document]
ramaraochavali: This can not be applied at proxy level just. This has to be a mesh config for it to work - all proxies that the request traverses through needs to have this config enabled to make this work

ramaraochavali:
@nrjpoddar @howardjohn How do we move forward with this?

costinm:
On Sun, May 30, 2021 at 12:31 AM Rama Chavali ***@***.***>
wrote:

>
> <https://github.com/istio/api/pull/2004#discussion_r642034322>:
>
> > @@ -847,8 +847,13 @@ message MeshConfig {
>    // If not set, the NormalizationType.DEFAULT configuration will be used.
>    ProxyPathNormalization path_normalization = 61;
>
> +  // Istio proxies by default normalize HTTP/1.1 header keys to lowercase as per the HTTP/1.1 specification.
> +  // This can potentially lead to issues when migrating existing systems that rely on header casing being preserved.
> +  // Set this true to preserve HTTP/1.1 header casing.
>
> This can not be applied at proxy level just. This has to be a mesh config
> for it to work - all proxies that the request traverses through needs to
> have this config enabled to make this work
>

Mesh is a very large level - if one workload has this requirement, do we
need the entire mesh to be messed up ?

The workload and the gateway in front of it - plus maybe the workloads that
call it.

I wouldn't mind it too much if it didn't have the risk of breaking other
applications that may expect the headers
to be normalized.

ramaraochavali:
> Mesh is a very large level - if one workload has this requirement, do we
need the entire mesh to be messed up ?

This has to be applied across all sidecars that request traverses through - it wont be enough if we enable it one entry point workload and I do not see a downside of enabling it across mesh because for proxies that does not care about case - it will be lowercase (i.e. http 1.1 compliant proxies) and this is no op for them.

costinm:
Besides - ProxyConfig has a default, which has the equivalent MeshConfig
global scope - but
can also be overridden. We just need to be careful with the merging and
default value.

In terms of header-related options - max size is another common one, as
well as 'block'
and 'set' header ( to customize which headers will be blocked, or set
specific headers globally ).
We do have some options in VirtualHost, but they are for egress usually.

ramaraochavali:
I can introduce header operations, I like that idea to be future proof.

costinm:
Proxies may not care about case - but apps may.

Apache preserves the case - and there are apps that rely on this (bad)
behavior.
HAProxy normalizes - and I assume there are apps that rely on this as well
- which is equally bad.

I think normalizing is safer and more common - and I would expect some apps
will start assuming this
is the case and skip doing their own normalization. I don't know all http
stacks apps use.

costinm:
What are we debating here ? ProxyConfig allows setting it as a mesh wide
option as well,
and I don't see any downside to allowing users to override - instead of
making it mesh-admin only.

ramaraochavali:
> What are we debating here ? ProxyConfig allows setting it as a mesh wide
option as well,
and I don't see any downside to allowing users to override - instead of
making it mesh-admin only.

:-) . I agree. I was just explaining the use case for enabling at mesh wide. I will make changes.

nrjpoddar:
> One big downside (or upside, depending on how you see it), is proxy config requires all pods to restart to update?

I agree. I don't think adding in ProxyConfig is a good idea until we fix the restart limitation. All "dynamic" config getting housed under ProxyConfig but requiring restarts is very confusing to users..

ramaraochavali:
>  At the same time if we need per workload overrides I don't know if we have a better option Sidecar/Gateway may be?

That is a good option but do we want to go it right now or wait till a use case comes?

mesh/v1alpha1/config.proto
howardjohn: Maybe would make esnse. Things like https://github.com/istio/istio/issues/13861 would make a lot of sense there

costinm:
There is no 'restart' limitation - if you modify MeshConfig.default, this
is automatically picked up by Istio and a new push happens,
with the config based on the new config ( including ProxyConfig ).

With the CR proposal - more fields would be 'dynamic' - and we will
document which fields are (currently) static.

Of course, if you use the ProxyConfig annotation you need to restart - but
if you add them to MeshConfig you need
to reinstall Istiod and you can't override it anyways, so it's not better.
Restart is needed only if you override - until
we have the new CR.

costinm:
We clearly have a use case - since we have this thread :-)

We can add it to Sidecar as 'alpha', apparently there is a way now (
besides just documentation ).

Neeraj's proposal to merge Sidecar/ProxyConfig sounds very good to me - if
we can do it we end in the same
place.

ramaraochavali:
@howardjohn @nrjpoddar @costinm moved to proxy config with header_options as discussed. Can you PTAL?

nrjpoddar:
Hi @ramaraochavali we discussed this extensively in yesterday's TOC and the notes are captured [here](https://docs.google.com/document/d/13lxJqtlaQhmV2EwsNnS6h-_O4pobZQZuMjrzOeMgVI0/edit#bookmark=id.q87xoxdhvz36).

Basically, we landed on the decision to not add any API for this option, nor preserve case by default as when we switch to H2 transport for HTTP/1.1 we will loose header casing and all headers will be normalized to lower case. For now, EnvoyFilter at a workload level granularity is the best bet. Happy to chat more if needed.

ramaraochavali:
@nrjpoddar Thanks for the update. Fair enough. I will close this PR.

PR #2012, opened 2021-05-28
Added a note that wildcard names are not exported.

tbartelmess:
See: https://github.com/istio/istio/issues/14817

### DIFF
--- a/networking/v1beta1/destination_rule.proto
+++ b/networking/v1beta1/destination_rule.proto
@@ -236,9 +236,7 @@ message DestinationRule {
   // the destination rule is declared in. Similarly, the value "*" is reserved and
   // defines an export to all namespaces.
   //
-  // Services that have a wildcard "*" in the host name (e.g.
-  // "*.default.svc.cluster.local") are not exported to other namespaces.
-  // The only exception is the "istio-system" namespace.
+  // Destination rules can only export services from the same namespace.
   repeated string export_to = 4;
 }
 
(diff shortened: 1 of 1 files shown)


tbartelmess:
@googlebot I signed it!

networking/v1beta1/destination_rule.proto
```diff
@@ -235,6 +235,10 @@ message DestinationRule {
   // The value "." is reserved and defines an export to the same namespace that
   // the destination rule is declared in. Similarly, the value "*" is reserved and
   // defines an export to all namespaces.
+  //
+  // Services that have a wildcard "*" in the host name (e.g.
+  // "*.default.svc.cluster.local") are not exported to other namespaces.
+  // The only exception is the "istio-system" namespace.
```
howardjohn, line 241: I think this is not quite accurate if I understand right. It's not just wildcards, anything in the wrong namespace.

The lookup order is client namespace, server namespace, root namespace (typically istio-system - we should change the text here to mention root namespace instead of istio-system). The issue here is that a rule defined is in none of these 3 namespaces so it's not applied

tbartelmess: Sorry, you are right, it's everything from a different namespace, I'll update it.

networking/v1beta1/destination_rule.proto
```diff
@@ -235,6 +235,8 @@ message DestinationRule {
   // The value "." is reserved and defines an export to the same namespace that
   // the destination rule is declared in. Similarly, the value "*" is reserved and
   // defines an export to all namespaces.
+  //
+  // Destination rules can only export services from the same namespace.
```
linsun, line 239: Is this sentence much different from ` The value "." is reserved and defines an export to the same namespace that
 the destination rule is declared in.` in line 235?

PR #2016, opened 2021-06-04
Make defaultRevision BoolValue instead of raw bool

Monkeyanator:
Need to be able to distinguish between unset `defaultRevision` and user explicitly setting `defaultRevision: false`


operator/v1alpha1/operator.proto
```diff
@@ -97,7 +99,8 @@ message IstioOperatorSpec {
 
     // Identify whether this revision is the default revision for the cluster
     // This option is currently experimental.
-    bool default_revision = 17;
+    google.protobuf.BoolValue default_revision = 17;
```
costinm, line 102: That's still pretty inconvenient. Bools are bad, we should avoid them.

Also 'default_revision' is not ideal - can we turn this into an enum ? Can have 2 values for start, but I suspect we may have others in future. 

Comments are not clear - what does it mean for a user that it's a 'default revision' ?

Monkeyanator: Other than `DEFAULT` and `NOT DEFAULT` what other types of revisions could there be? Agreed an enum would be better but having a field like `revisionType: DEFAULT` or `revisionType: NOT DEFAULT` feels awkward.

The comments are outdated, I'll update.

costinm:
I would also avoid using the term 'revision' in the API.

With string:
installType: "" ( default value is empty )
installType: "default"

In future we may refine it - for example add "injection" for an install
that will be used for injection only,
or "ca" for an install that will be used for CA only ( while Istiod has all
the functions, it can also be
deployed in a micro-services model ).

operator/v1alpha1/operator.proto
costinm: It's not going to be 'NOT_DEFAULT' - just empty ( default value ).

install_type: default  

In future we may have 'inject_only' or some other modes of installing.

hzxuzhonghu: How about make it string, if unset, default it to the version installed

PR #2020, opened 2021-06-10
Fix config_source in EnvoyFilter docstring

dgn:
Fixes istio/istio#32415

This example doesn't actually work (see referenced issue). I copied the correct example from the docs: https://istio.io/latest/docs/ops/configuration/extensibility/wasm-module-distribution/

### DIFF
--- a/networking/v1alpha3/envoy_filter.proto
+++ b/networking/v1alpha3/envoy_filter.proto
@@ -362,12 +362,7 @@ import "networking/v1alpha3/sidecar.proto";
 //         name: my-wasm-extension # This must match the name above
 //         config_discovery:
 //           config_source:
-//             api_config_source:
-//               api_type: GRPC
-//               transport_api_version: V3
-//               grpc_services:
-//               - envoy_grpc:
-//                   cluster_name: xds-grpc
+//             ads: {}
 //           type_urls: ["envoy.extensions.filters.http.wasm.v3.Wasm"]
 // ```
 //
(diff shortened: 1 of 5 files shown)


bianpengyuan (approved):

networking/v1alpha3/envoy_filter.proto
```diff
@@ -362,12 +362,8 @@ import "networking/v1alpha3/sidecar.proto";
...
 //         config_discovery:
 //           config_source:
-//             api_config_source:
-//               api_type: GRPC
-//               transport_api_version: V3
-//               grpc_services:
-//               - envoy_grpc:
-//                   cluster_name: xds-grpc
+//             ads: {}
+//             initial_fetch_timeout: 0s # wait indefinitely to prevent bad Wasm fetch
```
howardjohn, line 366: Is this a good idea? This means a single missing filter can block the entire workload startup?

For Gateway SDS we do not set this btw, to prevent a single secret from blocking startup.

bianpengyuan: Hmm yeah, indeed. Should only be used in ECDS.

dgn: should I remove the line? after all it's just an example

bianpengyuan: Yeah let's remove. People tend to copy paste and not check every field.

howardjohn (approved):

PR #2024, opened 2021-06-17
Updated CODEOWNERS to reflect the surviving TOC

nrjpoddar:


smawson (approved):
Survival of the fittest?

CODEOWNERS
```diff
@@ -1,4 +1,4 @@
-*                @linsun @dcberg @louiscryan @smawson @duderino @nrjpoddar
+*                @linsun @louiscryan @smawson @nrjpoddar
```
brian-avery, line 1: Out of curiosity, why don't we use `istio/technical-oversight-committee` in codeowners?

nrjpoddar:
> Survival of the fittest?

I hope that's the case ;)

CODEOWNERS
nrjpoddar: Good question. I don't know, not an expert on all of this but I would happily take recommendation from someone like you who is. We can also do some trials as follow-on PRs?

duderino:
A small sob was heard in the distance... around the vicinity of the great america amusement park

CODEOWNERS
linsun: I think we have individual listed here to ensure the approval from at least one TOC on api repo.  group didn't work when we tried it 2 years ago

linsun (approved):

PR #2025, opened 2021-06-17
Fix jwt rule note

shamsher31:
To fix https://github.com/istio/istio.io/issues/9914

### DIFF
--- a/security/v1beta1/jwt.proto
+++ b/security/v1beta1/jwt.proto
@@ -89,13 +89,13 @@ message JWTRule {
   //
   // Example: `https://www.googleapis.com/oauth2/v1/certs`
   //
-  // Note: Only one of `jwks_uri` and `jwks` should be used.
+  // Note: Only one of `jwksUri` and `jwks` should be used.
   string jwks_uri = 3;
 
   // JSON Web Key Set of public keys to validate signature of the JWT.
   // See https://auth0.com/docs/jwks.
   //
-  // Note: Only one of `jwks_uri` and `jwks` should be used.
+  // Note: Only one of `jwksUri` and `jwks` should be used.
   string jwks = 10;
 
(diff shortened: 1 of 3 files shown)


howardjohn (approved):

shamsher31:
cc: @nrjpoddar @linsun

security/v1beta1/jwt.proto
```diff
@@ -89,13 +89,13 @@ message JWTRule {
   //
   // Example: `https://www.googleapis.com/oauth2/v1/certs`
   //
-  // Note: Only one of jwks_uri and jwks should be used. jwks_uri will be ignored if it does.
+  // Note: Only one of `jwks_uri` and `jwks` should be used.
```
hzxuzhonghu, line 92: jwksUri

shamsher31: I have seen `jwks_uri` mentioned in multiple release notes [this](https://github.com/istio/istio/pull/30014/files#diff-01aa61a7126fd7b057a98c3354466573817b5610e443038c0fcaab72887fa2d7R8) and [this](https://github.com/istio/istio/pull/30679/files#diff-9f7f93e2212215ca68c7d07ad7f19b0343ba73b6729c2f59d2b1887f6cf5186dR8)

Also `RequestAuthentication` CRD has both `jwks_uri` and `jwksUri`, so IMO one of them will be deprecated.

Any reason we should write `jwksUri` here?

howardjohn: They are not two different fields, just two different representations of the same field. I believe `jwks_uri` is not valid in YAML so we probably don't want to use it in docs

hzxuzhonghu: Yes, the proto is to developer, but the comments are to user

hzxuzhonghu (approved):

PR #2030, opened 2021-06-21
Components for remote and config cluster installations.

frankbu:
Part of https://github.com/istio/istio/issues/33455

Release notes will be in https://github.com/istio/istio/pull/33527


operator/v1alpha1/operator.proto
```diff
@@ -171,6 +171,11 @@ message IstioComponentSetSpec {
     repeated GatewaySpec ingress_gateways = 40;
     repeated GatewaySpec egress_gateways = 41;
 
+    // Resources on a remote cluster using an external control plane.
+    ComponentSpec remote = 42;
```
howardjohn, line 175: I am a bit confused here, sorry if I missed some discussion. I thought we were moving away from `istiod_remote` to using some setting in values. So why we are now adding back `remote`?

And if we do want to do this, why are we making a new component `remote` instead of using the existing `istiod_remote`?

frankbu: > I am a bit confused here, sorry if I missed some discussion. I thought we were moving away from istiod_remote to using some setting in values. So why we are now adding back remote?

@costinm in the last Environments WG call, as well as the fact that a bunch of new unwanted resources were recently moved from base to istio-discovery and would require more values settings to filter for external istiod, convinced me that the previous cleanup approach was resulting in a lot of spaghetti code, and was a failed experiment, so this is effectively reverting the previous change, but with hopefully a cleaner design.

> And if we do want to do this, why are we making a new component remote instead of using the existing istiod_remote?

The previous `istiod_remote` component was both the new `remote` and `config` components combined. This new `remote` component will only install what is needed on a remote cluster to use an external istiod (i.e., essentially just the mutatingwebhook). The `config` component is what is needed of a config cluster for an external istiod. A config cluster will typically have both components enabled, while additional remote clusters will only enable the `remote` component.

howardjohn: This feels like a lot of churn for the user? Why don't we just keep the original `istiod_remote` component and add some flag like cluster.type={config,remote} or similar?

We are making so many install changes to 1.11 that I am increasingly worried that something is going to blow up...

smawson (approved):

operator/v1alpha1/operator.proto
frankbu: I would prefer a cleaner separation, but I see your point. I've reverted back to the original istiod_remote component.

linsun:
Any 1.10 notes we need to update?  not sure if you deprecated the istiod remote component there.

linsun (approved):

linsun:
Frank replied offline no need to update anything in 1.10 notes.

Issue #2033, opened 2021-06-23
Doc.: 'The format is “[/]”'  on Global Mesh Options page.


dsbos:
The row currently at https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ExtensionProvider-EnvoyExternalAuthorizationHttpProvider-service, along with several others, says:

> The format is “[/]”. 

That doesn't make any sense.

From looking at the raw (source) HTML, it seems that someone forgot to encode certain "<" characters in your documentation source, and that that someone didn't even look at the rendered output after making changes.

yangminzhu:
I don't think we should encode the "<" character in the protobuf file. It should be the html generation tool's responsibility to handle these issues.
But I think it's probably better and much easier to just quote the string with the ` character, will send out a PR.

PR #2034, opened 2021-06-23
fix the html issue of unescaped < character

yangminzhu:
Fixes https://github.com/istio/api/issues/2033


smawson (approved):

linsun (approved):

nrjpoddar:
@yangminzhu does this need to cherry-picked to 1.9 and possibly older versions?

yangminzhu:
> @yangminzhu does this need to cherry-picked to 1.9 and possibly older versions?

ah, thanks for the reminder, it should be cherry-picked for 1.9, it's new in 1.9 so we do not need for older versions.

PR #2038, opened 2021-06-29
add explicit call out about the subtle difference between empty vs omitted

rsalmond:
I think I made this change in the wrong place with istio/istio.io#10015, if this is where those docs are generated from I will leave this open and close that docs PR.

### DIFF
--- a/security/v1beta1/authorization_policy.proto
+++ b/security/v1beta1/authorization_policy.proto
@@ -243,6 +243,9 @@ message AuthorizationPolicy {
   //
   // If not set, the match will never occur. This is equivalent to setting a
   // default of deny for the target workloads.
+  //
+  // If set but left empty, the match will always occur. This is equivalent to
+  // setting a default of allow for the target workloads.
   repeated Rule rules = 2;
 
   // Action specifies the operation to take.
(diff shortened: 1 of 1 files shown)


security/v1beta1/authorization_policy.proto
```diff
@@ -243,6 +243,9 @@ message AuthorizationPolicy {
   //
   // If not set, the match will never occur. This is equivalent to setting a
   // default of deny for the target workloads.
+  //
+  // If set but left empty, the match will always occur. This is equivalent to
```
yangminzhu, line 247: this is a list field so technically you cannot `set but left empty`, you are actually adding an empty rule to the list.

An empty rule means the policy will always match, but the behavior (allow or reject) depends on the action. ALLOW action means always allowed, DENY action means always rejected.

rsalmond: Agreed, I also did not like that wording but I wanted to try to use terms consistent with the existing documentation. If you are okay with it I will rewrite both statements to be more accurate. What do you think of something like this?

`If there are no elements in the list, the match will never occur...`

`If an empty match rule is added to the list, the match will always occur ...`

>the behavior (allow or reject) depends on the action.

I based that explanation of "empty means always allow" on the example of an empty rule which appears directly above [the table describing these fields in the docs](https://istio.io/latest/docs/reference/config/security/authorization-policy/#AuthorizationPolicy)

image

If your characterization is more correct then we should we change this example as well?

howardjohn (approved):

PR #2040, opened 2021-06-29
Introduce InsecureSkipVerify to DR

Kmoneal:
* hide VerifyCertificateAtClient in ProxyConfig to eventually be
removed.
* Add InsecureSkipVerify bool to allow users to prevent any certificate
validation on desired external host.


jacob-delgado:
cc @ericvn can you go over the documentation for this PR? Thanks!

brian-avery:
This is part of https://docs.google.com/document/d/100Jvwjr9KkURiF2HtiOtRlAON8TqgskUyLfT17_xoRM/edit#

releasenotes/notes/25652.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+issue:
+  - 25652
+
+releaseNotes:
+  - |
+    **Added** InsecureSkipVerify to DestinationRule. Users can choose to opt-out of validating external site certificates on hosts basis if a user has enabled certificate validation.
```
craigbox, line 9: on a per-host basis?

releasenotes/notes/25652.yaml
```diff
@@ -0,0 +1,10 @@
...
+    **Modified** VerifyCertificateAtClient is now deprecated in meshconfig. This will be removed in a future release and was never implemented.
```
craigbox, line 10: `meshConfig`? 

"This feature was not implemented, and will be removed in a future release" ?

craigbox:
while Eric is out, Brian asked me to look at this; the only thing I think that is relevant to docs is

```// $hide_from_docs```

so LGTM with some comments on the release note

File: networking/v1alpha3/destination_rule.proto [hunk 3]
```diff
@@ -964,6 +964,12 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // Boolean specifying if proxy should not check the CA signature for the host
+  // certificate. `InsecureSkipVerify` is `false` by default. If set to true,
+  // the proxy will skip verification of the host's certificate regardless of
```
craigbox, line 970:
```suggestion
  // the proxy will skip verification of the host's certificate, regardless of
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -964,6 +964,12 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // Boolean specifying if proxy should not check the CA signature for the host
```
craigbox, line 968: if _the_ proxy?

hzxuzhonghu:
lg, need to run `make gen`

networking/v1alpha3/destination_rule.proto
```diff
@@ -964,6 +964,12 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // Boolean specifying if the proxy should not check the CA signature for the host
+  // certificate. `InsecureSkipVerify` is `false` by default. If set to true,
+  // the proxy will skip verification of the host's certificate, regardless of
+  // any CA certificate specified elsewhere.
```
nrjpoddar, line 971:
```suggestion
  // InsecureSkipverify specifies whether the proxy should skip verifying the CA signature for the server
  // certificate. `InsecureSkipVerify` is `false` by default. If set to true,
  // the proxy will skip verification of the server's certificate, regardless of
  // CA certificate path specified elsewhere.
```

nrjpoddar (commented):
You need to sync v1alpha3 to v1beta1 APIs. @howardjohn recently wrote a tool for it so you can just run it and ensure the proto files are in sync.

networking/v1alpha3/destination_rule.proto
nrjpoddar: Does it also not verify the SAN field?

Kmoneal: This would also not verify the SAN past the cert and host name match

Kmoneal:
@howardjohn @nrjpoddar If you have any additional thoughts or feedback, I would greatly appreciate it!

networking/v1alpha3/destination_rule.proto
nrjpoddar: Ok basically certificate signature and Host in SAN is validated. If that's correct we should not just mention signature here.

releasenotes/notes/25652.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+issue:
+  - 25652
+
+releaseNotes:
+  - |
+    **Added** InsecureSkipVerify to DestinationRule. Users can choose to opt-out of validating external site certificates on a per-host basis if a user has enabled certificate validation.
```
nrjpoddar, line 9:
```suggestion
    **Added** InsecureSkipVerify to DestinationRule. Users can opt-out of validating server certificates on a per-host basis if certificate validation is enabled globally.
```

Kmoneal: I like this a lot more. I just wanted to clarify; `InsecureSkipVerify` disables validation that was enabled anywhere so I left out globally but would be happy to add it if I have a misunderstanding

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,12 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // Boolean specifying if the proxy should not check the CA signature or SAN
```
nrjpoddar, line 971:
```suggestion
  // Boolean specifying if the proxy should not check the CA signature and SAN
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,12 @@ message ClientTLSSettings {
...
+  // for the host certificate. `InsecureSkipVerify` is `false` by default. If
```
nrjpoddar, line 972:
```suggestion
  // for the server certificate corresponding to the host. `InsecureSkipVerify` is `false` by default. If
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,12 @@ message ClientTLSSettings {
...
+  // set to true, the proxy will skip verification of the host's certificate,
```
nrjpoddar, line 973:
```suggestion
  // set to true, the proxy will skip verification of the server certificate,
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,12 @@ message ClientTLSSettings {
...
+  // regardless of any CA certificate specified elsewhere.
```
nrjpoddar, line 974:
```suggestion
  // regardless of any certificate validation configuration provided elsewhere.
```

nrjpoddar:
Few more minor suggestions, rest look good!

nrjpoddar (approved):

Kmoneal:
@brian-avery  if you have time, I would appreciate any input!

networking/v1beta1/destination_rule.proto
```diff
@@ -967,6 +967,13 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // Boolean specifying if the proxy should not check the CA signature and SAN
+  // for the server certificate corresponding to the host. `InsecureSkipVerify`
+  // is `false` by default. If set to true, the proxy will skip verification of
+  // the server certificate, regardless of any certificate validation
+  // configureation provided elsewhere.
```
howardjohn, line 975:
```suggestion
  // configuration provided elsewhere.
```

mesh/v1alpha1/config.proto
```diff
@@ -436,7 +437,7 @@ message MeshConfig {
   // For wildcard host name in DestinationRule, client-side proxy will do a suffix match. For example,
   // if host is `*.x.y.com`, client-side proxy will verify the presented server certificate SAN matches
   // ``.x.y.com` suffix.
-  google.protobuf.BoolValue verify_certificate_at_client = 54;
+  google.protobuf.BoolValue verify_certificate_at_client = 54 [deprecated=true];
```
linsun, line 440: Can you clarify default here given this is not configurable?

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,13 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // Boolean specifying if the proxy should not check the CA signature and SAN
+  // for the server certificate corresponding to the host. `InsecureSkipVerify`
+  // is `false` by default. If set to true, the proxy will skip verification of
+  // the server certificate, regardless of any certificate validation
+  // configureation provided elsewhere.
```
linsun, line 975:
```suggestion
  // configuration provided elsewhere.
```

mesh/v1alpha1/config.proto
linsun: Is the scenario that this is set to true by default globally but certain service wants to skip the validation by setting insecure_skip_verify?

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,13 @@ message ClientTLSSettings {
...
+  google.protobuf.BoolValue insecure_skip_verify = 8;
```
linsun, line 976: what about `skip_verify_cert` or `skip_verify_certificate`?  seems easier to understand and consistent with `VerifyCertificateAtClient`

louiscryan, line 976: Is the certificate expiration checked?

@nrjpoddar If we're confident that we won't need more control then I'm OK with not being an enum.

mesh/v1alpha1/config.proto
nrjpoddar: By default, we don't verify server certificates unless you specify `caCertificates` in DR which is the bug (CVE) we are trying to address. @Kmoneal is adding an ENV var in istio/istio so that going forward CA verification can be turned on globally (it will still be default off for backwards compatibility). Once you have it enabled globally, the new config option in DR insecureSkipVerify disables it on a per host (service) basis.

networking/v1alpha3/destination_rule.proto
nrjpoddar: Negation makes more sense and is consistent with tools like curl/openssl that provide "-k" flags.

mesh/v1alpha1/config.proto
Kmoneal: Neeraj summed it up well. This was an attempt to add in enabling mesh-wide external server certificate validation. By planning to default verification to on, we wanted to make sure that we can allow users to disable it mesh-wide.

networking/v1alpha3/destination_rule.proto
Kmoneal: We wanted to make sure it was explicit that it was an insecure flag to enable and I prefer to keep that. If `insecure_skip_verify_cert` is agreeable I would be happy to change it.

nrjpoddar: I guess so? I'm assuming Envoy uses the openssl call to verify most expected attributes like signature, SAN, expiration, not valid before etc. We should refine the comment to not explicitly list out just signature and SAN.

I think starting out with boolean is good here. I haven't seen many requests for signature verification but no SAN or skip expiration etc.

This is what curl documents for -k:

`The server connection is verified by making sure the server's certificate contains the right name and
  verifies successfully using the cert store.`

Kmoneal: The certificate would be checked that it is a usable cert (expiration, not malformed, etc) but it will not check against a CA.

howardjohn: I like InsecureSkipVerify because it has a wellknown meaning already

howardjohn: `curl -k` doesn't verify expiration (you can check with https://expired.badssl.com/) fwiw

nrjpoddar (commented):
If folks questions are answered can we approve this? @howardjohn @linsun @louiscryan? Thanks!

networking/v1beta1/destination_rule.proto
```diff
@@ -967,6 +967,13 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // InsecureSkipverify specifies whether the proxy should skip verifying the
+  // CA signature and SAN for the server certificate corresponding to the host.
+  // `InsecureSkipVerify` is `false` by default. If set to true, the proxy will
```
howardjohn, line 973: is "`false` by default" accurate?

howardjohn (approved):

howardjohn:
Put a hold for small comment

networking/v1beta1/destination_rule.proto
Kmoneal: Feel free to correct me if I am missing something but `InsecureSkipVerify` would disable any other added CA Root Certs or SAN specified which means it should be `false`. Defaulting to `false` would have no changes to current environments once the istio modifications are made. This is a new feature to disable security checks per host and the [Env Var portion](https://github.com/istio/istio/pull/33472) is adding the additional security setting which will be disabled by default as well.

nrjpoddar:
> Put a hold for small comment

John, @Kmoneal responses to your comment so PTAL and remove hold so this can collapse.

networking/v1beta1/destination_rule.proto
howardjohn: This seems misleading. If someone told me InsecureSkipVerify is false by default, and I configure a DR for `https://google.com` (with no caCertificate config), I think the vast majority of users would expect that we are verifying the TLS.
Even without this comment, today most people (including most Istio developers), thought it was verifying. This furthers the confusion IMO.

Kmoneal: I agree, it is misleading and may provide a false sense of security for the standard settings. I took a second look and wanted to know if any of these solutions would be able to fix any concerns?

1. Hide `InsecureSkipVerify` from docs until OS CA Certificates verification is enabled by default.
2. Switch to an enum like the one below. This could also give a false sense of security unless `VERIFY` is added only when it is available. It will be a change from RFC but considering it has been talked about a couple of time I think it would be worth it
```
    enum InsecureSkipVerifyMode {
        // Keep the legacy behavior, no change
        UNSPECIFIED = 0;
        // Do not verify certificates against a CA but verify that the certificate is valid
        // with basic checks (i.e. expiration check, issued to the expected host, etc.)
        NO_VERIFY = 1;
        // Allow any verification enabled to be unchanged
        VERIFY = 2;
    }
```
3. Change wording to specify that this just disables added DestinationRule and ServiceEntry CA/SAN settings verification.

If there is anything else you would recommend or any modifications I would be happy to hear them.

nrjpoddar: I would suggest adding clarification around how this flag works in relationship with the environment variable i.e. this flag should only be used when the env var is enabled and you have verification turned on globally but want it disabled for a specific host. I would even add which Istio releases have the env var default off and later when we toggle it add from the release it is by default on. WDYT?

howardjohn: +1 to documenting correlation to the flag.

networking/v1alpha3/destination_rule.proto
```diff
@@ -968,11 +968,17 @@ message ClientTLSSettings {
...
-  // validation configuration provided elsewhere.
+  // This flag should only be set if global CA signature verifcation is
+  // enabled, `VerifyCertAtClient` environmental variable is set to `true`,
+  // but no verification is desired for a specific host. If enabled with or
+  // without `VerifyCertAtClient` enabled, verification of the CA signature and
+  // SAN will be skipped.
+  //
+  // `InsecureSkipVerify` is `false` by default.
+  // `VerifyCertAtClient` is `false` by default in Istio version 1.9 until
+  // Istio version 1.13 where, going forward, it will be enabled by default.
```
howardjohn, line 981: Nit: would avoid exact date, historically we don't hit our targets a lot :slightly_smiling_face:

howardjohn (approved):

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,20 @@ message ClientTLSSettings {
 
   // SNI string to present to the server during TLS handshake.
   string sni = 6;
+
+  // InsecureSkipVerify specifies whether the proxy should skip verifying the
+  // CA signature and SAN for the server certificate corresponding to the host.
+  // This flag should only be set if global CA signature verifcation is
+  // enabled, `VerifyCertAtClient` environmental variable is set to `true`,
```
nrjpoddar, line 974: // This flag should only be used if server certificate validation has been globally enabled
// via setting the environment variable `VERIFY_CERTIFICATE_AT_CLIENT` to true
// but no verification is desired for a specific host.

networking/v1alpha3/destination_rule.proto
```diff
@@ -967,6 +967,20 @@ message ClientTLSSettings {
...
+  // but no verification is desired for a specific host. If enabled with or
+  // without `VerifyCertAtClient` enabled, verification of the CA signature and
+  // SAN will be skipped.
+  //
+  // `InsecureSkipVerify` is `false` by default.
+  // `VerifyCertAtClient` is `false` by default in Istio version 1.9 but will
```
nrjpoddar, line 980: // Environment variable `VERIFY_CERTIFICATE_AT_CLIENT` is `false` by default in Istio versions 1.9 - 1.11
// but will be `true` by default in a later version where, going forward, it will be enabled by default.
// It is strongly recommended to set this environment variable to true in production environments.

nrjpoddar (commented):
@Kmoneal I missed updates to this PR so added few suggestions. PTAL.

PR #2041, opened 2021-06-30
[on-hold]add cert-signer into proxyconfig

irisdingbj:
To support workload define cert signer which will be used to sign workload certs.

details in : https://docs.google.com/document/d/1rxTSLuepcE_OJHj4C23RMJcgyIZzlBI-GqO8Y37SPko/edit#heading=h.b5qprpxs2o1a


irisdingbj:
@howardjohn A ping on this. Thanks!

mesh/v1alpha1/proxy.proto
```diff
@@ -521,6 +521,10 @@ message ProxyConfig {
   // The plugin certificates (the 'cacerts' secret), self-signed certificates (the 'istio-ca-secret' secret)
   // are added automatically by Istiod.
   repeated string ca_certificates_pem = 34;
+
+  // The cert signer which will be used to sign certificates for workload-to-workload communication.
+  // This only takes effect if Istiod is acting as RA (registration authority)
+  string cert_signer = 35;
```
howardjohn, line 527: I am a bit concerned about adding another top level field here.

https://docs.google.com/document/d/1KUQlAGnpajDJC9G6Em-OGGCn_xEftpw-J6TXcMB_MNE/edit# proposed adding a new struct here for `certificate` and related settings. Can you work with @jacob-delgado to see if it makes sense to merge these in under a common message type?

irisdingbj: @jacob-delgado I added your proposed changes in https://docs.google.com/document/d/1KUQlAGnpajDJC9G6Em-OGGCn_xEftpw-J6TXcMB_MNE/edit#  for proxyConfig part.  Also one additonal field cert-signer which can be used to sepcify the cert -signer.

Let me know if this relfects your proposal.

@howardjohn please add your comments as well.

jacob-delgado: Added some comments https://docs.google.com/document/d/1rxTSLuepcE_OJHj4C23RMJcgyIZzlBI-GqO8Y37SPko/edit?disco=AAAANDtF3PE


but changes made could be similar to

```proto
message ProxyConfig {
  ...
  CertificateParameters certificate = 34;
}

message CertificateParameters {
  // The cert signer which will be used to sign certificates for workload-to-workload communication.
  // This only takes effect if Istiod is acting as RA (registration authority)
  string cert_signer = 1;
}
```

irisdingbj: Thanks @jacob-delgado !  @howardjohn given that can go ahead with above solution? This makes it possible to add other certificate parameters in the future and also lands this cert-signer things.

howardjohn:
Sorry to run in circles a bit. There was a new proposal that has substantial overlap as well: https://docs.google.com/document/d/1QACaJW8LpQMvdbPd-EoVtYNcNprTb6vScj72k-qHyT4/

We should make sure we reconcile the two @liminw

irisdingbj:
> Sorry to run in circles a bit. There was a new proposal that has substantial overlap as well: https://docs.google.com/document/d/1QACaJW8LpQMvdbPd-EoVtYNcNprTb6vScj72k-qHyT4/
> 
> We should make sure we reconcile the two @liminw

@liminw Good idea. But can we add the k8s CSR as an option into your proposal? I also left a comment here. 

@howardjohn Can we make cert-signer into `proxyMetadata` for now to unblock the dependency? We can move it into the API @liminw is working on in the future once it is landed.

mesh/v1alpha1/certificate.proto
```diff
@@ -0,0 +1,43 @@
...
+  google.protobuf.Duration duration = 1;
+
+  message Key {
+    // Specify the signature algorithm to use when generating private keys
+    string algorithm = 1;
+
+    // Specify the private key size
+    int32 size = 2;
+   }
+   Key private_key = 3;
```
costinm, line 38: Add a comment that the names are from CertManger. Pretty poor names - but good to be consistent.

costinm (commented):
I'm ok with this PR, moving the env variables to proto and mostly reusing a small subset of CertMangaer API.

costinm (approved):

mesh/v1alpha1/certificate.proto
```diff
@@ -0,0 +1,43 @@
...
+// CertificateParameters are the parameters used to configure certificates issued
+// adapted from cert-manager (see https://cert-manager.io/docs/reference/api-docs)
+message CertificateParameters {
+  // The cert lifetime of the self-signed CA certificate.
+  // Default TTL duration is `8760h0m0s` for self-signed root
+  // cert and `24h0m0s` for created workloads.
+  google.protobuf.Duration duration = 1;
+
+  message Key {
+    // Specify the signature algorithm to use when generating private keys
```
howardjohn, line 32: Can we list which are valid?

mesh/v1alpha1/certificate.proto
```diff
@@ -0,0 +1,43 @@
...
+    string algorithm = 1;
+
+    // Specify the private key size
+    int32 size = 2;
```
howardjohn, line 36: Can we list valid sizes? IIRC 2048 is the minimum

mesh/v1alpha1/certificate.proto
howardjohn: Also just a general comment how it is used

mesh/v1alpha1/proxy.proto
```diff
@@ -521,6 +522,8 @@ message ProxyConfig {
   // The plugin certificates (the 'cacerts' secret), self-signed certificates (the 'istio-ca-secret' secret)
   // are added automatically by Istiod.
   repeated string ca_certificates_pem = 34;
+
+  CertificateParameters certificate = 35;
```
howardjohn, line 526: Add comments

howardjohn (commented):
cc @jacob-delgado @shankgan @myidpt @liminw please take a look

mesh/v1alpha1/certificate.proto
costinm: +1 - but 'which is supported by Istiod (citadel)' - if external CAs are used it's not our concern really, but the policy.

Same for size.

PR #2043, opened 2021-07-01
General topology aware loadbalancing

hzxuzhonghu:
This is to support general prioritized load balancing specified by a set of general labels. 

This is based on the WG meeting consensus.  https://docs.google.com/document/d/1qDmSwi_83VxvWFz1Um4dX4Cr1zW2PHq-UCWG4v7R0IA/edit#bookmark=id.lgmdj9jie0d

To replace #2002

~TODO: for per service lb, we need to further discuss, which can be a follow-up.~ (Got consensus on providing the flexibility)


mesh/v1alpha1/config.proto
```diff
@@ -283,6 +283,37 @@ message MeshConfig {
...
 
+  // TopologyKeys is a preference-order list of topology keys used to sort endpoints to do priority load balancing.
+  // This is to support traffic fail over across different group of endpoints.
+  // Suppose there are total N keys specified:
+  //   endpoints match topologyKeys[0:-1] with downstream proxy have priority P(0).
+  //   endpoints match topologyKeys[0:-2] with downstream proxy have priority P(1).
+  //   by analogy, endpoints match topologyKeys[0:-N+1] with downstream proxy have priority P(N-1).
+  //   endpoints match no keys have priority P(N)
+  //
+  // In theory the topology key can be any label, currently only "NETWORK", "REGION", "ZONE", "SUBZONE" are supported.
```
howardjohn, line 294: I think the consensus in the WG was to use the real label names (ie `topology.istio.io/network`, etc). In this way, we do not have any special values and everything is just a label.

mesh/v1alpha1/config.proto
```diff
@@ -283,6 +283,37 @@ message MeshConfig {
...
+  //   4. endpoints in same `["NETWORK"]` but different `REGION`, `ZONE`, `SUBZONE` with the client proxy have the third-level priority.
+  //   5. all the other endpoints that does not match any key have the lowest priority.
+  // ```yaml
+  // topologyKeys:
+  // - "NETWORK"
+  // - "REGION"
+  // - "ZONE"
+  // - "SUBZONE"
+  // ```
+  // Note: topologyKeys can not be specified simultaneously with localityLoadBalancerSetting
```
howardjohn, line 314: I was thinking that `topology_keys` modifies how `localityLoadBalancer` works. Why can they not be set at the same time?

mesh/v1alpha1/config.proto
hzxuzhonghu: Look at the below comments https://github.com/istio/api/pull/2043/files#diff-b5c7ece76d5d50226c89f3a2b7a70b22a91f3f8fc0bf079dc0e7fcd2ecc98938R295
They can be set via multi ways, but same meaning. If we allow all ways of setting, the implementation will be painful too.

Whatever this does not change any behavior.

mesh/v1alpha1/config.proto
hzxuzhonghu: say if locality failover is set in `localityLoadBalancer` but the behavior is different from here.

mesh/v1alpha1/config.proto
howardjohn: There would not be multiple ways of setting it in the API here; There is one way - `topology.istio.io/network`. These labels are already a part of the API - they can be used in DestinationRule subsets. So I think consistency there, plus the ability to use any labels, is preferred

mandarjog: This is an advanced feature, however if we were to make it default at some point network,region,zone,sub zone seems like a good default.
Since it is an advanced feature we don’t need an enum, we can just use labels as they are.
For example I may also want to use use  kubernetes.io/hostname (node) as a label, we don’t need to define the enum.
I can imagine someone defining some other label like physical-host to keep traffic within the same hypervisor.

hzxuzhonghu: Yeah, thanks for the comment. We get the consensus on WG meeting today

hzxuzhonghu:
@howardjohn ptal

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,50 @@ message LocalityLoadBalancerSetting{
...
 
+    // TopologyKeys is a preference-order list of topology keys used to sort endpoints to do priority load balancing.
+    // This is to support traffic fail over across different group of endpoints.
+    // Suppose there are total N keys specified:
+    //   endpoints match topologyKeys[0:-1] with downstream proxy have priority P(0).
+    //   endpoints match topologyKeys[0:-2] with downstream proxy have priority P(1).
+    //   by analogy, endpoints match topologyKeys[0:-N+1] with downstream proxy have priority P(N-1).
+    //   endpoints match no keys have priority P(N)
+    //
+    // In theory the topology key can be any label, currently only `topology.istio.io/network`, `topology.kubernetes.io/region`,
```
howardjohn, line 1076: Why don't we support arbitrary labels?

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,50 @@ message LocalityLoadBalancerSetting{
...
+    // Optional: only one of distribute, failover or topologyKeys can be set.
     // Explicitly specify the region traffic will land on when endpoints in local region becomes unhealthy.
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
 
+    // TopologyKeys is a preference-order list of topology keys used to sort endpoints to do priority load balancing.
+    // This is to support traffic fail over across different group of endpoints.
+    // Suppose there are total N keys specified:
+    //   endpoints match topologyKeys[0:-1] with downstream proxy have priority P(0).
```
howardjohn, line 1071: nit: I suspect this won't render as expected in the HTML docs

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,50 @@ message LocalityLoadBalancerSetting{
...
 
-    // Optional: only one of distribute or failover can be set.
+    // Optional: only one of distribute, failover or topologyKeys can be set.
     // Explicitly specify loadbalancing weight across different zones and geographical locations.
     // Refer to [Locality weighted load balancing](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/locality_weight)
     // If empty, the locality weight is set according to the endpoints number within it.
     repeated Distribute distribute = 1;
 
-    // Optional: only failover or distribute can be set.
+    // Optional: only one of distribute, failover or topologyKeys can be set.
```
howardjohn, line 1062: I think we technically could support `failover`, although it would be a bit odd. Maybe better to keep it orthogonal for now

hzxuzhonghu: Yes, better not to couple them atm.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: Will remove

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: what should i do?

hzxuzhonghu: I think i can refer to this https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-ServiceSettings-Settings

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,54 @@ message LocalityLoadBalancerSetting{
...
     // Explicitly specify the region traffic will land on when endpoints in local region becomes unhealthy.
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
 
+    // TopologyKeys is a preference-order list of topology keys used to sort endpoints to do priority load balancing.
```
ramaraochavali, line 1068: nit: TopologyKeys is a list of topology keys in the order provided, used to sort endpoints to do priority load balancing.

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,54 @@ message LocalityLoadBalancerSetting{
...
+    // This is to support traffic fail over across different group of endpoints.
+    // Suppose there are total N keys specified:
+    //
+    // 1. endpoints match topologyKeys[0:-1] with downstream proxy have priority P(0).
```
ramaraochavali, line 1072: This is not clear to me. What does topologyKeys[0:-1] mean here?

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,54 @@ message LocalityLoadBalancerSetting{
...
+    // 2. endpoints match topologyKeys[0:-2] with downstream proxy have priority P(1).
+    // 3. by analogy, endpoints match topologyKeys[0:-N+1] with downstream proxy have priority P(N-1).
+    // 4. endpoints match no keys have priority P(N).
+    //
+    // It can be any label specified on both client and server workload, also some special labels `topology.istio.io/network`, `topology.kubernetes.io/region`,
```
ramaraochavali, line 1077: It can be any label specified on both client and server workload.

The following special labels are also supported. May be better to read instead of repeating the labels twice

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: Sorry, go doesnot support [0:-1], it is python grammer. It means all the keys matched

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: SGTM

ramaraochavali:
lgtm

nrjpoddar:
@hzxuzhonghu I was under the impression that the WG agreement was to put topologyKeys under meshconfig and leave DR API unchanged for first draft. I see that this PR is doing the inverse. Was there a different decision reached afterwards?

hzxuzhonghu:
@nrjpoddar Yes, in the networking WG meeting we reached this consensus. It is just an optional field and could provide much flexibility.

nrjpoddar:
@hzxuzhonghu where is the PR to change meshconfig or are not changing meshconfig now? I'm kind of confused because the decision we made in one Networking WG meeting seems to be reversed by the next one. cc @howardjohn for clarification.

hzxuzhonghu:
@nrjpoddar `LocalityLoadBalancerSetting` is referred by meshConfig

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,54 @@ message LocalityLoadBalancerSetting{
...
     repeated Failover failover = 2;
 
+    // TopologyKeys is a list of topology keys in the order provided, used to sort endpoints to do priority load balancing.
+    // This is to support traffic fail over across different group of endpoints.
+    // Suppose there are total N keys specified:
+    //
+    // 1. endpoints match topologyKeys[0:N] with downstream proxy have priority P(0).
+    // 2. endpoints match topologyKeys[0:N-1] with downstream proxy have priority P(1).
+    // 3. by analogy, endpoints match topologyKeys[0:1] with downstream proxy have priority P(N-1).
+    // 4. endpoints match no keys have priority P(N).
```
nrjpoddar, line 1075: I understand it but it is quite hard to parse..Can we simplify it a bit?

networking/v1alpha3/destination_rule.proto
```diff
@@ -1053,18 +1053,54 @@ message LocalityLoadBalancerSetting{
...
+    //   - `topology.kubernetes.io/zone` is used to match the zone metadata of an endpoint, which maps to Kubernetes node label `topology.kubernetes.io/zone` or the deprecated label `failure-domain.beta.kubernetes.io/zone`.
+    //   - `topology.istio.io/subzone` is used to match the subzone metadata of an endpoint, which maps to Istio node label `topology.istio.io/subzone`.
+    //
+    // The below topology config indicates the following priority levels:
+    // 
+    // 1. endpoints match same [network, region, zone, subzone] attributes with the client proxy have the highest priority.
+    // 2. endpoints have same [network, region, zone] but different [subzone] attributes with the client proxy have the secondary-level priority.
+    // 3. endpoints have same [network, region] but different [zone, subzone] attributes with the client proxy have the third-level priority.
+    // 4. endpoints have same [network] but different [region, zone, subzone] attributes with the client proxy have the third-level priority.
```
nrjpoddar, line 1090:
```suggestion
    // 4. endpoints have same [network] but different [region, zone, subzone] attributes with the client proxy have the fourth-level priority.
```

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: any suggestion?

hzxuzhonghu:
@howardjohn @ramaraochavali Any other comments? We can move forward.

ramaraochavali (approved):
LGTM

howardjohn:
@nrjpoddar expressed some concerns about confusion about relation to `failover` and `topologyKeys`. I mocked up some alternative ideas:

```yaml
# 1a) Current proposal
localityLbSetting:
  topologyKeys:
  - topology.kubernetes.io/region
  - topology.kubernetes.io/zone
# howardjohn: some confusion since we cannot use "failover", but the behavior we provide *is* failing over

# 1b) Current proposal, but allow existing failover field+topologyKeys concurrently
localityLbSetting:
  failover: # applies to topology.kubernetes.io/region
  - from: us-west
    to: us-east
  topologyKeys:
  - topology.kubernetes.io/region
  - topology.kubernetes.io/zone
# howardjohn: seems reasonable but same concerns as (1a) and a bit confusing its only "region" in failover

# 1c) Current proposal, but allow customizable failover+topologyKeys concurrently
localityLbSetting:
  failover: # if key is unset, default to region. Either all or no keys must be set, to avoid confusion
  - key: topology.kubernetes.io/region
    from: us-west
    to: us-east
  - key: topology.kubernetes.io/zone
    from: zone-1
    to: zone-2
  topologyKeys:
  - topology.kubernetes.io/region
  - topology.kubernetes.io/zone
# howardjohn: same concern as (1). "failover" from zone to zone may not make sense without full hierarachy of region/zone (ie the zone name may be the same between regions and not make sense without region qualifier)


# 2) Hack existing failover. Allow either all elements to be "key" or "to/from"
localityLbSetting:
  failover:
  - key: topology.kubernetes.io/region
  - key: topology.kubernetes.io/zone
# howardjohn: this is extremely confusing. It does look decent in terms of YAML, but a bit confusing to have 2 APIs merged into a single field...

# 3) Tweak existing failover. If key is unset, it defaults to "topology.kubernetes.io/region";
# if any key is unset ALL keys must be unset
localityLbSetting:
  failover:
  - key: topology.kubernetes.io/region
    from: us-west
    to: us-east
  - key: topology.kubernetes.io/region
    from: us-east
    to: us-west
  - key: topology.kubernetes.io/zone
# howardjohn: I think this is confusing since now "key" can appear multiple times which is tricky

# 4) deprecate "failover" and create new field. Strawman is "prioritized". This can contain both topology keys and (advanced) overrides
localityLbSetting:
  prioritized:
  - key: topology.kubernetes.io/region
    overrides:
    - from: us-west
      to: us-east
    - from: eu
      to: cn
  - key: topology.kubernetes.io/zone
# howardjohn: probably a decent API, but deprecating a field is not great

# 5) Just renamed the API in current proposal. P
localityLbSetting:
  failoverTopologyKeys: # alternatives: failoverKeys, failoverLabels, failoverOrder
  - topology.kubernetes.io/region
  - topology.kubernetes.io/zone
# howardjohn: no strong preference
```

My comments are below each option. Overall, I think all of the options are pretty bad; I think the existing API is unfortunately not ideal so it makes it hard to get a great API when extending like this.

hzxuzhonghu:
@nrjpoddar WDYT?

nrjpoddar:
I think current failover field is restrictive and allows named regions to fail over which is different than prioritized failover that @hzxuzhonghu is proposing in this API. I can live with renaming `topologyKeys` to:

* failoverPriority (Only one of distribute/failover/failoverPriority is allowed)

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
     repeated Distribute distribute = 1;
 
-    // Optional: only failover or distribute can be set.
+    // Optional: only one of distribute, failover or topologyKeys can be set.
     // Explicitly specify the region traffic will land on when endpoints in local region becomes unhealthy.
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
 
+    // TopologyKeys is a list of topology keys in the order provided, used to sort endpoints to do priority load balancing.
```
nrjpoddar, line 1082:
```suggestion
    // TopologyKeys is an ordered list of labels used to sort endpoints to do priority load balancing.
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    // This is to support traffic fail over across different group of endpoints.
+    // Suppose there are total N keys specified:
+    //
+    // 1. endpoints match topologyKeys[0:N] with downstream proxy have priority P(0).
```
nrjpoddar, line 1086:
```suggestion
    // 1. endpoints matching all N labels with the client proxy have priority P(0) i.e. the highest priority.
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    // 2. endpoints match topologyKeys[0:N-1] with downstream proxy have priority P(1).
```
nrjpoddar, line 1087:
```suggestion
    // 2. endpoints matching the first N-1 labels with the client proxy have priority P(1) i.e. second highest priority.
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    // 3. by analogy, endpoints match topologyKeys[0:1] with downstream proxy have priority P(N-1).
```
nrjpoddar, line 1088:
```suggestion
    // 3. By extension of this logic, endpoints matching only the first label with the client proxy has priority P(N-1) i.e. second lowest priority.
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    // 4. endpoints match no keys have priority P(N).
```
nrjpoddar, line 1089:
```suggestion
    // 4. Endpoints matching no labels with the client proxy have priority P(N) i.e. lowest priority.
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    //
+    // It can be any label specified on both client and server workload.
```
nrjpoddar, line 1091:
```suggestion
    // It can be any label specified on both client and server workloads.
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    // The following special labels are also supported:
```
nrjpoddar, line 1092:
```suggestion
    // The following labels which have special semantic meaning are also supported:
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    //
+    //   - `topology.istio.io/network` is used to match the network metadata of an endpoint, which can be specified by pod/namespace label `topology.istio.io/network`, sidecar env `ISTIO_META_NETWORK` or MeshNetworks.
```
nrjpoddar, line 1094: What's MeshNetworks?

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    // 3. endpoints have same [network, region] but different [zone, subzone] attributes with the client proxy have the third-level priority.
+    // 4. endpoints have same [network] but different [region, zone, subzone] attributes with the client proxy have the fourth-level priority.
+    // 5. all the other endpoints that does not match any key have the lowest priority.
+    //
+    // ```yaml
+    // topologyKeys:
+    // - "topology.istio.io/network"
+    // - "topology.kubernetes.io/region"
+    // - "topology.kubernetes.io/zone"
+    // - "topology.istio.io/subzone"
```
nrjpoddar, line 1112:
```suggestion
    // The below configuration indicates the following priority levels:
    // ```yaml
    // topologyKeys:
    // - "topology.istio.io/network"
    // - "topology.kubernetes.io/region"
    // - "topology.kubernetes.io/zone"
    // - "topology.istio.io/subzone"
```
    // 
    // 1. endpoints match same [network, region, zone, subzone] label with the client proxy have the highest priority.
    // 2. endpoints have same [network, region, zone] label but different [subzone] label with the client proxy have the second highest priority.
    // 3. endpoints have same [network, region] label but different [zone, subzone] label with the client proxy have the third highest priority.
    // 4. endpoints have same [network] but different [region, zone, subzone] labels with the client proxy have the fourth highest priority.
    // 5. all the other endpoints that does not match any labels have the same lowest priority.
    //

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
 
-    // Optional: only failover or distribute can be set.
+    // Optional: only one of distribute, failover or topologyKeys can be set.
     // Explicitly specify the region traffic will land on when endpoints in local region becomes unhealthy.
     // Should be used together with OutlierDetection to detect unhealthy endpoints.
     // Note: if no OutlierDetection specified, this will not take effect.
     repeated Failover failover = 2;
 
+    // TopologyKeys is a list of topology keys in the order provided, used to sort endpoints to do priority load balancing.
+    // This is to support traffic fail over across different group of endpoints.
```
nrjpoddar, line 1083:
```suggestion
    // This is to support traffic failover across different groups of endpoints.
```

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: // networks:
//   network1:
//     endpoints:
//     - fromRegistry: registry1 #must match kubeconfig name in Kubernetes secret
//     - fromCidr: 192.168.100.0/22 #a VM network for example
//     gateways:
//     - registryServiceName: istio-ingressgateway.istio-system.svc.cluster.local
//       port: 15443
//       locality: us-east-1a
//     - address: 192.168.100.1
//       port: 15443
//       locality: us-east-1a

The above config can be used to get the network of an endpoint

networking/v1alpha3/destination_rule.proto
```diff
@@ -1067,18 +1067,54 @@ message LocalityLoadBalancerSetting{
...
+    // Suppose there are total N keys specified:
```
nrjpoddar, line 1084:
```suggestion
    // Suppose there are total N labels specified:
```

howardjohn:
`failoverPriority` LGTM

nrjpoddar (approved):
Thanks @hzxuzhonghu for incorporating all the changes.

PR #2056, opened 2021-07-16
Release 1.11

zhlsunshine:
Part of [issue#33913](https://github.com/istio/istio/issues/33913)


jwendell (changes requested):
This commit "[release-1.11] Automated branching step 4" should not be part of this PR

zhlsunshine:
Hi @brian-avery, as we have discussed in slack, it seems that there is no any higher release version than release-1.6 under the directory of `releaselocks` to record the API change, so I think we need to add a folder release-1.11 for it. However, the `build_api` test is failed, could you please help to check it when you are available? Thanks in advance!

zhlsunshine:
> This commit "[release-1.11] Automated branching step 4" should not be part of this PR

done

zhlsunshine:
It seems that there is no any releaselock folder for istio/api since release-1.6. It may be unnecessary to suddenly new a release-1.11. So I will close this PR and please let's know if it's necessary.

PR #2058, opened 2021-07-19
Add global HTTP retry policy API in MeshConfig

nrjpoddar:
Addresses: https://github.com/istio/istio/issues/6861


mesh/v1alpha1/config.proto
```diff
@@ -864,8 +865,16 @@ message MeshConfig {
   // If not set, the NormalizationType.DEFAULT configuration will be used.
   ProxyPathNormalization path_normalization = 61;
 
+  // Configure the default HTTP retry policy.
+  // The default number of retry attempts is set at 2 for these errors:
+  //   "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes".
+  // Setting the number of attempts to -1 disables retry policy globally.
```
howardjohn, line 871: `-1`? Doesn't (or shouldn't) `0` work?

nrjpoddar: I am trying to be compatible with this: https://github.com/istio/istio/blob/dcf81269280c87f096fa25a2f94e35a8e2d2396e/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go#L71

howardjohn: its checking `<=` though so I would think `0` works?

nrjpoddar: Jeez, yes, I saw it as < 0 for some reason. I will fix it.

nrjpoddar: Done

nrjpoddar:
@howardjohn any other comment/concerns?

mesh/v1alpha1/config.proto
```diff
@@ -864,8 +865,16 @@ message MeshConfig {
   // If not set, the NormalizationType.DEFAULT configuration will be used.
   ProxyPathNormalization path_normalization = 61;
 
+  // Configure the default HTTP retry policy.
+  // The default number of retry attempts is set at 2 for these errors:
+  //   "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes".
+  // Setting the number of attempts to 0 disables retry policy globally.
+  // This setting can be overriden on a per-host basis using the Virtual Service
+  // API.
+  istio.networking.v1alpha3.HTTPRetry default_http_retry_policy = 62;
```
ramaraochavali, line 874: Should we wrap this with some other top level message like "HttpRoutePolicy" or some thing similar so that later we can expand it to others like Cors, Fault injection etc if needed?

nrjpoddar: That's a great suggestion but I'm not sure what's the best path forward here. We already have following fields all rooted at meshconfig belong to HTTP:

* h2UpgradePolicy
* pathNormalization

I also don't know if setting Fault injection or CORS make sense on a global level to be honest as they are all by default off.

ramaraochavali: > I also don't know if setting Fault injection or CORS make sense on a global level to be honest as they are all by default off.

Those are just some examples I could think from top of my head. I meant to say other features.

nrjpoddar: Got it. This issue of global defaults affects HTTPRoute knobs in VS as there's no concept of merging (or default) with VS resources unlike DRs. Only other field that I found where a global default would be useful is timeout which is currently disabled by default.

I will wait for others to chime in if creating a top level field like http_routing is preferred I can do that. From a conceptual point of view I don't think having defaults for routing has concrete use cases. Adding timeout field in VS was a weird choice to begin with I think.

nrjpoddar:
@howardjohn can you take a look again?

mesh/v1alpha1/config.proto
```diff
@@ -864,8 +865,16 @@ message MeshConfig {
   // If not set, the NormalizationType.DEFAULT configuration will be used.
   ProxyPathNormalization path_normalization = 61;
 
+  // Configure the default HTTP retry policy.
```
howardjohn, line 868: Setting PerTryTimeout here may be a bit odd, since it has interactions with VirtualService.http.route.timeout.

Basically route.timeout is the total timeout, but PerTryTimeout is each attempt. So if we set PerTryTimeout=1s globally, and user sets timeout=10s, they are going to be pretty confused why they get timeout after 1s I think...

We can always add validation rejecting this if we don't want it, or just accept the caveat.

howardjohn (approved):

mesh/v1alpha1/config.proto
nrjpoddar: Agree. Let me add a clarifying comment that only setting attempts is supported at this point.

mesh/v1alpha1/config.proto
```diff
@@ -865,8 +866,18 @@ message MeshConfig {
...
+  // The default number of retry attempts is set at 2 for these errors:
+  //   "connect-failure,refused-stream,unavailable,cancelled,retriable-status-codes".
+  // Setting the number of attempts to 0 disables retry policy globally.
+  // This setting can be overriden on a per-host basis using the Virtual Service
+  // API.
+  // All settings in the retry policy except `perTryTimeout` can currently be
+  // configured globally via this field.
```
nrjpoddar, line 876: @howardjohn added a clarification around `perTryTimeout` PTAL?

nrjpoddar:
@howardjohn I added this setting in MeshConfig and not ProxyConfig as other HTTP settings like `normalizePath` etc were under MeshConfig. Should I move it in light of all the migration plans w.r.t Mesh and Proxy config?

nrjpoddar:
@linsun @smawson @louiscryan @brian-avery mind giving a second approval?

linsun (approved):
LGTM

costinm:
It seems my comment was lost ( I used 'reply' in the email message ). For reference, it was:

What is the goal for this setting ? To remain experimental / alpha for a very long time and be eventually moved to a proper API ( networking or ProxyConfig ) ? The goal was to shrink MeshConfig - not to add more.

And the usual question I ask for all MeshConfig changes: is this really something only the istio system owner can set for the entire mesh, and there is no  use case of setting it per namespace or workload ? And since we still use operator  to change MeshConfig - is this the kind of setting that would require admin to re-run istioctl install ( vs. updating the ProxyConfig or a networking CR ) ?  

--- 

I don't have any objection to the PR - like many other experimental things in MeshConfig, as long as users don't expect it 
to be a stable API and is used only for experimentation or as an 'istio install' option.

nrjpoddar:
The stable API for this exists in VirtualService so this belongs in MeshConfig as global mesh-wide setting. For overrides per host service owners can use existing VirtualService API. Basically, this is a class of config that I believe belongs in MeshConfig or equivalent mesh-wide configuration if and when the new API arrives.

jacob-delgado:
@zxh326 Yes

PR #2061, opened 2021-07-20
Add new guidelines to API compatibility

howardjohn:


GUIDELINES.md
```diff
@@ -198,6 +198,22 @@ protos.
         }
         ```
 
+- **Do not** remove fields.
```
howardjohn, line 201: This is a new guideline, although it has been de-facto standard, since in ~1.5+ we started validating unknown fields

GUIDELINES.md
```diff
@@ -198,6 +198,22 @@ protos.
...
+
+    - This is backwards compatible for protobuf, but not for CRDs which have strict validation preventing unknown fields.
+
+- **Do not** make validation stricter than in previous versions.
```
howardjohn, line 205: This is a new guideline, and the core motivation for this PR. Although it has been attempted to be enforced in PR review, there is lack of consensus that this is a requirement. I want to solidify our position one way or the other (preferably, the way stated in the docs), so there is consensus in the community.

GUIDELINES.md
```diff
@@ -226,15 +242,21 @@ message Gateway {
 
 Istio APIs should use a simple versioning strategy based on
 major versions and releases, such as `v1alpha`, `v2beta`, or
-`v3`. Within each release, there should not be any breaking
-change to released features, such as changing the type of
-a field type, renaming a field, or changing a field number.
-Breaking changes are allowed between different releases,
-such as `v1alpha1` and `v1alpha2`.
+`v3`. Due to current limitations in Istio's CRD versioning, namely
```
howardjohn, line 245: Just updating doc to match reality

GUIDELINES.md
therealmitchconnors: Would renaming a field constitute a removal?  Does this apply to all stability levels equally?

GUIDELINES.md
```diff
@@ -198,6 +198,22 @@ protos.
         }
         ```
 
+- **Do not** remove fields.
+
+    - This is backwards compatible for protobuf, but not for CRDs which have strict validation preventing unknown fields.
```
therealmitchconnors, line 203: This is optional for CRDs, and can be turned off.  Additionally, if the field has a logical successor (such as label selector map replaced by true label selectors), we can potentially up convert old resources to the new format, as long as the new format is at least as expressive than the original.  See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

GUIDELINES.md
therealmitchconnors: +1 for this requirement

GUIDELINES.md
howardjohn: Renaming is already discussed on L139.

Yes, this applies to all stability levels

GUIDELINES.md
howardjohn: > we can potentially up convert old resources to the new format, as long as the new format is at least as expressive than the original

I agree in theory, but we do not have the technical capability to do that in the project today though, right? See L245

GUIDELINES.md
kyessenov: Also, mention defaults. They are considered part of API for practical purpose (e.g. meshconfig).

GUIDELINES.md
howardjohn: And I agree it can be turned off, but I see no reason to turn it off to remove a field. Is there a valid use case for doing that? If not, I think it makes sense to leave this as is - we can always issue exceptions in exceptional cases. WDYT?

GUIDELINES.md
kyessenov: I think the reality should match the goals in this case. It's not realistic to have all versions be the same all the time.

GUIDELINES.md
howardjohn: Defaults are a tricky one. I feel like they are in a different class of incompatibility breakage. Here we are mostly discussing a configuration in one version not being valid in a new version.

Defaults are a bit more subtle - they configuration is valid, but behaves differently. The same really applies beyond just defaults, but in how functionality (API driven or otherwise) is implemented. For example, https://istio.io/latest/blog/2021/upcoming-networking-changes/.

I think a blanket statement that "validation cannot become stricter" is a safe claim to make, but I am not sure that "defaults (or behavior in general) can never change" is something we are willing to commit to. We should absolutely limit these and use extreme caution when making them, but its a bit less strict.

What do you think? I can add a separate section for defaults, a bit worried that may be more controversial than the rest of the change though

GUIDELINES.md
howardjohn: I think there is general agreement that we would like that, but no implementable design to actually do this unfortunately. Would be great if someone came up with one though.

GUIDELINES.md
kyessenov: I think the right way to signal breaking changes in defaults is by bumping the version number. Defaults are part of the API, not just behavior details. What you are saying is that we reserve the right to perform breaking API changes with respect to defaults, and that needs to be mentioned.

GUIDELINES.md
kyessenov: I understand the limitations. It's odd to canonicalize the limitations since it signals we will never actually add versions.

howardjohn: I was worried about the current wording to mislead developers that they can actually make breaking changes (via version bump) - when this is not the case for the foreseeable future. I didn't intend the new wording to be set in stone - if we lift the limitation we can update it?

kyessenov: Guidelines exist so that we don't have to keep changing our position on things. If we add and then remove guidelines, it'll lead to conflict when some things are no longer OK by guidelines.

howardjohn: Sure, but I am not sure how it is useful to say "You can make breaking changes between apiVersions", when that is not actually possible. The new guideline does indicate this is only due to "current limitations"

GUIDELINES.md
therealmitchconnors: I'm comfortable leaving these topics as something we'll tackle when we have sufficient need.  Until then, don't remove fields.

GUIDELINES.md
jasonwzm: One option to make defaults more transparent to users is to implement defaulting in CRDs: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

That way it is clearer contract on what the default value is.

GUIDELINES.md
```diff
@@ -226,15 +242,21 @@ message Gateway {
...
 Deprecating a feature in an API release is allowed by following
 the applicable deprecation process. The reason to allow
 deprecation of individual features in a release is that it is
 significantly cheaper and simpler for everyone involved. In
 practice, it works out much better than deprecating an entire
 API version.
+
+## Exceptions
+
+Many of the guidelines above are related to limiting breaking changes.
```
nrjpoddar, line 258:
```suggestion
Many of the guidelines above are related to limiting backward incompatible changes.
```

GUIDELINES.md
nrjpoddar, line 205: This is interesting and kind of agree in principle but can lead to us not being able to patch bugs e.g. many times istiod is missing validation for various fields in our APIs. So if it's once released can you safely add validation without violating this guideline?

ramaraochavali: I agree with @kyessenov . Defaults in general are considered part of API whether we explicitly specify in CRD or not. 
Regarding validations, I am not fully sure if we want to bump api version number for every small validation bug(enforcing new validation). It would be good to be compatible, but if we find a bug in validation code that crashes Istiod or makes Istiod behave differently because of some invalid config - I think we should be able to fix that validation bug. We could take approach similar to what Envoy does, advertise it in minor behaviour change in the release notes and possibly runtime guard it so that people can revert to old during upgrade with a clear deadline for removing that runtime guard.

howardjohn: I think the fix there would be to make Istiod crash, not to make the validation stricter. I also am not sure bumping api version is applicable here (at least currently, as we do not properly support multiple api versions anyways).

howardjohn: I think no, which is why validation should be strict in the initial release. Warnings can always be added.

nrjpoddar: That's an ideal state which is like saying we will write perfect code going forward.

I would advocate that if you need to add stricter validation, you can turn it on by default with an option to disable it for users who can be affected. Since this validation is hopefully better and users should migrate their imperfect configuration we can further provide tooling to assist them during this process.

therealmitchconnors (approved):

GUIDELINES.md
hzxuzhonghu: Do we allow deprecate a field for several releases and remove it?

GUIDELINES.md
hzxuzhonghu: If unreasonable api settings cause it does not take effect, don't we allow stricter validation to prevent it?

GUIDELINES.md
howardjohn: No, because if the user has a cr referencing that and user upgrades istio suddenly it will break. This has happened in the past with h2upgradepolicy and knative

GUIDELINES.md
howardjohn: No, because even if a user has a config that does nothing, an upgrade of Istio may now suddenly break their ci/cd pipelines. It's better to do nothing than to break.

If it's something that has always crashed istiod or something where no one can plausibly have the setting configured, then I think it's reasonable. But in general it's not safe

hzxuzhonghu:
@howardjohn There are exceptions in k8s api change

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

```
If your change does not meet these criteria, it is not considered compatible, and may break older clients, or result in newer clients
 causing undefined behavior. Such changes are generally disallowed, though exceptions have been made in extreme cases 
(e.g. security or obvious bugs).
```

hzxuzhonghu:
I think we should also allow exceptions for some cases.

howardjohn:
> I think we should also allow exceptions for some cases.


Done

hzxuzhonghu (approved):

PR #2076, opened 2021-08-09
update cluster label docs

linsun:
based on comments from https://github.com/istio/istio/issues/34366

PTAL and suggest changes.  The main point is to not confuse users to label their workloads with this `"topology.istio.io/cluster"` label.

cc @stevenctl @nmittler


label/labels.yaml
```diff
@@ -123,11 +123,10 @@ labels:
 
   - name: topology.istio.io/cluster
     featureStatus: Alpha
-    description: A workload label that indicates the name of the cluster that contains the
-      workload. This is typically configured during control plane installation, using either
-      an auto-generated or admin-specified value. Setting this allows workload selection by
-      cluster. For example, a service owner could create a DestinationRule containing
-      a subset per cluster and then use these subsets to control traffic flow to each
+    description: This label indicates the name of a cluster for its workloads, which can be
+      configured during Istio control plane installation. Setting this allows workload selection by
```
stevenctl, line 127: > Setting this

I think that's the misleading part. Users don't set this label, and we don't inject this label onto actual objects in apiserver. This label exists only in istiod's internal state, so we should write this as "Usiing this label for selection in Istio CR's" or similar.

linsun: Ah, good point. Updated!

stevenctl (approved):
lgtm - it might be good to explicitly set the expectation that users _won't_ see this on their Pods, but I defer

linsun:
@howardjohn @brian-avery comments?

label/labels.yaml
```diff
@@ -123,11 +123,10 @@ labels:
 
   - name: topology.istio.io/cluster
     featureStatus: Alpha
-    description: A workload label that indicates the name of the cluster that contains the
-      workload. This is typically configured during control plane installation, using either
-      an auto-generated or admin-specified value. Setting this allows workload selection by
-      cluster. For example, a service owner could create a DestinationRule containing
-      a subset per cluster and then use these subsets to control traffic flow to each
+    description: This label indicates the name of a cluster for its workloads, which can be
```
nmittler, line 126: I'm not sure that the new text accomplishes your goal ... it's not any more clear WRT not setting the label.

Maybe something like:

```
A label applied to a workload internally that identifies the Kubernetes cluster containing
the workload. The cluster ID is specified during Istio installation in each cluster. It should
be noted that this is only used internally within Istio and is not an actual label on workload
pods. If a pod contains this label, it will be overridden by Istio internally with the
cluster ID specified during Istio installation. 

This label provides a way to select workloads by cluster when using
DestinationRules. For example, a service owner could create a DestinationRule
containing a subset per cluster and then use these subsets to control traffic flow to
each cluster independently.
```

label/labels.yaml
```diff
@@ -123,12 +123,13 @@ labels:
...
   - name: topology.istio.io/cluster
     featureStatus: Alpha
-    description: A workload label that indicates the name of the cluster that contains the
-      workload. This is typically configured during control plane installation, using either
-      an auto-generated or admin-specified value. Setting this allows workload selection by
-      cluster. For example, a service owner could create a DestinationRule containing
-      a subset per cluster and then use these subsets to control traffic flow to each
-      cluster independently.
+    description: This label is applied to a workload internally that identifies the Kubernetes cluster containing
+      the workload. The cluster ID is specified during Istio installation for each cluster via `values.global.multiCluster.clusterName`.
```
linsun, line 127: @nmittler thank you for the suggestion - I like it!  Question - I added configuration for cluster ID to help users to connect the dots from their install config to cluster ID, does this look correct to you?

nmittler: LGTM

nmittler (approved):

label/labels.yaml
linsun: Thank you @nmittler

linsun:
@howardjohn @brian-avery could i get an approval from one of you?  Thanks

stevenctl (approved):

hzxuzhonghu:
The description is really what we do today. My question may not relate, why did we expose a label that is intended to be used internally rather than users.

linsun:
That is a good q @hzxuzhonghu!  I remember I questioned about exposing it early on too.  Perhaps @nmittler @stevenctl has some reasonings.

howardjohn (approved):

PR #2080, opened 2021-08-17
Add ProxyConfig CRD

Monkeyanator:
[ProxyConfig CRD RFC](https://docs.google.com/document/d/1Ari-c-nvIr7kMy8be8zywltp0lQSdc9ktsB13ixGQqg/edit?usp=chrome_omnibox&ouid=101546357587773542516)

There are more fields that are proposed to make it into the CRD (`includeOutboundPorts`, `excludeOutboundPorts`, etc) but the idea is to promote the absolute minimum number of fields into the CRD initially.

cc @costinm


networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+// To set the mesh wide defaults, create a `ProxyConfig` resource in the `istio-system` namespace.
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1beta1
+// kind: ProxyConfig
+// metadata:
+//   name: my-proxyconfig
+//   namespace: istio-system
+// spec:
+//   discoveryAddress: https://example.com
```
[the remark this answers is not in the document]
howardjohn: Although I guess setting it to the default is possibly confusing to users - they may think they need to set it.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/wrappers.proto";
+import "networking/v1beta1/sidecar.proto";
+
+package istio.networking.v1beta1;
+
+option go_package="istio.io/api/networking/v1beta1";
+
+// ProxyConfig defines variables for individual Envoy instances. This can be configured on a per-workload basis, a per-revision
```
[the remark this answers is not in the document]
howardjohn: Also I would indicate its not a required resource - there are also defaults

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+//   name: per-workload-proxyconfig
+//   namespace: istio-system
+// spec:
+//   workloadSelector:
+//     labels:
+//       app: ratings
+//   discoveryAddress: https://example.com
+// ```
+//
+// If multiple of these are configured they are merged with per field semantics; the field set in annotation will fully replace the field from mesh config defaults.
```
howardjohn, line 68: "annotation" seems out of place here

howardjohn: In current impl proxy_metadata is a deep merge actually

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+  string discovery_address = 2;
+
+  // The number of worker threads to run.
+  // If unset, this will be automatically determined based on CPU requests/limits.
+  // If set to 0, all cores on the machine will be used.
+  // Default is 2 worker threads.
+  google.protobuf.Int32Value concurrency = 3;
+
+  // Path to the proxy bootstrap template file
+  string proxy_bootstrap_template_path = 4;
```
howardjohn, line 88: I am not sure we need this in the initial revision

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+  // Additional environment variables for the proxy.
+  // Names starting with `ISTIO_META_` will be included in the generated bootstrap and sent to the XDS server.
+  map<string,string> proxy_metadata = 5;
+
+  // The amount of time allowed for connections to complete on proxy shutdown.
+  // On receiving `SIGTERM` or `SIGINT`, `istio-agent` tells the active Envoy to start draining,
+  // preventing any new connections and allowing existing connections to complete. It then
+  // sleeps for the `termination_drain_duration` and then kills any remaining active Envoy processes.
+  // If not set, a default of `5s` will be applied.
+  google.protobuf.Duration  termination_drain_duration = 6;
```
howardjohn, line 99: fwiw its likely we will want to change the behavior of this (https://github.com/envoyproxy/envoy/issues/17056)

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+
+  // The unique identifier for the [service mesh](https://istio.io/docs/reference/glossary/#service-mesh)
+  // All control planes running in the same service mesh should specify the same mesh ID.
+  // Mesh ID is used to label telemetry reports for cases where telemetry from multiple meshes is mixed together.
+  string mesh_id = 7;
```
howardjohn, line 104: If it must in the same for all pods then it doesn't seem to fit in ProxyConfig

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+
+  // Boolean flag for enabling/disabling the holdApplicationUntilProxyStarts behavior.
+  // This feature adds hooks to delay application startup until the pod proxy
+  // is ready to accept traffic, mitigating some startup race conditions.
+  // Default value is 'false'.
+  google.protobuf.BoolValue hold_application_until_proxy_starts = 8;
```
howardjohn, line 110: IMO this is not beta level. I can dig up some more discussion on this if needed; the original plan was to keep it as experimental indefinitely until k8s adds native support

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+//
+// Note: fields in ProxyConfig are not dynamically configured; changes will require restart of workloads to take effect.
+message ProxyConfig {
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `ProxyConfig` configuration should be applied.
+  WorkloadSelector workload_selector = 1;
+
+  // Address of the discovery service exposing xDS with mTLS connection.
+  // The inject configuration may override this value.
+  string discovery_address = 2;
```
howardjohn, line 79: ca_address seems like a good addition as well

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+// If multiple of these are configured they are merged with per field semantics; the field set in annotation will fully replace the field from mesh config defaults.
+// This is different than a deep merge provided by protobuf.
+//
+// Note: fields in ProxyConfig are not dynamically configured; changes will require restart of workloads to take effect.
+message ProxyConfig {
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `ProxyConfig` configuration should be applied.
+  WorkloadSelector workload_selector = 1;
+
+  // Address of the discovery service exposing xDS with mTLS connection.
```
howardjohn, line 77: The connection is not mTLS (except on VMs) anymore. Just TLS+JWT

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+  // This feature adds hooks to delay application startup until the pod proxy
+  // is ready to accept traffic, mitigating some startup race conditions.
+  // Default value is 'false'.
+  google.protobuf.BoolValue hold_application_until_proxy_starts = 8;
+
+  // The PEM data of the extra root certificates for workload-to-workload communication.
+  // This includes the certificates defined in MeshConfig and any other certificates that Istiod uses as CA.
+  // The plugin certificates (the 'cacerts' secret), self-signed certificates (the 'istio-ca-secret' secret)
+  // are added automatically by Istiod.
+  repeated string ca_certificates_pem = 9;
```
howardjohn, line 116: We might want to consider making this more nested - there have been various proposals to add a lot of cert related options to proxy config that may make that more future proof. We don't necesarily need to block on adding those, but should consider what the future API could look like.

@shankgan @jacob-delgado

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+//   namespace: istio-system
+// spec:
+//   workloadSelector:
+//     labels:
+//       app: ratings
+//   discoveryAddress: https://example.com
+// ```
+//
+// If multiple of these are configured they are merged with per field semantics; the field set in annotation will fully replace the field from mesh config defaults.
+// This is different than a deep merge provided by protobuf.
```
howardjohn, line 69: Can we describe what happens with annotation, `meshConfig.defaultConfig`, and `ProxyConfig`?

networking/v1beta1/proxy_config.proto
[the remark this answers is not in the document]
costinm: Also not for 'envoy instances' - but for workloads. It applies to proxyless grpc and probably other mesh options.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+import "google/protobuf/wrappers.proto";
+import "networking/v1beta1/sidecar.proto";
+
+package istio.networking.v1beta1;
+
+option go_package="istio.io/api/networking/v1beta1";
+
+// ProxyConfig defines variables for individual Envoy instances. This can be configured on a per-workload basis, a per-revision
+// basis, and mesh-wide.
+// To set the mesh wide defaults, create a `ProxyConfig` resource in the `istio-system` namespace.
```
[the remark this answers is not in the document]
costinm: How about removing this part - behavior of istio-system or root namespace can be documented independently of the spec of the ProxyConfig CRD.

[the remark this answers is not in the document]
costinm: If you really want - 'defaults may be added based on the mesh config and implementation'.

networking/v1beta1/proxy_config.proto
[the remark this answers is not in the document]
costinm: Agreed, probably not a good example. 
discoveryAddress is for advanced users who want to use an external XDS server or customize - should not be used by most users.

I'm starting to agree that maybe we should not make it part of the CRD...

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+// apiVersion: networking.istio.io/v1beta1
+// kind: ProxyConfig
+// metadata:
+//   name: my-proxyconfig
+//   namespace: istio-system
+// spec:
+//   discoveryAddress: https://example.com
+// ```
+//
+// This can be configured on a per-revision basis by configuring the `istio.io/rev` label on the `ProxyConfig` resource. For example:
```
costinm, line 39: Less is more...

This is a super advanced feature, it may be documented in another place but we don't need to specify it in each resource.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,117 @@
...
+// ```
+//
+// This can also be configured per-workload basis by setting the `workloadSelector` field on the `ProxyConfig` resource. For example:
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1beta1
+// kind: ProxyConfig
+// metadata:
+//   name: per-workload-proxyconfig
+//   namespace: istio-system
```
costinm, line 60: namespace: example

( istio-system is special )

networking/v1beta1/proxy_config.proto
costinm: Agreed, remove annotations, we can also be a bit light on merge - right now the only merge is between the exact match and the defaults. 

We can say 'fields can be merged or overridden by mesh wide settings'.

I don't think this PR is the right place to add a new feature of merging multiple ProxyConfig that match. We are promoting subset of current behavior, not adding more.

networking/v1beta1/proxy_config.proto
costinm: I think the agreement was that if ProxyConfig CR is present, it will be used and the annotation and old meshconfig.defaultConfig ignored ( i.e. using the CR is an opt-in to the new API )

networking/v1beta1/proxy_config.proto
costinm: That's an implementation detail, I think we can still offer client certs if asked by server and should support other auth methods as well. 

Also: I don't think inject configuration should override this value ( it is not the case with the current impl either AFAIK ). 

We should add warnings that this is an advanced feature that replaces the install default. And we might discuss again if it should be in the public API ( instead of internal )

networking/v1beta1/proxy_config.proto
costinm: If we keep discovery_address - yes.

networking/v1beta1/proxy_config.proto
costinm: +1 - only the fields we are absolutely sure we want to expose and support as user configurable.

networking/v1beta1/proxy_config.proto
costinm: agreed. it's mostly internal.

networking/v1beta1/proxy_config.proto
costinm: The question is if configuring root certificates to trust belongs to ProxyConfig - DestinationRule specifies per destination root certs. Sidecar may be a better place for this, if not a security API.

Also is there a use case to configure this at workload or namespace granularity ? It might be something to keep mesh-wide, or even to leave it outside of Istio ( CertManager or other systems may have their own ways to propagate the roots, and if users integrate we should not interfere or pretend we can change that ).

costinm (commented):
My suggestion is to reduce this to the smallest, most conservative subset - to get the template in and start working on the controller and impl. 

We can keep adding fields one by one - the impl will not change ( i.e. we can implement it so that we parse using the existing ProxyConfig any unknown fields )

networking/v1beta1/proxy_config.proto
howardjohn: I generally agree. I think this is a broader issue - there are user facing configs for a proxy that should be configured in this api at the granularity we provide.

Then there is the implementation details that we configure a proxy nearly exclusively by "PROXY_CONFIG". I think we need to clearly separate the two - the internal one should be a much larger superset

networking/v1beta1/proxy_config.proto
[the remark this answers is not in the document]
shankgan: is it also valid to qualify and say "init-time/install-time" variables for workloads?  variables sounds a little ambigious?

networking/v1beta1/proxy_config.proto
shankgan: this particular field is currently used to configure all additional trustAnchors for mesh ISTIO_MUTUAL mTLS. There is some effort to create a WorkloadCertificate CRD separately, to configure the cert params - we could potentially add it there, instead of here.

networking/v1beta1/proxy_config.proto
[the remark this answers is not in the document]
Monkeyanator: Removing from initial iteration of the CRD since there's still some discussion to be had around this

networking/v1beta1/proxy_config.proto
Monkeyanator: Removed.

networking/v1beta1/proxy_config.proto
Monkeyanator: Removed.

networking/v1beta1/proxy_config.proto
Monkeyanator: I thought that the most specific match would override the mesh wide settings rather than mesh wide settings overwriting fields?

networking/v1beta1/proxy_config.proto
Monkeyanator: What do you mean here? Agreed we shouldn't specify on each resource, this is just documentation showing how it can be done.

networking/v1beta1/proxy_config.proto
costinm: That's the intent - but for some settings we may make different choices. For example if we know a setting will break the app, or security reasons. Merging is very hard to understand and complicated - I wish we didn't expose users too much to it...

networking/v1beta1/proxy_config.proto
costinm: So let's keep it out of the CRD, it's an 'internal only' setting for now.

I don't even know if WorkloadCertificate CRD is needed - in many cases it might be desirable to let the vendors and integrations take care of this, without imposing requirements that may be hard to meet. MeshConfig and the internal-only settings work for now.

networking/v1beta1/proxy_config.proto
costinm: Do we have such doc on Gateway or Sidecar ? If not, why should we add it here ? 

I don't know if the feature of targetting a CR to a specific revision is 'beta' - there are some advanced use cases, and we may have a wiki or doc somewhere - but this is not the right place.

Also it is not clear how this will interact with tags - we want people to think in terms of tags, while Istiod doesn't know its own tags. That's the real reason to be cautious.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,88 @@
+// Copyright 2017 Istio Authors
```
nrjpoddar, line 1: Old header...

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,88 @@
...
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/wrappers.proto";
+import "networking/v1beta1/sidecar.proto";
+
+package istio.networking.v1beta1;
+
+option go_package="istio.io/api/networking/v1beta1";
+
+// ProxyConfig defines variables for individual Envoy instances. This can be configured on a per-workload basis, a per-revision
```
nrjpoddar, line 25:
```suggestion
// ProxyConfig defines variables for individual Envoy proxy instances. This can be configured on a per-workload basis, a per-revision
```

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,88 @@
...
+// This is different than a deep merge provided by protobuf.
+//
+// Note: fields in ProxyConfig are not dynamically configured; changes will require restart of workloads to take effect.
+message ProxyConfig {
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `ProxyConfig` configuration should be applied.
+  WorkloadSelector workload_selector = 1;
+
+  // The number of worker threads to run.
+  // If unset, this will be automatically determined based on CPU requests/limits.
```
nrjpoddar, line 80: Is it requests or limits?

networking/v1beta1/proxy_config.proto
Monkeyanator: Fixed, thanks. Copied from Sidecar proto :)

Monkeyanator: Fixed, thanks. Copied from Sidecar proto :)

networking/v1beta1/proxy_config.proto
Monkeyanator: Looks like it actually is both https://github.com/istio/istio/blob/master/pkg/kube/inject/inject.go#L789

networking/v1beta1/proxy_config.proto
Monkeyanator: Agree it makes sense to leave it out of docs for first iteration, removed.

networking/v1beta1/proxy_config.proto
Monkeyanator: Added.

Monkeyanator:
Addressed comments on the documentation and reduced the CRD to 3 initial fields: (1) the workload selector, (2) concurrency, and (3) proxy_metadata.

costinm (approved):

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,74 @@
...
+// Note: fields in ProxyConfig are not dynamically configured; changes will require restart of workloads to take effect.
+message ProxyConfig {
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `ProxyConfig` configuration should be applied.
+  WorkloadSelector workload_selector = 1;
+
+  // The number of worker threads to run.
+  // If unset, this will be automatically determined based on CPU requests.
+  // If set to 0, all cores on the machine will be used.
+  // Default is 2 worker threads.
```
howardjohn, line 68: How can we have a default and "if unset" not be the same? Its a bit confusing what the behavior would be here

Monkeyanator: Updated, looks like when it's unset the default is 2, when set to 0 it will use requests and limits to choose a value that uses all cores.

jacob-delgado (approved):
We had talked about getting in XFF and XFCC in as well, but that can wait for a follow up PR. It'd be great if we could get this into 1.12. Branch cut is 1 month away.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+import "google/protobuf/duration.proto";
+import "google/protobuf/wrappers.proto";
+import "networking/v1beta1/sidecar.proto";
+
+package istio.networking.v1beta1;
+
+option go_package="istio.io/api/networking/v1beta1";
+
+// `ProxyConfig` defines configuration for individual workloads. `ProxyConfig` can be configured on a per-workload basis, a per-revision
+// basis, a per-namespace basis, or mesh-wide. `ProxyConfig` is not a required resource; there are default values in place.
```
mandarjog, line 26: Merge semantics should be defined here, or pointer to a place where these are defined.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+//   namespace: example
+// spec:
+//   workloadSelector:
+//     labels:
+//       app: ratings
+//   concurrency: 0
+// ```
+//
+//
+// If a `ProxyConfig` CRD is defined that matches a workload it will take precedence over the annotation for that workload.
```
mandarjog, line 56: Precedence is not descriptive enough.
Does it replace at the top level or merge ?

costinm:
On Mon, Sep 20, 2021, 09:37 mandarjog ***@***.***> wrote:

>
> In networking/v1beta1/proxy_config.proto
> <https://github.com/istio/api/pull/2080#discussion_r712334205>:
>
> > +//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
> +//   See the License for the specific language governing permissions and
> +//   limitations under the License.
> +
> +syntax = "proto3";
> +
> +import "google/protobuf/duration.proto";
> +import "google/protobuf/wrappers.proto";
> +import "networking/v1beta1/sidecar.proto";
> +
> +package istio.networking.v1beta1;
> +
> +option go_package="istio.io/api/networking/v1beta1";
> +
> +// `ProxyConfig` defines configuration for individual workloads. `ProxyConfig` can be configured on a per-workload basis, a per-revision
> +// basis, a per-namespace basis, or mesh-wide. `ProxyConfig` is not a required resource; there are default values in place.
>
> Merge semantics should be defined here, or pointer to a place where these
> are defined
>


------------------------------
>
> In networking/v1beta1/proxy_config.proto
> <https://github.com/istio/api/pull/2080#discussion_r712335647>:
>
> > +//
> +// ```yaml
> +// apiVersion: networking.istio.io/v1beta1
> +// kind: ProxyConfig
> +// metadata:
> +//   name: per-workload-proxyconfig
> +//   namespace: example
> +// spec:
> +//   workloadSelector:
> +//     labels:
> +//       app: ratings
> +//   concurrency: 0
> +// ```
> +//
> +//
> +// If a `ProxyConfig` CRD is defined that matches a workload it will take precedence over the annotation for that workload.
>
> Precedence is not descriptive enough.
> Does it replace at the top level or merge ?
>
Current plan is to replace, not merge. Few reasons - probably most
important is that proxyconfig can have rbac control.

howardjohn:
> Current plan is to replace, not merge. Few reasons - probably most important is that proxyconfig can have rbac control.

Won't replace will break external control plane that relies on PROXY_METADATA to pass control plane connection info?

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+  WorkloadSelector workload_selector = 1;
+
+  // The number of worker threads to run.
+  // If unset, defaults to 2. If set to 0, this will be configured to use all cores on the machine using
+  // CPU requests and limits.
+  google.protobuf.Int32Value concurrency = 2;
+
+  // Additional environment variables for the proxy.
+  // Names starting with `ISTIO_META_` will be included in the generated bootstrap and sent to the XDS server.
+  map<string,string> proxy_metadata = 3;
```
smawson, line 72: This is already ProxyConfig, do we need to prefix this with proxy_ as well? Or can we just call this metadata? Or maybe environment_variables?

mandarjog:
All Istio apis need to deal with replace vs merge.
Replace gives you full control (removal of keys) and merge is easier to use.

I think a “smart merge” like we have in telemetry api / new mesh config merge  is what we need here. This normally behaves like merge but it can protect certain keys that should not be messed with.

And we need to specify it as part of the api :-)

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+// `ProxyConfig` defines configuration for individual workloads. `ProxyConfig` can be configured on a per-workload basis, a per-revision
+// basis, a per-namespace basis, or mesh-wide. `ProxyConfig` is not a required resource; there are default values in place.
+//
+// To set the mesh wide defaults, create a `ProxyConfig` resource in the root namespace.
+//
+// ```yaml
+// apiVersion: networking.istio.io/v1beta1
+// kind: ProxyConfig
+// metadata:
+//   name: my-proxyconfig
```
linsun, line 34: does this need to be named default?  What if user puts two proxy config in istio-system ns?

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+// spec:
+//   workloadSelector:
+//     labels:
+//       app: ratings
+//   concurrency: 0
+// ```
+//
+//
+// If a `ProxyConfig` CRD is defined that matches a workload it will take precedence over the annotation for that workload.
+// Furthermore, options configured in a `ProxyConfig` CRD matching a workload will take precedence over mesh wide defaults.
```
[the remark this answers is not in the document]
linsun:
```suggestion
// Furthermore, options configured in a `ProxyConfig` CR matching a workload will take precedence over mesh wide defaults.
```

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+//
+// Note: fields in ProxyConfig are not dynamically configured; changes will require restart of workloads to take effect.
```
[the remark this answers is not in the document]
linsun: Yep, +100, this is a big deal and not obvious to users.

linsun:
If I update pod annotation to configure proxy (for example update currency or add a new env var in my proxy metadata), do I also need to manually restart the pod?

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+// apiVersion: networking.istio.io/v1beta1
+// kind: ProxyConfig
+// metadata:
+//   name: my-proxyconfig
+//   namespace: istio-system
+// spec:
+//   concurrency: 0
+// ```
+//
+// This can also be configured per-workload basis by setting the `workloadSelector` field on the `ProxyConfig` resource. For example:
```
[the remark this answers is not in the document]
Monkeyanator: Yep leaving unset should make it apply namespace-wide, added. From what I can tell the `WorkloadSelector` field descriptions are mostly copy-pasted from each other, don't think there's a central place.

networking/v1beta1/proxy_config.proto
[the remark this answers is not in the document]
Monkeyanator: Added namespace-wide defaults here. Can create an issue to track creating a page describing our config model but I think we can track separately since we'd want to update a bunch of API docs at once.

networking/v1beta1/proxy_config.proto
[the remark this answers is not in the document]
Monkeyanator: Added a bold **NOTE** to the beginning of the line.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,73 @@
...
+//
+//
+// If a `ProxyConfig` CRD is defined that matches a workload it will take precedence over the annotation for that workload.
+// Furthermore, options configured in a `ProxyConfig` CRD matching a workload will take precedence over mesh wide defaults.
+//
+// Note: fields in ProxyConfig are not dynamically configured; changes will require restart of workloads to take effect.
+message ProxyConfig {
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `ProxyConfig` configuration should be applied.
+  WorkloadSelector workload_selector = 1;
```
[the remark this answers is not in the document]
Monkeyanator: Makes sense, looks like the more recently added APIs use "selector", done.

networking/v1beta1/proxy_config.proto
Monkeyanator: Good point... "environment_variables" is clearer to me, especially since only envvars prefixed with `ISTIO_META_` get used as node metadata.

networking/v1beta1/proxy_config.proto
Monkeyanator: Ah good point, might make sense to line up with the behavior of the Telemetry API where:

> For any namespace, including the root configuration namespace, it is only valid to have a single workload selector-less Telemetry resource.

and also:

> For resources with a workload selector, it is only valid to have one resource selecting any given workload

Although it doesn't seem much worse to allow multiple and merge them together at each tier? @douglas-reid what was the reason it was decided to disallow this for telemetry API? Ambiguity with how resources should merge in the same tier (e.g. two resources both selecting same workload specifying different values for the same fields)?

Monkeyanator:
> Without proxy config CRD in Istio today, if I update pod annotation to configure proxy (for example update currency or add a new env var in my proxy metadata), do I also need to manually restart the pod?

Yes, the annotation is used at inject-time so pod restart is required to get the new values injected (although it seems like there's some existing work to dynamically receive ProxyConfig updates through XDS https://github.com/istio/istio/blob/925e6e61704893fdbaa1e5bdf4b1b86115cdea37/pkg/istio-agent/agent.go#L167 so maybe more fields could be made dynamic?)

>   I think a “smart merge” like we have in telemetry api / new mesh config merge is what we need here. This normally behaves like merge but it can protect certain keys that should not be messed with.

Agreed a smart merge makes sense here.

smawson:
I'm happy with this but would like to see merge semantics define per @mandarjog request.

Monkeyanator:
From the Environments WG discussion this morning, I think there are two questions:

> (1) Handling global fields that should not be overwritable on lower levels (i.e. namespace-scoped and workload-scoped ProxyConfig CR should *NOT* be able to override fields like `CA_ADDR`, `XDS_CA_ADDR`, `discoveryAddress` etc) 

For this I think we just document that certain fields cannot be set except through global (`meshConfig.defaultConfig` or CR in root namespace) config. This also precludes the a "replace" based API that could unset these values at lower levels. Other than these special fields, the resources will merge global->namespace->workload with values from the most specific matching CR taking precedence (the usual model).

> (2) Precedence between existing `proxy.istio.io/config` (per-workload) and `meshConfig.defaultConfig` (global) configuration and the new CRs.

The discussed solution was to __completely ignore__ the global CR if there are fields set in `meshConfig.defaultConfig` and do the same for workload-scoped CRs if `proxy.istio.io/config` is set. While this entirely avoids the question of merging I don't think it's correct because it completely disallows usage of the new CR until there are _no usages_ of the old config sources. Also, the new CR provides a tiny subset of the options that the old config does, so users would be stuck not using the CR if they wanted to set those values.

My proposal is to merge the old (annotations + defaultConfig) and the new (CR), and when there are overlapping values, _the value from the old config takes precedence_. This allows users to incrementally adopt the CR without moving all their values over at once. I think we could check for and warn on overlap in an analyzer if that's something we wanted to do.

There's a tradeoff with this being slightly more difficult for users to understand (e.g. "why are some of my CR values applying and others not?") but I think if we document this interplay between old and new configs up-front and clearly it shouldn't be an issue.

WDYT @mandarjog @costinm @smawson?

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,101 @@
...
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1beta1
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message ProxyConfig {
+  // Criteria used to select the specific set of pods/VMs on which this `ProxyConfig` configuration should be applied.
+  // If not set, the `ProxyConfig` resource will be applied to all workloads in the same namespace.
+  WorkloadSelector selector = 1;
```
howardjohn, line 91: should use `istio.type.v1beta1.WorkloadSelector` NOT `networking.WorkloadSelector` IMO. This is used by telemetry, request auth, authz, peer auth, and wasm. networking one is used by sidecar, SE, and envoyfilter.

Seems the general trend is in the direction of `istio.type.v1beta1.WorkloadSelector`.

Side note, outside of scope of this PR: we should probably consolidate them. I don't know if we can make the proto backwards compat to merge messages, but the YAML interface can be identical.

howardjohn (commented):
LGTM

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,101 @@
...
+
+// $schema: istio.networking.v1beta1.ProxyConfig
+// $title: ProxyConfig
+// $description: Provides configuration for individual workloads.
+// $location: https://istio.io/docs/reference/config/networking/proxy-config.html
+// $aliases: [/docs/reference/config/networking/v1beta1/proxy-config]
+// $mode: none
+
+// `ProxyConfig` defines configuration for individual workloads. `ProxyConfig` can be configured on a per-workload basis, a per-revision
+// basis, a per-namespace basis, or mesh-wide. `ProxyConfig` is not a required resource; there are default values in place.
```
nrjpoddar, line 28: Where can users find out defaults?

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,101 @@
...
+// apiVersion: networking.istio.io/v1beta1
+// kind: ProxyConfig
+// metadata:
+//   name: my-proxyconfig
+//   namespace: istio-system
+// spec:
+//   concurrency: 0
+// ```
+//
+// For workload level configuration, set the `selector` field on the `ProxyConfig` resource:
```
nrjpoddar, line 43: Can we state what's the behavior if multiple workload level ProxyConfig CRs select the same workloads?

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,101 @@
...
+//
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1beta1
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message ProxyConfig {
+  // Criteria used to select the specific set of pods/VMs on which this `ProxyConfig` configuration should be applied.
+  // If not set, the `ProxyConfig` resource will be applied to all workloads in the same namespace.
```
nrjpoddar, line 90:
```suggestion
  // Selectors specify the set of pods/VMs on which this `ProxyConfig` resource should be applied.
  // If not set, the `ProxyConfig` resource will be applied to all workloads in the namespace where this resource is defined.
```

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,101 @@
...
+  WorkloadSelector selector = 1;
+
+  // The number of worker threads to run.
+  // If unset, defaults to 2. If set to 0, this will be configured to use all cores on the machine using
+  // CPU requests and limits.
```
nrjpoddar, line 95: Which one is used requests or limits if both are defined and are different?

networking/v1beta1/proxy_config.proto
Monkeyanator: Done.

networking/v1beta1/proxy_config.proto
Monkeyanator: Documented with the fields, added comment clarifying.

networking/v1beta1/proxy_config.proto
Monkeyanator: Done.

networking/v1beta1/proxy_config.proto
Monkeyanator: Done.

networking/v1beta1/proxy_config.proto
Monkeyanator: Limits takes precedence over requests when estimating concurrency here, changed the comment.

releasenotes/notes/proxy-config.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+releaseNotes:
+  - |
+    **Added** a CRD for configuring `ProxyConfig` values containing a stable subset of the configuration from `MeshConfig.DefaultConfig`.
```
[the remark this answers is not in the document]
Monkeyanator: Done.

networking/v1beta1/proxy_config.proto
```diff
@@ -0,0 +1,101 @@
...
+// with the CR taking precedence over the annotation for overlapping fields. Furthermore, options configured in a `ProxyConfig`
+// CR matching a workload will similarly merge with mesh-wide and namespace-wide defaults.
+//
+// **NOTE**: fields in ProxyConfig are not dynamically configured- changes will require restart of workloads to take effect.
+
+package istio.networking.v1beta1;
+
+option go_package= "istio.io/api/networking/v1beta1";
+
+// `ProxyConfig` defines configuration for individual workloads.
```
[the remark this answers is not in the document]
Monkeyanator: Agreed, changed to `ProxyConfig exposes proxy level configuration options`

PR #2083, opened 2021-08-21
update api docs for retriable status codes

ramaraochavali:
Our default retry policy sets retriable status codes and defaults it to 503. https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/route/retry/retry.go#L36

This PR makes it configurable so that retry policy specified in VS can also make use of this feature.


ramaraochavali:
I have realized the retry_on processes http status codes also and adds them to retriable status code. This configuration is not needed. I will update docs to reflect the same.

ramaraochavali:
@howardjohn @hzxuzhonghu @nrjpoddar can you PTAL?

networking/v1alpha3/virtual_service.proto
```diff
@@ -1518,7 +1520,7 @@ message StringMatch {
 //     retries:
 //       attempts: 3
 //       perTryTimeout: 2s
-//       retryOn: gateway-error,connect-failure,refused-stream
+//       retryOn: gateway-error,connect-failure,refused-stream,503
```
hzxuzhonghu, line 1523: AFAIK gateway-error includes [502. 503, 504], this is not a good example

ramaraochavali: Fixed

howardjohn (approved):

hzxuzhonghu (approved):

PR #2084, opened 2021-08-23
Add support for Envoy's connection type LOGICAL_DNS in ServiceEntry

jacob-delgado:
Fixes https://github.com/istio/istio/issues/35475

Istio does not support LOGICAL_DNS in Envoy via ServiceEntry. This is not optimized for users that use external APIs from providers that have rotating DNS records.

See https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#arch-overview-service-discovery-types-logical-dns


networking/v1alpha3/service_entry.proto
```diff
@@ -908,6 +908,15 @@ message ServiceEntry {
...
     DNS = 2;
+
+    // Attempt to resolve the IP address by querying the ambient DNS,
+    // during request processing. Unlike DNS, connections made to
+    // hosts will be retained even if DNS records change frequently under
+    // LOGICAL. This is best suited for large web scale services that
+    // must be accessed via DNS. The proxy will resolve the DNS address
+    // specified in the hosts field, if wildcards are not used. DNS resolution
+    // cannot be used with Unix domain socket endpoints.
+    LOGICAL = 3;
```
howardjohn, line 919: I think LOGICAL is a bad name for this. We should not copy Envoy's ~illogical~ confusing names, IMO

networking/v1alpha3/service_entry.proto
```diff
@@ -908,6 +908,15 @@ message ServiceEntry {
...
     DNS = 2;
+
+    // Attempt to resolve the IP address by querying the ambient DNS,
+    // during request processing. Unlike DNS, connections made to
+    // hosts will be retained even if DNS records change frequently under
+    // DNS_ROUND_ROBIN. This is best suited for large web scale services that
+    // must be accessed via DNS. The proxy will resolve the DNS address
+    // specified in the hosts field, if wildcards are not used. DNS resolution
+    // cannot be used with Unix domain socket endpoints.
+    DNS_ROUND_ROBIN = 3;
```
hzxuzhonghu, line 919: Could I understand the difference with strict dns is that connection drain or not? Under what cases strict dns should be used instead of logic dns?

jacob-delgado: I need to perform some experiments and gather some tcp and envoy metrics and config. I setup an EnvoyFilter to get this behavior, but I should examine the socket connections for the case it presents. Here's some info I've found:

From https://dzone.com/articles/getting-started-with-lyft-envoy-for-microservices
> strict_dns: Envoy will monitor DNS, and every matching A record will be assumed valid.

> logical_dns: Envoy will basically use the DNS to add hosts, but will not discard them if they’re no longer returned by DNS (think round-robin DNS with hundreds of hosts — we’ll get more into this in subsequent articles).

and from https://groups.google.com/g/envoy-users/c/i5PjUJJBjF8/m/WFM1ilArAgAJ
> Are you using STRICT_DNS or LOGICAL_DNS? For AWS services (or any massive web service in which DNS will continuously return different results) I recommend using LOGICAL_DNS.

The number of DNS queries made is unchanged relative to STRICT_DNS, but when DNS queries are performed their results do not invalidate any connections made or anything in the connection pool. For services that perform many API requests to the same hostname over a long period of time this would be a nice optimization.

Strict DNS should be used in cases where DNS round robin isn't being used or the number of DNS records is low. There's also the case where if DNS fails to return any connection it appears that all connections are drained; this does not appear to be the case with LOGICAL_DNS.

Maybe @mattklein123 has some more insight on the underlying behavior or if I've stated anything wrong.

jacob-delgado: cc @howardjohn Here's the EnvoyFilter that may fix the issue reported

```yaml
kind: EnvoyFilter
metadata:
  name: sleep
  namespace: sleep
spec:
  workloadSelector:
    labels:
      app: sleep
  configPatches:
  - applyTo: CLUSTER
    match:
      context: SIDECAR_OUTBOUND
      cluster:
        service: api.github.com
    patch:
      operation: MERGE
      value:
        type: LOGICAL_DNS
```

hzxuzhonghu: make sense

ramaraochavali: // Attempt to resolve the IP address by querying the ambient DNS,
    // during request processing.
Is incorrect. I know the same comment is there for DNS but we should update both to say

    // Attempt to resolve the IP address by querying the ambient DNS,
    // asynchronously

networking/v1alpha3/service_entry.proto
```diff
@@ -908,6 +908,15 @@ message ServiceEntry {
     // the destination IP address.  DNS resolution cannot be used with Unix
     // domain socket endpoints.
     DNS = 2;
+
+    // Attempt to resolve the IP address by querying the ambient DNS,
+    // during request processing. Unlike DNS, connections made to
```
ramaraochavali, line 913: Just a suggestion to explain things more clearly

Unlike DNS, DNS_ROUND_ROBIN only uses the first IP address returned when a new connection needs to be initiated without relying on complete results of DNS resolution and  connections made to  hosts will be retained even if DNS records change frequently eliminating draining connection pools and connection cycling.

networking/v1alpha3/service_entry.proto
jacob-delgado: Came up with `DNS_ROUND_ROBIN`, but I'm not sure if it's any better.

howardjohn (approved):

ramaraochavali (approved):

PR #2086, opened 2021-08-24
Clarify request authentication root namespace scope.

incfly:
No behavior change just clarification.

We have the clear documentation in authorization policy of what would happen for "istio-system" authz, but not request authn.
Verified the behavior is as documented in this PR. This would give some clarify of the use case.

Also to give some pointers for the impl,
https://github.com/istio/istio/blob/37b717fc897730710e224c1f0e55303e374d22d9/pilot/pkg/model/authentication.go#L231.


incfly:
where authz is doc, https://github.com/istio/api/blob/2af152115f81571984b4146e93274f82ecc73f3e/security/v1beta1/authorization_policy.proto#L137

security/v1beta1/request_authentication.proto
```diff
@@ -64,6 +64,21 @@ option go_package="istio.io/api/security/v1beta1";
 //         requestPrincipals: ["*"]
 // ```
 //
+// - A policy in root namespace ("istio-system") applies to workloads in all namespaces in a mesh.
+// The following policy makes all sidecars perform JWT authentiation.
```
yangminzhu, line 68: not only sidecars, I think this also applies to gateways.

incfly: good point. updated.

incfly:
friendly ping, thanks.

@smawson @howardjohn @nrjpoddar @brian-avery

security/v1beta1/request_authentication.proto
```diff
@@ -64,6 +64,33 @@ option go_package="istio.io/api/security/v1beta1";
 //         requestPrincipals: ["*"]
 // ```
 //
+// - A policy in root namespace, "istio-system" by default, applies to workloads in all namespaces
```
smawson, line 67:
```suggestion
// - A policy in the root namespace ("istio-system" by default) applies to workloads in all namespaces
```

security/v1beta1/request_authentication.proto
```diff
@@ -64,6 +64,33 @@ option go_package="istio.io/api/security/v1beta1";
...
+// in a mesh. The following policy makes all workloads only accept the requests that contain a
```
smawson, line 68:
```suggestion
// in a mesh. The following policy makes all workloads only accept requests that contain a
```

linsun (approved):
LGTM after @smawson's suggestions

smawson (approved):

PR #2088, opened 2021-08-30
Add scheme and port to redirect

howardjohn:
This adds a `port` and `scheme` field to the virtual service `redirect`. The primary motivation here is to provide feature parity with the gateway-api: https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRequestRedirect; this will additionally be used for the implementation of that API.

Port redirect is useful if you are using non-standard (80/443) ports; today the redirect will *always* go to port 80/443 (depending on scheme), meaning it will likely not work for these services at all.

Scheme redirect is similar to HttpsRedirect on gateway, but because it is on the VS level we have more flexibility. For example, httpsRedirect is unusable with ACME (https://github.com/istio/istio/issues/27643) because it will redirect the `.well-known/acme-challenge` calls! With this PR, you can do it by redirecting `prefix: /` while still choosing to *not* redirect other urls. Example config in https://github.com/istio/istio/issues/27643#issuecomment-908786514.

Alternatives considered:
* https://github.com/istio/istio/pull/34910, which hacked `authority` to allow setting the full location header. For example, `authority: https://:8443` would set the port and scheme, leaving authority untouched. This was deemed a incorrect usage of the field by @ramaraochavali @hzxuzhonghu , which preferred an API change. I tend to agree.
* Use headers.set.location. This is *almost* workable with something like `"%REQ(:scheme)%://%REQ(:authority)%:%DOWNSTREAM_LOCAL_PORT%"`. The problem is `:authority` may have the port, so to do this we would need an envoy API change to allow a more complex header set. This is also nearly unusable for real users, and only helpful in translated config from gateway-api
* Do not expose this change in VirtualService at all, pass the redirect through an internal config. The generated VirtualServices from gateway-api persist in memory only so they do not need to represent exactly the same as the virtual service CRD. However, this is a bit hard to do with anything other than annotations, as maintaining a separate go struct requires a lot of effort (dev and runtime)

Validation webhook changes:
* Currently requires URI or authority. Now we just need `URI || authority || port || scheme`. Any combination of those is valid as long as we have at least one.
* port: must be 0-max_port_num
* scheme: only http and https are allowed

Open questions:
We need to expose three options for port: default port (80/443) - must be the default option if nothing set for backwards compat, explicitly set port, and retain request port. Note that "retain request port" is required as a single VirtualService can apply to multiple ports. In this PR, I have chosen to add a special value `0` to indicate `retain request port`. Some alternatives:

oneof. useRequestPort=false is invalid:
```
port:
  number: <set this to explicitly declare>
  useRequestPort: <set this to true to use request port>
```

presence check port:
```
port: 
  number: 8080 # explicit
```
OR
```
port: {} # use request port
```


hzxuzhonghu (approved):

costinm (approved):

networking/v1alpha3/virtual_service.proto
```diff
@@ -1413,6 +1413,16 @@ message HTTPRedirect {
   // this value.
   string authority = 2;
 
+  // On a redirect, overwrite the port portion of the URL with this value.
+  // If unset, the port will be unset (resulting in port 80 for HTTP and 443 for HTTPS).
+  // If set to 0, the port from the original request will be used
+  google.protobuf.UInt32Value port = 4;
```
ramaraochavali, line 1419: Is this change now forcing user to specify `0` for the original port to be used? i.e. when user just specifies authority earlier we are using the port in the request now he has to specifically set to 0 to get the same behaviour?

hzxuzhonghu: The default value for uint32 is 0

hzxuzhonghu: So i think the behavior is same as before

howardjohn: nil (ie default, what all current users have) - so not set port at all
0 - explicitly set to request port

ramaraochavali: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-redirectaction - Based on this I think if we do not set port, the original port in URL is used. May be comment here needs update if that is the default behaviour in Envoy?

howardjohn:
There's actually no way to tell Envoy "use the port from original request",
but we can hardcode it since all our routes are per-port so we know which
port the original request is from at the control plane.

ramaraochavali:
Ok. If you have already tested that is good. This statement in Envoy docs

"port_redirect
(uint32) The port value of the URL will be swapped with this value." is confusing to me. I thought it uses the original port if it is not provided. It does not talk about using 80/443.

ramaraochavali (approved):

louiscryan:
If we can't actually use the port of the request maybe we should rename the enum to be ACCESSED_PORT or SERVING_PORT instead of REQUEST_PORT and doc that it doesn't work with services that support port wildcarding

howardjohn:
> If we can't actually use the port of the request maybe we should rename the enum to be ACCESSED_PORT or SERVING_PORT instead of REQUEST_PORT and doc that it doesn't work with services that support port wildcarding

I am not sure I understand the different between ACCESSED_PORT and REQUEST_PORT?

While envoy doesn't support a native 'request port' concept, Istio doesn't have port wildcarding so it doesn't actually matter. If we add them the API can remain, we just need to extend envoy to have a native 'request port'.

howardjohn:
@louiscryan can you take another look?

networking/v1alpha3/virtual_service.proto
```diff
@@ -1413,6 +1413,24 @@ message HTTPRedirect {
   // this value.
   string authority = 2;
 
+  enum RedirectPortSelection {
+    INFER = 0;
```
louiscryan, line 1417: Nit:

derive_port: INFER | REQUEST_PORT

reads a little clunkily

derive_port: USE_PROTOCOL_DEFAULT | USE_REQUEST_PORT
derive_port: FROM_PROTOCOL_DEFAULT | FROM_REQUEST_PORT

read a little more fluently

networking/v1alpha3/virtual_service.proto
```diff
@@ -1413,6 +1413,24 @@ message HTTPRedirect {
...
+    uint32 port = 4;
+    // On a redirect, dynamically set the port:
+    // * INFER: automatically set to 80 for HTTP and 443 for HTTPS.
+    // * REQUEST_PORT: automatically use the port of the request.
+    RedirectPortSelection derive_port = 5;
+  }
+
+  // On a redirect, overwrite the scheme portion of the URL with this value.
+  // For example, "http" or "https".
+  // If unset, the original scheme will be used.
```
louiscryan, line 1431: Doc should connect the setting of this value with INFER above so folks understand the behavior. Could make SCHEME part of the enum value name to help with cognition.

networking/v1alpha3/virtual_service.proto
```diff
@@ -1413,6 +1413,24 @@ message HTTPRedirect {
   // this value.
   string authority = 2;
 
+  enum RedirectPortSelection {
+    INFER = 0;
+    REQUEST_PORT = 1;
+  }
+  oneof port_type {
```
louiscryan, line 1420: Nit:

s/port_type/redirect_port

Doesn't seem like a 'type' in the classical sense.

louiscryan (approved):

PR #2089, opened 2021-09-03
add disable_clear_route_cache to the internal authn filter config

yangminzhu:
Note: This only changes the internal authn filter to allow opt-out of the JWT claim based routing feature (see https://github.com/istio/proxy/pull/3477).

This is NOT end-user facing API and does not change the Istio CRDs at all.

@kyessenov @bianpengyuan @mandarjog @lambdai

### DIFF
--- a/envoy/config/filter/http/authn/v2alpha1/config.proto
+++ b/envoy/config/filter/http/authn/v2alpha1/config.proto
@@ -50,4 +50,9 @@ message FilterConfig {
   // When this field is set, the skip_validate_trust_domain field is ignored.
   // This field has no effect for plaintext traffic.
   repeated string allowed_trust_domains = 4;
+
+  // If true, the authn filter will clear the route cache so that the validated
+  // JWT token claims can be used in routing.
+  // This should be set to true for routing based on JWT claims.
+  bool clear_route_cache = 5;
 }
--- a/envoy/config/filter/http/authn/v2alpha1/config.proto
+++ b/envoy/config/filter/http/authn/v2alpha1/config.proto
@@ -51,8 +51,10 @@ message FilterConfig {
   // This field has no effect for plaintext traffic.
   repeated string allowed_trust_domains = 4;
 
-  // If true, the authn filter will clear the route cache so that the validated
+  // By default the authn filter will clear the route cache so that the validated
   // JWT token claims can be used in routing.
-  // This should be set to true for routing based on JWT claims.
-  bool clear_route_cache = 5;
+  // Advanced users can set this to true to disable the behavior if they do not
+  // want the authn filter to clear the route cache for any reasons.
+  // Warning: setting this to true will break the JWT claim based routing.
+  bool disable_clear_route_cache = 5;
 }


envoy/config/filter/http/authn/v2alpha1/config.proto
```diff
@@ -50,4 +50,9 @@ message FilterConfig {
   // When this field is set, the skip_validate_trust_domain field is ignored.
   // This field has no effect for plaintext traffic.
   repeated string allowed_trust_domains = 4;
+
+  // If true, the authn filter will clear the route cache so that the validated
+  // JWT token claims can be used in routing.
+  // This should be set to true for routing based on JWT claims.
+  bool clear_route_cache = 5;
```
mandarjog, line 57: If we expect the default case to clear the cache, this should have the opposite polarity.

yangminzhu: yeah that makes sense, renamed to disable_clear_route_cache.

yangminzhu:
@bianpengyuan @mandarjog @kyessenov could you take a look again, thanks.

bianpengyuan (approved):

mandarjog (approved):
lgtm

yangminzhu:
@brian-avery @howardjohn @linsun @louiscryan @nrjpoddar @smawson hi, could any of you give an approval to this PR to unblock the work? Note this has nothing to do with the Istio user-facing API, it's the config of the internal istio authn filter to allow users to opt-out the feature in case something goes wrong. Thanks.

howardjohn:
There was a lot of discussion about if this was appropriate, how it relates to WASM, third party plugins, etc. Can we summarize that discussion here?

yangminzhu:
> There was a lot of discussion about if this was appropriate, how it relates to WASM, third party plugins, whether Envoy should automatically do that etc. Can we summarize that discussion and resolution here?

@howardjohn , Sure, please see more detailed explanation below

**Background**: the opt-out config is added out of extra cautious to not break some advanced users with very special use cases mentioned by @kyessenov .

**The special use case explanation**: an advanced customer might inject a custom filter (e.g. Lua) that changed the header but not want the changed header to affect the route decision. They did this by intentionally not calling the clearRouteCache() in their custom filter and their custom filter is added before the Istio authn filter.

This is a really special case that won't be needed by majority normal customers. That's why it's an opt-out flag in the internal authn config. We  will not expose it in any Istio CRDs or MeshConfig.

**Resolution:**  The JWT claim based routing feature will change the Istio authn filter to call clearRouteCache() so that the JWT claims can be used in routing. The advanced users with the above special use case may NOT want the Istio authn filter to call refreshRouteCache() because they do not want the changes in their custom filter to affect the routing decision. They could now to set the opt-out flag (via EnvoyFilter) so that the Istio authn filter does clear the route cache.

**Relation to WASM**: If you change the headers in the WASM filter, it will always implicitly call the clearRouteCache() and this can NOT be disabled in the WASM filter, in other words, if the customer uses WASM filter, they are refreshing the route cache anyway and this flag won't help them.

This also means the Istio built-in metadata-exchange WASM filter is already always calling the clearRouteCache() method because it changes the metadata exchange header. (@bianpengyuan I think this also means always calling clearRouteCache() in Istio authn should have near zero performance impact because the router filter will need to re-calculate the route anyway because it was already cleared in the metadta-exchange filter)

**Expected behavior in Envoy**: According to this [Envoy comment](https://github.com/envoyproxy/envoy/blob/main/envoy/http/filter.h#L206-L207): `If a filter has modified the headers in a way that affects routing, clearRouteCache() must be called to clear the cache.`, so the standard behavior for a filter is to call clearRouteCache() if it does anything that could affect routing.

bianpengyuan:
> This also means the Istio built-in metadata-exchange WASM filter is already always calling the clearRouteCache() method because it changes the metadata exchange header. 

A subtle point here is that we do not set exchange header on inbound (server side sidecar), but I think most JWT usages should be at gateway and MX headers will be set there, so performance concern for route recompute might not be much of a problem.

howardjohn:
So the users impacted by this are users that are using JWT filter, and NOT using metadata exchange, and using a lua script that sets header, and are do NOT want the added header to impact routing, and are willing to change their lua script/expectations (because this is a temporary flag, right?), but are not able to change their lua script/expectations before upgrading to Istio 1.12?

This seems like quite a niche group. Do we have evidence we actually need to do the opt out?

yangminzhu:
> So the users impacted by this are users that are using JWT filter, and NOT using metadata exchange, and using a lua script that sets header, and are do NOT want the added header to impact routing, and are willing to change their lua script/expectations (because this is a temporary flag, right?), but are not able to change their lua script/expectations before upgrading to Istio 1.12?
> 
> This seems like quite a niche group. Do we have evidence we actually need to do the opt out?

I'm not sure, I only heard it from @kyessenov. Note after some more research I realized that it seems Lua filter will also always clear the route cache (see comment in https://github.com/envoyproxy/envoy/issues/8719#issuecomment-548460186). But I guess this is still possible because filters like ext-authz can modify the headers but only clear the route cache when the `clear_route_cache` field is set to true.

But I think the flag probably makes more sense for the following reasons instead of the above special use case:

1) Align with the similar design in other Envoy filters. The ext_authz filter has `clear_route_cache` (it's opt-in as the ext-authz filter is a public filter used for many different purposes, the Istio authn filter is not supposed to be used by other people and we want the user experience to be as smooth as possible, that's why we have opt-out here). 

    I also have a separate PR to add the same `clear_route_cache` to the Envoy JWT filter which will help the OSS customers but not us.

2) I think it is generally considered safer to have a way to opt-out such behavior change in Istio authn in case anything goes wrong, the user can quickly workaround it without re-installation.

3) We don't need the feature on sidecars anyway so we can simply disable the behavior on sidecar, it should avoid the extra performance overhead even though it is already very minimal.

howardjohn (approved):

PR #2090, opened 2021-09-08
add signer:rootcert map into mesh config

irisdingbj:
If istio acting as RA, user can specify signers for the CSR, We need a way to let user specify root cert for each of the signers. Detailed discussion can be found in: https://github.com/istio/istio/pull/34486


mesh/v1alpha1/config.proto
```diff
@@ -877,6 +877,11 @@ message MeshConfig {
   // configured globally via this field.
   istio.networking.v1alpha3.HTTPRetry default_http_retry_policy = 62;
 
+  // Map of root certs for kubernetes CSR signers when Istiod is acting as RA(registration authority)
+  // Key is the signer name and value is the secretName which stores the root certs for the signer
+  // signer1: signer-root-cert.ns1
+  map<string, string> signer_root_certificate = 63;
+
   // $hide_from_docs
   // Next available field number: 63
```
rveerama1, line 886: Update "Next available field number: 64". You used 63.

rveerama1:
I think you have to run "make proto-commit" to update proto.lock with new changes.

mesh/v1alpha1/config.proto
irisdingbj: good catch! Thanks!

irisdingbj:
@myidpt @costinm @howardjohn @shankgan Please take a look at this PR to see whether it is on the right direction. Thanks!

mesh/v1alpha1/config.proto
```diff
@@ -877,8 +877,13 @@ message MeshConfig {
   // configured globally via this field.
   istio.networking.v1alpha3.HTTPRetry default_http_retry_policy = 62;
 
+  // Map of root certs for kubernetes CSR signers when Istiod is acting as RA(registration authority)
+  // Key is the signer name and value is the secretName which stores the root certs for the signer
+  // signer1: signer-root-cert.ns1
+  map<string, string> signer_root_certificate = 63;
```
JoshVanL, line 883: Would there be room for specifying the CA as an in-line in the config itself rather than only being able to reference a Secret?

I wonder from a security standpoint whether this is _slightly_ better, since in line would mean only one resource can be used to modify the root CAs, rather than 2...

irisdingbj: Thanks! Issue for inline is since we can support multi-signers, not sure this will make the meshconfig huge..

irisdingbj:
@myidpt  @shankgan looking forward to your comments for this!

mesh/v1alpha1/config.proto
shankgan: nit: does this need to be a secret, as opposed to a configmap? Don't have any objection to using secret, just curious

shankgan (approved):

mesh/v1alpha1/config.proto
irisdingbj: The benefit for a secret is k8s will do format validation for the specified root cert.

howardjohn: Can you elaborate on how k8s will validate the format? I have never heard of this before. 

IMO inline is preferred *unless* we already expect the user to have  the root certs in configmap or secret for other reasons (say if cert manager writes them or similar)

costinm:
Storing public keys in secrets seems unnecessary and complex. We already
have a dozen ways to handle public keys and CAs - can we extend one of the
existing ones ?

costinm:
Also, why in meshconfig ? We have other ca settings in ProxyConfig - why
would this be a global setting instead of per namespace/workload ?

mesh/v1alpha1/config.proto
irisdingbj: for tls type of secret k8s will do some basic validationwhen you create it,  eg: 
```
 k apply -f secret-sample-bad.yaml
The request is invalid: patch: Invalid value: "map[data:map[tls.crt:sssssssssssss\n] metadata:map[annotations:map[kubectl.kubernetes.io/last-applied-configuration:{\"apiVersion\":\"v1\",\"data\":{\"tls.crt\":\"sssssssssssss\\n\"},\"kind\":\"Secret\",\"metadata\":{\"annotations\":{},\"name\":\"test-tls\",\"namespace\":\"default\"}}\n]]]": error decoding from json: illegal base64 data at input byte 13
```
The reason why not using inline is becasue if we have multiple singers, it will make the meshconfig more huge..

howardjohn: That's just the yaml representation it's no different than configmap. It's not providing meaningful tls cert verification it just requires secrete to be base64 encodes. When istiod reads them they aren't even I'm base64

irisdingbj:
@costinm Thanks for the review! 

The reason to add this into meshConfig instead of ProxyConfig is because 
1. istiod need to validate the returned certs chain which are signed from different signers against the root certs specified in meshConfig
2. The cert for istiod itself can be signed via the specified signer as well.  


I also went through the related  fileds in meshConfig and have not found a good match for this task: sepcify root certs for different signers:


1.  message CertificateData {
    oneof certificate_data {
      string pem = 1;
      SPIFFE_Trust_Domain_and_Bundle.md#the-spiffe-trust-domain-and-bundle
      string spiffe_bundle_url = 2;
    }
  }  

  this can not be used as it lack of places to specify signer name.

  2.  message CA {
    // REQUIRED. Address of the CA server implementing the Istio CA gRPC API.
    string address = 1;
    istio.networking.v1alpha3.ClientTLSSettings tls_settings = 2;
    google.protobuf.Duration request_timeout = 3;
    bool istiod_side = 4;
  }
  this can not be used as `address` is a required filed but we do no need this. 

  3. message Certificate {
  string secret_name = 1;
}
this can not be used as it will use built-in signers and have no place to define signers.

howardjohn:
If we are going to move forward with https://github.com/istio/api/pull/1923 we may want consistency there. @mathetake @lizan is there still interest in that API?

If so, it may make sense to add two fields there: "trust_domains" for the use case in 1923 and "signers" for the use case in 1923.

mathetake:
oh yeah we are still interested in 1923 so I would really appreciate it if we could leverage the API introduced here (trust_domains field).

mesh/v1alpha1/config.proto
```diff
@@ -877,8 +877,13 @@ message MeshConfig {
   // configured globally via this field.
   istio.networking.v1alpha3.HTTPRetry default_http_retry_policy = 62;
 
+  // Map of root certs for kubernetes CSR signers when Istiod is acting as RA(registration authority)
```
mathetake, line 880: drive-by comment: maybe we should stick to the term "trust_domain" instead of signer across Istio APIs, though "singer" is used by Kubernetes CSR but "trust_domain" is already used by Istio where both of them mean the same thing. Or maybe signerNameDomain? Related to @howardjohn's commen: https://github.com/istio/api/pull/2090#issuecomment-925378106

irisdingbj: Does this works for both cases?  we at least need to way to distinguish signer from trustDomain:
```
message CertificateData {
oneof certificate_data {
string pem = 1;
SPIFFE_Trust_Domain_and_Bundle.md#the-spiffe-trust-domain-and-bundle
string spiffe_bundle_url = 2;
}
oneof target {
	string signer = 1;
	repeated string trust_domains = 3; 
}
}
```

irisdingbj: Resuing this field will cause trouble to our `ISTIO_MULTIROOT_MESH` feature so not sure this is a good option

howardjohn: After reading the SPIFFE docs more, I tend to agree with @mathetake . With the current approach, we have many pods with the same trust domain but different root cert. This violates the SPIFFE guidelines in https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/#trust-domain:

> All workloads identified in the same trust domain are issued identity documents that can be verified against the root keys of the trust domain.

shankgan: The SPIFFE document says "root keys" of the trust-domain - implying that there can be more than one root of trust for a given trustDomain. The same document defines Trust Bundle as a "collection of one or more certificate authority (CA) root certificates" - don't see any issue with having multiple Root CA's in a trustDomain. Also, any attempt to join the fields may be confusing here. In https://github.com/istio/api/pull/1923, the goal of introducing additional root-certs is to help with mesh federation. The goal of this PR is to introduce an additional way to configure additional roots of trust within the same trustDomain

howardjohn: Thanks for the clarifications. It does still seem like we could reuse the same field:

* certificate_data with nothing associate: applies to all trust domains and default signer
* certificate_data with trust domain X: applies to X trust domain with default signer
* certificate_data with signer Y: applies to all trust domains with signer Y
* certificate_data with trust domain X and signer Y: applies to trust domain X with signer Y -- maybe this one doesn't make sense, we could make it a oneof

I do like the idea of reusing the field which already has a defined way to fetch the actual contents.

mathetake: +1 to the above reusable certificate_data

irisdingbj: So it means a seperate filed for trustdomain and signer like below? 

```
message CertificateData {
oneof certificate_data {
string pem = 1;
#2104 SPIFFE_Trust_Domain_and_Bundle.md#the-spiffe-trust-domain-and-bundle
string spiffe_bundle_url = 2;
}
repeated string signers = 3
repleated string trust_domains = 4
}
```


Also currently Root certs defined here will get populated to all workloads.  If we reuse this field, shall we add some filter function to get them populated to workloads?

irisdingbj: @howardjohn @shankgan your thoughts? Thanks!

irisdingbj:
@howardjohn @shankgan PTAL to see the latest change which is reusing the certificate_data filed.  @mathetake PTAL to see whether  the adding for trust_domains is approriate or you want to take care of it in a separate PR.  

Thanks!

mesh/v1alpha1/config.proto
```diff
@@ -223,6 +223,16 @@ message MeshConfig {
       // The certificate is retrieved from the endpoint.
       string spiffe_bundle_url = 2;
     }
+    // Optional. Specify the kubernetes signer to which this certificate data belongs to
+    // when Istiod is acting as RA(registration authority)
+    string cert_signer = 3;
```
howardjohn, line 228: This comment is not clear what the interaction is if trust_domains is also set.

Also, should this be repeated like trust_domains?

It is also not clear _how_ this is used, unlike the `trust_domains` field which says "if set..."

irisdingbj: @howardjohn refined comments and changed cert-signer to repeated. PTAL, Thanks!

mathetake (approved):
Thanks LGTM from me! (though I'm not sure if I can finish the multiple trust domains impl part for 1.12 as I am going into vacation for two weeks beginning the next week).

mesh/v1alpha1/config.proto
```diff
@@ -223,6 +223,21 @@ message MeshConfig {
       // The certificate is retrieved from the endpoint.
       string spiffe_bundle_url = 2;
     }
+    // Optional. Specify the kubernetes signers to which this certificate data belongs to
```
shankgan, line 226: nit: "//Optional. Specify the External CA signers that use this CA certificate"

mesh/v1alpha1/config.proto
```diff
@@ -223,6 +223,21 @@ message MeshConfig {
...
+    // when Istiod is acting as RA(registration authority)
+    // If set, they are used for these signers. Otherwise, this root CA is used for default signer.
+    repeated string cert_signers = 3;
+
+    // Optional. Specify the list of trust domains to which this certificate data belongs.
```
shankgan, line 231: nit: could we replace the word certificate with trustAnchor in the comments? This entire field is used to indicate CA certificate, which is used as a trustAnchor within the mesh. The use of the word "certificate" is ambiguous IMO

howardjohn (approved):
A few small comments to address but overall lgtm

irisdingbj:
> A few small comments to address but overall lgtm

Thanks! I addressed the comments and removed the on-hold label to get it in.  Glad to address new comments in following up PR if needed.

PR #2094, opened 2021-09-09
Add log formatting options to Telemetry API

douglas-reid:
This PR brings logging customization controls up to parity with Tracing / Metrics, allowing tailoring of log formats on a per-workload basis.  The API is based on the existing Open Telemetry Logging API specification, which should enable mapping into existing provider formats (`file`, `stream`, `open_telemetry`, and `stackdriver`).

This addresses several open FRs, including istio/istio#33307.


douglas-reid:
@mandarjog @kyessenov @howardjohn i'm trying to kick off Telemetry API work for 1.12. please take a look and provide some feedback. I opt'd out of trying to pull in the OTel protos directly for an initial round of discussion, etc., but they can be put back as required. I think it should be _relatively_ straightforward to map into existing providers here, and this avoids having to try to model structured logging formats, etc., in an envoy-specific way.

telemetry/v1alpha1/telemetry.proto
```diff
@@ -562,4 +562,75 @@ message AccessLogging {
...
+// It adopts portions of the Open Telemetry model for record formatting,
+// allowing mapping into various access logging providers.
+message LogFormat {
+  // Optional. A value containing the body of the log record. Valid uses include 
+  // a human-readable free-form string message and  structured data composed of arrays and maps
+  // of other values.
+  //
+  // TODO(dougreid): discuss mapping into supported providers?
+  //
+  // Example: ``body { string_value: "%PROTOCOL%" }``.
```
howardjohn, line 581: This is not yaml so its a bit confusing how a user is supposed to apply it?

telemetry/v1alpha1/telemetry.proto
```diff
@@ -562,4 +562,75 @@ message AccessLogging {
...
+  // If not specified, the istio default logging format will be used.
+  LogFormat log_format = 3;
+}
+
+// LogFormat provides control over the generated access log records.
+// It adopts portions of the Open Telemetry model for record formatting,
+// allowing mapping into various access logging providers.
+message LogFormat {
+  // Optional. A value containing the body of the log record. Valid uses include 
+  // a human-readable free-form string message and  structured data composed of arrays and maps
```
howardjohn, line 576: Do we have evidence structure format is useful? As far as I can tell NGINX, HAProxy, and Envoy all use a free form string. It seems quite a bit simpler to understand.

I couldn't tell how to use the structured form in this API (or the free form, really...), and the otel spec is like 20 pages which is a bit of a red flag it may be over complicated

howardjohn: Actually it makes a bit more sense after poking around the upstream docs a bit. is the otel proto meant for humans writing it though? It seems pretty unfriendly

douglas-reid: re: structured format utility -- support by major vendors, including cloud logging, and envoy itself, suggests that there are indeed those that use JSON-formatted logs. i know i have open issues against me about structured logging.

This is a bit verbose. That is definitely true. Adopting something like [SubstitutionFormatString](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/substitution_format_string.proto#envoy-v3-api-msg-config-core-v3-substitutionformatstring) also seemed a bit onerous. 

This is sort of what I'm trying to tease out here.  Do we want to provide any sort of structured customization (perhaps even CEL-based), or just say "here's a string, put whatever you want here" ?

I think we do want an ability to provide structured customization (say adding fields to Cloud Logging formats), but I'm not clear on what the "best" way forward is.  In past API reviews, we rejected putting the customization directly in the providers (having provider-specific overrides). Trying to generalize across formats unfortunately leads to bloat.

Maybe we could try to slim down the options, or abstract them in a way that makes _writing_ them easier. I'll have to do more thinking about what that might look like.

telemetry/v1alpha1/telemetry.proto
douglas-reid: apologies, i'll update with some yaml examples soon. i was just trying to put a strawman together quickly to spur discussion.

telemetry/v1alpha1/telemetry.proto
howardjohn: What is the functionality gap between just plain JSON (and raw string) vs the full otel proto? We already support json format today and its not hard to use.

telemetry/v1alpha1/telemetry.proto
```diff
@@ -562,4 +562,75 @@ message AccessLogging {
...
+
+  // Optional. Controls the format of the generated log records.
+  // If not specified, the istio default logging format will be used.
+  LogFormat log_format = 3;
+}
+
+// LogFormat provides control over the generated access log records.
+// It adopts portions of the Open Telemetry model for record formatting,
+// allowing mapping into various access logging providers.
+message LogFormat {
```
mandarjog, line 574: Do we need to make this provider agnostic? It can just be pass thru, no?

kyessenov:
I think the main point that needs a decision is whether we want to support Envoy format strings as Istio APIs. The alternative is to make the format string pre-defined and not expose it in API. Otherwise, this is fairly consistent with other telemetry providers.

telemetry/v1alpha1/telemetry.proto
douglas-reid: we support JSON format, but do we support easy customization of the JSON format ? does `accessLogFormat` string in proxy config count as easy to use for JSON format cases?

if we want to customize the attributes in any way, having some structured control over the key/values seems useful.

The bulk of my experience is with Cloud Logging. Having a way to separate out the `body` from additional `attributes` is useful for that case. That actually maps to some of the old [Mixer configuration](https://github.com/istio/old_mixer_repo/blob/master/adapter/stackdriver/config/config.proto#L154-L182), where we had a `payload` and `http_mapping` sections.

That may be too much for most users though?

We could have a `oneof` with a text format and a json format to allow for slightly simpler config, at the risk of inventing yet another logging API. The initial appeal of OTel to me is two-fold: (1) it exists already and (2) it can be mapped to other formats.

douglas-reid:
> I think the main point that needs a decision is whether we want to support Envoy format strings as Istio APIs. The alternative is to make the format string pre-defined and not expose it in API. Otherwise, this is fairly consistent with other telemetry providers.

I suspect that users _will_ want to customize the log formats -- to add/remove attributes, etc. (based on the forums, open issues, etc.) Pre-defining the format string seems fairly limiting. Providing an Istio layer that translates down into Envoy format strings also seems like a step too far. So, my instinct is to allow users to pass in format strings opaquely, with some control over the structure that results. Curious to hear thoughts of others.

telemetry/v1alpha1/telemetry.proto
howardjohn: > we support JSON format, but do we support easy customization of the JSON format ?

We support it but not ideal (IMO) (https://github.com/istio/istio/blob/473453566db2c0d86a32c9f10f9f3d27475fa647/pilot/pkg/networking/core/v1alpha3/accesslog.go#L249).

A better schema would be a oneof string or struct, instead of json-in-string.

What exactly would be in body vs attributes? It seems like *everything* is an attribute - but most of my experience is just with envoy basic logs

douglas-reid: `body`, to me, is _the_ log message about the event, and this can be in any sort of format (text, json, proto). `attributes` are meta info about the event being recorded (`http.url`, `http.status_code`).

so, you might have something like:

```
body: "Failure retrieving foo from bar; too many bazs"
attributes:
   http.url: "host/path?..."
   http.status_code: 500
```

The distinction between the two is more convention than anything else,  I suppose.

howardjohn: It makes sense in logs in general like application logs.. but what could the body be here? seems like it would be something trivial like "request complete" and then all the associated attributes

mandarjog:
Yeah, I think pass thru opaque string is fine. We can add an enum later to catalog "standard formats", but we don't need to start there.  

Freeform opaque string is both necessary and sufficient.  `Struct` offers some validation, however we will have provider specific validation that can do what ever it wants with the string.

telemetry/v1alpha1/telemetry.proto
douglas-reid: That's a good point, for default cases. But you could put anything you want in, potentially. so you could have: `body: "serviceA --> serviceB"` and the `attributes` be all the other info. The precedent for this would be something like the existing stackdriver TCP logs ([example](https://github.com/istio/proxy/blob/master/testdata/stackdriver/client_tcp_access_log_entry.yaml.tmpl)).

I'd be fine with starting without it, if we feel it is overkill.

howardjohn: Interesting... I guess you would need templating in the body as well for it to be useful.

Really it seems like body is just a specialized attribute that may be displayed different in various logging systems

kyessenov:
If you want envoy formatters, those are evolving and we need some way to selectively trigger extensions. E.g. some old envoys do not have REQ_WITHOUT_QUERY while newer ones do.

telemetry/v1alpha1/telemetry.proto
```diff
@@ -562,4 +562,75 @@ message AccessLogging {
   // selected above. Customization controls will be added to this API in
   // future releases.
   google.protobuf.BoolValue disabled = 2;
+
+  // Optional. Controls the format of the generated log records.
+  // If not specified, the istio default logging format will be used.
+  LogFormat log_format = 3;
```
hzxuzhonghu, line 568: What we need is far from customizing log format, but we also want specify headers to add

douglas-reid: The intent behind `LogFormat` is to allow exactly what you ask for. The idea would be to pass in format specifiers that reference headers, etc., that would be included in the log.

Perhaps we need a better name?  Or maybe something less opaque to make that more explicit?

hzxuzhonghu: Can you add an example

zirain: show we use some config like k8s volume for better usage, something like:
```yaml
- name: file-log
  file:
    path: /dev/stdout
    format: "[%START_TIME%]"
- name: grpc-http-log
  gprc:
    grpcService:
      address: grpc-als.istio-system.svc:9811
    additionalRequestHeadersToLog:
    - x-request-user
    additionalResponseHeadersToLog:
    - x-response-cost
- name: grpc-http-log
  grpcOtel:
    grpcService:
      address: grpc-als.istio-system.svc:9811
      
```

douglas-reid: @zirain does this mean that you want to log different things per provider?

douglas-reid:
@hzxuzhonghu I added some potential examples to the top comments in the proto file.

@hzxuzhonghu @mandarjog @kyessenov @howardjohn @zirain I took a quick stab at an alternate provider-based approach for the same functionality in #2110. Would appreciate your feedback on that approach versus this style.

PR #2104, opened 2021-09-17
add option to skip sidecar generated requestID based sampling decision

Shikugawa:
Add an option to set whether the Request ID generated by the sidecar should be used when determining the sampling strategy for tracing.

cc @basvanbeek @wu-sheng


mesh/v1alpha1/config.proto
```diff
@@ -650,6 +650,10 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. Sets whether the Request ID generated by the sidecar should be used when determining the sampling strategy in tracing.
```
bianpengyuan, line 654: How does request id affect sampling decision?

Shikugawa: See more context. https://www.envoyproxy.io/docs/envoy/v1.19.0/api-v3/extensions/request_id/uuid/v3/uuid.proto#envoy-v3-api-field-extensions-request-id-uuid-v3-uuidrequestidconfig-use-request-id-for-trace-sampling

In default behavior, front envoy will generate request ID and use it to identify single request. Also tracing strategy depends on the value of request ID (It is uuid and some bits will be used to determine whether sample or not.) But with this feature, we can disable this behavior, then we are allowed to use only provider specific request header value. e.g. x-b3-sample

bianpengyuan: I see. We should set it as true by default then.

bianpengyuan: IIUC If user set this to false, envoy will basically stop making any sampling decision and rely on sample intent from the downstream?

Shikugawa: Yes. if false, envoy relies on tracing provider's sampling policy like this option https://github.com/openzipkin/b3-propagation#sampling-state-1

bianpengyuan: Then how is it different from setting `random_sampling_percentage` to 0? https://github.com/istio/api/blob/master/telemetry/v1alpha1/telemetry.proto#L280-L289

Shikugawa: The reason this option is needed is to prevent sampling from failing when Envoy handles request IDs generated by other proxies (e.g. Nginx). In the current Envoy implementation, sampling is based on the value of the request ID, but if this request ID is not generated by Envoy, sampling will never take place. By setting this to false, Envoy can disable sampling policy decisions that depend on the request ID. As a result, no matter what format of request ID is handled by Envoy, the problem of no sampling can be prevented.

mesh/v1alpha1/config.proto
```diff
@@ -706,6 +710,10 @@ message MeshConfig {
 
       // Optional. The SkyWalking OAP access token.
       string access_token = 3;
+
+      // Optional. Sets whether the Request ID generated by the sidecar should be used when determining the sampling strategy in tracing.
```
howardjohn, line 714: Only sidecar? or all proxies?

Can we add some explanation as to *why* a user would set this? To non-experts of tracing I don't think it is clear

mesh/v1alpha1/config.proto
bianpengyuan: I see.. This seems a quite subtle setting which ties closely to Envoy implementation details. More comment is needed, and we probably should rename it to something more understandable (e.g.`use_zipkin_sampling`)

bianpengyuan: We probably also should think about hoist this up to common tracing configuration instead of make it per provider.

Shikugawa: I think it can be provider-specific value it is because it is not necessary all provider having their own sampling decision. But I'm not sure it exclude zipkin and apache skywalking. WDYT? cc @douglas-reid

mesh/v1alpha1/config.proto
Shikugawa: It should support all proxies. let me change docs.

mesh/v1alpha1/config.proto
douglas-reid: The underlying Envoy setting is not provider-specific, as far as I can tell, which I think leans towards exposing this as a higher-level concept.

I wonder how frequently this is used, and whether or not this should be in the API. It is fairly Envoy-specific. I'd like to hear other thoughts here.

mandarjog: Does this need to go in mesh config?
@douglas-reid is there home for something like this in the new API ?

Shikugawa: Fixed. And it will perform on any tracing providers.

lizan:
@istio/wg-policies-and-telemetry-maintainers ping, what blocks here?

mesh/v1alpha1/config.proto
```diff
@@ -877,6 +877,15 @@ message MeshConfig {
   // configured globally via this field.
   istio.networking.v1alpha3.HTTPRetry default_http_retry_policy = 62;
 
+  // This value is true by default; Envoy decides whether or not to sample based on
```
howardjohn, line 880: > This value is true by default

I think this is not really possible with the current API. `false` == `unset`.

We should make it a BoolValue wrapper?

mesh/v1alpha1/config.proto
```diff
@@ -877,6 +877,15 @@ message MeshConfig {
...
+  // the value of the Request ID generated by Ingress in distributed tracing.
```
howardjohn, line 881: what is "Ingress"? This is not an existing concept in Istio APIs

douglas-reid:
echoing @mandarjog, instead of adding this into meshconfig, should this be exposed as part of the Tracing bits of the telemetry API, alongside `random_sampling_percentage`, etc. ?

mesh/v1alpha1/config.proto
Shikugawa: fixed

mesh/v1alpha1/config.proto
Shikugawa: fixed. that means the proxy receives user traffic fitst

mandarjog:
@Shikugawa can recast it in the telemetry api instead of here? New features there is exactly what will help migration to the new api and it is cleaner.

Shikugawa:
> @Shikugawa can recast it in the telemetry api instead of here? New features there is exactly what will help migration to the new api and it is cleaner.

@mandarjog @douglas-reid I think it can't be accomplished for our requirements. It is because if malformed (from the perspective of Envoy) format of RID will notify to discard trace sampling. So even if you set random sampling percentage, nothing will be changed. https://github.com/envoyproxy/envoy/blob/365acdbe02ed7bfd2f8fdc616d057b430da2621e/source/common/http/conn_manager_utility.cc#L308-L312

telemetry/v1alpha1/telemetry.proto
```diff
@@ -335,6 +335,17 @@ message Tracing {
...
+
+
+  // This value is true by default; Envoy decides whether or not to sample based on
+  // the value of the Request ID generated by Ingress in distributed tracing.
+  // The format of this Request ID is specific to Envoy, and if the Request ID generated
+  // by the proxy that receives user traffic first is not specific to Envoy, Envoy
+  // will break the trace because it cannot interpret the Request ID. By setting
+  // this value to false, we can prevent Envoy from sampling based on the Request
+  // ID. As a result, the trace will not be broken even if the Request ID is
+  // not in the Envoy format.
```
douglas-reid, line 347: nit: can we add something like: `[Trace Context Propagation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/tracing#trace-context-propagation) provides more information on Request ID handling.` to the end of the comment ?

nrjpoddar:
Can we mark this hidden until the implementation is ready to go?

howardjohn (approved):

PR #2105, opened 2021-09-21
support x-jwt-claim in request authentication

yangminzhu:
Update the request authentication API for the support of `x-jwt-claim`, no CRD changes, reuse the `headers` field.

Design doc: https://docs.google.com/document/d/1s8Lykf8-KUZTLiG4YMSCHvxBDwZzHlrrHyZ7cJymc_8/edit#heading=h.6p29w9kcduab

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -186,11 +186,11 @@ option go_package = "istio.io/api/networking/v1alpha3";
 // +cue-gen:VirtualService:subresource:status
 // +cue-gen:VirtualService:scope:Namespaced
 // +cue-gen:VirtualService:resource:categories=istio-io,networking-istio-io,shortNames=vs
-// +cue-gen:VirtualService:printerColumn:name=Gateways,type=string,JSONPath=.spec.gateways,description="The names of gateways and sidecars 
+// +cue-gen:VirtualService:printerColumn:name=Gateways,type=string,JSONPath=.spec.gateways,description="The names of gateways and sidecars
 // that should apply these routes"
 // +cue-gen:VirtualService:printerColumn:name=Hosts,type=string,JSONPath=.spec.hosts,description="The destination hosts to which traffic is being sent"
-// +cue-gen:VirtualService:printerColumn:name=Age,type=date,JSONPath=.metadata.creationTimestamp,description="CreationTimestamp is a timestamp 
-// representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. 
+// +cue-gen:VirtualService:printerColumn:name=Age,type=date,JSONPath=.metadata.creationTimestamp,description="CreationTimestamp is a timestamp
+// representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations.
 // Clients may not set this value. It is represented in RFC3339 form and is in UTC.
 // Populated by the system. Read-only. Null for lists. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata"
 // +cue-gen:VirtualService:preserveUnknownFields:false
@@ -1061,8 +1061,24 @@ message HTTPMatchRequest {
   //
   // - `regex: "value"` for RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
   //
-  // If the value is empty and only the name of header is specfied, presence of the header is checked.
-  // **Note:** The keys `uri`, `scheme`, `method`, and `authority` will be ignored.
+  // If the value is empty and only the name of header is specified, presence of the header is checked.
+  // **Note 1:** The keys `uri`, `scheme`, `method`, and `authority` will be ignored.
+  //
+  // **Note 2:** The special key `x-jwt-claim` is reserved for matching with JWT claims on ingress gateway.
+  // Claim of type string or list of string is supported. Nested claims is also supported using the separator "." or "-"
+  // between each claim. For example,
+  //
+  // Use "." as the separator:
+  // - "x-jwt-claim.admin" matches the claim "admin".
+  // - "x-jwt-claim.group.id" matches the nested claims "group" and "id".
+  // - "x-jwt-claim.some-key.another" matches the nested claims "some-key" and "another".
+  // Use "-" as the separator:
+  // - "x-jwt-claim-admin" matches to the claim "admin".
+  // - "x-jwt-claim-group-id" matches to the nested claims "group" and "id".
+  // - "x-jwt-claim-www.example.com-name" matches the nested claims "www.example.com" and "name".
+  //
+  // This key is only supported on ingress gateway and requires a proper RequestAuthentication applied on the ingress
+  // gateway to validate the JWT and make the claims available for route matching.
   map<string, StringMatch> headers = 5;
 
   // Specifies the ports on the host that is being addressed. Many services
@@ -1567,15 +1583,15 @@ message StringMatch {
 //
 message HTTPRetry {
   // Number of retries to be allowed for a given request. The interval
-  // between retries will be determined automatically (25ms+). When request 
-  // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute) 
+  // between retries will be determined automatically (25ms+). When request
+  // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute)
   // or `per_try_timeout` is configured, the actual number of retries attempted also depends on
   // the specified request `timeout` and `per_try_timeout` values.
   int32 attempts = 1 [(google.api.field_behavior) = REQUIRED];
 
   // Timeout per attempt for a given request, including the initial call and any retries. Format: 1h/1m/1s/1ms. MUST BE >=1ms.
-  // Default is same value as request 
-  // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute), 
+  // Default is same value as request
+  // `timeout` of the [HTTP route](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRoute),
   // which means no timeout.
   google.protobuf.Duration per_try_timeout = 2;
 
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -186,11 +186,11 @@ option go_package = "istio.io/api/networking/v1alpha3";
 // +cue-gen:VirtualService:subresource:status
 // +cue-gen:VirtualService:scope:Namespaced
 // +cue-gen:VirtualService:resource:categories=istio-io,networking-istio-io,shortNames=vs
-// +cue-gen:VirtualService:printerColumn:name=Gateways,type=string,JSONPath=.spec.gateways,description="The names of gateways and sidecars
+// +cue-gen:VirtualService:printerColumn:name=Gateways,type=string,JSONPath=.spec.gateways,description="The names of gateways and sidecars 
 // that should apply these routes"
 // +cue-gen:VirtualService:printerColumn:name=Hosts,type=string,JSONPath=.spec.hosts,description="The destination hosts to which traffic is being sent"
-// +cue-gen:VirtualService:printerColumn:name=Age,type=date,JSONPath=.metadata.creationTimestamp,description="CreationTimestamp is a timestamp
-// representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations.
+// +cue-gen:VirtualService:printerColumn:name=Age,type=date,JSONPath=.metadata.creationTimestamp,description="CreationTimestamp is a timestamp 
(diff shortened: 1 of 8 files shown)


networking/v1alpha3/virtual_service.proto
```diff
@@ -1061,8 +1061,24 @@ message HTTPMatchRequest {
   //
   // - `regex: "value"` for RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
   //
-  // If the value is empty and only the name of header is specfied, presence of the header is checked.
-  // **Note:** The keys `uri`, `scheme`, `method`, and `authority` will be ignored.
+  // If the value is empty and only the name of header is specified, presence of the header is checked.
+  // **Note 1:** The keys `uri`, `scheme`, `method`, and `authority` will be ignored.
+  //
+  // **Note 2:** The special key `x-jwt-claim` is reserved for matching with JWT claims on ingress gateway.
+  // Claim of type string or list of string is supported. Nested claims is also supported using the separator "." or "-"
```
yangminzhu, line 1068: The `.` and `-` together should be good for 99% claim names in JWT token.

If we want to support the case when both `.` and `-` is used in JWT token, we could consider:

option 1: Support more allowed special characters (`"!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"` defined in https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) for valid separator.

option 2: Support all characters (including those not allowed in header name) to be used as claim separator.

option 3: add a dedicated jwt claim field in virtual service instead of reusing the header field. This has been discussed in the design doc and is not favored by the networking WG as it makes it harder to align with the new gateway-api in the future.

cc @howardjohn @mandarjog @costinm @louiscryan

howardjohn: I think 1 and 2, and the existing option are too complex. I haven't seen any similar APIs that have dynamic implicit separators; it took me quite a while to grok this at all.

costinm:
Or: for the first iteration just support the common/standard claims, which
are simpler ( aud, iss, etc).

We will still need to reconcile this with claims from client mTLS
certificate - where the set of options are more limited.

yangminzhu:
> Or: for the first iteration just support the common/standard claims, which are simpler ( aud, iss, etc).

@costinm there are many non standard claims we need support (e.g. `group`, `groups`, etc.), I think we just don't need to restrict the claim name here.

Regarding nested claims, it's actually common and can be found in JWT issued by many identity provider, the same feature was asked by many users for the authorization policy before. I'm just afraid that it may block the adoption for some customers without the nested claim support.

costinm:
It is fine to iterate on a feature - provide a basic feature first, get
feedback and resolve the problem for some customers.

It is also fine if some customers and use cases can't be satisfied - if
other products provide such feature
they can use them as well. Or use WASM, EnvoyFilters, etc.

Having an API that is too complex or tries to solve all problems is usually
bad for everyone, in particular
for security where complexity almost always results in security bugs. Keep
in mind all this needs tests, docs,
etc - it is much better to take small iterative steps.

yangminzhu:
> It is fine to iterate on a feature - provide a basic feature first, get feedback and resolve the problem for some customers. It is also fine if some customers and use cases can't be satisfied - if other products provide such feature they can use them as well. Or use WASM, EnvoyFilters, etc. Having an API that is too complex or tries to solve all problems is usually bad for everyone, in particular for security where complexity almost always results in security bugs. Keep in mind all this needs tests, docs, etc - it is much better to take small iterative steps.

Nested claims is a common and pretty basic use case and not even close to the "all problem" statement.

To start with a simpler API, we can support the `.` as the separator: `x-jwt-claim.key1` and `x-jwt-claim.key1.key2` are both very natural formats, there is no extra complexity to support nested claim at all. It won't work for claim name with `.` but we can figure it out in later iterations.

networking/v1alpha3/virtual_service.proto
yangminzhu: @howardjohn I just updated to use only `.` as the separator, PTAL.  This won't work for 100% customers but should be good for most users and significantly simplify the API. I think we can iterate to support the corner cases (e.g. claim name with `.`) later.

costinm:
How about not making any change except mentioning that VirtualService may
route using metadata.

VirtualService is beta, x-jwt-claim is at best alpha - not a good idea to
confuse users to believe
this scheme is at beta stability and won't change.

Also keep in mind that the VirtualService API - and it's
K8S-Gateway/HttpRoute replacement -
are not tied to envoy, we support at least proxyless. Maybe x-jwt-claim
will move to beta at some point,
maybe will change based on feedback and making it consistent with mTLS
claims.

yangminzhu:
> How about not making any change except mentioning that VirtualService may route using metadata. VirtualService is beta, x-jwt-claim is at best alpha - not a good idea to confuse users to believe this scheme is at beta stability and won't change. Also keep in mind that the VirtualService API - and it's K8S-Gateway/HttpRoute replacement - are not tied to envoy, we support at least proxyless. Maybe x-jwt-claim will move to beta at some point, maybe will change based on feedback and making it consistent with mTLS claims.

@costinm I'm very confused by your comment. Can you clarify what you mean by `not making any change except mentioning that VirtualServie may route using metadata`? Isn't that's what the PR already doing? We're not changing the VirtualService CRD schema because we reuse the header field with a special key `x-jwt-claim`. This PR just adds the comment to mention it.

I'm not too worried about the alpha/beta thing, we can clarify it in the doc and API, the virtual service API itself is still using v1alpha1 API version.

We also shouldn't mention metadata because that's internal implementation detail in the data plane, different client may implement the support in a different way, for example, proxyless does not support metadata today.

ramaraochavali:
Simplified version of nested jwt claims with `.` seems reasonable to me.  As @yangminzhu mentioned we can iterate later.

yangminzhu:
@istio/technical-oversight-committee  please take a look, thanks.

yangminzhu:
@howardjohn @linsun @louiscryan @smawson @nrjpoddar could someone review the API change, thanks

networking/v1alpha3/virtual_service.proto
```diff
@@ -1061,8 +1061,19 @@ message HTTPMatchRequest {
   //
   // - `regex: "value"` for RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
   //
-  // If the value is empty and only the name of header is specfied, presence of the header is checked.
-  // **Note:** The keys `uri`, `scheme`, `method`, and `authority` will be ignored.
+  // If the value is empty and only the name of header is specified, presence of the header is checked.
+  // **Note 1:** The keys `uri`, `scheme`, `method`, and `authority` will be ignored.
+  //
+  // **Note 2:** The special key `x-jwt-claim` is reserved for matching with JWT claims on ingress gateway.
```
smawson, line 1067: Does this really need to be documented on VirtualService? I don't think we need to document all headers here, so how do we decide which headers are documented here vs. elsewhere?

My preference would be to not include this documentation on VirtualService, and instead document it as part of the documentation on JWT claim support on ingress gateways. Either that or we need to decide that we are going to document all relevant headers in VirtualService.

smawson:
My comment is basically the same thing Costin is saying, which is that this doesn't seem like the right place to document this, since I don't think we want to be exhaustive in VirtualService about what headers can be used.

That said, if we want to introduce a new capability that headers can have sub-structure, using '.' notation to access it, that is a new capability that we can document as part of virtual service. We could use x-jwt-claim as an example but the documentation should be about sub-header access, not about jwt routing. VirtualService should remain generic so we don't end up documenting a bunch of features in it.

yangminzhu:
@smawson thanks for the comment, I think there are 2 things here:

1. where to document the header? I'm fine to move it to request authentication, we can mention once the user applies the request authentication on ingress gateway, they can then also use the `x-jwt-claim` header for JWT claim based matching.

2. the syntax to access the jwt claim in the header? previously costin asked for a simpler syntax and I have updated it to `x-jwt-claim.key1.key2`, Are we okay with this syntax now? I think it's simple enough and covers the majority use cases in the jwt claim, note it's not supposed to be a generic sub-structure for all headers, it's specific to JWT claims, this will be more clear once we move the documentation to RequestAuthentication instead of VirtualService. cc @costinm

smawson:
> @smawson thanks for the comment, I think there are 2 things here:
> 
> 1. where to document the header? I'm fine to move it to request authentication, we can mention once the user applies the request authentication on ingress gateway, they can then use the `x-jwt-claim` header for JWT claim based matching in the virtual service.

Yes that's my preference, seems like a better place since not all VirtualService users will care about this, only users of JWT authentication.

> 2. the syntax to access the jwt claim in the header? previously costin asked for a simpler syntax and I have updated it to `x-jwt-claim.key1.key2`, Are we okay with this syntax now? I think it's simple enough and covers the majority use cases in the jwt claim, note it's not supposed to be a generic sub-structure for all headers, it's specific to JWT claims, this will be more clear once we move the documentation to RequestAuthentication instead of VirtualService. cc @costinm

I'm happy with the . syntax, but I do think we're introducing a new standard here, since we want consistency for our users. I'm good with this just being documented in RequestAuthentication but if we add other sub-structure matching in the future let's use the same syntax.

networking/v1alpha3/virtual_service.proto
yangminzhu: thanks for the review, updated to add in request authN.

yangminzhu:
@smawson thanks for the review, ptal for the update.

security/v1beta1/request_authentication.proto
```diff
@@ -156,6 +156,71 @@ option go_package="istio.io/api/security/v1beta1";
 //         paths: ["/healthz"]
 // ```
 //
+// - When applied on gateway, you can also use the special header name `x-jwt-claim` for matching with JWT claims in
```
smawson, line 159:
```suggestion
// - When applied on a Gateway, you can also use the special header name `x-jwt-claim` for matching JWT claims in
```

security/v1beta1/request_authentication.proto
```diff
@@ -156,6 +156,71 @@ option go_package="istio.io/api/security/v1beta1";
...
+// the virtual service. Claim of type string or list of string is supported and nested claims is also supported using
```
smawson, line 160:
```suggestion
// the VirtualService. Claims of type string or list of string are supported and nested claims are also supported using
```

security/v1beta1/request_authentication.proto
```diff
@@ -156,6 +156,71 @@ option go_package="istio.io/api/security/v1beta1";
...
+// the `.` as separator for claim names. Examples: `x-jwt-claim.admin` matches the claim "admin" and `x-jwt-claim.group.id`
```
smawson, line 161:
```suggestion
// `.` as a separator for claim names. Examples: `x-jwt-claim.admin` matches the claim "admin" and `x-jwt-claim.group.id`
```

security/v1beta1/request_authentication.proto
```diff
@@ -156,6 +156,71 @@ option go_package="istio.io/api/security/v1beta1";
...
+// matches the nested claims "group" and "id".
+//
+// The following example creates the request authentication and authorization policy for JWT validation on ingress
```
smawson, line 164:
```suggestion
// The following example creates the request authentication and authorization policies for JWT validation on ingress
```

security/v1beta1/request_authentication.proto
```diff
@@ -156,6 +156,71 @@ option go_package="istio.io/api/security/v1beta1";
...
+// gateway and route the request based on the "version" claim in the validated JWT.
```
smawson, line 165:
```suggestion
// gateway and routes requests based on the "version" claim in the validated JWT.
```

security/v1beta1/request_authentication.proto
```diff
@@ -156,6 +156,71 @@ option go_package="istio.io/api/security/v1beta1";
...
+//         host: foo.prod.svc.cluster.local
+//         subset: v2
+//   - name: "default-route"
+//     route:
+//     - destination:
+//         host: foo.prod.svc.cluster.local
+//         subset: v1
+// ```
+//
+// **Note:** This only supports gateway and a proper request authentication must first be applied to validate the JWT.
```
smawson, line 221:
```suggestion
// **Note:** This routing is only supported on Gateways and proper request authentication must first be applied to validate the JWT.
```

security/v1beta1/request_authentication.proto
```diff
@@ -156,6 +156,71 @@ option go_package="istio.io/api/security/v1beta1";
...
+// The special key `x-jwt-claim` is matched with internal metadata populated after successful JWT validation.
```
smawson, line 222: Not sure what this means, what should the reader take away from this comment?

yangminzhu: oh, the original purpose was to explain a little bit about the implementation but now I feel it looks redundant and not very helpful, should not mention too much detail about the implementation as it could change anytime, removed.

smawson (approved):

yangminzhu:
@costinm @howardjohn could you take a look again? thanks. I have updated the PR per sven's comment.

howardjohn (approved):

PR #2107, opened 2021-09-30
Move istio/proxy protos into istio/api

howardjohn:
Fixes https://github.com/istio/istio/issues/35210.

This does two things:
* Move some protos we have in istio/istio that are manually generating
the go protos, using an outdated protobuf library and outdated proto.
* Move some protos we have in istio/proxy that are not generating any go
protos at all.

A next step will be to have istio/istio import these go libraries, and
istio/proxy to import these as proto_libraries.


howardjohn:
@kyessenov @yangminzhu any concerns with this?

kyessenov:
Yeah, I think this is just adding overhead. These are not user visible APIs. This does not help proxy build at all (which is all bazel).

howardjohn:
> Yeah, I think this is just adding overhead. These are not user visible APIs. This does not help proxy build at all (which is all bazel).

I wasn't sure if it was use for proxy build but I think it is for istio/istio. This was mostly inspired by your comment, which I agree with: https://github.com/istio/istio/pull/33583#discussion_r667149402.

We don't have to change istio/proxy if its just annoying. WDYT?

kyessenov:
You can manually generate and commit protos to istio/istio with standard toolchain. It's not like they change every month. It's just confusing to require TOC approval for machine protos so this repo is not appropriate.

howardjohn:
We do already have half of them here. I would prefer to be consistent
personally. It's also useful to be able to import just the Protos without
the dependency hell of Istio/istio - Protos living in Istio repo has been
an issue for at least 3 projects I know of

howardjohn:
we can make Network team the owners of envoy/?

kyessenov:
As I mentioned earlier, this is just pure overhead IMHO. Yes, it can be done, but the value of tooling in istio/api for Envoy build is close to zero if not negative. Some protos in istio/proxy are not even protos, as they are just documentation for Wasm configs (which is not using protobuf).

howardjohn:
I mentioned above "We don't have to change istio/proxy if its just annoying", so it shouldn't impact istio/proxy developers.

We have to put them _somewhere_ else (can be a copy), since istio/istio importing istio/proxy is probably not a good idea. That leaves us istio/api or istio/istio.

IMO istio/api is a better choice, as we already have half the protos here, already have all the tooling setup to generate protos, and it allows folks to import *just* the protos without all of istio/istio.

yangminzhu:
I like this PR for making it auto generated instead of relying on manually generating the proto code. It's always better to be automatic even if these files does not change often.

The tooling changes seems very small in the PR so I guess there should not be too much overhead of it?

howardjohn:
Blocked by https://github.com/istio/tools/pull/1718

howardjohn:
@kyessenov do you have concerns if we do not touch istio/proxy?

kyessenov:
My vote is not binding, feel free to merge. It can always be reverted if it becomes a hassle.

douglas-reid:
These configs are definitely not meant to be user-facing, but we've stored similar schema in istio/api before (mixer client config). 

Is there a way to annotate them, or highlight the implementation-detail-ish nature of them prominently?  Would shoving them into `internal/...` or similar work?

We definitely need to keep them in sync (either through remote deps in istio/proxy, or some sort of sync job if pushing).

Will we need to update whatever is generating the reference docs for them as well (https://istio.io/latest/docs/reference/config/proxy_extensions/) ?

louiscryan (approved):
Approved. Please add the suggested README prior to commit

kyessenov:
@douglas-reid Thanks for reminding me of the horrors of mixerclient protos from here used in proxy bazel (with all the gogo crap).

howardjohn:
README added

douglas-reid (approved):

kyessenov:
These have actually diverged now like I was afraid of, so hopefully there's no user facing repercussions.

PR #2108, opened 2021-09-30
telemetry: add stackdriver logging options to provider

douglas-reid:
This PR would add some (hidden) configuration options for the Stackdriver provider, to control specific logging options.

I expect that the `only_errors` option _may_ ultimately be replaced by a common filtering option at the Telemetry API level (supported across a large number of providers), however the design of such a mechanism and the API for that mechanism are probably at least one release cycle away. In the meantime, this would expose some of the prior mechanisms for controlling Stackdriver logging volumes via the provider.


mesh/v1alpha1/config.proto
```diff
@@ -729,6 +729,19 @@ message MeshConfig {
...
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 5;
+
+      message LoggingOptions {
+        // Turns on gzip-compression of logs when sending to Cloud Logging.
+        google.protobuf.BoolValue compress_logs = 1;
+
+        // For access logs, only generate for requests with a response.code >= 400.
+        // TODO: replace with filter options in Telemetry API (when ready)
+        google.protobuf.BoolValue only_errors = 2;
```
bianpengyuan, line 739: Do we have any policy on deprecating field which is hidden from docs? Wondering how hard would it be to make the switch to filter.

bianpengyuan (commented):
lgtm

mesh/v1alpha1/config.proto
mandarjog: The alternative is to only allow a single expression here and then loosen the restriction?

douglas-reid: I assume that unpublished configuration options have no deprecation policy, but I don't know the official answer.

PR #2110, opened 2021-09-30
telemetry: add provider-specific format options

douglas-reid:
This PR adds a way to configure access log providers (the format of their content) and an access log filter to the Telemetry API.

The configuration of extension providers in mesh configuration is a temporary state. It is expected that the `ExtensionProviders` configuration will move out of `MeshConfig` in a follow-on effort.


zirain:
> ```yaml
> access_logging
> ```

@douglas-reid how about like this?

```yaml
spec:
  access_logging:
    - name: file-log
      file:
        path: /dev/stdout
        format: "[%START_TIME%]"
    - name: grpc-http-log
      gprc:
        grpcService:
          address: grpc-als.istio-system.svc:9811
        additionalRequestHeadersToLog:
        - x-request-user
        additionalResponseHeadersToLog:
        - x-response-cost
    - name: grpc-http-log
      grpcOtel:
        grpcService:
          address: grpc-als.istio-system.svc:9811
```

zirain:
> > @douglas-reid how about like this?
> > ```yaml
> > spec:
> >   access_logging:
> >     - name: file-log
> >       file:
> >         path: /dev/stdout
> >         format: "[%START_TIME%]"
> >     - name: grpc-http-log
> >       gprc:
> >         grpcService:
> >           address: grpc-als.istio-system.svc:9811
> >         additionalRequestHeadersToLog:
> >         - x-request-user
> >         additionalResponseHeadersToLog:
> >         - x-response-cost
> >     - name: grpc-http-log
> >       grpcOtel:
> >         grpcService:
> >           address: grpc-als.istio-system.svc:9811
> > ```
> 
> @zirain i don't think we'd want to put service addresses in the Telemetry API itself. that was the main motivation behind the `ExtensionProviders` configuration (currently in MeshConfig). Referencing those providers is also part of the rest of the Telemetry API, so skipping having `ProviderRefs` would be a major deviation.
> 
> Is it intentional that you have two loggers with the same name and different configuration? What is the expected output of such a configuration?
> 
> It seems like the second `grpc-http-log` has no workload-specific configuration -- unless you intend to have individual workloads send to different backends. Is that the case?

I think we should allow user send envoy access log via grpc to multi backend, especially we should support user customize additional_request_headers_to_log additional_response_headers_to_log per workload. 

BTW, maybe we should allow users config more then one grpc `ExtensionProviders` configuration?

costinm:
My main concern is: not in MeshConfig. We have a telemetry API now, which has proper CRDs that can be set in namespaces and protected with RBAC and managed by telemetry admins, without having to touch the root mesh config. 

The second concern is that we shouldn't be the ones defining how opentelemetry or prometheus is configured. Short term - it may be ok to have a CRD that has a placeholder config - but ideally each project would define and own its APIs and configuration, independent of Istio. 

We now have a mechanism to delegate ( policy attachment in Gateway ).

mandarjog:
The provider config model in meshconfig will migrate out to its own CR if and when it makes sense. The thought it to delay creating a CR/CRD until we exactly know what it should be. If some of this config already has CR, we can use them.
@costinm we are not configuring prometheus. We are configuring how *Envoy* generates prometheus output. If EnvoyPrometheus creates its own CR, we may perhaps use it.

Not defining own format != using their CRs. We can have a passthru format defined by the respective projects or import a snapshot into our proto.

douglas-reid:
@howardjohn @mandarjog @costinm please take another look. I believe this represents the tentative agreement reached in design discussions. Follow-on work will be undertaken to design and define an `ExtensionProviders` API that is distinct from `MeshConfig`.

releasenotes/notes/add-access-logging-filter.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: release-notes/v2
+
+kind: feature
+area: telemetry
+issue: []
+
+releaseNotes:
+- |
+    **Added** experimental support in the Telemetry API for access logging filters. Additional
```
mandarjog, line 9: Nit: accesslog providers.

mandarjog (approved):
Lgtm

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +802,87 @@ message MeshConfig {
...
+        string text_format = 1;
+
+        // Structured format for the envoy access logs. Envoy command operators
+        // can be used as values for fields within the Struct. Values are rendered
+        // as strings, numbers, or boolean values, as appropriate
+        // (see: [format dictionaries](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-dictionaries)).
+        //
+        // Example
+        // ```
+        // json_format:
```
hzxuzhonghu, line 828: jsonFormat

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +802,87 @@ message MeshConfig {
...
+      // This may be used to write to streams, via `/dev/stderr` and `/dev/stdout`
       // If unspecified, defaults to `/dev/stdout`.
       string path = 1;
+
+      message LogFormat {
+        // Textual format for the envoy access logs. Envoy command operators may be
+        // used in the format. The [format string documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-strings)
+        // provides more information.
+        //
+        // Example: `text_format: "%LOCAL_REPLY_BODY%:%RESPONSE_CODE%:path=%REQ(:path)%\n"`
```
hzxuzhonghu, line 818: textFormat

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +802,87 @@ message MeshConfig {
...
+      repeated string filter_state_objects_to_log = 4;
+
+      // Optional. Additional request headers to log.
+      repeated string additional_request_headers_to_log = 5;
+
+      // Optional. Additional response headers to log.
+      repeated string additional_response_headers_to_log = 6;
+
+      // Optional. Additional response trailers to log.
+      repeated string additional_response_trailers_to_log = 7;
```
hzxuzhonghu, line 865: I understand these fields can be different for different services, so i feel it is better to make it under the telemetry api

hzxuzhonghu: This is UX issue, which should be per service.

douglas-reid: I agree that this will be somewhat limiting for per-workload customizations in the current form. It does require editing `MeshConfig` to add providers for each customization needed. However, it is worth considering that this, while not ideal, is still an improvement over the existing mechanism, as customizations are not possible with the existing `MeshConfig`-based config.

This was a compromise, as we don't have an agreed upon mechanism for abstract mechanism for dealing with differing logging formats yet (what does "add this trailer" mean for textual logging formats?). I'm hoping as this progresses from experimental -> alpha -> beta -> GA, we can smooth out the rough edges based on user feedback.

When `ExtensionProviders` becomes a separate, dynamic piece of config (moves out of `MeshConfig`) that should alleviate some of the UX issues. Admins/service owners will hopefully be able to add differently configured providers, and then reference them from the Telemetry resources.

mesh/v1alpha1/config.proto
```diff
@@ -744,6 +745,12 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 5;
+
+      message LoggingOptions {
+        map<string, string> dimensions = 1;
+        google.protobuf.Duration report_duration = 2;
+        google.protobuf.BoolValue errors_only = 3;
```
howardjohn, line 756: I thought we were planning a common CEL filtering?

mesh/v1alpha1/config.proto
```diff
@@ -744,6 +745,12 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 5;
+
+      message LoggingOptions {
+        map<string, string> dimensions = 1;
+        google.protobuf.Duration report_duration = 2;
```
howardjohn, line 755: What makes this different than a `duration` dimensions?

mesh/v1alpha1/config.proto
```diff
@@ -744,6 +745,12 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 5;
+
+      message LoggingOptions {
+        map<string, string> dimensions = 1;
```
howardjohn, line 754: before merging some examples would help

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +802,87 @@ message MeshConfig {
...
       // Path to a local file to write the access log entries.
+      // This may be used to write to streams, via `/dev/stderr` and `/dev/stdout`
       // If unspecified, defaults to `/dev/stdout`.
       string path = 1;
+
+      message LogFormat {
+        // Textual format for the envoy access logs. Envoy command operators may be
+        // used in the format. The [format string documentation](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-strings)
+        // provides more information.
+        //
```
howardjohn, line 821: nit: I recommend we inject the `\n` and don't have users put it there. Otherwise its a big footgun to forget it, and YAML makes trailing newlines hard to insert

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +802,87 @@ message MeshConfig {
...
+        // as strings, numbers, or boolean values, as appropriate
+        // (see: [format dictionaries](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-dictionaries)).
+        //
+        // Example
+        // ```
+        // json_format:
+        //   status: "%RESPONSE_CODE%"
+        //   message: "%LOCAL_REPLY_BODY%"
+        // ```
+        google.protobuf.Struct json_format = 2;
```
howardjohn, line 836: nit: `json` as the name?

Then usage looks like

```
logFormat:
  json: { .... }
```

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +802,87 @@ message MeshConfig {
...
 
+    // Defines configuration for Envoy-based access logging that writes to
+    // local files (and/or standard streams).
     message EnvoyFileAccessLogProvider {
       // Path to a local file to write the access log entries.
+      // This may be used to write to streams, via `/dev/stderr` and `/dev/stdout`
       // If unspecified, defaults to `/dev/stdout`.
       string path = 1;
+
+      message LogFormat {
```
howardjohn, line 817: nit: may want a comment this is a oneof; you cannot set both formats. Or literally make it a oneof perhaps?

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +802,87 @@ message MeshConfig {
...
+        // Textual format for the envoy access logs. Envoy command operators may be
```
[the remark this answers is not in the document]
douglas-reid: updated with reference.

releasenotes/notes/add-access-logging-filter.yaml
douglas-reid: changed this in attempt to clarify.

mesh/v1alpha1/config.proto
douglas-reid: thanks! done.

mesh/v1alpha1/config.proto
douglas-reid: done.

howardjohn (commented):
LGTM

mesh/v1alpha1/config.proto
douglas-reid: we are. i went back and forth on this, as SD doesn't currently function like the other access loggers. i'm ok with removing and taking the AI to make SD accept CEL.

howardjohn: Oh I had thought we already made CEL for SD. I think that is the long term path, if we need this in the interim maybe its fine but hidden from docs?

mesh/v1alpha1/config.proto
douglas-reid: i'm not entirely clear what you mean here?  SD has a configurable logs reporting duration (iirc, defaults to `10s`). This would allow configuring that, if needed. I'm fine with leaving this out for now. I had thought it was used in tests, but I don't see it on a second look.

howardjohn: Oh its the _interval_? Duration confused me - I thought this meant something like "report the duration of a request".

mesh/v1alpha1/config.proto
douglas-reid: done. same with `text`.

mesh/v1alpha1/config.proto
douglas-reid: added comments. ptal.

mesh/v1alpha1/config.proto
douglas-reid: done.

mesh/v1alpha1/config.proto
douglas-reid: added. ptal.

mesh/v1alpha1/config.proto
douglas-reid: removed.

mesh/v1alpha1/config.proto
douglas-reid: removed. will do the work in SD extension.

mesh/v1alpha1/config.proto
costinm: The word 'dimensions': is anyone else using it for logging ? OTel proto has 'attributes', stackdriver 'labels'.

mesh/v1alpha1/config.proto
```diff
@@ -744,6 +749,21 @@ message MeshConfig {
...
+        //
+        // Example:
+        //   dimensions:
+        //     path: request.url_path
+        //     foo: request.headers['x-foo']
+        map<string, string> dimensions = 1;
+      }
+
+      // Optional. Controls Stackdriver logging behavior.
+      LoggingOptions logging_options = 6;
```
costinm, line 766: In my never ending fight with redundant naming - can we use Logging ? Everything in this file is an option ( or config, policy ), no need for the suffixes. At least it's not logging_policy...

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +815,93 @@ message MeshConfig {
...
+          // as strings, numbers, or boolean values, as appropriate
+          // (see: [format dictionaries](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#config-access-log-format-dictionaries)).
+          //
+          // Example
+          // ```
+          // json:
+          //   status: "%RESPONSE_CODE%"
+          //   message: "%LOCAL_REPLY_BODY%"
+          // ```
+          google.protobuf.Struct json = 2;
```
costinm, line 848: Since 'dimensions' ( or attributes, labels ) are defined in the top level - do we need to duplicate it here ? Isn't json just an encoding of the structured dimensions from above ? 

And if that is the case - can we leave json as the default ( if text is not specified ), and not bother with this redundant setting ?

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +815,93 @@ message MeshConfig {
...
+        }
+      }
+
+      // Optional. Allows overriding of the default text log format.
+      LogFormat log_format = 2;
```
costinm, line 853: Also - given that 'text' format would be common to multiple implementations - I wouldn't mind having it top level, with the semantic: if text is provided, will be used as format ( if the driver allows ), otherwise the default structured output of the provider is used ( json, proto, etc) with the defined dimensions/labels/attributes.

mesh/v1alpha1/config.proto
howardjohn: Dimension is in stackdriver message, this is the envoy one. They are distinct so the config is never shared

howardjohn (approved):
My comments were addressed and I am happy with this so marking my approval. I added a hold since there were other comments

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +815,93 @@ message MeshConfig {
...
+    // Defines configuration for an Envoy [Access Logging Service](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/grpc/v3/als.proto#grpc-access-log-service-als)
+    // integration for HTTP traffic.
+    message EnvoyHttpGrpcV3LogProvider {
+      // REQUIRED. Specifies the service that implements the Envoy ALS gRPC authorization service.
+      // The format is `[<Namespace>/]<Hostname>`. The specification of `<Namespace>` is required only when it is insufficient
+      // to unambiguously resolve a service in the service registry. The `<Hostname>` is a fully qualified host name of a
+      // service defined by the Kubernetes service or ServiceEntry.
+      //
+      // Example: "envoy-als.foo.svc.cluster.local" or "bar/envoy-als.example.com".
+      string service = 1;
```
costinm, line 865: Do we use this style in the other APIs ? I would expect 'service' to be the name of a k8s service that has port inside, and port usually is name or string. But also I think we use host:port commonly - in particular in gRPC addresses.

costinm: And also 'service' - we have discoveryAddress, etc. 

service instead of Address is a bit confusing given the 100 uses of service.

mesh/v1alpha1/config.proto
costinm: I agree with both of you - we should make incremental progress and expose each provider options ( no major harm ), but at the same time for very common things - as the selection of attributes, a text format, filter - we should use the top level.

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +815,93 @@ message MeshConfig {
...
+
+      // REQUIRED. Specifies the port of the service.
+      uint32 port = 2;
+
+      // REQUIRED. The friendly name of the access log
```
costinm, line 900: It's not very friendly to have required things - for the port too. Why not defaults ?

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +815,93 @@ message MeshConfig {
...
+      string log_name = 3;
+
+      // Optional. Additional filter state objects to log.
+      repeated string filter_state_objects_to_log = 4;
```
costinm, line 904: Same as above - filters should probably be outside of the provider-specific config.

And my English is not strong enough to understand this name or the comment. What state ? What objects ? and to what log - isn't this the config for logging ?

costinm (commented):
Despite my comments - it looks good in general, and as an alpha API we can still iterate a bit.

mesh/v1alpha1/config.proto
```diff
@@ -506,6 +507,10 @@ message MeshConfig {
       PrometheusMetricsProvider prometheus = 10;
       // Configures an Envoy File Access Log provider.
       EnvoyFileAccessLogProvider envoy_file_access_log = 11;
+      // Configures an Envoy Access Logging Service provider for HTTP traffic.
+      EnvoyHttpGrpcV3LogProvider envoy_http_als = 12;
```
costinm, line 511: Do we need both, would a user configure envoy grpc logging just for TCP and not HTTP ? And is 'als' an acronym commonly used ? 

Also for the one above and all other mentions of "envoy" in the names - is this configuration exclusive to Envoy in any way ? Even if envoy defines a gRPC service for logging ( and one for config - XDS ), it doesn't mean they are specific to envoy. Same for file access log above.

The API should be supportable on proxyless gRPC or uProxy or other non-envoy based implementations, it's an Istio API.

howardjohn: I think it can still be reasonable to call it 'Envoy' despite it being used by non-Envoy things. It is Envoy's defined API (ie defined in https://github.com/envoyproxy/), but it is used by many. The same applies to all the APIs they have really - XDS, ext_authz, rate limit, ALS, etc

Also fine with another name if there is a reasonable one though.

costinm: For example we could use
- AccessLogGrpc ( or whatever is the service name - if it doesn't include envoy in the name)
- FileAccessLog
- OpenTelementyLog

howardjohn: OpenTelemetryLog is really not much different than EnvoyLog other than "Envoy"  being overloaded as an API and a Proxy, at least how I see it

I am glad we have enough agreement that only the names are up for debate 🙂

costinm: I guess we use "EnvoyAuthorizationProvider" - but it's a pretty bad naming for an API or service. I guess if it means "don't use except for envoy, use otel or open standards instead" it's fine...

mesh/v1alpha1/config.proto
costinm: Sure, but they can move up - and use the same word for both. Virtually any provider would support some attributes/labels, and we don't need to use all possible synonyms in our config file.

costinm:
I'm more concerned with the structure than the name actually, ok to use
Envoy prefix to everything that is envoy-only,
at least will eliminate confusion on what will be supported outside envoy.

But for structure - at least have 'attributes' at the top level and used
for any/app loggers that support structured output,
and same for the text format. Which we already had for a long time -
access_log_format, no need to deprecate or replace it !

mesh/v1alpha1/config.proto
douglas-reid: i really don't care at all about the provider-specific naming here. `dimensions` merely matches the existing stackdriver extension configuration terminology. if we feel `labels` is clearer, that's perfectly fine.

mesh/v1alpha1/config.proto
douglas-reid: we can't move it up without agreeing on how to universally represent formats. that was the other proposal, which was rejected.

mesh/v1alpha1/config.proto
douglas-reid: we use this exact style/convention for the rest of the `ExtensionProviders`. it was required when we first started adding them. let's not re-open that can of worms for this PR.

mesh/v1alpha1/config.proto
douglas-reid: We can use a default, if we agree on one. The underlying extension config makes this a required field, so I just duplicated here.

mesh/v1alpha1/config.proto
costinm: Stackdriver https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry seems to call them labels. 

By primary concern here is that this should be top level, and used by any provider that supports selection of attributes/labels.

mesh/v1alpha1/config.proto
douglas-reid: this is why it is in the extension-specific config. `filter` here and `filter_state_objects` are Envoy terms. It doesn't make sense to move it up, as it is not universal. `to_log` means to include in the generated logging record.

mesh/v1alpha1/config.proto
costinm: We don't have to agree on an universal representation of formats. 

Just that all providers that need to select a set of attributes to include will use the same common config. Similar for the filter. 

Middle ground is better - beside using different words, all providers have a selection of key/value pairs to include.

mesh/v1alpha1/config.proto
douglas-reid: These aren't really part of the Telemetry API. They are referenced by the API. I think it is completely fine to name the configuration for specific providers after the implementation of those providers. In this case, the Envoy extensions take differing configuration.

We can abstract them, if needed/desired, when we pull `ExtensionProviders` out to a separate API.

mesh/v1alpha1/config.proto
costinm: Ok, got it. This is the problem with Istio APIs- we rush something in as experimental ( authz provider ), and then it becomes the pattern everything else must follow. One more reason to adopt K8S or open APIs wherever possible and not move MeshConfig out of alpha...

mesh/v1alpha1/config.proto
costinm: How about we at least use 'labels' or 'attributes' in all providers - here and in the 'json' ( which is actually used to select what to include ). And we keep the structure as it is until k8s or otel defines a proper config ?

mesh/v1alpha1/config.proto
costinm: See comment above - ok to keep in in this structure, but at least let's use the same name (attributes or labels).

mesh/v1alpha1/config.proto
costinm: Is this the literal envoy config ? Can you add a comment indicating this is equivalent with having an EnvoyFilter, and the config is exactly what is configured in envoy ? ( if that's the case )
And this is how they comment their field ?

douglas-reid:
> But for structure - at least have 'attributes' at the top level and used for any/app loggers that support structured output, and same for the text format. Which we already had for a long time - access_log_format, no need to deprecate or replace it !

That plan still doesn't allow for configuration of ALS extensions. Having to map into that configuration from generic attributes will prove challenging, if not ultimately futile (`istiod` having to parse struct fields to determine if they reference a request header or response trailer, etc.). Which means we would likely have multiple configuration types in the top-level, each of which maps to a ~one provider. That design, fwiw, was an alternate proposal which was rejected earlier.

This PR allows room for follow-on work that add fields to the Telemetry API for format control.

mesh/v1alpha1/config.proto
costinm: Ok, never mind - for all EnvoyXXX we can keep them as complicated as they are, they should die eventually and complexity in config helps...

mesh/v1alpha1/config.proto
costinm: Resolving - same as the other one, I no longer have any objections for anything in an Envoy named proto.

mesh/v1alpha1/config.proto
costinm: Same comment - if this is what envoy specific config has, we should mirror it. 

Please just add a comment on each of them indicating this is a copy of Envoy config and expected to only be used with envoy proxies.

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +815,93 @@ message MeshConfig {
 
     message PrometheusMetricsProvider {}
 
+    // Defines configuration for Envoy-based access logging that writes to
+    // local files (and/or standard streams).
```
costinm, line 819: I resolved my other comments related to envoy-specific stuff - but file based access log is in no way specific to envoy and we already have fields in the top level config, there is no need to deprecated them or create confusion. 

Can you at least leave this as "FileAccessLogProvider" and mention that the log format, path are the existing ones in the MeshConfig top level ?

mesh/v1alpha1/config.proto
```diff
@@ -506,6 +507,10 @@ message MeshConfig {
       PrometheusMetricsProvider prometheus = 10;
       // Configures an Envoy File Access Log provider.
       EnvoyFileAccessLogProvider envoy_file_access_log = 11;
+      // Configures an Envoy Access Logging Service provider for HTTP traffic.
+      EnvoyHttpGrpcV3LogProvider envoy_http_als = 12;
+      // Configures an Envoy Access Logging Service provider for TCP traffic.
+      EnvoyTcpGrpcV3LogProvider envoy_tcp_als = 13;
```
costinm, line 513: No OTel provider ? Given how many envoy-specific options we have, maybe we should have at least a neutral one...

mesh/v1alpha1/config.proto
douglas-reid: I'm not talking about Stackdriver the backend. I'm talking about the Envoy Stackdriver extension, which is the provider being configured.  See: https://github.com/istio/api/blob/master/envoy/extensions/stackdriver/config/v1alpha1/config.proto#L38.

I'm fine with using `labels` here.  `json`, on the other hand, i'm less sure about. it provides information about the format of the output. I'll update and add documentation to see if we think that is clearer.

costinm:
The format of the output ( json or text ) is already specified at top
level, and I don't think for file access log we need to deprecate/move
that. We have the enum with format type.

If it is json - the selection of attributes to include can use the same
'labels/attributes' semantic.

The stackdriver extension is our own implementation - with terms probably
from Mixer NIH times,
I don't think envoy filter implementation should override the naming used
in the real protocol/provider,
any stackdriver user would probably be confused if they show up in the API.

mesh/v1alpha1/config.proto
douglas-reid: I'm happy to go back and add it, but it had not been specifically requested to date (the envoy extension, fwiw, is still not considered production ready). Since there was concern about adding a bunch to `MeshConfig`, it seemed safer to go with the minimal requested set for 1.13. We can always add in a separate PR.

mesh/v1alpha1/config.proto
douglas-reid: You want me to change the existing name of the provider message?  

I'll add comments about these formats matching the `MeshConfig` ones.

The goal here is to make the top-level fields more flexible, in a way.  Support more than a monolithic format, and allow selection per namespace or workload (or per operational mode), and support dynamic updates. Ultimately, we plan to migrate usage from those fields to wherever ExtensionProviders live.

mesh/v1alpha1/config.proto
douglas-reid: It isn't common to the other (Istio-supported) implementations (ALS and Stackdriver), strangely enough.

mesh/v1alpha1/config.proto
douglas-reid: Updated both to be `labels`. PTAL.

mesh/v1alpha1/config.proto
douglas-reid: done.

douglas-reid:
We need to move the file access log into a provider for several reasons:
- it is how the rest of the API works (we need a reference to a provider)
- it allows for multiple mesh-admin-configured formats
- it allows for dynamic selection of the formats via provider selection

mesh/v1alpha1/config.proto
```diff
@@ -795,10 +815,96 @@ message MeshConfig {
...
+      //
+      // Example: "envoy-als.foo.svc.cluster.local" or "bar/envoy-als.example.com".
+      string service = 1;
+
+      // REQUIRED. Specifies the port of the service.
+      uint32 port = 2;
+
+      // Optional. The friendly name of the access log.
+      // Defaults to `istio-proxy`.
+      string log_name = 3;
```
hzxuzhonghu, line 873: can you explain what  it is used for

hzxuzhonghu: got it,  seems the default value violates currently used by istio
```
	httpEnvoyAccessLogFriendlyName     = "http_envoy_accesslog"
	tcpEnvoyAccessLogFriendlyName      = "tcp_envoy_accesslog"
	listenerEnvoyAccessLogFriendlyName = "listener_envoy_accesslog"
	httpEnvoyAccessLogFriendlyName     = "http_envoy_accesslog"
	tcpEnvoyAccessLogFriendlyName      = "tcp_envoy_accesslog"
	listenerEnvoyAccessLogFriendlyName = "listener_envoy_accesslog"
```

douglas-reid: thanks! I had forgotten that. fixed.

hzxuzhonghu (approved):

mesh/v1alpha1/config.proto
costinm: I don't think 'migrate usage' is the best option for users. We'll have to keep them working until a stable API emerges ( possibly outside of Istio ), too many people are using them and upgrades must be fully automated.

mesh/v1alpha1/config.proto
costinm: It doesn't have to be in all - if a workload is using a driver that supports text, it should use the format, otherwise it can be ignored.

There is a big reason for having a single text format - when access logs are collected/aggregated, it is very hard to process them if the text format is different between workloads.

costinm (approved):

howardjohn:
@douglas-reid I think good to remove to hold I added whenever you are ready

mesh/v1alpha1/config.proto
douglas-reid: I think we are aligned here. Support is not being dropped. And automation around upgrades, etc. is being planned / designed. `migrate usage` was my term for the (automated) management of that transition.

douglas-reid:
> @douglas-reid I think good to remove to hold I added whenever you are ready

Let's do it. There is a lot of work left to do here for this API, but I think this puts in a good foundation for 1.13. Having this in will allow us to explore more early in the dev cycle.

PR #2126, opened 2021-10-15
Move the Annotation to istio/api

xichengliudui:
ref: https://github.com/istio/istio/pull/35617


xichengliudui:
@Monkeyanator  
what is the `buf` command
```sh
./gen.sh: line 23: buf: command not found
```

PR #2129, opened 2021-10-20
Add verbose flag for envoy tracing

commixon:
Here we add the `verbose` flag for envoy tracing.

As it is right now, istio annotates spans with additional data by default, which will produce sometimes a lot more "span events" than intended. This is an easy way to control span "verbosity" which is already supported in envoy.

Reference: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html?highlight=verbose


commixon:
Ping comment to maybe get some traction? :)

vlinevych:
I am interested in this feature as well 👍

PR #2131, opened 2021-10-27
Add AlpnProtocols to Gateway API

liuxu623:
Fix https://github.com/istio/istio/issues/13578 https://github.com/istio/istio/issues/14708 https://github.com/istio/istio/issues/25081 

Allow to config alpnProtocols in ServerTLSSettings like
```yaml
kind: Gateway
metadata:
  name: mygateway
spec:
  selector:
    istio: ingressgateway # use istio default ingress gateway
  servers:
  - port:
      number: 443
      name: https
      protocol: HTTPS
    tls:
      mode: SIMPLE
      credentialName: httpbin-credential # must be the same as secret
      alpnProtocols:
      - h2
      - http/1.1
    hosts:
    - httpbin.example.com
```


zengyuxing007:
Any progress on this? @liuxu623

guydc:
It seems that they gateway-api consensus is to not represent HTTP3 as a distinct protocol: https://github.com/kubernetes-sigs/gateway-api/issues/687. Istio’s experimental support for HTTP3 is also aligned with this state: https://github.com/istio/istio/wiki/Experimental-QUIC-and-HTTP-3-support-in-Istio-gateways. 

Considering these updates, the proposal by @albertlockett to use the new `alpn_protocols` to [override ALPN settings](https://github.com/istio/istio/issues/44729) only for non-QUIC/ISITO_MUTUAL sounds reasonable. @liuxu623, @mandarjog  - WDYT?

AnPiakhota:
Why has it been closed? Is there any way to force Istio to use HTTP 1.1 protocol only?

wtzhang23:
FYI I'm happy to continue off this work.

wtzhang23:
@howardjohn Created demo here:
* https://github.com/wtzhang23/istio/pull/new/support-custom-server-alpns
* https://github.com/istio/api/compare/master...wtzhang23:istio-api:specify-server-alpns?expand=1
If the Istio group thinks this change is ok, I can open them as PRs and we can progress with them further (Also, for future reference, if y'all are ok with me opening PRs directly, I can do that in the future).

PR #2134, opened 2021-10-28
update jwt based routing

yangminzhu:
This PR changes the format of the special header used for JWT claim based routing in virtual service from `x-jwt-claim...` to `@request.auth.claims...`:

1. The use of the `@` makes the header name drastically different from a normal real HTTP header. This change is made to:

    - Prevent possible customer confusion of the special header name with real HTTP headers, and any security and usability issues caused by the confusion.

    - Prevent potential collision with existing values because a customer might already be using the same header in the virtual service.

    The `@` makes it invalid for HTTP header which is intentionally to make sure it is never used by customer before and will never collide with existing HTTP headers. For example, Envoy and google.com will simply reject the header using `@` with `400` code. This also makes sure the user will not confuse it with real HTTP headers.

2. The use of the `request.auth.claims` is to align with the existing attribute (`request.auth.claims`) in the authorization policy for JWT claim matching (https://istio.io/latest/docs/reference/config/security/conditions/#supported-conditions). By reusing the same attribute (under the hood they are actually also sharing the same matching code), this helps to reduce the user learning efforts for understanding and adopting the feature as they might be already familiar with the name. And this makes our API consistent across different places.

@howardjohn @costinm @louiscryan PTAL, thanks.

### DIFF
--- a/security/v1beta1/request_authentication.proto
+++ b/security/v1beta1/request_authentication.proto
@@ -208,7 +208,7 @@ option go_package="istio.io/api/security/v1beta1";
 //   - name: "v2-route"
 //     match:
 //     - headers:
-//         @request.auth.claims.version:
+//         "@request.auth.claims.version":
 //           exact: "v2"
 //     route:
 //     - destination:
--- a/security/v1beta1/request_authentication.proto
+++ b/security/v1beta1/request_authentication.proto
@@ -156,19 +156,16 @@ option go_package="istio.io/api/security/v1beta1";
 //         paths: ["/healthz"]
 // ```
 //
-// - [Experimental] When applied on a Gateway, a special attribute `@request.auth.claims` will be available for using
-// in the header field of the virtual service for JWT claim based routing. The prefix `@` indicates that it is
-// matching with special attributes derived from validated JWT and not with normal HTTP headers.
+// [Experimental] Routing based on internally derived [metadata](https://istio.io/latest/docs/reference/config/security/conditions/)
+// is now supported. A prefix '@' is used to denote a match against internal metadata instead of the headers in the request.
+// Currently this feature is only supported for the following metadata:
 //
-// Currently JWT claims of type string, list of string and nested claims are supported. The claim name itself must
-// not include the `.` character for now.
-// Examples: `@request.auth.claims.admin` refers to the claim "admin" and `@request.auth.claims.group.id` refers to the nested claims "group" and "id".
+// - `request.auth.claims.{claim-name}[.{sub-claim}]*` which are extracted from validated JWT tokens. The claim name
+// currently does not support the `.` character. Examples: `request.auth.claims.name` and `request.auth.claims.group.id`.
 //
-// **Note:** The JWT claim routing and the special attribute is an experimental feature and only supported on Gateways. The
-// request authentication must first be applied on gateways to validate the JWT to make the JWT claims available for routing.
-//
-// The following example creates the request authentication and authorization policy for JWT validation on ingress
-// gateway and virtual service for routing based on the "version" JWT claim.
+// You could use the `request.auth.claims` attribute for JWT claim based routing on gateways. The following example creates
+// the request authentication and authorization policy for JWT validation on ingress gateway and virtual service for
+// routing based on the "version" JWT claim.
 //
 // ```yaml
 // apiVersion: security.istio.io/v1beta1
--- a/security/v1beta1/request_authentication.proto
+++ b/security/v1beta1/request_authentication.proto
@@ -156,14 +156,13 @@ option go_package="istio.io/api/security/v1beta1";
 //         paths: ["/healthz"]
 // ```
 //
-// - [Experimental] When applied on a Gateway, a special attribute `@request.auth.claims[...]` will be available to be
-// used in the header field in the virtual service route matching. This can be used for routing based on JWT claims.
-// The special prefix `@` indicates that it is matching with attributes derived from validated JWT and not with normal
-// HTTP headers.
+// - [Experimental] When applied on a Gateway, a special attribute `@request.auth.claims` will be available for using
+// in the header field of the virtual service for JWT claim based routing. The prefix `@` indicates that it is
+// matching with special attributes derived from validated JWT and not with normal HTTP headers.
 //
-// JWT claims of type string or list of string are supported and nested claims are also supported.
-// Examples: `@request.auth.claims[admin]` refers to the claim "admin" and
-// `@request.auth.claims[group][id]` refers to the nested claims "group" and "id".
+// Currently JWT claims of type string, list of string and nested claims are supported. The claim name itself must
+// not include the `.` character for now.
+// Examples: `@request.auth.claims.admin` refers to the claim "admin" and `@request.auth.claims.group.id` refers to the nested claims "group" and "id".
 //
 // **Note:** The JWT claim routing and the special attribute is an experimental feature and only supported on Gateways. The
 // request authentication must first be applied on gateways to validate the JWT to make the JWT claims available for routing.
@@ -212,7 +211,7 @@ option go_package="istio.io/api/security/v1beta1";
 //   - name: "v2-route"
 //     match:
 //     - headers:
-//         @request.auth.claims[version]:
+//         @request.auth.claims.version:
 //           exact: "v2"
 //     route:
 //     - destination:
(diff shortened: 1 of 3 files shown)


security/v1beta1/request_authentication.proto
```diff
@@ -156,13 +156,18 @@ option go_package="istio.io/api/security/v1beta1";
 //         paths: ["/healthz"]
 // ```
 //
-// - When applied on a Gateway, you can also use the special header name `x-jwt-claim` for matching JWT claims in
-// the VirtualService. Claims of type string or list of string are supported and nested claims are also supported using
-// `.` as a separator for claim names. Examples: `x-jwt-claim.admin` matches the claim "admin" and `x-jwt-claim.group.id`
-// matches the nested claims "group" and "id".
+// - [Experimental] When applied on a Gateway, a special variable `{request.auth.claims[...]}` will be available for use in
```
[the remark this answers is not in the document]
yangminzhu: We discussed this in the gateway community and I believe the community is preferring a dedicated first-class field for the JWT and extension for vendor specific matching.

I have updated the PR to use the format `@request.auth.claims.key`, the `@` makes it different from the real HTTP header and the remaining part will eventually be converted and supported in a first-class field in gateway API, note since we use the dot syntax, it will NOT support claim name itself with dot.

@costinm @louiscryan could you take a look again? thanks.

security/v1beta1/request_authentication.proto
```diff
@@ -156,13 +156,19 @@ option go_package="istio.io/api/security/v1beta1";
 //         paths: ["/healthz"]
 // ```
 //
-// - When applied on a Gateway, you can also use the special header name `x-jwt-claim` for matching JWT claims in
-// the VirtualService. Claims of type string or list of string are supported and nested claims are also supported using
-// `.` as a separator for claim names. Examples: `x-jwt-claim.admin` matches the claim "admin" and `x-jwt-claim.group.id`
-// matches the nested claims "group" and "id".
+// - [Experimental] When applied on a Gateway, a special attribute `@request.auth.claims` will be available for using
```
louiscryan, line 159: I would document this a little differently. Something like

[Experimental] Routing based on internally derived metadata (linkify) is now supported. A prefix '@' is used to denote a match against internal metadata instead of the headers in the request. Currently this feature is only supported for the following metadata 

request.auth.claims.{claim-name}[.{sub-claim}]

which are extracted from validated JWT tokens <link to schema definition in envoy reference for attributes.proto>

Examples: request.auth.claims.group.id, ...

yangminzhu: thanks for the review, updated.

security/v1beta1/request_authentication.proto
```diff
@@ -156,19 +156,16 @@ option go_package="istio.io/api/security/v1beta1";
...
-// Examples: `@request.auth.claims.admin` refers to the claim "admin" and `@request.auth.claims.group.id` refers to the nested claims "group" and "id".
+// - `request.auth.claims.{claim-name}[.{sub-claim}]*` which are extracted from validated JWT tokens. The claim name
+// currently does not support the `.` character. Examples: `request.auth.claims.name` and `request.auth.claims.group.id`.
 //
-// **Note:** The JWT claim routing and the special attribute is an experimental feature and only supported on Gateways. The
-// request authentication must first be applied on gateways to validate the JWT to make the JWT claims available for routing.
-//
-// The following example creates the request authentication and authorization policy for JWT validation on ingress
-// gateway and virtual service for routing based on the "version" JWT claim.
+// You could use the `request.auth.claims` attribute for JWT claim based routing on gateways. The following example creates
```
louiscryan, line 166: Avoid the use of 'could' in user-facing documentation. Is the intent to say that this feature is only available for Gateways ? Is so then a more prescriptive way to write this would be:

"The use of matches against JWT claim metadata is only supported in Gateways"

yangminzhu: this is better, thanks for the suggestion, updated.

yangminzhu:
@louiscryan @costinm could you take a look again? thanks.

security/v1beta1/request_authentication.proto
```diff
@@ -156,13 +156,16 @@ option go_package="istio.io/api/security/v1beta1";
 //         paths: ["/healthz"]
 // ```
 //
-// - When applied on a Gateway, you can also use the special header name `x-jwt-claim` for matching JWT claims in
-// the VirtualService. Claims of type string or list of string are supported and nested claims are also supported using
-// `.` as a separator for claim names. Examples: `x-jwt-claim.admin` matches the claim "admin" and `x-jwt-claim.group.id`
-// matches the nested claims "group" and "id".
+// [Experimental] Routing based on internally derived [metadata](https://istio.io/latest/docs/reference/config/security/conditions/)
+// is now supported. A prefix '@' is used to denote a match against internal metadata instead of the headers in the request.
```
costinm, line 160: remove 'internal' - just 'metadata'.

security/v1beta1/request_authentication.proto
```diff
@@ -156,13 +156,16 @@ option go_package="istio.io/api/security/v1beta1";
...
+// Currently this feature is only supported for the following metadata:
 //
-// The following example creates the request authentication and authorization policies for JWT validation on ingress
-// gateway and routes requests based on the "version" claim in the validated JWT.
+// - `request.auth.claims.{claim-name}[.{sub-claim}]*` which are extracted from validated JWT tokens. The claim name
+// currently does not support the `.` character. Examples: `request.auth.claims.name` and `request.auth.claims.group.id`.
```
costinm, line 164: Very few providers have a 'name' claim - can you use one of the commonly used claims in the example ( maybe from https://www.iana.org/assignments/jwt/jwt.xhtml#claims )

'sub' is probably most used.

For 'group' - AFAIK there is a 'groups' claim in rfc7643, but it has an array value and 'value' instead of id. Maybe find another standard claim that is nested ? Or is there a vendor that only has a group in the JWT with id ?

security/v1beta1/request_authentication.proto
```diff
@@ -205,7 +208,7 @@ option go_package="istio.io/api/security/v1beta1";
 //   - name: "v2-route"
 //     match:
 //     - headers:
-//         x-jwt-claim.version:
+//         "@request.auth.claims.version":
```
costinm, line 211: missed this in the original review - I don't think 'vesrsion' is a standard claim either.

costinm (commented):
Looks good in general - the docs are certainly better explaining the experimental nature, and the @ prefix avoids potential problems.

security/v1beta1/request_authentication.proto
yangminzhu: I can change this but I want to clarify we support all claim names, not just standard claim, a lot identity provider and customer-issued JWT includes custom claim names.

security/v1beta1/request_authentication.proto
yangminzhu: this is just an example, many identity providers and customer issued JWT uses non standard claim names, even the example in JWT RFC (https://www.rfc-editor.org/rfc/rfc7519.html#section-3.1) uses non standard claim names...

yangminzhu: we can use `request.auth.claims.name.givenName`: https://www.rfc-editor.org/rfc/rfc7643#section-8.2

security/v1beta1/request_authentication.proto
```diff
@@ -156,13 +156,16 @@ option go_package="istio.io/api/security/v1beta1";
...
+// is now supported. A prefix '@' is used to denote a match against internal metadata instead of the headers in the request.
+// Currently this feature is only supported for the following metadata:
 //
-// The following example creates the request authentication and authorization policies for JWT validation on ingress
-// gateway and routes requests based on the "version" claim in the validated JWT.
+// - `request.auth.claims.{claim-name}[.{sub-claim}]*` which are extracted from validated JWT tokens. The claim name
+// currently does not support the `.` character. Examples: `request.auth.claims.sub` and `request.auth.claims.name.givenName`.
+//
+// The use of matches against JWT claim metadata is only supported in Gateways. The following example creates
+// the request authentication and authorization policy for JWT validation on ingress gateway and virtual service for
```
louiscryan, line 167: I would spell out why all 3 resources are needed. E.g

The following example shows:
- RequestAuthentication <link> to decode and validate a JWT
- AuthorizationPolicy <link> to check for valid principals in the request
- VirtualService to route the request based on a claim

howardjohn (commented):
lgtm

security/v1beta1/request_authentication.proto
yangminzhu: thanks, updated.

yangminzhu:
@costinm @louiscryan ptal, thanks

costinm (approved):

louiscryan (approved):
LGTM

PR #2138, opened 2021-11-02
Align ProxyConfig merge semantics with other Istio resources

Monkeyanator:
Previously, the defined merge semantics with multiple PC CRs in the same namespace or matching the same workload was to merge with no guaranteed precedence. This does not align with other Istio CRs: usually, we document that it's only valid to have 1 workload / namespace and then take the oldest.


Monkeyanator:
Bump @costinm @howardjohn when you have the chance, this aligns the API description with what actually ended up merging for the ProxyConfig impl.

howardjohn (approved):

ericvn (approved):

PR #2148, opened 2021-11-10
Add ProxyImage to defaultConfig

mandarjog:
Add ProxyImage to defaultConfig and ProxyConfig.
This overrides and eventually takes place of the values API.

Signed-off-by: Mandar U Jog <mjog@google.com>


networking/v1beta1/proxy_config.proto
```diff
@@ -118,4 +124,32 @@ message ProxyConfig {
...
+
+  // The release tag of the image.
+  // example: 1.11.1
+  string tag = 2;
+
+  // The image type of the image.
+  // Istio publishes base(""), debug and distroless images.
+  // Other values are allowed if those image types (example: centos) are published to the specified hub.
+  // supported values: ""(default),  debug, distroless.
+  string image_type = 3;
```
bianpengyuan, line 150: Should we make this an enum?

networking/v1beta1/proxy_config.proto
```diff
@@ -118,4 +124,32 @@ message ProxyConfig {
...
+
+  // $hide_from_docs
+  // image_name: proxyv2 (default)
+  string image_name = 4;
```
bianpengyuan, line 154: What could this value be other than proxyv2?

networking/v1beta1/proxy_config.proto
mandarjog: Enum will be too restrictive. If a user or a vendor rolls centos image, they need to wait for api change.

networking/v1beta1/proxy_config.proto
mandarjog: We could get rid of it now, instead of hide from docs.

mesh/v1alpha1/proxy.proto
```diff
@@ -521,6 +522,9 @@ message ProxyConfig {
   // The plugin certificates (the 'cacerts' secret), self-signed certificates (the 'istio-ca-secret' secret)
   // are added automatically by Istiod.
   repeated string ca_certificates_pem = 34;
+
+  // Specifies the details of the proxy image.
+  istio.networking.v1beta1.ProxyImage image = 35;
```
Monkeyanator, line 527: General question: do we want to keep adding fields to meshConfig.defaultConfig? In versions with the ProxyConfig CR users should use the global CR in the root namespace.

howardjohn: Wouldn't it need to be here since its the proto we use internally? Perhaps hide_from_docs, but I would think this one needs to be a superset of the ProxyConfig CR?

mandarjog: In order make migration possible our hierarchy is

meshconfig.defaultConfig (This should eventually become the same proto that is `istio.networking.v1beta1.ProxyConfig` ) -->  proxyconfig-cr.istio-system --> workload specific etc 

This means that the "installation-admin" can have say `distroless` as the default, but you can still override it.

howardjohn (commented):
LGTM once image_name is removed

mandarjog:
@costinm  @howardjohn 

1.  We *can* move `meshConfig.defaultConfig.image` --> `meshConfig.proxyImage`
2.  `minimal` vs `distroless`, it would be difficult to map one image to another, without the mapping.


Alternative is to introduce ProxyClass, in this case `ProxyBootstrapClass` is more apt since these properties need to be present *before* the proxy starts. This is based on the named provider mechanism used in telemetry API/
 
It is initially backed by .Values to ease migration.
`default` class comes from values, if not explicitly defined.
Mode properties from the `sidecar.istio.io/*` annotations will apply here.

```
meshConfig {

defaultProxyClass: minimal

proxyClasses: 
 - name: default
    tag: 1.11
    hub:  github.com/istio
    image_type: ""
    
- name: minimal
    tag: 1.11
    hub: github.com/istio
    image_type: "distroless"
- name:  trafficDirector
    tag: 1.11
    hub: github.com/istio
    image_type: "distroless"
    discoveryAddress: trafficdirector.googleapis.com
    bootstrap_path: /tmp/gc_bootstrap
- name:  mcp
    tag: 1.11
    hub: github.com/istio
    image_type: "distroless"
    discoveryAddress: meshconfig.googleapis.com 
]
}
```
1. Is there inheritance or a prototype of a `boostrapClass` ? otherwise when upgrading, we will need to update tag from many elements.
2. The number of properties is large. and we would like to keep most of these out of `proxyConfig`

howardjohn:
Thanks for the alternative. I personally do not like the 'class' idea much here. The main goal of ProxyConfig was to move things _out_ of mesh config, and to configure properties of the proxy. ProxyImage is almost the canonical example of the type of thing that belongs in ProxyConfig, IMO. If we don't think it does then I am a bit concerned about the future of ProxyConfig.

Since I imagine someone will mention 'this doesn't work for proxyless' - that is fine. An API doesn't have to work for all cases. We have a clear demand to customize the proxy sidecar and should expose that, even if it cannot be used in 100% of usages. Much like HTTP header routing doesn't work for TCP.

networking/v1beta1/proxy_config.proto
```diff
@@ -118,4 +124,28 @@ message ProxyConfig {
...
+}
+
+// The following values are used to construct proxy image url.
+// $hub/$image_name/$tag-$image_type
+// example: docker.io/istio/proxyv2:1.11.1 or docker.io/istio/proxyv2:1.11.1-distroless
+// This information was previously part of the Values API.
+message ProxyImage {
+
+  // The repository of the released image.
+  // hub: docker.io/istio (default), gcr.io/gke-release/asm, or a private repository.
```
howardjohn, line 139:
```suggestion
  // example: docker.io/istio (default), gcr.io/istio-release, or a private repository.
```

networking/v1beta1/proxy_config.proto
```diff
@@ -118,4 +124,28 @@ message ProxyConfig {
...
+  string hub = 1;
+
+  // The release tag of the image.
+  // example: 1.11.1
+  string tag = 2;
+
+  // The image type of the image.
+  // Istio publishes base(""), debug and distroless images.
+  // Other values are allowed if those image types (example: centos) are published to the specified hub.
+  // supported values: ""(default),  debug, distroless.
```
howardjohn, line 149:
```suggestion
  // supported values: default, debug, distroless.
```

And then make empty string == default, like how enums would?

networking/v1beta1/proxy_config.proto
howardjohn: The UX for users over yaml is identical anyways so I think string is fine?

howardjohn (approved):

ericvn (approved):

mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
```diff
@@ -2520,6 +2520,17 @@ <h2 id="ProxyConfig">ProxyConfig</h2>
...
 are added automatically by Istiod.</p>
 
+</td>
+<td>
+No
+</td>
+</tr>
+<tr id="ProxyConfig-image">
+<td><code>image</code></td>
+<td><code><a href="https://istio.io/docs/reference/config/networking/proxy-config.html#ProxyImage">ProxyImage</a></code></td>
```
ericvn, line 2530: This PR breaks istio.io link checking when this link was added.

PR #2160, opened 2021-11-18
Change dr MaxConnection data type from int32 to uint32

ningyougang:
- [x] Change dr tcpSetting's MaxConnection int type from int32 to uint32
  
  Refer to: https://istio.io/v1.5/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-TCPSettings
image

  When configure a value to above red part value: `4294967295`, the value can't be parsed due to int32 can't meet it, so need to use `int64` to meet it.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -211,7 +211,7 @@ spec:
                                 maxConnections:
                                   description: Maximum number of HTTP1 /TCP connections
                                     to a destination host.
-                                  format: int32
+                                  format: int64
                                   type: integer
                                 tcpKeepalive:
                                   description: If set then set SO_KEEPALIVE on the
@@ -456,7 +456,7 @@ spec:
                                       maxConnections:
                                         description: Maximum number of HTTP1 /TCP
                                           connections to a destination host.
-                                        format: int32
+                                        format: int64
                                         type: integer
                                       tcpKeepalive:
                                         description: If set then set SO_KEEPALIVE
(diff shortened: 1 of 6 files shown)


ningyougang:
My latest commit's author info as below
image

I don't know why `CLA` check failed
image

Someone knows about it? what should i do?

hzxuzhonghu:
We could not change the type of an API.
The confusing docs actually means, if no value set, the value that istio applied to envoy is max uint32(2^32-1)

ningyougang:
> We could not change the type of an API. The confusing docs actually means, if no value set, the value that istio applied to envoy is max uint32(2^32-1)

Thanks for your reply.

howardjohn (changes requested):
Thanks for the PR!

I am not sure this is backwards compatible in protobuf?  Envoy's is a uint32 so I am not sure using int64 here is appropriate as well

ningyougang:
> Thanks for the PR!
> 
> I am not sure this is backwards compatible in protobuf? Envoy's is a uint32 so I am not sure using int64 here is appropriate as well

uint32 type is in [0 , 4294967295] , also meet the scope.
I already changed it to `uint32`

howardjohn (commented):
Per proto, this does seem ok (https://developers.google.com/protocol-buffers/docs/proto3#updating). However, I am not sure I understand the motivation. Do you really need a number that int32 will not satisfy?

ningyougang:
> Per proto, this does seem ok (https://developers.google.com/protocol-buffers/docs/proto3#updating). However, I am not sure I understand the motivation. Do you really need a number that int32 will not satisfy?

The motivation is the int32 is not enough for below value
image
so it is better to change it from int32 to uint32.

btw, regarding `Per proto, this does seem ok`
You mean need some changes for some proto file? but i didn't find it.

PR #2165, opened 2021-11-22
added tls to sidecar api

shriramsharma:
Fixes istio/istio#35111

RFC: [Support hybrid sidecar mode](https://docs.google.com/document/d/15Qhr7errbylXEzxxCK7ij_oUpn4E5SFU2uDdl_n2GIc/edit#)

Added TLS settings to the sidecar API which enables TLS/mTLS from outside of the mesh

---
**NOTE**

I had mistakenly created a PR on `release-1.12` branch. Here is the old [PR](https://github.com/istio/api/pull/2161) for historical reference.

---


howardjohn (commented):
One thing I wanted to clarify. In the doc, its not clear if we need TLS termination or TLS termination AND routing. This seems to implement only the former. I think this is the right approach but I wanted to make sure that meets the goals.

shriramsharma:
@aattuluri @kfaseela PTAL

kubernetes/customresourcedefinitions.gen.yaml
```diff
@@ -3388,6 +3388,66 @@ spec:
                         targetPort:
                           type: integer
                       type: object
+                    tls:
+                      properties:
+                        caCertificates:
```
kfaseela, line 3393: should be caCertificate?

hzxuzhonghu: This is referring legacy code, so could not change

kfaseela: okay, thanks!

kubernetes/customresourcedefinitions.gen.yaml
```diff
@@ -3388,6 +3388,66 @@ spec:
...
+                          type: array
+                        verifyCertificateHash:
+                          items:
+                            type: string
+                          type: array
+                        verifyCertificateSpki:
+                          items:
+                            type: string
+                          type: array
+                      type: object
```
kfaseela, line 3450: @shriramsharma want to know how multiple hosts config as mentioned in https://istio.io/latest/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-tls-ingress-gateway-for-multiple-hosts will be achieved with the current sidecar extension

shriramsharma: @kfaseela I think this can be achieved by using `subjectAltNames` inside the `tls` config.
```
ingress:
    - port:
        number: 80
        name: http
        protocol: HTTP
      defaultEndpoint: 127.0.0.1:80
      tls:
        mode: SIMPLE
        privateKey: "httpbinkey.pem"
        serverCertificate: "httpbin.pem"
        subjectAltNames:
        - httpbin.example.com
    - port:
        number: 5000
        name: http
        protocol: HTTP
      defaultEndpoint: 127.0.0.1:5000
      tls:
        mode: SIMPLE
        privateKey: "helloworldkey.pem"
        serverCertificate: "helloworld.pem"
        subjectAltNames:
        - helloworld.example.com
```

```
k exec deployment.apps/sleep -c sleep -n app2 -- curl https://httpbin.example.com:8000/hello -vso /dev/null --resolve httpbin.example.com:8000:10.100.133.38 --cacert /etc/sleep/tls/rootCA.pem
```

kfaseela: Thanks @shriramsharma

kfaseela: @shriramsharma if this does not work for same port number as given in the ingress example, do you see any other config option?

kubernetes/customresourcedefinitions.gen.yaml
```diff
@@ -3388,6 +3388,66 @@ spec:
...
+                        caCertificates:
+                          description: REQUIRED if mode is `MUTUAL`.
+                          type: string
+                        cipherSuites:
+                          description: 'Optional: If specified, only support the specified
+                            cipher list.'
+                          items:
+                            type: string
+                          type: array
+                        credentialName:
```
kfaseela, line 3402: Does specifying credentialName instead of certificate paths work at the sidecar level?

kubernetes/customresourcedefinitions.gen.yaml
shriramsharma: @kfaseela , as discussed offline, right now the sidecar API does not allow to use same port numbers in ingress.

kubernetes/customresourcedefinitions.gen.yaml
shriramsharma: @kfaseela , as of now, it does not

networking/v1alpha3/sidecar.proto
```diff
@@ -491,6 +491,8 @@ message IstioIngressListener {
 
   reserved "localhost_client_tls";
   reserved 5, 6;
+
```
costinm, line 494: Comments please. Including limitations, etc.

costinm: You can also link to the (istio) Gateway tls setting, and describe differences. Wouldn't hurt to also link to the new K8S Gateway API tls settings.

howardjohn: Would also like to see:
* Full Sidecar YAML example. Its not super trivial to configure
* How it relates to `protocol`, in the sidecar.ingress config and Service. I think the Service should be `http` and the Sidecar config should be `https`?
* What traffic is allowed: plaintext, tls, mTLS. How it relates to PeerAuthentication config

shriramsharma: @costinm @howardjohn , I've added some comments and example YAML. Can you please review them? Thanks

networking/v1alpha3/gateway.proto
```diff
@@ -615,6 +615,7 @@ message Port {
 message ServerTLSSettings {
   // If set to true, the load balancer will send a 301 redirect for
   // all http connections, asking the clients to use HTTPS.
+  // Cannot be used in the Sidecar IstioIngressListener API.
```
hzxuzhonghu, line 618: Why canot be used in sidecar?

shriramsharma: @hzxuzhonghu , and this limitation based on this comment https://github.com/istio/istio/pull/36210#discussion_r778271615

louiscryan, line 618: Note that 'IstioIngressListener' isn't a documentable part of the user-visible API. Better to refer to what it's contained by - i.e. "Cannot be used for ports in [Sidecar.ingress](link)"

hzxuzhonghu: TBH, i think this can be useful for sidecar too

networking/v1alpha3/sidecar.proto
```diff
@@ -401,6 +401,81 @@ import "networking/v1alpha3/virtual_service.proto";
...
+// kind: Service
+// metadata:
+//   name: ratings
+//   labels:
+//     app: ratings
+//     service: ratings
+// spec:
+//   ports:
+//   - port: 8443
+//     name: http
```
howardjohn, line 453: should be `https`. Otherwise when clients try to talk to this service we will see issues; they *should* be using HTTPS, but clientside envoy will expect http because of this config and not match.

networking/v1alpha3/gateway.proto
```diff
@@ -650,6 +652,7 @@ message ServerTLSSettings {
     // gateway workload identity, generated automatically by Istio
     // for mTLS authentication. When this mode is used, all other
     // fields in `TLSOptions` should be empty.
+    // Cannot be used in Sidecar.ingress API..
```
howardjohn, line 655:
```suggestion
    // Cannot be used in Sidecar.ingress API.
```

howardjohn (approved):

aattuluri (approved):
lgtm

networking/v1alpha3/gateway.proto
shriramsharma: @hzxuzhonghu , thanks for your feedback. Unfortunately, we're getting too close to the next release and this might require some more discussions. So I can open an issue to track this and maybe we can include it in a future release.

shriramsharma:
@costinm @hzxuzhonghu , can I please get another approval here?
This [PR](https://github.com/istio/istio/pull/36210) is dependant on this one to be merged. Thanks.

networking/v1alpha3/sidecar.proto
```diff
@@ -401,6 +401,81 @@ import "networking/v1alpha3/virtual_service.proto";
...
+// `prod-us1` namespace for all pods with labels `app: ratings`
+// belonging to the `ratings.prod-us1` service.  The service accepts
+// inbound HTTPS traffic on port 8443 and the sidecar proxy terminates
+// one way TLS using the given server certificates.
+// The traffic is then forwarded to the attached workload instance
+// listening on a Unix domain socket.
+// It is expected that PeerAuthentication policy would be configured
+// in order to either set mTLS mode to "DISABLE" on specific
+// ports.
+// In this example, the mTLS mode is disabled on PORT 80.
```
hzxuzhonghu, line 414: I am thinking about this example is not common, actually it is very special.

networking/v1alpha3/gateway.proto
hzxuzhonghu: Maybe `Sidecar.Ingress`

shriramsharma: @hzxuzhonghu for now I have removed these comments.
as per @howardjohn's recommendation, I've added `hide-from-docs` for this feature for now.

networking/v1alpha3/sidecar.proto
shriramsharma: @hzxuzhonghu for now, I have put a note that this feature is experimental.

hzxuzhonghu (approved):

ramaraochavali:
@shriramsharma Since this is implemented and we have feature flag to control this behaviour, can we unhide this from docs?

costinm:
Yes, we should unhide it.

shriramsharma:
yes, @ramaraochavali @costinm, I'll create a PR to unhide this. Thanks

PR #2166, opened 2021-11-22
Fix docs generation for proxy_config

howardjohn:
Currently ProxyConfig is not generating any docs. The reason is fairly
complex. The docs generator supports a `mode`; this must be the same for
a whole *package*. So we make all v1beta1 types generate docs.

On the istio.io side, we will just copy these into the same place. This
means for ones where we have alpha and beta, we will replace beta with
alpha if both are present (a PR to the docs repo will make beta > alpha,
but the result is identical right now).


howardjohn:
https://github.com/istio/api/pull/2167 is better

PR #2170, opened 2021-11-23
add LogOption for Telemetry API

zirain:
this PR try to add LogOption on Telemetry API allow user change access log config per POD.

@douglas-reid would you please take a look at this, as ProviderRef is slice, there‘s a little troubled when  returns the logging configuration for a given proxy [AccessLogging](https://github.com/istio/istio/blob/e358a5974e2bc8fadf8b89d8c70e04a6bea9ec5e/pilot/pkg/model/telemetry.go#L181)


howardjohn:
Perhaps I am missing context but this seems to be contradictory with https://github.com/istio/api/pull/2110 which was just merge?

hzxuzhonghu:
I am thinking this is to allow overriding global options from provider, so it could be more friendly to use

zirain:
> Perhaps I am missing context but this seems to be contradictory with #2110 which was just merge?

yes, try to add options to override global options from provider per pod

looking forward to hearing your opinions

howardjohn:
It is still possible to do per pod overrides. You just need to create a new
provider first.

douglas-reid:
I would like to hold off on this approach. We have been trying to avoid putting provider-specific details in Telemetry API. And as @howardjohn relates, we have the ability to add per-pod customization via different providers in the current API.

I do think that we will need to revisit the override situation in one of a number of ways, including: (1) a potential separate Providers API and (2) possible "universal" logging api.

One loose thought I've had around (2) is maybe having a set of key-value pairs that, when `structured` logging is selected help build a JSON entry and when `unstructured` logging is selected result in an ordered, delimited serialization of the values (and/or a single field be the entire entry (maybe specially-named?)).

douglas-reid:
@zirain @hzxuzhonghu @howardjohn I took a quick shot at something that _may_ be tolerable in https://github.com/istio/api/pull/2171.

It would require putting additional smarts in the control plane, as well as some potential indirection for users (`additional_request_headers_to_log` -> `attributes.foo: "%REQ(my_header)%` for the ALS stuff), so it is not perfect. I'd be curious to know if you think such an approach is worth pursuing now that we have a baseline to fall back on.

zirain:
> @zirain @hzxuzhonghu @howardjohn I took a quick shot at something that _may_ be tolerable in #2171.
> 
> It would require putting additional smarts in the control plane, as well as some potential indirection for users (`additional_request_headers_to_log` -> `attributes.foo: "%REQ(my_header)%` for the ALS stuff), so it is not perfect. I'd be curious to know if you think such an approach is worth pursuing now that we have a baseline to fall back on.

@douglas-reid I'm using an [EnvoyFilter](https://gist.github.com/zirain/1416750c5aada49dfd349c3f04a84b24) to customize request headers. There still some merge issue as https://github.com/istio/istio/issues/34911#issuecomment-916569749

IMO, the Telemetry API Logging still have something not that good:
- too mush provider in MeshConfig, specially put some pod related config/provider in global MeshConfig
- maybe we can make ALS like RateLimit which require a non-static cluster

telemetry/v1alpha1/telemetry.proto
```diff
@@ -589,4 +589,29 @@ message AccessLogging {
...
+
+    message EnvoyHttpGrpcV3LogOption{
+      // Optional. Additional request headers to log.
+      repeated string additional_request_headers_to_log = 1;
+
+      // Optional. Additional response headers to log.
+      repeated string additional_response_headers_to_log = 2;
+
+      // Optional. Additional response trailers to log.
+      repeated string additional_response_trailers_to_log = 3;
```
hzxuzhonghu, line 612: Instead of treating these options as provider specific, i would rather treat them as application or service specific.  They should be set by the application operator but not mesh operator who has no knowledge of specific applications.

douglas-reid: The issue, as I see it, is that the options themselves are specific to ALS. `additional_request_headers_to_log` is Envoy configuration for the HTTP ALS access logging extension. It doesn't easily translate to other access logging extensions. That's what I mean by provider-specific.

I believe we need something more abstract and "portable" -- something that works across providers. I admittedly do not have a great solution however. I do think there is promise in a separate provider API + some common overrides. That will take some effort to get started, however.

zirain:
duplicate with https://github.com/istio/api/pull/2171

PR #2171, opened 2021-11-29
telemetry(access logging): add format override option

douglas-reid:
DRAFT PR to explore format options within the Telemetry resource itself.

### DIFF
--- a/telemetry/v1alpha1/telemetry.proto
+++ b/telemetry/v1alpha1/telemetry.proto
@@ -589,4 +589,61 @@ message AccessLogging {
   // Optional. If specified, this filter will be used to select specific
   // requests/connections for logging.
   Filter filter = 3;
+
+  message CustomLoggingFormat {
+    // Optional. Textual message format to generate for logging event. This
+    // field is mapped into different provider formats in provider-specific
+    // ways.
+    //
+    // For some structured logging providers (including Envoy Access Logging
+    // Service providers), this field may be dropped entirely, as there is no
+    // place for it in the target schema.
+    //
+    // For unstructured providers, this field will be used exclusively,
+    // regardless if additional attributes are provided.
+    string message = 1;
+
+    // Optional. Used in structured logging to control the attributes captured
+    // in the access logs.
+    //
+    // For providers that define rigid schemas, use of this field is restricted
+    // to adding *additional* attributes to the existing set (where possible).
+    // Examples include adding additional headers/trailers to Envoy Access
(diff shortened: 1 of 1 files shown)


telemetry/v1alpha1/telemetry.proto
```diff
@@ -589,4 +589,61 @@ message AccessLogging {
   // Optional. If specified, this filter will be used to select specific
   // requests/connections for logging.
   Filter filter = 3;
+
+  message CustomLoggingFormat {
```
costinm, line 593: Granularity ? 

For attributes - it is very reasonable to have this per vhost ( in a gateway for example ). What is the finest granularity in envoy ? 

For text format - I still believe it is not right or maintainable to allow workload owners to define the log format - and it MUST be per log collector, since I can't imagine any logging system where different lines have different order for the field. 

Mixing the 2 together forces the wrong granularity for one.

Also for attributes - I believe selecting the subset of attributes is not specific to logging - tracing, metrics have similar needs, so it may be better to treat all telemetry consistently.

douglas-reid: The granularity here is still per-workload at the finest level.

The distinction between attributes and text format is not clear to me here. They are two sides of the same coin, in many ways (just a serialization difference). Put another way, with both, I'm selecting the content of the log entry. JSON provides more flexibility here, perhaps, but ultimately it is the same selection.

Example: `text_format: %REQ(x-custom-header)%` ~= `json_format:  foo: %REQ(x-custom-header)%`

telemetry/v1alpha1/telemetry.proto
```diff
@@ -589,4 +589,61 @@ message AccessLogging {
...
+    // translate Envoy command operators for request headers and trailers into
+    // known provider formats.
+    //
+    // Example:
+    // ```
+    // attributes:
+    //   status: "%RESPONSE_CODE%"
+    //   message: "%LOCAL_REPLY_BODY%"
+    // ```
+    google.protobuf.Struct attributes = 2;
```
hzxuzhonghu, line 622: Feels `struct` is too hard to use for end users.

mandarjog: Struct just translates to json. Why is it hard?

hzxuzhonghu: They donot know the valid key and values

zirain: image

IMO, this looks too complex

PR #2173, opened 2021-11-30
Authorization policy: allow specifying string match for path and conditions

sschepens:
Add StringMatcher options to Path and Condition in AuthorizationPolicy.
We use the same StringMatcher as used in VirtualHosts to provide consistent matching in both.
If both String and StringMatcher are present, they should be appended.

Fixes https://github.com/istio/istio/issues/16585
Related: https://github.com/istio/istio/pull/35641

@howardjohn wdyt? this way we can prevent introducing breaking changes in the exiting APIs but still support matching by regex as needed.


sschepens:
@howardjohn what is your opinion on this? I would like to make progress on this feature, it is something that many people need.

teddy-wahle:
I would love for this PR to be merged! We really need this feature.

timAdrien:
Yes please someone validate the PR, a lot of people are waiting this feature for so long

craigbox:
Closing due to implementation in https://github.com/istio/istio/pull/50365

PR #2174, opened 2021-12-01
proxyconfig(service-cluster): add control for service cluster naming

douglas-reid:
This PR provides enumerated options for `--service_cluster` generation. This can be used to control the service names used in trace spans.

This is meant to address a regression in exposed functionality that appeared in 1.12 (istio/istio#36162), which itself was a workaround for the original FR (istio/istio#12644).

This is not proposed as part of the new Telemetry API, as it is a bootstrap parameter (and not currently modifiable at runtime). The existing scope of the Telemetry API is dynamic runtime behaviors (read: no proxy restarts required).

If approved, there will be a strong desire to cherry-pick this change and a subsequent implementation PR in istio/istio into 1.12.


mesh/v1alpha1/proxy.proto
```diff
@@ -324,20 +324,46 @@ message ProxyConfig {
...
+    CANONICAL_SERVICE_NAME_AND_NAMESPACE = 2;
+    // Similar to `APP_LABEL_AND_NAMESPACE` but without the namespace appended.
+    APP_LABEL_ONLY = 3;
+    // Similar to `WORKLOAD_NAME_AND_NAMESPACE` but without the namespace appended.
+    WORKLOAD_NAME_ONLY = 4;
+    // Similar to `CANONICAL_SERVICE_NAME_AND_NAMESPACE` but without the namespace appended.
+    CANONICAL_SERVICE_NAME_ONLY = 5;
+  }
+  
+  oneof cluster_name {
```
howardjohn, line 347: just making sure: this does not show up in yaml right?

douglas-reid: good question. i would not think the `oneof` should up show in YAML, but I'm always a bit surprised by YAML.

any thoughts on a better name here, btw?

mesh/v1alpha1/proxy.proto
```diff
@@ -324,20 +324,46 @@ message ProxyConfig {
...
+    // a cluster name.
+    WORKLOAD_NAME_AND_NAMESPACE = 1;
+    // Uses the canonical service name and namespace.
+    CANONICAL_SERVICE_NAME_AND_NAMESPACE = 2;
+    // Similar to `APP_LABEL_AND_NAMESPACE` but without the namespace appended.
+    APP_LABEL_ONLY = 3;
+    // Similar to `WORKLOAD_NAME_AND_NAMESPACE` but without the namespace appended.
+    WORKLOAD_NAME_ONLY = 4;
+    // Similar to `CANONICAL_SERVICE_NAME_AND_NAMESPACE` but without the namespace appended.
+    CANONICAL_SERVICE_NAME_ONLY = 5;
```
howardjohn, line 344: Given canonical service is already taking app and workload into account as defaults, should we only have canonical service name as the option?

douglas-reid: I'm hesitant to start with only canonical service name since, iirc, the current code uses `app` + ns. Though those should align for some super high percentage, I'm not sure it is "safe" to swap them.

@Stono thoughts here?

Stono: I find it all a bit confusing... sorry.  Correct me if my understanding below is wrong.

The injection template currently has:

```
service.istio.io/canonical-name: {{ index .ObjectMeta.Labels `service.istio.io/canonical-name` | default (index .ObjectMeta.Labels `app.kubernetes.io/name`) | default (index .ObjectMeta.Labels `app`) | default .DeploymentMeta.Name  | quote }}
```

Which I read as:
- take `service.istio.io/canonical-name`
- failing that take `app.kubernetes.io/name`
- failing that take `app`
- failing that take `Deployment.spec.name`

In my case, `canonical-name` then would always be `labels.app`, yes?

Then there's `canonical-service name`, which I assume is the service that this app belongs to.  The service name for me is always called "app" (we deploy everything in the same way).  So I think `canonical-service` would be `app.` + `labels.app`, eg `app.some-app`.

Because of design decisions for us, the service name is irrelevant as we have a single service and workload per namespace, hence why historically we've always set `serviceCluster` === labels.app.

Therefore, to regain the ability to do that I'd probably want `WORKLOAD_NAME_ONLY` or `CANONICAL_WORKLOAD_NAME` perhaps to marry up with the labels being added by the injector?

douglas-reid: OK. Based on this feedback, I've attempted to simplify the options. Please let me know what you think.

Stono:
lgtm

douglas-reid:
@howardjohn any further thoughts here?

howardjohn (approved):

nrjpoddar:
@douglas-reid IIRC the default used to be just the app name without namespace?

douglas-reid:
Gentle ping on second approval.

nrjpoddar (approved):

PR #2175, opened 2021-12-02
add otel ALS provider

zirain:
This PR adds a way to configure [Envoy Open Telemetry ALS](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/open_telemetry/v3/logs_service.proto) providers in MeshConfig.

@douglas-reid PTAK


mesh/v1alpha1/config.proto
```diff
@@ -910,6 +912,32 @@ message MeshConfig {
...
+      // Defaults:
+      // - "otel_envoy_accesslog"
+      string log_name = 3;
+
+      // Optional. Format for the proxy access log
+      // Empty value results in proxy's default access log format
+      string body = 4;
+
+      // Optional. Additional attributes that describe the specific event occurrence.
+      map<string,string> attributes = 5;
```
hzxuzhonghu, line 939: please add more comments about what does it mean

mesh/v1alpha1/config.proto
```diff
@@ -910,6 +912,32 @@ message MeshConfig {
...
+      uint32 port = 2;
+
+      // Optional. The friendly name of the access log.
+      // Defaults:
+      // - "otel_envoy_accesslog"
+      string log_name = 3;
+
+      // Optional. Format for the proxy access log
+      // Empty value results in proxy's default access log format
+      string body = 4;
```
hzxuzhonghu, line 936: something like `format` maybe better

mesh/v1alpha1/config.proto
howardjohn: If this and `body` are what they seem like, we should use the same `LogFormat` from `EnvoyFileAccessLogProvider` (or same names anyhow) IMO

douglas-reid: I think this is an important decision point. Should the provider configuration use the provider terminology (here: `body` and `attributes`) or should we have commonality across provider configuration?

zirain: Agree, `LogFormat` is better than `body`.

should we keep use `attributes` as [Envoy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/open_telemetry/v3/logs_service.proto) does.

mesh/v1alpha1/config.proto
zirain: fixed, PTAK!

mesh/v1alpha1/config.proto
zirain: so far, `LogFormat` support text and json.
IMO, `body`(rename to 'format') prefer to set some like [EnvoyTextLogFormat](https://github.com/istio/istio/blob/ae120753863586bdc807722d13ce7511152ce36e/pilot/pkg/networking/core/v1alpha3/accesslog.go#L43), and `attributes` used for json output
```yaml
envoyOtelAls:
  service: otel-collector.monitoring.svc.cluster.local
  port: 55680
  format: "[%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %RESPONSE_CODE_DETAILS% %CONNECTION_TERMINATION_DETAILS% \"%UPSTREAM_TRANSPORT_FAILURE_REASON%\" %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\" %UPSTREAM_CLUSTER% %UPSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_REMOTE_ADDRESS% %REQUESTED_SERVER_NAME% %ROUTE_NAME%\n"
  attributes:
  - "user_agent": "%REQ(USER-AGENT)%"
  - "protocol": %PROTOCOL%"
```

zirain:
@howardjohn @douglas-reid Updated, PTAK.

mesh/v1alpha1/config.proto
```diff
@@ -910,6 +912,37 @@ message MeshConfig {
...
+      uint32 port = 2;
+
+      // Optional. The friendly name of the access log.
+      // Defaults:
+      // - "otel_envoy_accesslog"
+      string log_name = 3;
+
+      // Optional. Format for the proxy access log
+      // Empty value results in proxy's default access log format
+      string format = 4;
```
douglas-reid, line 936: is `format` here `body` in the Envoy config?  If so, can we add that into the comment for the field?

mesh/v1alpha1/config.proto
```diff
@@ -511,6 +511,8 @@ message MeshConfig {
       EnvoyHttpGrpcV3LogProvider envoy_http_als = 12;
       // Configures an Envoy Access Logging Service provider for TCP traffic.
       EnvoyTcpGrpcV3LogProvider envoy_tcp_als = 13;
+      // Configures an Envoy Open Telemetry Access Logging Service provider.
+      EnvoyOpenTelemetryLogProvider envoy_otel_als = 14;
```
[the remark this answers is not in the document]
douglas-reid: The Envoy* providers have followed the `envoy.access_loggers` extension category naming scheme, for the most part. I think this goes back to the field naming discussion above: should we follow the underlying extension configuration closely, or try to abstract over it to make the Istio configuration more consistent across the extensions?  I'm not sure which approach will ultimately be clearer for the admin that is configuring the extensions (do you have to know Envoy configuration and Istio configuration? only one? which one?).

```
envoy.access_loggers.file
envoy.access_loggers.http_grpc
envoy.access_loggers.tcp_grpc
envoy.access_loggers.open_telemetry
...
```

mesh/v1alpha1/config.proto
zirain: see https://github.com/istio/api/pull/2175#discussion_r761048363, I'll update the comment.

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
zirain: > Above we have `http` and `tcp`. Having `otel` is the right API, I think, but it may be confusing based on the comments. We may want to clarify here the distinction between these?
> 
> I assume otel handles http+tcp together?

yes

mesh/v1alpha1/config.proto
howardjohn: IMO all provider should use standard `LogFormat`. If we need to support a body AND attributes, we can just make `LogFormat` not have a oneof (and allow setting both fields).

douglas-reid: @howardjohn and what do you do when you can only have one (file/stream provider)?  define `body` as the priority field?

My thinking is that provider config should be provider-specific. If there is truly common config, it should be elevated to the API. `LogFormat` won't cover the ALS providers (directly), so that would leave a decision around having bits in the API that aren't useful for all providers (or adding control plane logic to map into specific providers).

howardjohn: I think as much as possible the yaml API should be the same. So maybe this isn't `LogFormat` *directly*, but the fields are named `text` and `labels`. For things that are truly different like `filter_state_objects_to_log` - not much we can do.

zirain: @howardjohn do you means something like this:
```yaml
envoyOtelAls: 
  service: otel-collector.monitoring.svc.cluster.local 
  port: 55680 
  log_format: 
    text: "[%START_TIME%] \"%REQ(:METHOD)% %REQ(X-ENVOY-ORIGINAL-PATH?:PATH)% %PROTOCOL%\" %RESPONSE_CODE% %RESPONSE_FLAGS% %RESPONSE_CODE_DETAILS% %CONNECTION_TERMINATION_DETAILS% \"%UPSTREAM_TRANSPORT_FAILURE_REASON%\" %BYTES_RECEIVED% %BYTES_SENT% %DURATION% %RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)% \"%REQ(X-FORWARDED-FOR)%\" \"%REQ(USER-AGENT)%\" \"%REQ(X-REQUEST-ID)%\" \"%REQ(:AUTHORITY)%\" \"%UPSTREAM_HOST%\" %UPSTREAM_CLUSTER% %UPSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_LOCAL_ADDRESS% %DOWNSTREAM_REMOTE_ADDRESS% %REQUESTED_SERVER_NAME% %ROUTE_NAME%\n" 
    labels: 
      user_agent: "%REQ(USER-AGENT)%"
      protocol: %PROTOCOL%"
```

howardjohn: Yes, that is my preference. Then it looks the same for users using envoy or otel or others.

(small note: log_format -> logFormat is preferred format, especially in docs. `log_format` in proto will accept both forms in yaml)

douglas-reid: This seems OK, especially if that unblocks this work. I think it starts to present a template for promotion into the API as well.

hzxuzhonghu (approved):

hzxuzhonghu:
@douglas-reid @howardjohn Can you approve if no other comments

howardjohn (approved):

PR #2182, opened 2021-12-06
remove derelict/broken bazel build files

pjjw:
these files use rules_go conventions that are four years old and deprecated and just cause problems referencing this repo from a bazel project. the BUILD file here can't even work, as it refers to deleted files. let's just toss it.


pjjw:
apologies, i've signed the CLA but this seems not to be reflected by the tests. tried using a few explicitly approved emails.

howardjohn (approved):

ericvn (approved):

PR #2192, opened 2021-12-21
Loosen weight validation

hzxuzhonghu:
For https://github.com/istio/istio/issues/36069

According to api guide https://github.com/istio/api/blob/master/GUIDELINES.md `Loosening validation is permitted`

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -1258,10 +1258,9 @@ message HTTPRouteDestination {
   // to which the request/connection should be forwarded to.
   Destination destination = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // The proportion of traffic to be forwarded to the service
-  // version. Must be in range [0, infinite).
-  // If there is only one destination in a rule, the weight value is assumed to
-  // be 100.
+  // Weight specifies the relative proportion of traffic to be forwarded to the destination. A destination will receive `weight/(sum of all weights)` requests.
+  // If there is only one destination in a rule, it will receive all traffic.
+  // Otherwise, if weight is `0`, the destination will not receive any traffic.
   int32 weight = 2;
 
   reserved 3 to 6;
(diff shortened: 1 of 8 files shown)


networking/v1alpha3/virtual_service.proto
```diff
@@ -1259,7 +1259,7 @@ message HTTPRouteDestination {
   Destination destination = 1 [(google.api.field_behavior) = REQUIRED];
 
   // The proportion of traffic to be forwarded to the service
-  // version. (0-100). Sum of weights across destinations SHOULD BE == 100.
+  // version. Must be in range [0, infinite).
   // If there is only one destination in a rule, the weight value is assumed to
   // be 100.
```
howardjohn, line 1264:
```
Weight specifies the relative proportion of traffic to be forwarded to the destination. A destination will receive `weight/(sum of all weights)` requests.
If weight is `0`, the destination will not receive any traffic.
If there is only one destination in a rule, it will receive all traffic
```

Then validation should check that `sum(weights) > 0` I think, since I am not sure what we would do in that case otherwise?

howardjohn (commented):
LGTM

networking/v1alpha3/virtual_service.proto
hzxuzhonghu: Right, i also wonder should we change the int432 to uint32, as you said this would not influence protobuf coding, maybe discuss later

howardjohn (approved):

networking/v1alpha3/virtual_service.proto
```diff
@@ -1258,10 +1258,9 @@ message HTTPRouteDestination {
   // to which the request/connection should be forwarded to.
   Destination destination = 1 [(google.api.field_behavior) = REQUIRED];
 
-  // The proportion of traffic to be forwarded to the service
-  // version. (0-100). Sum of weights across destinations SHOULD BE == 100.
-  // If there is only one destination in a rule, the weight value is assumed to
-  // be 100.
+  // Weight specifies the relative proportion of traffic to be forwarded to the destination. A destination will receive `weight/(sum of all weights)` requests.
+  // If there is only one destination in a rule, it will receive all traffic.
```
ericvn, line 1262: What happens in the case of a single rule with a value of 0? The validation of sum of the weights being zero should catch this case, just not sure if the wording could be better.

hzxuzhonghu: it behaves as before

ericvn (approved):

PR #2196, opened 2021-12-24
add support for starttls mode

ramaraochavali:
This is mostly needed for server first protocols that expect startTLS mode from client. Fix for https://github.com/istio/istio/issues/29761

When this is used, the plan is to generate start TLS config and set upstreamTLSContext https://github.com/envoyproxy/envoy/blob/9d5627a0879b0a029e90515137c108e1d2884bfc/api/envoy/extensions/transport_sockets/starttls/v3/starttls.proto#L45 - so that usecases like postgres can be handled easily.


ramaraochavali:
@howardjohn @hzxuzhonghu any objections to this?

howardjohn (commented):
is there "mutual start tls"? Or only simple?

ramaraochavali:
AFAIK, only SIMPLE

howardjohn:
Seems like it might be possible: https://stackoverflow.com/a/26919123/374797

ramaraochavali:
I see. Thanks for the link. Is that a concern for us? We can setup UpstreamTLSContext accordingly?

howardjohn:
Well it seems odd to have normal TLS (simple+mutual) but STARTTLS only
simple. I would think mutual vs simple is orthogonal to TLS flavor

ramaraochavali:
Sorry - I did not understand your comment fully. What are you suggesting?

howardjohn:
We need the API to support simple and mutual starttls

costinm:
I think 'startTLS' may be a bit too much for Istio APIs. I understand there are protocols that require this - but most of them have actual implementations that deal with startTLS already, not clear how this would even be supported with the sidecar or gateway.

And if envoy supports it - a filter may be appropriate, it is not that common.

costinm (commented):
To confirm: what do you mean by 'start_tls' ? Normally that's the in-band negotiation of TLS, where client/server exchange few plain text messages and then 'start tls' is sent to do a handshake. https://en.wikipedia.org/wiki/Opportunistic_TLS

If you just mean starting in tls (or mtls) mode for server-first protocols - please call it something else, and add some extra docs or example on what it actually does.

ramaraochavali:
> If you just mean starting in tls (or mtls) mode for server-first protocols - please call it something else, and add some extra docs or example on what it actually does.

Yes.  I can add. Can you please suggest a better name for it?

costinm:
We already have the setting to use TLS or mTLS - maybe just 'serverFirst'
? Forgot, do we have an explicit API to
specify this for plain TCP ? If not - we could use 'serverFirst' for both
TCP and TLS.

Also: we use protocol name in a lot of places, this behavior is really part
of the protocol specification.
IMO having a (configurable) list of protocols that are 'server first' would
be more helpful and clear,
and we can have defaults that cover the known ones, most new protocols are
client first (and I don't think
a lot of new non-TLS non-H2 protocols are getting created )

howardjohn:
Costin - I think this is referring to in band negotiation of TLS (per the
wiki link). It's not just server first protocol

costinm:
On Wed, Jan 26, 2022 at 6:39 AM John Howard ***@***.***>
wrote:

> Costin - I think this is referring to in band negotiation of TLS (per the
> wiki link). It's not just server first protocol
>

Right - but was using the exact same name as the in-band negotiation key.

costinm:
As someone who had to implement IMAP/POP3 and XMPP - I suspect at some
point someone will ask for (real) startTLS,
which will be a huge pain. For this PR - my only concern is to not create
additional confusion, it's already a pain to deal
with the in-band startTLS.

Also I think the right way to specify 'server first' may be by configuring
based on protocol, not at mtls level.

howardjohn:
I think I am getting confused - AFAIK this PR is for enabling
https://www.envoyproxy.io/docs/envoy/v1.17.4/api-v3/extensions/transport_sockets/starttls/v3/starttls.proto#envoy-v3-api-msg-extensions-transport-sockets-starttls-v3-starttlsconfig
.

I don't know what real startTls is if not this config?

ramaraochavali:
> I think I am getting confused - AFAIK this PR is for enabling
https://www.envoyproxy.io/docs/envoy/v1.17.4/api-v3/extensions/transport_sockets/starttls/v3/starttls.proto#envoy-v3-api-msg-extensions-transport-sockets-starttls-v3-starttlsconfig

Yes.  That is correct.

> I don't know what real startTls is if not this config?

Same question. @costinm can you please explain a bit more?

@howardjohn I think if we agree that STARTTLS is an additional TLS mode, Based on your comments earlier - I agree we should add support for SIMPLE and MUTUAL by qualifying it. But let us first settle on whether adding here is the right way - which I think is the right way.

costinm:
The link I included earlier should explain what the 'real STARTTLS' is - has been in use for a very long time in many widely used protocols ( IMAP, XMPP, etc ) with pretty clear semantics, to start the TLS handshake after a number of plain text messages have been exchanged.  We may have to support that in the future if we start supporting some of those protocols ( Apache, Nginx and other servers do support 'real' STARTTLS AFAIK ).

It is unfortunate that envoy proto used a keyword that is in broad use in RFCs for something very different, but we don't have to do the same. 

Second, I still don't understand what this mode does - is there a RFC or protocol specification describing this behavior ? Is the client starting a TCP connection, and server sending ClientHello ? At minimum the docs should have a link or some indication of what it does - and as John mentioned how it apples to mTLS.

Also if this is a protocol-specific thing, I don't think it should be another TLS mode - the enum indicates if the server should ask for
client certificates or not, has nothing to do with a custom protocol. I would instead just require users to set the protocol to 'postgress-tls' (if this is something that postgress protocol does when TLS is enabled), and treat this as a specific handling of the protocol.  If we ever implement XMPP or IMAP support - I would guess the (real) STARTTLS will also be implemented in the same way, not by adding a mode to the tls mode enum, but as part of the protocol handling.  

But first - please add some more info to the comments about what the protocol does, because 'whatever envoy implemented for a keyword that has been defined by RFCs to do something different' is not clear enough...

howardjohn:
As far as I know this *is* adding 'real STARTTLS' as used by IMAP, XMPP,
etc - what makes you think it is not?

costinm:
How ? AFAIK the implementation of (real) StartTLS requires some understanding of the protocol before TLS is switched on, for example for XMPP there is some XML exchanged before, while for IMAP some text, plus some protocol-defined negotiation.

costinm:
For example: https://xmpp.org/extensions/xep-0035.html

howardjohn:
Hm.. looking at the envoy tests it may also require using the envoy native
postgres filter as well (which we also don't support)

costinm:
Yes - it's perfectly fine to add 'posgress' protocol support, like we have redis and other things - and for posgress to use internally StartTLS ( if negotiation allows it ).

However - I'm even more confused by 'server first protocols' in the comments, https://www.postgresql.org/docs/9.3/protocol-flow.html doesn't appear to be a server-first.

costinm:
Also - to define a Istio API as a new TLS mode but with semantic "setting TLS mode really mean postgress protocol" is pretty strange, even by Istio API standards...

ramaraochavali:
> Hm.. looking at the envoy tests it may also require using the envoy native
postgres filter as well (which we also don't support)

We do not need to support postgres filter. We have tested this internally with an Envoy filter to a postgres database instance without postgres filter (as plain TCP) and it works.

howardjohn:
How does that work? Is it like sniffing the bytes for the starttls signal?

costinm:
If it's another sniffing hack - and specific to postgresql - please close this PR, and add whatever is needed if protocol is postgresql.

When other protocols are supported - we can do the same, there are not too many and unlikely a lot of new protocols will use this,  and it requires special code specific to the protocol in envoy anyways.

ramaraochavali:
> How does that work? Is it like sniffing the bytes for the starttls signal?

I have no idea about how envoy implemented this. May be @lizan knows it better.

> If it's another sniffing hack - and specific to postgresql - please close this PR, and add whatever is needed if protocol is postgresql.

It is not specific to postgressql though.  This should be useful for any serverfirst protocols.  But I do npt know the internal details of Envoy implementation. Let us wait till @lizan responds and based on that I can close this PR.

costinm:
Rama - I checked the spec for postgress, and it does not appear to be a
'server first' protocol. If you find any docs that suggest otherwise - let
me know. And StartTLS in general has nothing to do with 'server first' or
'client first'.

It is used for starting TLS after both client and server have exchanged a
number of plain text messages.
And the way it is implemented ( in at least 3-4 cases I know ) is by having
the proxy understand the
actual protocol - because for each protocol it is different. In XMPP it's
some XML element - based on
previous attributes in the plain text, in SMTP and IMAP it is in the
respective formats. I highly doubt
envoy (or anyone) could implement starttls in a generic way and without
partial specific protocol
implementation.

StartTLS is generally considered 'opportunistic TLS' - i.e. based on the
plain text handshake, if both
ends are capable of TLS, then startTLS is used. Usually there is nothing to
configure - user may
want to indicate that if TLS can't be negotiated the connection should be
closed, but we already have the
TLS mode for that.

This also can't be used for 'mesh' - multi-network is based on SNI routing
and that can't work with
startTLS protocols.

I'm not saying it's a bad feature for gateways/proxies - SMTP/IMAP  are
still in pretty broad use ( XMPP maybe
less), but this would be part of implementing the protocol.

ramaraochavali:
@costinm You may be right - It is not serverFirst looks like (I could not find where I found that).

Also now I remember - I tried with this  looking at postgres option - open ssl s_client must be using postgres protocol - So  it might be specific to postgres.

`openssl s_client -msg -debug -starttls postgres  -connect`

Assuming that it is posgres specific - I agree with you. This PR does not make sense, but we can startTLS config based on protocol in Istiod it self. 

Does it make sense to you?

lizan:
I believe this meant to configure STARTTLS transport socket for Envoy. It's not completely specific for Postgres, but it need the network filter to support it. So configuring this based on protocol makes more sense.

hzxuzhonghu:
What are the rest protocols that need this? Does istio have network filters configured for them?

ramaraochavali:
@lizan Thank you for the response.

I think having network filter is not mandatory. We have tried this with the following setup

Progres JDBC Driver in client APP (initiates plain text) -> Client Envoy (with start tls configured to upgrade the connection to tls with tcp protocol) -> Postgres instance (AWS) and it works. So strictly having network filter is not needed.

But I agree and make it more sense to tie this with protocol so that in future istio can add network filters if needed for better support.

@costinm @howardjohn @hzxuzhonghu are you fine with it? If yes - I will close this PR and implement it in Istiod based on protocol (for postgres to start with)

costinm:
Sounds good, and we can do the same with other protocols as envoy adds support.

ramaraochavali:
closing this infavor of https://github.com/istio/istio/pull/37290

cpakulski:
Some late comments - just noticed this discussion. I agree that starttls transport socket should be added as part of the protocol, not separately. So, something like "postgres-tls" should automatically add underlying starttls transport socket and configure postgres filter to drive transport socket to achieve TLS termination. Note that two scenarios are possible here: termination of TLS (client sends encrypted traffic to Envoy) or initiation of TLS (Envoy sends encrypted traffic to Postgres server, but client may talk clear-text to Envoy).
As for STARTTLS name, yeah - maybe the naming is bit unfortunate, but in Envoy implementation it is really used for scenarios where several cleartext messages are exchanged to decide whether to start TLS handshake on the same TCP session.
starttls transport socket cannot exist by itself and requires a filter which understands cleartext initial exchange and signals to starttls transport socket to start TLS handshake.
I also looked at using starttls transport socket to terminate MySQL's encryption at Envoy, but it cannot be done at the moment without additional development on transport socket side.

PR #2197, opened 2021-12-25
Make termination drain duration default to 30s

hzxuzhonghu:
The current default value is quite  different from k8s pod default value. (5s vs 30s) So when pod is terminated, envoy will be killed after 5s, but application container can be still there, maybe requiring communication with the outside.  It was  not a graceful way before.

### DIFF
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
@@ -1406,7 +1406,7 @@
             "$ref": "#/components/schemas/istio.mesh.v1alpha1.Topology"
           },
           "terminationDrainDuration": {
-            "description": "The amount of time allowed for connections to complete on proxy shutdown. On receiving `SIGTERM` or `SIGINT`, `istio-agent` tells the active Envoy to start draining, preventing any new connections and allowing existing connections to complete. It then sleeps for the `termination_drain_duration` and then kills any remaining active Envoy processes. If not set, a default of `5s` will be applied.",
+            "description": "The amount of time allowed for connections to complete on proxy shutdown. On receiving `SIGTERM` or `SIGINT`, `istio-agent` tells the active Envoy to start draining, preventing any new connections and allowing existing connections to complete. It then sleeps for the `termination_drain_duration` and then kills any remaining active Envoy processes. If not set, a default of `30s` will be applied.",
             "type": "string"
           },
           "meshId": {
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
@@ -2739,7 +2739,7 @@ No
 On receiving <code>SIGTERM</code> or <code>SIGINT</code>, <code>istio-agent</code> tells the active Envoy to start draining,
 preventing any new connections and allowing existing connections to complete. It then
 sleeps for the <code>termination_drain_duration</code> and then kills any remaining active Envoy processes.
-If not set, a default of <code>5s</code> will be applied.</p>
+If not set, a default of <code>30s</code> will be applied.</p>
 
 </td>
 <td>
--- a/mesh/v1alpha1/proxy.pb.go
+++ b/mesh/v1alpha1/proxy.pb.go
@@ -1232,7 +1232,7 @@ type ProxyConfig struct {
 	// On receiving `SIGTERM` or `SIGINT`, `istio-agent` tells the active Envoy to start draining,
 	// preventing any new connections and allowing existing connections to complete. It then
 	// sleeps for the `termination_drain_duration` and then kills any remaining active Envoy processes.
-	// If not set, a default of `5s` will be applied.
+	// If not set, a default of `30s` will be applied.
 	TerminationDrainDuration *types.Duration `protobuf:"bytes,29,opt,name=termination_drain_duration,json=terminationDrainDuration,proto3" json:"terminationDrainDuration,omitempty"`
 	// The unique identifier for the [service mesh](https://istio.io/docs/reference/glossary/#service-mesh)
 	// All control planes running in the same service mesh should specify the same mesh ID.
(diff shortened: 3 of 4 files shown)


howardjohn (changes requested):
This will mean that all pods take 30s to shutdown. I do not think this is a good default, and even if it was the behavioral change is too great to make in a backwards compatible manner, IMO.

Users can easily set this themselves if they desire.

howardjohn:
Current logic is: `sleep(drain duration)`. There are a number of efforts to make it smarter, which I think are a better investment. When we do that, it should be based on the actual terminationGracePeriodSeconds, IMO, instead of a fixed 30s

hzxuzhonghu:
The value is to comply with pod default terminationGracePeriodSeconds, otherwise sidecar will exit before application container. Which may cause termination forcibly.

howardjohn:
Yes, the downside is that ALL pods will take the full 30s now with this
change. At the time the 5s default was introduced this was deemed
unacceptable, and I still feel it is today - especially since we have 2-3
prototypes of proper fixed for this in flight. Changing the default now
will cause a behavioral regression and lots of churn again when we undo it
in the near future.

Users are free to set this themselves

costinm:
Envoy stopping before the app is stopped is clearly a bug and needs to be
fixed, we all agree on this.
Keeping envoy and the pod running AFTER the app stopped is also a bug.

The correct behavior is "same as if no sidecar was injected".

I don't think the current TerminationDrainDuration can solve this problem,
regardless of default value.

I only see few possible solutions:
1. (long term) work with k8s for improvements in the pod lifecycle.
2. injection can modify the startup command for the pod, running a 'init'
that handles startup/shutdown.
I have a prototype for the init, but injecting it is tricky and seems too
complicated ( works great for VMs
or containers or with the 'injection-less' mode we discussed - but too
tricky with normal injection)
3. detecting if no connections are in use - I think it's an improvement,
but still not the correct solution.
The app may wait for all connections to be closed (which envoy would also
detect and exit ) - but make
additional outbound requests after that.
4. (long term) when the sidecar is out of the pod, the problem goes away.

We may also be able to use the CNI - which technically has high permissions
and can see all containers
on the host. I think it would be worth moving more functionality into the
CNI and starting to require it,
for example a lot of the logic in pilot-agent could be moved there (cert
provisioning, XDS proxy and DNS are better
handled in the CNI ).  There are some problems - it may require vendor or
CRI-specific plugins,
but that's probably useful for (4) as well.

IMO the best short (and long) term solution is a variant of (2), providing
an 'istio-init' app and
library, that users can explicitly add to their docker images and use as
entry point, and also handle the
init zombie problem.  For proxyless gRPC this would also handle bootstrap
generation - since
proxyless doesn't need a sidecar or init container. It does have the
drawback that users will
have to make a small change to their code and the 'magic' will not work -
but I think serious
users who actually depend on such subtle startup/shutdown logic are better
served if they gain
visibility and control over startup instead of the current magic black box.

By 'library' I mean sample code and documentation for the interaction
between containers, i.e.
a stable and supported mechanism for the app to be notified and send events
about lifecycle,
which might at some point become (1), and also provide a more efficient way
to propagate
readiness/health status to Istiod ( instead of watching Endpoint and using
the periodic probes).

hzxuzhonghu:
>2. injection can modify the startup command for the pod, running a 'init'
that handles startup/shutdown.

This is also a solution proposed by @howardjohn and @rittneje, but personally, i think it is not a good way

hzxuzhonghu:
https://github.com/istio/istio/issues/36640

costinm:
On Wed, Dec 29, 2021 at 5:24 PM Zhonghu Xu ***@***.***> wrote:

>
>    1. injection can modify the startup command for the pod, running a
>    'init'
>    that handles startup/shutdown.
>
> This is also a solution proposed by @howardjohn
> <https://github.com/howardjohn> and @rittneje
> <https://github.com/rittneje>, but personally, i think it is not a good
> way
>
I tend to agree with you - it is certainly not the ideal solution for k8s,
but I think it's the most realistic
short-term option for users who refuse to make small changes to their
docker images.

The bottom line is that we must have some coordination for startup and
shutdown, and until K8S
provides it we must find some way to do it - either in the CNI (or some
other node-level solution) or in the
pod itself. We need to at least detect when the app container(s) are done
before stopping envoy.

I don't see another choice for detecting container status - has to be
either on the node or in the pod.
And if it is on the pod - either we inject it or the user adds it to their
container explicitly.

Unless someone finds a different way to get the app exit status - it's all
we have, since it's not safe
to stop envoy before.

hzxuzhonghu:
> This will mean that all pods take 30s to shutdown. I do not think this is a good default, and even if it was the behavioral change is too great to make in a backwards compatible manner, IMO.

@howardjohn  If user really wants to shutdown application container with sidecar container at the same time, we have /quitquitquit for them, what all we need to do is https://github.com/istio/istio/pull/36624.   And also how could this related to compatibility.

rittneje:
@hzxuzhonghu Fundamentally, `terminationDrainDuration` is not the right solution for any use case.

1. If your application meets the "normal" use case of just processing incoming requests that are routed through Envoy, and doesn't do anything unusual at shutdown, then the recently add `terminateOnZeroActiveConnections` is sufficient.
2. For all other use cases, there is no proper solution at present. You are forced to just guess how long your application takes to finish, which means you have to go with the worst case always, which is probably the same as your pod termination grace period anyway. In these cases, some explicit synchronization between the containers is really needed, to tell istio-proxy when to exit. Either you need to call /quitquitquit at the end (and fix it to make the sidecar exit immediately), or you need to leverage a UDS. The former can work well within a prestop hook, while the latter is much more robust and flexible, but does require amending application code. Neither of these can work "magically", with no explicit effort on the client's part.

With regards to compatibility, one thing I will say about this change is that since pod shutdown will be artificially delayed, you may end up putting more pressure on the cluster autoscaler during rolling updates, and it will lead to greater latency during deployments, which annoys developers. And in any case, 30 seconds is still a completely arbitrary number. Why not 60 seconds? Or 5 minutes?

hzxuzhonghu:
We are in a diverge way, this has been discussed in several issues. We should get to a concensue  ASAP. 

If user really cares about one second speed, they have to notify sidecar to exit immediately. Otherwise it is no easy for us to do

PR #2199, opened 2022-01-06
fix httpbin.com -> example.com

xichengliudui:
ref: https://github.com/istio/istio.io/pull/10722

### DIFF
--- a/networking/v1alpha3/service_entry.pb.go
+++ b/networking/v1alpha3/service_entry.pb.go
@@ -266,7 +266,7 @@
 //   namespace : egress
 // spec:
 //   hosts:
-//   - httpbin.com
+//   - httpbin.org
 //   exportTo:
 //   - "."
 //   location: MESH_EXTERNAL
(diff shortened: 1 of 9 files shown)


hzxuzhonghu (changes requested):
Donot see the necessity of changing it, it is just an example

howardjohn:
ideal is probably example.com - it's specifically meant for cases like this

xichengliudui:
> 

the `httpbin.com` domain name does not exist and using it can be confusing for beginners，`httpbin.org` exists

xichengliudui:
do i need to close or update to `example.com`?

ericvn:
> do i need to close or update to `example.com`?

I think updating to example.com would solve the issue of using a link that isn't found, and still imply to a user that they should change it.

howardjohn (approved):

ericvn (approved):

hzxuzhonghu (approved):

PR #2202, opened 2022-01-07
Add NONE mode to interception mode

howardjohn:
This is already used in our injection template, but cannot actually be set, which seems like a bug. Its also in the pilot code: https://github.com/istio/istio/blob/e0207105bb09baf1299a5752b98d83240ba981af/pilot/pkg/model/context.go#L1044

I am actually a bit confused how this could have worked at all - is everyone using annotation to set it? @costinm @ramaraochavali ?


hzxuzhonghu:
Actually when do we need `None` for proxy, doesn't it mean no interception?  And if so, why do users inject sidecar?

hzxuzhonghu:
And there is also `CaptureMode` in Sidecar

zirain:
> Actually when do we need `None` for proxy, doesn't it mean no interception? And if so, why do users inject sidecar?

bypass for debug?

howardjohn:
This is needed for injection though, we don't read Sidecar there

hzxuzhonghu:
There is also an annotation who tells injector not to inject sidecar

howardjohn:
May want sidecar but not iptables setup, for http_proxy mode.

NONE mode already exists - it is just only able to be set using hacks like
annotations and env vars, instead of the proper proxy config enum

mesh/v1alpha1/proxy.proto
```diff
@@ -412,6 +412,9 @@ message ProxyConfig {
     // filtering and manipulation. This mode also configures the sidecar to run with the
     // `CAP_NET_ADMIN` capability, which is required to use `TPROXY`.
     TPROXY = 1;
+    // The `NONE` mode does not configure redirect to Envoy at all. This is an advanced
+    // configuration that typically requires changes to user applications.
```
linsun, line 416: What does it mean "typically requires changes to user applications"?  Do you mean user needs to change their app to handle app network connectivity in the app?

linsun:
> Actually when do we need `None` for proxy, doesn't it mean no interception? And if so, why do users inject sidecar?

same q from me :) would like to get clarity on this.

howardjohn:
> Actually when do we need None for proxy, doesn't it mean no interception? And if so, why do users inject sidecar?

> same q from me :) would like to get clarity on this.

There is other modesthan iptables. HTTP_PROXY, sending directly to localhost explicitly, etc. \

All of this already exists, it just can only be configured through back doors.

mesh/v1alpha1/proxy.proto
costinm: It means apps need to connect to 127.0.0.1:xxx instead of foo.ns.svc for TCP, and set HTTP_PROXY - most of the time it's a config change, not a code change - but sometimes code as well.

costinm: Worth mentioning 'this mode is needed for cases where iptables and NET_CAP_ADMIN are not possible' - and/or add to the other modes that it requires NET_CAP_ADMIN and iptables_restore.

costinm (approved):

costinm:
We still inject the sidecar - but not the init container, no iptables interception. This is for 2 main cases:
- the cluster doesn't allow NET_ADMIN or installing the CNI - i.e. the user has no permissions. Maybe not even permission to install an injector ( kube-inject can be used instead ). 

- the original reason - you want to avoid the overhead of iptables. Maybe you configure envoy to listen on a UDS socket and the app is using that. AFAIK iptables redirect with large number of connections can be less efficient. 

We had this feature for a very long time - and there are other ways to configure it.

hzxuzhonghu (approved):
Much clearer now

PR #2207, opened 2022-01-14
WorkloadSelector support for DestinationRule

kfaseela:
Adding support for workloadSelector in DR
as per RFC [Simplify Sidecar Egress For MTLS](https://docs.google.com/document/d/1UXpT3rZpE2uFeMh5KffAZAPAoZBLNSHc/)


bianpengyuan:
Curious, how does this interact with subset? Is it that the unmatched proxy won't have subset configured? If so should we also have workload selector at virtual service to make it consistent. There are definitely demand for this https://github.com/istio/istio/issues/36240 and https://github.com/istio/istio/issues/29434

howardjohn:
VirtualService already basically has workload selector with sourceLabels? I
*think* its functionally equivalent

hzxuzhonghu:
First, i wanna say i am totally for this proposal.
But i think this will break backward compatibility. Before a DR without wokloadSelector, it would be applied to all namespaces.

But with this, without workloadSelector set, it will be applied to the same namespace

linsun:
This is an interesting concept.  Could we think through what impact it will have for DR priority ordering: client side, server side, namespace, global.

BTW, is this workload selector for client side or server side?

kfaseela:
> First, i wanna say i am totally for this proposal. But i think this will break backward compatibility. Before a DR without wokloadSelector, it would be applied to all namespaces.
> 
> But with this, without workloadSelector set, it will be applied to the same namespace

I think the confusion came due to my wrong description of the attribute that it will be namespace specific. will correct it

howardjohn:
> workSelector by default is global, and we can use "exportTo" to limit namespace list.

I disagree, if workloadSelector is present it should only match current namespace. This matches all of our resources (except Gateway, and its been replaced by another mechanism in gateway-api and was a mistake IMO).

> DR will impact server side in TLS case, TLS mode in DR can be changed to override global TLS mode (e.g.STRICT), from this perspective, we can limit the "workSelector" to client side only, and keep server side behaviour as before.

DR only impacts server for circuit breakers, not TLS.

bianpengyuan:
> DR will impact server side in TLS case, TLS mode in DR can be changed to override global TLS mode (e.g.STRICT), from this perspective, we can limit the "workSelector" to client side only, and keep server side behaviour as before.

https://github.com/istio/istio/issues/36731#issuecomment-1015521037 as an example of why it would also benefit server side.

howardjohn:
> istio/istio#36731 (comment) as an example of why it would also benefit server side.

I think its still somewhat ambiguous and https://github.com/istio/api/pull/1754 would be preferred. For example, say I have a DR for `httpbin.default.svc.cluster.local` and selector `app=httpbin`. We would apply config on inbound and outbound side - use may have actually only wanted it on one or the other. Sidecar makes it very explicit

bianpengyuan:
> I think its still somewhat ambiguous and #1754 would be preferred

Yeah I agree. Adding selector to DR is more like a workaround for the inbound configuration issue, which might complicate the configuration or not even cover some of the use cases.

ramaraochavali:
not stale

hzxuzhonghu:
It is also useful to control outbound settings for a subset of clients.

networking/v1alpha3/destination_rule.proto
```diff
@@ -162,6 +163,28 @@ import "gogoproto/gogo.proto";
 //         simple: ROUND_ROBIN
 // ```
 // {{</tab>}}
+//
+// {{<tab name="v1alpha3" category-value="v1alpha3">}}
```
howardjohn, line 167: This is adding a third tab to a tab group of {v1alpha3,v1beta1}.

If we want to provide and example with workloadSelector, we should look like:


```
You can also use a workload selector .. blah blah...

{{<tabset category-name="selector-example">}}
{{tab }}
....
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -239,6 +262,11 @@ message DestinationRule {
   // the destination rule is declared in. Similarly, the value "*" is reserved and
   // defines an export to all namespaces.
   repeated string export_to = 4;
+  //
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `DestinationRule` configuration should be applied. If omitted, the `DestinationRule`
+  // configuration will be applied to all workload instances in the same namespace.
```
howardjohn, line 268: "Same namespace" is not require accurate, since it can apply cross namespace,

We should also note that with workload selector set it is NOT cross namespace.

howardjohn (commented):
Code LGTM, just needs some docs improvements

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: It should be clearly documented, whether this is selecting cross namespaces or within the same namespace. 

Also move this field up to first position.

networking/v1alpha3/destination_rule.proto
kfaseela: Not sure, if I fixed it properly. Could you please take another look now?

networking/v1alpha3/destination_rule.proto
kfaseela: I have tried to rephrase the comment.

networking/v1alpha3/destination_rule.proto
howardjohn: Looks great

howardjohn (approved):

networking/v1alpha3/destination_rule.proto
```diff
@@ -239,6 +288,13 @@ message DestinationRule {
   // the destination rule is declared in. Similarly, the value "*" is reserved and
   // defines an export to all namespaces.
   repeated string export_to = 4;
+  //
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `DestinationRule` configuration should be applied. If specified, the `DestinationRule`
+  // configuration will be applied only to the workload instances matching the workload selector
+  // label in the same namespace. Workload selectors do not apply across namespace boundaries.
+  // If omitted, the `DestinationRule` falls back to its default behavior.
+  istio.type.v1beta1.WorkloadSelector workload_selector = 5;
```
ramaraochavali, line 297: May be we should explain when this will be used vs. default behaviour? That is not clear to me while reading this comment.

@howardjohn WDYT?

kfaseela: Clarified in the comment

ramaraochavali (approved):
Thank you

kfaseela:
Thank you!

networking/v1alpha3/destination_rule.proto
```diff
@@ -239,6 +288,16 @@ message DestinationRule {
   // the destination rule is declared in. Similarly, the value "*" is reserved and
   // defines an export to all namespaces.
   repeated string export_to = 4;
+  //
+  // Criteria used to select the specific set of pods/VMs on which this
+  // `DestinationRule` configuration should be applied. If specified, the `DestinationRule`
+  // configuration will be applied only to the workload instances matching the workload selector
+  // label in the same namespace. Workload selectors do not apply across namespace boundaries.
```
hzxuzhonghu, line 295: I donot think this is consistent with other API @howardjohn 

Not a good idea to be inconsistent

howardjohn: Can you explain what about it is inconsistent? My expectation is that this behaves the same way as every other Istio API that has a workload selector?

hzxuzhonghu: Workload selectors do not apply across namespace boundaries.

For gateway, the selector works across namespaces.

hzxuzhonghu: Another point it from implement view: currently dr is used both fro outbound and inbound setting. Then if this only applies to the same namespace, does that mean we need to create two DRs for a service if the caller resides in different namespace as server?

kfaseela: @hzxuzhonghu : implementation can be seen at : https://github.com/istio/istio/pull/37174

kfaseela: @hzxuzhonghu regarding your comment about the default behaviour  of selectors in Gateway, please take a look at John's comment above already: https://github.com/istio/api/pull/2207#issuecomment-1016587329

ramaraochavali:
> But i think this will break backward compatibility. Before a DR without wokloadSelector, it would be applied to all namespaces.

> But with this, without workloadSelector set, it will be applied to the same namespace

@kfaseela Is this correct? I thought there is no change in behaviour for the case of DR with no workload selector?

kfaseela:
> > But i think this will break backward compatibility. Before a DR without wokloadSelector, it would be applied to all namespaces.
> 
> > But with this, without workloadSelector set, it will be applied to the same namespace
> 
> @kfaseela Is this correct? I thought there is no change in behaviour for the case of DR with no workload selector?

This is not correct. The namespace restriction is there only for workloadSelector set cases.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: I donot see how to deal with the inconsistency

hzxuzhonghu: it is different from not only gateway. but also authorization policy

```
If the authorization policy is in the root namespace, the selector
	// will additionally match with workloads in all namespaces.
```

howardjohn: But not EnvoyFilter, Sidecar, Telemetry, ProxyConfig, WasmPlugin, PeerAuthentication, or Request Authentication.

Gateway was a mistake IMO - we do not do this cross namespace in the gateway-api and we have a flag to disable cross namespace for Istio gateway.

AuthzPolicy is the odd one out, not DR, IMO.

hzxuzhonghu: It is very confusing and error-prone. What's the long-term plan? After upgrading API version, we should remove all this bad inconsistency

pkgulati:
How to achieve workloadSelector for incoming traffic. I want to apply max connections limit using Destination Rule only on SERVER side of the service using istio proxy (side car).

kfaseela:
@pkgulati : This was discussed under: https://github.com/istio/istio/issues/41235

PR #2212, opened 2022-01-20
add injectionSelectors

liuxu623:
`InjectionSelectors`  is a list of Kubernetes selectors that specify the set of namespaces, Istio will create istio-ca-root-cert configmap in selected namespaces.If omitted, Istio will create istio-ca-root-cert configmap in all namespaces except kubernetes system namespaces.

Fix https://github.com/istio/istio/issues/31115


Arhell (commented):
/ok-to-test

kfaseela:
@liuxu623 you need a rebase on this PR as well?

liuxu623:
> @liuxu623 you need a rebase on this PR as well?

@kfaseela Please help me review code, thank you.

kfaseela:
> > @liuxu623 you need a rebase on this PR as well?
> 
> @kfaseela Please help me review code, thank you.

sure..will do.. but considering this is an API change, I assume you would have already discussed this in one of the working group meetings, and got a consensus for this enhancement?

liuxu623:
> > > @liuxu623 you need a rebase on this PR as well?
> > 
> > 
> > @kfaseela Please help me review code, thank you.
> 
> sure..will do.. but considering this is an API change, I assume you would have already discussed this in one of the working group meetings, and got a consensus for this enhancement?

No, I didn't discuss this with working group  yet...

mesh/v1alpha1/config.proto
```diff
@@ -1063,8 +1063,31 @@ message MeshConfig {
   // configured globally via this field.
   istio.networking.v1alpha3.HTTPRetry default_http_retry_policy = 62;
 
+  // A list of Kubernetes selectors that specify the set of namespaces,
+  // Istio will create istio-ca-root-cert configmap in selected namespaces.
```
howardjohn, line 1067: If this is *only* for the istio-ca-root-cert configmap it seems like the name `namespaceSelectors` is far too broad?

emike922: Maybe this could become `injectionSelectors` and additionally control whether the sidecar injector accepts an injection request from a namespace?

howardjohn: That is already defined by the mutating webhook; a mesh config setting cannot actually control it at all (unless it becomes an operator for the Webhook)

emike922: I understand, but since scoping the istio-ca-root-cert configmap will allow deploying multiple distinct istiod instances, there may then also be multiple mutating webhook configs that _each_ invoke /inject on the same pod but different webhook services.
An istiod instance with `injectionSelector` could return false at injectRequired(), similar to the handling for "special" k8s namespace at https://github.com/istio/istio/blob/098cc47dfcbdca202208bd241d9dee35cd230a4d/pkg/kube/inject/inject.go#L201-L205, if the pod namespace does not match the selector.
I would not currently expect that turning it into an operator for the webhook will be necessary

howardjohn: They should use revisions so they only match their own distinct `istio.io/rev`.

if we want to define a full multitenancy proposal then I would recommended we have a holistic design doc on how this will work.

> An istiod instance with injectionSelector could return false at injectRequired()

This is much worse than the current Webhook scoping of istio.io/rev, since now ALL istiods are points of failure for a pod starting. Doing it in webhook selector -- which already exists using revision -- filters at the k8s level and does not add single point of failures

emike922: Agreed, filtering at k8s level using the webhook's namespace and object selectors is **much** preferable, but would require the webhook operator after all. Relying only on `istio.io/rev` will not offer the same functionality and "convenience" as the current, cluster-global deployment. Out of scope for this PR.

But maybe scoping for rootCA provisioning could still be accomplished here. Writing to every namespace, even those not intended to really act as active members of the service mesh, feels a bit intrusive. Wdyt?

howardjohn: I don't understand how istio.io/rev doesn't solve this. It was explicitly designed to handle the case if ensuring different namespace use different control planes for injection. Can you elaborate?

hzxuzhonghu:
Do you have a high-level design? There are too many factors to consider for multi-tenants, we need a design.

mesh/v1alpha1/config.proto
liuxu623: istio.io/rev can isolate differenet verison istiod，but it seem couldn't isolate multi same verison istiod.
https://files.slack.com/files-pri/T382U9E4U-F036F2RK07Q/image.png

liuxu623: @howardjohn John, do you think there is a way with existing methods to solve the isolation of config-map creation, so that I don't have to add an additional namespaceSelector like this.

emike922: > I don't understand how istio.io/rev doesn't solve this. It was explicitly designed to handle the case if ensuring different namespace use different control planes for injection. Can you elaborate?

Yes, revisions DO solve parts of the problem. But if we are talking multi-tenancy, each tenant should have its own rootCA and control a subset of namespaces both with and without revisions. This means that the same revision could occur in each tenant and yet be distinct/unrelated. Effectively, xSelectors (placeholder) should represent a namespace boundary beyond which a particular control plane instance should not make changes (s.a. creating config map or injecting proxies).

howardjohn: > istio.io/rev can isolate differenet verison istiod，but it seem couldn't isolate multi same verison istiod.
It does, there is no reason they need to be different versions.

> This means that the same revision could occur in each tenant and yet be distinct/unrelated.

I don't agree this is valid use case. Kubernetes is fundamentally not entirely multitenant. There are shared resources that Istio depends on - CRDs and webhooks. There _must_ be something distinguishing these between tenants, and some higher level admin ensuring these do not conflict. In Istio, that is revision. If you would like to introduce a different mechanism for this I would suggest a design doc holistically describing how multitenancy is expected to work.

emike922: You are right, CR consumption must also be limited along with the aforementioned CM creation and proxy injection. That is why I had expressed interest when extending discovery selectors to CRs was proposed in istio/istio#36639. But probably it will be better if both are part of the same config toggle instead of having to fine-tune several. And webhooks of course need some attention. I will try to see if I can condense all that into a design doc when I have some time..

kfaseela:
> Do you have a high-level design? There are too many factors to consider for multi-tenants, we need a design.

https://docs.google.com/document/d/1y4liRJbQW0NCMeQtqMma46flVqs-izV1/

PR #2213, opened 2022-01-20
Revert 1.13

stevenctl:
cc @zirain @douglas-reid 

These commits break istio/istio; I'm assuming these changes aren't actually implemented in the 1.13 timeframe. If they are we will have to cherry-pick later.


howardjohn (approved):
Approved. Note to others - these are still on master, just not 1.13

douglas-reid:
@stevenctl @howardjohn the PR for the proxyconfig change was out for review.  it fixes a critical breaking change issue that first appeared with 1.12. we cannot imho ship another release with the same breaking change.

stevenctl:
As soon as that merges in istio/istio master I'll get it into the 1.13 branch along with these API changes.

zirain:
@stevenctl https://github.com/istio/istio/pull/36809 is merged, I think it's ready to get everything back.

PR #2216, opened 2022-01-21
Clarify concurrency setting from proxy config

timonwong:
Signed-off-by: Tianpeng Wang <tpwang@alauda.io>

It seems since https://github.com/istio/istio/pull/32928, the concurrency behavior is changed so we need update the doc.

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -411,9 +411,10 @@ message ProxyConfig {
   int32 stat_name_length = 15;
 
   // The number of worker threads to run.
-  // If unset, this will be automatically determined based on CPU requests/limits.
-  // If set to 0, all cores on the machine will be used.
-  // Default is 2 worker threads.
+  // For sidecars:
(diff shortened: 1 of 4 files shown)


howardjohn:
I am not sure this is quite accurate. It is still the old behavior for some gateway install methods (https://istio.io/latest/docs/setup/additional-setup/gateway/).

Arhell (commented):
/ok-to-test

mesh/v1alpha1/proxy.proto
```diff
@@ -411,9 +411,10 @@ message ProxyConfig {
   int32 stat_name_length = 15;
 
   // The number of worker threads to run.
-  // If unset, this will be automatically determined based on CPU requests/limits.
-  // If set to 0, all cores on the machine will be used.
-  // Default is 2 worker threads.
+  // For sidecars:
+  // If set to 0, this will be automatically determined based on CPU requests/limits.
+  // If positive n, then n worker threads will be used.
+  // Otherwise, defaults to 2 worker threads.
```
linsun, line 417: Is this only for sidecar?

#2247

PR #2217, opened 2022-01-24
Move more examples to `credentialName`

howardjohn:
This is the recommended method for 6 releases now. There is still one
reference to the file mount option.


ericvn:
Do we want this in 1.13 as well?

ericvn (approved):

hzxuzhonghu (approved):

PR #2220, opened 2022-01-24
clarify multiple tokens case for request authentication API

yangminzhu:
Clarify the handling of multiple tokens and be consistent with the https://istio.io/latest/docs/concepts/security/#request-authentication

### DIFF
--- a/security/v1beta1/jwt.proto
+++ b/security/v1beta1/jwt.proto
@@ -119,6 +119,9 @@ message JWTRule {
   //   - name: x-jwt-assertion
   //     prefix: "Bearer "
   // ```
+  //
+  // Note: Requests with more than one tokens (at different locations) are not supported, the output principal of
+  // such requests is undefined.
   repeated JWTHeader from_headers = 6;
 
   // List of query parameters from which JWT is expected. For example, if JWT is provided via query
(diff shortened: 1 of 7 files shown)


security/v1beta1/jwt.proto
```diff
@@ -119,6 +119,9 @@ message JWTRule {
   //   - name: x-jwt-assertion
   //     prefix: "Bearer "
   // ```
+  //
+  // Note: Requests with more than one tokens (at different locations) are not supported, the output principal of
```
ericvn, line 123: Maybe
```suggestion
  // Note: Requests with multiple tokens (at different locations) are not supported, the output principal of
```
or maybe `more than one token` (removing the s).

Same comment in the other locations.

yangminzhu: done

ericvn (approved):

security/v1beta1/jwt.proto
```diff
@@ -119,6 +119,9 @@ message JWTRule {
   //   - name: x-jwt-assertion
   //     prefix: "Bearer "
   // ```
+  //
+  // Note: Requests with multiple tokens (at different locations) are not supported, the output principal of
+  // such requests is undefined.
```
hzxuzhonghu, line 124: The comment is good, it is confusing for users the fields below are repeated.

hzxuzhonghu (approved):

PR #2227, opened 2022-01-27
WasmPlugin add VmConfig

zirain:
https://github.com/envoyproxy/envoy/issues/14958 is merged

add `VmConfig` support get env in wasm extensions

@bianpengyuan @mathetake PTAL


extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +284,48 @@ enum PullPolicy {
...
+
+  // Allow the wasm file to include pre-compiled code on VMs which support it.
+  // Warning: this should only be enable for trusted sources as the precompiled code is not
+  // verified.
+  bool allow_precompiled = 4;
+
+  // If true and the code needs to be remotely fetched and it is not in the cache then NACK the configuration
+  // update and do a background fetch to fill the cache, otherwise fetch the code asynchronously and enter
+  // warming state.
+  bool nack_on_code_cache_miss = 5;
```
hzxuzhonghu, line 314: We donot need to expose all envoy options like these two

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +284,38 @@ enum PullPolicy {
...
   Always = 2;
 }
+
+// Configuration for a Wasm VM.
+message VmConfig {
+  // An ID which will be used along with a hash of the wasm code (or the name of the registered Null
+  // VM plugin) to determine which VM will be used for the plugin. All plugins which use the same
+  // *vm_id* and code will use the same VM. May be left blank. Sharing a VM between plugins can
+  // reduce memory utilization and make sharing of data easier which may have security implications.
+  string vm_id = 1;
```
bianpengyuan, line 294: Let's expose only the needed options first. I think we can start with only `environment_variables`.

zirain: agree, I will update

zirain:
release notes will add when implemented in istio repo

extensions/v1alpha1/wasm.proto
```diff
@@ -241,6 +242,9 @@ message WasmPlugin {
   // value, the ordering will be deterministically derived from name and
   // namespace of the `WasmPlugins`. Defaults to `0`.
   google.protobuf.Int64Value priority = 10;
+
+  // Configuration for finding or starting VM.
```
bianpengyuan, line 246: Comment needs to update. Current config does not involve `finding` or `starting` the VM. Also probably would be good to link a doc about what is Wasm VM here if Envoy has it somewhere.

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +284,22 @@ enum PullPolicy {
...
   Always = 2;
 }
+
+// Configuration for a Wasm VM.
+message VmConfig {
+  // Specifies environment variables to be injected to this VM which will be available through
+  // WASI's ``environ_get`` and ``environ_get_sizes`` system calls. Note that these functions are mostly implicitly
+  // called in your language's standard library, so you do not need to call them directly and you can access to env
+  // vars just like when you do on native platforms.
+  // Warning: Envoy rejects the configuration if there's conflict of key space.
```
bianpengyuan, line 294: This warning seems to be tie very closely to envoy's implementation. i.e. key space does not seem to make sense to istio users. Can we make it more relevant to istio user? And we probably should replace all presence of Envoy to Istio proxy.

bianpengyuan (approved):

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +285,19 @@ enum PullPolicy {
   // this plugin.
   Always = 2;
 }
+
+// Configuration for a Wasm VM.
+// more details can be found [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto#extensions-wasm-v3-vmconfig).
+message VmConfig {
+  // Specifies environment variables to be injected to this VM.
+  EnvironmentVariables environment_variables = 1;
```
howardjohn, line 293: nit: should we use `env` to match Kubernetes naming?

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +285,19 @@ enum PullPolicy {
...
+}
+
+message EnvironmentVariables {
+  // The keys of *Envoy's* environment variables exposed to this VM. In other words, if a key exists in Envoy's environment
+  // variables, then that key-value pair will be injected. Note that if a key does not exist, it will be ignored.
+  repeated string host_env_keys = 1;
+
+  // Explicitly given key-value pairs to be injected to this VM in the form of "KEY=VALUE".
+  map<string, string> key_values = 2;
```
howardjohn, line 302:
```
keyValues:
   foo: bar
```

seems a bit odd UX here.

perhaps we could use the same as Kubernetes?

```yaml
env:
- name: FOO # explicit
  value: bar
- name: HOST_BAR # value from host, implicit since value is not set
- name: HOST_BAR
  fromHost: true # alternative to above, explicit set host
```

wdyt? cc @bianpengyuan @louiscryan

bianpengyuan: Yeah this looks better. Probably should make from host explicit, to make it not ambiguous about whether the env variable is set to empty string or read from host.

zirain: @howardjohn @bianpengyuan how do you think about this?

```yaml
hostEnv:
- POD_NAME
- POD_NAMESPACE
env:
  MESH_NAME: mesh1
  MESH_ZONE: cn-east1
```

bianpengyuan: Seems to me it is still more intuitive by keeping the same UX as k8s env.

zirain:
```yaml
env:
- name: FOO # explicit
  value: bar
- name: HOST_BAR # value from host, implicit since value is not set
   value: host_bar
- name: HOST_BAR
  fromHost: true # alternative to above, explicit set host
```
with this format, how do we pass `HOST_BAR`? looks like related to order of slice.

zirain:
@bianpengyuan @howardjohn ptal

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +313,41 @@ enum PullPolicy {
...
+	// Escaped references will never be expanded, regardless of whether the variable
+	// exists or not.
+	// Defaults to "".
+	// +optional
+  optional string value = 2;
+}
+
+enum EnvValueSource {
+  // *Istio-proxy's* environment variables exposed to this VM.
+  HOST = 0;
```
howardjohn, line 349: I would make this:
```
INLINE = 0;
HOST =1;
```

That way users do not need to set `valueFrom` at all when specifying `key`. So it looks like

```
//     env:
//	   - name: POD_NAME
//       valueFrom: HOST
//     - name: TRUST_DOMAIN
//       value: "cluster.local"
```

The other option is to make `value` and `valueFrom` a oneof. With valueFrom only having `UNSPECIFIED` and `HOST`. That ends up looking the same in YAML form though

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +313,41 @@ enum PullPolicy {
...
+  // Required
+  // Name of the environment variable. Must be a C_IDENTIFIER.
+  string name = 1;
+
+  // Required
+  // Source for the environment variable's value.
+  EnvValueSource value_from = 3;
+
+  // Variable references $(VAR_NAME) are expanded
+	// using the previously defined environment variables in the container and
```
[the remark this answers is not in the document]
zirain: updated

zirain:
blocked by https://github.com/istio/api/issues/2235

zirain:
updated, @howardjohn @bianpengyuan ptal

extensions/v1alpha1/wasm.proto
```diff
@@ -77,6 +78,33 @@ import "type/v1beta1/selector.proto";
 //     openid_realm: ingress
 // ```
 //
+// This is the same as the last example, but using VmConfig.
```
howardjohn, line 81:
```suggestion
// This is the same as the last example, but using VmConfig to configure environment variables in the VM.
```

extensions/v1alpha1/wasm.proto
```diff
@@ -77,6 +78,33 @@ import "type/v1beta1/selector.proto";
...
+// spec:
+//   selector:
+//     labels:
+//       istio: ingressgateway
+//   url: oci://private-registry:5000/openid-connect/openid:latest
+//   imagePullPolicy: IfNotPresent
+//   imagePullSecret: private-registry-pull-secret
+//   phase: AUTHN
+//   pluginConfig:
+//     openid_server: authn
```
howardjohn, line 98: when should a user use pluginConfig vs vmConfig?

howardjohn (approved):

howardjohn:
lgtm, I removed @bianpengyuan review since it was stale. @bianpengyuan can you take a look to make sure its good? Thanks!

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +312,41 @@ enum PullPolicy {
...
+  // Name of the environment variable. Must be a C_IDENTIFIER.
+  string name = 1;
+
+  // Required
+  // Source for the environment variable's value.
+  EnvValueSource value_from = 3;
+
+  // Variable references $(VAR_NAME) are expanded
+  // using the previously defined environment variables in the container and
+  // any service environment variables. If a variable cannot be resolved,
```
bianpengyuan, line 335: This seems to be copied from k8s doc, which is a bit confusing. I don' understand what service environment mean here. Can we update the wording a bit to make it related to Istio / wasm?

zirain: updated

extensions/v1alpha1/wasm.proto
```diff
@@ -280,3 +312,35 @@ enum PullPolicy {
...
+message EnvVar {
+  // Required
+  // Name of the environment variable. Must be a C_IDENTIFIER.
+  string name = 1;
+
+  // Required
+  // Source for the environment variable's value.
+  EnvValueSource value_from = 3;
+
+  // Variable references $(VAR_NAME) are expanded
```
bianpengyuan, line 333:
```suggestion
  // Value for the environment variable.
```

bianpengyuan (approved):

PR #2230, opened 2022-02-04
Running proto-lock

kfaseela:
As per https://github.com/istio/api#updating `make proto-lock`
has to be run as part of new api changes. I was trying
to run the same for my PR https://github.com/istio/api/pull/2207
and saw some additional changes added in proto.lock which
was not introduced by my change. Hence putting them in a separate
PR here to check if these changes are needed or not.

### DIFF
--- a/proto.lock
+++ b/proto.lock
@@ -32468,6 +32468,11 @@
                 "type": "AccessLogging"
               },
               {
+                "id": 17,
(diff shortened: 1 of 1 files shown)


howardjohn (approved):

kfaseela:
> /retest

Thanks!

ericvn (approved):

PR #2232, opened 2022-02-07
Updating LB enum to allow switch in default

nmittler:
Switching the default zero value to UNSPECIFIED. This allows us
to better detect when the user has set a value for the enum. If
the user previously specified ROUND_ROBIN, it will still be
handled properly, but will be assigned a new enum value.

In addition, the value LEAST_CONN is now deprecated in favor of
a new value LEAST_REQUEST. It was previously poorly named and
only remains for backward compatibility.

There is a chance that this change could break existing tools
if they rely on binary marshaling/unmarshaling of the proto.
There are currently no tools within Istio that do this,
however.


howardjohn (commented):
I believe this is safe because:

If a user was using wire protocol, previously `0` was ROUND_ROBIN. Now its UNSPECIFIED. If they keep using RR, that is fine as a default.

If a user is using YAML, there is no change

cc @louiscryan @costinm @smawson if you have any experience why this might not be OK

louiscryan:
I agree with @howardjohn here. I believe this is safe.

louiscryan (approved):

networking/v1alpha3/destination_rule.proto
```diff
@@ -442,13 +442,12 @@ message Subset {
 message LoadBalancerSettings {
   // Standard load balancing algorithms that require no tuning.
   enum SimpleLB {
-    // Round Robin policy. Default
-    ROUND_ROBIN = 0;
+    // No load balancing algorithm has been specified by the user. Istio
+    // will select an appropriate default.
+    UNSPECIFIED = 0;
```
linsun, line 447: Could you clarify which one would be used when user doesn't specify one?  @nmittler

linsun: I assume it is UNSPECIFIED

linsun:
LGTM, happy to approve tmr if there is no objections.

networking/v1alpha3/destination_rule.proto
nmittler: Yeah that's right.

ericvn (approved):
LGTM

PR #2239, opened 2022-02-08
Add option to set runtime flags

howardjohn:
For https://github.com/istio/istio/issues/37202

I was originally going to use env var hacks - we already have similar
for `ISTIO_META` and `XDS_HEADER`. However, envoy runtime vars take the
format of `foo.bar_baz=something`. Env vars do not allow `.`, or really
any special charecters other than `_`, so we cannot easily use them.
Instead, seems better to have an explicit field.


ericvn (approved):

mesh/v1alpha1/proxy.proto
```diff
@@ -460,6 +460,10 @@ message ProxyConfig {
   // Names starting with `ISTIO_META_` will be included in the generated bootstrap and sent to the XDS server.
   map<string,string> proxy_metadata = 24;
 
+  // Envoy [runtime configuration](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/runtime) to set during bootstrapping.
+  // This enables setting experimental, unsafe, unsupported, and deprecated features that should be used with extreme caution.
+  map<string,string>  runtime_values = 37;
```
ramaraochavali, line 465: nit: may be runtime_flags?

PR #2241, opened 2022-02-09
Add HeadersToDownstreamOnAllow to ext_authz

hanxiaop:
First step to implement https://github.com/istio/istio/issues/36950


mesh/v1alpha1/config.proto
```diff
@@ -619,8 +619,20 @@ message MeshConfig {
       // - Suffix match: "*abc" will match on value "abc" and "xabc".
       repeated string headers_to_downstream_on_deny = 8;
 
+      // List of headers from the authorization service that should be forwarded to downstream when the authorization
+      // check result is allowed (HTTP code 200).
+      // If not specified, the original request will not be modified and forwarded to backend as-is.
```
yangminzhu, line 624: `If not specified, the original request will not be modified and forwarded to backend as-is`

This is field is about the response and has nothing to do with the request to backend right?

hanxiaop: Thanks! Fixed.

mesh/v1alpha1/config.proto
```diff
@@ -619,8 +619,20 @@ message MeshConfig {
       // - Suffix match: "*abc" will match on value "abc" and "xabc".
       repeated string headers_to_downstream_on_deny = 8;
 
+      // List of headers from the authorization service that should be forwarded to downstream when the authorization
+      // check result is allowed (HTTP code 200).
+      // If not specified, the original request will not be modified and forwarded to downstream as-is.
```
yangminzhu, line 624: "If not specified, the original **response** will not be modified ..."

This field is adding extra headers to the response to the downstream, please update the comment.

yangminzhu (approved):
Looks good, just a small comment.

mesh/v1alpha1/config.proto
hanxiaop: Fixed. Thanks!

hanxiaop:
@ericvn @howardjohn PTAL.

PR #2242, opened 2022-02-10
fix fuzz

zirain:
fixed https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_istio/37254/unit-tests_istio/1491600327076483072


extensions/v1alpha1/wasm.proto
```diff
@@ -334,7 +334,7 @@ message EnvVar {
   // Note that if `value_from` is `HOST`, it will be ignored.
   // Defaults to "". 
   // +optional
```
bianpengyuan, line 336: Actually we probably should also remove `+optional`? I don't think it mean anything at Istio doc gen.

extensions/v1alpha1/wasm.proto
```diff
@@ -334,7 +334,7 @@ message EnvVar {
...
-  optional string value = 2;
+  string value = 2;
```
bianpengyuan, line 337: Curious why optional is causing the issue?

zirain: have no idea about this, remove first to unblock anything.

extensions/v1alpha1/wasm.proto
zirain: you're right

bianpengyuan (approved):

extensions/v1alpha1/wasm.proto
howardjohn: Technically it can work - there is a test which needs an Allowlist of interface types - which optional generates - due to some limitations. But we could have just added it.

That being said I think `optional` is hard to use and we do not use it anywhere else so this is the right fix, IIMO

howardjohn (approved):

PR #2247, opened 2022-02-11
Update stale descriptions

### LINKED ISSUE #1783: ProxyConfig.concurrency documentation is unclear
**Bug description**

`istio.mesh.v1alpha1.ProxyConfig.concurrency` documentation is unclear

[x] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience

**Expected behavior**

The docs should not contain self-contradictions, especially in the same paragraph. :)

**Steps to reproduce the bug**

Visit https://istio.io/v1.8/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig, rendered from https://github.com/istio/api/blob/c2df7c41fd8e4f27ef9aea3bdbcd1fa1eede4a00/mesh/v1alpha1/proxy.proto#L372-L376

Notice that the comment for `concurrency` is self-contradicting:

> The number of worker threads to run. 
> **If unset**, this will be automatically determined based on CPU requests/limits. 
> If set to 0, all cores on the machine will be used. 
> **Default** is 2 worker threads.

since it claims that there are two different behaviors for the same case ("if unset" and "default"), something that should be impossible outside of quantum computing. In our deployments at least it seems that the actual behavior is the latter (2 worker threads). The apparent self-contradiction was added in #1493. 

It is also possible we're simply misinterpreting the comment, but then in this case we really have no idea of what the correct meaning could be.

There is a separate point to be made about the sanity of such a default (arguably automatically calculating it as in the former case would make more sense), but let's focus on at least getting the docs right for now. Likely related: https://github.com/istio/istio/issues/23470.

cc @shamsher31 @howardjohn

ktalg:
according to: https://github.com/istio/istio/issues/23470#issuecomment-630302121
fix #1783

### DIFF
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.gen.json
@@ -1410,7 +1410,7 @@
             "format": "int32"
           },
           "concurrency": {
-            "description": "The number of worker threads to run. If unset, this will be automatically determined based on CPU requests/limits. If set to 0, all cores on the machine will be used. Default is 2 worker threads.",
+            "description": "The number of worker threads to run. If set to 0, this will be automatically determined based on CPU requests/limits. Default is 2 worker threads.",
             "type": "integer",
             "nullable": true
           },
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
@@ -2759,8 +2759,7 @@ No
 <td><code><a href="https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#int32value">Int32Value</a></code></td>
 <td>
 <p>The number of worker threads to run.
-If unset, this will be automatically determined based on CPU requests/limits.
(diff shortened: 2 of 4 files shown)


ktalg:
> @ktalg: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> release-notes_api 	[d5a2a3d](https://github.com/istio/api/commit/d5a2a3de7715dfbf75fb47bfa46be31dbb5f155d) 	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_api/2247/release-notes_api/1492482691881242624) 	false 	`/test release-notes_api`

@ericvn 
Should I add a release note?

ericvn:
Related to https://github.com/istio/api/pull/2216

ericvn:
Looking through the referenced PRs and some other PRs, I'm not sure this is accurate. @howardjohn probably has the most knowledge of the gateway/sidecar settings noted here.

PR #2253, opened 2022-02-22
update envoyfilter EXTENSION_CONFIG demo

zirain:
the original demo will be rejected:
```console
ACK ERROR httpbin-74fb669cc6-tpjfh.default-4 Internal:Error adding/updating listener(s) virtualInbound: Error: terminal filter named envoy.filters.http.router of type envoy.filters.http.router must be the last filter in a http filter chain.
```

### DIFF
--- a/networking/v1alpha3/envoy_filter.proto
+++ b/networking/v1alpha3/envoy_filter.proto
@@ -354,16 +354,20 @@ import "networking/v1alpha3/sidecar.proto";
 //   # The second patch instructs to apply the above Wasm filter to the listener/http connection manager.
 //   - applyTo: HTTP_FILTER
 //     match:
-//       context: SIDECAR_INBOUND
+//       listener:
+//         filterChain:
+//           filter:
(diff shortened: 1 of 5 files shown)


networking/v1alpha3/envoy_filter.proto
```diff
@@ -354,16 +354,20 @@ import "networking/v1alpha3/sidecar.proto";
 //   # The second patch instructs to apply the above Wasm filter to the listener/http connection manager.
 //   - applyTo: HTTP_FILTER
 //     match:
-//       context: SIDECAR_INBOUND
+//       listener:
```
howardjohn, line 357: Should we recommend users use a WasmPlugin instead?

howardjohn (approved):

networking/v1alpha3/envoy_filter.proto
zirain: maybe we should not place a wasm ecds demo here?

PR #2261, opened 2022-03-03
Enable configuring Envoy private key provider through ProxyConfig

rveerama1:
**Introduction**

Since 1.20 Envoy has supported configuring private key providers over xDS. Also, the first in-tree private key provider (CryptoMB) was released in Envoy 1.20 contrib build (Ref PR: https://github.com/envoyproxy/envoy/pull/17826 ).

The idea in CryptoMB private key provider is that incoming TLS handshakes’ RSA operations are accelerated using AVX-512 multi-buffer instructions. For more details, see this Intel whitepaper which contains some more information about the AVX-512 instructions and potential performance increase: https://www.intel.com/content/www/us/en/architecture-and-technology/crypto-acceleration-in-xeon-scalable-processors-wp.html. 
An introductory presentation is available here :  https://www.intel.com/content/www/us/en/developer/videos/accelerate-envoy-with-advanced-vector-extensions.html

The private key providers can be configured using SDS protocol: the private_key field in Envoy Secret is replaced with the private key provider configuration. For example,

```
tls_certificates:
   certificate_chain: { "filename": "/path/cert.pem" }
   private_key: { "filename": "/path/key.pem" }

could be replaced with:

 tls_certificates:
   certificate_chain: { "filename": "/path/cert.pem" }
   private_key_provider:
     provider_name: cryptomb
     typed_config:
       "@type": type.googleapis.com/envoy.extensions.private_key_providers.cryptomb.v3alpha.CryptoMbPrivateKeyMethodConfig
       private_key: { "filename": "/path/key.pem" }
       poll_delay: 0.1s
```

The problem is that private key providers are typically host-dependent: for example, CryptoMb provider requires a host with support for AVX-512 instruction set. Therefore, we propose an extension mechanism, which allows for dynamic configuration of private key providers.

To set the mesh wide defaults, configure the `defaultConfig` section of `meshConfig`. For example:
```
    meshConfig:
      defaultConfig:
        privateKeyProvider:
          cryptomb:
            pollDelay: 0.01s
```
This can also be configured on a per-workload basis by configuring the `proxy.istio.io/config` annotation on the pod.
For example:
```
    annotations:
      proxy.istio.io/config: |
        privateKeyProvider:
          cryptomb:
            pollDelay: 0.01s
```


mesh/v1alpha1/proxy.proto
```diff
@@ -292,6 +293,42 @@ message Topology {
   ForwardClientCertDetails forward_client_cert_details = 2;
 }
 
+// PrivateKeyProvider defines configuration for gateways and sidecars. This can be confgured
```
irisdingbj, line 296: // PrivateKeyProvider defines private key configuration for gateways and sidecars. This can be configured

irisdingbj:
@howardjohn  This is the API change part PR and the relevant istio change PR is https://github.com/istio/istio/pull/37681
Please help to give a review for this, Thanks!

rveerama1:
Istio proxy relevant change is here: https://github.com/istio/proxy/pull/3752

mesh/v1alpha1/proxy.proto
```diff
@@ -292,6 +293,42 @@ message Topology {
...
+//         privateKey:
+//           inlineString: "This field will be filled by istio-agent for sidecars, users should not to set it in the yaml"
+//         pollDelay: 0.01s
+// ```
+message PrivateKeyProvider {
+  // PrivateKeyProvider Name: "cryptomb"
+  string name = 1;
+
+  // Private key provider specific configuration
+  google.protobuf.Struct config = 2;
```
howardjohn, line 329: Why we need a whole generic config struct? Should we just have an Enum with options like `DEFAULT, CRYPTOMB, ...`?

howardjohn (changes requested):
I am not a huge fan of exposing users to the raw underlying Envoy types -- especially since they are NOT passthrough and istio needs to have code to handle each type.

mesh/v1alpha1/proxy.proto
irisdingbj: generic type is  used here to make sure we do not  expose users to the raw underlying Envoy type.  We can not limit supported private key provider via Enum --suppose in the future there might be other new private key provider be introduced in envoy.  

@rveerama1 please add any info I am missing here.

rveerama1: 1). If we are expecting some values from the user (e.g poll_delay value from user in CryptoMB private key provider configuration, private_key will be filled by istiod/istio-agent). It will allow us to validate and map from struct to respective type in source code.

2.) [EnvoyFilter](https://github.com/istio/api/blob/master/networking/v1alpha3/envoy_filter.proto#L771) uses generic struct type to pass different types of dynamic data. Thought this kind of mechanism is recommended.

3) It will allow others to introduce new private key providers for Envoy without affecting istio.io/api. Just add relevant code in istio.io/istio sources.

lambdai:
The happy path is exciting. 
How wide is the AVX-512? What if the the host doesn't offer AVX-512, e.g on a stale VM?

mesh/v1alpha1/proxy.proto
rveerama1: How about something like this ?
`
// CryptoMBPrivateKeyProvider configuration
message CryptoMBPrivateKeyProvider {
  google.protobuf.Duration poll_delay = 1;
}

// PrivateKeyProvider defines private key configuration for gateways and sidecars. This can be configured
// mesh wide or individual per-workload basis.
message PrivateKeyProvider {
  // REQUIRED. PrivateKeyProvider name ("cryptomb").
  string name = 1;

  // REQUIRED. Specifies detailed configuration for the Private key provider.
  oneof provider {
    CryptoMBPrivateKeyProvider cryptomb = 2;
  }
}
`

rveerama1:
> What if the the host doesn't offer AVX-512, e.g on a stale VM?

This can be configured mesh wide and per-workload. In this case, user/admin has to configure per-workload basis and deploy it based on some label to that particular node which supports AVX-512.

ipuustin:
> The happy path is exciting. How wide is the AVX-512? What if the the host doesn't offer AVX-512, e.g on a stale VM?

The operations are https://en.wikipedia.org/wiki/AVX-512#IFMA calls, so the registers are 512 bits wide. Internally https://github.com/intel/ipp-crypto/tree/develop/sources/ippcp/crypto_mb library is called to perform the cryptographic operations.

irisdingbj:
@howardjohn Can you please review the latest change? Thanks!

mesh/v1alpha1/proxy.proto
```diff
@@ -292,6 +292,29 @@ message Topology {
   ForwardClientCertDetails forward_client_cert_details = 2;
 }
 
+// PrivateKeyProvider defines private key configuration for gateways and sidecars. This can be configured
+// mesh wide or individual per-workload basis.
+message PrivateKeyProvider {
```
zirain, line 297: I'm not sure can we make this as an kind of ExtensionProvider? 
I think there's a long term goal about move all things like this out of `MeshConfig`, https://docs.google.com/document/d/1WDKX-Rk_FYD2L8Ner_w7kic93ZUMPvyG84ymcZZ9BGY/edit#

irisdingbj: Thanks for the review!  we have explored the possibility to make it part of extensionProvider.    However it does not work due to two reasons:
1.  We do not have a specific CRD as telemetry API or authorization CRD to reference extensionProvider from meshconfig for a specific workload
2. We need support per pod/workload defination besides one global mesh one.

mesh/v1alpha1/proxy.proto
```diff
@@ -292,6 +292,29 @@ message Topology {
...
+    // How long to wait until the per-thread processing queue should be processed. If the processing queue
+    // gets full (eight sign or decrypt requests are received) it is processed immediately.
+    // However, if the queue is not filled before the delay has expired, the requests already in the queue
+    // are processed, even if the queue is not full.
+    // In effect, this value controls the balance between latency and throughput.
+    // The duration needs to be set to a non-zero value.
+    google.protobuf.Duration poll_delay = 1;
+  }
+
+  // REQUIRED. PrivateKeyProvider name (e.g: "cryptomb").
```
howardjohn, line 309: name is no longer required, right? It is implied by the oneof type

howardjohn (commented):
This LGTM other than the small comment.

I would like to see approval from some set of @shankgan @myidpt @jacob-delgado @costinm before merging though to make sure there is broader agreement (not all of them needed)

jacob-delgado:
~~We are trying to get rid of `meshConfig` and `meshConfig.defaultConfig`. I'm thinking this should this be moved to `networking/v1beta1/proxy_config.proto` and not be used in`meshConfig.defaultConfig` or annotations. I'd like to see what the status is of `ProxyConfig` though before we commit a change.~~

Since the user cannot put node labels/tolerations and use affinity/anti-affinity on namespaces it needs to go on something that can be placed on a deployment, statefulset or daemonset. For this case `ProxyConfig` won't suffice and this is the right place to put it.

mesh/v1alpha1/proxy.proto
jacob-delgado: I looked at the WIP PR, if this is kept then this should be an enum and not a string. Ideally we could key off of the provider type.

jacob-delgado:
LGTM, but it'd be nice to get rid of the `name` field so that we don't have magic strings in the istio codebase.

rveerama1:
Thanks for the reviews @howardjohn @jacob-delgado, I have removed the 'name' part and updated the PR. PTAL.

ericvn (approved):

PR #2266, opened 2022-03-04
Telemetry API: Add WorkloadMode selection to Tracing/Logging

douglas-reid:
There is a need to filter tracing and logging behavior based on role of the involved workload within Istio. This PR provides a structured API control for such behavior (as was initially discussed in the API design review). This brings tracing and logging behavior to parity with metrics. 

Addresses: https://github.com/istio/istio/issues/37665


telemetry/v1alpha1/telemetry.proto
```diff
@@ -337,19 +340,33 @@ message Tracing {
...
+  // Envoy, Envoy will break the trace because it cannot interpret the Request
+  // ID. By setting this value to false, we can prevent Envoy from sampling
+  // based on the Request ID. As a result, the trace will not be broken even if
+  // the Request ID is not in the Envoy format. [Trace Context
+  // Propagation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/tracing#trace-context-propagation)
   // provides more information on Request ID handling.
   // $hide_from_docs
   google.protobuf.BoolValue use_request_id_for_trace_sampling = 6;
+
+  message TracingSelector {
```
zirain, line 356: can this share with `LogSelector`?

zirain: at least share the `enum`?

telemetry/v1alpha1/telemetry.proto
```diff
@@ -600,4 +635,19 @@ message AccessLogging {
...
+      OUTBOUND_ONLY = 2;
+    }
+
+    // This determines whether or not to apply the access logging configuration
+    // based on the direction of traffic relative to the proxied workload.
+    TrafficDirection traffic_direction = 1;
+  }
+
+  // Allows tailoring of behavior to specific conditions.
+  LogSelector match = 4;
```
zirain, line 652: what's the different between `match` and `filter`? Is `selector` an kind of `filter`?

douglas-reid: Structured vs. unstructured, really. For instance, `traffic direction` is not an attribute exposed in order to filter on, currently.

telemetry/v1alpha1/telemetry.proto
douglas-reid: It definitely can share the enum. The only hang-up, really, is that `WorkloadMode` in the `MetricSelector` covers the same concept. Perhaps we should reuse that in both Tracing/Logging ?

As for the same selector, my current feeling is that maybe they diverge (like `MetricSelector` did) ?

telemetry/v1alpha1/telemetry.proto
zirain: make sense.

telemetry/v1alpha1/telemetry.proto
zirain: > As for the same selector, my current feeling is that maybe they diverge (like MetricSelector did) ?
agree

telemetry/v1alpha1/telemetry.proto
```diff
@@ -600,4 +635,19 @@ message AccessLogging {
...
+      INBOUND_ONLY = 1;
+      OUTBOUND_ONLY = 2;
+    }
+
+    // This determines whether or not to apply the access logging configuration
+    // based on the direction of traffic relative to the proxied workload.
+    TrafficDirection traffic_direction = 1;
+  }
+
+  // Allows tailoring of behavior to specific conditions.
```
bianpengyuan, line 651: Probably need to better document the relationship between match and filter (i.e. AND, nor OR).

telemetry/v1alpha1/telemetry.proto
```diff
@@ -337,19 +340,33 @@ message Tracing {
...
   google.protobuf.BoolValue use_request_id_for_trace_sampling = 6;
+
+  message TracingSelector {
+    enum TrafficDirection {
+      ALL_TRAFFIC = 0;
+      INBOUND_ONLY = 1;
+      OUTBOUND_ONLY = 2;
+    }
+
+    // This determines whether or not to apply the access logging configuration
```
bianpengyuan, line 363: apply the tracing configuration?

telemetry/v1alpha1/telemetry.proto
```diff
@@ -337,19 +340,33 @@ message Tracing {
...
+    // based on the direction of traffic relative to the proxied workload.
+    TrafficDirection traffic_direction = 1;
```
hzxuzhonghu, line 365: How could we extend if we need another select condition in the future? Is the logic `And`/`OR`

douglas-reid: The original idea was `AND`'d clauses. For example, if you had a second component for `protocol`:

```
traffic_direction: INBOUND
protocol: TCP
```
would mean that the configuration matched **only** inbound TCP traffic.

```
protocol: TCP
```
would mean any TCP.

If you wanted `OR`, you would (potentially) use a different `access_logging` stanza (with some ordering logic potentially -- like first to match, or similar).

telemetry/v1alpha1/telemetry.proto
douglas-reid: Tried to add some commentary above. Does that help?

telemetry/v1alpha1/telemetry.proto
douglas-reid: Added some comments. Probably need more. PTAL.

telemetry/v1alpha1/telemetry.proto
zirain: can we merge the `filter` to `match`? 

```yaml
match:
  filter: <some-cel-filter-expression>
  wordloadMode: CLIENT
  somefilterMayAddInTheFuture: <>
```

douglas-reid: Not without breaking backwards compatibility with existing resources. Maybe as we consider a promotion to Beta ?

There is some distinction (though, importantly, maybe not for the user) between `match.mode` and `filter`. Primarily, the control plane can use `match.mode` to decide where to apply the configuration, whereas the data plane must evaluate `filter`. As a result, the `match` clause can allow for specific tailoring of generated config in a way that `filter` cannot.  I don't know whether or not to keep this distinction in the presented API.

telemetry/v1alpha1/telemetry.proto
```diff
@@ -600,4 +631,19 @@ message AccessLogging {
...
+  // configuration sent to individual workloads. Finer-grained logging behavior
+  // can be further configured via `filter`.
+  message LogSelector {
+    // This determines whether or not to apply the access logging configuration
+    // based on the direction of traffic relative to the proxied workload.
+    WorkloadMode mode = 1;
+  }
+
+  // Allows tailoring of logging behavior to specific conditions.
+  LogSelector match = 4;
```
hzxuzhonghu, line 648: How about moving in front

telemetry/v1alpha1/telemetry.proto
```diff
@@ -337,19 +340,29 @@ message Tracing {
...
+  // based on the Request ID. As a result, the trace will not be broken even if
+  // the Request ID is not in the Envoy format. [Trace Context
+  // Propagation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/tracing#trace-context-propagation)
   // provides more information on Request ID handling.
   // $hide_from_docs
   google.protobuf.BoolValue use_request_id_for_trace_sampling = 6;
+
+  // TracingSelector provides a coarse-grained ability to configure tracing
+  // behavior based on certain traffic metadata (such as traffic direction).
+  message TracingSelector {
```
hzxuzhonghu, line 358: `TracingMatch` looks better? 

Selector usually is based on labels, this  is deeply rooted in the heart of user

douglas-reid: For the `metrics` bit, we already used `MetricSelector`.  `TracingSelector` aligns with that choice.

Users will interact via `match` in config names, however:

```yaml
match:
  mode: CLIENT
```

telemetry/v1alpha1/telemetry.proto
hzxuzhonghu: I mean move this field in front, just like WorkloadSelector. Could be more clear to user

```
message AccessLogging {
  // Allows tailoring of logging behavior to specific conditions.
  LogSelector match = 4;
  repeated ProviderRef providers = 1;
```

douglas-reid: Ah ok thanks. Wanted to be sure I understood. Moved them per your suggestion.

PR #2274, opened 2022-03-11
WIP: add auto-sni and auto-san fields in ClientTLSSettings

kfaseela:
RFC: [auto-sni and auto-san support](https://docs.google.com/document/d/1pTUl-Ng3nXAWJb7UGJtalftznpxQEfID/edit)

Previous activity can be seen in #1773.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -691,6 +691,10 @@ spec:
                                 description: TLS related settings for connections
                                   to the upstream service.
                                 properties:
+                                  autoSan:
+                                    type: boolean
+                                  autoSni:
+                                    type: boolean
                                   caCertificates:
                                     type: string
                                   clientCertificate:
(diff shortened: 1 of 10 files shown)


nrjpoddar (commented):
Thanks for the PR. I think we need to make sure UX is correct here as there are 2-3 fields all related to the same area and we need to understand how they all work together.

I also see that the RFC never got approved.

kfaseela:
> Thanks for the PR. I think we need to make sure UX is correct here as there are 2-3 fields all related to the same area and we need to understand how they all work together.
> 
> I also see that the RFC never got approved.

@nrjpoddar Thanks for the review. Have added points in the RFC to clarify the behavior when auto-sni and sni fields come together. I presented the RFC only in the last working group meeting, so might need some more time to refine all usecases, and get the RFC approved, however the group was positive about the proposal. There was even a point whether we should be doing this without the need for an additional API change, and make this the default behavior. The discussion is still in progress. I thought it would be easier to comment on the code-review, so thought of pushing the changes.

kfaseela:
As per the latest discussions in the networking working group, an API change is not needed for supporting the same. The details are updated in the RFC, and closing this PR. Will be working on the agreed upon approach

PR #2283, opened 2022-03-15
Extend DestinationRule with tunneling settings

jewertow:
Signed-off-by: Jacek Ewertowski <jewertow@redhat.com>

#### Background context
This change aims to enable tunneling outbound traffic. I described the idea in detail in this [RFC](https://docs.google.com/document/d/1QbjsBwM0QyQxJIs2LJBUvITap2QeqOKmsHQrGFK9lWQ).

The API in this pull request differs slightly from that described in the document, but I decided to make it as close to Envoy API as possible. What's more, I didn't mention about field `destination_port` in the RFC, because I forgot that port matching is optional in a VirtualService configuration, so without explicitly defined destination port it might be not possible to configure `TcpProxy.tunneling_config.hostname` in some cases.

This API change is related to this pull request: https://github.com/istio/istio/pull/37968.


networking/v1beta1/destination_rule.proto
```diff
@@ -282,6 +282,18 @@ message TrafficPolicy {
...
+    // Specifies a port to which the downstream connection is tunneled.
+    uint32 destination_port = 1 [(google.api.field_behavior) = REQUIRED];
+
+    // Specifies whether to use CONNECT or POST http method for the upstream tunnel request.
+    // If set to true, then POST is used, otherwise CONNECT.
+    bool use_post = 2;
+  }
+
+  // Configuration for tunneling TCP over HTTP.
+  TunnelingSettings tunneling = 6;
```
hzxuzhonghu, line 296: same question :

>IUUC, this will be used as tunneling_config.hostname
>
> 1. Then what if this is wildcard hosts?
> 2. What's the expected behavior if  dr has subsets

jewertow: 1. That's one of the limitations that I mentioned in the RFC. Tunneling cannot be applied when a hostname contains a wildcard, because it's not possible to determine what hostname should be set in the `tunneling_config.hostname`. I started to work on setting `tunneling_config.hostname` automatically based on SNI, but it might work only for TLS connections. It's also important to note that the same limitation applies to virtual services with multiple hostnames. In such a case, it's also not possible to determine which hostname to choose, so the destination rule must be ignored.
2. Subsets with tunneling settings are handled like any other traffic policy for a subset.

networking/v1alpha3/destination_rule.proto
```diff
@@ -282,6 +282,18 @@ message TrafficPolicy {
...
   // to fields omitted in port-level traffic policies.
   repeated PortTrafficPolicy port_level_settings = 5;
+
+  message TunnelingSettings {
+    // Specifies a port to which the downstream connection is tunneled.
+    uint32 destination_port = 1 [(google.api.field_behavior) = REQUIRED];
+
+    // Specifies whether to use CONNECT or POST http method for the upstream tunnel request.
+    // If set to true, then POST is used, otherwise CONNECT.
+    bool use_post = 2;
```
linsun, line 292: Not a big fan of use_post as a boolean. What is the default?  CONNECT?

I think i prefer a simple method string with CONNECT as the default.

networking/v1alpha3/destination_rule.proto
```diff
@@ -282,6 +282,18 @@ message TrafficPolicy {
...
+  }
+
+  // Configuration for tunneling TCP over HTTP.
```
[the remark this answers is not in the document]
jewertow: Yes, of course.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: Maybe not expose to users

jewertow: > Maybe not expose to users

Since CONNECT is probably the most common use case, it sounds reasonable. We might not expose this setting for now and consider enabling it in the future if necessary.

jewertow: @howardjohn @costinm could you share your opinions?

networking/v1alpha3/destination_rule.proto
```diff
@@ -340,6 +340,18 @@ message TrafficPolicy {
   // overridden by port-level settings, i.e. default values will be applied
   // to fields omitted in port-level traffic policies.
   repeated PortTrafficPolicy port_level_settings = 5;
+
+  message TunnelingSettings {
```
howardjohn, line 344: I definitely agree with the high level goal of improving/enabling tunneling in Istio. However, I want to make sure this aligns with our intentions to use HTTP/2 CONNECT ('bts' or 'hbone') as base line transport protocol in Istio, as well as general plans to improve Egress traffic in Istio. Can you sync up with @lambdai (the first topic) and @costinm (second topic) to make sure these are aligned?

Additionally, it could be useful to compare initiating the tunnel in Envoy vs in the app; most users that are going to use an egress proxy probably already have it when they adopt Istio, so their apps already have HTTP(s)_PROXY configured. AFAIK this is broken with Istio today.

costinm: Yes, there is a doc describing "Better Transport Security" - we have been slowly working toward it, the support for CONNECT in Envoy and other changes are in part based on that design.

Support for POST as a fallback is IMO a strong requirement, but I don't think it should be exposed in the API, but be implemented at discovery level, like auto MTLS.

networking/v1alpha3/destination_rule.proto
costinm: string "protocol" ( but we already have that in the Service, so not clear why would we need it in DR as well ). 

And we should define some protocol names - I think "hbone" can be the normal CONNECT and the default, and we can use hbone-post and other variants.

I think it is pretty clear that tunneling will not be a boolean, but we may need to support multiple protocols and modes. MASQUE and related standards, etc.

costinm (changes requested):
I commented on the istio PR as well - I personally don't think we should add this to DestinationRule at all, in particular in context of the BTS/Hbone proposal to have all in-mesh TCP streams use CONNECT. 

We already have Auto-MTLS as an example - and we know the UX is far easier and better than
before, user not having to manually configure this each time. 

The Service already has protocol - and we use that across Istio to determine how to connect 
(h2, grpc, https, etc), it is IMO much better UX to just use discovery information and auto-configure the protocol used to tunnel.

costinm:
I am also not completely opposed to have something in DestinationRule - if we find some strong use cases where using discovery info is not possible. But I can't think of a use case where we need DR and can't use the Service protocol name.

networking/v1alpha3/destination_rule.proto
jewertow: > but we already have that in the Service, so not clear why would we need it in DR as well

In my opinion it would be unclear how to route traffic to a service which has specified a tunnel protocol. Let's say you define a service entry with protocol `hbone`. Which route type do you use in a virtual service? TCP or TLS? How to determine which one should be matched. I think it's tricky and we can't find an intuitive solution.

> I think "hbone" can be the normal CONNECT and the default, and we can use hbone-post and other variants.

In my opinion "hbone" might be confusing and users might have problems to understand what is it and how does it work, because I found only one [section](https://istio.io/latest/docs/ops/deployment/requirements/) in Istio docs with this term, but it's not explained there. I also searched in Google the following terms: "hbone", "hbone connect", "hbone protocol" and I didn't get any related article.
On the other hand, even if we will document in detail what hbone is, we still can't assume that users will find what they need, because it's a new term and I guess that users will search in Google terms like "Istio connect support" or "Istio tunneling" rather than "Istio hbone".

networking/v1alpha3/destination_rule.proto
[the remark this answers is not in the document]
jewertow: Done

networking/v1alpha3/destination_rule.proto
jewertow: I see another important problem. Using the service protocol to determine tunneling would make it possible to tunnel traffic only directly from a sidecar proxy to the forward proxy. Tunneling traffic through a gateway would be impossible, because it requires to route traffic to a gateway and then to the forward proxy. So there are two steps of routing. With the destination rule we are able to choose on which stage to apply tunneling config. On the other hand, when you only have a service protocol, it's not possible to determine when exactly to enable tunneling.

I was investigating how it might be implemented only with service protocol, but there is more problems to identify and discuss. In my opinion it's much more difficult to implement and much less flexible.

costinm: How about 'h2-connect', 'h2-post' ? BTW, 'hbone' and 'better transport' proposals are inspired from https://datatracker.ietf.org/wg/masque/about/ - we could use 'h2-masque'. 

Why would be impossible to tunnel through a gateway ? The same would work, we would set the protocol on the gateway port that supports CONNECT. Same for traffic to another app with a sidecar.

We already rely a lot on Service port identifying the protocol - grpc, tls, h2, http, https, etc. 

I was thinking about this a bit more - extending DR with a 'protocol' setting that would override the Service is not 
a bad thing, there are cases where the port name/protocol in Service can't be modified ( existing apps installed from helm), or needs to be overridden.

networking/v1alpha3/destination_rule.proto
```diff
@@ -340,6 +340,19 @@ message TrafficPolicy {
   // overridden by port-level settings, i.e. default values will be applied
   // to fields omitted in port-level traffic policies.
   repeated PortTrafficPolicy port_level_settings = 5;
+
+  message TunnelSettings {
```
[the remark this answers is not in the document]
costinm: Also 'protocol' is typically lowercase, and we may need to be more precise - h2-connect, h2-post, http-connect (http-post is not possible since http/1.1 POST is not bi-directional in most implementations).

networking/v1beta1/destination_rule.proto
litong01: @jewertow should there be some validation defined in DR analyzer to ensure all these limitations? Seems to me this can people into trouble quite easily, it will be super nice to reflect the logic/limitation in a specific filter analyzer.

jewertow: I included relevant [validations](https://github.com/istio/istio/pull/37968/files#diff-1488ede8a94d615faf5bbbd17f81fdb16b0409a3cc21ae6525a91991e55df34f) in Istio. Should it be done somehow in the API repository? Could you share an example of such a validation?

networking/v1alpha3/destination_rule.proto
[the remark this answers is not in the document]
jewertow: Used HTTP version depends on the protocol defined in the `ServiceEntry` configured for a given proxy, so "h2-" and "http-" prefixes can't be used to choose the protocol.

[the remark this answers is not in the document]
jewertow: > Can we flatten it out ? I don't know if this is exclusive to 'tunnels'

You're right that these properties might not be exclusive to tunnels, but on the other hand they are used to set `tunneling_config` field in Envoy, so how would you use these properties in any other context? I understand your point and agree that these fields might be used to override protocol and port, but without `tunnel` wrapper how to decide whether it is for `tunneling_config` or not?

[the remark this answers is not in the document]
jewertow: > Can we flatten it out ? I don't know if this is exclusive to 'tunnels'

If you see other potential use cases for these fields, I'm okay with that.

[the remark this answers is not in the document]
jewertow: I was trying to document these fields without the `TunnelingSettings` wrapper, but then it becomes difficult to specify requirements and semantics clearly. Then configuration lacks the context and validation will be tricky. So if you think about reusing these fields for other types of proxies and protocols, what do you think about wrapper `ProxySettings`?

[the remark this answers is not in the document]
jewertow: > Also not sure if 'which HTTP method' is right - we may want in future to extend this to MASQUE ( over QUIC ), or support HAProxy prefix, or tunnel over WS.

But why to document something that is not implemented? It could be document as well once it's implemented.

jewertow:
I considered all suggested concepts, but after investigation all of them, I think that the current version is the best trade-off. Please review it once again. I really have to push this change.

jewertow:
@linsun could you take a look at this PR once again?

costinm (approved):
Looks good.

networking/v1alpha3/destination_rule.proto
```diff
@@ -340,6 +340,22 @@ message TrafficPolicy {
...
+    // Specifies which protocol to use for tunneling the downstream connection.
+    // Supported protocols are:
+    //   connect - uses HTTP CONNECT;
+    //   post - uses HTTP POST.
+    // HTTP version for upstream requests is determined by the service protocol defined for the proxy.
+    string protocol = 1 [(google.api.field_behavior) = REQUIRED];
```
ramaraochavali, line 350: Sorry. I am late to this review. Would it be better if we define this as enum?

jewertow: I don't know downsides of using enum, so if @linsun and @costinm don't mind, I can change it to enum. They suggested to use string type and because of [port protocol](https://github.com/istio/api/blob/master/networking/v1beta1/gateway.proto#L603) is string, I didn't protest to use string as type for tunnel protocol.

ramaraochavali: I am not so strong about enum but it would be better to control it with fixed set of values rather than validations.

costinm:
Please don't use enum ! It is perfectly possible to validate the name - enum create far more problems than they solve ( backward/forward compatibility, etc ).

networking/v1alpha3/destination_rule.proto
costinm: I am very strongly against enums. Protocols are not a fixed set of values in the API - please update the doc maybe to mention 'connect' and 'post' are examples, not an exclusive list of protocols. 

While your use case is only for supporting a traditional HTTP CONNECT proxy as tunnel - this is the API definition, and an API defining that a destination will be tunneled can't be restricted by one particular use case and implementation. There is a long list of tunneling protocols and variants.

hzxuzhonghu, line 350: protocol here is a little confusing

jewertow: I understand your point, but read the discussion above. At the beginning I named this field `method`, because Istio will support tunneling via HTTP proxy using CONNECT or POST methods for now. But in the future Istio might support MASQUE protocol as well, so we decided to name this field `protocol`.

costinm: How about tunnel_protocol ? Agree protocol is very confusing, and it really is tunnel.protocol in the yaml. 

BTW - what is the URL in case of POST ?

linsun: I prefer protocol because it is already in tunnelSettings

linsun: basically user will be configuring:

```
tunnel:
  protocol:
```

so it is pretty clear it is tunnel protocol.

linsun: I think we need clear guidance in the doc on what protocols are supported for tunneling.  We can't leave this open choice by the user.  enum provides a way to do this to ensure user only enters what we support.

jewertow: I don't see significant difference between `tunnel.protocol` and `tunnel.tunnel_protocol`, so I am not sure if it changes anything for a user.

> BTW - what is the URL in case of POST ?

There is no option to specify URL, because on the TcpProxy level we can't perform URL-based routing. TcpProxy knows only a cluster and clusters don't define URLs.

jewertow:
The `destination_port` is needed to enable tunneling TCP traffic to HTTP services. The documentation of VirtualService says about the `TcpRoute`:
```
An ordered list of route rules for opaque TCP traffic. TCP routes will be applied to any port
that is not a HTTP or TLS port. The first rule matching an incoming request is used.
```
So let's say I want to tunnel traffic to an external HTTP service. I create service entries:
```
kind: ServiceEntry
metadata:
  name: example-com
spec:
  hosts:
  - example.com
  location: MESH_EXTERNAL
  ports:
  - number: 80
    name: http
    protocol: HTTP
resolution: DNS
---
kind: ServiceEntry
metadata:
  name: external-proxy
spec:
  hosts:
  - external-proxy.com
  location: MESH_EXTERNAL
  ports:
  - number: 3128
    name: http
    protocol: HTTP
resolution: DNS
```
and routing rules to forward proxy:
```
kind: VirtualService
metadata:
  name: tunnel-traffic-to-example-com
spec:
  hosts:
  - example.com
  gateways:
  - mesh
  tcp:
  - match:
    - gateways:
      - mesh
      port: 80
    route:
    - destination:
        host: external-proxy.com
        port:
          number: 3128
```
Now you can assume that the service entry for example.com has defined port 80, so I don't need to define this port once again in the destination rule, but in fact the sidecar proxy will not have configured TcpProxy listening on port 80, because on port 80 will be configured "Route: 80", which is an HttpConnectionManager that redirects traffic to virtual hosts "allow_any" or "block_all" (depending on the Istiod settings). So conceptually the port might be determined from service entry, but in this particular case it will not work and the workaround with `destination_port` is necessary.

networking/v1alpha3/destination_rule.proto
costinm: I noticed envoy also includes a "hostname" option ( which is sent as the Host header when making this request).

Should we add it as well ?

jewertow: I assumed that the hostname is determined from the `hosts` list specified in a virtual service and this is the reason why a destination rule with tunnel settings may only be applied to a virtual service that has only one host without wildcard.

This is very good question, because having `hostname` or `targetHost` field in tunnel settings we could avoid the limitation for `hosts` in virtual service. Since the target port must be defined, maybe it's a good idea to define also the hostname explicitly. Then the destination rule would be:
```
kind: DestinationRule
metadata:
  name: tunnel-traffic-to-example-com-via-external-forward-proxy
spec:
  host: external-forward-proxy.com
  subsets:
  - name: example-com-80
    trafficPolicy:
      tunnel:
        protocol: connect
        targetHost: example.com
        targetPort: 80
```
I didn't consider this approach, but it seems to be simpler. What do you think?

@linsun could you also share your opinion?

costinm:
I think I understand your use case now, and it is a bit more complicated - maybe too complicated. 

What you actually need is for HTTP traffic to be sent to the proxy. The 'CONNECT' stuff is implemented for TCP proxies. 
In your case, example.com:80 is actually HTTP - we want it to be treated as HTTP, so metrics, rules, etc are applied as HTTP.
We certainly don't want sidecar to configure a TCP listener on port 80, or the API to be driven by this - just because Envoy can't 
use CONNECT for this. For HTTP, using CONNECT or POST are both wrong anyways - HTTP_PROXY is a different protocol, 
using an absolute URL ( GET http://example.com:80 HTTP/1.1 ).

I think the CONNECT support should be limited to HTTPS/TCP - where the app originates plain-text HTTP or http2 we should still have a http connection manager. 

On the proxy side, I doubt any standard-compliant proxy will know how to deal with 'HTTP in CONNECT' - but most of them support 'transparent proxy' mode, so if we just forward the HTTP request to squid, it should be handled correctly, without losing L7 semantics.  I think ideally Envoy would add proper support for HTTP_PROXY on client side - it already has 'allow_absolute_url' on the listener side, so Envoy can act as a non-transparent proxy. 

I would add that virtually all client libraries now use https - besides our examples I don't think it is very common to have code making calls to http://example.com as plain text - just common on the intranet where network security may be used. 

Treating HTTP as TCP is technically possible - if the example-com ServiceEntry is declared as TCP instead of HTTP, and a virtual IP is added (plus DNS interception I think) - Istio will match IP:80 first, and treat it as TCP. But that should almost never be done, and certainly not an example or recommended in the docs, it would violate both the RFC and Istio metrics/policy/etc. And I suspect it would result in a plain-text stream from the egress gateway to example.com, which is the worse ( a CONNECT proxy will not encrypt the stream ).

linsun:
> The `destination_port` is needed to enable tunneling TCP traffic to HTTP services. The documentation of VirtualService says about the `TcpRoute`:
> 
> ```
> An ordered list of route rules for opaque TCP traffic. TCP routes will be applied to any port
> that is not a HTTP or TLS port. The first rule matching an incoming request is used.
> ```
> 
> So let's say I want to tunnel traffic to an external HTTP service. I create service entries:
> 
> ```
> kind: ServiceEntry
> metadata:
>   name: example-com
> spec:
>   hosts:
>   - example.com
>   location: MESH_EXTERNAL
>   ports:
>   - number: 80
>     name: http
>     protocol: HTTP
> resolution: DNS
> ---
> kind: ServiceEntry
> metadata:
>   name: external-proxy
> spec:
>   hosts:
>   - external-proxy.com
>   location: MESH_EXTERNAL
>   ports:
>   - number: 3128
>     name: http
>     protocol: HTTP
> resolution: DNS
> ```
> 
> and routing rules to forward proxy:
> 
> ```
> kind: VirtualService
> metadata:
>   name: tunnel-traffic-to-example-com
> spec:
>   hosts:
>   - example.com
>   gateways:
>   - mesh
>   tcp:
>   - match:
>     - gateways:
>       - mesh
>       port: 80
>     route:
>     - destination:
>         host: external-proxy.com
>         port:
>           number: 3128
> ```
> 
> Now you can assume that the service entry for example.com has defined port 80, so I don't need to define this port once again in the destination rule, but in fact the sidecar proxy will not have configured TcpProxy listening on port 80, because on port 80 will be configured "Route: 80", which is an HttpConnectionManager that redirects traffic to virtual hosts "allow_any" or "block_all" (depending on the Istiod settings). So conceptually the port might be determined from service entry, but in this particular case it will not work and the workaround with `destination_port` is necessary. I know that it's a single case, but unfortunately it's probably the most common case for not encrypted traffic and we can't assume that this particular case will not be handled and users can't tunnel traffic to port 80, because in case of external services, users may have no control of ports configuration.

Agreed on the value of the targetPort. Just to make sure I understand - are you saying the above 2 SEs, 1 VS to be used in conjunction with the 1 DR (https://github.com/istio/api/pull/2283#discussion_r844973379)?

networking/v1alpha3/destination_rule.proto
jewertow: There is no obstacle to extend tunneling setting to support headers in the future. I want to support at least the simplest case for now - I mean tunneling without appending headers.

> I think targetHost is important for TCP capture cases because otherwise TcpProxy wouldn't know what hostname to put in the request - Envoy has a listener on a IP:port, and I believe will do a CONNECT with that IP:port (?).

As I said earlier, I assumed that the implementation will determine hostname from VirtualService hosts. You can look at the implementation [here](https://github.com/istio/istio/pull/37968/files#diff-b68d6b51c18d25e1ac92021387695f2e06f71321d262f0bd7634c227e9f581c9R39). But it implies the limitation for `hosts` in a virtual service, so I suggested to explicitly define `targetHost` instead of setting it automatically.

> So maybe targetHost will be used in the implementation, and set to external-forward-proxy.com ?

No, the targetHost in this case is example.com.

howardjohn:
I would like to review this but am OOO this week. will give it a look
Monday

jewertow:
I added field `target_host`. Now this API is a mirror of Envoy API and everything must be specified explicitly, but it also eliminates limitations for virtual service. It means that a virtual service may contain multiple hosts and these hosts may contain wildcards.

linsun:
For the record, I don't have objection with the PR.  Since @howardjohn wants to review on Monday, i'll defer to him to stamp the approval.  I'll be on vacation next week.

networking/v1alpha3/destination_rule.proto
costinm: We don't have an enum for any other protocol field, or in Service.

We can have docs that list protocols that are supported - also a reminder that Istio also support proxyless gRPC and in theory it is not tied to Envoy or a single implementation.

costinm (commented):
I think we need to figure out what to do about HTTP - because that's clearly going to be broken.

We can't treat HTTP as TCP.

hzxuzhonghu:
Yes, envoy only supports tunnel config under tcp proxy now.

hzxuzhonghu:
cc @lambdai is an expert on this aspect.

costinm:
Right - and CONNECT is only meant for TCP, plain text HTTP is not supposed to use CONNECT, the proxies will get confused and send it as plain text. 

We need to document ( and ensure ) that tunnel DR is only used for TCP ( including HTTPS/TLS), at least until envoy has support for proper HTTP_PROXY.

jewertow:
> I think we need to figure out what to do about HTTP - because that's clearly going to be broken.

Why? What is going to be broken?

> We can't treat HTTP as TCP.

Why? HTTP is built on top of TCP, so what's wrong with it? If you suggest to not support HTTP, because it's not encrypted, we should not support any other non-TLS traffic, but it doesn't make sense, because for plain traffic users can apply TLS origination.

> Right - and CONNECT is only meant for TCP, plain text HTTP is not supposed to use CONNECT, the proxies will get confused and send it as plain text.

HTTP is TCP as well, so CONNECT proxy don't care what is this traffic. It accepts HTTP CONNECT, establishes connection and send bytes back and forth. Proxies will not get confused, because it's responsibility is only to forward traffic to a target socket. The whole point of such proxies is to not understand the traffic. So there is no difference in tunneling HTTP or HTTPS.

costinm:
On Fri, Apr 8, 2022 at 4:11 AM Jacek Ewertowski ***@***.***>
wrote:

> I think we need to figure out what to do about HTTP - because that's
> clearly going to be broken.
>
> Why? What is going to be broken?
>
The protocol ? The RFC is pretty clear about how HTTP proxy is supposed to
work, and it isn't CONNECT.

Plus telemetry, any RBAC rules that expect http attributes, routing, etc.



> We can't treat HTTP as TCP.
>
> Why? HTTP is built on top of TCP, so what's wrong with it? If you suggest
> to not support HTTP, because it's not encrypted, we should not support any
> other non-TLS traffic, but it doesn't make sense, because for plain traffic
> users can apply TLS origination.
>

I meant: our typical example is upgrading HTTP to HTTPS via egress. User
originates HTTP, egress gateway upgrade to HTTPS after applying various
RBAC rules
or routing. If HTTP is treated as TCP - all this is gone.

Besides - if the goal is to support 'standard proxies' like SQUID - you
must follow the RFC, and that is pretty clear about how http proxy is
handled.
I also doubt many properly configured proxies will allow CONNECT on port
80, usually they are used for some policy enforcement.


> Right - and CONNECT is only meant for TCP, plain text HTTP is not supposed
> to use CONNECT, the proxies will get confused and send it as plain text.
>
> HTTP is TCP as well, so CONNECT proxy don't care what is this traffic. It
> accepts HTTP CONNECT, establishes connection and send bytes back and forth.
> Proxies will not get confused, because it's responsibility is only to
> forward traffic to a target socket. The whole point of such proxies is to
> not understand the traffic. So there is no difference in tunneling HTTP or
> HTTPS
>

There is a big difference - https is following the RFC semantics.

Costin

jewertow:
It seems that I didn't understand what you mean by treating HTTP as TCP and we talk about different things.
I didn't wanted to say that HTTP proxy should be treated as TCP. I was thinking that you mean that outbound HTTP traffic can't be treat as TCP. HTTP proxy obviously shouldn't be treated as TCP service. To avoid misunderstanding please take a look at this pull request: https://github.com/istio/istio/pull/37968. There are examples in the directory tests/integration/pilot/tunneling.

costinm:
The examples seem to configure port 8080 as TCP - so a tcp proxy is
created. That's wrong IMO, it means http telemetry and any RBAC in the
egress proxy
that uses http attributes will be gone. It may pass the test, since it's
not including this.

Neither in-cluster nor outbound HTTP should be treated as TCP, and I've
seen many proxies that reject CONNECT traffic that doesn't look encrypted
or even if dest is not 443 (or some small set of ports), in some cases
doing MITM to apply further policies.

networking/v1alpha3/destination_rule.proto
```diff
@@ -340,6 +340,26 @@ message TrafficPolicy {
   // overridden by port-level settings, i.e. default values will be applied
   // to fields omitted in port-level traffic policies.
   repeated PortTrafficPolicy port_level_settings = 5;
+
```
howardjohn, line 343: I mentioned this in the WG meeting but will reiterate it here. I think it's critical we have a strategy planned out for how this relates to user initiated CONNECT; not suggesting implementing that is a blocker but i would like to see some discussion of the long term plans around that so we can guide users, or docs, and our future plans.

jewertow:
> The examples seem to configure port 8080 as TCP - so a tcp proxy is
created. That's wrong IMO, it means http telemetry and any RBAC in the
egress proxy
that uses http attributes will be gone. It may pass the test, since it's
not including this.

Yes, but for TLS traffic you also can't apply HTTP telemetry, so it's not a matter of whether the traffic is encrypted or not.

On the other hand why do we have to assume that CONNECT proxies must reject not encrypted traffic? It's not the problem of Istio. I don't care how users configure external components.

> so a tcp proxy is created. That's wrong IMO

It's not wrong. It's absolutely intentional configuration, because only TcpProxy enables tunneling, so HttpConnectionManager can't be used.

ramaraochavali:
For http, does n't some thing similar will help https://github.com/envoyproxy/envoy/blob/8cb6862fe6099cd8583a64ff037ecdeaf0e939fa/configs/proxy_connect.yaml#L36?

networking/v1alpha3/destination_rule.proto
nrjpoddar: Tunneling protocols are much more bounded than application protocols so it made sense to me keep service protocol as string so that we don't change our APIs often. If we expect new tunneling protocols to be supported by Envoy very frequently then I can understand keeping this as string else I would prefer a ENUM here.

nrjpoddar (commented):
Adding this to DR has few confusing semantics i.e. can you specify all other settings in the TrafficPollcy if you want to use the Tunnel protocol? Settings like TLS, connection level settings and load balancer settings?

Since the tunneling happens over HTTP the VS associated with this should be an HTTP block and the service port prefix of the upstream should be http*, is that correct?

costinm:
There is no such thing as "CONNECT tunneling for HTTP" - the standard that
defines CONNECT also defines how to proxy HTTP, using
the absolute URL - you can't change the verb from POST/GET/etc to CONNECT
or use CONNECT and then send the HTTP request as the body.

We need to stop inventing weird protocols and follow the standards.

jewertow:
> We need to stop inventing weird protocols and follow the standards.

I don't know what's weird for you. It's as simple as possible. If you don't agree to support plain TCP traffic, let's support at least TLS traffic.

The main purpose of this PR is to provide API which enable users to eliminate configuring HTTP proxies in their apps by setting envs like `HTTP_PROXY` or properties like `java -Dhttps.proxyHost=host -Dhttps.proxyPort=port`.
If "tunneling API" is too broad topic and you have other plans that conflict with this use case and submitted API, what do you think to change it to `HttpsProxy` and keep it simple and support only TLS?

costinm:
I do agree with supporting TLS and TCP traffic, via TcpProxy.
App-originated HTTPS too.

I am also ok with supporting HTTP proxies - equivalent with setting
HTTP_PROXY and HTTPS_PROXY.

But setting HTTP_PROXY is NOT using CONNECT, and all HTTP proxies implement
the RFC, which is to use
the absolute URL. The configs you have - treating HTTP as a TCP connection
- is not the same thing with setting HTTP_PROXY and
is not supported/used by http proxies, and is changing the semantics of
Istio, that's what I don't like.

jewertow:
> But setting HTTP_PROXY is NOT using CONNECT, and all HTTP proxies implement the RFC

I tested multiple commonly used HTTP tools and libraries (Java, NodeJS, curl...) and all of them use HTTP CONNECT when HTTP proxy is configured, so I don't understand why you still say that it's weird protocol and why you suggest that proxies does not support it. I also inspected the traffic with Wireshark and it works as expected.

Could you share the RFC which you are referring to?
I read the following RFCs:
- HTTP CONNECT: https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6
- Tunneling TCP based protocols through Web proxy servers: https://datatracker.ietf.org/doc/html/draft-luotonen-web-proxy-tunneling-01

and I can't see there any information about absolute URL which you talk about. Both these RFCs says that tunnel should be established with the following request
```
     CONNECT server.example.com:80 HTTP/1.1
     Host: server.example.com:80
```
and it makes sense, because the absolute URL that you want connect to will be sent once the connection is established.

costinm:
Quick search:
https://stackoverflow.com/questions/7577917/how-does-a-http-proxy-utilize-the-http-protocol-a-proxy-rfc

Can you share wireshark trace for any of those languages - with HTTP_PROXY
set  ( for a http:// request, not https://  ) ?
The go implementation is net/http/request.go is

if usingProxy && r.URL.Scheme != "" && r.URL.Opaque == "" {
 ruri = r.URL.Scheme + "://" + host + ruri
....

and transport.go

case cm.targetScheme == "http":
pconn.isProxy = true
if pa := cm.proxyAuth(); pa != "" {
pconn.mutateHeaderFunc = func(h Header) {
h.Set("Proxy-Authorization", pa)
}
}
case cm.targetScheme == "https":
conn := pconn.conn
var hdr Header
if t.GetProxyConnectHeader != nil {
var err error
hdr, err = t.GetProxyConnectHeader(ctx, cm.proxyURL, cm.targetAddr)
if err != nil {
conn.Close()
return nil, err
}

I can probably find the java implementation as well.

jewertow:
Ok, now I understand what you mean and you're right that setting HTTP or HTTPS proxies in most clients don't use HTTP CONNECT for HTTP connection initiated by a client, but for TLS connections proxies use HTTP CONNECT.
I misunderstood your comment and I thought you mean that setting HTTP/HTTPS proxies in clients does not use CONNECT at all.

I think we look at this feature from completely different perspectives.
You are focused on that this is supposed to be a typical web/HTTP proxy (in the RFC called "proxy"). In that case your thoughts about forwarding HTTP would be absolutely correct and reasonable.
But this PR aims to enable integration with tunnel proxies (in the RFC called "tunnel").
Look at section "2.3. Intermediaries":
```
   A "tunnel" acts as a blind relay between two connections without
   changing the messages.  Once active, a tunnel is not considered a
   party to the HTTP communication, though the tunnel might have been
   initiated by an HTTP request.
```
and then at section "2.6. Protocol Versioning":
```
   Intermediaries that process HTTP messages (i.e., all intermediaries
   other than those acting as tunnels) MUST send their own HTTP-version
   in forwarded messages.
```
As you can see there is no information how tunnel proxies have to treat HTTP requests, because their purpose is to be **blind**, to **not process** received messages and just **establish** TCP connection between a client and an origin server.
So I think that this PR is compatible with the RFC you are referring to.

As far as this API is named `TunnelSettings` there should be no confusion why the connection is not initiated as typical HTTP clients which support `HTTP_PROXY` (or similar) settings.
I would like to cancel the previous idea to name it `HttpProxy` or `HttpsProxy` - it was not well thought out.

networking/v1alpha3/destination_rule.proto
jewertow: We (at RedHat) have no other plans related to tunneling or initiating CONNECT requests. I proposed the simplest API I can see. I know that you had other ideas similar to auto-mtls or service protocol, but both have flaws which I explained above. As far as Envoy is not going to provide nor extend its tunneling capabilities, I can't see alternative solutions.

It's important to note that this API does not block you to provide other approach for this problem in the future - as is in case of mTLS which can be applied both with DestinationRule and PeerAuthentication.

costinm:
I am not actually opposed to supporting the CONNECT tunnel for HTTP
connections ( in particular for HTTP/2 plain text), if:
- we find some non-Istio proxies that support this and don't get confused
- we implement it in a way that doesn't break Istio own APIs.

Maybe I was not clear in my comments - the later point is very important
and what started the discussion, in your example the HTTP port is labeled
as TCP to
force the use of the TCP proxy which supports CONNECT. That breaks
telemetry and policies operating on HTTP attributes.

We could either make changes in Envoy, or generate a config that still
treats the request as HTTP, applies all policies/telemetry, including mTLS
- but does an internal forward to
 a local Tcp cluster that does the upgrade to CONNECT. This is IMO the
'right' use of CONNECT, for tunneling encrypted traffic and preserving mTLS.

jewertow:
> we find some non-Istio proxies that support this and don't get confused

Do you mean to ensure which proxies would it work with? I was testing it mostly with Envoy and there are no restrictions. I was also using squid and nginx with proxy_connect module. I could also test it with apache http.

It's definitely possible to support HTTP port without changing anything in Envoy, but it's much more complex solution. I was testing redirecting traffic from HttpConnectionManager to a TcpProxy listening on unix domain socket and it worked fine.

So what do you think to start with support only for TLS traffic and then investigate support for plain TCP and HTTP in another pull request? Tunneling TLS is the most important case and I don't want to block it with discussions about plain HTTP.

costinm:
On Fri, Apr 15, 2022 at 7:54 AM Jacek Ewertowski ***@***.***>
wrote:

> we find some non-Istio proxies that support this and don't get confused
>
> Do you mean to ensure which proxies would it work with? I was testing it
> mostly with Envoy and there are no restrictions. I was also using squid and
> nginx with proxy_connect module. I could also test it with apache http.
>

I'm sure all will "work" - treating it as a TCP stream. What I mean is that
proxy may have http-level settings/policies - like allow some domains or
block others, or keep track of the domains. It's the main reason proxies
were required. I don't know how much those days with https.

The fundamental problem is treating HTTP as TCP - both in Istio and in the
proxy.


> It's definitely possible to support HTTP port without changing anything in
> Envoy, but it's much more complex solution. I was testing redirecting
> traffic from HttpConnectionManager to a TcpProxy listening on unix domain
> socket and it worked fine.
>

Yes, it is more complex, but I think it is necessary. There are some
changes to use an internal redirection and further optimize.


> So what do you think to start with support only for TLS traffic and then
> investigate support for plain TCP and HTTP in another pull request?
> Tunneling TLS is the most important case and I don't want to block it with
> discussions about plain HTTP.
>

I agree. Let's just document in the API that it will not apply to HTTP or
H2 for now - and not include any of the examples treating http as tcp.

networking/v1alpha3/destination_rule.proto
```diff
@@ -340,6 +340,27 @@ message TrafficPolicy {
...
+    // Target host must be an FQDN.
+    string target_host = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // Specifies a port to which the downstream connection is tunneled.
+    uint32 target_port = 3 [(google.api.field_behavior) = REQUIRED];
+  }
+
+  // Configuration of tunneling TCP over other transport or application layers
+  // for the host configured in the DestinationRule.
+  // Tunnel settings can be applied to TCP or TLS routes and can't be applied to HTTP routes.
```
jewertow, line 362: I will remove this note once I submit support for HTTP routes.

networking/v1alpha3/destination_rule.proto
```diff
@@ -340,6 +340,27 @@ message TrafficPolicy {
...
+  message TunnelSettings {
+    // Specifies which protocol to use for tunneling the downstream connection.
+    // Supported protocols are:
+    //   connect - uses HTTP CONNECT;
+    //   post - uses HTTP POST.
+    // HTTP version for upstream requests is determined by the service protocol defined for the proxy.
+    string protocol = 1 [(google.api.field_behavior) = REQUIRED];
+
+    // Specifies a host to which the downstream connection is tunneled.
+    // Target host must be an FQDN.
```
costinm, line 353: Why not IP ? HTTPS_PROXY does allow this, and it is common to not have a DNS name for proxies (many times
they are used over VPN or hosts where the DNS resolver is hard to change, since setting the proxy can be done by non-root users).

jewertow: That's true, thanks. I just forgot to mention about it. I added this information and pushed new commit.

networking/v1alpha3/destination_rule.proto
```diff
@@ -340,6 +340,27 @@ message TrafficPolicy {
...
+    string target_host = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // Specifies a port to which the downstream connection is tunneled.
+    uint32 target_port = 3 [(google.api.field_behavior) = REQUIRED];
```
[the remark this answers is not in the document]
jewertow: To be honest, it completely doesn't matter for me. So I would like to know the opinion of API maintainers.
What do you think @linsun?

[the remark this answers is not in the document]
linsun: I think targetHost and targetPort is pretty simple for user to use, i could be convinced to use but also felt targetHost and targetPort offers more clarity.
```
target: {host:port}
```

linsun (approved):
LGTM

Thank you @jewertow for your hard work on this!

PR #2286, opened 2022-03-16
operator: use camelCase

howardjohn:
operator relies on merging using K8s libraries. K8s libraries use the
`json` tag. protobuf doesn't have a way to configure this (`json_tag` is
for `protobuf.json`, not `json`) and never will.

JSON tag comes from the protobuf name. So we can make it camelCase to
fix the above issue. This will have no impact on unmarshalling with jsonpb -- which we use in
all cases aside from the merging -- any casing is allowed.

This may be a breaking change in the protobuf sense (but NOT to Istio
users), but it doesn't really matter as until last week the operator
protobuf was not a real protobuf at all and relied on `sed` to make it
into the shape we want. Therefor, its not possible someone depended on
the operator protobuf being protobuf-compatible.


richardwxn (approved):
LGTM as long as it does not affect the unmarshalling behavior.

would you mind adding some comments to the operator.proto explaining why we need to use camelCase? just to make sure other maintainers are aware of this uncommon pattern

ericvn (commented):
LGTM. Will approve after comments added to operator.proto per prior review.

howardjohn:
added comment

operator/v1alpha1/operator.proto
```diff
@@ -72,7 +76,7 @@ message IstioOperatorSpec {
     //
     // * /tmp/istio-installer/nightly (local file path)
     //
-    string install_package_path = 11;
```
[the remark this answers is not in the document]
howardjohn: Only in operator do we try to use K8s strategic merge.

Previously, we also use camel case here but we did it with `sed` which makes it not valid proto

Makefile.core.mk
```diff
@@ -108,10 +108,12 @@ release-lock-status:
 # lint-protos is different for istio/api. List all other lint-all targets and add local-lint-protos
 local-lint-protos:
 	@buf lint
+	@./scripts/check-operator-proto.sh
 
 lint: lint-dockerfiles lint-scripts lint-yaml lint-helm lint-copyright-banner lint-go lint-python lint-markdown lint-sass lint-typescript lint-licenses local-lint-protos
  	@$(htmlproofer) . --url-swap "istio.io:preliminary.istio.io" --assume-extension --check-html --check-external-hash --check-opengraph --timeframe 2d --storage-dir $(repo_dir)/.htmlproofer --url-ignore "/localhost/"
 
+
```
ericvn, line 116: nit: extra line.

operator/v1alpha1/operator.proto
```diff
@@ -21,6 +21,10 @@ syntax = "proto3";
 // $mode: package
 
 // Configuration affecting Istio control plane installation version and shape.
+// Note: unlike other Istio protos, field names must use camelCase. This is asserted in tests.
+// Without camelCase, the `json` tag on the Go struct will not match user's the JSON representation.
```
ericvn, line 25: nit: match the user's

scripts/check-operator-proto.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# Copyright 2019 Istio Authors
```
ericvn, line 3: nit: don't need the year.

ericvn (approved):

PR #2293, opened 2022-03-22
Migrate from gogo/protobuf to golang/protobuf

howardjohn:
This PR mirrors https://github.com/istio/istio/pull/38055; see that PR for more details. This closes a 2+ year T&R roadmap item.

This PR is split up into 2 commits - the first represents changes I made by hand, and the latter running `make gen`. I strongly recommend just looking at the first commit


networking/v1alpha3/envoy_filter.pb.go
```diff
@@ -1198,506 +1195,265 @@ type EnvoyFilter_ListenerMatch struct {
...
-var xxx_messageInfo_EnvoyFilter_Patch proto.InternalMessageInfo
-
-func (m *EnvoyFilter_Patch) GetOperation() EnvoyFilter_Patch_Operation {
-	if m != nil {
-		return m.Operation
-	}
-	return EnvoyFilter_Patch_INVALID
-}
-
-func (m *EnvoyFilter_Patch) GetValue() *types.Struct {
```
zirain, line 1559: this will impact users import istio.io/api or istio.io/istio, when they are trying to upgrade the import packages.

howardjohn: Yes, this is correct. Users that import istio/api will need to make some minor changes to their application. These are mostly just switching imports and minor syntax changes.

Unfortunately there is no alternative. The gogo project is dead for multiple years and continues to drift from the mainline protobuf, with a number of critical bugs and security issues unaddressed. It's use also brings substantial complexity to Istio as we need to mix 2 incompatible libraries together and maintain forks for a number of protos.

Almost all major projects are moving in this direction

zirain: I agree with this's something we must do.
Do we need post an blog like [this](https://istio.io/latest/blog/2021/upcoming-networking-changes/). before we merged this?

howardjohn: I don't think so, just a release note. That issue would *subtly* break users when they upgraded. In this case they will run `go get` and the build will immediately fail to compile; it will be very obvious why it broke and there is nothing a user can do in advanced to prepare for it, so I don't think a blog is needee

sdake:
really nice work @howardjohn. 👍

jacob-delgado (approved):

ericvn (approved):
LGTM. Thanks for doing this.

PR #2305, opened 2022-03-29
move node selector annotation to api repo

xichengliudui:
ref: https://github.com/istio/istio/pull/38167

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -306,7 +306,7 @@ annotations:
     deprecated: false
     hidden: false
     resources:
-      - Node
+      - Service
 
   - name: traffic.sidecar.istio.io/includeOutboundIPRanges
     featureStatus: Alpha
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -302,7 +302,10 @@ annotations:
 
   - name: traffic.istio.io/nodeSelector
     featureStatus: Stable
-    description:  the value for this annotation is a set of key value pairs (node labels) that can be used to select a subset of nodes from the pool of k8s nodes It is used for multi-cluster scenario, and with nodePort type gateway service.
+    description: This annotation is a set of node-labels (key1=value,key2=value). If the
+      annotated Service is of type NodePort and is a multi-network gateway (see
+      topology.istio.io/network), the addresses for selected nodes will be used for
+      cross-network communication.
     deprecated: false
     hidden: false
     resources:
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -300,6 +300,14 @@ annotations:
     resources:
       - Pod
 
+  - name: traffic.istio.io/nodeSelector
+    featureStatus: Stable
+    description:  the value for this annotation is a set of key value pairs (node labels) that can be used to select a subset of nodes from the pool of k8s nodes It is used for multi-cluster scenario, and with nodePort type gateway service.
+    deprecated: false
+    hidden: false
+    resources:
+      - Node
```
hzxuzhonghu, line 309: Not right, it is not applied to node

xichengliudui: Why is `nodeSelector` not a node? what should it be?

hzxuzhonghu: It is on a nodePort east-west gw svc.

xichengliudui: done

xichengliudui:
cc @stevenctl

xichengliudui:
cc @stevenctl

annotation/annotations.yaml
```diff
@@ -300,6 +300,14 @@ annotations:
     resources:
       - Pod
 
+  - name: traffic.istio.io/nodeSelector
+    featureStatus: Stable
+    description:  the value for this annotation is a set of key value pairs (node labels) that can be used to select a subset of nodes from the pool of k8s nodes It is used for multi-cluster scenario, and with nodePort type gateway service.
```
hzxuzhonghu, line 305: The desc is not clear to me

stevenctl: Is this clearer: 

This annotation is a set of node-labels (ex: `key1=value,key2=value`). If the annotated Service is of type `NodePort` and is a multi-network gateway (see `topology.istio.io/network`), the addresses for selected nodes will be used for cross-network communication.

hzxuzhonghu: Looks good

stevenctl (approved):

hzxuzhonghu (approved):

xichengliudui:
@hzxuzhonghu 
It seems that a release-notes-none is needed.

zirain:
> @hzxuzhonghu It seems that a release-notes-none is needed.

take it easy, reviewer will add `release-note-none` if it's needed.

ericvn:
When a PR was opened, the creator can check a box that there are no user visible updates. If the box was checked at creation time, then the `release-notes-none` tag is automatically added. If the box isn't checked, it implies that there are user visible updates and that release notes are required so no label is added. Reviewers can update the label as well.

PR #2306, opened 2022-03-29
Update the package where the variable resides

xichengliudui:
`IstioGatewayPortLabel` should be in the label package
I will update istio if merge

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -266,18 +266,6 @@ var (
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


howardjohn (commented):
cc @stevenctl

stevenctl (approved):

ericvn:
So the istio/istio repo is failing to merge this as noted above: https://prow.istio.io/view/gs/istio-prow/logs/update_api_dep_istio_api_postsubmit/1516471354218516480

PR #2312, opened 2022-04-05
add support for configuring internal addresses

ramaraochavali:
Envoy now supports configuring cidrs that should be treated as internal addresses https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L209. 

This affects how Envoy treats incoming connections as internal or external. In some of envs, RFC 1918 is not followed for internal addresses. 

This allows a mesh to configure internal addresses cidrs


mesh/v1alpha1/config.proto
```diff
@@ -1088,8 +1088,12 @@ message MeshConfig {
   // Configuration of mTLS for traffic between workloads within the mesh.
   TLSConfig mesh_mTLS = 63;
 
+  // List of CIDR ranges that are treated as internal. If unset, then RFC1918 / RFC4193
+  // IP addresses will be considered internal.
+  repeated string internal_addresses = 64;
```
costinm, line 1093: But what happens if it is set ? Are those extras ?

costinm (commented):
I was also looking at something like this - not sure what your use case is, but I am trying to simplify the configuration of 'egress gateway' by redirecting all non-internal addresses to the gateway. Still working on a RFC..  

What I was thinking was to actually extend the existing 'networks'. In other words, a user may have a multi-network mesh, with each network having a set of 'internal cidr ranges' - and we already have a config to define that via MeshNetworks.fromCidr.

The equivalent of what you propose is using 
default:
 - fromCidr: ....

mesh/v1alpha1/config.proto
ramaraochavali: If this is set, the cidr range specified here would be considered as internal address instead of default one

costinm:
On Tue, Apr 5, 2022 at 9:13 AM Rama Chavali ***@***.***>
wrote:

>
> <https://github.com/istio/api/pull/2312#discussion_r843000324>:
>
> > @@ -1088,8 +1088,12 @@ message MeshConfig {
>    // Configuration of mTLS for traffic between workloads within the mesh.
>    TLSConfig mesh_mTLS = 63;
>
> +  // List of CIDR ranges that are treated as internal. If unset, then RFC1918 / RFC4193
> +  // IP addresses will be considered internal.
> +  repeated string internal_addresses = 64;
>
> If this is set, the cidr range specified here would be considered as
> internal address instead of default one
>

Maybe add this to the doc, and list the IP4 and IP6 ranges that are set as
default ( so if the user wants to _add_ an extra network they don't
miss some of the ranges in RFC1918/...)

hzxuzhonghu:
Can you elaborate how internal address is treated differently

ramaraochavali:
Envoy sets "[x-envoy-internal](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers.html?highlight=internal%20addresses#x-envoy-internal)" header and services can use it to decide whether the call is from internal mesh/public. That is a common usecase.

ramaraochavali:
@costinm That makes sense. I updated the comments.

hzxuzhonghu:
>@hzxuzhonghu Envoy sets "[x-envoy-internal](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers.html?highlight=internal%20addresses#x-envoy-internal)" header and services can use it to decide whether the call is from internal mesh/public. That is a common usecase.

Why could not infer this from other headers?

ramaraochavali:
> Why could not infer this from other headers?

Which header?

hzxuzhonghu:
x-envoy-xxx

ramaraochavali:
> x-envoy-xxx 

which header? x-envoy-internal is the one which tells whether the address is internal or external and it is derived from xff (which is again based on this value)

hzxuzhonghu:
Oh, I misunderstood a little about the background.  Since for istio internal mesh sidecar-sidecar communication, no xff is appended. only  external-call --> gateway--> server, xff is generated. By looking at this header, could know if it is external

ramaraochavali:
Yes.

ramaraochavali:
@costinm @hzxuzhonghu Can you please approve if you don't have any further questions?

mesh/v1alpha1/config.proto
zirain: do we need set all clusters' pod cidrs here for best practice?

costinm: Not for 'best practice' - it won't work otherwise... But I think if you add the ranges above it'll cover all the mesh.

costinm (commented):
few remaining issues:
- please add the IP6 ranges as well to the doc - I bet users will forget to add them
- I think we need more discussion if we want to add this here, or use the existing 'networks', which define the ranges of pods in each internal network.

In a multi-network case the user shouldn't have to duplicate the ranges both here and in networks. The networks also has the benefit that it is also associating the ranges with network names, so I would prefer that - we can get the same behavior you have here by setting the fromRanges on the  'default' network with no API changes.

mesh/v1alpha1/config.proto
ramaraochavali: > do we need set all clusters' pod cidrs(in a multi clusters mesh) here for best practice
No.

Yes. If we add the ranges above, it covers the entire mesh

ramaraochavali:
> I think we need more discussion if we want to add this here, or use the existing 'networks', which define the ranges of pods in each internal network.

If we go down this path - we need to change https://github.com/istio/api/blob/master/mesh/v1alpha1/network.proto#L53 to repeated. If you are OK with that, I can change that as well

mesh/v1alpha1/config.proto
```diff
@@ -1088,8 +1088,18 @@ message MeshConfig {
   // Configuration of mTLS for traffic between workloads within the mesh.
   TLSConfig mesh_mTLS = 63;
 
+  // List of CIDR ranges that are treated as internal. If unset, then RFC1918 / RFC4193
```
howardjohn, line 1091: can you explain in the comments what the implications of "internal" are?

mesh/v1alpha1/config.proto
```diff
@@ -1088,8 +1088,18 @@ message MeshConfig {
...
+  // IP addresses will be considered internal.
+  // RFC 1918 defines the following address ranges as private,
+  // 10.0. 0.0/8 (addresses 10.0. 0.0 through 10.255. 255.255 inclusive)
```
howardjohn, line 1094: nit: weird spacing in the IPs?

costinm:
BTW - if you do make changes to the Networks proto, I would strongly
recommend

```
 from:
    - ipBlock:
        cidr: 172.17.0.0/16
        except:
        - 172.17.1.0/24
    - ipBlock:
        cidr: 1.1.1.0/24
    -....
```

( copied from NetworkPolicy ) - instead of another variation. Will also
make it easier to write NetworkPolicies - I expect users will want to do
this as well
for 'in-depth' defence.

mesh/v1alpha1/config.proto
ramaraochavali: Sure

mesh/v1alpha1/config.proto
ramaraochavali: yeah. Copied from RFC. Will correct

ramaraochavali:
@costinm My preference would be to leave it here as it s mesh wide config and people running with single network do not have to worry about additional network config. But I see your point about redundancy, but I expect people who use it for multi network cases, do not have to worry about it. And I am not sure where/how that from_cidr is used.
@howardjohn  WDYT?

costinm:
Networks are also 'mesh wide', and it is fine to define a single network (
default ).

In reality "mesh config" is not 'mesh wide' - but specific to one revision
in one cluster - promoting the network description to CRD could make it
mesh wide,
and seems a generally useful feature. If multi-cluster code starts watching
them - we won't have to copy the same thing in all clusters and keep it in
sync.

There are other related use cases - for example non-Istio gateways that use
a dedicated IP range ( separate node pool, etc) and we 'trust' them using
 L3 (assuming network security).

nrjpoddar:
Are we still adding fields to MeshConfig or have frozen it to provide a way to migrate to the CRD?

ramaraochavali:
This is not for proxyConfig - Are you thinking it is for proxy config?

ramaraochavali:
@costinm @howardjohn Please let me know your preference - I want to move forward with one way or other. I am not fully familiar with network use cases so I am good with what you think is best place to put this. But I would like to add API support for this

costinm:
My preference remains to use networks. It may not even require any API change - just a flag/env variable in Istiod to indicate that networks will be used to trigger this behavior.

User will have to configure a network ('default') with all the ranges. They may also configured additional networks. All defined networks will be treated as internal, if this flag is set.

ramaraochavali:
I am fine with that.

@howardjohn Are you good with that as well?

ramaraochavali:
Closing this in-favour of https://github.com/istio/istio/pull/38801

costinm:
Endpoints are already repeated.

- fromCidr: 1.1.1.0/24
- fromCidr: .....


I don't think making a fromCidr repeated is backward compatible, but we
could add a new one

endpoints:
  fromCidrs:
   - ....
  - ....

How about getting the implementation with the current API, and sync in next
NetworkingWG -  I think having a good way to
describe the IP ranges and networks is pretty important but may be time to
move it to a proper CRD and figure out how
it can be used with K8S Gateway attachment policy, etc.

Meanwhile - I don't think the feature you need is blocked by API change if
you use networks, just not ideal UX.

PR #2336, opened 2022-05-02
Extends the scope of ImagePullPolicy to cover http[s]:// and file:// as …

ingwonsong:
This PR proposes extending the scope of `ImagePullPolicy` to deal with not only OCI Image but also the direct fetching of Wasm module via `http[s]`(or `file`). 

There is no explicit API change, but comments are changed for the documentation purpose.

In addition, more specific meaning of `Always` is described in the comment. 
When `Always` policy is applied, then Wasm module will be pulled for each time when WasmPlugin resource is changed.
This is exactly the same with [the original definition in the design doc](https://docs.google.com/document/d/15GQnhOpLfvgyOZ8VDF5KD-q2cjXEMLdTQMm0YPswIu0/edit#heading=h.u37h4u1pn1d4)

This PR was started from the discussion on https://github.com/istio/istio/pull/38475#issuecomment-1111534522

cc: @zirain @dgn


ingwonsong:
@dgn do you have any thoughts about this PR?

extensions/v1alpha1/wasm.proto
```diff
@@ -231,14 +256,16 @@ message WasmPlugin {
...
+  // with `http[s]://` or `file://` as well.
   // The pull behaviour to be applied when fetching an OCI image. Only
   // relevant when images are referenced by tag instead of SHA. Defaults
   // to IfNotPresent, except when an OCI image is referenced in the `url`
   // and the `latest` tag is used, in which case `Always` is the default,
   // mirroring K8s behaviour.
   // Setting is ignored if `url` field is referencing a Wasm module directly
   // using `file://` or `http[s]://`
-  PullPolicy image_pull_policy = 4;
+  PullPolicy image_pull_policy = 4 [deprecated=true];
```
howardjohn, line 268: I would prefer to keep the same field, its a pain dealing with deprecation, `imagePullPolicy` is extremely familiar to k8s users, and I think HTTPS urls could still be called 'images' if you squint your eyes a bit.

ingwonsong: I understood your point.

One thing I worried, if we want to keep `imagePullPolicy`, I felt that `imagePullSecret` may give a nuance that it supports https as well somehow. (Are there any good idea or previous thoughts about https auth?)

WDYT?

dgn: +1 to @howardjohn's point. I think using `ImagePullPolicy` is fine here. we should implement it for OCI images as well

ingwonsong: I see. Then, I will just update the comments. Thank you very much for the comments!

ingwonsong:
I think it would be better to mention this change in the main PR https://github.com/istio/istio/pull/38475 because there is no explicit API change.

ingwonsong:
@ericvn Could you please take a look?

ingwonsong:
With https://github.com/istio/istio/pull/39212, the implementation change is settled.
We need to update the comment for explaining the changed behavior in API layer.

louiscryan (approved):

PR #2341, opened 2022-05-09
add support for max connection duration

ramaraochavali:
Use case: When mesh is configured to talk to an external service, envoy validates TLS certs during handshake and if the certs are rotated, the existing connections would continue to the expired certs. This is fine with mesh to mesh calls but if some external services validate client cert expiry on every request, the requests from old connections would fail. This PR allows users to configure max_connection_duration so that connections can be closed with TTL of certs.

Related issue https://github.com/istio/istio/issues/7630


networking/v1beta1/destination_rule.proto
```diff
@@ -648,6 +648,11 @@ message ConnectionPoolSettings {
 
     // If set then set SO_KEEPALIVE on the socket to enable TCP Keepalives.
     TcpKeepalive tcp_keepalive = 3;
+
+    // The maximum duration of a connection. The duration is defined as the period since a connection
+    // was established. If not set, there is no max duration. When max_connection_duration
+    // is reached the connection will be closed. Duration must be at least 1ms.
+    google.protobuf.Duration max_connection_duration = 4;
```
howardjohn, line 655: if this is about rotating certs not re-establishing, maybe it would be safer to have connections dropped on cert rotation instead of some max con duration. This field seems hard to use

hzxuzhonghu:
>some external services validate client cert expiry on every request,

If this fails, isn't the connection closed?

ramaraochavali:
> If this fails, isn't the connection closed?

No. If the external service just returns 403.

networking/v1beta1/destination_rule.proto
ramaraochavali: There is no way to drop connections in Envoy today. The external service can do it but that needs changes to external service code which is not ideal.

Are you suggesting we enhance Envoy to do that? If we have to do it Envoy, this again has to be based on some config in cluster and (possibly listener)

ramaraochavali: @lizan might know details of how easy it is to do in Envoy

Another option is to add a env var in Istiod "MAX_CONNECTION_DURATION_FOR_SERVICE_ENTRIES" or some thing like that instead of this API

lizan: It isn't easy to implement in Envoy today but we might want it in longer term. Though it is still useful to expose `max_connection_duration` in DR anyway?

ramaraochavali: @howardjohn what do you think? How should we proceed on this? I kind of agree that it is little hard to use but given where we are I see no harm in exposing this and user can configure if they want to.

ramaraochavali:
> Using max connection duration seems like a bad pattern, even for your own use case?

There is no other option now especially if you do not have control over the server. The requests from old connections with expired certs, keep failing till the pod is restarted.

ramaraochavali:
Changing envoy is not that easy as mentioned by @Lizan https://github.com/istio/api/pull/2341#discussion_r868899007. 

And Envoy's recommendation is to use max_connection_duration https://github.com/envoyproxy/envoy/issues/17999

howardjohn:
I don't really understand how it solves the problem. Say I set `max_con_duration` to 1 min (pretty aggressive). Doesn't that mean I can have up to 1min of expired certs?

ramaraochavali:
Let us your cert TTL is 6 days and the cert is refreshed every 4 days by your certificate tooling, you can configure the max_connection_duration < 2 days i.e. any connection opened just before it is refreshed on the 4th day is guaranteed to be closed before it expires.

lizan:
> Using max connection duration seems like a bad pattern, even for your own use case?

It is actually better than trigger cluster/listener warming process when you get cert rotated, as you don't want many connections/instance to be reconnect at the same time.

As long as the time between cert rotation and cert expiration < max connection duration you won't have expired cert during the connection lifespan.

hzxuzhonghu:
The tricky point is you never know precisely when the cert is signed, so it is hard to align the time.  If `max connection duration` is too aggressive, it will influence normal requests

ramaraochavali:
> The tricky point is you never know precisely when the cert is signed, so it is hard to align the time. If max connection duration is too aggressive, it will influence normal requests

I think as @lizan mentioned if know the times, you can set conservatively - and we have done that using Envoy Filter and it works. Since this goes through draining, it should not impact normal requests.

Also it is generally used for other use cases also
https://www.envoyproxy.io/docs/envoy/latest/faq/configuration/timeouts#connection-timeouts
https://github.com/envoyproxy/envoy/issues/15107
https://github.com/projectcontour/contour/issues/2247

So as @lizan mentioned earlier, I think would be a generally useful setting for various use cases. Since we are not providing  any default value, it is up to the user to configiure.

ramaraochavali:
@howardjohn Do you think we can move forward with this given that it is generally useful?

ramaraochavali:
@howardjohn gentle ping. WDYT? I am OK to close this if you think it is not useful but our experience of dealing with lot of external services on which we do not have control over cert validations, this is going to be a problem and max_connection_duration in general is a useful feature as @lizan mentioned.

howardjohn (approved):
sorry forgot about this

ramaraochavali:
no problem

ericvn (approved):

PR #2343, opened 2022-05-11
Hide verification_key from documentation

ingwonsong:
This PR proposes to hide `verification_key` field from documentation, and declare that the field will be not implemented until the detailed design is elaborated.

In current situation, there are two major stream for signing/verifying the OCI image. On the other hand, it seems to be no apparent major stream for signing and verifying Wasm binary. 
In this case, we can think of more generalized API interface to deal with such multiple mechanisms, and it need to be designed more general way rather than providing just a `verification_key` field.

Therefore, I would like to propose to hold this field at this time and get it back later.

cc: @dgn @mathetake @PiotrSikora @mandarjog @bianpengyuan

### DIFF
--- a/extensions/v1alpha1/wasm.gen.json
+++ b/extensions/v1alpha1/wasm.gen.json
@@ -84,7 +84,7 @@
             "type": "string"
           },
           "verificationKey": {
-            "description": "Public key that will be used to verify signatures of signed OCI images or Wasm modules. Must be supplied in PEM format.",
+            "description": "Public key that will be used to verify signatures of signed OCI images or Wasm modules.",
             "type": "string"
           },
           "pluginConfig": {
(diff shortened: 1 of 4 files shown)


howardjohn (approved):

ingwonsong:
>

PiotrSikora (approved):
> I am not saying "not supporting both", rather, I would like to say that we need more generalized API spec for dealing with such multiple ways, like "typed config".

I'm not convinced if that's the case, thought. Users need to specify `verification_key` regardless of which container signature was used, and rest can be detected based on signature in the container.

But I agree that we should hide this field until it's implemented, so LGTM.

ingwonsong:
> rest can be detected based on signature in the container

Yes. It makes sense. 

I just considered more complicated situation like verification with KMS, DNSSec, .... 
In addition, I think that there might be some possibilities to use PKI more aggressively, so the certificate, which is used for signing, is included (or referenced) in the signed Wasm code (like Windows' signed executable). Then, the trusted CA lists should be given to the verification logic.

Of course, if all the things can be set without complicated configuration, it would be the best. 
I just wanted to touch the possibility because we do not have the detailed design yet.

Thanks.

ingwonsong:
> I will cherry-pick it on top of release-1.14 in a new PR and assign it to you.

I assigned myself for that. :)

PR #2364, opened 2022-05-25
fix proxy-config reference doc

nicole-lihui:
fix proxy-config reference doc:
1.  yaml error: selector.matchLabels
2. formating error 
image


ericvn:
Will need to be cherry-picked to 1.14 to make it into preliminary.istio.io right now (istio.io after 1.14 is released).

hzxuzhonghu (approved):

howardjohn (approved):

PR #2368, opened 2022-05-27
[WIP] Mesh partitioning policy

nmittler:
The main use case for this currently is for on-boarding a new cluster into the mesh. For brownfield clusters, it may be undesirable to allow cross-cluster traffic initially, but rather allow traffic in and out of the cluster to be enabled in a staged manner by a mesh admin.


mesh/v1alpha1/config.proto
```diff
@@ -1088,6 +1088,45 @@ message MeshConfig {
...
+  // If there are policies for workloads in both "cluster1" and "cluster2", communication
+  // between the two clusters will be determined based on the most restrictive policy for
+  // ingress and egress.
+  message PartitionPolicy {
+    // The labels for selecting the subject workloads for this isolation policy.
+    map<string, string> select = 1;
+
+    // The labels for selecting which workloads are allowed to send traffic to
+    // the subject workloads. If empty, ingress is allowed from anywhere in the mesh.
+    map<string, string> allow_ingress_from = 2;
```
howardjohn, line 1117: I think the "allow" terminology is misleading here unless we are doing Authz enforcement on this (which we cannot since cluster is not part of the identity?)

howardjohn: Although I don't understand what "ingress" even means if not enforced... so I think i am confused what this does :slightly_smiling_face:

mesh/v1alpha1/config.proto
```diff
@@ -1088,6 +1088,45 @@ message MeshConfig {
...
+  //     allowIngressFrom:
+  //       "topology.istio.io/cluster": "cluster1"
+  //     allowEgressTo:
+  //       "topology.istio.io/cluster": "cluster1"
+  // ```
+  //
+  // If there are policies for workloads in both "cluster1" and "cluster2", communication
+  // between the two clusters will be determined based on the most restrictive policy for
+  // ingress and egress.
+  message PartitionPolicy {
```
howardjohn, line 1111: On all three of these - what labels are we selecting?

mesh/v1alpha1/config.proto
```diff
@@ -1088,6 +1088,45 @@ message MeshConfig {
...
+    // The labels for selecting which workloads are allowed to receive traffic from
+    // the subject workloads. If empty, egress is allowed to anywwhere in the mesh.
+    map<string, string> allow_egress_to = 3;
+  }
+
+  // $hide_from_docs
+  // Policy for generating partitions of workloads within the mesh. If no policy is
+  // provided for a given workload, ingress and egress traffic from anywhere in the mesh
+  // is allowed.
+  repeated PartitionPolicy partition_policy = 64;
```
hzxuzhonghu, line 1128: I am thinking maybe we only need to support  cluster & namespace partition.

Allow any label selecting partition in meshConfig seems not a good practice.

costinm:
IMO it is very confusing. What does 'partition' even mean ? If it's about enforcement - we have RBAC rules that can do pretty much everything in controlling who is allowed to talk. 

If we are talking visibility - we could extend Sidecar visibility, which while complicated it is now relatively well used.

We also have a lot of configs related to tenants and what namespaces to watch - that can also be extended. 

I think adding a 4th way to do things will not help as much as adjusting the existing mechanisms.

PR #2372, opened 2022-06-02
Hide dns certificate management from docs

howardjohn:
Aligns with https://github.com/istio/istio.io/pull/11369


linsun:
Do we need a release note for this in case anyone is using this?

howardjohn:
We aren't removing it, just hiding it

hzxuzhonghu:
I think we need to deprecate it, donot think any user is using this.

linsun:
approval pending on resolving merge conflict...

ericvn (approved):
LGTM. Just need to fix the rebase.

PR #2374, opened 2022-06-03
Document new sidecar annotation to exclude interfaces 

obaranov1:
Regarding the added feature istio/istio#34832

### DIFF
--- a/releasenotes/notes/2374.yaml
+++ b/releasenotes/notes/2374.yaml
@@ -1,8 +1,8 @@
 apiVersion: release-notes/v2
-kind: bug-fix
+kind: feature
 area: documentation
 issue:
-  - https://github.com/istio/api/pull/2374
+  - https://github.com/istio/istio/pull/34832
 releaseNotes:
   - |
     **Added** sidecar `traffic.sidecar.istio.io/excludeInterfaces` annotation documentation.
--- a/releasenotes/notes/2374.yaml
+++ b/releasenotes/notes/2374.yaml
@@ -1,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: documentation
+issue:
+  - https://github.com/istio/api/pull/2374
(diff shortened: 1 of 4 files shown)


ramaraochavali:
incorrect PR link in the description

howardjohn (changes requested):
I don't see such an annotation anywhere in the codebase?

obaranov1:
the link has been updated. the codebase is there

howardjohn:
But no annotation was ever defined? If you grep the code there is no mention

rveerama1:
We have a use case around this feature. When we were exploring about this feature in the documentation https://istio.io/latest/docs/reference/config/annotations/ we didn't find any thing. Later we noticed about this https://github.com/istio/istio/pull/34832. We haven't tried it on any environment yet. Thought only documentation might be missing. If there are no objections can we proceed with adding pod annotation in the code base?

linsun:
Is annotation better than label here? We are trying to shift away from annotation for sidecar injection.

Can you please document your scenario in a doc as well?

howardjohn:
Annotation is fine here, label is only needed for the webhook stuff since
its used in label selectors

rveerama1:
@howardjohn PTAL about documentation.

howardjohn (approved):

releasenotes/notes/2374.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: bug-fix
+area: documentation
+issue:
+  - https://github.com/istio/api/pull/2374
```
ericvn, line 5: Nit: Maybe point to feature: https://github.com/istio/istio/pull/34832

ericvn (approved):
Approved. Marked DNM for the nit. Feel free to fix or not. Then remove the DNM to let this in. Thanks.

PR #2385, opened 2022-06-17
Namespace annotation for cluster leaders

frankbu:
New namespace annotation used to configure leader(s) for remote clusters.

See https://github.com/istio/istio/pull/39406

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -498,8 +498,8 @@ annotations:
   - name: topology.istio.io/controlPlaneClusters
     featureStatus: Alpha
     description: A comma-separated list of clusters (or * for any) running istiod that should attempt leader election
-      for a remote cluster thats system namespace includes this annotation. Istiod will attempt to lead unannotated
-      remote clusters by default.
+      for a remote cluster thats system namespace includes this annotation. Istiod will not attempt to lead unannotated
+      remote clusters.
     deprecated: false
     hidden: false
     resources:
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -494,3 +494,11 @@ annotations:
     hidden: true
     resources:
       - WorkloadEntry
+
+  - name: topology.istio.io/istiodClusters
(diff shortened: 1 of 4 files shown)


annotation/annotations.yaml
```diff
@@ -494,3 +494,11 @@ annotations:
     hidden: true
     resources:
       - WorkloadEntry
+
+  - name: topology.istio.io/istiodClusters
+    featureStatus: Alpha
+    description: Comma-separated list of clusters (or * for any) with an istiod that should attempt leader election for a remote cluster.
```
ramaraochavali, line 500: Comma-separated list of clusters (or * for any) running istiod - may sound better

hzxuzhonghu: It is not clear where to set the annotation: remote cluster or config? And what is the behavior when not set?

annotation/annotations.yaml
```diff
@@ -494,3 +494,13 @@ annotations:
     hidden: true
     resources:
       - WorkloadEntry
+
+  - name: topology.istio.io/istiodClusters
+    featureStatus: Alpha
+    description: A comma-separated list of clusters (or * for any) running istiod that should attempt leader election
+      for a remote cluster whose system namepace includes this annotation. Istiod will not attempt to lead other remote
+      clusters by default.
```
frankbu, line 502: We could, instead, make istiod always attempt to lead by default (when there is no annotation).
That way this annotation would just be an optimization to prevent it from logging failure messages when it doesn't have permission, or otherwise fail the leader election to a more local istiod. WDYT?

cc @howardjohn @stevenctl

frankbu: It's sort of backwards compatible either way since it only comes into play with istiodless remotes which we still haven't changed the MC docs to use (but I'm going to, once these PRs are merged).

External control plane with multiple remotes would be affected. I don't think that is really used by anybody yet, but to be safe I suppose you're right and we should make it optional.

howardjohn (approved):

PR #2388, opened 2022-06-23
add support for resources in OTel provider

zirain:
follow https://github.com/envoyproxy/envoy/pull/21603
original idea is here: https://github.com/envoyproxy/envoy/pull/20883#issue-1208019077


mesh/v1alpha1/config.proto
```diff
@@ -971,6 +971,17 @@ message MeshConfig {
...
+      // Optional. Additional attributes that describe the node.
+      // Alias to `resource_attributes` filed in [Open Telemetry](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/open_telemetry/v3/logs_service.proto)
+      //
+      // Example:
+      // ```
+      // resources:
+      //   pod_name: "fortio-deploy-5f75cc56b5-d5pc4"
+      //   pod_namespace: "default"
+      // ```
+      google.protobuf.Struct Resources = 5;
```
costinm, line 984: In the otel proto, are they 'struct' or map<string,string> ? I.e. will it work if someone passes a complex struct (all the way to otel proto ) ?

mesh/v1alpha1/config.proto
```diff
@@ -971,6 +971,17 @@ message MeshConfig {
           // ```
           google.protobuf.Struct labels = 2;   
       }
+
+      // Optional. Additional attributes that describe the node.
+      // Alias to `resource_attributes` filed in [Open Telemetry](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/access_loggers/open_telemetry/v3/logs_service.proto)
```
howardjohn, line 976: Can we point to the OTEL spec, not envoy?

costinm:
Looking at otel - it's key/value pairs.

Since this is a global setting - I doubt users just want to add some
constant key/value pairs in logs from ALL workloads, I can't imagine any
use cases for wasting the storage space.

There may be a value in providing a list of dynamic attributes or
baggage attributes to include - the value will by based on
source (for example pod id) or request ( if they're from the baggage).

However I am not sure this belongs in MeshConfig - instead of the new
telemetry API - and if it should be only for access logs
or across metrics/trace/logs.

mesh/v1alpha1/config.proto
zirain: in OTel here is [KeyValue ](https://github.com/open-telemetry/opentelemetry-proto/blob/9e7c1b3b3bff53cdce5dcd7fd139343be36862b2/opentelemetry/proto/resource/v1/resource.proto#L32), we have use in `LogFormat.Labels`, keep both consistent.

mesh/v1alpha1/config.proto
zirain: sure, will update later.

mesh/v1alpha1/config.proto
```diff
@@ -971,6 +971,17 @@ message MeshConfig {
...
+      //
+      // Example:
+      // ```
+      // resources:
+      //   pod_name: "fortio-deploy-5f75cc56b5-d5pc4"
```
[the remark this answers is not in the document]
zirain: the original purpose is send constant values(e.g. pod_name, cluster_name, mesh_name etc) per connection, instead of per request.

I may write a bad example:

```
resource:
  pod_name: ${POD_NAME} # replace in istiod
```

zirain:
> I doubt users just want to add some
constant key/value pairs in logs from ALL workloads, I can't imagine any
use cases for wasting the storage space.

my original idea is sending constant values like pod_name, pod namespace per connection

> However I am not sure this belongs in MeshConfig 

I think it belongs to `ExtensionProvider`

costinm:
Why would Istiod replace the 'constant' and send all this config to the
pods ? Will impact the caching of the config, config size, etc.
They already know their own pod id.

zirain:
> Why would Istiod replace the 'constant' and send all this config to the pods ? Will impact the caching of the config, config size, etc. They already know their own pod id.


if we do just add some constant values like envoy do here:

https://github.com/envoyproxy/envoy/blob/4809663c73ae7bc7963de2ddb61bc8cd7420f667/source/extensions/access_loggers/open_telemetry/grpc_access_log_impl.cc#L54

this api change is unneeded.

costinm:
You mean to envoy otel implementation ? We do pass the pod name and a lot
of other info in the bootstrap config - so envoy should
have access to everything without Istiod having to generate an un-cacheable
config to tell an envoy plugin a value it already has...

The log entry content should be customizable - but that should go to
Telemetry CRD, consistent with how tracing tags are customized and
independent of implementation.

zirain:
I did what you said in https://github.com/envoyproxy/envoy/pull/20883, but in the end it was not accepted by the envoy maintainers, more details is here https://github.com/envoyproxy/envoy/pull/20883#discussion_r870690674.

can we just add `pod_name`, `pod_namespace` by default(we can get them in istiod), but still keep resoucre_attributes not configurable to end-users?

costinm:
I don't think pod_name or pod_namespace are a good idea by default - they
are specific to k8s and I think we are still supporting VMs and non-k8s
environments,
but the source workload, etc are good substitutes and will also match what
the user has in the dashboards/alerts/etc.
I don't mind having any of the labels we use in metrics also present in the
access log. I think one of the ideas of OTel is that access logs, metrics
and
traces are using the same protocol and should be consistent with each
other.

I think the more interesting question is how does it relate to the labels
that OTel itself defines - the resource semantic convention,
https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/

If we want to go with pod - I would use the names from that doc,
k8s.pod.name - but I don't think this PR is the right place to make
 a decision to move away from our labels to OTel ( in particular since the
k8s resource names are marked as 'experimental' ).

However I have a different problem now - can you confirm if the access log
config is in bootstrap or generated in the LDS/RDS ?
If it is in bootstrap - bootstrap is generated by the agent and doesn't
have access to MeshConfig or any Istio CRs - only the
injected ProxyConfig and few other injected env variables. My concerns
about caching are also irrelevant there - but I would still
want the labels to follow Istio or at least OTel naming.

zirain:
> However I have a different problem now - can you confirm if the access log
config is in bootstrap or generated in the LDS/RDS ?

it's generated in LDS.

> I think we are still supporting VMs and non-k8s environments

my fault, always forgot this. I will think about all of this, wish I could come up with a better solution.

zirain:
closed as need more investigation

PR #2392, opened 2022-06-24
Deprecate MeshConfig.Certificates

hzxuzhonghu:
Using istio to sign certificates will be no longer supported

replace #2372


mesh/v1alpha1/config.proto
```diff
@@ -355,8 +355,9 @@ message MeshConfig {
   // - `%SERVICE%` will use reviews.prod as the stats name.
   string outbound_cluster_stat_name = 45;
 
+  // $hide_from_docs
   // Configure the provision of certificates.
-  repeated Certificate certificates = 47;
+  repeated Certificate certificates = 47 [deprecated=true];
```
ericvn, line 360: We probably need a release note if we are deprecating this.

mesh/v1alpha1/config.proto
```diff
@@ -356,6 +356,11 @@ message MeshConfig {
   string outbound_cluster_stat_name = 45;
 
   // Configure the provision of certificates.
+  //
+  // Note: Deprecated, please refer to cert-manager to sign certificates.
```
costinm, line 360: 'please refer to Cert-Manager or other cert provisioning solutions to sign DNS certificates.' 

And remove the 'secrets generated are of different keys' - or add:
"An equivalent behavior can be achieved with Cert-Manager using an Issuer using the same 
private key that Istio is using. Please note that CertManager issuer expects secrets with 
tls.key, tls.crt, and Istio self-signed root CA is using key.pem, cert-chain.pem"

But not sure we need such details in the API doc - a wiki maybe.

hzxuzhonghu: > But not sure we need such details in the API doc 

maybe the transition is enough, users will refer the docs from official website.

hzxuzhonghu:
@howardjohn @costinm

howardjohn (approved):

ericvn (approved):

PR #2403, opened 2022-06-30
feat(accesslog): support config omit_empty_values

j2gg0s:
Support user config omit_empty_values for Envoy's AccessLog, reference [Issue/31060](https://github.com/istio/istio/issues/31060).

Related envoy's [SubstitutionFormatString.omit_empty_values](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/substitution_format_string.proto#envoy-v3-api-msg-config-core-v3-substitutionformatstring)


mesh/v1alpha1/config.proto
```diff
@@ -866,6 +866,10 @@ message MeshConfig {
 
       // Optional. Allows overriding of the default access log format.
       LogFormat log_format = 2;
+      // Optional. If set to true, when command operators are evaluated to null,
+      // for text, the output of the empty operator is changed from - to an empty string, so that empty values are omitted entirely.
+      // for labels, the keys with null values are omitted in the output structure.
+      bool omit_empty_values = 3;
```
zirain, line 872: can you and some example for this?

j2gg0s: > can you and some example for this?

Of course, but as a newbie of istio, i don't know where to add example, and what should i add? Example for usage?

j2gg0s:
> @j2gg0s: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> release-notes_api	[71e0d4d](https://github.com/istio/api/commit/71e0d4d91e8625dd1f3b23323c83a37143b4a1c5)	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_api/2403/release-notes_api/1542431536324808704)	false	`/test release-notes_api`

this pr need release notes?

mesh/v1alpha1/config.proto
```diff
@@ -866,6 +871,10 @@ message MeshConfig {
 
       // Optional. Allows overriding of the default access log format.
       LogFormat log_format = 2;
+      // Optional. If set to true, when command operators are evaluated to null,
```
ramaraochavali, line 874: nit: empty line before this?

mesh/v1alpha1/config.proto
```diff
@@ -121,6 +121,11 @@ message MeshConfig {
   // Empty value results in proxy's default access log format
   string access_log_format = 24;
 
+  // If set, omity empty in proxy's access logging.
```
ramaraochavali, line 124: nit: omits empty

mesh/v1alpha1/config.proto
ramaraochavali: yes . Add it comments like https://github.com/istio/api/blob/master/mesh/v1alpha1/config.proto#L863

mesh/v1alpha1/config.proto
```diff
@@ -121,6 +121,11 @@ message MeshConfig {
   // Empty value results in proxy's default access log format
   string access_log_format = 24;
 
+  // If set, omit empty in proxy's access logging.
+  // for text format, the empty is change from - to an empty string.
+  // for json format, the keys with null values are omitted.
+  bool access_log_omit_empty_values = 64;
```
douglas-reid, line 127: I feel like we should not add this in two different places in MeshConfig.  If we are defining new configuration, we should do it in the place we expect new configuration to live -- provider config (as this PR does below). Adding this config in two places makes it less clear which is the proper place to for that configuration to live and will create more maintenance churn, etc., as provider configuration evolves. It seems fine to me to tell users that the new features will be available via the new configuration mechanism.

j2gg0s: Agree and I did spend a lot of time debugging similar issues when I started using istio.

However, if we dont support this config to MeshConfig, what about people who only use MeshConfig?
Should we only offer this feature with Telemetry? If so, i can revert unnecessary code.

douglas-reid: My feeling is that since we've survived this long without having this as a MeshConfig-supported feature, there isn't a great demand among current MeshConfig users for this feature. That leads me to believe that we should introduce it with the new mechanism only.

I'd like to see a follow-on to have the install options for MeshConfig update the envoy provider config to match to make the transition for early adopters as seamless as possible, but that's beyond the scope of this effort.

j2gg0s: I undo the related changes

linsun:
Yes, can we add a release note for this pls?

mesh/v1alpha1/config.proto
```diff
@@ -867,6 +867,13 @@ message MeshConfig {
 
       // Optional. Allows overriding of the default access log format.
       LogFormat log_format = 2;
+
+      // Optional. If set to true, when command operators are evaluated to null,
+      // for text, the output of the empty operator is changed from - to an empty string, so that empty values are omitted entirely.
+      // for labels, the keys with null values are omitted in the output structure.
+      //
+      // Example: `omit_empty_values: true`
+      bool omit_empty_values = 3;
```
linsun, line 876: Could we explain why this  is useful and give 1 or two examples?  I had to lookup the original issue and envoy docs.

douglas-reid: @j2gg0s maybe the action here is to link to the envoy docs directly as a "for more information" style helper?

j2gg0s: @douglas-reid Did i understand correctly? With latest note

douglas-reid: That seems reasonable to me. @linsun wdyt ?

zirain:
kindly ping @linsun

akashjain971:
Anyone taking this forward?

akashjain971:
> > Anyone taking this forward?
> 
> @akashjain971 I switched to EnvoyFilters to manage logs, such as:
> 
> ```yaml
> kind: EnvoyFilter
> metadata:
>   name: log
>   namespace: istio-system
> spec:
>   configPatches:
>   - applyTo: NETWORK_FILTER
>     match:
>       listener:
>         filterChain:
>           filter:
>             name: envoy.filters.network.http_connection_manager
>     patch:
>       operation: MERGE
>       value:
>         typed_config:
>           '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
>           accessLog:
>           - name: envoy.access_loggers.file
>             typedConfig:
>               '@type': type.googleapis.com/envoy.extensions.access_loggers.file.v3.FileAccessLog
>               logFormat:
>                 jsonFormat:
>                   authority: '%REQ(:AUTHORITY)%'
>                   ...
>                   upstream_transport_failure_reason: '%UPSTREAM_TRANSPORT_FAILURE_REASON%'
>                 omit_empty_values: true
>               path: /dev/stdout
> ```

Thank you @j2gg0s. This should sail my boat for now.
But this is good to have in Istio anyway. Why is this PR on hold?

akashjain971:
Looks like **OmitEmptyValues** is already supported here.
This required Istio code changes and cannot be done from the IstioOperator.
https://github.com/istio/istio/blob/8f73141f83f6610943286e4fc65fc15bc7fd41c8/pilot/pkg/model/telemetry_logging.go#L250

PR #2405, opened 2022-07-02
add stat prefix for routes

ramaraochavali:
https://github.com/envoyproxy/envoy/pull/21302 added support for route level stats in Envoy. This PR enables that in Istio


linsun:
@ramaraochavali can you add a release note for this?

ramaraochavali:
Sure. I was thinking of adding it when the feature is implemented in Istio. I can add it if you prefer to add it here also.

ericvn (approved):
LGTM

louiscryan:
paging @douglas-reid

louiscryan:
How is this intended to function with the Telemetry API ?

douglas-reid:
@louiscryan I think this is beyond the current scope of the Telemetry API. As I understand it, this is adding prefixes to envoy-generated stats, which has generally been distinct from the domain of the Telemetry API (which has focused on controlling Istio generation). Mostly, the envoy-level interactions for stats tend to be at the bootstrap level, which the Telemetry API has not tried to cover. 

The use case is not fully fleshed out here, but I imagine that @ramaraochavali desires to include some envoy-level metrics specific to endpoints. @ramaraochavali is that correct?

The Telemetry API-based equivalent, iiuc, would be adding new labels to the istio metrics for path -- and using attribute-gen (or equivalent) to manage scoping of that generation for specific "high-value" destinations. That pattern is a bit complex, and involves bits outside the Telemetry API as well (existing [docs](https://istio.io/latest/docs/tasks/observability/metrics/classify-metrics/#classify-metrics-by-request)). 

I think at a minimum we should be clearer in the documentation provided that this prefix is **only** for proxy-level stats (`envoy_*`) and not service-level (`istio_*`) stats. It _may_ also be worth noting that using this prefix _may_ require bootstrap changes to `stats_matcher` configuration, etc., to properly use (although @ramaraochavali may solve that in the control plane implementation?).

ramaraochavali:
>  I imagine that @ramaraochavali desires to include some envoy-level metrics specific to endpoints. @ramaraochavali is that correct?

Yes. This is to set the stat prefix to envoy route config so that it can generate route level stats.

> I think at a minimum we should be clearer in the documentation provided that this prefix is only for proxy-level stats (envoy_*) and not service-level (istio_*) stats. It may also be worth noting that using this prefix may require bootstrap changes to stats_matcher configuration, etc., to properly use (although @ramaraochavali may solve that in the control plane implementation?).

I will update the docs.

networking/v1alpha3/virtual_service.proto
```diff
@@ -1112,6 +1112,11 @@ message HTTPMatchRequest {
   // If the VirtualService has a list of gateways specified in the top-level `gateways` field,
   // it must include the reserved gateway `mesh` for this field to be applicable.
   string source_namespace = 13;
+
+  // The human readable prefix to use when emitting statistics for this route.
+  // The statistics are generated with prefix route.<stat_prefix>.
+  // This should be set for highly critical routes that one wishes to get “per-route” statistics on.
+  string stat_prefix = 14;
```
howardjohn, line 1119: can we please document that this is for the unstable envoy metrics, not the core Istio ones? thanks?

and maybe link to envoy docs for them

hzxuzhonghu, line 1119: I think this should not be in HTTPMatchRequest, which contains all matching conditions

ramaraochavali: For each match we create a route - In Envoy it is exposed at route. I did not get why it can not be at Match? This is for path based routing. For each match, we want stat to be generated.

hzxuzhonghu: I think `stat_prefix ` is to specify a behavior for a route after matching conditions, however this pr mess it in the match conditions.

ramaraochavali: It is per match in Istio (translates to route in Envoy) to get path level stats.

hzxuzhonghu: I can see it is aligned with envoy api

PR #2407, opened 2022-07-04
Virtual Service Direct Response

sschepens:
Fixes https://github.com/istio/istio/issues/29264


networking/v1beta1/virtual_service.proto
```diff
@@ -1436,6 +1443,73 @@ message HTTPRedirect {
...
+  HTTPBody body = 2;
+}
+
+message HTTPBody {
+  oneof specifier {
+    // response body as a string
+    string string = 1;
+
+    // response body as base64 encoded bytes.
+    bytes bytes = 2;
```
hzxuzhonghu, line 1509: how would this be used?

sschepens: This is mostly yo copy envoy's API, but this could be used to specify bodies which are not text based, for example gRPC.

ericvn:
It could be that there is a difference in tooling used when running locally vs within the pipeline. I hope to fix that here shortly.

sschepens:
@ericvn I pushed a commit using the previous image so that CI is happy.

Another thing, OpenApi generation is assigning `format: binary` to the `bytes` field which I think is not exactly correct, I think `format: byte` would be best, which is theoretically a base64 encoded bytes.
Is there any way to customize the type being assigned to a field?

linsun (commented):
LGTM overall, would like an example on bytes

sschepens:
@linsun do you have any idea on the question I asked in my previous comment? when using `bytes` in protobuf it would seem that we're translating to a `format: binary` in OpenApi which would not be correct in this case, `format: byte` would be more appropiate. Is there any way to change the translated format?

networking/v1alpha3/virtual_service.proto
```diff
@@ -1436,6 +1443,73 @@ message HTTPRedirect {
...
+  HTTPBody body = 2;
+}
+
+message HTTPBody {
+  oneof specifier {
+    // response body as a string
+    string string = 1;
+
+    // response body as base64 encoded bytes.
+    bytes bytes = 2;
```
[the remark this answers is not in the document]
sschepens: added

linsun (approved):
LGTM

louiscryan:
Note that the envoy filter has an "unknown security posture" that makes me quite nervous about including this in the API

louiscryan:
See https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/direct_response/v3/config.proto

sschepens:
This is what we will be using: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#envoy-v3-api-msg-config-route-v3-directresponseaction

louiscryan (approved):
LGTM

networking/v1beta1/virtual_service.proto
```diff
@@ -1441,6 +1448,177 @@ message HTTPRedirect {
...
+//   hosts:
+//   - ratings.prod.svc.cluster.local
+//   http:
+//   - match:
+//     - uri:
+//         exact: /v1/getProductRatings
+//     directResponse:
+//       status: 503
+//       body:
+//         string: "unknown error"
```
hzxuzhonghu, line 1493: is this right?

PR #2412, opened 2022-07-09
Add a match scheme to WasmPlugin for passing more specific traffic to Wasm module

ingwonsong:
This PR proposes to add a match scheme "TrafficSelector" to WasmPlugin for passing more specific traffic to Wasm module.

Currently, there is no way to select the traffic by its direction or port. So, all the HTTP traffics at the proxy, which is selected by "WorkloadSelector" and namespace, are passed to Wasm module.

This is a part of efforts for fixing: https://github.com/istio/istio/issues/39345 and https://github.com/istio/istio/issues/36238.

For the details and discussion, please refer to the [draft design doc](https://docs.google.com/document/d/1qknaZqXItogEmROVnQD61144pEW1X47bEQraI5D52uY/edit?usp=sharing&resourcekey=0-wHAgrakvVaMcnBm40bKuSg).


extensions/v1alpha1/wasm.proto
```diff
@@ -314,6 +315,42 @@ message WasmPlugin {
...
   VmConfig vm_config = 11;
+
+  // $hide_from_docs
+  // Hide this from doc until implementing this.
+  //
+  // TrafficSelector provides a method to select a specific traffic which will be
+  // passed to the Wasm plugin.
+  // When all the sub conditions in the TrafficSelector are satisfied, 
+  // the traffic will be selected.
+  message TrafficSelector {
```
zirain, line 326: IIRC, this is mostly like some struct in `Telemetry`?

ingwonsong: Yes. There are similar ones.

* [Log Selector](https://github.com/istio/api/blob/6a4e706e001820d590e9e4da87dbb25acd3b8310/telemetry/v1alpha1/telemetry.proto#L615-L622)
* [Tracing Selector](https://github.com/istio/api/blob/6a4e706e001820d590e9e4da87dbb25acd3b8310/telemetry/v1alpha1/telemetry.proto#L276-L285)

But, they are also defined as a nested message at this moment. So, I proposed this as a separate selector message. 
Also, the selectors in Telemetry do not have "port".

howardjohn:
cc @douglas-reid @kyessenov

extensions/v1alpha1/wasm.proto
```diff
@@ -314,6 +315,42 @@ message WasmPlugin {
...
+    // $hide_from_docs 
+    // Hide this from the doc until implementing this.
+    //
+    // Criteria for selecting traffic by their direction.
+    // Note that CLIENT and SERVER are analogous to INBOUND and OUTBOUND,
+    // respectively. If the selected workload is a gateway, the field is ignored.
+    // If not specified, this condition is evaluated to true for any traffic
+    // direction.
+    istio.telemetry.v1alpha1.WorkloadMode mode = 1;
```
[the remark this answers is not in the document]
ingwonsong: I agree with you. (Actually, my first version of this change was like that).

If there is no other opinion, I would like to move them to type/v1beta/selector.proto. 

In addition, I think "WorkloadModeSelector" seems to be more consistent if we want to move them there.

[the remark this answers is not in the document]
ingwonsong: Done to move WorkloadMode and PortSelector to `type/v1beta1/selector.proto`.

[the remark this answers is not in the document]
ingwonsong: For future use of "PortSelector" from Telemetry or other places, I have moved it as well.

extensions/v1alpha1/wasm.proto
```diff
@@ -314,6 +314,42 @@ message WasmPlugin {
...
+    istio.type.v1beta1.WorkloadMode mode = 1;
+  
+    // $hide_from_docs 
+    // Hide this from the doc until implementing this.
+    //
+    // Criteria for selecting traffic by their destination port.
+    // If one of the given ports is matched, this condition is evaluated to true.
+    //
+    // If not specified, this condition is evaluated to true for any port.
+    repeated istio.type.v1beta1.PortSelector ports = 2;
```
hzxuzhonghu, line 343: It is not clear what port it is? For outbound, I guess it is service port. What about inbound?

ingwonsong: In "Inbound" case, it would be the "destination port" that means.

Specifically, in [`inboundChainConfig`]( https://github.com/istio/istio/blob/89cdfc2760660912b8f5fcb64067cb8648e566bf/pilot/pkg/networking/core/v1alpha3/listener_inbound.go#L56 ), `targetPort` of `port` which has the type `ServiceInstancePort` will be used for the match.

hzxuzhonghu: I mean this should be documented

ingwonsong: Documented.

extensions/v1alpha1/wasm.proto
```diff
@@ -314,6 +314,45 @@ message WasmPlugin {
...
+  }    
+
+  // $hide_from_docs
+  // Hide this from doc until implementing this.
+  //
+  // Specifies the criteria to determine which traffic is passed to WasmPlugin.
+  // If a traffic satisfies any of TrafficSelectors, 
+  // the traffic passes the WasmPlugin.
+  repeated TrafficSelector match = 12;
```
ramaraochavali, line 355: matches? as it is repeated field?

ramaraochavali (commented):
LGTM. Couple of minor comments

extensions/v1alpha1/wasm.proto
```diff
@@ -314,6 +314,45 @@ message WasmPlugin {
   // Configuration for a Wasm VM.
   // more details can be found [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto#extensions-wasm-v3-vmconfig).
   VmConfig vm_config = 11;
+
+  // $hide_from_docs
+  // Hide this from doc until implementing this.
+  //
+  // TrafficSelector provides a method to select a specific traffic which will be
```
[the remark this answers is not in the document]
ingwonsong: Done. Thanks.

ramaraochavali (approved):

douglas-reid (commented):
This generally looks good.

I have two related questions:

1. Will there be a follow-on to update `Telemetry` to use the common `WorkloadMode` type?
1. Will there be a follow-on to move `TrafficSelector` to be a common type?

It _feels_ like `TrafficSelector` could be valuable outside of just `Wasm` resources.

ingwonsong:
> * Will there be a follow-on to update `Telemetry` to use the common `WorkloadMode` type?
Sounds good to me. I can add it to this PR. WDYT?

> * Will there be a follow-on to move `TrafficSelector` to be a common type?
Moving `TrafficSelector` to the common place also sounds good to me.
Is there any plan to regard `PortSelector` in `Telemetry`?

ingwonsong:
@douglas-reid

> * Will there be a follow-on to update `Telemetry` to use the common `WorkloadMode` type?

Sounds good to me. But, I think it seems to be possible to change it when `Telemetry API` is promoted to Beta or GA.

> * Will there be a follow-on to move `TrafficSelector` to be a common type?

Regarding `TrafficSelector`, IMO, it would be better to keep it in `WasmPlugin` at this moment, because it is possibly changed later.

zirain (approved):

ingwonsong:
Please note that https://github.com/istio/istio/pull/40684 is in WIP as an corresponding implementation of this PR.

type/v1beta1/selector.proto
```diff
@@ -35,3 +35,34 @@ message WorkloadSelector {
...
 }
+
+// PortSelector is the criteria for specifying if a policy can be applied to 
+// a listener having a specific port.
+message PortSelector {
+  oneof port {
+    // Valid port number
+    uint32 number = 1;
+    // Port name
+    string name = 2;
```
howardjohn, line 46: I am a bit suspicious of this. It is somewhat acceptable for inbound, since you control the service itself, but for outbound I think its unsafe. I don't think there are other APIs (including K8s) that match on service port _name_ of a service in another namespace?

howardjohn: Actually I am wrong, Ingress does... Gateway-api does not though. Maybe still be good to err on the side of more minimal API

type/v1beta1/selector.proto
```diff
@@ -35,3 +35,34 @@ message WorkloadSelector {
...
+// WorkloadMode allows selection of the role of the underlying workload in
+// network traffic. A workload is considered as acting as a SERVER if it is
+// the destination of the traffic (that is, traffic direction, from the
+// perspective of the workload is *inbound*). If the workload is the source of
+// the network traffic, it is considered to be in CLIENT mode (traffic is
+// *outbound* from the workload).
+enum WorkloadMode {
+  // Selects for scenarios when the workload is either the
+  // source or destination of the network traffic.
+  CLIENT_AND_SERVER = 0;
```
howardjohn, line 59: Doesn't the default need to be SERVER or CLIENT (I forget which...) to meet current WASM impl?

I would make `0` `UNDEFINED` so each usage can decide the default, and bump CLIENT_AND_SERVER down

type/v1beta1/selector.proto
```diff
@@ -35,3 +35,34 @@ message WorkloadSelector {
...
+
+  // Selects for scenarios when the workload is the
```
howardjohn, line 61: Can we mention which one a gateway is

type/v1beta1/selector.proto
```diff
@@ -35,3 +35,34 @@ message WorkloadSelector {
...
+  uint32 number = 1;
+}
+
+// WorkloadMode allows selection of the role of the underlying workload in
+// network traffic. A workload is considered as acting as a SERVER if it is
+// the destination of the traffic (that is, traffic direction, from the
+// perspective of the workload is *inbound*). If the workload is the source of
+// the network traffic, it is considered to be in CLIENT mode (traffic is
+// *outbound* from the workload).
+// When the workload is a gateway, WorkloadMode will be ignored.
```
howardjohn, line 52: Does this match Telemetry implementation? I thought its considered a server?

ingwonsong: Done to change the description. 

Please note that [source code](https://github.com/istio/istio/blob/69b1e0f7bc04fcc6f32f0eab8c796cfed78b4c02/pilot/pkg/model/telemetry.go#L207) shows that Gateway is mapped to "CLIENT" mode.

howardjohn (approved):

PR #2413, opened 2022-07-12
Validate and fix all samples

howardjohn:
Run `scripts/validate.sh` until it passes. Does need https://github.com/istio/istio/pull/39881 to do this.

Note: it would be great to run this in CI but we have a circular dependency on istio/istio to run `istioctl validate` unfortunately.


ericvn (approved):
LGTM

hzxuzhonghu (approved):

PR #2428, opened 2022-07-23
clarify destination rule docs

ramaraochavali:
Clarify Destination Rule docs that Max Pending Requests/ Max Requests apply to both Http1 and Http2

See https://github.com/istio/istio/issues/31548 and https://github.com/envoyproxy/envoy/pull/9668 (See version history and doc updates)

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -709,11 +709,14 @@ message ConnectionPoolSettings {
 
   // Settings applicable to HTTP1.1/HTTP2/GRPC connections.
   message HTTPSettings {
-    // Maximum number of pending HTTP requests to a destination. Default 2^32-1.
+    // Maximum number of requests that will be queued while waiting for
+    // a ready connection pool connection. Default 1024.
+    // Refer to https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking
+    // under which conditions a new connection is created for HTTP2.
     // Please note that this is applicable to both HTTP/1.1 and HTTP2.
     int32 http1_max_pending_requests = 1;
 
-    // Maximum number of requests to a backend. Default 2^32-1.
+    // Maximum number of active requests to a destination. Default 1024.
     // Please note that this is applicable to both HTTP/1.1 and HTTP2.
     int32 http2_max_requests = 2;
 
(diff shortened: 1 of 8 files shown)


networking/v1alpha3/destination_rule.proto
```diff
@@ -710,9 +710,11 @@ message ConnectionPoolSettings {
   // Settings applicable to HTTP1.1/HTTP2/GRPC connections.
   message HTTPSettings {
     // Maximum number of pending HTTP requests to a destination. Default 2^32-1.
+    // Please note that this is applicable to both HTTP1.1/ and HTTP2.
```
ericvn, line 713: HTTP/1.1 instead of HTTP1.1/ or leave the / out entirely as it's not in HTTP2.

ericvn (approved):

networking/v1alpha3/destination_rule.proto
```diff
@@ -710,9 +710,11 @@ message ConnectionPoolSettings {
   // Settings applicable to HTTP1.1/HTTP2/GRPC connections.
   message HTTPSettings {
     // Maximum number of pending HTTP requests to a destination. Default 2^32-1.
+    // Please note that this is applicable to both HTTP/1.1 and HTTP2.
     int32 http1_max_pending_requests = 1;
 
     // Maximum number of requests to a backend. Default 2^32-1.
+    // Please note that this is applicable to both HTTP/1.1 and HTTP2.
     int32 http2_max_requests = 2;
```
howardjohn, line 718: so is it identical to http1_max_pending_requests then

ramaraochavali: It is not exactly same. http2_max_requests is how many concurrent active requests can be there. max_pending_requests is how many requests can be in pending queue waiting for connection. 

Specifically for pending requests,
"For HTTP/2 connections, if [max concurrent streams](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-max-concurrent-streams) and [max requests per connection](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-max-requests-per-connection) are not configured, all requests will be multiplexed over the same connection so this circuit breaker will only be hit when no connection is already established"

So there is subtle difference on how http2 protocol options are setup.

howardjohn: Should we clarify this difference in our api comments?

ramaraochavali: yeah. good point. Let me update it.

ramaraochavali: updated the docs. Defaults are also incorrect. Fixed them

ramaraochavali:

howardjohn (approved):

PR #2469, opened 2022-08-30
add stateful session load balancer support

ramaraochavali:
API support needed for https://github.com/istio/istio/issues/39740 and https://github.com/istio/istio/issues/37910


networking/v1alpha3/destination_rule.proto
```diff
@@ -444,6 +444,51 @@ message Subset {
...
 
+// Describes a HTTP cookie that will be used for session stickiness.
+// If the cookie is not present, it will be generated.
+message HTTPCookie {
+  // Name of the cookie.
+  string name = 1 [(google.api.field_behavior) = REQUIRED];
+  // Path to set for the cookie.
+  string path = 2;
+  // Lifetime of the cookie.
+  google.protobuf.Duration ttl = 3 [(google.api.field_behavior) = REQUIRED];
```
howardjohn, line 455: should we have a default?

networking/v1alpha3/destination_rule.proto
```diff
@@ -444,6 +444,51 @@ message Subset {
...
+};
+
+// Standard load balancing algorithms that require no tuning.
+enum SimpleLB {
```
howardjohn, line 459: I am not sure moving these out is a backwards compatible change in proto?

ramaraochavali: It is not actually

networking/v1alpha3/destination_rule.proto
ramaraochavali: Sure. We can have.

networking/v1alpha3/destination_rule.proto
```diff
@@ -623,6 +624,19 @@ message LoadBalancerSettings {
     uint64 minimum_ring_size = 4 [deprecated=true];
   };
 
+  // StatefulSession load balancer load balances across backend hosts using 'strong' stickiness. 
+  // This load balancer enforces stable session stickiness by routing requests to the same backend
```
hzxuzhonghu, line 628: In the case one backend instance is down, how will this work? I do not see any explicitly saying about this

networking/v1alpha3/destination_rule.proto
```diff
@@ -623,6 +624,19 @@ message LoadBalancerSettings {
...
+  // consistently till the backend host on which the session lives goes down.
+  message StatefulSession {
+    // Load balancer to use for the initial request, before a session is established.
+    SimpleLB load_balancer = 1;
```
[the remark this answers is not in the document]
hzxuzhonghu: from envoy doc 

```
HTTP session stickiness in Envoy is generally achieved through hash-based load balancing.
```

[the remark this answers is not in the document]
ramaraochavali: That is one of the benefits of using stateful session.  See my comments above the field

and from Envoy docs

The case where a non hash-based load balancer (Random, Round Robin, etc.) is used and session stickiness is still required. If stateful sessions are enabled in this case, requests for new sessions will be routed to the corresponding upstream host based on the result of load balancing. Requests belonging to existing sessions will be routed to the session’s upstream host.

networking/v1alpha3/destination_rule.proto
ramaraochavali: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/stateful_session_filter#how-it-works 

TL;DR; The destination is encoded in the cookie and will be routed irrespective of other nodes going down,

hzxuzhonghu: Not sure i can get the answer,  i am asking if the one that session indicates is down. How does envoy do

hzxuzhonghu: And a further question, what if the instance is down and actually not in the eds?

ramaraochavali: when the instance that hosts the session is down, session is recreated on another node.

networking/v1alpha3/destination_rule.proto
```diff
@@ -444,6 +444,51 @@ message Subset {
   TrafficPolicy traffic_policy = 3;
 }
 
+// Describes a HTTP cookie that will be used for session stickiness.
+// If the cookie is not present, it will be generated.
+message HTTPCookie {
+  // Name of the cookie.
+  string name = 1 [(google.api.field_behavior) = REQUIRED];
+  // Path to set for the cookie.
+  string path = 2;
```
costinm, line 453: I think the main problem is path. 

Normally in envoy session affinity can be set in routes - so map  /example to example.ns.svc, and the path of the cookie will be set to /example 

If another route /example2 is also pointing to example.ns - the path will have to be /example2. 

In Istio, having the path in DR - which is associated with a destination - it is not possible to do that. 

With the Gateway API, sticky session would be attached to a route - and we could generate a working config by setting a per route config.

Please add a lot of docs on this field - with examples and what is not possible ( effectively a destination can only be mapped to a single path ).

This is very important because you may have 
example.com/example -> example.svc - with stickyEndpoint1
example.com/admin -> admin.svc with stickyEndpoint2.

The admin and example can't have same path - and path can't be set to "/" unless there is a single service using sticky sessions.

costinm: I would rather not have path in the API then have it broken and confusing. Without path - we use "/", and it works as expected for normal mesh calls ( as long as user doesn't have routes that redirect to multiple backends ), and is mostly broken for ingress ( only one of the backends for a hostname can be sticky )

networking/v1alpha3/destination_rule.proto
costinm: +1 on a TTL - probably short ( 1h ?)

networking/v1alpha3/destination_rule.proto
costinm: Probably not - but why is it required for this PR ?

networking/v1alpha3/destination_rule.proto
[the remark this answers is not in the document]
costinm: My understanding is that the 2 features are separated - use hash load balancer for affinity, use the cookie for exact routing together with another load balancer.

Using both cookie and hash is probably possible but doesn't make a lot of sense.

networking/v1alpha3/destination_rule.proto
```diff
@@ -631,8 +645,11 @@ message LoadBalancerSettings {
   oneof lb_policy {
     SimpleLB simple = 1;
     ConsistentHashLB consistent_hash = 2;
+    StatefulSession stateful_session = 5;
```
costinm, line 648: I don't think this is correct. I would leave lb_policy as is - and add the stateful_session as a separate option, since it applies to any lb_policy. That doesn't require shuffling protos around and seems safer.

costinm: I would check if consistent_hash is allowed with stateful_session - I don't see why not, AFAIK envoy will use the cookie if present or whatever LB is set.

costinm:
On Wed, Aug 31, 2022 at 2:24 AM Zhonghu Xu ***@***.***> wrote:

>
> In networking/v1alpha3/destination_rule.proto
> <https://github.com/istio/api/pull/2469#discussion_r959364231>:
>
> > @@ -623,6 +624,19 @@ message LoadBalancerSettings {
>      uint64 minimum_ring_size = 4 [deprecated=true];
>    };
>
> +  // StatefulSession load balancer load balances across backend hosts using 'strong' stickiness.
> +  // This load balancer enforces stable session stickiness by routing requests to the same backend
>
> And a further question, what if the instance is down and actually not in
> the eds?
>

I think another important question: what if the instance is up - but not in
EDS ?

Does envoy do any checks to make sure this is not an open proxy - I hope
so, and it needs to be clearly
explained ( and tested !).

networking/v1alpha3/destination_rule.proto
sanjaypujare: In gRPC the way it is proposed in https://github.com/grpc/proposal/blob/e49869d41d91ca5e0ea538c833a1480f90b348a0/A55-stateful-session-affinity.md the "child policy" could be consistent_hash (ring_hash) so if the cookie is not present it will go to the child policy

sanjaypujare:
> Specifically document the difference between consistentHash would be nice
as well

Described in https://github.com/grpc/proposal/blob/e49869d41d91ca5e0ea538c833a1480f90b348a0/A55-stateful-session-affinity.md#rationale

costinm:
IMO we should not mix LB - which selects an endpoint based on different rules, including affinity - with session stickiness, i.e. regardless of LB go to this specific IP:port.

I did more research on how other implementations work - in particular when stickiness is exposed trough gateways or other
infrastructure is involved - I am leaning more towards just "enable sticky sessions" in the API - and maybe the duration, and 
leaving the name of the cookie, path, etc out of the API.  If we integrate with other infra - they may have already defined cookies
and mechanisms to attach them to paths. I don't think use defining a different cookie name for each service is that valuable or needed, and if we get requests we can add it later. And for the path - I would add it after we have a good understanding on how 
it'll work with gateways and routing. 

A smaller PR with just 'enable sticky' and time would move faster and allow us to get feedback - we can add more later, it's harder
to remove.

networking/v1alpha3/destination_rule.proto
costinm: That seems very reasonable - users who want sticky (stateful) sessions probably also want consistent_hash or maglev, and 
that may use a different cookie or parameters.

networking/v1alpha3/destination_rule.proto
costinm: is recreated on another node - using the LB policy in effect. Sticky/stateful just means 'original dest' routing is used
when the cookie with the IP is present, otherwise regular LB takes place.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: Yes, it is not a must to change this. 

You can just add the session sticky in

  oneof lb_policy {

networking/v1beta1/destination_rule.proto
```diff
@@ -572,6 +573,19 @@ message LoadBalancerSettings {
...
   };
 
+  // StatefulSession load balancer load balances across backend hosts using 'strong' stickiness. 
+  // This load balancer enforces stable session stickiness by routing requests to the same backend
+  // consistently till the backend host on which the session lives goes down.
+  message StatefulSession {
+    // Load balancer to use for the initial request, before a session is established.
+    SimpleLB load_balancer = 1;
+    // The session state to use.
+    oneof session_state {
```
hzxuzhonghu, line 583: this is  not required too, add it when we need to support another option would not break compat

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: +1

costinm:
One more question - I've looked at proxyless gRPC and gateway use cases, the client or browser handles the cookie jar.

For 'mesh' use case - is the client-side sidecar holding a cookie jar when making the requests ? Or does the user have to change their code to track cookies ? 

And if the client is gRPC ( with envoy sidecar for security/telemetry/etc - not proxyless ) - does gRPC use a cookie jar already ? 

If you find out - please add it to the comments, in particular if user code needs to track the cookie, I haven't figured it out from envoy docs.

ramaraochavali:
> If you find out - please add it to the comments, in particular if user code needs to track the cookie, I haven't figured it out from envoy docs.

Yes. User code has to track the cookie. And we are supporting this https://github.com/envoyproxy/envoy/issues/22475 in Envoy so that non browser based apps can use this header to track instead of cookie.

ramaraochavali:
>  and maybe the duration, and leaving the name of the cookie

I think leaving path seems OK. But name should be given. We can't assume/standarize the name from some annotations. If name is not given we can derive based on other infrastructure data as you mentioned. But for typical Istio/envoy use case, name is important

ramaraochavali:
@costinm I agree with most of the comments except the name of cookie. If we agree on that, I will delink it from LB policy and introduce statefulSesstion config, if provided, sticky ness would be enabled.

costinm:
Not sure using header instead of cookie helps much - most http clients have
support for cookie tracking, just has to be enabled - and I assume gRPC
will add some easy option if they support sticky sessions for proxyless
gRPC.
Cookie is a header too - so if an app can set a header, they can set a
cookie too.

I remember in Java it was a request parameter - jsessionid - and the cookie
name was also standardized. I'm not saying to hard code - just keep it in
meshConfig or some env variable. I would guess java applications will have
to use the name that is defined in the servlet spec ( or was 20 years ago,
I have a strong deja vue...). Concern is that each service defines a
different cookie and we end up with a mess. It may be a way around path
problems - I'm not completely opposed to customizing the cookie name, just
want to make sure we evaluate pros and cons.

ramaraochavali:
@costinm changed. PTAL.

networking/v1alpha3/destination_rule.proto
```diff
@@ -623,6 +623,23 @@ message LoadBalancerSettings {
     uint64 minimum_ring_size = 4 [deprecated=true];
   };
 
+  // Stateful Session configuration.
```
costinm, line 626: Some more comments here on client changes requirement ( use a cookie jar ).

And maybe some mention this is mainly for service-to-service, the gateway behavior is not yet supported ( since a hostname may route to different backends - we don't have a way to represent this AFAIK, cookie is single IP for single cluster )

networking/v1alpha3/destination_rule.proto
```diff
@@ -643,6 +660,19 @@ message LoadBalancerSettings {
...
+  // a backend host is removed or added, the hash mapping of 1/N client sessions
+  // gets affected (where N is the total number of backends) which would lead to 
+  // dropped requests in some cases which is unacceptable.
+  // StatefulSession enforces session stickiness by routing requests to the same 
+  // backend till the backend host on which the session lives goes down or is not available
+  // in the load balaner pool.
+  // When the backend host that hosts the session goes down, sesion is recreated
+  // on another backend host. A new backend host is selected based on the the configured
+  // load balancer policy.
+  StatefulSession stateful_session_config = 5;
```
costinm, line 675: I am a bit allergic to using 'config', 'policy', 'data' as suffix to fields.
There is nothing on the API that is not a config - or used as policy or data.

networking/v1alpha3/destination_rule.proto
```diff
@@ -623,6 +623,23 @@ message LoadBalancerSettings {
...
+    // SessionCookie is used to maintain the session state across client and server. 
+    message SessionCookie {
+       // Name of the cookie to be used. If not specified, environment specific 
+       // configuration is used.
+       string name = 1; 
+       // Session duration. If not specified, a duration of 1 hour is used.
+       google.protobuf.Duration duration = 2;
+    }
+    // The session state to use.
+    oneof session_state {
```
costinm, line 637: I never understood oneof with a single value or all this nesting in protos.

Just one message - StatefulSession - with 2 fields ( duration, cookie_name ) would be easier to use and less confusing. 
Nothing exclusive about having either cookie or header or request parameter, it is possible to configure more than one.

costinm (commented):
Much better - left few cosmetic/docs comments and waiting for others to review, but I think it's pretty close.

networking/v1alpha3/destination_rule.proto
```diff
@@ -623,6 +623,23 @@ message LoadBalancerSettings {
...
   };
 
+  // Stateful Session configuration.
+  message StatefulSession {
+    // SessionCookie is used to maintain the session state across client and server. 
+    message SessionCookie {
+       // Name of the cookie to be used. If not specified, environment specific 
+       // configuration is used.
+       string name = 1; 
+       // Session duration. If not specified, a duration of 1 hour is used.
```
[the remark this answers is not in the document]
ramaraochavali: you are right. It is duration of cookie.

networking/v1alpha3/destination_rule.proto
ramaraochavali: I wont increase your allergy :-)

networking/v1alpha3/destination_rule.proto
ramaraochavali: > Nothing exclusive about having either cookie or header or request parameter, it is possible to configure more than one.

Envoy does not support more than one https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/stateful_session/v3/stateful_session.proto#envoy-v3-api-field-extensions-filters-http-stateful-session-v3-statefulsessionperroute-stateful-session

costinm: Envoy may not support more than one now - but I wouldn't design a long-term API around limitations on the first implementation. 

Persistent sessions are not new - and it is clear other forms of tracking will need to be added, so better to have the correct API with comments explaining that current version of envoy only supports one. Note that the Istio APIs are not officially 'envoy only',
and we are starting to support the multi-vendor K8S Gateway APIs.

As an extra benefit - it would allow us to avoid the nested message and oneof which are pretty confusing for both devs and users.

costinm (commented):
@howardjohn IMO it may be a good idea to not add this feature in DestinationRule, but to a new 'policy' CR - unlike many other things in DR, persistent session is not really 'host based', a host may route to multiple clusters. 

Current Envoy implementation is inadequate and just works in the most trivial case, but
other vendors may have proper persistent support and Envoy will likely be forced to do the same.
The session cookie should be able to handle tracking the endpoints for multiple clusters, the 
and using multiple cookies or path is unlikely to work as a general solution.  

I don't mind approving this in DR as 'alpha' and to unblock the most basic use case, my only concern is getting again stuck with an experiment becoming 'v1' because we don't really make any distinction.

networking/v1alpha3/destination_rule.proto
ramaraochavali: I am not saying Envoy may not support others. All I am saying the state would be exclusive i.e. either cookie or header or some thing else. But not combination of all. I can drop oneof here but I think nested message would help us to add oneof later if needed rather than spraying fields across.

costinm:
What is the difference between a repeated oneof with 10 fields and a
message with 10 fields ? In yaml it's far cleaner with a message:

persistentSession:
   - cookie:
       - name: JSESSIONID
         duration: 1h
   - header:
       - name: foo


persistentSession:
  cookie: JESSIONID
  header: foo

Besides extra indentation and complexity ( and marshalling overhead ) - you
don't gain much, spraying fields
is far cleaner than spraying messages in hige oneofs ( which end up looking
just like the message ).

costinm:
BTW - it may be best to start with the UX ( including how many levels of
indentation and how much pain the user must deal with),
not with internal proto. The API we expose is yaml/json - we just happen to
use proto internally, but that's not exposed to users.

ramaraochavali:
how do you specify either cookie or header can be set but not both?
```
persistentSession:
  cookie: JESSIONID
  header: foo
```

 Are you saying, we will document that but not enforce via protobuf?

costinm:
I mean: we should not have a convoluted API to enforce something that is
more of an implementation bug than intended behavior.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: I would agree with removing oneof here. And once we need another option, we can add it back, which is compatable

hzxuzhonghu:
@ramaraochavali You can provide a demo usage to be more clear

ramaraochavali:
@costinm One question on this 
```
persistentSession:
  cookie: JESSIONID
  header: foo
```

Will there be ever a case of supporting stateful session persistence with more than one? If not this may be better. Otherwise what you suggested my be better.

persistentSession:
   - cookie:
       - name: JSESSIONID
         duration: 1h
   - header:
       - name: foo

costinm:
Based on my tomcat experience - I can almost guarantee cookie-only is not sufficient,
and certainly not single-value plain-text/base64 IP address.  We keep thinking of this as a direct service-to-single-service API - but gateways are a far bigger user of persistent sessions. 
It is a good start - but if you look at other load balancers supporting persistent sessions it's more refined.

I think we should keep the API as flexible as possible - including default values, i.e.
most users should just say:

 persistentSession: {}

Since persistent sessions may need to span multiple components ( in particular when 
exposed on ingress gateway ) it is far better for most users ( persona == namespace admin/app owner) to not mess with how persistentSession is handled. The actual infrastructure may also include external (non-Envoy) LBs  and backends implemented in Java that may have their own mechanisms (HttpSession) - and it may be better for cluster admin who is aware of all infra to fine tune this. 

We can keep the cookie name/header - users may want to override, but real persistent session across multiple sticky backends  ( i.e. the common case when a browser is accessing via ingress) may not interact well with overrides.

How about keeping it as simple as possible - we can add header support on a separate PR, just keep 

persistentSession:
  - cookie:
      name: foo
      maxAge: 1h # using the cookie terminology to be more clear it's not session duration

If you think header is important - we can also add it, but with docs that in current 
envoy impl only one of header and cookie can be used ( ingress gateways other than
envoy may support both, proxyless gRPC too). 

And I would add a MeshConfig "defaultPersistentCookie" section for mesh admins to set the defaults, and include a default value in the code, so most user can just declare intent to use  ` persistentCookie: {} ` without worry about the details. I would not use JSESSIONID as default because it is typically an encrypted or opaque blob.

costinm:
BTW - https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/ has nginx config - https://kubernetes.github.io/ingress-nginx/examples/affinity/cookie/ the equivalent API when nginx is used as ingress, https://www.haproxy.com/blog/load-balancing-affinity-persistence-sticky-sessions-what-you-need-to-know/ and https://doc.traefik.io/traefik/routing/services/#sticky-sessions for others.

sanjaypujare:
@va7ish you might be interested in this API

ramaraochavali:
> How about keeping it as simple as possible - we can add header support on a separate PR, just keep

persistentSession:

cookie:
name: foo
maxAge: 1h # using the cookie terminology to be more clear it's not session duration

I agree with this.  Changed. PTAL

mesh/v1alpha1/config.proto
```diff
@@ -1121,8 +1121,11 @@ message MeshConfig {
   // Configuration of mTLS for traffic between workloads within the mesh.
   TLSConfig mesh_mTLS = 63;
 
+  // Default persistent cookie to be used for configuring persistent sessions.
+  istio.networking.v1alpha3.PersistentCookie default_persistent_cookie = 64;
```
hzxuzhonghu, line 1125: Donot think the global config makes much sense, this option is per service

networking/v1alpha3/destination_rule.proto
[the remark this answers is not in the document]
va7ish: @costinm statefull affinity feature is orthogonal to LB choice. Theoretically it should be usable with any LB, including the hash-based algorithms. This is because hash-based LB algorithms only provide "soft-affinity", I.e. when the backend set changes, at least 1/N of the sessions will land on new backend hosts which will break affinity.

networking/v1alpha3/destination_rule.proto
va7ish: +1

va7ish:
> @howardjohn IMO it may be a good idea to not add this feature in DestinationRule, but to a new 'policy' CR - unlike many other things in DR, persistent session is not really 'host based', a host may route to multiple clusters.

I agree with @costinm. It seems that if this is done at the DR level it will break the canary functionality. We need to maintain affinity across clusters. Not sure if Envoy supports that in the currently.

costinm:
> > @howardjohn IMO it may be a good idea to not add this feature in DestinationRule, but to a new 'policy' CR - unlike many other things in DR, persistent session is not really 'host based', a host may route to multiple clusters.
> 
> I agree with @costinm. It seems that if this is done at the DR level it will break the canary functionality. We need to maintain affinity across clusters. Not sure if Envoy supports that in the currently.

I don't think it breaks canary - anything that has persistent session should keep going to the sticky backend, canary will apply 
to new sessions. 

And we can define a new 'policy' - possibly for the Gateway/GAMMA API - in addition to adding it in DR, we should not 
block on that.

costinm (commented):
Looks good to me, ready to approve - waiting for more people to check it.

mesh/v1alpha1/config.proto
costinm: Yes, this is just a global option to set the default values for users enabling persistent cookie without specifying a cookie name. 

meshconfig - mesh admin:

    default_persistent_cookie: 
      cookie_name: foo
      maxAge: 1h 

destinationRule - user:

    persistent_session: {} # use defaults for the mesh

This way the admin can figure out what cookie to use - for example if they use some other infra or load balancers
( ingress using nginx. etc). It is a bit strange - the value of the cookie my not match, but this may evolve, at some point
envoy will need to support more flexible format for the cookie value, not only the raw IP.

networking/v1alpha3/destination_rule.proto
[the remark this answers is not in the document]
costinm: I understand - my point was that there is little reason to use hash-based affinity when session is enabled ( since it provides even stronger stickiness). It is better to combine 'stateful session' with load-based or random or one of the other LBs. 
Hash would work - just don't think it's a good choice.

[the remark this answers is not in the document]
va7ish: ok, makes sense.

ramaraochavali:
@howardjohn Any concerns with this?

mesh/v1alpha1/config.proto
howardjohn: Is there going to be a default default cookie?

networking/v1alpha3/destination_rule.proto
```diff
@@ -1127,6 +1148,15 @@ message ClientTLSSettings {
   google.protobuf.BoolValue insecure_skip_verify = 8;
 }
 
+// PersistentCookie is used to maintain the session state across client and server. 
+message PersistentCookie {
+  // Name of the cookie to be used. If not specified, environment specific
```
howardjohn, line 1153: isn't it "mesh config default is used"?

networking/v1beta1/destination_rule.proto
```diff
@@ -592,6 +600,19 @@ message LoadBalancerSettings {
...
+  // a backend host is removed or added, the hash mapping of 1/N client sessions
+  // gets affected (where N is the total number of backends) which would lead to 
+  // dropped requests in some cases which is unacceptable.
+  // PersistentSession enforces session stickiness by routing requests to the same 
+  // backend till the backend host on which the session lives goes down or is not available
+  // in the load balaner pool.
+  // When the backend host that hosts the session goes down, sesion is recreated
+  // on another backend host. A new backend host is selected based on the the configured
+  // load balancer policy.
+  PersistentSession persistent_session = 5;
```
howardjohn, line 615: needs a comment it is HTTP only. And possibly indicate it exposes internal details in headers which users may not find appropriate?

howardjohn (commented):
Isn't it kind of broken without path...?

For example, if we have 50/50% weighting to v1 and v2

1. Randomly hit v1 cluster, I get cookie for 1.2.3.4 (part of v1)
2. Randomly hit v2 cluster. There is no 1.2.3.4, so I pick a random host and cookie is set to 2.3.4.5 (part of v2)
3. Randomly hit v1. There is no cookie for 2.3.4.5, so I pick a random host...
4. Repeat

networking/v1beta1/destination_rule.proto
costinm: I suspect with ambient/HBONE it may not have to be HTTP only. I was planning to use it as part of the east/west multi-network gateway improvements I mentioned in one of the docs. 

We should mention that current implementation of headers has some limitations and should not be exposed on the internet or on gateways - but this is the API, envoy ( and proxyless ,etc) will need to evolve the implementation.

networking/v1alpha3/destination_rule.proto
costinm: mesh config, and probably we should define a default one if mesh config is not set ( but please no 'istio' or 'envoy' in the name...)

costinm:
> Isn't it kind of broken without path...?
> 
> For example, if we have 50/50% weighting to v1 and v2
> 
> 1. Randomly hit v1 cluster, I get cookie for 1.2.3.4 (part of v1)
> 2. Randomly hit v2 cluster. There is no 1.2.3.4, so I pick a random host and cookie is set to 2.3.4.5 (part of v2)
> 3. Randomly hit v1. There is no cookie for 2.3.4.5, so I pick a random host...
> 4. Repeat

Yes, it is pretty broken in other ways as well - as implementation in envoy. The API however seems fine, and the impl can
be fixed in time. 

For example the cookie can encode a list (cluster1=IP1,cluster2=IP2), or it can be combined with external authz to 
support encrypted cookies or cookies as IDs and lookup appropriate backend dynamically, for example in cases where we authenticate and send the session to a server where user data is stored. 

I think ext authz is great for 'composing' with the persistent session, I believe it can inject or modify headers including cookie
and support most of the advanced use cases.

networking/v1beta1/destination_rule.proto
howardjohn: using this in out ambient layer has pretty big impacts on the implementation of HBONE. While its possibly Envoy would need tons of changes to make this work

howardjohn:
IMO we should have an idea of a non-broken Envoy implementation before moving forward with an API

sanjaypujare:
> Isn't it kind of broken without path...?
> 
> For example, if we have 50/50% weighting to v1 and v2
> 
> 1. Randomly hit v1 cluster, I get cookie for 1.2.3.4 (part of v1)
> 2. Randomly hit v2 cluster. There is no 1.2.3.4, so I pick a random host and cookie is set to 2.3.4.5 (part of v2)
> 3. Randomly hit v1. There is no cookie for 2.3.4.5, so I pick a random host...
> 4. Repeat

Not sure I understand "hit the v2 cluster" part. If a cookie is present then the RPC is sent to that IP address (skipping cluster selection). This is how both Envoy and proxyless gRPC will behave

howardjohn:
> > Isn't it kind of broken without path...?
> > For example, if we have 50/50% weighting to v1 and v2
> > 
> > 1. Randomly hit v1 cluster, I get cookie for 1.2.3.4 (part of v1)
> > 2. Randomly hit v2 cluster. There is no 1.2.3.4, so I pick a random host and cookie is set to 2.3.4.5 (part of v2)
> > 3. Randomly hit v1. There is no cookie for 2.3.4.5, so I pick a random host...
> > 4. Repeat
> 
> Not sure I understand "hit the v2 cluster" part. If a cookie is present then the RPC is sent to that IP address (skipping cluster selection). This is how both Envoy and proxyless gRPC will behave

How can it skip cluster selection? That has more than IP - has TLS settings, etc

sanjaypujare:
> ...
> > Not sure I understand "hit the v2 cluster" part. If a cookie is present then the RPC is sent to that IP address (skipping cluster selection). This is how both Envoy and proxyless gRPC will behave
> 
> How can it skip cluster selection? That has more than IP - has TLS settings, etc

This is the design in Envoy https://docs.google.com/document/d/1IU4b76AgOXijNa4sew1gfBfSiOMbZNiEt5Dhis8QpYg/edit#heading=h.x3i1k73a3v8z

"When a new request arrives, the router filter tries to resolve the upstream host address from the cookie.
The Load Balancer gets this address through the LoadBalancerContext (requires extending the LoadBalancerContext API) and prefers this address."

That means it skips cluster selection? That is the whole point of cookie based stateful session affinity.

howardjohn:
AFAIK that is to pick the host _within_ the cluster _after_ picking a cluster

sanjaypujare:
> AFAIK that is to pick the host _within_ the cluster _after_ picking a cluster

Hmmm, interesting. @htuch could you throw some light on this or CC someone who can?

htuch:
Yes, only within the picked cluster (and only amongst valid hosts there). @wbpcode can also speak to this work more directly.

wbpcode:
> AFAIK that is to pick the host _within_ the cluster _after_ picking a cluster

Yeah, this stateful session could select host cross different clusters. So, this function is conflict with weighted cluster.

networking/v1beta1/destination_rule.proto
costinm: I don't think it's that bad - but will need some changes, of course. My point was that the HBone protocol is intended for this kind of use cases - where a TCP protocol is augmented with metadata like the cookie and can support persistence and 'original dst' across one or more hops. 

Except for the initial capture and final local delivery - the traffic in the hbone mesh is H2 and can use most features.

costinm:
> IMO we should have an idea of a non-broken Envoy implementation before moving forward with an API

For the most common use case - persistent session for a single Service/Cluster - it works. That's what 99% of users 
need for the basic use case. 

We can document that traffic split, ingres, etc are not currently supported by the API ( with envoy ), but I don't think
there is anything wrong in the API - or supporting the basic case.

costinm:
> > ...
> > > Not sure I understand "hit the v2 cluster" part. If a cookie is present then the RPC is sent to that IP address (skipping cluster selection). This is how both Envoy and proxyless gRPC will behave
> > 
> > 
> > How can it skip cluster selection? That has more than IP - has TLS settings, etc
> 
> This is the design in Envoy https://docs.google.com/document/d/1IU4b76AgOXijNa4sew1gfBfSiOMbZNiEt5Dhis8QpYg/edit#heading=h.x3i1k73a3v8z
> 
> "When a new request arrives, the router filter tries to resolve the upstream host address from the cookie. The Load Balancer gets this address through the LoadBalancerContext (requires extending the LoadBalancerContext API) and prefers this address."
> 
> That means it skips cluster selection? That is the whole point of cookie based stateful session affinity.

It can't skip cluster selection - that's where mTLS checks are made. 

But I'm concerned if it is restricted to IPs in the Cluster current set - it'll break draining.

costinm:
Maybe a different way to look at this API is: it enables persistent session support, as defined by the data plane implementation.
It is no different from adding the envoy filter using EnvoyFilter. 

How envoy will evolve the implementation - or how proxyless gRPC will implement this, or other gateways that use proper 
cookies and cross-cluster tracking is not specified in the API - we can only document the current limitations of the impl.

sanjaypujare:
> ...
> How envoy will evolve the implementation - or how proxyless gRPC will implement this, or other gateways that use proper cookies and cross-cluster tracking is not specified in the API - we can only document the current limitations of the impl.

I agree. Also to go back to the cluster selection: the selected cluster will be different only if the routing config changed in the middle of a session or they are using weighted cluster. We can have the caveat that cookie based session affinity will not work if any of those things are present (routing config change in the middle of sessions and weighted clusters).

CC @yanavlasov

howardjohn:
> I agree. Also to go back to the cluster selection: the selected cluster will be different only if the routing config changed in the middle of a session or they are using weighted cluster. We can have the caveat that cookie based session affinity will not work if any of those things are present (routing config change in the middle of sessions and weighted clusters).

Or they just have `/foo -> cluster foo, /bar -> cluster bar` and send a different request on the next call.

wbpcode:
IMO, the stateful session is more like supplement to load balancing. Just like hash policy to consistent hash.
So, would it be better to make this API be part of VS or separated policy?

costinm:
Cookie prefix doesn't solve much - but setting different cookie for each
cluster might.

Eventually we'll need to improve envoy to match the rest of the load
balancers support persistent sessions - but we can still make progress for
the little that
is supported.

wbpcode:
May be it possible to record the cluster in the future. I will have a try. 🤔

costinm:
Certainly not VS - it is specific to clusters, not routes.

I think we agreed that long term we should have a separate policy, part of
Gateway and GAMMA model - ideally not vendor specific ( but it can start as
a vendor ).
However adding something in DR - as proposed in this PR - is a good start
and works for most common cases.

There is much more - this model of indicating a specific IP while using the
cluster settings may also be used to improve stateful sets, which are
another
form of persistent session ( even harder - user explicitly targets 5.svc.ns
), and we'll need to fix it to work with E/W multi-network GW, etc.

wbpcode:
> Certainly not VS - it is specific to clusters, not routes.

Emmm, but in current implementation of envoy, the stateful session is specific for routes/vhs/listeners, not for clusters.

May it could be specific to upstream clusters in future after the upstream filters of envoy is completed. But for now, it's function for downstream.

costinm:
Envoy does define it per route, not clear why - but obviously can't work
with path prefix unless all the routes in envoy are path-prefix, which is
not the case.
And the current implementation is very unlikely to pass a security/privacy
review for ingress use cases, so I think it is reasonable to assume envoy
will
need to make quite a bit of change to the implementation, and not tie the
API to the initial impl.

Routing based on the cookie or IP override is also clearly a property that
needs to be enabled/controlled at the cluster level, and typically the
server (backend or proxy) generates the cookie - it may not be an envoy.

wbpcode:
> Envoy does define it per route, not clear why

Because it's designed as a function that to overwrite upstream load balancing result based on the downstream attributes and decision. And we want this function to be **protocol independent and extendable**. 
The upstream module doesn't care about which downstream attributes (HTTP headers/cookies, thrift headers, dubbo method, etc.) are used. It just cares about the final override address.
So, most of the config is specified in the downstream (routes/vhs/listeners). 
It's hard to change this core design if there is no better design or convincing reason. (May be the new `upstream filters` can help, but I haven't to try it.)

> but obviously can't work with path prefix unless all the routes in envoy are path-prefix, which is not the case.

I didn't see this restriction. But the cookie base stateful session is indeed just the simplest implementation to ensure the above mechanism can work as expected.
On the one hand, we can optimize the implementation of this cookie-based extension (encryption support, etc.), on the other hand, it would be better to make the API extendable to make sure we can add stateful session support that implemented by other ways.

va7ish:
> Certainly not VS - it is specific to clusters, not routes. I think we agreed that long term we should have a separate policy, part of Gateway and GAMMA model - ideally not vendor specific ( but it can start as a vendor ). However adding something in DR - as proposed in this PR - is a good start and works for most common cases. There is much more - this model of indicating a specific IP while using the cluster settings may also be used to improve stateful sets, which are another form of persistent session ( even harder - user explicitly targets 5.svc.ns ), and we'll need to fix it to work with E/W multi-network GW, etc.

It seems not right to me. Generally speaking the notion of session affinity across an infra node N1 (Envoy or anything else) needs to maintain such affinity through the entire set of "layers" (routing, LB, etc). Such that to guarantee that the particular session (how it would be uniquely identified and persistent is a separate point) gets routed and load balanced to the same destination host regardless of any branches in the decision path in the node's N1 logic. Thus such configuration should probably appear at the top most configuration layer (i.e. VS).

costinm:
I think you have a good point - most existing implementations of persistent
sessions are not tied to a single 'cluster'.

The reason it was added to DR seems to be that we traditionally configure
LB features in DR - and the cookie and
IP is specific to a cluster. If a VirtualService has multiple routes, with
multiple Services and subsets - defining a
'user persistent session' will not work well with current Envoy impl, the
cookie has the IP of a specific cluster only.

What other proxies do - I think haproxy ? - is use a Service annotation,
which is an option and could cover all subsets
as well - but I'm not sure it will work with Envoy impl. That was my first
proposal for this ( in the grpc-only PR ).

Long term we need a way to define persistent sessions across multiple
clusters/backends - so it may make sense
to just define in VirtualService that the entire host is using persistent
session and what cookie, and have a change
in envoy to have a list of cluster:IP pairs. Even in this case - we still
need to know which of the clusters need persistent
sessions and which don't.

costinm:
On Wed, Sep 14, 2022 at 9:09 AM code ***@***.***> wrote:

> Envoy does define it per route, not clear why
>
> Because it's designed as a function that to overwrite upstream load
> balancing result based on the downstream attributes and decision. And we
> want this function to be *protocol independent and extendable*.
>
I don't know what this means - cookies are pretty HTTP specific, and all
other load balancers support persistent sessions in a pretty consistent
way.

Decision is clearly based on downstream attributes (presence and value of
the cookie) - but it is also based on cluster (Service) properties.
Depending on which cluster the requests gets directed to - we may use
persistent or not, it's a property of the backend.


> The upstream module doesn't care about which downstream attributes (HTTP
> headers/cookies, thrift headers, dubbo method, etc.) are used. It just
> cares about the final override address.
>
Upstream module in envoy doesn't care - but it's a property of the service
( backend ) if it is designed for persistent sessions or not.
For example in Java persistent session is enabled in user code, by calling
getHttpSession(). In K8S typically stateful sets are
used when persistent sessions are used - this feature adds a slightly
different way to do the same.



> So, most of the config is specified in the downstream
> (routes/vhs/listeners).
> It's hard to change this core design if there is no better design or
> convincing reason. (May be the new upstream filters can help, but I
> haven't to try it.)
>
I don't think having the config in envoy in the downstream is a major
problem.

If Istio knows a particular Service expects session affinity - it can
configure the relevant listeners/routes/etc.

With current envoy impl - the only way I know to have a persistent session
across multiple clusters seems to be to have
each cluster use a different cookie name.



> but obviously can't work with path prefix unless all the routes in envoy
> are path-prefix, which is not the case.
>
> I didn't see this restriction. But the cookie base stateful session is
> indeed just the simplest implementation to ensure the above mechanism can
> work as expected.
> On the one hand, we can optimize the implementation of this cookie-based
> extension (encryption support, etc.), on the other hand, it would be better
> to make the API extendable to make sure we can add stateful session support
> that implemented by other ways.
>

Yes - and using external authz it is already mostly possible to do this -
ext authz can check a cookie, decrypt and set the per cluster cookie as
needed. I'm not saying it's impossible - what we have is a great first
step, but we should not design the Istio API around how it is implemented
now by envoy.

This discussion is also in large part around how proxyless gRPC is going to
implement the same feature - and how we can move
this API into K8S Gateway as a vendor-neutral API ( i.e. Nginx, haproxy and
others implementing the common API )

sanjaypujare:
@ramaraochavali  is there an ETA for when all the outstanding issues will be resolved and this will be merged?

va7ish:
> With current envoy impl - the only way I know to have a persistent session across multiple clusters seems to be to have each cluster use a different cookie name.

As far as I understand this would still not work correctly when using weight based routing. This is because when a cookie set for Cluster C1 ends up later being routed to C2 (based on weight based probability) that cookie will be ignored and a new cookie would be created with a different mapping to an endpoint which is a member of C2. Correct?

costinm:
On Wed, Sep 14, 2022 at 4:05 PM Israel Shapiro ***@***.***>
wrote:

> With current envoy impl - the only way I know to have a persistent session
> across multiple clusters seems to be to have each cluster use a different
> cookie name.
>
> As far as I understand this would still not work correctly when using
> weight based routing. This is because when a cookie set for Cluster C1 ends
> up later being routed to C2 (based on weight based probability) that cookie
> will be ignored and a new cookie would be created with a different mapping
> to an endpoint which is a member of C2. Correct?
>

Yes, that's even more difficult to support. Not sure how we can express
that a set of Clusters are 'related' and part of the same Service -
the stateful application is the same even if we implement this as multiple
envoy clusters. Maybe we can add some label to all
clusters belonging to a service - and see if envoy can change the code so
the cookie is applied across clusters based on the label ?
It certainly needs envoy changes.

Proxyless gRPC clients likely have more flexibility here.

wbpcode:
cc @costinm 

> I don't know what this means - cookies are pretty HTTP specific, and all
other load balancers support persistent sessions in a pretty consistent
way.

It's hard to explain everything clearly in short message. There are some inner details of envoy which I think the control plane shouldn't cares about.

In conclusion, the mechanism of upstream address overriding is hard to be changed, so main config should be specified in downstream (routes/vhs/listeners) for envoy.
But as an extension, the cookie based impl is easy to be optimized to support some new reasonable features (Encryption, cluster record, etc.).

wbpcode:
> I don't think having the config in envoy in the downstream is a major
problem.
> If Istio knows a particular Service expects session affinity - it can
configure the relevant listeners/routes/etc.

IMO, this would create a strong dependency between rds and cds, which I don't think is a good idea.

costinm:
I think it's a mistake to design Istio API around how one (pretty new and
strange) implementation in Envoy happens to work.

In K8S, there is first class support for (very) session affinity via
StatefulSets. As a user, if you deploy an app as stateful set
you are guaranteed distinct DNS entries that allows each individual
workload to be directly used. In addition few apps
implement strong session affinity by discovering the endpoints and using
them directly.

This feature provides a different way - maybe better, since it can be
combined with a load balancer if the session is
not used. But ultimately the developer still deploys a specific app  that
requires sticky sessions -
and this is typically one K8S Service and one or more Deployments. I don't
think it is very common for user to
even use routes for this service, i.e. to redirect requests to
statefulsvc.ns.svc to other services by path or other things.
With StatefulSets it would be quite hard, since each instance has its own
hostname. If someone has use cases
where requests for stateful.ns.svc would need to be routed to other
services and how routing will work - we can
address with a different API.

Other load balancers allow and use this pattern without any problems, so we
know the API is valid and implementation
possible.

For service-to-service without subsets - what Envoy implements now is
sufficient to implement this pattern. From the user
perspective, this is a  property of the Service - making it behave closer
to a StatefulSet. From Istio perspective,
we set this mode for the Service - which we usually configure with
DestinationRule ( or my proposed Service annotation,
as others are doing). Internally in envoy - we do config it as a http
filter, but users don't need to worry about this.

For ingress, where routing may be a concern - this is nowhere close to
being used due to privacy/security and many
 other reasons, so we shouldn't worry about it for now.

Main problem remains how to handle subsets/traffic shifting - if there is a
strong requirement for the IP in the cookie to be
present in the EDS response for that cluster. We apply the setting as a
http filter - so if Envoy must verify the IP is
part of the EDS, it could maybe check for all clusters that are targets ? I
don't understand why this check would be needed -
at least in Istio all security is based on TLS certificates and a malicious
IP wouldn't be a problem.

wbpcode:
> From Istio perspective,
we set this mode for the Service - which we usually configure with
DestinationRule ( or my proposed Service annotation,
as others are doing). Internally in envoy - we do config it as a http
filter, but users don't need to worry about this.

Yeah, I agree the users don't need to worry that. 

If the control plane ensures that we have no route-level requirement, I think a separated policy would be better than DR here to avoid the strong dependency to the DR. (where can get the context about the service annotation?).

wbpcode:
> Main problem remains how to handle subsets/traffic shifting - if there is a
strong requirement for the IP in the cookie to be
present in the EDS response for that cluster. We apply the setting as a
http filter - so if Envoy must verify the IP is
part of the EDS, it could maybe check for all clusters that are targets ? I
don't understand why this check would be needed -
at least in Istio all security is based on TLS certificates and a malicious
IP wouldn't be a problem.

A host belongs to a cluster and is managed by that cluster. And the same IP may be considered as different hosts that exist in multiple clusters. Therefore, there is no global table of hosts that can be searched by IP/hostname in the envoy. Moreover, it is too expensive to search all clusters one by one.

It's easier to just record the cluster in the cookie and try to overwrite the cluster of the route by this record.

va7ish:
> I think it's a mistake to design Istio API around how one (pretty new and strange) implementation in Envoy happens to work.


I agree with you on this hence I think this needs to be thought through considering the longterm clean approach. 

> In K8S, there is first class support for (very) session affinity via StatefulSets. As a user, if you deploy an app as stateful set you are guaranteed distinct DNS entries that allows each individual workload to be directly used. In addition few apps implement strong session affinity by discovering the endpoints and using them directly.

I don't think these are comparable. What you described for existing functionality in K8s StatefulSet is actually not really session affinity but rather C->S affinity. In other words this I a very coarse affinity managed entirely by implementation logic within the client application code. The client must be made aware of the full set of server endpoints and must connect to each endpoint individually. The client must also directly discover the full set of endpoints at every point in time (as the service scales down/up). Beyond that the negative caching in DNS prevents the clients from accessing newly created pods quickly.  Load balancing and other service mesh functionality is not feasible with such approach either.

ramaraochavali:
> @ramaraochavali is there an ETA for when all the outstanding issues will be resolved and this will be merged?

@sanjaypujare Sorry. I am down with Viral fever and not able to keep track of all the discussion going on here. I will spend time understanding what needs to change based on all these discussions and make changes accordingly.

@costinm thank you for driving all the discussion.  I will take a look at the discussion again but any thing need to change to get this going?

costinm:
On Wed, Sep 14, 2022 at 10:09 PM Israel Shapiro ***@***.***>
wrote:

> I think it's a mistake to design Istio API around how one (pretty new and
> strange) implementation in Envoy happens to work.
>
> I agree with you on this hence I think this needs to be thought through
> considering the longterm clean approach.
>
> In K8S, there is first class support for (very) session affinity via
> StatefulSets. As a user, if you deploy an app as stateful set you are
> guaranteed distinct DNS entries that allows each individual workload to be
> directly used. In addition few apps implement strong session affinity by
> discovering the endpoints and using them directly.
>
> I don't think these are comparable. What you described for existing
> functionality in K8s StatefulSet is actually not really session affinity
> but rather C->S affinity. In other words this I a very coarse affinity
> managed entirely by implementation logic within the client application
> code. The client must be made aware of the full set of server endpoints and
> must connect to each endpoint individually. The client must also directly
> discover the full set of endpoints at every point in time (as the service
> scales down/up). Beyond that the negative caching in DNS prevents the
> clients from accessing newly created pods quickly. Load balancing and other
> service mesh functionality is not feasible with such approach either.
>

You are right, the 3 variants are very different - in particular around how
much implementation logic is needed on the client side - but all 3 solve a
very similar problem,
client wants to control the specific instance it talks to and to bypass the
load balancer. For StatefulSet - client does this by using the individual
DNS entry, for 'custom discovery'
the client watches APIserver and gets the IPs directly, and with the new
option it keeps a cookie.

The API (and feature) we are discussing in this PR has many benefits
compared to the other 2 - it is simpler for client, works with a lot of
non-mesh internet infrastructure ( cookies
are how almost everyone has implemented sticky sessions for >20 years :-) -
and is far better match for Istio model. With StatefulSet we pay a huge
price tracking and configuring
the IPs and with the client-discovery we are forced to use original dst and
lose most of the features. In addition this is the only option that will
work with multi-network and E/W gateways,
so it's pretty important to get it right.

costinm:
On Wed, Sep 14, 2022 at 10:08 PM code ***@***.***> wrote:

> Main problem remains how to handle subsets/traffic shifting - if there is a
> strong requirement for the IP in the cookie to be
> present in the EDS response for that cluster. We apply the setting as a
> http filter - so if Envoy must verify the IP is
> part of the EDS, it could maybe check for all clusters that are targets ? I
> don't understand why this check would be needed -
> at least in Istio all security is based on TLS certificates and a malicious
> IP wouldn't be a problem.
>
> A host belongs to a cluster and is managed by that cluster. And the same
> IP may be considered as different hosts that exist in multiple clusters.
> Therefore, there is no global table of hosts that can be searched by
> IP/hostname in the envoy. Moreover, it is too expensive to search all
> clusters one by one.
>

A host ( K8S Service ) is split between multiple clusters - by subsets and
other means, and the mapping is not fixed, it may change when new versions
are deployed.

I was not suggesting search one by one - a hash map would do, and given how
much redundant info is passed in XDS - it would barely be visible.
Obviously even better is to skip this step and not tie this feature to EDS.
Any reason persistent sessions should not work with DNS clusters
for example ?



> It's easier to just record the cluster in the cookie and try to overwrite
> the cluster of the route by this record.
>

Can we at least have an option to allow the same behavior with the
"original dest" clusters - where the header is not checked
against any EDS ? I still don't understand why looking up the IP is needed
- as I mentioned it has no security benefits (we
check the certificate of the server against Cluster-specified SAN), breaks
any XDS server ability to use subsetting for
large clusters - and breaks traffic shifting and other LB features that use
multiple clusters. Plus plenty of corner cases
on draining and other issues - but I haven't heard a single benefit, except
maybe in cases where mTLS is not used
for security - but EDS having an IP is not really the right way to secure,
anyone can put arbitrary IPs in the EDS response
by creating Endpoints or ServiceEntries - and at least in sidecar case (we
won't enable it for ingress) the app can simply
directly connect to the IP, doesn't need to forge a cookie.

wbpcode:
> A host ( K8S Service ) is split between multiple clusters - by subsets and
other means, and the mapping is not fixed, it may change when new versions
are deployed.

What I mean here by `host` is an instance/IP, not the k8s domain host.

costinm:
On Thu, Sep 15, 2022 at 10:55 AM code ***@***.***> wrote:

> I am not saying it's impossible.
> But for now, in the envoy, we must select a cluster first, and then select
> host (abstraction of service instance/IP) from this cluster. The selected
> host/IP must be managed by the cluster. It's designed as this type to
> manage all upstream resource.
> And for now, only the orignal dst cluster is allowed to create new host
> dynamically based on the IP.
>
> If some cross-cluster functions are necessary, we can record the cluster,
> and find the cluster before the load balancing.
>
The problem is that the IP is not forever tied to a cluster - users can
modify 'canary' and 'prod' subsets (== clusters) to select a different set
of workloads, and the 'canary' and 'prod'
cluster will have different IPs. A workload may start in the 'canary'
cluster, later be mapped to 'prod'. This is the simplest example - there
are more complicated ones.

From user perspective - it's the same k8s service, but we map it
dynamically to clusters.

While the cluster is managing the connections - there was never a
requirement that a cluster knows ALL the IPs. For example DNS cluster
clearly can't, and
for very large services spanning multiple clusters/regions - it is
reasonable to expect XDS servers to not return ALL the endpoints to ALL the
workloads, but subset.
We don't do it now - but I wouldn't want to have an API that prevents us
from doing it in the future.

wbpcode:
> The problem is that the IP is not forever tied to a cluster

Yeah, this is a problem that cannot be solved for now.

sanjaypujare:
> @sanjaypujare Sorry. I am down with Viral fever and not able to keep track of all the discussion going on here. I will spend time understanding what needs to change based on all these discussions and make changes accordingly.
> 

oops, sorry. I hope you feel better soon. Thanks for the update.

costinm:
On Thu, Sep 15, 2022 at 11:48 AM code ***@***.***> wrote:

> The problem is that the IP is not forever tied to a cluster
>
> Yeah, this is a problem that cannot be solved for now.
>
Can't be solved with Envoy - but I don't think this should block the API,
and proxyless gRPC implementation may be able to solve it,
as well as in future with Envoy ( or other implementations of Istio/Gateway
API).

Unless there are other major objections or blocking comments - I think we
cover most possible limitations, we'll need to document
them but IMO we should move ahead with this PR and focus on how to
implement it...

wbpcode:
> Can't be solved with Envoy - but I don't think this should block the API,
and proxyless gRPC implementation may be able to solve it,
as well as in future with Envoy ( or other implementations of Istio/Gateway
API).

👍Agree this detail shouldn't block this PR. Although I still worry about this DR dedicated API.

costinm (approved):

ramaraochavali:
@howardjohn updated docs. Please let me know if there are there are any other concerns.

howardjohn:
There was a LOT of discussion here so I may have missed some nuance, but it sounds like our proposal is to merge the API as-is, despite it not working in Envoy and having no concrete plan to make it work in Envoy. Is that accurate?

If so, I don't think that is the right approach. These APIs last quite a long time. I am not sure how we can show this is the right design if it doesn't actually work today and we don't have plans to make it work.

I believe the motivation here is proxyless gRPC, with the argument that our APIs should not be Envoy specific. I agree with that, but the APIs need to work on Envoy, we just want them to _also_ work outside of Envoy -- "not envoy specific" doesn't mean it is acceptable to not work on Envoy, which is used by 99.9% of users.

If there is desired to get this working on the experimental proxyless gRPC area, perhaps we need an experimental proxyless specific API - or use existing experimental options like annotations and environment variables?

I really don't feel great about merging an API to DestinationRule that only works in a very small number of scenarios.

sanjaypujare:
> There was a LOT of discussion here so I may have missed some nuance, but it sounds like our proposal is to merge the API as-is, despite it not working in Envoy and having no concrete plan to make it work in Envoy. Is that accurate?

My understanding is that it *does* work in Envoy except the case of weighted clusters or other cases involving multiple clusters for a service. 

> 
> If so, I don't think that is the right approach. These APIs last quite a long time. I am not sure how we can show this is the right design if it doesn't actually work today and we don't have plans to make it work.
> 
> I believe the motivation here is proxyless gRPC, with the argument that our APIs should not be Envoy specific. I agree with that, but the APIs need to work on Envoy, we just want them to _also_ work outside of Envoy -- "not envoy specific" doesn't mean it is acceptable to not work on Envoy, which is used by 99.9% of users.
> 
> If there is desired to get this working on the experimental proxyless gRPC area, perhaps we need an experimental proxyless specific API - or use existing experimental options like annotations and environment variables?
> 
> I really don't feel great about merging an API to DestinationRule that only works in a very small number of scenarios.

About "a very small number of scenarios" , I am not sure; but @costinm @ramaraochavali or @va7ish may want to add their comments.

howardjohn:
Its not just weighting, any case where requests will land on different endpoints. For example, if I have a route for `/bar -> bar, /foo -> foo`. Or even if I simply call `curl foo` then `curl bar` using the same cookie set on the client each time, since they use the same cookie they would be reset between each call?

sanjaypujare:
> Its not just weighting, any case where requests will land on different endpoints. For example, if I have a route for `/bar -> bar, /foo -> foo`. Or even if I simply call `curl foo` then `curl bar` using the same cookie set on the client each time, since they use the same cookie they would be reset between each call?

The way cookie based stateful session affinity works (in Envoy currently and in gRPC as proposed) is that if you use a cookie (which has the backend IP encoded) on the request and the LB in Envoy/gRPC does not route the request to that IP (for whatever reason - say that particular backend is unhealthy or the routing caused the request to go to a different cluster and obviously a different backend) then on the response will contain a new cookie with the new backend IP address and the session affinity is broken. This is by design.

In your example, your first request to /foo returned a cookie (backend BE1 handled the request) which you used in the next request to /bar and because of the routing setup the LB ended up sending the request to BE2 (say) so the cookie will now have BE2 IP address and the session affinity is broken - which is by design.

On the other hand if the user really wanted to maintain session affinity across paths like /foo and /bar then they need to make sure these requests will have the exact same routing (routes and clusters), same LB policies and the same cookie based session affinity configuration. So I think it makes even more sense to set up this config on the destination rule because in the outer (or preceding) layer the user could have set up complex routing in which case the requests will get routed to different clusters and then there is obviously no session affinity to maintain.

ramaraochavali:
> There was a LOT of discussion here so I may have missed some nuance, but it sounds like our proposal is to merge the API as-is, despite it not working in Envoy and having no concrete plan to make it work in Envoy. Is that accurate?

This is not correct. As @sanjaypujare explained, it supports most of the basic use cases. Limitations have been documented. Happy to add more docs if it is not clear.  The plan is to fix any missing pieces in Envoy. So the current api is definitely an usable api.

sanjaypujare:
> ...
> This is not correct. As @sanjaypujare explained, it supports most of the basic use cases. Limitations have been documented. Happy to add more docs if it is not clear. The plan is to fix any missing pieces in Envoy. So the current api is definitely an usable api.

Yes, I think it will be helpful to add some documentation especially examples illustrating capability and limitations to prove the API works for the intended use-case(s).

costinm:
John - persistent sessions have been around for a long time, and are well
supported by all load balancers (except envoy). Current Envoy
implementation has limitations - and we need
to document/fix them  - but it is already used by at least a few users via
EnvoyFilter, and for the application types they have Istio is not usable
without this feature ( or without
rewriting the app to bypass Istio by directly connecting to pods, losing
most security benefits of Istio).

Having a single cookie for multiple (separate) services has known solutions
-  and AFAIK it can be used today by adding an ExtAuthz server or by using
separate cookies for
each service. With a change in Envoy to encode foo=IP1,bar=IP2 it can also
work with a single cookie name.

What doesn't work - only because of implementation choices in envoy - is
traffic shifting or any LB that generates multiple clusters from the same
service and moves endpoints
across clusters.

The API itself just indicates user intent to have persistent sessions - and
is similar with what all other servers that have persistent sessions use.
My first proposal was to just
use an annotation on the Service, like few of the other LBs do - but DR has
the same semantics.

costinm:
How about first merging the other change in istio/istio, using an
annotation - and revisiting the DestinationRule after we have more usage
and feedback ?

We do use annotations for some features - and all other proxies implement
persistent sessions (and had them for a long time), and do this using
annotations.
Maybe we can start a discussion in Gateway group and go directly to a
cross-vendor 'policy' - while we validate if external authz could be used
with ingress
to allow users to plug-in an encrypted or obfuscated cookie, and we see if
envoy can fix their implementation to match what other proxy vendors
provide ?

I think it is important to start from use cases and existing solutions for
defining APIs - and not from envoy limitations or a 'original'/unique
approach to a common
problem.

ramaraochavali:
> How about first merging the other change in istio/istio, using an
annotation - and revisiting the DestinationRule after we have more usage
and feedback ?

I am fine with that. What does that annotation look like? And is it implmemented only for proxyless gRPC or regular proxies? If it is implemented (or planned to be implemented) for both - I am fine.

costinm:
In my WIP PR the annotation is actually a label - x-session-cookie - but
was mostly a placeholder to get started.

It is not specific to proxyless gRPC or envoy - and based on the discussion
on your PR I think we should keep it as small as possible,
ie. just something like "networking.istio.io/peristent-session=ID" to
indicate the provider intent for clients to use persistent sessions for the
service ( and all its subsets - when Envoy is fixed to allow this), with
the documentation indicating implementation MAY use the ID
in cookie or header name, but without guarantees that it'll be the actual
cookie name, so future changes can be made.

I think there are few advantages:
- it avoids some confusion associated with DR - like client overrides,
merging, etc - the Service annotation is more clearly a property
of the service, indicating the associated workloads require and use
persistent sessions.
- it is very similar to what the other load balancers support
- should work as well for Gateway/GAMMA users without requiring istio APIs
- hopefully would be easier to promote to an "Gateway policy attachment" -
since most LB vendors support something similar.

I think 'official' labels and annotations also need to be defined in
https://github.com/istio/api/blob/master/annotation/annotations.yaml -
if John is ok with the annotation we can modify this PR to change that
field instead, to keep the comment history ?

For the annotation name bike-shedding, existing names I found are:
- nginx.ingress.kubernetes.io/affinity: "cookie" ( plus few more
annotations to specify the name, expires )
- traefik.ingress.kuberenetes.io/affinity: "true" ( plus "
trafik.ingress.kubernetes.io/session-cookie-name" )
- router.openshift.io/cookie_name: "name"
- run.googleapis.com/sessionAffinity: true - for CloudRun ( as an extension
to KNative service spec )

I personally think 'persistent session' is a better name than 'affinity'
since we already use affinity for the weaker variant, and
java has used this term since ~2000, but happy with any other name the
community want.

va7ish:
> the Service annotation is more clearly a property of the service, indicating the associated workloads require and use persistent sessions.

Where would you apply the annotation to? Is this for a K8s Service or some other construct in Istio? 
If the former then Istio would still need to configure the clients through xDS to inject and handle some sort of affinity tag (I.e. cookie). This also means that two different service subsets must must still be defined by a single K8s service, does it?

costinm:
Annotation will be on the Service.

The value of the annotation should be an ID that can be used to associate
different subsets - in Istio with DestinationRule it is not
a problem, we can define subsets from the same Service. In the K8S Gateway
the versions/subsets are each an individual Service.
In other words - for a set of workloads that are part of the persistent
session pool, the same ID ( and indirectly cookie/header name)
 must be used and it should be shared by all Services or DRs that may
select subsets. We still need to make changes to
envoy to deal with an endpoint moving to a different Cluster - but at least
we'll have a way to associate all the clusters that
belong to the same persistent-session workload set.

ramaraochavali:
>. For the annotation name bike-shedding, existing names I found are:
- nginx.ingress.kubernetes.io/affinity: "cookie" ( plus few more
annotations to specify the name, expires )
- traefik.ingress.kuberenetes.io/affinity: "true" ( plus "
trafik.ingress.kubernetes.io/session-cookie-name" )
- router.openshift.io/cookie_name: "name"
- run.googleapis.com/sessionAffinity: true - for CloudRun ( as an extension
to KNative service spec )

I personally think 'persistent session' is a better name than 'affinity'
since we already use affinity for the weaker variant, and
java has used this term since ~2000, but happy with any other name the
community want.


I am fine with that. Should we use `traffic.istio.io/session-persistence-cookie` and move forward?

costinm:
If we are bikeshedding -
https://istio.io/latest/docs/reference/config/annotations/ has existing
annotations, the prefix is all over the place,
with networking.istio.io, traffic.istio.io, traffic.sidecar.istio.io.  I'm
not sure the prefix in front of istio.io helps much - I would just use
istio.io/persistent-session=true or istio.io/persistent-session-cookie=NAME
but I am ok with your proposed name too.

John, any preference ?

I am in doubt about how to express HOW the persistence will be implemented.
In Java the implementations are free to pick -
a client that doesn't support cookies may get the session using the query
param (?jsessionid=XXX), and we prototyped ( don't remember
 if it was shipped ) tracking sessions based on TLS session ID from the
handshake.

Would be good to hear from potential users - is there a need to explicitly
control the cookie name ( for example if other infra or code
is generating the cookie ), or leave flexibility to the implementation and
just have a bool ( we can refine it and add more options when
the proper CRD is defined ) ?

costinm:
From API perspective ( and how persistent sessions work with other serves )
- both header and cookie ( and other forms like TLS session ID,  request
parameter,
etc)  can be used.

The doc should mention that envoy currently supports only one. There is no
reason to believe it is impossible for envoy
to support multiple, and we should not design the API based on an envoy
limitation.

ramaraochavali:
@costinm @howardjohn any suggestions on annotation naming? I would like to move them to annotations and implement it.

costinm:
I merged https://github.com/istio/istio/pull/40660 - it has an
implementation for proxyless gRPC.

It would be great if you can use the same label and implement it in envoy,
happy to make changes as well.

sanjaypujare:
> May be it possible to record the cluster in the future. I will have a try. 🤔

@wbpcode have you been able to try it? Are there any plans to add support for weighted clusters? Any idea what changes would be required?

arhamGH:
Any idea when this can be merged ?

ramaraochavali:
i will open a fresh PR

PR #2470, opened 2022-08-31
Add support contains condition in Authz conditions

rrobb-aspen:
Fix Issue: https://github.com/istio/istio/issues/40784

We have customers who want to be able to create their own set of regular expressions for header matching in authorization policies.  The approach done in PR https://github.com/istio/istio/pull/40499 wanted to change the semantic behavior of `value`.  We believe that it is best not to change the current behavior but to add additional behavior to allow regex.  We need to document that users need to use this with extreme caution due to security and potential performance issues.

Edited: Our customer actually wants to be able to match on a substring within the entire header rather than doing full regex.  This should be able to be handled with a contains match instead.


liminw:
@rrobb-aspen We have not added Regex support in authorization policy due to security and performance concerns. In the past, we have been discussing alternative ways (e.g., glob matching) instead of supporting full regex. Do you mind sharing your use cases so we can evaluate how to extend the existing policies?

rrobb-aspen:
> @rrobb-aspen We have not added Regex support in authorization policy due to security and performance concerns. In the past, we have been discussing alternative ways (e.g., glob matching) instead of supporting full regex. Do you mind sharing your use cases so we can evaluate how to extend the existing policies?

@liminw Thank you for your concerns, we are also trying to figure out a good way on how to make this work with those concerns in mind.  The case we're looking at is being able to match on headers using a syntax similar to `*string to match*` where we're trying to match on a substring in the header when the user inputs a value surrounded by two `*`.  This might be able to be handled by `HeaderMatch_ContainsMatch`, which we think would be safer than adding regex.

liminw:
Thanks @rrobb-aspen for explaining your use case. We currently support exact match, prefix match, suffix match, and any match ("*"). Adding "contains match" sounds ok to me. However, the proto change should not be in "conditions". Instead, we are extending the way "values" or "not_values" are matched. We can explore alternatives, e.g., changing "values"/"not_values" types from "repeated string" to a repeated structure type for different matches, or come up with a semantic to express contains match using string type (similar to using "foo*" to express prefix match). We need a design proposal to explore different options and compare PROs and CONs.

rrobb-aspen:
@liminw and @costinm We evaluated our use case and what the customer is requesting is a contains feature and not a full regex.  We evaluated CEL and believe that this would be a better mechanism for this and that Envoy is currently [adding more support for CEL](https://www.envoyproxy.io/docs/envoy/latest/xds/type/matcher/v3/cel.proto). However, we do not know when they will have support for this available as it currently is still a work in progress.  Support for this feature is a more immediate concern for our customer considering their security posture.  We're planning on putting up two different PRs up for this change, one which will change the current implementation of values/not_values to add this additional code change without changing the API and another which will implement the change in the case where it's decided that an API change is required.  I will try to have the first up today, though the second may be delayed as I will be unavailable next week.

kyessenov:
From Envoy's perspective, using the unified matchers is the most flexible and sustainable option, and it supports various regexes (some special ones like hyperscan) and CEL (which is indeed will continue to be used in Envoy and gRPC).

rrobb-aspen:
@liminw Created PR https://github.com/istio/istio/pull/40786 with one option on adding contains matching.

sschepens:
In our use case, we definitely need regexes, we need to match some of the same things we can match on VirtualServices.

I understand the risks involved using regexes, but why not add this and make it a custom option to use or not?

howardjohn:
> In our use case, we definitely need regexes, we need to match some of the same things we can match on VirtualServices.
> 
> I understand the risks involved using regexes, but why not add this and make it a custom option to use or not?

Regex have been a source of a number of CVEs which, regardless of how much warning we do in the API, we become responsible for by maintaining an API to reach it. Additionally, users don't actually read our warnings so there isn't a feasible way to make it so that users acknowledge the risks of regex - if it's there it will be used. Given the complexity of regex's it will almost certainly be misused

security/v1beta1/authorization_policy.proto
```diff
@@ -464,4 +464,7 @@ message Condition {
   // Optional. A list of negative match of values for the attribute.
   // Note: at least one of values or not_values must be set.
   repeated string not_values = 3;
+
+  // Optional. A list of contained strings to match for the attribute.
+  repeated string contains = 4;
```
hzxuzhonghu, line 469: And may also need not contains


Similar with k8s label selector

costinm:
I think we need to have a very serious design doc before adding any support like this, in particular around performance/scale. 
This is common to both authz and routing. And in particular for using Authz in ingress/egress gateway. 

Besides CVEs - both regex and CEL force a linear-search without pretty complex optimizations ( which I doubt Envoy has ). 
To make it worse - current authz is added at listener level in Istio. Envoy does support per route - but we do not use it. 

At the very least:
- we should restrict the use of CEL or RE to HTTP headers - not all fields in the authz ( for example TCP-level values should stil
be restricted ).
- use of CEL/RE should require at least a hostname and path prefix match that are associated with a route - and make sure the CEL/RE rules are added to the route, and don't impact all other hosts. This is far better handled using the new K8S Gateway policy, by attaching CEL/RE to routes.

On top of this, the Authz CRD is moving to v1 - and it is already very confusing that some fields are beta. Obviously
any new feature like RE would start as 'experimental'.

Giving the above - I would be ok with CEL/RE if it is added in a SEPARATE CRD defined as a K8S Gateway policy
instead of extending v1 Authz. 

I would prefer we start with CEL - http structured headers are very hard to parse with RE alone ( and extremely likely
to get wrong and have a CVE if RE is used for security with structured headers ).  I can live with RE too if it is strictly
restricted to one format.

liminw:
I agree with Costin that we need a serious design doc for this feature. This change will have big impact on the security and performance aspects of Istio authorization policy. We need to review the options of introducing glob, CEL/RE. In either case, we need to avoid impacting the performance in the common cases, and the semantics should be simple and clear (likely a restricted subset of CEL or glob). The feature can be introduced as experimental initially. 

@rrobb-aspen I suggest we pause this API PR and the associated code change. We need to start with a design and plan for this feature.

jacob-delgado:
We are evaluating our options and making sure it passes the CI/CD. We will have a formal doc in the coming weeks detailing our findings. We understand that it can take time to get community feedback before anything is approved, but we would like to understand the pros/cons better before we document this and present it to the community. I'll put up another PR up soon that has a CEL approach (hopefully) to aid in this discussion.

I would, personally, not let us allow RegEx, as having users be able to specify their own regular expressions would open them up to issues they may not do well. Our customers have been asking for support for a "contains" like feature. We've had numerous issues with regex parsing and sanitization issues with previous releases. We understand that this can be *__very__* contentious.

Our customers have a use case to parse `X-Forwarded-Client-Cert` and to only allow specific `DNS` and `URI` services that have been qualified in their chain to authorize a service. They definitely do not need full regex support, but a supplement to the `*` prefix and suffix functionality that already exists; a contains-like behavior. But as we learned when we did https://github.com/istio/istio/pull/40786 there would be a behavior change regarding this as  `*` suffix behavior further complicating this change. So please disregard these PRs as we just try and get early feedback before writing a formal design doc.

We have timelines for this and are trying to come up with solutions that wouldn't fork the API.

howardjohn:
Parsing `X-Forwarded-Client-Cert` is the exact type of thing `regex` and `contains` _should not_ be used for IMO. I am nearly certain any rule that someone could come up with to parse these by regex is either incredibly complicated or incorrect.

I think @costinm was looking into adding first class support for XFCC?

jacob-delgado:
Getting real first-class support for XFCC is definitely something that we need to be involved and we can start talking about that solution instead. We aren't trying to bypass the design phase, but start teasing out these concerns.

Istio should probably not allow either XFF or XFCC be specified as a header you can use as even the prefix or suffix solution presents potential security issues, but that would also be a behavior change if we added that validation. I'm not comfortable with anyone doing this right, but there aren't any other means for people to validate them right now.

howardjohn:
Yep totally get that. We can always add a warning instead of error btw, to
avoid breaking changes. Parsing these is super important - but tricky -
which is why I was leaning towards native support

costinm:
What I explored is transparent support for XFCC - i.e. low level authz code will parse it, extract SANs/etc and use it directly in the existing authz rules ( assuming the caller is authenticated and authorized for  'trust delegation' - like an ingress gateway ).

Agree with John, the worse thing you can do is attempt substring or regex for XFCC. Don't forget quotes and 
random strings that may show up in the SAN value ( it's a URL - an attacker can insert a lot of stuff ). I think XFCC and 
in general 'structured http headers' ( there is a standard for this ) are best handled with CEL or similar path-based expressions.
It would be a disaster to allow regex for this.

rrobb-aspen:
We've taken @costinm's thoughts and the feedback from the community into consideration.  We are hoping to have a solution presented to the community in the on-coming weeks.

PR #2476, opened 2022-09-07
Fix the comments explaining Workload mode of gateways

ingwonsong:
Fix the comments describing the workload mode of Gateway.

To have consistency with the existing Telemetry API, we have changed the semantics of Workload mode in Gateway, but there was an omitted comments which had to be changed as well.

### DIFF
--- a/extensions/v1alpha1/wasm.proto
+++ b/extensions/v1alpha1/wasm.proto
@@ -330,7 +330,7 @@ message WasmPlugin {
     // Note that CLIENT and SERVER are analogous to INBOUND and OUTBOUND,
     // respectively.
     // For the gateway, the field should be CLIENT or CLIENT_AND_SERVER.
-    // If not specified, CLIENT_AND_SERVER will be the default value.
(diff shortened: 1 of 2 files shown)


howardjohn (approved):

extensions/v1alpha1/wasm.proto
```diff
@@ -328,9 +328,9 @@ message WasmPlugin {
     //
     // Criteria for selecting traffic by their direction.
     // Note that CLIENT and SERVER are analogous to INBOUND and OUTBOUND,
-    // respectively. If the selected workload is a gateway, the field is ignored.
-    // If not specified, this condition is evaluated to true for any traffic
-    // direction.
+    // respectively.
+    // For the gateway, the field should be CLIENT or CLIENT_AND_SERVER.
+    // If not specified, CLIENT_AND_SERVER will be the default value.
```
ericvn, line 333: Nit: Maybe..
```suggestion
    // If not specified, the default value is CLIENT_AND_SERVER.
```

ericvn (approved):

PR #2477, opened 2022-09-08
Added gateway server snis

hzxuzhonghu:
FYI https://discuss.istio.io/t/http-404-when-authority-and-requested-server-name-are-different/10713

This is to handle cases SNI and Host not equal

@ramaraochavali


howardjohn (commented):
Is this a use case we want to support? It seems not to be something other products expose and an uncommon pattern

hzxuzhonghu:
It is a small option, i think to be consistent with https://github.com/istio/api/blob/master/networking/v1alpha3/destination_rule.pb.go#L1562. We should do

networking/v1alpha3/gateway.proto
```diff
@@ -728,4 +728,10 @@ message ServerTLSSettings {
   // Optional: If specified, only support the specified cipher list.
   // Otherwise default to the default cipher list supported by Envoy.
   repeated string cipher_suites = 9;
+
+  // An optional list of server names (e.g. SNI for TLS protocol) to consider when determining
```
costinm, line 732: I don't get it.  Client presents a SNI, it is matched against one of the certificates the server has - and that's it. 

Either gateway has a cert for that SNI and returns it or it doesn't - how would a list of SNIs in the gateway config change this ? Another Envoy invented TLS 'standard' ?

costinm: I assume this is not for SNI routing - but for terminating TLS. 

I would guess it can be useful if the cert has some SANs not matching the hosts - but then what will the request look like ? Should we even allow requests if SNI does not match one of the expected hostname ?

hzxuzhonghu: Yes, it is not for SNI routing.

Yep, in some case, the client can use a different SNI as `Host` header https://discuss.istio.io/t/http-404-when-authority-and-requested-server-name-are-different/10713

And in gateway, it use sni as a matching condition for filterchain instead of host name, while route domain match is based on host.

costinm: I know - but it is debatable if this is a 'feature' or a security vulnerability. 

Given that the SNIs are ( or should be) generated from the host names, with the assumption that the cert has the same SANs - can this be simply done by just adding 'hostnames' for all additional SNIs ? Or change envoy implementation 
to auto-extract the SNIs from the certs ?

I think we need to seriously prioritize the move to the official K8S Gateway API - if something can't be done with the 
K8S API we should consider proposing a change there or use a policy attachment - this also means APIs need to consider other servers, not only envoy behavior.

howardjohn: This is not currently possible in the gateway-api BTW

costinm:
I thought if I define all those names as hostnames - it will generate the
SNI filters ?

Ie. if I define hostnames: example.com, foo.com - it wouldn't work if the
generated config didn't recognize
both domains as SNI names.

howardjohn:
Istio and gateway-api work the same - Gateway hostname and Route hostname form a match. The Gateway hostname becomes an SNI match, but the route hostname must be a subset of that match (i.e. the Host header needs to match the SNI).

What this is adding, by my understanding, is a way to allow completely arbitrary SNIs without constraining the set of route hostnames.

costinm:
Do we actually check that the host header matches the SNI ?

I'm starting to think this is better left to an envoy filter, I don't think
it's a common enough use case - and the risks of misuse and security issues
seem high enough...

networking/v1alpha3/gateway.proto
linsun: What is the conclusion for this?  Do we have other users ask for this besides that discuss link?

PR #2482, opened 2022-09-19
Update ConfigSource.address examples in docs

rajatsharma94:
The ConfigSource.address field mentions that the MCP backend address can be an IP address but doesn't clarify the protocol that needs to be used (xds:// or k8s://) in the docs.

Just adding a line to the doc to make it clear in the docs so that people don't have to dive into the code to see the possible address formats

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -1190,7 +1190,8 @@ message MeshConfig {
 message ConfigSource {
  // Address of the server implementing the Istio Mesh Configuration
  // protocol (MCP). Can be IP address or a fully qualified DNS name.
- // Use fs:/// to specify a file-based backend with absolute path to the directory.
+ // Use xds:// to specify a grpc-based xds backend, k8s:// to specify a k8s controller or
+ // fs:/// to specify a file-based backend with absolute path to the directory.
  string address = 1;
  // Use the tls_settings to specify the tls mode to use. If the MCP server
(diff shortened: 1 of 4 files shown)


mesh/v1alpha1/config.proto
```diff
@@ -1190,7 +1190,8 @@ message MeshConfig {
 message ConfigSource {
  // Address of the server implementing the Istio Mesh Configuration
  // protocol (MCP). Can be IP address or a fully qualified DNS name.
- // Use fs:/// to specify a file-based backend with absolute path to the directory.
+ // Use xds:// to specify a grpc-based xds backend, k8s:// to specify a k8s controller or
```
hzxuzhonghu, line 1193: do we really need `k8s`? It is the default way, once we document it, it would be hard to remove that.

rajatsharma94: I added that to keep the documentation in-line with the code. Also, since not adding to docs doesn't imply that its not being used, I guess removing/changing that should follow the general decommissioning process ?

linsun (approved):

costinm:
K8S is the default - but AFAIK if the config sources are used, the default
is not added ? Don't remember exactly - but the original
use case was for the tests using fs:// - and that meant k8s would not be
used. If a user wants to combine k8s and fs sources,
or multiple k8s source - k8s:// is needed ?

PR #2498, opened 2022-10-08
telemetry: fix typo

zirain:
fixes: https://github.com/istio/istio/issues/41307


telemetry/v1alpha1/telemetry.proto
```diff
@@ -367,7 +367,7 @@ message Tracing {
 // Used to bind Telemetry configuration to specific providers for
 // targeted customization.
 message ProviderRef {
-  // Required. Name of Telemetry provider in MeshConfig.
```
hzxuzhonghu, line 370: It is required, so donot remove it

PR #2502, opened 2022-10-11
Added note tls version below 1.2 should set valida cipher suites

hzxuzhonghu:


networking/v1alpha3/gateway.proto
```diff
@@ -719,7 +719,11 @@ message ServerTLSSettings {
     TLSV1_3 = 4;
   }
 
-  // Optional: Minimum TLS protocol version.
+  // Optional: Minimum TLS protocol version. By default, it is `TLSV1_2`.
+  // TLS protocol versions below TLSV1_2 require setting compatible ciphers with the
+  // `cipherSuites` setting as the default ciphers no longer include compatible ciphers.
```
ericvn, line 724: Reading this the default is 1.2. Versions below 1.2 require setting compatible ciphers. Then the last part says that 1.2, the default, no longer includes compatible ciphers. I think we either need to say 1.2 and below don't include compatible ciphers or more likely its change 
```suggestion
  // `cipherSuites` setting as they no longer include compatible ciphers.
```

howardjohn:
Can we add a warning to the validation webhook for this as well? Or alterntively, just automatically set the ciphers?

hzxuzhonghu:
> Can we add a warning to the validation webhook for this as well? Or alterntively, just automatically set the ciphers?

SG, this will bring better UX

networking/v1alpha3/gateway.proto
hzxuzhonghu: done

hzxuzhonghu:
@howardjohn

ericvn (approved):

networking/v1beta1/gateway.proto
```diff
@@ -719,7 +719,11 @@ message ServerTLSSettings {
     TLSV1_3 = 4;
   }
 
-  // Optional: Minimum TLS protocol version.
+  // Optional: Minimum TLS protocol version. By default, it is `TLSV1_2`.
```
louiscryan, line 722: @lei-tang please review

lei-tang (commented):
LGTM. 
Nit: Consider replacing "Using TLS protocol versions below TLSV1_2 has serious security considerations and risks." with "Using TLS protocol versions below TLSV1_2 has serious security risks."

networking/v1beta1/gateway.proto
lei-tang: LGTM.
@hzxuzhonghu 
Nit: Consider replacing "Using TLS protocol versions below TLSV1_2 has serious security considerations and risks." with "Using TLS protocol versions below TLSV1_2 has serious security risks."

hzxuzhonghu: SG, i will update

PR #2509, opened 2022-10-14
Run `make gen` with the new build-tools to generate new html

ericvn:
The new build-tools has a new markdown generator for the html files (same as the one used by istio.io. The PR is simply `make gen` with the new build-image to update the html files.

This PR is needed to prevent other PRs from having to check these changes in.

### DIFF
--- a/analysis/v1alpha1/message.pb.html
+++ b/analysis/v1alpha1/message.pb.html
@@ -104,7 +104,7 @@ No
 <td><code>template</code></td>
 <td><code>string</code></td>
 <td>
-<p>A go-style template string (https://golang.org/pkg/fmt/#hdr-Printing)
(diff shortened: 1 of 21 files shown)


ericvn:
The gen-check test will pass when the pipeline moves to the same image as the istio/api repos is using.

howardjohn (approved):

zirain:
@ericvn this should merge first?

ericvn:
@zirain I was waiting for the 2 needed approvals for this PR. Once the approvals are in place, I will merge https://github.com/istio/test-infra/pull/4331 which will update the pipeline to use the new image. At that point, the gencheck for this PR will pass and then merge.

If I simply merge the test-infra PR, then the gencheck will fail in other PRs unless they also include these changes for all files, not just what they are changing.

If a user tries to run `make gen` on the current commit and submit the PR, they will have the same gencheck failure this PR does as the repo and tooling are out of sync.

Hopefully, we can get the two PRs merged soon.

kfaseela (approved):

PR #2515, opened 2022-10-20
hpa: Policies move to v2beta2

zirain:
fixes: https://github.com/istio/istio/issues/41309

cc @richardwxn


howardjohn (commented):
LGTM if @richardwxn approves it, as they are much more of an expert here

richardwxn (approved):

howardjohn (approved):

PR #2522, opened 2022-10-21
Fix abort and delay percentage docs

hzxuzhonghu:
If the percentage unset, istio makes it equal to 0, which means no request will be aborted and delayed.

We have talked about whether to change istio implementation to make unset = 100%, but this can break compatibility.

https://github.com/istio/istio/issues/39969


ericvn (approved):
LGTM - Agree we can't change the behavior to match the original comments.

hzxuzhonghu:
@howardjohn

howardjohn (approved):

PR #2526, opened 2022-10-24
remove comment about not support regex for delegete vs

### LINKED ISSUE #2527: Istio support regex for delegate vs and docs not changed
**Bug description**

image

[X] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience

**Expected behavior**
 Should remove this notes about not support regex in delegate vs because by "https://github.com/istio/istio/pull/38692"
**Steps to reproduce the bug**

**Version (include the output of `istioctl version --remote` and `kubectl version`)**

**How was Istio installed?**

**Environment where bug was observed (cloud vendor, OS, etc)**

Additionally, please consider attaching a [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) by attaching
the dump file to this issue.

luohua13:
Resolves #2527 
Signed-off-by: luohua13 <jcwang@alauda.io>

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -1005,8 +1005,8 @@ message TCPRoute {
 //
 // HTTPMatchRequest CANNOT be empty.
 // **Note:**
-// 1. If root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
-// 2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
+// 1. If a root VirtualService have matched any property (path, header etc.) by regex, delegate VirtualServices should not have any other matches on the same property.
+// 2. If a delegate VirtualService have matched any property (path, header etc.) by regex, root VirtualServices should not have any other matches on the same property.
 message HTTPMatchRequest {
   // The name assigned to a match. The match's name will be
   // concatenated with the parent route's name and will be logged in
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -1004,7 +1004,9 @@ message TCPRoute {
 // {{</tabset>}}
 //
 // HTTPMatchRequest CANNOT be empty.
-// **Note:** No regex string match can be set when delegate VirtualService is specified.
+// **Note:**
+// 1. If root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
+// 2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
 message HTTPMatchRequest {
   // The name assigned to a match. The match's name will be
   // concatenated with the parent route's name and will be logged in
--- a/networking/v1beta1/virtual_service.proto
+++ b/networking/v1beta1/virtual_service.proto
@@ -1005,8 +1005,8 @@ message TCPRoute {
 //
 // HTTPMatchRequest CANNOT be empty.
 // **Note:**
-// 1. If root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
-// 2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
+// 1. If root VirtualService specifies regex, delegate VirtualService should not have any other matches.
+// 2. If delegate VirtualService specifies regex, root VirtualService should not have any other matches.
 message HTTPMatchRequest {
   // The name assigned to a match. The match's name will be
   // concatenated with the parent route's name and will be logged in
--- a/networking/v1beta1/virtual_service.proto
+++ b/networking/v1beta1/virtual_service.proto
@@ -1004,7 +1004,9 @@ message TCPRoute {
 // {{</tabset>}}
 //
 // HTTPMatchRequest CANNOT be empty.
-// **Note:** No regex string match can be set when delegate VirtualService is specified.
+// **Note:**
+// 1. If root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
+// 2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
 message HTTPMatchRequest {
   // The name assigned to a match. The match's name will be
   // concatenated with the parent route's name and will be logged in
--- a/networking/v1beta1/virtual_service.proto
+++ b/networking/v1beta1/virtual_service.proto
@@ -1005,8 +1005,8 @@ message TCPRoute {
 //
 // HTTPMatchRequest CANNOT be empty.
 // **Note:**
-// 1. If root VirtualService specifies regex, delegate VirtualService should not have any other matches.
-// 2. If delegate VirtualService specifies regex, root VirtualService should not have any other matches.
+// 1. If a root VirtualService have matched any property (path, header etc.) by regex, delegate VirtualServices should not have any other matches on the same property.
+// 2. If a delegate VirtualService have matched any property (path, header etc.) by regex, root VirtualServices should not have any other matches on the same property.
 message HTTPMatchRequest {
   // The name assigned to a match. The match's name will be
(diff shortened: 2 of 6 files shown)


luohua13:
Remove this comment because by "https://github.com/istio/istio/pull/38692"

networking/v1alpha3/virtual_service.proto
```diff
@@ -1004,7 +1004,6 @@ message TCPRoute {
 // {{</tabset>}}
 //
 // HTTPMatchRequest CANNOT be empty.
-// **Note:** No regex string match can be set when delegate VirtualService is specified.
```
ramaraochavali, line 1007: Can you please clarify here that the "regex" can be specified either at root or delegate but not both?

ramaraochavali: Also if delegate specifies regex, root should not have other matches.

ramaraochavali (commented):
Thank you for fixing this comment. Need to clarify couple of points about "regex" usage.

networking/v1alpha3/virtual_service.proto
luohua13: If "regex" at root, then "regex" or "prefix" or "exact" at delegate is invalid.
If "regex" at delegate, and if any of "regex" and "prefix" and "exact" at root, then delegate is invalid, otherwise delegate “regex” is valid.

luohua13: > Can you please clarify here that the "regex" can be specified either at root or delegate but not both?

You can specified both, just the delegate is invalid.

luohua13: @ramaraochavali I thought for a moment. I can do as you say. And I will clarify it in the code.

luohua13:
Can you review it again? @ramaraochavali

networking/v1alpha3/virtual_service.proto
```diff
@@ -1004,6 +1004,7 @@ message TCPRoute {
 // {{</tabset>}}
 //
 // HTTPMatchRequest CANNOT be empty.
+// **Note:** Regex string match can be specified either at root or delegate VirtualService but not both. Also if delegate VirtualService specifies regex, the root VirtualService should not have other matches.
```
ramaraochavali, line 1007:
```suggestion
// **Note:** 
// 1. Regex string match can be specified either at root or delegate VirtualService but not at both. 
// 2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
```

luohua13: @ramaraochavali I have modified, then we merge it?

ramaraochavali (commented):
LGTM @howardjohn @ericvn PTAL

networking/v1alpha3/virtual_service.proto
```diff
@@ -1004,7 +1004,9 @@ message TCPRoute {
 // {{</tabset>}}
 //
 // HTTPMatchRequest CANNOT be empty.
-// **Note:** No regex string match can be set when delegate VirtualService is specified.
+// **Note:**
+// 1. Regex string match can be specified either at root or delegate VirtualService but not at both.
+// 2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
```
hzxuzhonghu, line 1009: if either vs specifies regex, no other matching can be specified

luohua13: I should edit as follows:
1. If root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.

luohua13:
can you merge it now? @howardjohn

hzxuzhonghu (approved):

releasenotes/notes/delegate-vs-regex.yaml
```diff
@@ -0,0 +1,14 @@
...
+# issue is a list of GitHub issues resolved in this note.
+issue:
+  - 2527
+
+docs:
+ - '[usage] https://istio.io/latest/docs/reference/config/networking/virtual-service/'
+
+releaseNotes:
+  - |
+    **Fixed** remove comment about not support regex for delegete VirtualService
```
ericvn, line 14:
```suggestion
    **Removed** comment about not supporting regex for the delegate VirtualService
```

networking/v1beta1/virtual_service.proto
```diff
@@ -1004,7 +1004,9 @@ message TCPRoute {
 // {{</tabset>}}
 //
 // HTTPMatchRequest CANNOT be empty.
-// **Note:** No regex string match can be set when delegate VirtualService is specified.
+// **Note:**
+// 1. If root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
+// 2. If delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
```
ericvn, line 1009: I'm not sure if `the` should be included, but we sometimes are and sometimes aren't. I'll error in my comments that we should.
```suggestion
// 1. If the root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
// 2. If the delegate VirtualService specifies regex, the root VirtualService should not have any other matches.
```

networking/v1beta1/virtual_service.proto
```diff
@@ -1004,7 +1004,9 @@ message TCPRoute {
 // {{</tabset>}}
 //
 // HTTPMatchRequest CANNOT be empty.
-// **Note:** No regex string match can be set when delegate VirtualService is specified.
+// **Note:**
+// 1. If root VirtualService specifies regex, the delegate VirtualService should not have any other matches.
```
louiscryan, line 1008: This definition is not aligned with that elsewhere in the doc 

"The delegate’s HTTPMatchRequest must be a strict subset of the root’s, otherwise there is a conflict and the HTTPRoute will not take effect."

I don't know what 'any other matches' means in this context? Do you mean that the delegate must 'have only one route with the exact same match as the root? Or do you mean something else.

luohua13: If the root VirtualService specifies regex, the delegate VirtualService's matchs will be invalid even if it is the same match.

luohua13: @louiscryan  Refer to this link: "https://github.com/istio/istio/pull/38692"

networking/v1beta1/virtual_service.proto
luohua13: The pr comment of [](https://github.com/istio/istio/pull/38692) not include **the**

ericvn: Then I would remove it in the second part of the sentence to be consistent.

networking/v1beta1/virtual_service.proto
louiscryan: OK. What I read in the code doesn't seem to align with the description here. I think the following is more accurate.

"If a root VirtualService matches any property (path, header etc.) by regex then a delegate VirtualService must not match on the _same property_. Similarly the delegate VirtualService can not match by regex any property _matched_ by the root."

@ramaraochavali ?

luohua13: Yeah, you are right.I will edit it again.

luohua13: @louiscryan I have modified it. Would you like to review it again?

ramaraochavali: [14 hours ago](https://github.com/istio/api/pull/2526#discussion_r1011024951) - Yes this is more accurate.

linsun (approved):
/ok-to-test

LGTM - all comments from other reviews are resolved

PR #2530, opened 2022-10-25
Matched horizontal pod autoscaler behaviour field with that of k8s

trick392:
fixes: https://github.com/istio/istio/issues/41309


zirain:
duplicate with https://github.com/istio/api/pull/2515

PR #2532, opened 2022-10-29
remove thrift config from config

ramaraochavali:


mesh/v1alpha1/config.proto
```diff
@@ -361,18 +361,8 @@ message MeshConfig {
   // Note: Deprecated, please refer to Cert-Manager or other cert provisioning solutions to sign DNS certificates. 
   repeated Certificate certificates = 47;
 
-  message ThriftConfig {
```
hzxuzhonghu, line 364: How about hide it

ramaraochavali: We have removed the code completely and do not have plans to support it. So not see any benefits of hiding. Do you see any benefits of hiding vs. removing?

hzxuzhonghu: Will it break compatability removing a field?

ramaraochavali: AFAIK. it should not

ericvn:
Note that this change is causing a failure in the client-go update: https://prow.istio.io/job-history/gs/istio-prow/logs/update_client-go_dep_client-go_postsubmit

ramaraochavali:
@ericvn sorry. Here is the PR https://github.com/istio/istio/pull/41711 to fix it

PR #2539, opened 2022-11-02
stats: support custom histogram bucket

zirain:
after https://github.com/istio/istio/pull/41441, add first class api to support custom histogram bucket

cc @kyessenov


mesh/v1alpha1/proxy.proto
```diff
@@ -579,6 +580,22 @@ message ProxyConfig {
 
   // Specifies the details of the Private Key Provider configuration for gateway and sidecar proxies.
   PrivateKeyProvider private_key_provider = 38;
+
+  // Defines rules for setting the histogram buckets.
+  message ProxyStatsHistogramBucketSetting {
```
hzxuzhonghu, line 585: should document how to use it and provide an example

mesh/v1alpha1/proxy.proto
```diff
@@ -579,6 +580,22 @@ message ProxyConfig {
...
+    // Specifies a matcher for stats and the buckets that matching stats should use.
+    // The match is applied to the original stat name before tag-extraction,
+    // for example `cluster.exampleclustername.upstream_cx_length_ms`.
```
howardjohn, line 588: Does this apply for istio_request_total type metrics? Or only envoy ones?

zirain: https://github.com/istio/istio/pull/41441 change the implement of istio.stats (with Telemetry API), after that this work for all stats.

howardjohn: ok

Telemetry already has a way of selecting metrics. I am a bit worried about using a different mechanism..

zirain:
@hzxuzhonghu @howardjohn should support this on `ProxyConfig`?

mesh/v1alpha1/proxy.proto
zirain: updated

mesh/v1alpha1/proxy.proto
hzxuzhonghu: How can it be applied to counter metric?

hzxuzhonghu:
ProxyConfig should only include xds client metadata, but later it is extended to include some other data plane config, which is not covered by API.  For this case, telemetry api can apply to a specific proxy, so I feel proxy-related metric setting should be there.

mesh/v1alpha1/proxy.proto
zirain: it only works for `Histogram` metric

zirain:
> ProxyConfig should only include xds client metadata, but later it is extended to include some other data plane config, which is not covered by API. For this case, telemetry api can apply to a specific proxy, so I feel proxy-related metric setting should be there.

this's actually part of statsconfig in bootstrap, similar to `ProxyStatsMatcher.InclusionPrefixes`

mesh/v1alpha1/proxy.proto
```diff
@@ -579,6 +580,50 @@ message ProxyConfig {
...
+  //   - 10000
+  //   - 30000
+  //   - 60000
+  //   - 300000
+  //   - 600000
+  //   - 1800000
+  //   - 3600000
+  //   - 7200000
+  // ```
+  message ProxyStatsHistogramBucketSetting {
```
hzxuzhonghu, line 613: Proxy prefix is unneccessary, it is in ProxyConfig

type/v1alpha1/string_match.proto
```diff
@@ -0,0 +1,37 @@
...
+// $title: String Match
+// $description: Definition how to match a given string.
+// $location: https://istio.io/docs/reference/config/type/string-match.html
+
+package istio.type.v1alpha1;
+
+option go_package="istio.io/api/type/v1alpha1";
+
+// Describes how to match a given string. Match is case-sensitive.
+message StringMatch {
```
hzxuzhonghu, line 25: I think there is already such a API, which is used by VS

hzxuzhonghu:
>this's actually part of statsconfig in bootstrap, similar to ProxyStatsMatcher.InclusionPrefixes

IC,

type/v1alpha1/string_match.proto
zirain: there is two `message StringMatch` in this repo, they may share same API just like `WorkloadSelector`?

mesh/v1alpha1/proxy.proto
zirain: just keep same as `ProxyStatsMatcher`

kyessenov:
LGTM. I think it's useful for customizing standard Envoy metrics, too, so it can't be in just the stats filter. Looks like it's mostly verbatim copied into bootstrap, so it should be noted that it requires a proxy restart.

zirain:
> LGTM. I think it's useful for customizing standard Envoy metrics, too, so it can't be in just the stats filter. Looks like it's mostly verbatim copied into bootstrap, so it should be noted that it requires a proxy restart.

update notes

nrjpoddar:
This is a bit confusing - We were trying to get away from ProxyConfig options for Telemetry in lieu of Telemetry API. Why is the new API not used?

Secondly, it's unclear what all metrics does it effect.

nrjpoddar:
This is a bit confusing - We were trying to get away from ProxyConfig options for Telemetry in lieu of Telemetry API. Why is the new API not used?

Secondly, it's unclear what all metrics does it effect. What happens if you update a histogram metrics buckets? Can you make sense of old and new data? Does it use the same name?

zirain:
> This is a bit confusing - We were trying to get away from ProxyConfig options for Telemetry in lieu of Telemetry API. Why is the new API not used?

this's actually part of bootstrap, Temeletry API cannot achieve 

> Secondly, it's unclear what all metrics does it effect. What happens if you update a histogram metrics buckets? Can you make sense of old and new data? Does it use the same name?

it will not change metrics name, only change the default buckets of histogram metric.

zirain:
kindly ping @nrjpoddar

zirain:
Telemetry API cannot change configuration in bootstrap, BDS(bootstrap discovery request) also will not help in this scenario.

mesh/v1alpha1/proxy.proto
```diff
@@ -591,6 +592,42 @@ message ProxyConfig {
...
+  //   - 5000
+  //   - 10000
+  //   - 30000
+  // ```
+  // Note: as part of bootstrap configuration, your may need restart the proxy to make this effective
+  message ProxyStatsHistogramBucketSetting {
+    // Specifies a matcher for stats and the buckets that matching stats should use.
+    // The match is applied to the original stat name before tag-extraction,
+    // for example `cluster.exampleclustername.upstream_cx_length_ms`.
+    istio.type.v1alpha1.StringMatch match = 1;
```
nrjpoddar, line 621: Why a new type?

costinm (commented):
I have a few concerns here:
- ProxyConfig vs Telemetry - not sure there is a clear answer on why it's kept in ProxyConfig
- the regex matching - I don't think it's a good idea in general, but for metrics in particular.
- the shape of the API - I think it's reasonable to customize metrics, but I would expect an
API where for each metric (by name - we don't have that many) we would have some settings associated to that metric - buckets, etc.

But most important concern: we are in the middle of few major changes, with 
OpenTelemetry maturing, Ambient, K8S Gateway happening. We also know that
Istio telemetry 'cardinality' and miss-match with other telemetry standards is a problem.
One of the options to improve and better integrate with OTel - for example generate the of the high-cardinality metrics from otel access logs ( historically istio metrics  were designed with Mixer - which was access log based, so less of a problem ). 

I would really like to see a doc or have a discussion about the future of telemetry and 
how it fits with the evolving world around us - and which APIs belong in Istio and which 
in open telemetry.

costinm:
Regarding the bootstrap - I think this is an implementation detail of Envoy that leaks into API without any good reason. 

We can use any of the Istio APIs in bootstrap - just need to inject them into the pod like we do for ProxyConfig, or get them over XDS before starting envoy ( agent already gets the root CAs from Istiod ). We can get all the config in agent
already. 

Ambient on the other hand is likely to not have any injection at all - and a lot of the changes that depend on injecting
to a pod will no longer work ( Waypoint is shared ). I think any new API should take into account ambient, given 
the major simplification and perf improvements it brings.

mesh/v1alpha1/proxy.proto
zirain: there is two protoo `StringMatch` in this repo, they may share same API just like `WorkloadSelector`?

I can reuse the one with VS.

mesh/v1alpha1/proxy.proto
zirain: unlike `Telemetry API`, `The match is applied to the original stat name before tag-extraction`

zirain:
> ProxyConfig vs Telemetry - not sure there is a clear answer on why it's kept in ProxyConfig

AFAIK, `Telemetry` used to generate configuration for stats filter, and as a part of bootstrap config, that's why I put this on `ProxyConfig`.

> the regex matching - I don't think it's a good idea in general, but for metrics in particular.

I agree that regex it's not a good idea, but it's widely used in envoy statistic system.

> I would really like to see a doc or have a discussion about the future of telemetry and
how it fits with the evolving world around us - and which APIs belong in Istio and which
in open telemetry.

that's another topic, which is too large for this PR (IMO, this's minor improvement)
Before v1.17, this won't work on a wasm filter, it's not a bad good idea to provide an api from the beginning of switching to native filter.

costinm:
On Fri, Dec 9, 2022 at 11:31 PM zirain ***@***.***> wrote:

> ProxyConfig vs Telemetry - not sure there is a clear answer on why it's
> kept in ProxyConfig
>
> AFAIK, Telemetry used to generate configuration for stats filter, and as
> a part of bootstrap config, that's why I put this on ProxyConfig.
>
We keep confusing implementation details in envoy with Istio API.

The fact Envoy happens to configure something in bootstrap should not have
any impact on the API surface.
The agent can get telemetry configs and generate whatever bootstrap it
needs - and when we move to
ztunnel in ambient it may generate whatever rust needs.

the regex matching - I don't think it's a good idea in general, but for
> metrics in particular.
>
> I agree that regex it's not a good idea, but it's widely used in envoy
> statistic system.
>
Envoy configs are huge and usually just a reflection of its internal
implementation details.
There are a lot of features in Envoy we choose not to expose in Istio API -
and even more that
won't be exposed in K8S Gateway or OTel.

Istio API should be implementable with ztunnel or proxyless gRPC.

If users absolutely need an envoy config - we have EnvoyFilter, and I
believe it is now possible
to also have the bootstrap generated dynamically by Istiod, for users who
want to take strong
dependencies on envoy internal impl.


> I would really like to see a doc or have a discussion about the future of
> telemetry and
> how it fits with the evolving world around us - and which APIs belong in
> Istio and which
> in open telemetry.
>
> that's another topic, which is too large for this PR (IMO, this's minor
> improvement)
> Before v1.17, this won't work on a wasm filter, it's not a bad good idea
> to provide an api from the beginning of switching to native filter.
>
Any API we add is going to make the transition to ambient and OTel a bit
harder.

I have nothing against the feature - it seems quite useful - but we must
take into account what
happens outside of Istio and where we want to go long term.

For example, if this historgram setting will be used for TCP metrics - how
would they be implemented
by ztunnel (where there is no ProxyConfig - since it's shared per node -
and no envoy ) ?
And if we switch to OTel - does the collector or instrumentation have a
similar config ?

kyessenov:
To resolve some ambiguities:
1. Histogram buckets affect all sinks (OTel in the future, prometheus now), except Wasm ones (due to ABI)
2. Telemetry API should define them for Istio metrics, and we need to ask users to use BDS or set them before injection. @costinm this would require capability to read telemetry APIs in injection for latter.
3. Leave Envoy metrics alone, continue using BDS or custom bootstrap which works already.

lei-tang:
LGTM.

zirain:
cannot make an agreement, close this first.

PR #2553, opened 2022-11-07
security policy graduation to v1

aryan16:
This PR involves API changes required to graduate security policies from V1beta1 to V1 and some documentation changes for authorization policy.


security/v1/jwt.proto
```diff
@@ -0,0 +1,157 @@
+// Copyright 2019 Istio Authors
```
howardjohn, line 1: We should be adding the `istio.io/sync-from` indicators like we have for networking protos so we don't need to manually keep them in sync.

howardjohn (commented):
Can we split out new docs from the rest of the changes? They are ~unrelated and make it harder to review

security/v1/jwt.proto
aryan16: done, thanks

howardjohn (approved):

linsun (changes requested):
Only concern I have is the JWT.proto - some of the changes are too recent, for example ClaimToHeaders were changed few days ago and this feature is still relatively new IMHO, I don't feel comfortable promoting them to stable.  Others are fine.

Thoughts?

aryan16:
cc : @liminw

linsun (approved):
Thanks @aryan16 for clarifying the claimtoheader. Given we don't have a better way to track each field other than the comment, LGTM

PR #2556, opened 2022-11-09
telemetry: add support for reporting_interval

zirain:
for https://github.com/istio/istio/issues/41763

allow end-users config `tcp_reporting_duration` via Telemetry API

in v1.17 `istio.stats` will use native filter, the default value is `5s` as kuat's suggestions: https://github.com/istio/istio/issues/41763#issuecomment-1306090046


nrjpoddar:
Is there ay other user facing impact to switching to the native filter? Since this config option only applies to TCP metrics we need to evaluate where to add this field in the API.

zirain:
> Is there ay other user facing impact to switching to the native filter? 

IMO no, but @kyessenov can confirm this.

kyessenov:
1. Native vs wasm - the difference is where the timer is installed. In Wasm, it's one big timer per worker, in native, it's per individual stream. The duration applies to both.
2. 5s is reasonable because that's the default envoy flush interval. Anything faster won't help.
3. gRPC stream metrics are also reported continuously for HTTP, but that's a little known feature.

kyessenov:
LGTM, users are asking for this, and we need to stop EnvoyFilter method.
Can you rename it to "reporting_interval"? We _may_ use this for long duration HTTP streams.

zirain:
> LGTM, users are asking for this, and we need to stop EnvoyFilter method. Can you rename it to "reporting_interval"? We _may_ use this for long duration HTTP streams.

done

telemetry/v1alpha1/telemetry.proto
```diff
@@ -396,6 +397,11 @@ message Metrics {
   // a best practice to list any universal overrides first, with tailored
   // overrides following them.
   repeated MetricsOverrides overrides = 2;
+  
+  // Optional. Allows configuration of the time between calls out to for metrics reporting.
+  // This only work for TCP metrics, we may use this for long duration HTTP streams in the future.
+  // The default duration is `5s`.
```
nrjpoddar, line 403:
```suggestion
  // Optional. Reporting interval allows configuration of the time between calls out to for metrics reporting.
  // This currently only supports TCP metrics but we may use this for long duration HTTP streams in the future.
  // The default duration is `5s`.
```

nrjpoddar (approved):

telemetry/v1alpha1/telemetry.proto
```diff
@@ -396,6 +397,11 @@ message Metrics {
   // a best practice to list any universal overrides first, with tailored
   // overrides following them.
   repeated MetricsOverrides overrides = 2;
+  
+  // Optional. Reporting interval allows configuration of the time between calls out to for metrics reporting.
+  // This currently only supports TCP metrics but we may use this for long duration HTTP streams in the future.
```
ericvn, line 402: Maybe:
```suggestion
  // Optional. Reporting interval allows configuration of the time between calls for metrics reporting.
  // This currently supports TCP metrics only, but we may use this for long duration HTTP streams in the future.
```

howardjohn (approved):

PR #2561, opened 2022-11-11
update authz policy docs

aryan16:


security/v1beta1/authorization_policy.proto
```diff
@@ -110,6 +110,28 @@ import "type/v1beta1/selector.proto";
 //         methods: ["POST"]
 // ```
 //
+// The following is another example that sets `action`` to "DENY" to create a deny policy.
```
howardjohn, line 113:
```suggestion
// The following is another example that sets `action` to `DENY` to create a deny policy.
```

security/v1beta1/authorization_policy.proto
```diff
@@ -110,6 +110,28 @@ import "type/v1beta1/selector.proto";
...
+// It denies requests all the "POST" method request on port "8080" on all workloads
```
howardjohn, line 114:
```suggestion
// It denies all requests with the "POST" method on port "8080" on all workloads
```

security/v1beta1/authorization_policy.proto
```diff
@@ -110,6 +110,28 @@ import "type/v1beta1/selector.proto";
...
+// in the "foo" namespace.
```
howardjohn, line 115:
```suggestion
// in the `foo` namespace.
```

security/v1beta1/authorization_policy.proto
```diff
@@ -110,6 +110,28 @@ import "type/v1beta1/selector.proto";
...
+//   action: DENY
+//   rules:
+//   - to:
+//     - operation:
+//         methods: ["POST"]
+//         ports: ["8080"]
+//
+// If we remove the ports from the rules of the above-mentioned deny policy, then it will just have the http attribute
+// in the rules and it will deny all the traffic on raw tcp ports. It's better to always add ports in the deny policy
+// if you don't want to deny all the TCP traffic. [Authorization Policy for TCP Ports] (https://istio.io/latest/docs/tasks/security/authorization/authz-tcp/)
```
howardjohn, line 133: This is misleading IMO. The port doesn't behave differently for TCP and non-TCP.

I would say

When this rule is applied to TCP traffic, the `method` field (as will all HTTP based attributes) cannot be processed. For a `DENY` rule, missing attributes are treated as matches. This means all TCP traffic on port 8080 would be denied in the example above. If we were to remove the `ports` match, _all_ TCP traffic would be denied. As a result, its recommended to always scope `DENY` policies to a specific port, especially when using HTTP attributes

security/v1beta1/authorization_policy.proto
aryan16: I followed how it is represented in all the parts of the docs. It was "", but I think it's better to highlight that. Updated as per your comment. Thanks

howardjohn (approved):

security/v1beta1/authorization_policy.proto
```diff
@@ -110,6 +110,29 @@ import "type/v1beta1/selector.proto";
...
+//   action: DENY
+//   rules:
+//   - to:
+//     - operation:
+//         methods: ["POST"]
+//         ports: ["8080"]
+//
+// When this rule is applied to TCP traffic, the `method` field (as will all HTTP based attributes) cannot be processed. 
+// For a `DENY` rule, missing attributes are treated as matches. This means all TCP traffic on port 8080 would be denied in the example above.
+// If we were to remove the `ports` match, all TCP traffic would be denied. As a result, its recommended to always scope `DENY` policies to a specific port,
```
kfaseela, line 133: nit: As a result, it is recommended

security/v1beta1/authorization_policy.proto
```diff
@@ -110,6 +110,29 @@ import "type/v1beta1/selector.proto";
...
+//   rules:
+//   - to:
+//     - operation:
+//         methods: ["POST"]
+//         ports: ["8080"]
+//
+// When this rule is applied to TCP traffic, the `method` field (as will all HTTP based attributes) cannot be processed. 
+// For a `DENY` rule, missing attributes are treated as matches. This means all TCP traffic on port 8080 would be denied in the example above.
+// If we were to remove the `ports` match, all TCP traffic would be denied. As a result, its recommended to always scope `DENY` policies to a specific port, 
+// especially when using HTTP attributes [Authorization Policy for TCP Ports] (https://istio.io/latest/docs/tasks/security/authorization/authz-tcp/)
```
ericvn, line 134:
```suggestion
// especially when using HTTP attributes [Authorization Policy for TCP Ports] (https://istio.io/latest/docs/tasks/security/authorization/authz-tcp/).
```

ericvn (approved):
Thanks for the updates.

PR #2565, opened 2022-11-15
Enable configuring Envoy QAT private key provider

rveerama1:
Brief background: Enabling PrivateKeyProvider through ProxyConfig was originally introduced in https://github.com/istio/api/pull/2261 for CryptoMB. This PR is an addition for that. This will provide configuration for enabling QAT private key provider.

QAT support in Envoy is merged and available from https://github.com/envoyproxy/envoy/pull/21984. 

Next generation Intel® QAT support with Intel® Xeon® Scalable processors will feature an Intel® QAT cryptography and compression acceleration engine.

QAT private key provider extension will use qatlib library (https://github.com/intel/qatlib) to accelerate RSA operations in handshakes. The extension will look a bit like the existing cryptomb private key provider. The use case is to move the expensive cryptographic operations away from the CPU to the accelerator device, leaving CPU cycles for other use.

Additional Description:
Support for Intel® QAT is already present in the mainline Linux kernel and in Kubernetes device plugins (to expose the device files to containers). There are previous generations of Intel QAT® hardware devices, but they are not supported by this extension.


howardjohn:
I thought I saw somewhere QAT only supports OpenSSL. but Envoy uses boringssl. Did I understand wrong, or is this somehow enabling openssl?

ipuustin:
> I thought I saw somewhere QAT only supports OpenSSL. but Envoy uses boringssl. Did I understand wrong, or is this somehow enabling openssl?

QAT can be used with both OpenSSL and BoringSSL. In BoringSSL the QAT support is done using private key methods, which are hooks for overriding BoringSSL private key operations. Envoy contrib has a a QAT private key provider extension, which uses these hooks for accelerating RSA sign/decrypt operations.

rveerama1:
@howardjohn ping

howardjohn (approved):

mesh/v1alpha1/proxy.proto
```diff
@@ -306,13 +306,25 @@ message PrivateKeyProvider {
     // However, if the queue is not filled before the delay has expired, the requests already in the queue
     // are processed, even if the queue is not full.
     // In effect, this value controls the balance between latency and throughput.
-    // The duration needs to be set to a non-zero value.
+    // The duration needs to be set to a value greater than or equal to 1 millisecond.
+    google.protobuf.Duration poll_delay = 1;
+  }
+
+  // QAT PrivateKeyProvider configuration
```
linsun, line 313: Can you specify in comment what QAT stands for?

linsun (commented):
LGTM, minor comment

mesh/v1alpha1/proxy.proto
rveerama1: fixed.

PR #2570, opened 2022-11-17
copy jwt claim to header

aryan16:
Updated the api to support the newly added feature in envoy - https://github.com/envoyproxy/envoy/pull/23444


security/v1beta1/jwt.proto
```diff
@@ -155,3 +168,13 @@ message JWTHeader {
...
   string prefix = 2;
 }
+
+// This message specifies the detail for copying claim to header.
+message ClaimToHeader {
+  // The name of the header to be created. The header will be overridden if it already exists in the request.
+  string header = 1;
+
+  // The name of the claim to be copied from. Only claim of type string/int/bool is supported.
+  // The header will be set to an empty value if the claim does not exist or the type of the claim is not supported.
```
howardjohn, line 178: Do we want to set empty or not set the header at all?

security/v1beta1/jwt.proto
```diff
@@ -143,6 +143,19 @@ message JWTRule {
...
+
+  // This field specifies a list of operations to copy the claim to HTTP headers on a successfully verified token.
+  // This differs from the `output_payload_to_header` by allowing outputting individual claims instead of the whole payload.
+  // The header specified in each operation in the list must be unique.
+  // ```
+  //   outputClaimToHeaders:
+  //   - header: x-my-company-jwt-group
+  //     claim: my-group
+  //   - header: x-test-environment-flag
+  //     claim: test-flag
```
howardjohn, line 156: are nested expressions supported?

security/v1beta1/jwt.proto
aryan16: oh yes, it won't set anything if claim is invalid. I will update the language.

security/v1beta1/jwt.proto
aryan16: yes, will add an example for that. Thanks

costinm (approved):

howardjohn (approved):

PR #2573, opened 2022-11-21
consistent hash: add warnings

howardjohn:
I have helped on >10 issues from these confusion, so hoping to clear things up.


linsun (approved):
Thanks for the clarification

hzxuzhonghu (approved):

Issue #2501, opened 2022-10-10
WasmPlugin.Priority field should be an int32 to avoid failing on creating the resource with Go lang client


jonathanvila:
(NOTE: This is used to report product bugs:
  To report a security vulnerability, please visit <https://istio.io/about/security-vulnerabilities/>
  To ask questions about how to use Istio, please visit <https://discuss.istio.io>
)

**Bug description**
Currently WasmPlugin proto definition contains the field `Priority` of type Int64. 
When using Protobuf with Go lang to apply an instance of a WasmPlugin to Kubernetes, this marshal will set the value as a string, quoting the contents.
https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L739
https://github.com/golang/protobuf/issues/1414

This will make Kubernetes to complain and not allowing to apply the resource.

Also considering that a Pod.Priority is an int32 field, I wonder why an int64 is needed to express the plugin priority.

**Affected product area (please put an X in all that apply)**

[X] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience

**Expected behavior**
Set the numeric value for WasmPlugin.Priority and no errors expected.

**Steps to reproduce the bug**
Execute this code : https://gist.github.com/acbbae4ba6a00768b1e7c21db8454d91

**Version (include the output of `istioctl version --remote` and `kubectl version`)**
istio : 1.15.1
kubectl : Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.2", GitCommit:"5835544ca568b757a8ecae5c153f317e5736700e", GitTreeState:"clean", BuildDate:"2022-09-21T14:25:45Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7

**How was Istio installed?**
helm

**Environment where bug was observed (cloud vendor, OS, etc)**
kind clusters in GCP

Additionally, please consider attaching a [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) by attaching
the dump file to this issue.

howardjohn:
same as https://github.com/istio/istio/issues/41256, we should use the same solution for both

ericvn:
Reopening as this isn't fully merged yet. The changes haven't made it to istio/istio.

ericvn:
[...]

PR #2577, opened 2022-11-23
Changing the WasmPlugin.Priority field from int64 to int32

jonathanvila:
Fixes: https://github.com/istio/api/issues/2501

Protobuf is generating an int64 into a string, and this is generating issues with regular marshallers in order to generate a K8s resource from a proto, because K8s expects `priority` to be a number not a string.

Also considering that this is a Priority, looks like the possibility to need an int64 is not a common one but a remote one, and also considering that Envoy filter priority field is an int32.

### DIFF
--- a/extensions/v1alpha1/wasm.pb.go
+++ b/extensions/v1alpha1/wasm.pb.go
@@ -489,7 +489,7 @@ type WasmPlugin struct {
 	// If `priority` is not set, or two `WasmPlugins` exist with the same
 	// value, the ordering will be deterministically derived from name and
 	// namespace of the `WasmPlugins`. Defaults to `0`.
-	Priority *wrappers.Int64Value `protobuf:"bytes,10,opt,name=priority,proto3" json:"priority,omitempty"`
+	Priority *wrappers.Int32Value `protobuf:"bytes,10,opt,name=priority,proto3" json:"priority,omitempty"`
 	// Configuration for a Wasm VM.
 	// more details can be found [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto#extensions-wasm-v3-vmconfig).
 	VmConfig *VmConfig `protobuf:"bytes,11,opt,name=vm_config,json=vmConfig,proto3" json:"vm_config,omitempty"`
@@ -597,7 +597,7 @@ func (x *WasmPlugin) GetPhase() PluginPhase {
 	return PluginPhase_UNSPECIFIED_PHASE
 }
 
-func (x *WasmPlugin) GetPriority() *wrappers.Int64Value {
+func (x *WasmPlugin) GetPriority() *wrappers.Int32Value {
 	if x != nil {
 		return x.Priority
 	}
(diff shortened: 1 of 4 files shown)


howardjohn (commented):
I am not sure this is a compatible protobuf change, won't the message encoding be different?

jonathanvila:
> I am not sure this is a compatible protobuf change, won't the message encoding be different?

You mean that already existing messages for WasmPlugin will not be compatible with this new change ?

jonathanvila:
@howardjohn following this doc about protobuf looks like int32 and int64 are compatible:
https://developers.google.com/protocol-buffers/docs/proto#updating

```int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility. If a number is parsed from the wire which doesn't fit in the corresponding type, you will get the same effect as if you had cast the number to that type in C++ (for example, if a 64-bit number is read as an int32, it will be truncated to 32 bits).```

**Comment by howardjohn** [comment]
Its not an int64 though its a Int64Value

**Comment by jonathanvila** [comment]
> Its not an int64 though its a Int64Value

I know, but they are the same wrappers for an internal field that is the one that changes.......
```
message Int64Value {
  // The int64 value.
  int64 value = 1;
}
```
vs
```
message Int32Value {
  // The int32 value.
  int32 value = 1;
}
```

**Comment by jonathanvila** [comment]
@howardjohn is there anything we could do to progress in this PR ?

**Review by hzxuzhonghu — APPROVED** [review]

**Review by kebe7jun — APPROVED** [review]

**Comment by ericvn** [comment]
Note that this broke the pipeline: https://prow.istio.io/view/gs/istio-prow/logs/update_client-go_dep_client-go_postsubmit/1661344222390784000:
```
pilot/pkg/model/addressmap.go:1: : # istio.io/istio/pilot/pkg/model [istio.io/istio/pilot/pkg/model.test]
pilot/pkg/model/push_context.go:1862:17: cannot use prio.Value (variable of type int32) as int64 value in assignment
pilot/pkg/model/push_context.go:1866:17: cannot use prio.Value (variable of type int32) as int64 value in assignment
pilot/pkg/model/push_context_test.go:478:15: cannot use &wrappers.Int64Value{…} (value of type *wrapperspb.Int64Value) as *wrapperspb.Int32Value value in struct literal
pilot/pkg/model/push_context_test.go:486:15: cannot use &wrappers.Int64Value{…} (value of type *wrapperspb.Int64Value) as *wrapperspb.Int32Value value in struct literal
pilot/pkg/model/push_context_test.go:502:15: cannot use &wrappers.Int64Value{…} (value of type *wrapperspb.Int64Value) as *wrapperspb.Int32Value value in struct literal
pilot/pkg/model/push_context_test.go:514:15: cannot use &wrappers.Int64Value{…} (value of type *wrapperspb.Int64Value) as *wrapperspb.Int32Value value in struct literal
pilot/pkg/model/push_context_test.go:521:15: cannot use &wrappers.Int64Value{…} (value of type *wrapperspb.Int64Value) as *wrapperspb.Int32Value value in struct literal
pilot/pkg/model/push_context_test.go:539:15: cannot use &wrappers.Int64Value{…} (value of type *wrapperspb.Int64Value) as *wrapperspb.Int32Value value in struct literal
pilot/pkg/model/push_context_test.go:557:15: cannot use &wrappers.Int64Value{…} (value of type *wrapperspb.Int64Value) as *wrapperspb.Int32Value value in struct literal (typecheck)
```

howardjohn:
I think that is ok, we just need to manually update it. We only offer
yaml+proto compatibility which this change meets, just breaks Go.

PR #2585, opened 2022-11-28
Extend DestinationRule with retry budget

quraynai:


networking/v1alpha3/destination_rule.proto
```diff
@@ -791,6 +791,9 @@ message ConnectionPoolSettings {
     // Note that when this is set to true, h2_upgrade_policy will be ineffective i.e. the client
     // connections will not be upgraded to http2.
     bool use_client_protocol = 7;
+
+    // Specifies a limit on concurrent retries in relation to the number of active requests. This parameter is optional.
+    RetryBudget retry_budget = 8;
```
howardjohn, line 796: It seems confusing to configure retries in virtual service, but a retry budget in destination rule. Wouldn't a user configure those together?

quraynai: @howardjohn 
Retry in virtualservice I think it's for configuring route.

But base on envoy document this is for configuring cluster. 
cluster.circuit_breakers.thresholds.retry_budget

Also other fields of HTTPSettings will try to configure
cluster.circuit_breakers.thresholds.*

I think it might be reasonable to put it here.

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/circuit_breaker.proto

quraynai:
@howardjohn hi can you help review this again?

networking/v1alpha3/destination_rule.proto
linsun: I agree with @howardjohn 's concern on DR. With your porposal, what if user configures retry budget on DR but forgot to configure retry in VS?

quraynai: @linsun 
I understand that this will make user to maintain 2 different resources (VS, DR) in case if they want to include retry_budget

but not sure if we move this configuration to VS will cause another confusion to user or not?
due to DR is for configuring cluster and VS is for configuring route

if user create VS user has to check into envoy config about route only. then user has to check for route and cluster instead

and another point is retry_budget from istio is about playing around with max_retries that already existed in DR.
also base on envoy Circuit Breaker other related config is already in DR as well
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/circuit_breaker.proto#envoy-v3-api-msg-config-cluster-v3-circuitbreakers-thresholds-retrybudget
https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings

quraynai: also this will make implementation on istio/istio more align with current implementation
https://github.com/istio/istio/pull/42232/files

hzxuzhonghu: CircuitBreakers_Thresholds.MaxRetries can  be used to limit the concurrent retries.  `RetryBudget` seems an advanced option. So I wonder this is not a must

ramaraochavali: I think this is not to restrict max retries but allowing additional retries even if the retry count in route exceeds if cluster config allows. Currently if retry count exceeds, we just stop retrying. This allows users to configure retries well  instead of giving a static retry count (and stop after that is reached), allow retries as a percentage of current active requests. I think there is value in providing this configuration, especially since it is [difficult](https://github.com/istio/istio/issues/27419) to configure this via Envoy filter.
 
While I agree with the concern of confusion caused by defining it at DR vs. on VS. -  I think there is no better place to add this to VS because it spawns across routes and primarily a property of destination. We just have to document this more clearly and refer it to VS docs from here.

quraynai: could you please help review again?
Already added document to link with HttpRetry from VirtualService

quraynai: @ramaraochavali @linsun @howardjohn @hzxuzhonghu 
hi could you please help review again?

ramaraochavali: This seems reasonable to me

hzxuzhonghu: I think this is a confusing feature. At least the relationship with CircuitBreakers_Thresholds and route.RetryPolicy is not obvious.

ramaraochavali: > route.RetryPolicy is not obvious. 

This is part is documented in comments.

> CircuitBreakers_Thresholds

circuit breaker thresholds are the max allowed limits defined statically.  Beyond which no more retries would be allowed. This field allows retries to be a function of current active request count which is more useful. 
BTW, there is a plan to deprecate max_retries in CB and move it to retry budgets in Envoy it self https://github.com/envoyproxy/envoy/issues/10869

quraynai: @ramaraochavali @linsun @howardjohn @hzxuzhonghu
Can you help on this MR again?
We kinda need this functionality to solve problem about retry storm.
Currently we are building our forked version to run in our company. Ideally we want this to be supported officially from istio. then moving forward we can keep upgrading version without maintain our forked one as well.

Kinda want this to be released with 1.17

quraynai: @ramaraochavali @linsun @howardjohn @hzxuzhonghu
sorry to tag again really need help on this PR

ramaraochavali (approved):

networking/v1alpha3/destination_rule.proto
linsun: Is there any validation work to alert users other than what is doc-ed here where user may forget to configure VS retry?

A few other concerns:
- How is this impact ambient sidecarless?
- GAMMA API doesn't have destination rule, is it possible you use EnvoyFilter for as a stop gap solution?

ramaraochavali: > Is there any validation work to alert users other than what is doc-ed here where user may forget to configure VS retry?

Since this is cross resource - not sure validation can work. But we can have IstioAnalayze flag this inconsistency?

>GAMMA API doesn't have destination rule, is it possible you use EnvoyFilter for as a stop gap solution?

Envoy filter while works is not a practical approach. See https://github.com/istio/istio/issues/27419#issuecomment-1322925405 - we need to replace entire circuit breaker

quraynai: @linsun 
from @ramaraochavali replied. Is it okay to have it on istioAnalyze instead?

quraynai: @linsun hi could you please help check this MR again?

quraynai: @linsun can you help on this? 1.17 is already released. we will try to aim for 1.18 instead.
In our company we are running 1.16 forked to support retry budget at the moment.
we really want this functionality support from istio and then we can aim for next istio upgrade.

quraynai: @howardjohn @hzxuzhonghu can you help on this?

quraynai: @linsun @howardjohn @hzxuzhonghu could you please help review this again?
About confusion of retry is it okay to go with IstioAnalayze instead?

PR #2592, opened 2022-12-12
fix: envoy statistics config documentation

chinaran:
refer: https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats
associate to: https://github.com/istio/istio.io/pull/12357

1. Istio uses `outlier_detection` for `circuit_breakers`
2. There are 6 retry metrics (`cluster.<name>.upstream_rq_retry*`) 
3. There are many upstream connections metrics (`cluster.<name>.upstream_cx_*`) 
4. Added `upstream_rq_timeout` suffixed example

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -558,15 +558,16 @@ message ProxyConfig {
   // (`sidecar.istio.io/statsInclusionPrefixes`,
   // `sidecar.istio.io/statsInclusionRegexps`, and
   // `sidecar.istio.io/statsInclusionSuffixes`). For example, to enable stats
-  // for circuit breaker, retry, and upstream connections, you can specify stats
-  // matcher as follow:
+  // for circuit breaker, retry, upstream connections and request timeout,
+  // you can specify stats matcher as follow:
   // ```yaml
   // proxyStatsMatcher:
   //   inclusionRegexps:
-  //     - .*circuit_breakers.*
-  //   inclusionPrefixes:
-  //     - upstream_rq_retry
-  //     - upstream_cx
+  //     - .*outlier_detection.*
+  //     - .*upstream_rq_retry.*
+  //     - .*upstream_cx_.*
+  //   inclusionSuffixes:
+  //     - upstream_rq_timeout
   // ```
   // Note including more Envoy stats might increase number of time series
   // collected by prometheus significantly. Care needs to be taken on Prometheus
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -558,8 +558,8 @@ message ProxyConfig {
   // (`sidecar.istio.io/statsInclusionPrefixes`,
   // `sidecar.istio.io/statsInclusionRegexps`, and
   // `sidecar.istio.io/statsInclusionSuffixes`). For example, to enable stats
-  // for circuit breaker, retry, upstream connections and request timeout,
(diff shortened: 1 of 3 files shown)


zirain (approved):
thanks for update the examples

mesh/v1alpha1/proxy.proto
```diff
@@ -558,15 +558,16 @@ message ProxyConfig {
...
   // `sidecar.istio.io/statsInclusionRegexps`, and
   // `sidecar.istio.io/statsInclusionSuffixes`). For example, to enable stats
-  // for circuit breaker, retry, and upstream connections, you can specify stats
-  // matcher as follow:
+  // for circuit breaker, retry, upstream connections and request timeout,
+  // you can specify stats matcher as follow:
   // ```yaml
   // proxyStatsMatcher:
   //   inclusionRegexps:
-  //     - .*circuit_breakers.*
```
ericvn, line 566: You mention circuit breaker, but then it seems to be removed in this change. Outlier detection is added.

mesh/v1alpha1/proxy.proto
```diff
@@ -558,15 +558,16 @@ message ProxyConfig {
   // (`sidecar.istio.io/statsInclusionPrefixes`,
   // `sidecar.istio.io/statsInclusionRegexps`, and
   // `sidecar.istio.io/statsInclusionSuffixes`). For example, to enable stats
-  // for circuit breaker, retry, and upstream connections, you can specify stats
-  // matcher as follow:
+  // for circuit breaker, retry, upstream connections and request timeout,
+  // you can specify stats matcher as follow:
```
ericvn, line 562: Some of these are singular and some plural. Maybe best to stick with one. Possibly:
```suggestion
  // for circuit breakers, request retries, request timeouts, and upstream connections,
  // you can specify stats matcher as follows:
```

mesh/v1alpha1/proxy.proto
```diff
@@ -558,15 +558,16 @@ message ProxyConfig {
...
   // ```yaml
   // proxyStatsMatcher:
   //   inclusionRegexps:
-  //     - .*circuit_breakers.*
-  //   inclusionPrefixes:
-  //     - upstream_rq_retry
-  //     - upstream_cx
+  //     - .*outlier_detection.*
```
ericvn, line 566: Do we mention somewhere that Istio uses outlier detection for circuit breakers? Or would it be could to maybe at a note here?

mesh/v1alpha1/proxy.proto
chinaran: I agree with you, but the order here is the same as that of the example.

mesh/v1alpha1/proxy.proto
chinaran: The conclusion comes from this issue: https://github.com/istio/istio/issues/9375.
And Istio isable envoy circuit breaking by default (https://github.com/istio/istio/pull/17346)

mesh/v1alpha1/proxy.proto
ericvn: Cn you update the singular vs plural. Maybe: 
```suggestion
  // for circuit breakers, request retries, upstream connections, and request timeouts,
  // you can specify stats matcher as follows:
```

chinaran: @ericvn Thanks for your suggesstion.

ericvn (approved):

PR #2605, opened 2022-12-19
remove parent shutdown duration

ramaraochavali:
As discussed in https://github.com/istio/istio/pull/42253. This is no longer used. So removing it.


howardjohn (approved):

ericvn (approved):

ericvn:
Post-submit issue with client-go: https://prow.istio.io/view/gs/istio-prow/logs/update_client-go_dep_client-go_postsubmit/1604884861187788800

ericvn:
Maybe just need https://github.com/istio/istio/pull/42253 to be merged.

costinm:
Can someone confirm ( and maybe add a test ) that a user config with this
setting will still work ?

We should probably add a test with absolutely all configurable options set
to non-default values, and keep it around without removing anything.

I think this one is safe - it's in a config map, and we ignore unknowns (
until someone decides to change the impl to be more strict and we forget
about what we removed...).
Removing fields from CRDs may be more dangerous due to validation.

ramaraochavali:
> Maybe just need https://github.com/istio/istio/pull/42253 to be merged.

Yes. That should be merged

PR #2613, opened 2022-12-29
fix: the default of concurrency for sidecar and gateway is different.

j2gg0s:


dhawton:
@j2gg0s Changes should be made against the master branch, then if it's a bug fix backported to the release branches.

dhawton (changes requested):
Changes should be made against the master branch, then if it's a bug fix backported to the release branches.

PR #2615, opened 2023-01-02
Clarify that Sidecar is not applicable to gateways

jewertow:
Signed-off-by: Jacek Ewertowski <jewertow@redhat.com>

To avoid confusion like in this [issue](https://github.com/istio/istio/issues/32041), I think it's worth to clarify that `Sidecar` is not applicable to gateways. Users may guess that `Sidecar` should work for gateways, because both gateway and sidecar proxies are the same istio-proxy.


costinm (approved):

ericvn:
Verifying the html:
Screenshot 2023-01-03 at 8 06 01 AM

howardjohn (approved):

PR #2616, opened 2023-01-03
fix: default concurrency for sidecar and gateway is different.

j2gg0s:


j2gg0s:
@dhawton thanks for help.

I made new pull request aginst branch master.

networking/v1beta1/proxy_config.proto
```diff
@@ -114,8 +114,9 @@ message ProxyConfig {
   istio.type.v1beta1.WorkloadSelector selector = 1;
 
   // The number of worker threads to run.
-  // If unset, defaults to 2. If set to 0, this will be configured to use all cores on the machine using
-  // CPU requests and limits to choose a value, with limits taking precedence over requests.
+  // For sidecar, defaults to 2.
+  // If set to 0, this will be computed from sidecar's CPU requests and limits, with limits taking precedence over requests.
+  // For gateway, defaults to the number of hardware threads on the machine.
```
howardjohn, line 119: Does gateway not follow the logic on L118?

j2gg0s: Gateway will not be set concurrency, according to my understanding.

Gateway's injection template:
https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/files/gateway-injection-template.yaml#L26

howardjohn: hmm... I think its supposed to

j2gg0s: > hmm... I think its supposed to

im not good in english, you mean we should change this behavior? @howardjohn

PR #2621, opened 2023-01-05
Fix text indent

tomoyk:
I found a typo of text-indent in HTML.

image

### DIFF
--- a/networking/v1alpha3/envoy_filter.pb.html
+++ b/networking/v1alpha3/envoy_filter.pb.html
@@ -111,22 +111,22 @@ spec:
     patch:
       operation: INSERT_BEFORE
       value: # lua filter specification
-       name: envoy.filters.http.lua
-       typed_config:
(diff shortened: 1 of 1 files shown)


networking/v1alpha3/envoy_filter.pb.html
```diff
@@ -111,8 +111,8 @@
     patch:
       operation: INSERT_BEFORE
       value: # lua filter specification
-       name: envoy.filters.http.lua
-       typed_config:
+        name: envoy.filters.http.lua
+        typed_config:
```
ericvn, line 115: This change is in a generated file (run 'make gen'). The actual change should be done here: https://github.com/istio/api/blob/master/networking/v1alpha3/envoy_filter.proto#L139-L140, and after the `make gen` the html file should be updated.

howardjohn (commented):
The html is autogenerated, the .proto file needs to be updated then `make gen` will update html

PR #2626, opened 2023-01-11
Separate ServiceEntry port from Gateway Server and Sidecar ingress listener

hzxuzhonghu:
target port is only used by ServiceEntry, However when share a  single `Port`, it is very error-prone for users to set targetPort for Gateway and Sidecar.

In this pr, i define a separate `ServicePort`, quite similar with `Port`, just has an additional target port


networking/v1alpha3/service_entry.proto
```diff
@@ -850,7 +849,7 @@ message ServiceEntry {
   // The ports associated with the external service. If the
   // Endpoints are Unix domain socket addresses, there must be exactly one
   // port.
-  repeated Port ports = 3 [(google.api.field_behavior) = REQUIRED];
+  repeated ServicePort ports = 3 [(google.api.field_behavior) = REQUIRED];
```
howardjohn, line 852: This change is both JSON and wire backwards compatible, right? Since it has the same format?

howardjohn (commented):
LGTM, this is a great improvement to avoid confusion

howardjohn (approved):

ericvn (approved):

costinm:
I don't mind the addition - but not sure the removal of target_port in the beta gateway API is not a breaking change.

What happens if a user was using it ? I don't see how it'll not be broken ( unless targetPort was not implemented ). Anything that requires user to make config changes or breaks a documented feature in beta API requires a full deprecation process and a lot of overhead. Not sure if it's worth it ( we can just add the new stuff and keep the old
one around )

hzxuzhonghu:
@costinm Good question. In theory it should be compatible. But will verify in real env

hzxuzhonghu:
@costinm Veirified in real env， it has no side effects. The above theory is correct, istio --> kube-apiserver -> etcd, all use json marshal and unmarshal for CRs

costinm:
Thanks for checking - at least it won't break. But any chance some is using
that field and expect Gateway to select the targetPort  ?

In other words - did we ever implement it, and if so what happened to the
impl ? If it was not ever implemented - great to get rid of it.

hzxuzhonghu:
No, we never use it in istio

networking/v1alpha3/gateway.proto
```diff
@@ -605,9 +605,8 @@ message Port {
   // Label assigned to the port.
   string name = 3 [(google.api.field_behavior) = REQUIRED];
 
-  // The port number on the endpoint where the traffic will be
-  // received. Applicable only when used with ServiceEntries.
-  uint32 target_port = 4;
+  reserved 4;
```
howardjohn, line 608: I think we may need to mark this as deprecated and not remove it. Even with `reserved`, the CRD does not have it. The CRD rejects unknown fields. So someone could have set targetPort (even though it does nothing) and upgrade is broken.

Alternatively, we could modify the CRD generator to include `reserved` fields, but that seems harder

hzxuzhonghu: Upgrade istio will not need to create Gateway again. Not sure I understand what you mean.

hzxuzhonghu: It looks good to me to deprecate it though

howardjohn: I mean a user can currently set `targetPort` in a Gateway YAML. After upgrade it will start being rejected, blocking their CI/CD etc

PR #2634, opened 2023-01-16
Deprecate Port.TargetPort

hzxuzhonghu:
Without this, when we create a gateway with targetPort specified, it will error.


howardjohn (approved):

howardjohn:
can we add a warning in the webhook if its set (even better: automatically detect any deprecated field usage?? Not sure we can do that)

ericvn (approved):

PR #2635, opened 2023-01-17
mark Lightstep provider deprecated

zirain:
xref: https://github.com/istio/istio/issues/40027


mesh/v1alpha1/config.proto
```diff
@@ -487,10 +487,8 @@ message MeshConfig {
       // Configures a tracing provider that uses the Zipkin API.
       ZipkinTracingProvider zipkin = 4;
       // Configures a Lightstep tracing provider.
-      // Note: For Istio 1.15+, configuring this provider will result in
-      // using an OpenTelemetryTracingProvider configured specially for 
-      // Lightstep. This is part of the Lightstep transition to OpenTelemetry.
-      LightstepTracingProvider lightstep = 5;
+      // Deprecated: For Istio 1.15+, please use an OpenTelemetryTracingProvider instead, more details can be found at https://github.com/istio/istio/issues/40027
+      LightstepTracingProvider lightstep = 5 [deprecated=true];
```
hzxuzhonghu, line 491: need to hide from doc

zirain: I'd like to mark as deprecated in 1.17, then hided from doc in 1.18.

hzxuzhonghu: We usually make them together

zirain: updated

hzxuzhonghu (approved):

ericvn:
Should there be a release note if we are deprecating? Maybe it will be elsewhere?

zirain:
> Should there be a release note if we are deprecating? Maybe it will be elsewhere?

updated

linsun (approved):
Thx for adding the release note

PR #2645, opened 2023-01-20
RFC on the proto change

qudongfang:
I want to work on this issue https://github.com/istio/istio/issues/34169 by having a global `default_http2_protocol_options` in the global mesh config, which applies to both H2 listeners and clusters.

I tried to use `EnvoyFilter`, and it's working by having an `echo_h2_enabled: true` label on all the `echo` service provider pods and `echo` service consumer pods. e.g,
```
kind: EnvoyFilter
metadata:
  name: echo-h2
spec:
  configPatches:
  - applyTo: CLUSTER
    match:
      cluster:
        portNumber: 8080
    patch:
      operation: MERGE
      value:
        typed_extension_protocol_options:
          envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
            '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
            explicit_http_config:
              http2_protocol_options:
                max_concurrent_streams: 500
                initial_stream_window_size: 65536
                initial_connection_window_size: 1048576
  - applyTo: NETWORK_FILTER
    match:
      listener:
        portNumber: 8080
    patch:
      operation: MERGE
      value:
        typed_config:
          '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
          http2_protocol_options:
            max_concurrent_streams: 500
            initial_stream_window_size: 65536
            initial_connection_window_size: 1048576
  workloadSelector:
    labels:
      echo_h2_enabled: true
```

Things get worse when there're lots of H1 and H2 services in the mesh.

It would be great to have a global `default_http2_protocol_options`  as the community did for the [global HTTP retry policy](https://github.com/istio/api/pull/2058).

I drafted this PR to get some insights from the community.

I'm new to istio, but noticed that we're not adding the [envoy data-plane-api](https://github.com/envoyproxy/data-plane-api) as a dependency. I wonder about the reasons behind the decision. For this specific change, I think it would make things easier if we can import the envoy data-plane-api protocol directly. 

Any suggestions/guidance would be appreciated. 
 

Draft istiod changes https://github.com/istio/istio/compare/master...qudongfang:istio:ISSUE-34169

Similar PR https://github.com/istio/istio/pull/35506


howardjohn (changes requested):
I understand the use case but I don't think we should do this. It is a slippery slope to have 100s of special cases for every aspect of the Envoy API

costinm:
I'm not sure I understand the use case - but agree that we shouldn't copy
the entire Envoy API.

Is this something that can/will be supported by ztunnel ? If not - even
stronger reason to not do it, or restrict any new features to gateways only.
I would like to see a freeze on any API or feature applying to sidecars -
if it is not feasible for ambient.

linsun:
I think I understood the user case, but curious if we have many users requesting it. We added http retry policy globally as there are many user requests for it.

With ambient slicing L4 and L7, we would need to understand if it can be supported for ztunnel as well like @costinm asked.

kyessenov:
Large windows will improve latency but will expose proxies to memory overflows.
Same with the number of streams - we'll get faster latency but the head of line blocking over h2 TCP will be worse.

costinm:
Curious - how much is customizable in Quic ? 

Total concurrent requests is clearly needed ( rate limit), but not specific to h2. And any adjustments are per workload, can't be global. 

Like advanced TLS settings - it requires a lot of expertise and easy to mess up - if we decide to add, may be worth keeping them in a separate 'vendor extension' and not mix with common and safe settings. I'm not saying no, just need to be careful.

costinm:
Right - but is this something that a typical application owner will
understand and correctly configure ?

Using annotations as API is usually wrong - but there are cases where it is
useful to separate the rarely used and hard to
use configs.

I wrote a H2 implementation and used quite a few others - but I would have
a lot of trouble picking the right
number of streams or window size. The few times I tried - I was
actually wrong, and broke something else...

kyessenov:
I think the original request is to allow the platform owner to set protocol settings. What works well on GCP, may not work well on the edge clusters and vice versa. The only way to find out is experimentation and guidance, but the controls need to be there to begin with.

costinm:
This kind of settings don't make a lot of sense on a global platform level.
The H2 defaults are good enough for the common case.

For a very specific workload doing un-common stuff - yes, it makes sense to
fine tune them if you really know what you're doing.

costinm:
We also have plenty of other 'expert only' controls in the env variable
settings. Nobody disagrees (I hope) that expert users
who know what they're doing should have a way to customize stuff (
including in ztunnel ). But I hope we agree that
we can't really support an API where all the expert customizations are part
of the core API.

hzxuzhonghu:
We will meet many other requests for adding global settings to tune traffic.  IMo we can design a new API like `GlobalTrafficPolicy` to include all the global traffic settings.

Now what we have in `MeshConfig` is connectTimeout, TCP keepalive, localityLbSetting,H2UpgradePolicy, DefaultHttpRetryPolicy, etc. 

MeshConfig exist as configmap, it contains all kinds of config and it is actually not as efficient as CR. By splitting traffic related global config out, we can add more advanced options.

qudongfang:
I'm looking for an easier way to tune these H2 options other than `EnvoyFilter`, not necessarily to be a global configuration.
 `EnvoyFilter` is not very easy to work with in this case, I'll explain in detail.

Any suggestions/solutions would be great.

We have a grpc service with a small payload size. 

# Where the H2 options need to be set

There are 3 phases of H2 happening here.

1. client locally: client to local istio-proxy sidecar listener
2. client to server: local istio-proxy sidecar outbound cluster to remote istio-proxy sidecar inbound listener
3. server locally: remote istio-proxy inbound cluster(PassthroughCluster) to the server process

If we go with the **only** available solution(`EnvoyFilter`) at the moment, we need to apply H2 changes to 

- client-side local listener: protect client-side istio-proxy from client container/process
- client-side outbound cluster: limit client-side istio-proxy, don't overload server-side istio-proxy
- server-side virtual-inbound listener:  protect server-side istio-proxy from client-side istio-proxy
- server-side PassthroughCluster: limit server-side istio-proxy, don't overload server container/process

## Configure H2 options with `EnvoyFilter`

The `EnvoyFilter` would be 1 for the client side and 1 for the server side for each service.

```
---
# server side
kind: EnvoyFilter
metadata:
  name: echo-h2
  namespace: echo-dev
spec:
  configPatches:
  - applyTo: CLUSTER
    match:
      cluster:
        name: "inbound|8080||"
    patch:
      operation: MERGE
      value:
        typed_extension_protocol_options:
          envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
            '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
            explicit_http_config:
              http2_protocol_options:
                max_concurrent_streams: 500
                initial_stream_window_size: 65536
                initial_connection_window_size: 1048576
  - applyTo: NETWORK_FILTER
    match:
      listener:
        filterChainMatch:
            destinationPort: 8080           
    patch:
      operation: MERGE
      value:
        typed_config:
          '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
          http2_protocol_options:
            max_concurrent_streams: 500
            initial_stream_window_size: 65536
            initial_connection_window_size: 1048576
  workloadSelector:
    labels:
      app: echo
```

```
---
# client side
kind: EnvoyFilter
metadata:
  name: echo-h2
  # namespace: istio-system without workloadSelector or each of the consumer namespaces?
spec:
  configPatches:
  - applyTo: CLUSTER
    match:
      cluster:
        portNumber: 8080
        service: "echo-grpc.baas-dev.us-east-1.aws.dev.kube"
    patch:
      operation: MERGE
      value:
        typed_extension_protocol_options:
          envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
            '@type': type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
            explicit_http_config:
              http2_protocol_options:
                max_concurrent_streams: 500
                initial_stream_window_size: 65536
                initial_connection_window_size: 1048576
  - applyTo: NETWORK_FILTER
    match:
      listener:
# catchall traffic of this port(e.g `ListenerMatch. portNumber:8080`) listener, couldn't tune H2 separately per host/DNS
# https://github.com/istio/api/blob/522813ce750f222c8504315946eaf27925a07f4b/networking/v1alpha3/envoy_filter.proto#L707 
        portNumber: 8080
    patch:
      operation: MERGE
      value:
        typed_config:
          '@type': type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
          http2_protocol_options:
            max_concurrent_streams: 500
            initial_stream_window_size: 65536
            initial_connection_window_size: 1048576
  workloadSelector:
    labels:
      echo-h2-client: true
```

As we can see, `EnvoyFilter` works fine for tuning the server-side H2 options. 
While things are so not nice with client-side: Which namespaces to apply it to? as a global `EnvoyFilter` in the root namespace? or apply to all namespaces that need to consume this `echo-h2` service?

Things get complicated pretty quickly with more and more services onboarding. 

## What we gain from setting the H2 options

With the default H2 options, the memory usage of Pods of B is around **2GB**, default value listed below:
- Http2ProtocolOptions.max_concurrent_streams: 2147483647 (2^31 - 1)
- Http2ProtocolOptions.initial_stream_window_size: 256MB
- Http2ProtocolOptions.initial_connection_window_size: Currently, this has the same minimum/maximum/default as initial_stream_window_size. 256MB

With the following H2 options, the memory usage of Pods B is under **500 MB**, and the rps(average about 3.3K rps) is the same as without the H2 options, rps got more stabilized too.

- Http2ProtocolOptions.max_concurrent_streams: 500
- Http2ProtocolOptions.initial_stream_window_size: 64KB
- Http2ProtocolOptions.initial_connection_window_size: 1MB

## Other unmatured semi-opinions/semi-arguments

- An Istio cluster is most likely local to a data center, and the networking condition is preset, only the payload size changes from service to service within the same Istio deployment.
- Envoy official doc suggests some [H2 option values for the edge](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge), though internal is different with the edge, but still, the default H2 option values seem to be too much.

Envoy official doc suggestions for the edge:
> [HTTP/2 maximum concurrent streams limit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-max-concurrent-streams) and [HTTP/3 maximum concurrent streams limit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-quicprotocoloptions-max-concurrent-streams) to 100
> [HTTP/2 initial stream window size limit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-initial-stream-window-size) to 64 KiB,
> [HTTP/2 initial connection window size limit](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-initial-connection-window-size) to 1 MiB.

The default H2 options,:
- Http2ProtocolOptions.max_concurrent_streams: 2147483647 (2^31 - 1)
- Http2ProtocolOptions.initial_stream_window_size: 256MB
- Http2ProtocolOptions.initial_connection_window_size: Currently, this has the same minimum/maximum/default as initial_stream_window_size. 256MB

costinm:
Thanks for the example. I think the 'edge' use case is very good and it
aligns with my argument that such settings should not be 'global'
in MeshConfig, but specific to a deployment/pod. And you bring up a very
good point that even for a pod, it's not one setting,
but 3 ( inbound, envoy-to-workload on localhost so can use very large
numbers, no packet loss, and outbound as client).

With ambient and gateways - we'll also have a pair of settings for the
waypoint, but since the H2 listener can be shared by multiple servers -
we may have a problem on which setting to use.

I do agree this is a kind of advanced setting we should support - my
concern is how to future-proof it and avoid
confusion/miss-use.

My preference was to start with an annotation - what are your concerns with
it ?
And once GAMMA matures - it is something we can likely agree with other
vendors, and it can be a proper CRD
while we do gain some feedback and experience with it.

PR #2646, opened 2023-01-20
Clarify protocol description for Gateway port

kfaseela:

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -600,8 +600,6 @@ message Port {
   // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|TCP|TLS.
   // TLS can be either used to terminate non-HTTP based connections on a specific port
   // or to route traffic based on SNI header to the destination without terminating the TLS connection.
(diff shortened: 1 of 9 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -598,8 +598,10 @@ message Port {
 
   // The protocol exposed on the port.
   // MUST BE one of HTTP|HTTPS|GRPC|HTTP2|MONGO|TCP|TLS.
-  // TLS implies the connection will be routed based on the SNI header to
-  // the destination without terminating the TLS connection.
+  // TLS can be either used to terminate non-HTTP based connections on a specific port
```
howardjohn, line 601: In Zhonghu's PR, we split the Port type. I wonder if we should do that again, and make a GatewayPort and SidecarPort. That way we can clearly explain the semantics based on its usage. Because the VS comment, for example, doesn't make sense for Sidecar

kfaseela: I removed the TCPRoute Note.

howardjohn (approved):

ericvn (approved):

PR #2659, opened 2023-01-28
VirtualService docs minor formatting fixes

ufUNnxagpM:
The double asterisks currently cause the text to be italicized instead of showing as expected

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -253,7 +253,7 @@ message VirtualService {
   repeated string gateways = 2;
 
   // An ordered list of route rules for HTTP traffic. HTTP routes will be
-  // applied to platform service ports named 'http-\*'/'http2-\*'/'grpc-\*', gateway
+  // applied to platform service ports using HTTP/HTTP2/GRPC protocols, gateway
   // ports with protocol HTTP/HTTP2/GRPC/TLS-terminated-HTTPS and service
   // entry ports using HTTP/HTTP2/GRPC protocols.  The first rule matching
   // an incoming request is used.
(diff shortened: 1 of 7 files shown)


zirain (approved):

hzxuzhonghu (approved):

ericvn:
Still doesn't appear to be correct.

Before:
Screenshot 2023-01-29 at 4 32 03 PM

After:
Screenshot 2023-01-29 at 4 32 55 PM

networking/v1alpha3/virtual_service.proto
```diff
@@ -253,8 +253,8 @@ message VirtualService {
   repeated string gateways = 2;
 
   // An ordered list of route rules for HTTP traffic. HTTP routes will be
-  // applied to platform service ports named 'http-*'/'http2-*'/'grpc-*', gateway
-  // ports with protocol HTTP/HTTP2/GRPC/ TLS-terminated-HTTPS and service
+  // applied to platform service ports named 'http-\*'/'http2-\*'/'grpc-\*', gateway
```
howardjohn, line 256: I would wrap it in code quotes , then we don't need to escape either. `http-*`

ufUNnxagpM: I think it could also be changed to
```suggestion
  // applied to platform service ports using http/http2/grpc protocols, gateway
```
so it can match with the [protocol selection page](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection) which has other ways to set the protocol (at least on k8s).

ericvn: I like @ufUNnxagpM's suggestion, but we should match the casing in of the protocols in the rest of the sentence, so change protocols to uppercase.

howardjohn (approved):

PR #2660, opened 2023-01-29
settings for OCSP Stapling on Gateways

sebltm:


howardjohn (changes requested):
Thanks for the PR! Can you link the design doc in the PR description and make sure we get approval on that before proceeding?

PR #2661, opened 2023-01-30
telemetry: add MetricDefinition

zirain:
before promote Telemetry API to beta, Telemetry API must support all features EnvoyFilter done.

cc @kyessenov


costinm:
Not sure why it would need to support all features - we don't do this in any other API.

I would wait a bit on promoting - there are discussions on converging with OTel and reducing the costs.

telemetry/v1alpha1/telemetry.proto
```diff
@@ -654,3 +658,25 @@ message AccessLogging {
...
+  // Gauge metric type.
+  GAUGE = 1;
+
+  // Histogram metric type.
+  HISTOGRAM = 2;
+}
+
+message MetricDefinition {
+  // Name of the metric.
+  string name = 1;
```
costinm, line 675: How would this even work ? Is the name one of the istio-defined names ? If not, how would we deal with a change of name or type in the rest of the system ? 

And what feature or use case does it address ?

linsun: Some examples on this would be nice.

telemetry/v1alpha1/telemetry.proto
```diff
@@ -654,3 +658,25 @@ message AccessLogging {
   // requests/connections for logging.
   Filter filter = 3;
 }
+
+enum MetricType {
```
linsun, line 662: can we have an example for this as well?

kyessenov:
I imagine this discussion is going to happen multiple times. The problem this is trying to solve is removing the requirement to use EnvoyFilter for an old feature (since 1.5 I think?). We never restricted the scope of EnvoyFilter to a subset of Envoy configs, and we cannot sustainably support EnvoyFilter API. So there's little choice but to expose the entire surface of Envoy configs as Istio configs. Do you have any other idea how to deal with this problem?

telemetry/v1alpha1/telemetry.proto
```diff
@@ -402,6 +402,10 @@ message Metrics {
   // This currently only supports TCP metrics but we may use this for long duration HTTP streams in the future.
   // The default duration is `5s`.
   google.protobuf.Duration reporting_interval = 3;
+
+  // Optional. List of definitions of metrics to be generated.
```
howardjohn, line 406: Does this require proxy config changes to the stats tagging config?

howardjohn:
My 2c - Istio's goal is not to support the entire Envoy config surface via Istio APIs. So while it may be tedious to have this discussion every time, it is critical to ensure that we are exposing the right APIs. Too many and the system becomes to complex; too few and its not useful.

I disagree, at least, with the assessment that Telemetry API needs to have the full featureset of EnvoyFilter to become Beta; the importance is stability of a core set of APIs, not "Support every possible feature".

kyessenov:
> My 2c - Istio's goal is not to support the entire Envoy config surface via Istio APIs. So while it may be tedious to have this discussion every time, it is critical to ensure that we are exposing the right APIs. Too many and the system becomes to complex; too few and its not useful.

Right, but a similar discussion happens in Envoy as well. Envoy has no interest in adding features noone uses. There's no concern about API stability, their v3 is de facto forever.

> I disagree, at least, with the assessment that Telemetry API needs to have the full featureset of EnvoyFilter to become Beta; the importance is stability of a core set of APIs, not "Support every possible feature".

Telemetry API is alpha, and we can deprecate parts of it. We don't need to promote the entire Istio API to stable. We have no visibility into EnvoyFilter usages because of PII, so we can't even get usage patterns without having some syntax support.

A couple of more points. We haven't been able to make progress for many years in this area and continue punting on EnvoyFilter for core use cases. At some point, someone needs to make a movement here, since EnvoyFilter is not an answer.

costinm:
On Mon, Jan 30, 2023 at 8:56 AM Kuat ***@***.***> wrote:

> My 2c - Istio's goal is not to support the entire Envoy config surface via
> Istio APIs. So while it may be tedious to have this discussion every time,
> it is critical to ensure that we are exposing the right APIs. Too many and
> the system becomes to complex; too few and its not useful.
>
> Right, but a similar discussion happens in Envoy as well. Envoy has no
> interest in adding features noone uses. There's no concern about API
> stability, their v3 is de facto forever.
>
With Gateway/GAMMA - I think it is important to have a core set of APIs
that can be used with any proxy (and supportable in ztunnel).

It is also useful to have an API that allows exposign Envoy unique or
uncommon features.

EnvoyFilter is not that API - but that doesn't mean we have to copy ( or
worse - slightly change ) each envoy API.

We are waiting for Envoy Gateway to define vendor extensions - hopefully
using K8S GAMMA patterns of attaching to HttpRoute or Gateway,
and if they do - we can adopt them as well, and any other Envoy-based
product can.

The problem with EnvoyFilter is not only that it's exposing all APIs - but
also structural, it is too tied to internal Istio generated config, names
and structure
and too powerful (patching) to be supportable.




> I disagree, at least, with the assessment that Telemetry API needs to have
> the full featureset of EnvoyFilter to become Beta; the importance is
> stability of a core set of APIs, not "Support every possible feature".
>
> Telemetry API is alpha, and we can deprecate parts of it. We don't need to
> promote the entire Istio API to stable. We have no visibility into
> EnvoyFilter usages because of PII, so we can't even get usage patterns
> without having some syntax support.
>
> A couple of more points. We haven't been able to make progress for many
> years in this area and continue punting on EnvoyFilter for core use cases.
> At some point, someone needs to make a movement here, since EnvoyFilter is
> not an answer.
>

IMO Envoy community should own how Envoy integrates with Gateway/GAMMA and
how Envoy configs can be safely attached to routes/gateways/etc.
Nginx and other proxies should do the same.

We are (slowly) adding features that have a lot of interest from users -
and punting to EnvoyFilter is IMO the correct answer for the rest.
Istio should not be tied to a particular proxy or implementation, in
particular for sidecar.

Costin

kyessenov:
@costinm Are you suggesting we stop the telemetry API promotion and instead focus on Gateway-based APIs for it?
The value in telemetry API is that we can finally abandon using EnvoyFilter by default on GCP implementation, for example. Eventually, we'll have to migrate EnvoyFilters to Gateway, and it's much easier to migrate from Telemetry API.

costinm:
No, I'm suggesting promoting telemetry API as is, and not add more
features.

After it is promoted and stable - depending on what happens with OTel
migration, ambient, Gateway we can evaluate what and
where to add more features.

kyessenov:
@zirain are you OK with delaying custom metrics to a later stage of telemetry API?

costinm:
Do we even have a design doc with how the 'custom metrics' will be
supported in Ambient ?
So far I still don't understand how will they show up - are we talking
about existing metrics that envoys generate ? What are they measuring ?
And how does it relate with the list of documented metrics we support -
which itself is out of sync with what OTel and others are using
and far too large ?

If the proposal was to remove metrics and reduce cardinality, or match OTel
- I would be far more supportive than adding more metrics
we'll need to support and implement in ambient.

kyessenov:
@costinm Custom metrics are just like metric customizations - they override the metric value and the metric name. You can measure anything application-specific (number of items in GET requests) as well.  This is a metric production mechanism, you can output these metrics to other services (metrics service or any other sink).

In general, we have no idea to what extent this is used in production since it's all hidden by EnvoyFilters.

costinm:
File a bug - ask users to add comments/votes to it if they are using it and
want it.

No idea if anyone is using it doesn't mean we should add it. And I still
don't understand the use case and what kind of metrics we would add :-)
Maybe there is some common  metric we are missing and we should add -
instead of arbitrary names and metrics ?

kyessenov:
@costinm The whole point of customization is to enable bespoke tail-end cases. So it's not going to be one universal missing metric unfortunately.  Filing an issue doesn't work - majority of interactions with prod cases are they never check github, they just want things working as before.

costinm:
If we don't know about the use cases - we can't test it or support them.
Envoy may support/test them - EnvoyFilters still work and are clearly
marked as dangerous and not supportable. If users want features added to
Istio API and supported by Istio - we need to know more about
them. Either customers, or vendors working with them need to provide some
info - and how 'tail end' those users are. I don't think few very
advanced users continuing to use EnvoyFilter until a replacement is
available is a problem as big as surfacing arbitrary customization without
knowing how useful it is, and 'supporting' it long term without
understanding what it does.

kyessenov:
The issue we have now is that vendors simply wrap Istio APIs (often using EnvoyFilter) into another set of APIs, so all the UX concerns are addressed at the outer layer (just like we do with xDS). This is not a bad idea for a component, so I don't think we should expect users to change the behavior and break through the glass. But this is not a good interface model, since supporting EnvoyFilter is terrible on both sides. Everyone would benefit from a more reliable mechanism over EnvoyFilter. Stability is not a concern, xDS is forever and has unfortunately become part of Istio APIs from now on.

zirain:
IIRC, `MetricDefinition` is used by many end users, @christian-posta have a nice blog about it https://blog.christianposta.com/understanding-istio-telemetry-v2/.

> @zirain are you OK with delaying custom metrics to a later stage of telemetry API?

if all toc members agree with that will not blocking Telemetry API promotion, I agree with this.

zirain:
From https://github.com/istio/istio/issues/39772#issuecomment-1183631452, Telemetry API cannot work together with `EnvoyFilter` .

PR #2674, opened 2023-02-10
ambient upstream: bring mesh config to `master`

howardjohn:
This is part of https://github.com/istio/istio/issues/40879. See that issue description for context.

This introduces a new configuration to enable/disable ambient mesh. Enablement means that traffic is captured.

There are 3 modes: OFF, ON, and label selector.

To enable admins to globally enforce the mesh is enabled but skip a few namespaces, a disable selector is present which restricts the scope of ON mode.


mesh/v1alpha1/config.proto
```diff
@@ -1123,7 +1123,50 @@ message MeshConfig {
...
+    enum AmbientMeshMode {
+      // Sets AmbientMesh to a namespaced configuration. This is the default.
+      // Namespaces that have the label `istio.io/dataplane-mode=ambient` set.
+      DEFAULT = 0;
+
+      // Sets AmbientMesh to be applied to all namespaces.
+      ON = 1;
+
+      // Disables AmbientMesh for the traditional sidecar approach.
+      OFF = 2;
```
costinm, line 1137: what will happen to workloads that have the label ? 

Not sure we need this, I don't remember sidecars having an 'off'

mesh/v1alpha1/config.proto
```diff
@@ -1123,7 +1123,50 @@ message MeshConfig {
...
+    //   - matchLabels:
+    //       env: dev
+    //   - matchExpressions:
+    //     - key: app
+    //       operator: In
+    //       values:
+    //         - cassandra
+    //         - spark
+    // ```
+    repeated k8s.io.apimachinery.pkg.apis.meta.v1.LabelSelector disabled_selectors = 3;
```
costinm, line 1162: Is this a namespace selector ? Workload selector ( I see 'app' in the example ) ?

Does it have to be so complicated, are there cases for this kind of matching ?
How about a list of namespaces to exclude (and maybe one of namespaces to include - as alternative to the label) ? 

Note that we didn't generally allowed arbitrary label selectors for sidecar injection enable/disable - 
but well defined labels.

costinm (commented):
I'm a bit concerned with this. I agree on having an 'on', and that may need a way to turn it off.

But what does turning off really mean ? Don't we have controls on the CNI capture ? How is it different ? ZTunnel will still run - I assume turning off means capture will not redirect to ztunnel but instead redirect to sidecar or nothing ? 

It would be great to see a more consistent API oriented towards controlling the capture - 
not with namespace granularity but even at port level. That may take a while - meanwhile an 'on' and a list of namespaces may be simpler and safer.

costinm:
Another option is to keep this as an option on the CNI controller installer. 

Most of MeshConfig options are consumed by Istiod ( and the few that aren't probably should be moved). 
I would also add that a LOT of MeshConfig doesn't make sense for ambient - we have a subset of the API and features,
and at some point may use a lighter/cleaned up Istiod. 

The CNI daemon set will hopefully be common and shared, and part of the installation it should be able to control
what is captured ?

mesh/v1alpha1/config.proto
howardjohn: I am not quite sure I follow - for sidecars we have sidecar.istio.io/inject=false? Or simply not labeling the namespace.

When redirection is not enabled, you are right ztunnel will still run (mentioned above), since it is node based, but there will be no redirection so it will not do anything

mesh/v1alpha1/config.proto
howardjohn: The intent behind this was to allow *enforced* redirection, something we don't actually allow today with sidecars. So an admin could introduce a hard requirement that cannot be overriden by applications.

I am not too opposed to keeping this more like sidecars for now, especially given we don't have egress authz policies

howardjohn:
> 

Currently this is also read by istiod but I had some ideas to stop doing this. Let me see if that is feasible

costinm:
I am not opposed to 'enforced' - just to arbitrary matchers and complexity
:-)

"ON" is as enforced as any CNI plugin doing encryption - there is no opt
out from wireguard CNI.
I don't think I've seen any good reason to be more complex or flexible than
an encrypting CNI plugin
at namespace level.

I do believe we need better controls ( access controlled via RBAC, enforced
) for interception - replacing the
annotations that exclude ports and ranges from interception by CNI/init
container. We can start in MeshConfig,
but this is something we should allow delegation ( with RBAC control). I've
been thinking of a proposal for gRPC,
but there are other cases.

The actual use case for opt-out if ambient is ON is to have some special
workloads talk with workloads outside
of the mesh, or not be intercepted. But the granularity is port or workload
- not namespace. And that's more of
a 'interception control API' rather than an install API ( Meshconfig is
still mainly used for install - since it is
still overridden on upgrade ).

I wouldn't mind at all if whatever we do here is aligned and consistent
with sidecars - if you add a list of namespaces
to enforce or exclude in MeshConfig ( as an alternative to labeling the
namespace, controlled by mesh admin ) - we could
probably use the same mechanism for sidecars.

The other concern is: do we need this for v1 ? I think we rushed with 'on
by default' for sidecars as well, shouldn't we
just focus on the gaps that prevent v1 ambient for shipping, and gradually
add more features based on feedback/etc ?
Again - no problem with ON.

Final note - even if it is ON, it would still exclude workloads labeled
with the sidecar injection annotation. Worth mentioning
this in the comments. As well as what happens if istio 'do not inject'
label is set - I think it is intent to not use istio at all ?

mesh/v1alpha1/config.proto
linsun: OFF could be useful as admin says ambient has not gone through all reviews and don't want any team to use it.

I assume this would overwrite the namespace level config?

mesh/v1alpha1/config.proto
linsun: Yes this is namespace selector, i think this was proposed so that it has similar style as discovery selectors.

cc @dhawton

PR #2680, opened 2023-02-16
telemetry: allow disable 128bit trace id for zipkin provider

zirain:
xref: https://github.com/istio/istio/issues/43359

Allows users to be compatible with existing zipkin backend.


Stono:
👏 thank you for doing this!

mesh/v1alpha1/config.proto
```diff
@@ -684,6 +684,10 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. A 128 bit trace id will be used in Istio.
+      // If true, will result in a 64 bit trace id being used.
+      bool Disable128BitTraceId = 4;
```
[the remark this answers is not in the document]
zirain: good idea!

mesh/v1alpha1/config.proto
```diff
@@ -684,6 +684,10 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. A 128 bit trace id will be used in Istio.
+      // If true, will result in a 64 bit trace id being used.
+      bool Enable64BitTraceId = 4;
```
howardjohn, line 690: nit: use underscore casing like the rest.

Do we expect other formats to be present in the future? an ENUM may be more extensible if so?

zirain: for what I can see no, also I encourage people move to OpenTelemetry Tracing if there's no backport issue.

howardjohn: ok then this is probably fine

zirain:
cc @kyessenov @lei-tang, PTAL

lei-tang (approved):

howardjohn (approved):

PR #2683, opened 2023-02-16
Fix link

ericvn:
Missed another spot with an extra space.


ericvn:
One outstanding issue is that we have the same location called out in two files:

location: https://istio.io/docs/reference/config/security/request_authentication.html

AndreaM12345 (approved):

PR #2693, opened 2023-02-24
Clarify ecdh_curves api docs

kfaseela:
Small clarification to the doc to align better with the implementation PR: https://github.com/istio/istio/pull/43207

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -1142,8 +1142,9 @@ message MeshConfig {
   // Mesh mTLS does not respect ECDH curves.
   TLSConfig mesh_mTLS = 63;
   //
-  // Configuration of TLS for mesh external traffic(i.e. the traffic entering or leaving the mesh).
-  // This includes any TLS mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL settings, meshMTLS configuration should be used.
+  // Configuration of TLS for mesh external traffic.
+  // This includes any TLS traffic mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL traffic settings,
+  // meshMTLS configuration should be used.
   // Note: Mesh external does not respect min TLS version configured here currently.
   TLSConfig mesh_external_TLS = 64;
 
--- a/mesh/v1alpha1/config.pb.go
+++ b/mesh/v1alpha1/config.pb.go
@@ -856,8 +856,9 @@ type MeshConfig struct {
 	// Configuration of mTLS for traffic between workloads within the mesh.
 	// Mesh mTLS does not respect ECDH curves.
 	MeshMTLS *MeshConfig_TLSConfig `protobuf:"bytes,63,opt,name=mesh_mTLS,json=meshMTLS,proto3" json:"mesh_mTLS,omitempty"`
-	// Configuration of TLS for mesh external traffic(i.e. the traffic entering or leaving the mesh).
-	// This includes any TLS mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL settings, meshMTLS configuration should be used.
+	// Configuration of TLS for mesh external traffic.
+	// This includes any TLS traffic mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL traffic settings,
+	// meshMTLS configuration should be used.
 	// Note: Mesh external does not respect min TLS version configured here currently.
 	MeshExternal_TLS *MeshConfig_TLSConfig `protobuf:"bytes,64,opt,name=mesh_external_TLS,json=meshExternalTLS,proto3" json:"mesh_external_TLS,omitempty"`
 }
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
@@ -632,8 +632,9 @@ No
 <td><code>meshExternalTLS</code></td>
 <td><code><a href="#MeshConfig-TLSConfig">TLSConfig</a></code></td>
 <td>
-<p>Configuration of TLS for mesh external traffic(i.e. the traffic entering or leaving the mesh).
-This includes any TLS mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL settings, meshMTLS configuration should be used.
+<p>Configuration of TLS for mesh external traffic.
+This includes any TLS traffic mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL traffic settings,
+meshMTLS configuration should be used.
 Note: Mesh external does not respect min TLS version configured here currently.</p>
 
 </td>


kfaseela:
cc @ramaraochavali

howardjohn (approved):

kfaseela:
cc @hzxuzhonghu

mesh/v1alpha1/config.proto
```diff
@@ -1142,8 +1142,9 @@ message MeshConfig {
   // Mesh mTLS does not respect ECDH curves.
   TLSConfig mesh_mTLS = 63;
   //
-  // Configuration of TLS for mesh external traffic(i.e. the traffic entering or leaving the mesh).
-  // This includes any TLS mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL settings, meshMTLS configuration should be used.
+  // Configuration of TLS for mesh external traffic.
+  // This includes any TLS traffic mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL traffic settings,
+  // meshMTLS configuration should be used.
```
hzxuzhonghu, line 1147: Does this mean `mesh_mTLS` will only take effect on `ISTIO_MUTUAL`? That would be incorrect to me

kfaseela: @howardjohn : Both @ramaraochavali and @hzxuzhonghu thinks the implementation for ecdh_curves is very much similar to the config of TLS versions in mesh_mTLS, and it will be confusing to end user to put ecdh outside in a separate "mesh_external_TLS". Which is looking correct to me, when there was no easy way to distinguish between internal and external traffic for the mesh, and the only thing we are distinguishing is traffic modes. And the API doc which i currently added also, as @hzxuzhonghu pointed out would make confusions for people who were using mesh_MTLS for non ISTIO_MUTUAL cases for setting TLS versions. Would it be okay for you, if I remove the mesh_external field, and use ecdh curves inside mesh_mTLS itself?

kfaseela: @hzxuzhonghu : I ended up adding a separate config all because of line 1141 : which says mesh_MTLS is only for traffic between workloads within the mesh. So is the line wrong?

hzxuzhonghu: As L1131 shows `// The below configuration parameters can be used to specify TLSConfig for in-mesh and mesh-external TLS traffic.`

kfaseela: @howardjohn : Both Rama and Zhonghu thinks adding a separate ExternalTLS is confusing to end user, and it should be merged with mesh_mTLS. Please let me know your opinion. The current mesh_MTLS, even though the doc says, it is only for workloads within the mesh, seems to be used in all cases? Though I did not see that in the code for clusters.

ramaraochavali: I think we use mesh_mTLS to build inbound listeners when sidecar does not have TLS settings

kfaseela: @ramaraochavali : so looks like even mesh_mTLS docs are not properly matching with the implementation, if someone assumes that the TLS version set here is going to be set for outbound cases as well. As i told I added separate parameter, thinking this is something that cannot be used for ingress/egress gateway kind of scenarios. @howardjohn shall i then use ecdh_curves from within mesh_mTLS itself, putting proper API docs to clarify the behavior?

ramaraochavali: > so looks like even mesh_mTLS docs are not properly matching with the implementation, if someone assumes that the TLS version set here is going to be set for outbound cases as well

In general TLS version is enforced at the server level to cater to all services (both sidecar and sidecar less). so the the current implementation is correct except that we need to document the fact that `minTLSVersion` field is honoured only if that workload does not have server TLS settings specified via Sidecar. So the TLS version is added for all TLS types.

But your current PRs statement that mesh_TLS is only used for ISTIO_MUTUAL would be incorrect. you need to fix that. That is the confusion I was talking about earlier- some settings (like minTLS version are applied irrespective of TLS type) and now for ecdh_curves we are introducing a new field.

kfaseela: @ramaraochavali : so you still mean ecdh_curves should be moved inside mesh_mTLS with proper documentation, right?

ramaraochavali: That would be my preference because mesh_mTLS fields (tls version) is already being applied to all TLS modes and would be less confusing for users.

howardjohn: > But your current PRs statement that mesh_TLS is only used for ISTIO_MUTUAL would be incorrect

Can you show in the code wheremesh_mTLS is used for non- ISTIO_MUTUAL? The one place its used, afaik, is https://github.com/istio/istio/blob/53f24f14f874dc9c29f212fe7ec99d0024511e49/pilot/pkg/security/authn/v1beta1/policy_applier.go#L195 which is only ISTIO_MUTUAL

ramaraochavali: > Or are we saying that ISTIO_MUTUAL is only an outbound property, and for inbound "Istio mTLS" is a different thing?

Yes. This was my point. We do it based on [**MutualTLS** mode](https://github.com/istio/api/blob/master/security/v1beta1/peer_authentication.proto#L155) in PeerAuthentication rather than whether ISTIO_MUTUAL or not.

I do not still get the concern of merging these together and documenting clearly that `ecdh_curves` is only applicable for SIMPLE/MUTUAL.

hzxuzhonghu: @howardjohn I do feel mesh_tls and mesh_external_tls are not good API. Not to mention other users, we experienced developers can not clearly explain them to others.

I am tracking the old mesh_mtls design https://docs.google.com/document/d/1rU-zoeXlTsbuKosYcLIqcglb03tmDxo_9yDdw8WvEbw/edit?resourcekey=0-T8xYrkDOjdGl0AcTNjvagg#

IIRC, it is intended to configure any mesh internal traffic not only mtls.

howardjohn: > IIRC, it is intended to configure any mesh internal traffic not only mtls.

I don't think that is the case, if it was intended to apply to all traffic it would be be called mesh_mTLS - the *m* implies we are talking about 

> > Or are we saying that ISTIO_MUTUAL is only an outbound property, and for inbound "Istio mTLS" is a different thing?
> 
> Yes. This was my point. We do it based on [MutualTLS mode](https://github.com/istio/api/blob/master/security/v1beta1/peer_authentication.proto#L155) in PeerAuthentication rather than whether ISTIO_MUTUAL or not.
> 
> I do not still get the concern of merging these together and documenting clearly that `ecdh_curves` is only applicable for SIMPLE/MUTUAL.
> 
> However, I certainly do not want to block this work because of this. So approved the other PR

I think my use of ISTIO_MUTUAL may be have been a bit inprecise since that only applies to outbound. We clearly have a concept of "Istio mTLS", which is configured by auto_mtls and ISTIO_MUTUAL (outbound) and PA (inbound). mesh_mtls applies to that, and mesh_external_tls applies to everything else. Is my view.

ramaraochavali: Ok. I get your point.

It would have been nice if we have named the "mesh_mTLS" as "mesh_auto_mTLS".

How about calling "mesh_external_mtls" as
- mesh_user_configured_mTLS / mesh_user_specified_mTLS/ mesh_explicit_mTLS or any other that gives meaning that is not auto detected/applied but user configured?

howardjohn: auto_mtls is not very accurate either as "auto mtls" is already a thing that means Destinationrule was not used.

I am not very clear what the hang up here is. There is a very clear view of "Istio mTLS" and "User mTLS" being two separate things. This can be abused to blur the lines (using MUTUAL ("user mtls") to connect to sidecars using "Istio mTLS"), but that is such a niche case it should not pollute our API.


@hzxuzhonghu can you please be a bit more clear about what the objects are? I think @kfaseela feels like this is stalled with no path forward.

hzxuzhonghu: Do you mean `mesh_mTLS` is all for istio mutual TLS and `mesh_external_TLS` for the others?

howardjohn: Yes, that's my understanding/expectation

hzxuzhonghu: `mesh_external_TLS ` is not accurate, as talked above ingress gateway -> sidecar simple tls should use `mesh_external_TLS `. IMO this could not be called `external tls`

howardjohn: I agree "external" is a bit misleading. It's really "not 'mesh mTLS'". I am not sure a great name for that... any ideas?

hzxuzhonghu: How about global tls it will be used anywhere except in istio mtls case.

No matter what name it is called, other people may want to add some more options to it.

ramaraochavali: The easiest path forward is to merge both and just put every thing under `mesh_mTLS` so we do not have to invent a great name for other one. :innocent:

Thinking about this more, even if we do that since we are applying ecdh_curves for SIMPLE TLS that does not make sense. If we exclude SIMPLE TLS that may work.

How about just calling `mesh_tls_settings` or `tls_settings` and document that it is only applied for non Istio MTLS case? Is that less controversial?

kfaseela: How about making the new field, meshExtendedTLS , and have only extended tls settings inside that, and keep existing TLSConfig with versions independent of this. 

```  
  // meshConfig:
  //   meshMTLS:
  //     minProtocolVersion: TLSV1_3
  //   meshExtendedTLS:
  //     Note: currently applicable only for non ISTIO_MUTUAL scenarios.
  //     ecdhCurves:
  //       - P-256
  //       - P-512
```

howardjohn: Maybe?

```
meshMTLS: // Applies to "istio mutual tls"
tlsDefaults: // Applies to anything except "istio mutual tls", can be overriden by Gateway/etc
```

Note things like "minProtocolVersion" is actually useful in tlsDefaults

kfaseela: @howardjohn do you mean I implement only ecdhCurves under tlsDefaults now, and later we extend other params as and when needed?

howardjohn: yes

kfaseela: @hzxuzhonghu and @ramaraochavali : are you both okay with the last proposal from John ? "tlsDefaults"

ramaraochavali: I am fine with that. 
I think "minProtocolVersion" should be supported in both. ecdh_curves can only be in tlsDefaults.

ramaraochavali: >I think "minProtocolVersion" should be supported in both. - 

Mainly because to maintain BC

howardjohn: minProtocolVersion is also useful since we used to support 1.0 min and now set it to 1.2 (in 1.14 I think). So its a good way to revert to the old default.

hzxuzhonghu: I agree with the above suggestion, it eliminates the confusion with comments

kfaseela: Great, thanks :)

mesh/v1alpha1/config.proto
```diff
@@ -1128,24 +1128,23 @@ message MeshConfig {
...
+  // Configuration of mTLS for traffic between workloads with ISTIO_MUTUAL TLS traffic.
   TLSConfig mesh_mTLS = 63;
   //
-  // Configuration of TLS for mesh external traffic(i.e. the traffic entering or leaving the mesh).
-  // This includes any TLS mode except ISTIO_MUTUAL mTLS. For ISTIO_MUTUAL settings, meshMTLS configuration should be used.
-  // Note: Mesh external does not respect min TLS version configured here currently.
-  TLSConfig mesh_external_TLS = 64;
+  // Configuration of TLS for all traffic except for ISTIO_MUTUAL mode.
+  // For ISTIO_MUTUAL TLS settings, use meshMTLS configuration.
+  TLSConfig tls_defaults = 64;
```
hzxuzhonghu, line 1147: If they use the same struct TLSConfig,  `ecdh_curves` should also apply to istio mtls? 
Assume not, we should document it

kfaseela: I have added validation in the implementation PR to guard this, but will clarify the docs as well

kfaseela: @hzxuzhonghu : updated the docs, please have a look now.

hzxuzhonghu: Thanks

PR #2695, opened 2023-02-27
add missing labels and annotations used for ambient

GregHanson:
related issue:
https://github.com/istio/istio/issues/43589


GregHanson:
@howardjohn based on https://github.com/istio/istio/issues/43614 it sounds like this may be premature

howardjohn (commented):
I think we should wait for now before we get some more consensus on these

PR #2702, opened 2023-03-01
make inbound port configurable

zirain:
xref:  https://github.com/istio/istio/issues/43655

cc @howardjohn @ramaraochavali @hzxuzhonghu


mesh/v1alpha1/config.proto
```diff
@@ -35,10 +35,20 @@ option go_package="istio.io/api/mesh/v1alpha1";
...
   int32 proxy_listen_port = 4;
 
+  // Port on which Envoy should listen for all inbound traffic to the pod/vm will be captured to. 
+  // Default port is 15006.
+  // $hide_from_docs
+  int32 proxy_inbound_listen_port = 65;
+
+  // Port on which incoming HBone traffic will be captured.
+  // Default port is 15008.
+  // $hide_from_docs
```
costinm, line 49: Ok with the other 2, they apply to sidecar and config generation which is implemented in istiod.

For hbone:
- I don't think we are ready for the complexity of using a custom port. What happens in multi-cluster if different clusters have different ports ? Is ztunnel and all 'native hbone' aware of mesh config setting ? Would other non-Istio implementations have to somehow get MeshConfig to know the port ?
- if there are strong use cases - it should be an option on the ztunnel chart itself. It is bad practice to share mesh config across multiple independent deployments , mesh config is 'owned' by Istiod chart and used by Istiod.

costinm: Even for the other 2 ports I have doubts - I don't think I've seen a lot of tests using custom envoy interception port, nor a good use case or doc. I remember in the past we discussed about 'promote to beta or deprecate' for configs that linger without tests/docs - this is one example. 

Maybe a doc explaining the use cases and plans for testing/promotion/etc would be best ?

howardjohn: I am very against this

howardjohn (changes requested):

howardjohn:
15001 or 15006 are internal implementation details. 15008 is a public API

mesh/v1alpha1/config.proto
kyessenov: Can someone start IANA process to claim the port? It looks like Azure claimed the same range of ports so every installation there is affected.

zirain: @costinm linked with user's  comment.

zirain:
> 15001 or 15006 are internal implementation details. 15008 is a public API

revert 15008 part as suggested.

howardjohn (approved):
I don't particularly love this type of customization but will admit it brings consistency with 15001 being customizable, so I am ok with this. We should not add HBONE to this at any point, though

hzxuzhonghu (commented):
should better update L1151   // Next available field number: 65

zirain:
> should better update L1151 // Next available field number: 65

done.

hzxuzhonghu (approved):

PR #2714, opened 2023-03-08
update credentialName docs for DestinationRule and Gateway

ufUNnxagpM:
small update to the docs for the `credentialName` field. Main change is referring to the secret as Opaque instead of `generic` (the kubectl subcommand).

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -1093,16 +1093,16 @@ message ClientTLSSettings {
   string ca_certificates = 4;
 
   // The name of the secret that holds the TLS certs for the
-  // client including the CA certificates. Secret must exist in the
-  // same namespace with the proxy using the certificates.
-  // The secret (of type `generic`)should contain the
+  // client including the CA certificates. This secret must exist in
+  // the namespace of the proxy using the certificates.
+  // A secret of type `Opaque` should contain the
   // following keys and values: `key: <privateKey>`,
   // `cert: <clientCert>`, `cacert: <CACertificate>`.
   // Here CACertificate is used to verify the server certificate.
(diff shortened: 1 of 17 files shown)


ericvn (approved):
LGTM

networking/v1alpha3/destination_rule.proto
```diff
@@ -1093,16 +1093,16 @@ message ClientTLSSettings {
...
+  // A secret of type `Opaque` should contain the
   // following keys and values: `key: <privateKey>`,
   // `cert: <clientCert>`, `cacert: <CACertificate>`.
   // Here CACertificate is used to verify the server certificate.
   // For mutual TLS, `cacert: <CACertificate>` can be provided in the
   // same secret or a separate secret named `<secret>-cacert`.
-  // Secret of type tls for client certificates along with
-  // ca.crt key for CA certificates is also supported.
+  // A secret of type `kubernetes.io/tls` for client certificates along
+  // with `ca.crt` key for CA certificates is also supported.
```
howardjohn, line 1105: you cannot have `kubernetes.io/tls` with `ca.crt`.

howardjohn (commented):
This isn't really accurate. Only the key names matter, not the type

networking/v1alpha3/destination_rule.proto
ufUNnxagpM: Updated to just say "TLS" since the key name doesn't have an effect

> you cannot have `kubernetes.io/tls` with `ca.crt`.

I think this should be fine 🤔 k8s only checks for the two required keys ([code](https://github.com/kubernetes/kubernetes/blob/v1.26.0/pkg/apis/core/validation/validation.go#L5809))

howardjohn (approved):

PR #2718, opened 2023-03-09
Document available cipher list for ServerTLSSettings

kfaseela:
Documented as per https://github.com/istio/istio/blob/master/pkg/config/security/security.go#L190
Fixes https://github.com/istio/istio/issues/43212

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -731,10 +731,23 @@ message ServerTLSSettings {
   TLSProtocol max_protocol_version = 8;
 
   // Optional: If specified, only support the specified cipher list.
-  // Otherwise default to the default cipher list supported by Envoy.
-  // The supported list of ciphers are `ECDHE-ECDSA-AES128-GCM-SHA256, ECDHE-RSA-AES128-GCM-SHA256, ECDHE-ECDSA-AES256-GCM-SHA384,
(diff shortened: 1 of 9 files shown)


kfaseela:
cc @kyessenov

networking/v1alpha3/gateway.proto
```diff
@@ -732,5 +732,9 @@ message ServerTLSSettings {
 
   // Optional: If specified, only support the specified cipher list.
   // Otherwise default to the default cipher list supported by Envoy.
+  // The supported list of ciphers are `ECDHE-ECDSA-AES128-GCM-SHA256, ECDHE-RSA-AES128-GCM-SHA256, ECDHE-ECDSA-AES256-GCM-SHA384,
```
howardjohn, line 735: nit: putting in bullet format would be more readable.

Also, ~all vendors offer FIPS builds which this isn't actually true. Maybe we should just make the `supported by Envoy` part a link, since it has this FIPS nuance.

kyessenov: Envoy takes its list from bssl which changes as well :)

kyessenov: In general, we should avoid linking anywhere for API stuff since Istio is responsible for either migrating or preserving compatibility.

kfaseela: @howardjohn I have made the list as bullet format

howardjohn (approved):

ericvn (approved):
Thanks.

PR #2723, opened 2023-03-14
Update failoverPriority doc to reflect the new `key=value` support

hzxuzhonghu:


networking/v1alpha3/destination_rule.proto
```diff
@@ -1240,12 +1240,22 @@ message LocalityLoadBalancerSetting{
 
     // failoverPriority is an ordered list of labels used to sort endpoints to do priority based load balancing.
     // This is to support traffic failover across different groups of endpoints.
-    // Suppose there are total N labels specified:
+    // Two kinds of labels can be specified:
```
ramaraochavali, line 1243: I think it is difficult to read through the text. Can you please give an example for each of these kinds?

hzxuzhonghu: L1244 and L1252

ramaraochavali: I meant similar to how we have example in L1206 can you add example for this new model?

hzxuzhonghu: L1206?

ramaraochavali: https://github.com/istio/api/blob/master/networking/v1alpha3/destination_rule.proto#L1263 - similar to this yaml example

networking/v1alpha3/destination_rule.proto
```diff
@@ -1240,12 +1240,22 @@ message LocalityLoadBalancerSetting{
...
-    // 1. Endpoints matching all N labels with the client proxy have priority P(0) i.e. the highest priority.
-    // 2. Endpoints matching the first N-1 labels with the client proxy have priority P(1) i.e. second highest priority.
-    // 3. By extension of this logic, endpoints matching only the first label with the client proxy has priority P(N-1) i.e. second lowest priority.
-    // 4. All the other endpoints have priority P(N) i.e. lowest priority.
+    //   1. Endpoints matching all N labels with the client proxy have priority P(0) i.e. the highest priority.
+    //   2. Endpoints matching the first N-1 labels with the client proxy have priority P(1) i.e. second highest priority.
+    //   3. By extension of this logic, endpoints matching only the first label with the client proxy has priority P(N-1) i.e. second lowest priority.
+    //   4. All the other endpoints have priority P(N) i.e. lowest priority.
+    //
+    // - Specify labels with key and value `[key1=value1, key2=value2, key3=value3]`, istio would compare the labels with endpoints.
```
howardjohn, line 1252: can you do `[key1=value1, key2]`?

hzxuzhonghu: Intended not to couple this ussage here. Although we do support it, i think it is not common, and should not encourage such usage

ramaraochavali (approved):

PR #2732, opened 2023-03-16
Add to gateway documentation to clarify use of non-standard ports.

jhuttley-F5-com:
This attempt is generated from the gateway.proto  files.
did make gen

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -35,6 +35,12 @@ import "google/api/field_behavior.proto";
 // on these ports, it is the responsibility of the user to ensure that
 // external traffic to these ports are allowed into the mesh.
 //
+// Ports 9443 and 2379 must exist on the ingress-gateway, `my-gateway-controller` in the example below.
+// By default only ports 80 and 443 (http) are configured.
+// The available ports are shown in the Service for the ingress-gateway, in the istio-system namespace.
+// If a port requested by a `Gateway` resource does not exist on the ingress-gateway, 
+// `istioctl analyze` will report "The gateway refers to a port that is not exposed on the workload".
+//
 // {{<tabset category-name="example">}}
 // {{<tab name="v1alpha3" category-value="v1alpha3">}}
 // ```yaml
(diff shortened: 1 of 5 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -35,6 +35,12 @@ import "google/api/field_behavior.proto";
 // on these ports, it is the responsibility of the user to ensure that
 // external traffic to these ports are allowed into the mesh.
 //
+// Ports 9443 and 2379 must exist on the ingress-gateway, `my-gateway-controller` in the example below.
```
linsun, line 38: Why not mention 9080?

linsun: which I saw also in the gateway example below.

jhuttley-F5-com: Hi, yes there could be more, but at the moment I'm trying to figure out how the process works. it says I need to apply a release-notes-none   label  and I haven't figure out how to do that!!

jhuttley-F5-com: Hi @linsun  I cant figure out how to edit this PR to add the release-notes-none   label. In addition I set it to "draft" and now see no way to remove that label in the webpage.
https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels#applying-a-label
Does not seem to help as in my own fork, I don't see my PR.

Using gh I dont have access.

? Which repository should be the default? istio/api
✓ Set istio/api as the default repository for the current directory
 gh pr list
 gh pr edit
? What would you like to edit? Labels
? Labels do-not-merge/work-in-progress, ok-to-test, release-notes-none, size/M
? Submit? Yes
Message: jhuttley-F5-com does not have the correct permissions to execute `AddLabelsToLabelable`, Locations: [{Line:1 Column:54}]
--
Do I have to cancel the PR and start again?  --John

ericvn: > Hi, yes there could be more, but at the moment I'm trying to figure out how the process works. it says I need to apply a release-notes-none label and I haven't figure out how to do that!!

I believe only istio members, and maybe its maintainers, can apply the label. A secondary comment is if you look closely, the `release-notes_api` test is not required in this repo. As such, a failure in that test would not prevent a PR from being merged.

ericvn: Thanks for the contribution. Since this change is clarifying documentation and that usually doesn't have a release note, I'll add the `release-noes-none` label.

PR #2734, opened 2023-03-20
Support Retry Budget

quraynai:
Open new MR to solve conflict. Previous MR https://github.com/istio/api/pull/2585

Issue: https://github.com/istio/istio/issues/27419

Istio: https://github.com/istio/istio/pull/42232


rafatbiin:
hi @linsun & @quraynai, this is a bit unrelated, but how can you configure it for your service where the retry budget depends on something like active requests and active pending requests which is very low level and not easy to measure compared to the number amount of requests we're sending to upstream from a client. The same goes for min retry concurrency which we've less control over than retry per second(used by linked: https://linkerd.io/2.12/tasks/configuring-retries/#retry-budgets). I've opened a discussion on the [envoy repository](https://github.com/envoyproxy/envoy/issues/25660) but got no response from the maintainer unfortunately to discuss how we can measure and configure it for a production service.

quraynai:
@rafatbiin I think number of request is harder to configure and measure compare to number of active connections.
Retries base on number of request we might need to know RPS and interval that we want to measure and it's hard to make retry budget work if we configure incorrectly.

quraynai:
@howardjohn @linsun sorry for no activities with this MR for a while. 
We made our forked version and POC for a while. Could you please help review this MR again?

quraynai:
@howardjohn @linsun hi could you please help review on this PR?

rafatbiin:
> @rafatbiin I think number of request is harder to configure and measure compare to number of active connections. Retries base on number of request we might need to know RPS and interval that we want to measure and it's hard to make retry budget work if we configure incorrectly.

@quraynai sorry for the late reply. when you said `number of connections`, you meant `active requests`, right? how can we measure this in a reliable way to configure the retry budget percentage? because, from my experience, it's very low level and hard to measure. it can literally changes from 100 to 0 in a second and depending on when envoy is taking the measurement, you'll get a totally different retry budget percentage. it's specially true with service which doesn't have a lot of outgoing requests. because if you consider the average case, the retry budget percentage will be very high and when we'll have real retry storm, it'll actually bite us. do you have any suggestion on how to go about measuring this metrics?

hzxuzhonghu:
This is a very advanced setting provided by envoy, we haven't had very critical requirements for it

quraynai:
@hzxuzhonghu I understand about it's not critical to have this functionality at the moment from istio.
but it's kinda problem to configure it from EnvoyFilter. You can check more detail from this link
https://github.com/istio/istio/issues/27419

quraynai:
@rafatbiin maybe we can try with static number and readjust as global.
Understand tuning this value dynamically per service is really hard and might not worth it.

quraynai:
@hzxuzhonghu @linsun @howardjohn hi could you please help on this?

howardjohn:
I am concerned about this as retries are in VirtualService, so putting retry budget in DR is odd. It is to align with Envoy semantics only, not really designed from users perspective. Gateway-API is also looking into adding this in HTTPRoute; if we put it in DR it will be incompatible cc @mikemorris.

That being said, I don't know a way to put it in route without envoy changes.

quraynai:
@howardjohn I understand your concern. we will look into another option

ericdbishop:
> I am concerned about this as retries are in VirtualService, so putting retry budget in DR is odd. It is to align with Envoy semantics only, not really designed from users perspective. Gateway-API is also looking into adding this in HTTPRoute; if we put it in DR it will be incompatible cc @mikemorris.
> 
> That being said, I don't know a way to put it in route without envoy changes.

@howardjohn We are interested in having access to the `retry_budget` circuit breaker threshold via the DR api, as `max_retries` is a static value that requires tuning for every service within the Mesh, making it difficult for us to update when services begin receiving more traffic. In comparison, `retry_budget` configurations require no major tuning since the retry value is relative, and it also cuts new requests earlier than using `max_retries`, performing better in our tests. But using `retry_budget` via EnvoyFilter's is extremely difficult to maintain across our services. Do you think there is any path forward to exposing this within DestinationRule we could work on? Thanks.

mikemorris:
@ericdbishop We're starting to look at implementing retries (including retry budgets) upstream in Gateway API in https://github.com/kubernetes-sigs/gateway-api/issues/1731

One thing I'd like to get more clarity on is where/how this should be implemented - my initial thought was as a field on HTTPRoute (similar to [Timeouts](https://gateway-api.sigs.k8s.io/geps/gep-1742/)), but I've seen a few mentions of DestinationRule or [client-side config](https://github.com/istio/istio/issues/27419#issuecomment-2093659300) which I'd like to better understand the implications for (or any Envoy implementations constraints) as it would relate to Ambient mesh without client sidecars and reconciling this functionality with other Gateway API implementations such as [Linkerd's retry budgets](https://linkerd.io/2.15/tasks/configuring-retries/#retry-budgets).

PR #2737, opened 2023-03-22
Clarify that MeshConfig.localityLbSetting is enabled by default

kfaseela:
As per : https://github.com/istio/istio/blob/master/pkg/config/mesh/mesh.go#L92 this is enabled by default.
Better to clarify this in docs, because k8s has this disabled by default, and caused us come confusions while we started using this. Ideally this should have been disabled by default.

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -303,6 +303,8 @@ message MeshConfig {
 
   // Locality based load balancing distribution or failover settings.
   // If unspecified, locality based load balancing will be enabled by default.
+  // However, this requires outlierDetection to actually take effect for a particular
+  // service, see https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/failover/
   istio.networking.v1alpha3.LocalityLoadBalancerSetting locality_lb_setting = 35;
 
   // Configures DNS refresh rate for Envoy clusters of type `STRICT_DNS`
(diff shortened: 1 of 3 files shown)


mesh/v1alpha1/config.proto
```diff
@@ -302,6 +302,7 @@ message MeshConfig {
   string root_namespace = 34;
 
   // Locality based load balancing distribution or failover settings.
+  // If unspecified, locality based load balancing will be enabled by default.
```
howardjohn, line 305: I would maybe say "enabled by default. However, this requires outlierDetection to actually take effect for a particular service, see https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/failover/" or similar

kfaseela: Done

howardjohn (approved):

mesh/v1alpha1/config.proto
```diff
@@ -302,6 +302,9 @@ message MeshConfig {
...
+  // However, this requires outlierDetection to actually take effect for a particular
+  // service, see https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/failover/
```
ericvn, line 307: Should add a period to the end of the sentence. Maybe:
```suggestion
  // However, for locality load balancing to occur for a particular service, outlierDetection must be configured. See (https://istio.io/latest/docs/tasks/traffic-management/locality-load-balancing/failover/).
```
Need to verify the period at the end of the URL ends up being on the sentence, nit the URL.

PR #2741, opened 2023-03-23
small correction in peer auth example

kfaseela:
Fixes [#12933](https://github.com/istio/istio.io/issues/12933)

### DIFF
--- a/security/v1beta1/peer_authentication.proto
+++ b/security/v1beta1/peer_authentication.proto
@@ -57,7 +57,7 @@ option go_package="istio.io/api/security/v1beta1";
 // apiVersion: security.istio.io/v1beta1
 // kind: PeerAuthentication
 // metadata:
-//   name: default
+//   name: finance-peer-auth
 //   namespace: foo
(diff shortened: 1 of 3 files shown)


kfaseela:
@ericvn if we do not use release notes from api repo, why run release note checks here?

security/v1beta1/peer_authentication.proto
```diff
@@ -57,7 +57,7 @@ option go_package="istio.io/api/security/v1beta1";
 // apiVersion: security.istio.io/v1beta1
 // kind: PeerAuthentication
 // metadata:
-//   name: default
+//   name: finance-peer-auth
```
howardjohn, line 60: nit: I would call it "finance", no need to add the type in the name

howardjohn (approved):

security/v1beta1/peer_authentication.proto
kfaseela: Done

ericvn (approved):

PR #2743, opened 2023-03-29
Add annotation istio.io/rev

jewertow:
For more details, please look at this PR: https://github.com/istio/istio/pull/44161.

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -441,6 +441,8 @@ annotations:
   - name: istio.io/rev
     featureStatus: Alpha
     description: Specifies a control plane revision to which a given proxy is connected.
+      This annotation is added automatically, not set by a user. In contrary to the label istio.io/rev,
+      it represents the actual revision, not the requested revision.
     deprecated: false
     hidden: false
     resources:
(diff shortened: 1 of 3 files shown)


ericvn (approved):
LGTM

Also needed allow the istio.io tests to pass since the annotation use was already merged.

ericvn:
@howardjohn Would you be able to do a quick review so this can merge and we can move up the istio reference in the istio.io tests. Thanks.

ericvn:
Is there an issue having a label and annotation that are the same? I do see we already have a label.

annotation/annotations.yaml
```diff
@@ -438,6 +438,14 @@ annotations:
     resources:
       - AuthorizationPolicy
 
+  - name: istio.io/rev
+    featureStatus: Alpha
+    description: Specifies a control plane revision to which a given proxy is connected.
```
howardjohn, line 443: It may help to note that this is automatically added, not set by the user

Label: users intent
Annotation: actual result

jewertow: Done

howardjohn (approved):

PR #2745, opened 2023-04-03
Add gateway.istio.io/controller-version annotation

ericvn:
Introduced in https://github.com/istio/istio/pull/44174 and causes a `istioctl analyze` `Unknown Annotation` errors:
```
"Warning [IST0108] (Deployment default/bookinfo-gateway-istio) Unknown annotation: gateway.istio.io/controller-version
Warning [IST0108] (Pod default/bookinfo-gateway-istio-7f4d8dd7f6-cvg9d) Unknown annotation: gateway.istio.io/controller-version
Warning [IST0108] (Service default/bookinfo-gateway-istio) Unknown annotation: gateway.istio.io/controller-version"
```

I believe this should fix the issue, but this is my first time working with annotations. I believe we would keep this hidden and it seems most annotations are Alpha.

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -132,6 +132,18 @@ var (
 		},
 	}
 
+	GatewayControllerVersion = Instance {
+		Name:          "gateway.istio.io/controller-version",
+		Description:   "A version added to the Gateway by the controller "+
+                        "specifying the `controller version`.",
+		FeatureStatus: Alpha,
+		Hidden:        true,
+		Deprecated:    false,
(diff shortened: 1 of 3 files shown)


howardjohn (approved):
lgtm, will probably need backport as well

ericvn:
I did plan to cherry-pick once things worked on the main thread. I expect there will be a small change in the istio/istio PRs to directly reference the annotation as well.

howardjohn:
technically istio/istio doesn't need to reference it to fix the analyze
message.

PR #2748, opened 2023-04-05
docs: fix wasm plugin traffic selector docs

psinghal20:
The docs incorrectly mentioned CLIENT as INBOUND and SERVER as OUTBOUND in the traffic selector field docs in WASM Plugin. This PR changes the order of OUTBOUND and INBOUND terms to fix this confusion. 

The `WorkloadMode` docs correctly mention the CLIENT as OUTBOUND and SERVER as INBOUND: https://istio.io/latest/docs/reference/config/type/workload-selector/#WorkloadMode


howardjohn (approved):

psinghal20:
Hi team, this change would not require any release notes update. Can someone from the org please add `release-notes-none`? I don't have the necessary permissions to do so.

zirain (approved):

PR #2753, opened 2023-04-07
add regex rewrite support for uris

ramaraochavali:
Add regex rewrite support for uris. This will also help in natively support gateway API's FullPathRewrite (instead of current hack we have)

Fixes https://github.com/istio/istio/issues/22290


networking/v1alpha3/virtual_service.proto
```diff
@@ -1684,6 +1684,18 @@ message HTTPRewrite {
 
   // rewrite the Authority/Host header with this value.
   string authority = 2;
+
+  // rewrite the path portion of the URI with the specified regex.
+  RegexMatch uri_regex_rewrite = 3;
```
howardjohn, line 1689: Gateway API doesn't actuallyneed regex rewrite, we only need prefix or full. I also don't think we should be changing our public v1 api for gateway-api internal impl details; the API should stand on its own as useful

howardjohn: Actually i see you said it will _also_ help gateway, so I assume you have use cases beyond that?

ramaraochavali: Yes. We are seeing Regex rewrite use cases  especially people migrating from other proxies like nginx/haproxy to Istio. There it is natively supported.

ramaraochavali: Also https://github.com/istio/istio/issues/22290

ramaraochavali: @howardjohn Can you please take a look when you get chance? This is similar to what you have thought about [here](https://github.com/istio/istio/issues/22290#issuecomment-665266245)

mlitvinav: Hi 👋
@howardjohn the reason why Regex rewrites would be nice to have is because Istio currently only rewrites what it matches. If something is dynamic (mainly path parameters), they cannot be matched statically and a Regex would match everything from ^ to $, removing the dynamic part in the process. Regex groups would be handy since it would allow to grab a part of the path, e.g. only a uuid for a user or a sku for a product and placed into the new rewritten url path.

Often when websites are refactored the URL paths have to stay the same for SEO reasons and not everything is 1 to 1 translatable to the new site.

But i am not sure what the performance implications are for this Istio feature. Maybe allowing to rewrite a prefix and a postfix could be a solution but i cannot speak for all the use cases this should handle.

ramaraochavali:
@howardjohn PTAL when you get chance

networking/v1alpha3/virtual_service.proto
```diff
@@ -1684,6 +1684,18 @@ message HTTPRewrite {
...
+}
+
+message RegexMatch {
```
hzxuzhonghu, line 1692: RegexRewrite?

networking/v1alpha3/virtual_service.proto
```diff
@@ -1684,6 +1684,18 @@ message HTTPRewrite {
...
+  // RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
+  string match = 1;
+
+  // The string that should replace into matching portions of original URI. 
+  // Capture groups in the pattern can be referenced in the new URI.
+  string uri = 2;
```
hzxuzhonghu, line 1698: Looks envoy field `substitution` is much obvious

networking/v1alpha3/virtual_service.proto
ramaraochavali: Sure. I can change

networking/v1alpha3/virtual_service.proto
ramaraochavali: how about rewrite?

hzxuzhonghu: looks better than now

costinm:
Since you mentioned Gateway API - how would a user who wants to adopt the
new standard API handle this ? Assuming they have HttpRoute only.

We have discussed ways to 'attach' a VirtualService to a K8S Gateways, a
bit hacky but may work. However for new APIs we should consider this
as first class. And if we define it as an extension ( vendor filter, etc) -
maybe we can leave Virtualservice as is and only define the new thing for
the new API ?

ramaraochavali:
@costinm I agree with the need for being compatible with gateway API. I think we need to work on a strategy to figure that out in parallel.

hzxuzhonghu (approved):

hzxuzhonghu:
I think this is consistent with gateway api filter

ramaraochavali:
@howardjohn PTAL when you get chance

ramaraochavali:
https://gateway.envoyproxy.io/v0.3.0/user/http-urlrewrite.html - envoy gateway supports some thing similar

howardjohn:
> https://gateway.envoyproxy.io/v0.3.0/user/http-urlrewrite.html - envoy gateway supports some thing similar

Isn't it only prefix or full?

ramaraochavali:
I think it has both https://gateway.envoyproxy.io/v0.3.0/user/http-urlrewrite.html#rewrite-url-full-path

howardjohn:
ReplaceFullPath is not a regex. It just means `req.path = replaceFullPath` basically. Its even simpler than prefix.

Gateway API has no regex: https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPPathModifier.

networking/v1beta1/virtual_service.proto
```diff
@@ -1691,6 +1691,18 @@ message HTTPRewrite {
 
   // rewrite the Authority/Host header with this value.
   string authority = 2;
+
+  // rewrite the path portion of the URI with the specified regex.
+  RegexRewrite uri_regex_rewrite = 3;
```
howardjohn, line 1696: I would personally prefex to avoid regex wherever possible. But if we are going to have this can we have an example at least? its not obvious how to use this

ramaraochavali: I think there are valid use cases for regex rewrite especially for folks migrating from HAProxy to Envoy (Istio) as I mentioned. So if we are willing to add this, I can add examples for sure.

ramaraochavali:
@howardjohn added examples. PTAL.

howardjohn (approved):

rachitchauhan43:
@ramaraochavali / @howardjohn  : Which Istio version this feature will be available in ?

zirain:
@rachitchauhan43 usually istio don't cherry-pick new feature to a released version.

zirain:
@ramaraochavali is this landed in 1.18?

ramaraochavali:
The implementation was merged on Jun11th. So I guess 1.19 only

rachitchauhan43:
@ramaraochavali : Any chance that this can be back-ported into 1.17 as well ? We are using KServe and KServe is testing it's compatibility with Istio-1.17. This feature is a much needed feature by us.

ramaraochavali:
We do not backport features specially with api changes. So I think it would be difficult to backport. But I would leave the decision to @istio/technical-oversight-committee

PR #2758, opened 2023-04-12
Add cipher suites support in meshConfig for mesh-wide ciphers

jacob-delgado:
Partial fix for https://github.com/istio/istio/issues/28996

Allow users to specify the ciphers they want sidecars to use within their sidecars.


mesh/v1alpha1/config.proto
```diff
@@ -1133,6 +1133,16 @@ message MeshConfig {
     // If not specified, the default curves enforced by envoy will be used. For details about the default curves, refer to
     // [Ecdh Curves](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto)
     repeated string ecdh_curves = 2;
+    //
```
kfaseela, line 1136: we have two ways in which TLSProtocol is used. mesh_mTLS and tls_defaults. Are you planning to use this for both the cases? If not the corresponding docs should be updated as in line 1160

jacob-delgado:
We could also move this to `ProxyConfig` CRD. cc @costinm @howardjohn any thoughts?

mesh/v1alpha1/config.proto
howardjohn: useful for tls_defaults IMO, if you are using TLS origination to a server that only has legacy ciphers

howardjohn (approved):

kfaseela (approved):

kfaseela:
i think we can wait to merge until implementation is ready?

kfaseela:
@jacob-delgado did we implement this? otherwise we may have to revert the API PR

peterj:
Btw, I tried this with the beta bits and I can use the `cipherSuites` field in mesh config to use only a specific suite; if trying to force to use another one, the call fails (which is expected).

PR #2761, opened 2023-04-17
Add authz dry-run example

wulianglongrd:
dry-run is a very useful feature, but there is no mention in the authz reference document, so add an example.

### DIFF
--- a/security/v1/authorization_policy.proto
+++ b/security/v1/authorization_policy.proto
@@ -217,6 +217,30 @@ import "type/v1beta1/selector.proto";
 //    matchLabels:
 //      version: v1
 // ```
+//
+// The following example set up an authorization policy using annotation `istio.io/dry-run` to dry-run the policy without actually enforcing it.
+//
+// The dry-run annotation allows you to better understand the effect of an authorization policy before applying it to the production traffic.
+// This helps to reduce the risk of breaking the production traffic caused by an incorrect authorization policy.
+// For more information, see [dry-run tasks](https://istio.io/latest/docs/tasks/security/authorization/authz-dry-run/).
+//
+// ```yaml
+// apiVersion: security.istio.io/v1
+// kind: AuthorizationPolicy
+// metadata:
+//   name: deny-path-headers
+//   annotations:
+//     "istio.io/dry-run": "true"
+// spec:
+//   selector:
+//     matchLabels:
+//       app: httpbin
+//   action: DENY
+//   rules:
+//   - to:
+//     - operation:
+//         paths: ["/headers"]
+// ```
 package istio.security.v1;
 
 option go_package="istio.io/api/security/v1";
--- a/security/v1/authorization_policy.proto
+++ b/security/v1/authorization_policy.proto
@@ -218,7 +218,8 @@ import "type/v1beta1/selector.proto";
 //      version: v1
 // ```
 //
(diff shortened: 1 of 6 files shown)


wulianglongrd:
@istio/technical-oversight-committee Please take a look, if this is deemed useless, I'll be happy to close it. Thanks.

security/v1/authorization_policy.proto
```diff
@@ -217,6 +217,30 @@ import "type/v1beta1/selector.proto";
 //    matchLabels:
 //      version: v1
 // ```
+//
+// The following example set up an authorization policy using annotation `istio.io/dry-run` to dry-run the policy without actually enforcing it.
```
hanxiaop, line 221: Not sure if this is the right place to add the dry-run annotation, since it's not related to the fields of authorizationpolicy. Also, in the annotation page, we have mentioned the usage of this annotation https://istio.io/latest/docs/reference/config/annotations/.

security/v1/authorization_policy.proto
```diff
@@ -217,6 +217,30 @@ import "type/v1beta1/selector.proto";
...
+//
+// The dry-run annotation allows you to better understand the effect of an authorization policy before applying it to the production traffic.
```
hanxiaop, line 223: Also if this can be added, since it's explicitly marked as experimental, at least we need to add a note to indicate this is an experimental feature like the notes in https://istio.io/latest/docs/tasks/security/authorization/authz-dry-run/?

security/v1/authorization_policy.proto
wulianglongrd: Thanks for the reply, I think this is a very common feature, and the most commonly used should be the configuration reference document, so I think it is necessary to add an example here.

Also, this is an AuthZ-only feature, and using annotations to implement it is a bit unexpected (it might be applicable to a wider range of behavior control? Not sure), I think few people will think to confirm from the annotation description page whether support for a specific field of feature.

security/v1/authorization_policy.proto
wulianglongrd: thanks, i will fix it

security/v1/authorization_policy.proto
```diff
@@ -217,6 +217,31 @@ import "type/v1beta1/selector.proto";
...
 //      version: v1
 // ```
+//
+// The following example shows you how to set up an authorization policy using an [experimental annotation](https://istio.io/latest/docs/reference/config/annotations/)
+// `istio.io/dry-run` to dry-run the policy without actually enforcing it.
+//
+// The dry-run annotation allows you to better understand the effect of an authorization policy before applying it to the production traffic.
+// This helps to reduce the risk of breaking the production traffic caused by an incorrect authorization policy.
+// For more information, see [dry-run tasks](https://istio.io/latest/docs/tasks/security/authorization/authz-dry-run/).
+//
```
linsun, line 227: do we need the sample yaml here, given you already linked to istio.io authz dry run page?  would prefer no need for the yaml here in case there is update to this in the future.

linsun (commented):
overall lgtm, just some minor feedback

security/v1/authorization_policy.proto
wulianglongrd: Glad for your reply, I think the sample yaml should be added since other examples do this too.

howardjohn (approved):

PR #2799, opened 2023-06-01
wasm: add fail_open for telemetry plugins

kyessenov:
Exposes the [fail_open](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto#extensions-wasm-v3-pluginconfig) flag from Envoy, removing the need to use EnvoyFilter for Wasm filters.


extensions/v1alpha1/wasm.proto
```diff
@@ -311,6 +311,12 @@ message WasmPlugin {
   // namespace of the `WasmPlugins`. Defaults to `0`.
   google.protobuf.Int32Value priority = 10;
 
+  // Enables the fail open behavior for the Wasm plugin fatal errors to bypass
+  // the plugin execution. A fatal error can be a failure to fetch the remote
+  // binary, an exception, or abort() on the VM. This flag is not recommended
+  // for the authentication or the authorization plugins.
+  bool fail_open = 13;
```
howardjohn, line 318: is there any possibility we will want other values in the future that may point to an enum? Like

`STRICT|ALLOW_FETCH_FAILURE|ALLOW_ANY_FAILURE` or something possibly, if we distinguish from exception vs fetching failing?

howardjohn (commented):
Overall seems reasonable, just one minor syntax question

extensions/v1alpha1/wasm.proto
kyessenov: I don't think we need fine-grained fail-open modes in the medium term. For security filters, nothing should fail. For telemetry filters, there's no need to treat some failures as special.

kyessenov: Makes sense, changed to enum.

hzxuzhonghu (approved):

howardjohn (approved):

PR #2800, opened 2023-06-02
stats: allow custom histogram buckets

kyessenov:
Compromise to https://github.com/istio/api/pull/2539, to prefer over the bootstrap EnvoyFilter.

@zirain

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -201,8 +201,7 @@ annotations:
 
   - name: sidecar.istio.io/statsHistogramBuckets
     featureStatus: Alpha
-    description: |
-      Specifies the semicolon separated custom histogram buckets with a prefix matcher. Example: `istio:1,5,10,50,100,500,1000,5000,10000;envoy:1,5,10,25,50,100,250,500,1000,2500,5000,10000`.
+    description: Specifies the custom histogram buckets with a prefix matcher to separate the Istio mesh metrics from the Envoy stats, e.g. `{"istio":{1,5,10,50,100,500,1000,5000,10000},"envoy":{1,5,10,25,50,100,250,500,1000,2500,5000,10000}}`. Default buckets are `{0.5,1,5,10,25,50,100,250,500,1000,2500,5000,10000,30000,60000,300000,600000,1800000,3600000}`.
     deprecated: false
     hidden: false
     resources:
@@ -477,7 +476,7 @@ annotations:
   - name: sidecar.istio.io/extraStatTags
     featureStatus: Alpha
     description: An additional list of tags to extract from the in-proxy Istio Wasm telemetry. Each additional tag needs to be present in this list.
-    deprecated: true
+    deprecated: false
     hidden: false
     resources:
       - Pod
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -199,6 +199,15 @@ annotations:
     resources:
       - Pod
 
+  - name: sidecar.istio.io/statsHistogramBuckets
+    featureStatus: Alpha
+    description: |
+      Specifies the semicolon separated custom histogram buckets with a prefix matcher. Example: `istio:1,5,10,50,100,500,1000,5000,10000;envoy:1,5,10,25,50,100,250,500,1000,2500,5000,10000`.
+    deprecated: false
+    hidden: false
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -467,8 +476,8 @@ annotations:
 
   - name: sidecar.istio.io/extraStatTags
     featureStatus: Alpha
-    description: An additional list of tags to extract from the in-proxy Istio telemetry. each additional tag needs to be present in this list.
-    deprecated: false
+    description: An additional list of tags to extract from the in-proxy Istio Wasm telemetry. Each additional tag needs to be present in this list.
```
zirain, line 479: prefer to change this in a separate pr

zirain (approved):
this's a feature only supported by envoy, using annotation is LGTM.

annotation/annotations.yaml
```diff
@@ -199,6 +199,15 @@ annotations:
     resources:
       - Pod
 
+  - name: sidecar.istio.io/statsHistogramBuckets
+    featureStatus: Alpha
+    description: |
+      Specifies the semicolon separated custom histogram buckets with a prefix matcher. Example: `istio:1,5,10,50,100,500,1000,5000,10000;envoy:1,5,10,25,50,100,250,500,1000,2500,5000,10000`.
```
costinm, line 205: - document the default 
- why do we need istio and envoy ? Can one be used for both ?

costinm (approved):
I think it's reasonable - we have quite a few things exposed as annotations.

One concern about pod annotations is that in Ambient they will likely cease functioning - except for Waypoints, where the annotation will go on the Gateway resource.  But that's the case for most CRDs we have with workload selector, so nothing special.

On this note - would be good to add K8S Gateway as resource that will support the annotation ( they get copied to the deployment AFAIK ).

costinm:
Worth mentioning that customizing stats bucket for specific pods is actually best granularity for such an API - the defaults should work for most common cases, overrides are needed for pods that have unusual behavior, like long lived connections ( istiod itself ).

annotation/annotations.yaml
howardjohn: If we do need them separate, should we just make this json? Its not that complicated but we are essentially creating a micro-language we need to parse.

if we just make them one then its a simple CSV so no problem

howardjohn (commented):
Looks good overall as an experimental annotation, just one comment on the formatting (doesn't need to block, just want to make sure it is at least ACKed before moving forward)

annotation/annotations.yaml
kyessenov: Updated. Prefix is not needed but usually used.

annotation/annotations.yaml
```diff
@@ -199,6 +199,14 @@ annotations:
     resources:
       - Pod
 
+  - name: sidecar.istio.io/statsHistogramBuckets
+    featureStatus: Alpha
+    description: Specifies the custom histogram buckets with a prefix matcher to separate the Istio mesh metrics from the Envoy stats, e.g. `{"istio":{1,5,10,50,100,500,1000,5000,10000},"envoy":{1,5,10,25,50,100,250,500,1000,2500,5000,10000}}`. Default buckets are `{0.5,1,5,10,25,50,100,250,500,1000,2500,5000,10000,30000,60000,300000,600000,1800000,3600000}`.
```
howardjohn, line 204:
```suggestion
    description: Specifies the custom histogram buckets with a prefix matcher to separate the Istio mesh metrics from the Envoy stats, e.g. `{"istio":[1,5,10,50,100,500,1000,5000,10000],"envoy":[1,5,10,25,50,100,250,500,1000,2500,5000,10000]}`. Default buckets are `[0.5,1,5,10,25,50,100,250,500,1000,2500,5000,10000,30000,60000,300000,600000,1800000,3600000]`.
```

Current one is not valid json

howardjohn: can we specify what the unit is as well? milliseconds?

howardjohn (approved):
One small comment, otherwise LGTM. feel free to remove hold

annotation/annotations.yaml
kyessenov: There's no unit. We have both byte count and milliseconds in use.

PR #2811, opened 2023-06-09
fixes for Duplicate reference docs overwriting each other 

AndreaM12345:
fixes https://github.com/istio/istio.io/issues/12693


howardjohn:
is there any way we can test this?

howardjohn:
I mean make sure it doesn't happen again

AndreaM12345:
i was adding logic to the sync.sh script to check if the mode is present or not and then take corrective action

ericvn:
I suspect adding checks in both repos is appropriate. The reason for adding the check in istio.io is we do pull reference docs from multiple repos IIRC. There is a possibility someone could add something new that overlapped with another.

ericvn:
> Is this fix still needed? No objection to it - just wasn't sure how to validate it

I did add a check in the istio.io reference file update logic to verify we don't copy a file over another. Today, it's a warning, and after this merges, I will make it fail. That should cover duplicates with a repo, like these, or even duplicates across repos.

ericvn (approved):
I did review the tablets across the files and they all look OK.
There are 4 removed html files that coincide with the duplicate sets.

Thanks.

AndreaM12345:
@linsun @howardjohn Thanks for the earlier comments.  Let me know if there are any other areas that need to be address or if this PR is sufficient for fixing https://github.com/istio/istio.io/issues/12693.

howardjohn (approved):

PR #2812, opened 2023-06-09
Fix doc link for istio.operator.v1alpha1 helm options pointing at v1.5 docs

dhawton:
Fixes the link for `Helm installation options` link that appears https://istio.io/latest/docs/reference/config/istio.operator.v1alpha1/ linking to v1.5 instead of latest docs.

### DIFF
--- a/operator/v1alpha1/operator.proto
+++ b/operator/v1alpha1/operator.proto
@@ -112,7 +112,7 @@ message IstioOperatorSpec {
     map<string, ExternalComponentSpec> addonComponents = 51 [deprecated = true];
 
     // Overrides for default `values.yaml`. This is a validated pass-through to Helm templates.
-    // See the [Helm installation options](https://istio.io/v1.5/docs/reference/config/installation-options/) for schema details.
+    // See the [Helm installation options](https://istio.io/docs/reference/config/installation-options/) for schema details.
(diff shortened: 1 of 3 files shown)


operator/v1alpha1/operator.proto
```diff
@@ -112,7 +112,7 @@ message IstioOperatorSpec {
     map<string, ExternalComponentSpec> addonComponents = 51 [deprecated = true];
 
     // Overrides for default `values.yaml`. This is a validated pass-through to Helm templates.
-    // See the [Helm installation options](https://istio.io/v1.5/docs/reference/config/installation-options/) for schema details.
+    // See the [Helm installation options](https://istio.io/docs/reference/config/installation-options/) for schema details.
```
ericvn, line 115: The reason this points to 1.5 is that is the latest version of the docs with this reference page. The new link yields a page not found.

Maybe we could move it to https://istio.io/latest/docs/setup/install/helm/ or https://istio.io/latest/docs/setup/additional-setup/customize-installation-helm/.

ericvn: Actually, https://github.com/istio/istio.io/pull/7106 removed the page (helm was being deprecated and we were moving to operator (and the new rendered source of truth is:
https://preliminary.istio.io/docs/reference/config/istio.operator.v1alpha1/).

PR #2817, opened 2023-06-14
support HTTP/1.1 case preserve

kyessenov:
Enables stateful HTTP header case formatter per https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/header_casing on a per-proxy basis.

This has to be opt-in for backwards compatibility.

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -506,8 +506,8 @@ var (
 
 	SidecarPreserveHeaderCase = Instance {
 		Name:          "sidecar.istio.io/preserveHeaderCase",
-		Description:   "If set to `true`, Envoy preserve the header case for "+
-                        "proxied HTTP/1.1 requests.",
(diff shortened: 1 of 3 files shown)


howardjohn (changes requested):
I don't think we should do this, see https://github.com/istio/istio/issues/32008#issuecomment-988865470.

I believe this was a TOC decision, not just my personal opinion, but it was a while ago so I may be mistaken

kyessenov:
Can you explain why?
The goal to use http/2 everywhere does not seem reasonable, certainly not at ingress gateways.

howardjohn:
here was the discussion https://docs.google.com/document/d/13lxJqtlaQhmV2EwsNnS6h-_O4pobZQZuMjrzOeMgVI0/edit#bookmark=id.q87xoxdhvz36

PR #2819, opened 2023-06-15
Wasm fail open

kyessenov:
Fixes https://github.com/istio/api/issues/2801


hanxiaop (approved):

howardjohn (changes requested):
I don't think we should be modifying APIs in patch releases. We have actually made soft commitments to this and I know at least 2 vendors that will be impacted by this (not sure how much, but enough we should at least give them a heads up).

Is there a reason we need this in 1.18 or can it wait 2 months?

kyessenov:
It's a low risk control plane change. Defaults are the same, and the runtime has not changed in a long time. I think it's acceptable for an alpha API.

howardjohn:
I agree its low risk but our policy is to not add features -- especially
APIs -- in patch releases. Especially in this case there seems to be no
urgent need for this

linsun:
Agreed we don't usually add API changes to patch releases. Do you have a particular urgent need for this @kyessenov ?

kyessenov:
Fair enough. One of our customers expressed a desire to have this but I'll let them speak here if they really want it.

hzxuzhonghu:
I think users can cherry-pick some feature from master branch

craigbox (commented):
This is in master already, and the decision was broadly not to backport it to 1.18.  Is this still relevant @kyessenov or would you like to close?

hzxuzhonghu:
maybe not needed as 1.18 will be EOL in December, very close

linsun:
Seems the general consesus is to not have this API change in a patch release and the PR has been out for a few months without any movement. Going to close it - please reopen if I'm mistaken.

PR #2820, opened 2023-06-15
add a new TLS mode for validating client cert if presented

ramaraochavali:
As discussed and agreed up on https://github.com/istio/istio/issues/36863
https://github.com/istio/istio/issues/31984


networking/v1alpha3/gateway.proto
```diff
@@ -650,6 +652,15 @@ message ServerTLSSettings {
     // for mTLS authentication. When this mode is used, all other
     // fields in `TLSOptions` should be empty.
     ISTIO_MUTUAL = 4;
+
+    // Similar to MUTUAL mode, except that the client certificate
+    // is optional. A client certificate will also be requested 
+    // during handshake, but does not require that the client sends a
+	  // certificate.  When a client certificate is presented, it will
```
hobbytp, line 659: Seems there is a tab here.

hzxuzhonghu:
Has envoy supported such option?

networking/v1alpha3/gateway.proto
```diff
@@ -650,6 +652,15 @@ message ServerTLSSettings {
...
+    // be validated. ca_certificates should be specified for validating
+    // client certificates. Otherwise, client certificate validation will 
+    // be skipped.
+    OPTIONAL_MUTUAL = 5;
```
hzxuzhonghu, line 663: Why donot we reuse simple mode

ramaraochavali:
> Has envoy supported such option?

Yes. If we add validation_context and set require_client_certificate to false - envoy uses this mode

networking/v1alpha3/gateway.proto
ramaraochavali: yeah. thanks you . Will fix

networking/v1alpha3/gateway.proto
ramaraochavali: https://github.com/istio/istio/issues/36863#issuecomment-1015156351 - Please read this comment and below.  For simple we do not want envoy to request for a client certificate. In this mode, we want Envoy to request for client certificate but not mandate it.  And validate only if it is provided. Equivalent to `VerifyClientCertIfGiven` go TLS option

hzxuzhonghu: I mean we can use simple mode:

when the client provides a cert, we will verify it. When not we will not verify, just let it go.

ramaraochavali: In simple mode, we do not request for client certificate today. If we have to request for client certificate during simple TLS mode (whether client provides or not is a different issue), there is a change in behaviour during handshake. That is why we need to additional mode - that is why most language support it explicitly (not combine with SIMPLE)

```
	// NoClientCert indicates that no client certificate should be requested
	// during the handshake, and if any certificates are sent they will not
	// be verified.
	NoClientCert ClientAuthType = iota
	// RequestClientCert indicates that a client certificate should be requested
	// during the handshake, but does not require that the client send any
	// certificates.
	RequestClientCert
	// RequireAnyClientCert indicates that a client certificate should be requested
	// during the handshake, and that at least one certificate is required to be
	// sent by the client, but that certificate is not required to be valid.
	RequireAnyClientCert
	// VerifyClientCertIfGiven indicates that a client certificate should be requested
	// during the handshake, but does not require that the client sends a
	// certificate. If the client does send a certificate it is required to be
	// valid.
	VerifyClientCertIfGiven
	// RequireAndVerifyClientCert indicates that a client certificate should be requested
	// during the handshake, and that at least one valid certificate is required
	// to be sent by the client.
	RequireAndVerifyClientCert
```

hzxuzhonghu: Does  `require_client_cert = false` equal to VerifyClientCertIfGiven, when we also set tls context.

As the original issue https://github.com/istio/istio/issues/36863#issue-1106453390 seems a regression

hzxuzhonghu: After a quick look at the history, understand the background why we make simple mode not verify client cert. 

So lgtm adding another option

howardjohn:
I'm OOO back Tuesday but would like to review this before merging if
possible

/hold

ramaraochavali:
> I'm OOO back Tuesday but would like to review this before merging if
possible

Sure. Absolutely. There is no hurry and would like more eyes on it as it is an important API change. I have put DNM

costinm:
I have been working on using a waypoint/gateway in front of Istiod - this
feature is one of the biggest gaps, since istiod requests client certs but
can fall back to JWT.

So big +1 on having this available. But please check the Gateway proposals,
it would be very good to have this in the common/standard API and not
istio-only.

ramaraochavali:
Gateway already supports a generic [Options](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayTLSConfig) that people use to annotate specific mode like MUTUAL for example

networking/v1alpha3/gateway.proto
linsun: @ramaraochavali  Could we explain why SIMPLE mode & require_client_cert = false won't work in the comment of the proto so users don't need to read all these histories?

linsun:
For folks who are familar with k8s gateway API, is this new mode supported (or being discussed) in k8s gateway API?

ramaraochavali:
> For folks who are familiar with k8s gateway API, is this new mode supported (or being discussed) in k8s gateway API?

Gateways AFAIK do not need to explicitly support it. Gateway implementations should support it via Options similar to how MUTUAL is supported.

networking/v1alpha3/gateway.proto
ramaraochavali: Added

networking/v1beta1/gateway.proto
```diff
@@ -650,6 +653,14 @@ message ServerTLSSettings {
     // for mTLS authentication. When this mode is used, all other
     // fields in `TLSOptions` should be empty.
     ISTIO_MUTUAL = 4;
+
+    // Similar to MUTUAL mode, except that the client certificate
+    // is optional. Unlike SIMPL mode, A client certificate will
```
linsun, line 658:
```suggestion
    // is optional. Unlike SIMPLE mode, A client certificate will 
```

networking/v1alpha3/gateway.proto
```diff
@@ -650,6 +653,14 @@ message ServerTLSSettings {
     // for mTLS authentication. When this mode is used, all other
     // fields in `TLSOptions` should be empty.
     ISTIO_MUTUAL = 4;
+
+    // Similar to MUTUAL mode, except that the client certificate
+    // is optional. Unlike SIMPLE mode, A client certificate will 
+    // be requested during handshake, but does not require that 
+    // the client sends a certificate.  When a client certificate 
+    // is presented, it will be validated. ca_certificates should
+    // be specified for validating client certificates.
```
bleggett, line 662:
```suggestion
    // Similar to MUTUAL mode, except that the client certificate
    // is optional. Unlike SIMPLE mode, A client certificate will 
    // still be explicitly requested during handshake, but the client 
    // is not required to send a certificate. If a client certificate 
    // is presented, it will be validated. ca_certificates should
    // be specified for validating client certificates.
```

bleggett (approved):
Given context makes sense to me, minor wording nits on explanatory comment.

ramaraochavali:
@costinm @howardjohn Can you PTAL when you get chance?

costinm:
In vacation - looks good to me.

I was wondering - should we have a mode or variant where the cert is not
verified in the gateway, just passed through ? Can be useful for chained
cases ( front gateway just handshakes and does global LB, layer 2 or
waypoint applies policies). Maybe if no CA is specified for client
validation ? The identity from cert wouldn't be used.

ramaraochavali:
>.  Maybe if no CA is specified for client
validation ? The identity from cert wouldn't be used.

So yo want to request for client cert, but not validate it? It may need some changes in Envoy

hzxuzhonghu (approved):
I think we need to update website as well, since we invented a new mode

ramaraochavali:
> I think we need to update website as well, since we invented a new mode

Will do

howardjohn (approved):

PR #2830, opened 2023-06-22
added identity path prefix

nmnellis:


howardjohn:
In my opinion, and in discussion in previous WGs, the identity schema has essentially ossified. I don't see a way that we can reasonable change this and expect things to continue working -- not just Istio core, but integrations as well.

There is also no explanation of _why_ a user should do this (both in the code and/or PR) and why we need it. why do `spiffe://my_td/my_prefix` instead of `my_td_my_prefix`, for example?

nmnellis:
Companies that use Spire or Spiffe identites outside of kubernetes do not often follow the kubernetes naming scheme Istio does. This goes for Istio based VMs as well. This path prefix does not impact the current ossified structure Istio uses and only extends its capabilities to fit more use cases  such as bringing istio workloads into a global identity system.

howardjohn:
> Companies that use Spire or Spiffe identites outside of kubernetes do not often follow the kubernetes naming scheme Istio does.

But this doesn't address that, no? It just is slightly changing the format. SPIFFEs can be arbitrary, I am not sure how adding 1 path (but still keeping the /sa/ns part) helps align there?

>  This path prefix does not impact the current ossified structure Istio uses 

I am not sure I follow. The whole issue with ossification is we cannot change it. By definition changing it impacts the structure... Its like changing `http://` to allow `http-<prefix>://`. I am sure it fits more use cases, but it probably will also break the internet in unexpected ways

nmnellis:
> By definition changing it impacts the structure... Its like changing http:// to allow http-<prefix>://. I am sure it fits more use cases, but it probably will also break the internet in unexpected ways

Being ossified implies that this pattern is standard among the industry. Do you have examples of other technologies and use cases using this strict format outside of just Istio?

nmnellis:
closing until further traction

PR #2847, opened 2023-07-06
chore: Fix typo in VirtualService docs

mattkram:
Fixes https://github.com/istio/istio.io/issues/13514

A minor typo in the documentation.


howardjohn (approved):
ty

howardjohn:
/retest

nope, thanks!

ericvn (approved):
Thanks. This change will first show up in preliminary.istio.io docs. To make it to the istio.io docs, we need to cherry-pick this to the release-1.18 branch (if no cherry-pick it will show up when 1.19 is released which is where the preliminary site represents). If we do cherry-pick, we only update the istio.io docs on a patch release boundary so the change doesn't appear there quickly.

Edit: The cherry-pick will likely fail and need to be done manually.

PR #2860, opened 2023-07-11
proxy config: add support for PROXY protocol for gateways

kyessenov:
Adds an annotation on the gateway pods to enable PROXY parsing on the downstream. AWS and GCP support PROXY protocol in their TCP network load balancing infrastructures.  This annotation is a substitute for the difficult-to-support EnvoyFilter (e.g. https://preliminary.istio.io/latest/blog/2020/show-source-ip/#step-2-create-proxy-protocol-envoy-filter).


howardjohn (commented):
This seems like it could be appropriate here https://istio.io/latest/docs/ops/configuration/traffic-management/network-topologies/? I mean in `gatewayTopology`

kyessenov:
Yeah, that's the intended place for it. However, I don't think it makes sense to add it to ProxyConfig since that API applies to sidecars, and that's not the intent. We only want this for gateways.

howardjohn:
gatewayTopology also only applies to gateway though. You could make the
same argument for annotation and `gatewayTopology` I think - the only thing
making it "gateway" only is that it has "gateway" in the name and that we
only implement it for gateways?

kyessenov:
The difference is not only stylistic - pod granularity is the right abstraction, it should never be the default/installation config option. `proxy.istio.io/config` fields are also hard to discovery, when I go visit https://istio.io/latest/docs/reference/config/networking/proxy-config/, gateway topology is not there, so it makes harder to combine multiple options within `proxy.istio.io/config`.

howardjohn:
proxy config *is* at a pod granularity. The ProxyConfig CRD is not the same
as the meshconfig/proxy.istio.io struct
https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig
(which is confusing, but that's a problem unrelated to this that we can fix
out of band)

kyessenov:
Is there a reference page for things like gatewayTopology? I find it confusing why it's different in CRD and annotation, and I'm sure I'm not the only one.

howardjohn:
https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ProxyConfig
but for some reason its hide_from_docs. I think that is a mistake and we
can unhide it.

I also things its confusing so welcome improvements in that area

kyessenov:
Yeah, I like proxy config more as it's structured. Having it gateway topology exposed addresses my concern.

mesh/v1alpha1/proxy.proto
```diff
@@ -290,10 +287,17 @@ message Topology {
     ALWAYS_FORWARD_ONLY = 5;
   }
 
-  // $hide_from_docs
   // Configures how the gateway proxy handles x-forwarded-client-cert (XFCC)
   // header in the incoming request.
   ForwardClientCertDetails forward_client_cert_details = 2;
+
+  // PROXY protocol configuration.
+  message ProxyProtocolConfiguration {
```
costinm, line 295: I'm not a big fan of an empty message used instead of a boolean.

mesh/v1alpha1/proxy.proto
```diff
@@ -290,10 +287,17 @@ message Topology {
...
+  }
+
+  // Enables [PROXY protocol](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) for
+  // downstream connections on a gateway.
+  ProxyProtocolConfiguration enable_proxy_protocol = 3;
```
costinm, line 300: Is this something we want on ALL Gateway ports and listeners ? I guess with K8S Gateway targetRef can point to a section/port - but that won't work with Istio gateways. 

Not sure what's the right solution here - with K8S Gateway it would be bad to have a list of ports since targetRef section does the same. 

@howardjohn

costinm (commented):
Some of the comments you mentioned you want this for gateways only - any reasons sidecars shouldn't support this ? Technically an external load balancer could connect to a sidecar without an extra hop ( if it implements whatever policies the user needs ), we shouldn't require an extra hop.

kyessenov:
> Some of the comments you mentioned you want this for gateways only - any reasons sidecars shouldn't support this ? Technically an external load balancer could connect to a sidecar without an extra hop ( if it implements whatever policies the user needs ), we shouldn't require an extra hop.

The technical reason is that sidecars have a lot of sniffing applied that do not play well with PROXY (which explicitly discourages sniffing). The more important reason is that I'm not aware of any use case for a direct sidecar external TCP PROXY support.

mesh/v1alpha1/proxy.proto
kyessenov: This is to accommodate TLVs or port filters.

mesh/v1alpha1/proxy.proto
kyessenov: GCP recommends using PROXY on NEG health ports. Since Istio gateway expose both gateway listeners and default health listeners, I think it's OK to have PROXY on all ports.

costinm:
I don't mind having PROXY on all ports with the old Gateway API.

But in the new API - I think the model is to be able to attach extensions
to a section ( == listener ) only, and for use case it is
reasonable for a load balancer to have a port exposed to a global LB ( like
GCP or other vendors ) and another port
used in the VPC or for calls without PROXY.  The API also allows collapsing
multiple Gateway.

For the other use case - external load balancers directly balancing to
workloads - I would say this is the best practice
if the external load balancer supports mTLS or is on a secure network (
like a CNI implementing ambient or equivalent security). It
doesn't make sense to have an extra hop - we do it because most current
non-istio LBs don't support mTLS and ambient
is not yet ready.

I think new APIs should be forward looking and compatible with both ambient
and new Gateway API, there is no point
launching things that will force migrations later or will be incompatible
with the current standard - adding the targetRef is a must IMO.

BTW - I don't mind if the initial implementation only supports Gateways and
applies to all ports, we can document that.

On config options - before we add them, can you do a small survey on what
options are supported by AWS/GCP/other vendors
that currently support PROXY ? Since the purpose is to integrate with
existing infra - the API should be work on all, so I wouldn't
add any extra option in the enablement.

You mentioned the TLV and some mappings - if we can find out what people
actually use it would be good, and we can also add them
later, I think it's an advanced case. But we should document in the API
what TLVs are supported by default and how they are exposed/used.

kyessenov:
There are several comments that I'll try to unpack, so excuse me if I miss some:

1. Why ProxyConfig?

This is for consistency with the gateway topology setting for X-F-F. I think it makes sense to keep them together for the migration as well.

2. Why not ztunnel?

Simply, out of scope. There's a well-defined customer request to support inter-operation of external gateway with cloud infrastructure, and PROXY is a common solution. I am aware that PROXY can fully replace HBONE, in fact, I had a prototype that did it. Envoy implementation of PROXY is production ready, and ztunnel is not there yet.

3.  Why use a message for config?

To accommodate future evolution. However, I don't think the demand for TLVs or other settings is strong enough to require it now, but I can imagine it may become a necessity, so a simple boolean does not suffice.

howardjohn:
This LGTM as long as Costin is fine with moving forward. As a nit
`proxy_protocol: {}` seems like it may be better than
`enable_proxy_protocol: {}` as if we expanded this it feels a bit awkward.
I don't mind the current one if we want though, just a subjective nitpick.

``

kyessenov:
@ericvn @linsun  @louiscryan 
Need a secondary TOC review.

keithmattix (approved):
Based on the small scope and our discussion in the WG meeting, I have no issue with this

mesh/v1alpha1/proxy.proto
hzxuzhonghu: I am sure this is not good, as it is not agile to allow customizing on different port and different proxy protocol version, etc

kyessenov: We can add port and other guards with https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#envoy-v3-api-msg-config-listener-v3-listenerfilterchainmatchpredicate. But let's wait for it to be requested.

hzxuzhonghu: Thanks for the response. sgtm

PR #2871, opened 2023-07-28
mark opencensus deprecated

zirain:
xref: https://opentelemetry.io/blog/2023/sunsetting-opencensus/

### DIFF
--- a/mesh/v1alpha1/config.pb.go
+++ b/mesh/v1alpha1/config.pb.go
@@ -1902,6 +1902,7 @@ type MeshConfig_ExtensionProvider_Stackdriver struct {
 
 type MeshConfig_ExtensionProvider_Opencensus struct {
 	// Configures an OpenCensusAgent tracing provider.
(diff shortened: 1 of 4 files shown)


hzxuzhonghu (approved):

howardjohn (commented):
can we add a release note telling users what to do (move to otel, I assume).

otherwise LGTM

costinm (approved):
Looks good - should we add notes to the other providers and indicate the intent to deprecate and warn users off ? I assume there are no objections to adopting otel and using 
otel collectors as integration point with vendor protocols.

Also not sure if deprecate + hide from docs is the correct path - normally we want users to see the field and deprecation warning in the generated docs.

zirain:
> Looks good - should we add notes to the other providers and indicate the intent to deprecate and warn users off ? I assume there are no objections to adopting otel and using otel collectors as integration point with vendor protocols.
> 
> Also not sure if deprecate + hide from docs is the correct path - normally we want users to see the field and deprecation warning in the generated docs.

Plan to add warning notes in `istioctl analysis`

howardjohn (approved):

PR #2872, opened 2023-07-28
accesslog: ability to disable builtin labels

zirain:
xhref: https://github.com/envoyproxy/envoy/pull/28046


howardjohn (commented):
LGTM, do we want it to be default on? Probably users shouldn't be looking at these internals -- and if they do, they can always manually add them? So maybe not even have an option...

costinm:
Agree with John, should be off by default - we shouldn't have envoy specific metrics. 

But looking at this - in context of semantic conventions - if they used the right labels it may not be bad to have few of them.  Assuming we set them correctly at startup - zone is clear.   Is envoy 'cluster' equivalent with canonical service in Istio ?  Does anyone remember where we use it ?  And for node name - should we set it to the pod name instead ? 

Either way - better we set them explicitly when we generate the log and disable this envoy-specific info by default, I don't think we need a config.

zirain:
> LGTM, do we want it to be default on? Probably users shouldn't be looking at these internals -- and if they do, they can always manually add them? So maybe not even have an option...

Prefer to make it default on, that will make some properties will dismiss after upgrade to latest version, I'm not sure it's a good practice.

hzxuzhonghu (approved):

mesh/v1alpha1/config.proto
```diff
@@ -999,6 +999,11 @@ message MeshConfig {
           // ```
           google.protobuf.Struct labels = 2;   
       }
+
+      // $hide_from_docs
+      // If specified, Envoy will generate built-in resource labels
+      // like ``log_name``, ``zone_name``, ``cluster_name``, ``node_name``.
+      bool enable_builtin_labels = 5;
```
howardjohn, line 1006: Do we need this at all? Can't they just set it in `labels` field explicitly?

Sorry I didn't mention this earlier...

zirain: that would be a huge break change than I expected before.

I think only `node_name` is meaningful in a real world, but envoy seems cannot get node metadata in accesslog.

PR #2903, opened 2023-08-14
Update the doc for disabling access log

lei-tang:
This PR is for the issue: https://github.com/istio/istio/issues/46510.


howardjohn (changes requested):
I am a bit confused. why does

```yaml
apiVersion: telemetry.istio.io/v1alpha1
kind: Telemetry
metadata:
  name: default
spec:
  metrics:
  - overrides:
    - disabled: true
```

work but the access log one does not?

lei-tang:
https://github.com/istio/istio/issues/46510 was closed (fixes: https://github.com/istio/istio/pull/46531, https://github.com/istio/istio.io/pull/13730).

PR #2904, opened 2023-08-16
add support for network wasm filters

ramaraochavali:
Adds support for specifying plugin type


ramaraochavali:
@kyessenov PTAL

type/v1beta1/selector.proto
```diff
@@ -66,3 +69,14 @@ enum WorkloadMode {
   // source or destination of the network traffic.
   CLIENT_AND_SERVER = 3;
 }
+
+// ProtocolSelector is the criteria for specifying if a policy can be applied to
+// traffic of a specific protocol.
+enum ProtocolSelector {
+  // Applies Policy to HTTP traffic only.
+  HTTP = 0;
```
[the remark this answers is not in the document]
ramaraochavali: I was thinking of using HTTP to align with current default. But we can do that code.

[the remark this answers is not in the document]
kyessenov: I think it's fine to leave HTTP as the default. Saying UNDEFINED = HTTP indirectly adds to the confusion.

type/v1beta1/selector.proto
```diff
@@ -41,6 +41,9 @@ message WorkloadSelector {
 message PortSelector {
   // Port number
   uint32 number = 1;
+
+  // Protocol used by the port
+  ProtocolSelector protocol = 2;
```
kyessenov, line 46: Does this force protocol handling on the port? I'm confused what it means to say port: 80, protocol: TCP on inbound, for example. Does that apply to HTTP traffic as well?

type/v1beta1/selector.proto
```diff
@@ -66,3 +69,19 @@ enum WorkloadMode {
...
+// traffic of a specific protocol.
+enum ProtocolSelector {
+  // Default value, which will be interpreted by its own usage.
+  P_UNDEFINED = 0;
+
+  // Applies Policy to HTTP traffic only.
+  HTTP = 1;
+  
+  // Applies Policy to TCP traffic only.
+  TCP = 2;
```
kyessenov, line 83: I'd prefer us not relying on protocol sniffing at all. Instead, we should say that network filters apply on all ports (including HTTP), and then provide guidance how to detect HTTP traffic within Wasm.

ramaraochavali: Are you saying we ONLY apply network WASM (if a flag is chosen) on the selected port and if not apply HTTP default? But would n't that be too restrictive?

type/v1beta1/selector.proto
kyessenov: HTTP is TCP based though? How do you apply network filters to HTTP chains then?

hzxuzhonghu, line 46: I am doubting how it works.  Say i specify a port 80, it is HTTP, but can I also specify ProtocolSelector = TCP?

costinm: We keep having those discussions. Next will be - what if HTTP is over QUIC and UDP ?

The problem is that we model the API on how envoy generates config, not what makes sense to user. 

IMO we should keep L4 policies separate from L7 policies - in particular given that L4 policies may someday be implemented by ztunnel. Running WASM in rust  is not hard - I know we're now saying "we'll never add features
to ztunnel" - but maybe some other implementation will handle L4 and support WASM, or maybe we change our mind. 

So maybe we not use 'protocol' - but keep separate L4 under a different name/resource (or some other mechanism), and always apply L4.

costinm: Agreed, it is confusing. That's why they defined the OSI layers...

type/v1beta1/selector.proto
costinm: I don't know what would be restrictive about having deterministic and clear behavior. 

Completely agree - TCP policies don't require any sniffing ( obviously ), and for HTTP it is far better to be explicit, users typically know what ports are HTTP in particular if they have custom WASM code. 

100% of not adding any more APIs that depend or assume sniffing. In K8S we have HttpRoute and TcpRoute, there is no ambiguity.

type/v1beta1/selector.proto
kyessenov: +1, that's what I was thinking, too. Network wasm API is essentially a copy of HTTP Wasm API that can bind to any port, just like HTTP Wasm API. We add a field at the very top of API to indicate it's "network" to save on creating another CR.

Network Wasm can bind to a port that has HTTP routes, that's sort of irrelevant from L4 perspective.

ramaraochavali: Are you suggesting add a flag at higher level some where in WasmPlugin message `enable_network_filter` (defaults to false existing behaviour) and if set, add it in the listener directly?
If yes, that sounds reasonable to me.

kyessenov: Yes, probably enum, with the default value HTTP_FILTER and a new value NETWORK_FILTER. In the future we can add LISTENER_FILTER, etc.

ramaraochavali:
Updated to use Enum in the top level API that tells what type extension should be used

kyessenov (approved):

zirain (approved):

costinm (approved):

ramaraochavali:

hzxuzhonghu (approved):

extensions/v1alpha1/wasm.proto
```diff
@@ -344,6 +344,19 @@ message WasmPlugin {
...
+
+  // Specifies the type of Wasm Extension to be used.
+  PluginType type = 14;
+}
+
+
+// PluginType indicates the type of Wasm Extension to be used.
+enum PluginType {
+  // Use HTTP Wasm Extension.
+  HTTP = 0;
```
howardjohn, line 356: Please make this `undefined = 0`; we may not need it now but it always comes back to byte us when we don't.

extensions/v1alpha1/wasm.proto
```diff
@@ -344,6 +344,19 @@ message WasmPlugin {
...
   // the traffic passes the WasmPlugin.
   repeated TrafficSelector match = 12;
+
+  // Specifies the type of Wasm Extension to be used.
+  PluginType type = 14;
+}
+
+
+// PluginType indicates the type of Wasm Extension to be used.
+enum PluginType {
```
howardjohn, line 354: can we be a bit more descriptive one what these mean? Maybe link to the ABIs for HTTP and Network, whether network applies to HTTP as well or not, etc

ramaraochavali:
@howardjohn added docs. PTAL

extensions/v1alpha1/wasm.proto
```diff
@@ -344,6 +344,30 @@ message WasmPlugin {
...
+  // Specifies the type of Wasm Extension to be used.
+  PluginType type = 14;
+}
+
+
+// PluginType indicates the type of Wasm Extension to be used. 
+// There are two types of Extensions: HTTP and NETWORK.
+// HTTP Extension work at "Layer 7"(for example as a HTTP filters in Envoy). 
+// The detailed HTTP interface for can be found at [C++] (https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/b7e690703c7f26707438a2f1ebd7c197bc8f0296/include/proxy-wasm/context_interface.h#L199)
+// and [RUST] (https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/6b47aec926bc29971c727471d6f4c972ec407c7f/src/traits.rs#L309).
```
howardjohn, line 357:
```suggestion
// and [Rus] (https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/6b47aec926bc29971c727471d6f4c972ec407c7f/src/traits.rs#L309).
```

extensions/v1alpha1/wasm.proto
```diff
@@ -344,6 +344,30 @@ message WasmPlugin {
...
+// NETWORK Extension work at "Layer 4"(for example as a Network filters in Envoy).
```
howardjohn, line 358:
```suggestion
// NETWORK Extension work at "Layer 4" (for example, as a network filter in Envoy).
```

extensions/v1alpha1/wasm.proto
```diff
@@ -344,6 +344,30 @@ message WasmPlugin {
...
   repeated TrafficSelector match = 12;
+
+  // Specifies the type of Wasm Extension to be used.
+  PluginType type = 14;
+}
+
+
+// PluginType indicates the type of Wasm Extension to be used. 
+// There are two types of Extensions: HTTP and NETWORK.
+// HTTP Extension work at "Layer 7"(for example as a HTTP filters in Envoy).
```
howardjohn, line 355:
```suggestion
// HTTP Extension work at "Layer 7" (for example as an HTTP filter in Envoy). 
```

extensions/v1alpha1/wasm.proto
```diff
@@ -344,6 +344,30 @@ message WasmPlugin {
...
   // the traffic passes the WasmPlugin.
   repeated TrafficSelector match = 12;
+
+  // Specifies the type of Wasm Extension to be used.
+  PluginType type = 14;
+}
+
+
+// PluginType indicates the type of Wasm Extension to be used. 
+// There are two types of Extensions: HTTP and NETWORK.
```
howardjohn, line 354:
```suggestion
// There are two types of Extensions: `HTTP` and `NETWORK`.
```

howardjohn (approved):
LGTM just some small comment suggestions

extensions/v1alpha1/wasm.proto
zirain: rust

hanxiaop (approved):

PR #2910, opened 2023-08-22
add metadata headers for proxyHeaders API

GregHanson:
part of review discussion from https://github.com/istio/proxy/pull/4895 and https://github.com/istio/istio/pull/46579

in lieu of annotations or env var feature flags, include stripping the istio mx headers in the proxyHeaders API:

```
proxyHeaders:
  metadataExchangeHeaders:
    mode: ALWAYS|IN_MESH
    includePassthrough: true|false
```

Skip appending the `X-Envoy-Peer-Metadata` and `X-Envoy-Peer-Metadata-Id` headers on outbound requests to services not within the mesh.  Can optionally include `PassthroughCluster` in this behavior


GregHanson:
cc @howardjohn @linsun

mesh/v1alpha1/proxy.proto
```diff
@@ -640,6 +640,18 @@ message ProxyConfig {
...
+      // Default alwyas include istio metadata exchange headers in outbound requests from sidecars 
+      ALWAYS = 0;
+
+      // Only append the istio metadata exchange headers for services considered in mesh
+      IN_MESH = 1;
+    }
+    message MetadataExchangeHeaders {
+      MetadataExchangeMode mode = 1;
+      // If set, the metadata headers will also be removed for traffic directed through PassthroughCluster
+      google.protobuf.BoolValue includePassthrough = 2;
```
hzxuzhonghu, line 653: We should expose the implement detail to users. pass through cluster is a implement tech

linsun: Agreed, is there any value to users to expose this? e.g. if I configure in_mesh, would not that exclude passthrough for MX headers?

GregHanson: Kuat raised a concern with passthrough being used for in mesh as well:
>PASSTHROUGH can be used in the mesh, not just externally . . . . because it can be in-cluster it has a potential to be backwards incompatible

linsun: @kyessenov can you elaborate a scenario for this for in mesh? I think this would confuse users to have such a config and I would rather remove it.

linsun: @howardjohn thoughts?

kyessenov: Agree we don't need an API for this. We were always deliberate about treating MX headers as an implementation detail and not codify them as API or specification. So I'd prefer environment variables I think until we get some user adoption.

Re: PASSTHROUGH - sure, you can have mesh endpoint behind a PASSTHROUGH cluster for a variety of reasons I believe, e.g. mesh-in-mesh scenario where VIP/host is resolved within the client by gRPC or some other SDK. There'd be a breaking change by losing the associated telemetry if we drop MX then - so it's not a safe default.

GregHanson: removed `includePatthrough` flag in https://github.com/istio/api/pull/2910/commits/5eaa811977d49c942898e60e635cf8a63035e4b1

GregHanson:
the original ProxyHeaders API commit didn't have one.  I think the better place for release note will be in the actual istio/istio implementation PR

mesh/v1alpha1/proxy.proto
```diff
@@ -640,6 +640,16 @@ message ProxyConfig {
     message EnvoyDebugHeaders {
       google.protobuf.BoolValue disabled = 1;
     }
+    enum MetadataExchangeMode {
+      // Default alwyas include istio metadata exchange headers in outbound requests from sidecars
```
thiagogcm, line 644: typo: always

mesh/v1alpha1/proxy.proto
```diff
@@ -640,6 +640,16 @@ message ProxyConfig {
     message EnvoyDebugHeaders {
       google.protobuf.BoolValue disabled = 1;
     }
+    enum MetadataExchangeMode {
+      // Default alwyas include istio metadata exchange headers in outbound requests from sidecars 
+      ALWAYS = 0;
+
+      // Only append the istio metadata exchange headers for services considered in mesh
+      IN_MESH = 1;
```
kyessenov, line 648: API needs to say explicitly what is IN_MESH. Istio does not have authoritative info on peers being in mesh, so it should say how the user marks things in mesh or not.

mesh/v1alpha1/proxy.proto
GregHanson: done

mesh/v1alpha1/proxy.proto
GregHanson: @kyessenov updated PTAL

hzxuzhonghu (commented):
IIUC, we may remove metadata exchange very soon to use mds. @kyessenov correct me if i am wrong
 So i doubt if a feature env can be used temporarily.

mesh/v1alpha1/proxy.proto
kyessenov: Thanks, it's better. Do we have a better way to "doesn't match any listener"? I think that's confusing to end-users. Maybe @howardjohn has a better phrasing?

kyessenov:
@hzxuzhonghu Yes, we want WDS but not sure how soon that'll happen. No strong opinion on API vs env variables since ProxyConfig is also alpha so we can deprecate later.

linsun:
@kyessenov @howardjohn is this ready to be approved?

mesh/v1alpha1/proxy.proto
GregHanson: > We should not expose the implement detail to users. pass through cluster is a implement tech

Was trying to avoid saying PassthroughCluster directly for the reason above, so I copied the wording from the comments around use of PasthroughCluster in [istio/istio ](https://github.com/istio/istio/blob/master/pilot/pkg/networking/util/util.go#L61)

mesh/v1alpha1/proxy.proto
```diff
@@ -664,6 +675,11 @@ message ProxyConfig {
     // See the [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/router/v3/router.proto#envoy-v3-api-field-extensions-filters-http-router-v3-router-suppress-envoy-headers) for more details.
     // These headers are enabled by default if not configured.
     EnvoyDebugHeaders envoy_debug_headers = 5;
+    // Controls Istio metadata exchange headers `X-Envoy-Peer-Metadata` and `X-Envoy-Peer-Metadata-Id`.
+    // If enabled, these headers will be added on outbound request headers from sidecars.
+    // If disabled, these headers will not be set.
+    // These headers are enabled by default if not configured. These headers are not present in ambient enabled pods.
```
kyessenov, line 681: I'd suggest not saying they are enabled by default. The API goes both ways - once you say they are enabled they have to be always enabled, which is not what we want to state. Something more open-ended - the default is unspecified.

mesh/v1alpha1/proxy.proto
kyessenov: What about ambient upstreams from sidecars? Do we want to say they are in mesh?

GregHanson: you mean `sidecar -> ambient-enabled-pod`?  The ambient enabled pods are part of istio's service registry, and are not explicitly defined as `MESH_EXTERNAL` via ServiceEntry - so I would definitely include that as "in-mesh".  Envoy is aware of these pods and therefore will have listeners defined for them

mesh/v1alpha1/proxy.proto
howardjohn: If we want to leave that possibility (which i argee with) I think we also want the `0` enum to be `UNDEFINED`

howardjohn, line 681: I would remove the note about ambient, its unclear what it means and may just cause confusion

mesh/v1alpha1/proxy.proto
```diff
@@ -640,6 +640,17 @@ message ProxyConfig {
     message EnvoyDebugHeaders {
       google.protobuf.BoolValue disabled = 1;
     }
+    enum MetadataExchangeMode {
```
howardjohn, line 643: might want to document on L612

mesh/v1alpha1/proxy.proto
```diff
@@ -640,6 +640,17 @@ message ProxyConfig {
...
+      // Default always include istio metadata exchange headers in outbound requests from sidecars 
+      ALWAYS = 0;
+
+      // Only append the istio metadata exchange headers for services considered in mesh
+      // Out of mesh traffic includes requests to MESH_EXTERNAL ServiceEntries and traffic that does not match any listener in envoy
```
howardjohn, line 648:
```suggestion
      // Only append the istio metadata exchange headers for services considered in mesh.
      // Traffic is considered in-mesh if it is secured with Istio mutual TLS. This means that `MESH_EXTERNAL` services, unmatched passthrough traffic, and requests to workloads without Istio enabled will be considered out of mesh.
```

howardjohn (commented):
minor nits but lgtm

mesh/v1alpha1/proxy.proto
GregHanson: done

mesh/v1alpha1/proxy.proto
GregHanson: done

GregHanson: switched to default `UNDEFINED`

howardjohn (approved):

PR #2919, opened 2023-08-29
Update virtualservice HTTPRetry docs for the `attempt` field

Smeb:
Some folks communicated that they were confused about how `attempts` could impact the maximum number of requests made.


dominik-riha (approved):

hzxuzhonghu (commented):
you should update the proto file and then `make gen`

howardjohn (commented):
LGTM but you need to change the .proto files then run `make gen`

ericvn:
/ok-to-test

Will flag the `make gen` is required with the a test.

Will also flag a release note to pass (non-required).

Smeb:
Thanks for the feedback - should be correct now.

howardjohn (approved):
thanks!

ericvn (approved):

PR #2930, opened 2023-09-15
ProxyConfig: introduce ProxyStatsConfig

zirain:
inspired by https://github.com/istio/istio/issues/46933

this pacth will allow you more ability to change [stats_config](https://github.com/istio/istio/blob/b210df81d5c091d50734dcb9d1b2fb41b09e6b86/tools/packaging/common/envoy_bootstrap.json#L60)

<del>
case 1: enable `use_all_default_tags`
```yaml
meshConfig:
  proxyStatsConfig:
    enableDefaultTags: true
```
</del>

case 2: add welllkonwn tag like 
```yaml
meshConfig:
  proxyStatsConfig:
    tags:
      - name: "envoy.cluster_name"
        regex: "^cluster\\.((.+?)\\.).+"
      - tag_name: envoy.grpc_bridge_service
        regex: "grpc\\.((.+)\\.).+?\\..+$"
      - tag_name: envoy.grpc_bridge_method
        regex: "grpc\\..*\\.((.+?)\\.).+$"
```


zirain:
release-notes will be added when implement in istio/istio.

/retest

mesh/v1alpha1/proxy.proto
```diff
@@ -512,6 +512,32 @@ message ProxyConfig {
...
+    }
+
+    // Tags can be added by configuring the telemetry extension. Each additional tag needs to be present in this list.
+    // Extra tags emitted by the telemetry extensions must be listed here so that they can be processed
+    // and exposed as Prometheus metrics.
+    repeated Tag tags = 2;
+  }
+
+  // Configuration for processing of stats.
+  // $hide_from_docs
```
kyessenov, line 538: It would help to have an example config and some use case that it solves.

mesh/v1alpha1/proxy.proto
```diff
@@ -512,6 +512,32 @@ message ProxyConfig {
   // and exposed as Prometheus metrics.
   repeated string extra_stat_tags = 27;
 
+  // Proxy stats configuration.
+  message ProxyStats {
+    // Use all default tag regexes specified in .
```
howardjohn, line 517: specified in ....?

howardjohn (commented):
I don't really understand this. Why do we need this ? Can you give some background?

mesh/v1alpha1/proxy.proto
zirain: original stats:
```
cluster.greeter_service.grpc.helloworld.Greeter.SayHello.upstream_rq_time
cluster.greeter_service.grpc.helloworld.Greeter.SayHello.0: 9
cluster.greeter_service.grpc.helloworld.Greeter.SayHello.request_message_count: 9
cluster.greeter_service.grpc.helloworld.Greeter.SayHello.response_message_count: 9
cluster.greeter_service.grpc.helloworld.Greeter.SayHello.success: 9
cluster.greeter_service.grpc.helloworld.Greeter.SayHello.total: 9
```

with configuration:
```bash
- tag_name: envoy.cluster_name
  regex: "^cluster\\.((.+?)\\.).+"
- tag_name: envoy.grpc_bridge_service
  regex: "grpc\\.((.+)\\.).+?\\..+$"
- tag_name: envoy.grpc_bridge_method
  regex: "grpc\\..*\\.((.+?)\\.).+$"
```

result will be:
```
# TYPE envoy_cluster_grpc_0 counter
envoy_cluster_grpc_0{envoy_grpc_bridge_service="helloworld.Greeter",envoy_grpc_bridge_method="SayHello",envoy_cluster_name="greeter_service"} 9
# TYPE envoy_cluster_grpc_request_message_count counter
envoy_cluster_grpc_request_message_count{envoy_grpc_bridge_service="helloworld.Greeter",envoy_grpc_bridge_method="SayHello",envoy_cluster_name="greeter_service"} 9
# TYPE envoy_cluster_grpc_response_message_count counter
envoy_cluster_grpc_response_message_count{envoy_grpc_bridge_service="helloworld.Greeter",envoy_grpc_bridge_method="SayHello",envoy_cluster_name="greeter_service"} 9
# TYPE envoy_cluster_grpc_success counter
envoy_cluster_grpc_success{envoy_grpc_bridge_service="helloworld.Greeter",envoy_grpc_bridge_method="SayHello",envoy_cluster_name="greeter_service"} 9
# TYPE envoy_cluster_grpc_total counter
envoy_cluster_grpc_total{envoy_grpc_bridge_service="helloworld.Greeter",envoy_grpc_bridge_method="SayHello",envoy_cluster_name="greeter_service"} 9
```

costinm:
- Isn't this something that prometheus or otel collector can rewrite with their existing API ? 
- is it a mesh admin, global setting - or something we expect each user to want to customize differently for each app ? If the later - is it something we want to do via the annotation override - or does it belong to the telemetry API as a proper CRD ? 
- is it a static, boot-time option - or something that could be changed at runtime and without restarting ( as it would 
be the case if we used prometheus or otel settings ) ? 

Also: we do have an option to use a custom bootstrap config file. Assuming this is for advanced users, wouldn't be 
better to let users do that instead ?

zirain:
> * Isn't this something that prometheus or otel collector can rewrite with their existing API ?

Yes, I think they can

> * is it a mesh admin, global setting - or something we expect each user to want to customize differently for each app ? If the later - is it something we want to do via the annotation override - or does it belong to the telemetry API as a proper CRD ?

we can also override via annotations, like `sidecar.istio.io/extraStatTags`.
I don't think it belongs to Temeletry CRD, cause it required a rollout/restart.

> * is it a static, boot-time option - or something that could be changed at runtime and without restarting ( as it would
>   be the case if we used prometheus or otel settings ) ?

it's static, require to restarting.

> Also: we do have an option to use a custom bootstrap config file. Assuming this is for advanced users, wouldn't be better to let users do that instead ?

Custom bootstap is another way to achieve this.

BTW, after switch to native cpp stats fiter, I think `sidecar.istio.io/extraStatTags` is meaningless, this PR also introduce an alternative API.

mesh/v1alpha1/proxy.proto
howardjohn: `envoy_cluster_grpc_0` is it supposed to be 0? Seems like a bug

zirain: seems like, but not cause by the api.

kyessenov:
Let me understand the context for this. The way I understand it, the goal is to expose the Envoy filter metrics in "tagged" form (Istio metrics produce them as tags, so no need to do extraction). That seems important since 1) we do allow direct EnvoyFilter injection, and we need telemetry from them, 2) existing metrics for HCM etc are very narrow and could leverage tags when expanded

So in general, I agree we should support it. I don't have a strong opinion on whether do it via a separate annotation or proxyconfig anntotation, it's equivalent from the UX perspective.

I'd request a few minor changes:
1. Please reference Envoy docs about tag extraction, it's documented well there.
2. I don't think we need "all default tags" boolean, and inclusion prefix "envoy" does the same.
3. We should deprecate extraStatTags and give an example to migrate to this API.

mesh/v1alpha1/proxy.proto
```diff
@@ -510,8 +510,54 @@ message ProxyConfig {
...
+    repeated Tag tags = 1;
+  }
+
+  // Configuration for processing of stats.
+  // To enable all Envoy default tag regexes, For example:
+  //
+  // ```
+  // meshConfig:
+  //   proxyStatsConfig:
+  //     enableDefaultTags: true
```
kyessenov, line 542: stale, maybe supply stats matcher config?

kyessenov (approved):
I'm fine with placement either way - proxy stats matcher proto might be good enough, depends on what else we plan to add (otel stuff?)

mesh/v1alpha1/proxy.proto
zirain: yes, that's the future plan, moving all stats fields here.

zirain:
> I'm fine with placement either way - proxy stats matcher proto might be good enough, depends on what else we plan to add (otel stuff?)

Next, it will add stats sink under ProxyStatsConfig.

zirain:
kindly ping @istio/technical-oversight-committee

linsun:
Hi @zirain - we discussed this PR in today's WG meeting (thanks to @kyessenov), Qs for you- 

- Could you provide more details on the use cases? Do you have one or more users asking for it?
- Can we provide better usability than regex?  At least create some examples on how to use it?

Thanks!

zirain:
> * Could you provide more details on the use cases? Do you have one or more users asking for it?

After switch to native filter, `extra_stat_tags` is meaningless to `istio_*` metrics.
For user like https://github.com/istio/istio/issues/46933, how can I set tag extraction rule like following:
```yaml
      - name: "envoy.cluster_name"
        regex: "^cluster\\.((.+?)\\.).+"
      - tag_name: envoy.grpc_bridge_service
        regex: "grpc\\.((.+)\\.).+?\\..+$"
      - tag_name: envoy.grpc_bridge_method
        regex: "grpc\\..*\\.((.+?)\\.).+$"
```

`Custom Envoy Bootstrap Configuration` would be another option, if the TOC thinks it's good enough, I'm OK to close this proposal.

> * Can we provide better usability than regex?  At least create some examples on how to use it?

Envoy support Regex and FixedValue, what would be better?

https://github.com/envoyproxy/go-control-plane/blob/699014c269c004427f6fc3f10a68a9a6e259068a/envoy/config/metrics/v3/stats.pb.go#L372 

For https://github.com/istio/istio/issues/46933, you can use 

```yaml
meshConfig:
  proxyStatsConfig:
    tags:
      - name: "envoy.cluster_name"
        regex: "^cluster\\.((.+?)\\.).+"
      - tag_name: envoy.grpc_bridge_service
        regex: "grpc\\.((.+)\\.).+?\\..+$"
      - tag_name: envoy.grpc_bridge_method
        regex: "grpc\\..*\\.((.+?)\\.).+$"
```
instead of `Custom Envoy Bootstrap Configuration`.

PR #2934, opened 2023-09-20
Add DNSSettings for Tracer Configuration (API part)

aatarasoff:
A new section called `dns_settings` has been added to the Tracer Configuration. The purpose of this addition is to limit the Istio proxy's DNS queries to 10 minutes instead of the default 30 seconds when Zipkin tracing is enabled. This can be achieved by defining the below structure in `MeshConfig` using this PR:
```yaml
dns_settings:
  refresh_rate: "600s"
  respect_ttl: false
```


howardjohn (changes requested):
This DNS issue only applies to the legacy tracing configuration. I think a more appropriate fix would be to transition to https://istio.io/latest/docs/tasks/observability/telemetry/ ?

costinm:
Does this config apply to ALL tracers ? I agree with John, it's something that otel collector should handle.

linsun (commented):
@aatarasoff thank you for your contribution!

Could you check if it make sense to add it to the Telemetry API (per John mentioned earlier) given this is telemetry related config?

A small tip- for API change, might be good to present to our Wed's working group meeting.

hzxuzhonghu:
I think dns refresh rate is not only to tracers, but to all DNS services. We have a `DnsRefreshRate` in MeshConfig now, but actually we do not allow setting `respect_ttl`, which is hardcoded to true.

We can add a respect_ttl to meshConfig, and allow agent to get both values

hzxuzhonghu:
Find that we have also supported dynamic bootstrap discovery, which can be used to customize the whole bootstrap config

PR #2935, opened 2023-09-25
Add Ambient Custom Service Search Paths

kdorosh:
Add api to reflect solution 2 per design: https://docs.google.com/document/d/1uDWoWxHyMCE4oUc-nTJPfVoQyikZP-UMp-_BrAA-PQE

Resolves the security flaw with the ServiceEntry API by adding new mesh admin controls to handle how Istio handles DNS-intercept and resolving user-defined hosts.

Most of the context lives in the design doc, and has been discussed during the ambient contributor meetings now several weeks in a row with the general consensus that "this is strictly better than what we have today".

I don't think there is a perfect solution that is:
- backwards compatible
- aligns perfectly with ambient's "safe to enable by default"
- includes `ServiceEntry` in its current shape, which we cannot drop entirely at this point in the project

I do think this is the best compromise among all those requirements above, as well as those listed out in the design doc.


mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,33 @@ message MeshConfig {
   // For ISTIO_MUTUAL TLS settings, use meshMTLS configuration.
   TLSConfig tls_defaults = 64;
 
+  // An ordered namespace list for service resolution. Similar to DNS resolution
```
louiscryan, line 1177: I don't think we need to mention to the Envoy filter matching mechanism :)

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,33 @@ message MeshConfig {
...
+  // and envoy filter chain matching, progress down the list until a match is
+  // found. If no match is found then the hostname does not resolve.
+  //
+  // It’s recommended that users set:
+  // meshConfig:
+  //   serviceResolutionOrder:
+  //    - '.'
+  //
+  // If serviceResolutionOrder is omitted ambient workloads will see this
```
louiscryan, line 1186: I'd put the default first and use it to explain the syntax of "." and "*" even though its not quite ideal.

mesh/v1alpha1/config.proto
louiscryan, line 1177: Need to explain that this does not affect names that are K8S namespace scoped even if we're not letting users control what that means yet. Use examples to show it affecting non-scoped names.

costinm, line 1177: Isn't it too broad feature and possibly risky ?

An enum "ambientServiceEntryResolution" - with default value "." and an option "LEGACY" or "MIGRATION" would be safer and simpler.

We don't want users to start adding random namespaces and all the complexity that comes with it.

What about 'istio-system' (or istio-config) ? We can have 3 options - default (., istio-config), "legacy" (*), "restricted" (istio-config only)

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,33 @@ message MeshConfig {
   // For ISTIO_MUTUAL TLS settings, use meshMTLS configuration.
   TLSConfig tls_defaults = 64;
 
+  // An ordered namespace list for service resolution. Similar to DNS resolution
+  // and envoy filter chain matching, progress down the list until a match is
+  // found. If no match is found then the hostname does not resolve.
+  //
+  // It’s recommended that users set:
+  // meshConfig:
+  //   serviceResolutionOrder:
```
louiscryan, line 1183: nit:

Note sure I'd call the field and 'Order'.  'Paths' if we want to harken back to DNS or 'Namespaces'

also if we're going to hav this be for non namespace scoped things we might want to try to incorporate that into the name somehow..

E.g.

customServiceResolutionScopes
externalServiceResolutionScopes

louiscryan:
Just nits but overall LG

mesh/v1alpha1/config.proto
linsun: Good comment above from @louiscryan  - as we don't control k8s service resolution order so I'd also prefer not call it `serviceResolution..`

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,33 @@ message MeshConfig {
...
+  //    - '.'
```
linsun, line 1184: worth explain what . means in the comment

mesh/v1alpha1/config.proto
linsun: Worth clarify if this flag is only for ambient

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,33 @@ message MeshConfig {
...
+  //
+  // If serviceResolutionOrder is omitted ambient workloads will see this
+  // behavior as the default. For backwards compatability, sidecar workloads
+  // will see the following default:
+  // meshConfig:
+  //   serviceResolutionOrder:
+  //    - '.'
+  //    - '*'
```
linsun, line 1192: can this be arbitrary namespace? Also, what about istio root namespace (by default it is istio-systems but can be any namespace), is it always on the path automatically?

mesh/v1alpha1/config.proto
kdorosh: I added comments to the `'.'` and `'*'` fields as they're used to help explain the syntax.

I do think it's clear already since this matches the syntax of exportTo, and further I think it's clearer as written by building up towards the complex example from the simpler one.

mesh/v1alpha1/config.proto
kdorosh: good call out, added this

mesh/v1alpha1/config.proto
kdorosh: good call out, I renamed the field to customServiceResolutionPaths

mesh/v1alpha1/config.proto
kdorosh: I added a comment to this field to explain what this means. It's just like exportTo

mesh/v1alpha1/config.proto
kdorosh: It can be any namespace, I added a comment to demonstrate that. Istio-system is not included by default but users can add that if desired

mesh/v1alpha1/config.proto
kdorosh: > Isn't it too broad feature and possibly risky ?
>
> An enum "ambientServiceEntryResolution" - with default value "." and an option "LEGACY" or "MIGRATION" would be safer and simpler.

I don't think we want to add two fields here when only one is needed

> We don't want users to start adding random namespaces and all the complexity that comes with it.

I don't think this is any more complex than just doing it for istio-system, so might as well be more flexible.

How do you see your idea offering any less complexity? We would treat the first * the same as seeing legacy/migration on the implementation side. Same for istio-system vs any other named namespace.

kdorosh:
>  I would like to know the use cases for that extra flexibility.

What I had in mind:

```yaml
customServiceResolutionPaths:
 - 'my-admin-ns' # defined here, admin takes control of hosts. no override
 - '.' # per user service
 - 'my-admin-overrideable-ns' # admin can define a shared fallback for a host that users can override
```

This is the only use case that I think would merit the extra flexibility, otherwise I'm on board with an enum and limited controls.

This configuration allows support for two use cases simultaneously:
- defining globally shared external services (admin only)
- defining global service fallbacks while allowing per-user overrides

This use case is a combination of ideas @howardjohn mentioned in this [comment](https://docs.google.com/document/d/1uDWoWxHyMCE4oUc-nTJPfVoQyikZP-UMp-_BrAA-PQE/edit?disco=AAAA3axDE4Q). However, I will admit it strictly builds on and combines the initial ideas, so I'm happy to simplify the controls here if no one else has concerns. cc @louiscryan @linsun @nrjpoddar

costinm:
On Tue, Sep 26, 2023 at 10:44 AM Kevin Dorosh ***@***.***>
wrote:

> I would like to know the use cases for that extra flexibility.
>
> What I had in mind:
>
> customServiceResolutionPaths:
>  - 'my-admin-ns' # defined here, admin takes control of hosts. no override
>  - '.' # per user service
>  - 'my-admin-overrideable-ns' # admin can define a shared fallback for a host that users can override
>
> This is the only use case that I think would merit the extra flexibility,
> otherwise I'm on board with an enum and limited controls.
>

Ok, but we already have istio config - for admin controlled settings (
defaulting to istio-system, but can be set ).
What is so special about ServiceEntry that requires not one - but 2
user-configurable admin namespaces  ?

I don't mind having a dedicated namespace for SE - but do we really need 2
? And a list ?



> This configuration allows support for two use cases simultaneously:
>
>    - defining globally shared external services (admin only)
>    - defining global service fallbacks while allowing per-user overrides
>
> I don't think we have any requirement to support multiple cases
simultaneously on the same setting.

Security doesn't work well with flexibility and multiplexing.

What I would do:
1. Ambient default: only admin can set SE in istio config. It is the safest
option - compatible with external DNS.
2. Enum value to allow 'legacy/insecure' SE for migration and 'same
namespace' - the same namespace option will NOT be compatible with external
DNS and should be validated against 'enable DNS interception'
( and should have big warnings that if the user has a secure resolver in
the pod on linked in the app - it won't work )
3. If you really want 'my-admin-ns' for SE - while all other admin settings
are in istio config - add a string option with
'IstioConfigForServiceEntries' or similar name, and change the docs for
istio config to indicate that it will no longer be used for SE if admin
wants them in separate NS.

What I expect long term is users adopting secure resolvers and 'enterprise
DNS' - plus controlers and secure API to control the DNS, which hopefully
will be K8S-wide and not istio specific.
'external-dns' project is not perfect - but they have a lot of good code
and features.

Yes, some users will need our DNS capture - but we should keep it as narrow
as we can possibly can to reduce confusion and risks ( and complexity ).

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,36 @@ message MeshConfig {
...
+  //    - '.' # local namespace
+  //    - '*' # wildcard; any namespace
+  //
+  // Users can also set this proactively for migrations from classic to ambient
+  // Istio to protect workloads that may depend on the original (but insecure)
+  // global-by-default nature of ServiceEntry in classic Istio.
+  //
+  // With support for '*' there can still be conflicting ServiceEntry hosts;
+  // The tiebreak uses the oldest creation timestamp.
+  repeated string custom_service_resolution_paths = 66;
```
howardjohn, line 1203: I am a bit concerned about the implemention of this in Ztunnel. I don't think we can implement it without sending the mesh config? Or some other new API type?

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,36 @@ message MeshConfig {
   // For ISTIO_MUTUAL TLS settings, use meshMTLS configuration.
   TLSConfig tls_defaults = 64;
 
+  // An ordered namespace list for custom service resolution. Similar to DNS,
+  // progress down the list until a match is found. If no match is found then the
+  // hostname does not resolve. This does not affect namespaced-scoped services that
+  // cannot conflict (e.g. $namespace.$name.svc.* or $namespace.$name.svcentry.*)
```
howardjohn, line 1180: I very strongly feel we should not implement a new FQDN style (`svcentry`). There is no use cases that it is helping, at least not that I have seen

howardjohn: Also `$namespace.$name.svc.*` doesn't actually work, afaik? Since you can make ServiceEntry with those hostnames but in the wrong namespace.

(nit: `$namespace.$name.svc.*` has the name and ns flipped)

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,36 @@ message MeshConfig {
...
+  // with other hosts.
+  //
+  // It’s recommended that users set:
```
howardjohn, line 1183: If we are going to recommend it, we should probably say why?

howardjohn: Why would a user do this? They can use Sidecar import?

If you only trust their Services not ServiceImport, then jsut import them: `hosts: [untrusted/*.untrusted.svc.cluster.local]`? Pretty sure we support wildcards

linsun: My understanding is users would need this for ambient as we don't support Sidecar resources for ambient.  For sidecars, you are right, this could be configured with Sidecar resources.

mesh/v1alpha1/config.proto
```diff
@@ -1174,8 +1174,36 @@ message MeshConfig {
...
+  //    - '.' # local namespace
+  //    # users may also consider setting istio-system
+  //
+  // If customServiceResolutionPaths is omitted ambient workloads will see the
+  // above behavior as the default. For backwards compatability, sidecar workloads
+  // will see the following default:
+  // meshConfig:
+  //   customServiceResolutionPaths:
+  //    - '.' # local namespace
+  //    - '*' # wildcard; any namespace
```
linsun, line 1195: for sidecars, how would this work with any Sidecar resources users have if there are conflicts?

mesh/v1alpha1/config.proto
howardjohn: Good point on ambient. I really don't like this. We are getting ourselves in trouble because Istio has made *tons* of bad decisions in the past, we are replicating them all in ztunnel, and then trying to bandage over them after the fact. the end result is complex, unintuitive, and confusing for users.

mesh/v1alpha1/config.proto
kdorosh: We already have access to mesh config https://github.com/istio/ztunnel/blob/2205b13dbeb28b2d65ffa8de1f225cf1f3c832a8/src/config.rs#L390-L403

kdorosh (commented):
updated per PR feedback to remove custom hostname formats and limit the API blast radius to ambient-only. Also updated the default to be global-only (no per-namespace DNS) to better align with future external-DNS solutions

mesh/v1alpha1/config.proto
kdorosh: decided to limit the blast radius of this change to ambient only

mesh/v1alpha1/config.proto
kdorosh: agree; removed the custom FQDN style

mesh/v1alpha1/config.proto
kdorosh: Agree that the UX is not ideal; however I think the latest iteration has the best forward-looking defaults (`ServiceEntry` only for mesh admins in istio-system; no per-namespace overrides) as well as the necessary toggles to support important APIs that users expect to work istio. in classic or ambient

linsun:
Thanks @kdorosh! I think it LGTM, limiting it to ambient makes me feel more comfortable.

Any feedback from @louiscryan @howardjohn @costinm? John I think you said you don't quite like it, curious if you have any better proposal?

louiscryan:
@costinm 

I don't think telling users to put services into istio-system is a great idea for real deployments as it co-mingles responsibilities and permissions in ways that seem a little dangerous. 

AS a strawman it seems like you'd want to have something like a set of namespaces for common admin controlled use-cases. E.g. "istio-system", "istio-egress", "istio-ingress" or somesuch

This would make documenting use-cases and providing samples much simpler overall. In that situation we could use an annotation on the namespace to indicate that it was enabled for service sharing and use regular RBAC to control that. Namespace creation RBAC and management is already pretty well aligned with the management function so I think this is aligned with what you're trying to do.

If we had that, or something similar, then I would be fine with the enum proposal. WDYT?

costinm:
Completely agree - istio-system is not right. But we already added istio-config ( and is configurable ). Question is if SE needs another namespace - and I don't mind that either, but would prefer a well-defined name not arbitrary. Some people need to support and troubleshoot this...

costinm:
+1 on istio-egress - the role of egress admin should handle it

louiscryan:
I think its going to be hard to be completely reliant on well-known names, hence the annotation suggestion. Do we feel like we can insist on the creation of 'istio-egress' and just bake that in. Im a little skeptical.

@istio/technical-oversight-committee 

Its definitely a good idea to normalize the usage patterns and docs by using well-known namespaces OOB just not sure we can insist on it all the time.

costinm:
Like istio-config and istio-system - we can allow advanced users to
configure (for example in case of multi-tenant istiod),
but I don't know many Istio users who rename istio-system just for fun. I
have no problem with istio admin setting it just
like they can set istio-system.

We don't allow each namespace to re-configure istio-system or istio-config
with annotations - and given the security
risks it can't be done by each namespace admin. In other words - I don't
think we want to end up again with
each namespace having its own DNS setup and forcing the use of DNS
capture/istio-DNS.

A single namespace, with clear owner - that may be synced with an
enterprise DNS or control istio-DNS when user
doesn't have an enterprise DNS - is far easier to protect and make sure
doesn't end up with conflicts.

costinm:
I think we should also look at the bigger picture here - istio is not
limited to a single cluster, and enterprise DNS is generally
per network / VPC.

Maybe we can check with customers who use managed or enterprise DNS or have
requirements around egress
policy enforcement  on what policies they have around allowing
modifications - while also checking if they
 are planning to stick with plain-text UDP DNS resolvers in order to be
able to use our DNS capture.

I know it can take a long time for cloud/enterprise to catch up with what
is now standard in consumer devices
and browsers - but many of those security improvements may happen in
languages and frameworks or in the
base images they use - so they could end up accidentally with a secure DNS
resolver and break any Istio
API that depends on using plain text insecure resolvers.

kdorosh:
What can we do to move this forward?

I'm not convinced the enterprise DNS or secure DNS objection related to future ServiceEntry DNS API breaks holds; that would also break classic istio DNS, which uses the same APIs. We already established that ServiceEntry is not a perfect API and that the intended ambient support is for backwards compatibility, not GAMMA or better future APIs. So I don't think we have to block on hypothetical future requirements here.

Also I don't think we can force users to have a single egress namespace in all cases; users may depend on the service entry workload selector (which only selects within a namespace). Obviously we can recommend it to play nicely with enterprise DNS but I don't think we can require it.

kdorosh:
closing as an ambient non-priority for now upstream. can be revisited at a future date

PR #2937, opened 2023-09-27
fix broken envoy docs link

rsalmond:
looks like envoy dumped all the api-v2 docs

old link: https://web.archive.org/web/20210421173338/https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#envoy-api-enum-config-filter-network-http-connection-manager-v2-httpconnectionmanager-forwardclientcertdetails

new link: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto.html#enum-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-forwardclientcertdetails

### DIFF
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
@@ -4335,7 +4335,7 @@ policy.</p>
 <section>
 <p>ForwardClientCertDetails controls how the x-forwarded-client-cert (XFCC)
 header is handled by the gateway proxy.
-See <a href="https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#envoy-api-enum-config-filter-network-http-connection-manager-v2-httpconnectionmanager-forwardclientcertdetails">Envoy XFCC</a>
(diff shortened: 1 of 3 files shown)


kfaseela:
cc @phlax

rsalmond:
Hrm, if we're still using the V2 API's is this change a waste of time? I see a few other links to now broken V2 envoy docs that I could update if needed.

howardjohn (approved):

ericvn:
I believe there are more issues: https://github.com/istio/istio.io/issues/13494

phlax:
> looks like envoy dumped all the api-v2 docs

current versions no longer support v2 api so it is no longer documented

historical versions that do support it still have the relevant docs

rsalmond:
@phlax it appears in the link at the top of [this docs section](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#ForwardClientCertDetails).

phlax:
i see - so essentially istio latest - i guess linking to envoy latest is correct in that case but if you version these docs then you probably want to freeze the links

rsalmond:
Seems to be a common pattern.

```console
$ find . -name \*.proto -exec grep -Hi envoyproxy\.io\/docs\/envoy\/latest {} \; | wc -l
58
```

hzxuzhonghu (approved):

PR #2939, opened 2023-09-28
update expected filenames for tls terminating key/secret

rsalmond:
https://istio.slack.com/archives/C3TEGNZ7W/p1695876172041579

https://github.com/istio/istio/blob/c9b1d29e5a1ac5a7a193c999627a56308eedc858/security/pkg/pki/ca/ca.go#L55-L58

I believe this is what it's supposed to say.

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -684,7 +684,7 @@ message ServerTLSSettings {
   // For gateways running on Kubernetes, the name of the secret that
   // holds the TLS certs including the CA certificates. Applicable
   // only on Kubernetes. An Opaque secret should contain the following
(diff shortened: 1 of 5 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -684,7 +684,7 @@ message ServerTLSSettings {
   // For gateways running on Kubernetes, the name of the secret that
```
howardjohn, line 684: Both are valid, we should document both.

linsun:
@rsalmond can you get make gen passing? Thx for your contribution - i didn't even know both are valid!

hzxuzhonghu (approved):

linsun (approved):
LG, thank you @rsalmond !

PR #2940, opened 2023-10-02
Properly hideFromDocs targetRef

howardjohn:
Apparently it needs to be line 1.


ericvn:
Interestingly, there is no pb.html change with this there is no change in the docs?

ericvn (approved):

PR #2941, opened 2023-10-02
Migrate to protoc-gen-crd

howardjohn:
For https://github.com/istio/istio/issues/46151
Based on https://github.com/istio/api/pull/2946, will rebase once that merges

This PR moves our CRD generator from cue-gen to protoc-gen-crd. The details of these don't matter too much; the final generated CRD.yaml does. The migration value is not really exposed in this PR, as I tried to keep it minimal; a followup PR will be used where we implement https://github.com/istio/istio/issues/46151 (likely multiple PRs, to keep it small). For now, the following changes are present:

* REQUIRED fields are actually required. I validated each field that is currently set as REQUIRED is marked as required in the webhook, so this is not a breaking change
* The `oneOf` representation is simplified. In my testing, I have found these to be equivalent in functionality.

Overall this should have no user facing changes

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -25,7 +25,6 @@ repo_dir := .
 protolock = protolock
 annotations_prep = annotations_prep
 htmlproofer = htmlproofer
-cue = cue-gen -paths=common-protos
 
 #####################
 # Generation Rules
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -12,6 +12,9 @@ plugins:
 - name: golang-deepcopy
   out: .
   opt: paths=source_relative
+- name: crd
+  out: .
+  strategy: all
 - name: golang-jsonshim
   out: .
   opt: paths=source_relative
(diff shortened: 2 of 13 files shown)


ericvn (approved):

networking/v1alpha3/sidecar.proto
```diff
@@ -590,7 +590,7 @@ message IstioEgressListener {
   // specific ports while others have no port, the hosts exposed on a
   // listener port will be based on the listener with the most specific
   // port.
-  Port port = 1;
+  SidecarPort port = 1;
```
keithmattix, line 593: Possible dumb question but does changing the message break backwards compat?

howardjohn: I am 100% sure its YAML compatible, 80% sure its proto write compatible, and 100% sure its _not_ Go compatible (which we do not care about)

Also this was part of an old PR fwiw, just rebased. So the diff will be gone

keithmattix: Ah I missed this in the earlier one. Should be fine then since we're not sending protos across the wire

keithmattix (approved):

PR #2946, opened 2023-10-03
Align REQUIRED declaration with reality

howardjohn:
The `REQUIRED` proto label currently does nothing. As a result, some of
them are misaligned with reality -- fields marked REQUIRED are actually
not required, and often even have valid behaviors and are used in the
real world.

Unfortunately, many of these are fields that *should* be required but
are not. To maintain backwards compatibility, these cannot be made
required at this point, sadly.

In a future PR, I intend to make `REQUIRED` *actually make it required*,
so this PR is a pre-requisite to doing that.


ericvn (approved):

keithmattix (approved):
*sigh* I miss `required` and `optional` in proto2

hzxuzhonghu (commented):
why declare another SidecarPort? Does it break compatibility

PR #2951, opened 2023-10-06
Adopt CRD native validation

howardjohn:
Part of https://github.com/istio/istio/issues/46151. 

This is the first PR ramping up our CRD native validation. The intent here is to provide the same validation but without a webhook dependency.

To keep things small and start in a low-risk API, this PR introduces validation only for WasmPlugin. Most of this is pretty straightforward; the Go webhook validation logic is 1:1 mapped into native CRD concepts.

Testing this change (and future) ones has three levels:
* In this PR, introduced a new test suite. There is a big folder of CR yamls, and each one can declare itself valid or invalid. Each of these is run through validation and asserts the results.
* In a followup PR in istio/istio, the same test inputs are tested comparing the _webook_ response to the CRD validation response. This ensures we are not accidentally too strict or too lenient in our CRD validation; our goal is to be identical.
* In addition to ^, a fuzz test will be added that does the same but with randomized inputs. This will give us pretty solid evidence that this change is safe.

blocked by: https://github.com/istio/tools/pull/2687
blocked by: me running the fuzzer a bit locally to get confidence


costinm (commented):
+1 on the change - but I think we are over-complicating things. Great to have a test suite, but I don't think it needs to match 100% what the current validator does. If we can convert any tests the validator had to the new suite - it should be sufficient.

hzxuzhonghu (commented):
How does cue intergrate with kubebuilder

howardjohn:
We no longer use Cue because it did not support this

hzxuzhonghu (approved):

extensions/v1alpha1/wasm.proto
```diff
@@ -427,21 +438,27 @@ enum PullPolicy {
...
 
+// +kubebuilder:validation:XValidation:message="value may only be set when valueFrom is INLINE",rule="(has(self.valueFrom) ? self.valueFrom : '') != 'HOST' || !has(self.value)"
 message EnvVar {
-  // Required
-  // Name of the environment variable. Must be a C_IDENTIFIER.
-  string name = 1;
+  // Name of the environment variable.
+  // Must be a C_IDENTIFIER.
+  // +kubebuilder:validation:MaxLength=256
+  // +kubebuilder:validation:MinLength=1
```
ericvn, line 452: Nit: sometimes the min is before the max, sometimes its after.

PR #2952, opened 2023-10-06
Add max_concurrent_streams definition to DestinationRule

jaellio:
Part of https://github.com/istio/istio/issues/47166

**NOTE** This also impacts the sidecar resource which has a connectionPoolSetting field.


hzxuzhonghu:
Since we have `MaxRequestsPerConnection`, which is similar, why do we need another field?

keithmattix:
I personally would prefer to use a setting whose existence and behavior is ~guaranteed by the HTTP/2 spec rather than a proxy specific implementation.

jaellio:
Even though setting `max_requests_per_connection` for http/2 connections results in a similar outcome to setting `max_concurrent_streams`, adding support for `max_concurrent_streams` introduces minimal complexity and matches the user's expectation of settings available for http/2 connections.

hzxuzhonghu:
It is already confusing since we donot distinguish http1.1 and http2 connection pool settings,  I feel this will make things worse

jaellio:
How will we handle new HTTP settings that are only applicable to HTTP1 or HTTP2, but not both?

howardjohn (approved):

linsun:
@hzxuzhonghu do you have a better way to do this?

@jaellio the change LGTM, ready to approve if @hzxuzhonghu doesn't have strong objections to it.

hzxuzhonghu (changes requested):
I donot think `MaxConcurrentStreams` should be in tcp settings.
 If we want to explicitly distinguish it from HTTP setting, we can add it in http2Settings

keithmattix:
> I donot think `MaxConcurrentStreams` should be in tcp settings.
> 
>  If we want to explicitly distinguish it from HTTP setting, we can add it in http2Settings

This is a really good catch that I missed; indeed max_concurrent_streams does not belong in tcp settings. Good call out @hzxuzhonghu

linsun:
Great catch @hzxuzhonghu!  +1

jaellio:
> I donot think `MaxConcurrentStreams` should be in tcp settings. If we want to explicitly distinguish it from HTTP setting, we can add it in http2Settings

Yes, thank you, I missed that! Updated!

hzxuzhonghu (approved):

ramaraochavali:
Sorry for the late comment. To make this more clear that it only applies to http2, should we call it `http2_max_concurrent_streams` similar to how we distinguish other settings like `http2_max_requests` and `http1_max_pending_requests` . Ideally we should have different settings for http1/http2/http3 - but it is too late to change them

jaellio:
> Sorry for the late comment. To make this more clear that it only applies to http2, should we call it `http2_max_concurrent_streams` similar to how we distinguish other settings like `http2_max_requests` and `http1_max_pending_requests` . Ideally, we should have different settings for http1/http2/http3 - but it is too late to change them

@ramaraochavali good call out. I am fine with making this update, but I was confused about the descriptions of the fields that are prefixed with `http1` or `http2`. The descriptions for all of these fields include: "Please note that this is applicable to both HTTP/1.1 and HTTP2." https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings I understand that this field is specific to http2, but I didn't want to make it more confusing.

ramaraochavali:
I think those are different and this specific field is only for http2. I am OK with moving as is

PR #2961, opened 2023-10-11
Mirror DestinationRule connection pool configuration on Sidecar

ZackButcher:
Revives https://github.com/istio/api/pull/1754 to support the same `ConnectionPoolSettings` we use for DestinationRule on the inbound cluster via configuring the `Sidecar`.

Today developers _can_ configure the inbound cluster's connection pool, via `DestinationRule` -- Istiod programs both the client and server to use the same connection pool settings. However, when multiple services listen on the same port it can be challenging (impossible?) [for a user to ensure the correct DestinationRule is applying connection pooling settings on the server side](https://github.com/istio/istio/blob/8fbfd36e33cbe217c8218499ac775ceba6b21dbd/pilot/pkg/networking/core/v1alpha3/cluster_builder.go#L375-L379). This API allows the user to set their inbound connection pool explicitly (and differently than the DestinationRule they configure for callers) both at the top level of the sidecar as well as per-port.

There's a paired implementation PR in istio/istio that I'll link shortly. In that PR, I've implemented it so that the connection pool is configured for the server-side sidecar with the following precedence (highest to lowest):
- per-port `ConnectionPool` setting from Sidecar
- top-level `InboundConnectionPool` setting from Sidecar
- per-port `ConnectionPool` setting from DestinationRule
- top-level `TrafficPool.ConnectionPool` setting from DestinationRule

Client-side connection pools are unchanged, where per-port then top-level configuration from the DestinationRule still apply.


ZackButcher:
I'll get release notes and all the other bits in order once we're agreed on the bigger stuff like the field names, behavior of the API, and the documentation on it.

hzxuzhonghu:
@ZackButcher It can not express per port connection pool

ZackButcher:
What the DestinationRule does today, which I do in the implementation PR, is configure the connection pool on the inbound cluster that's generated for the pod (e.g. the listener `inbound|8080||`). We create one cluster like this per port for the service, so we can apply the settings per port. See https://github.com/istio/istio/pull/47313 and its test cases.

Again, this is exactly how the per-port connection pool settings configured by a `DestinationRule` work for inbound traffic, just made explicitly configurable via Sidecar in addition to inheriting from DestinationRule (which was not documented anywhere AFAIK).

networking/v1alpha3/sidecar.proto
```diff
@@ -576,6 +583,12 @@ message IstioIngressListener {
   // sidecar for requests originating from outside the mesh.
   // Currently supports only SIMPLE and MUTUAL TLS modes.
   ServerTLSSettings tls = 7;
+
+  // Settings controlling the volume of connections Envoy will accept from the network.
+  // This setting overrides the top-level default `inboundConnectionPool` to configure
+  // specific settings for this port. This configuration mirrors the `DestinationRule`'s
+  // [`PortTrafficPolicy.connectionPool`](https://istio.io/latest/docs/reference/config/networking/destination-rule/#TrafficPolicy-PortTrafficPolicy) field.
+  ConnectionPoolSettings connection_pool = 8;
```
howardjohn, line 591: Interesting, didn't consider we would want it at 2 levels. Part of me thinks we could do

```
ingress:
- connectionPool: ...
```

(no port)

for this case. But I think that is probably too complex.

networking/v1alpha3/sidecar.proto
```diff
@@ -528,6 +529,12 @@ message Sidecar {
   // detected defaults from the namespace-wide or the global default Sidecar.
   repeated IstioEgressListener egress = 3;
 
+  // Settings controlling the volume of connections Envoy will accept from the network.
+  // This default will apply for all inbound listeners and can be overridden per-port
+  // in the `Ingress` field. This configuration mirrors the `DestinationRule`'s
```
howardjohn, line 534: We may want to comment on the interaction between them? Its explained in the PR just could be in the comments here

howardjohn (approved):

networking/v1alpha3/sidecar.proto
ZackButcher: An ingress without a port is not allowed by validation today, hence the top level vs the per-port level.

networking/v1alpha3/sidecar.proto
ZackButcher: ACK, will include

ZackButcher:
@howardjohn PTAL at the updated comments

ZackButcher:
FWIW there's an argument we could support more than just `ConnectionPoolSettings` at the sidecar; we could potentially support the full `TrafficPolicy` from `DestinationRule` (though not all fields there make sense).

- One recurring ask of the project has been to enable the sidecar to terminate simple TLS from the app (effectively man-in-the-middle'ing the traffic) before re-encrypting with the mesh's mTLS. This would allow services that do TLS (e.g. HTTPS) internally to participate in the mesh at L7 without code changes to force the app to speak HTTP. In theory this could be achieved with `TrafficPolicy.TLS` settings on the `egress` stanza. (e.g. for in-mesh traffic, `app request > HTTPS > sidecar terminates with Sidecar.Egress.TLS data > sidecar does mesh mTLS to destination`)
- The opposite side of this is allowing apps that serves HTTPS to receive HTTP traffic from the mesh by having the sidecar initiate (and terminate) an HTTPS connection inside the pod (i.e. `client request via mesh mTLS > sidecar > HTTPS > app`). We could use the `TrafficPolicy.TLS` settings in the `Ingress` stanza to define this sidecar>app behavior as well.

Of course, some fields (like load balancing settings) wouldn't make sense in the sidecar inbound context. So then the whole discussion of do you use the same `TrafficPolicy` in `DestRule` and `Sidecar` or use a slightly different one in each, and so on.

As @keithmattix pointed out, seems to be a lot of similar `DestinationRule`-ish behavior to look at for sidecar inbound path.

howardjohn (approved):

networking/v1alpha3/sidecar.proto
```diff
@@ -479,6 +480,82 @@ import "networking/v1alpha3/virtual_service.proto";
...
+//   - port:
+//       number: 80
+//       protocol: HTTP
+//       name: somename
+//     connectionPool:
+//       http:
+//         http1MaxPendingRequests: 1024
+//         http2MaxRequests: 1024
+//         maxRequestsPerConnection: 1024
+//         maxRetries: 100
```
linsun, line 553: Would be good to show overrides here?

Or is it http connection pool is the same but user still has to list it here due to tcp setting is different?

networking/v1alpha3/sidecar.proto
```diff
@@ -576,6 +679,15 @@ message IstioIngressListener {
   // sidecar for requests originating from outside the mesh.
   // Currently supports only SIMPLE and MUTUAL TLS modes.
   ServerTLSSettings tls = 7;
+
+  // Settings controlling the volume of connections Envoy will accept from the network.
+  // This setting overrides the top-level default `inboundConnectionPool` to configure
```
linsun, line 684: Would be good to clarify if only overrides need to be specified or the entire connection pool config?

networking/v1alpha3/sidecar.proto
ZackButcher: Correct, I repeated mainly to highlight that the user has to repeat list because the settings are not merged.

The UX is probably improved if we merge the fields, but today for DestinationRule we do not merge at that level: we either replace the entire `ConnectionPoolSetting` or not but do not look inside it, and I'm trying to keep the behavior as close to DestinationRule as possible (principle of least surprise).

networking/v1alpha3/sidecar.proto
ZackButcher: We should probably have the conversation on merging: I think it would likely improve UX, but today we do not do it for `DestinationRule`. [Here's the code that merges per-Port and default DestinationRule config](https://github.com/istio/istio/blob/4482b870e91ce66b6cc776700cb4c94f1edc8dca/pilot/pkg/networking/util/util.go#L770). Merging wouldn't be super hard to do, but it would be a behavioral change compared to what we do today. In any case, whatever we do w.r.t. merging here I think we need to match in `DestinationRule` merging it's top-level and per-port settings, or the API gets harder for users to understand.

howardjohn: Prefer not to merge personally. Doesn't seem terribly compelling and adds a lot of complexity

networking/v1alpha3/sidecar.proto
```diff
@@ -576,6 +679,15 @@ message IstioIngressListener {
...
+  // specific settings for this port. This configuration mirrors the `DestinationRule`'s
```
hzxuzhonghu, line 685: This seems redundant comment as the `ConnectionPoolSettings` scheme is actually same one

networking/v1alpha3/sidecar.proto
hzxuzhonghu: +1, reduce merging

networking/v1alpha3/sidecar.proto
ZackButcher: Exactly -- I want to call that out explicitly for folks so they don't have to think as hard about it (rather than having them realize because the page links to the same place). Happy to cut this if we want to keep the docs more concise.

ZackButcher:
So, take away is no merging inside of ConnectionPoolSetting in this PR, which I agree with. If we want to take up that issue (and more generally, merging fields inside of `TrafficPolicy` in DestinationRule as well, to keep behavior consistent) that can happen in another issue/PR/discussion.

With that in mind, are there updates to the examples/docs that we should do to make it more explicitly than I already have that merging is not happening?

hzxuzhonghu (approved):

dasch:
Has there also been discussions about doing this for _egress_? In particular, we have applications that want to limit their exposure to specific dependency services, and e.g. setting up concurrency circuit breakers _on the source side_ would be beneficial. But the configuration would be specific to the source application (e.g. the level of concurrency), and so it would make sense to stick it in the Sidecar config for it.

howardjohn:
> Has there also been discussions about doing this for _egress_? In particular, we have applications that want to limit their exposure to specific dependency services, and e.g. setting up concurrency circuit breakers _on the source side_ would be beneficial. But the configuration would be specific to the source application (e.g. the level of concurrency), and so it would make sense to stick it in the Sidecar config for it.

The `workloadSelector` on DR can limit the source side

dasch:
> > Has there also been discussions about doing this for _egress_? In particular, we have applications that want to limit their exposure to specific dependency services, and e.g. setting up concurrency circuit breakers _on the source side_ would be beneficial. But the configuration would be specific to the source application (e.g. the level of concurrency), and so it would make sense to stick it in the Sidecar config for it.
> 
> The `workloadSelector` on DR can limit the source side.
> 
> Better docs around this would help, a lot

But that wouldn't help source service owners if they also want to retain the destination service's DR configuration, except for the specific circuit breakers that are necessary on the source side, right? Ideally there should be something similar to this, where source owners can have specific overrides for the stuff they care about.

PR #2988, opened 2023-11-13
Added transport socket match patch

hzxuzhonghu:
For https://github.com/istio/istio/issues/44342 https://github.com/istio/istio/issues/44342 https://discuss.istio.io/t/config-istio-proxy-protocol/14911 https://discuss.istio.io/t/enable-proxy-protocol-from-istio-to-upstreams/10321

Below is an example about how to insert a proxy protocol upstream transport. However, without this patch, users can only replace the whole cluster or append the proxy protocol transport at last. So `tlsMode-istio` `tlsMode-disabled` can take precedence over proxy protocol.

```yaml
// kind: EnvoyFilter
// metadata:
//   name: listener-filter-example
//   namespace: myns
// spec:
//   configPatches:
//   - applyTo: TRANSPORT_SOCKET_MATCH
//     match:
//       context: SIDECAR_OUTBOUND
//     patch:
//       operation: INSERT_FIRST
//       value:
//        name: envoy.transport_sockets.upstream_proxy_protocol
//        match: {}
//        transport_socket:
//          name: envoy.transport_sockets.upstream_proxy_protocol
//          typed_config:
//            '@type': type.googleapis.com/envoy.extensions.transport_sockets.proxy_protocol.v3.ProxyProtocolUpstreamTransport
//            config:
//              version: V2
//            transport_socket:
//              name: "envoy.transport_sockets.raw_buffer"
//              typed_config:
//                '@type': type.googleapis.com/envoy.extensions.transport_sockets.raw_buffer.v3.RawBuffer

### CODE REVIEW

**Comment by hzxuzhonghu** [comment]
It is for adding or replacing a transportsocket match,  from the original issues, adding upstream proxy protocol support is the most required

**Comment by hzxuzhonghu** [comment]
The semantics is very similar to listener filters patching

**Comment by howardjohn** [comment]
I mean can we add in the comments how it works? if I cannot understand it
probably users cannot?

**Comment by howardjohn** [comment]
and will be good to know why the existing support we have doesn't work

**Comment by hzxuzhonghu** [comment]
Understand, will add that

File: networking/v1alpha3/envoy_filter.proto [hunk 1]
```diff
@@ -500,6 +529,9 @@ message EnvoyFilter {
 
     // Applies the patch to the listener filter.
     LISTENER_FILTER = 11;
+
+    // Applies the patch to the cluster's transport socket match.
```
linsun, line 533: I saw the example is for `context: SIDECAR_OUTBOUND` - can you clarify if it can be used for other contexts as well in the comments?

linsun (commented):
left 1 comment, LG

networking/v1alpha3/envoy_filter.proto
hzxuzhonghu: In theory it can be added to any ctx,  I do believe this will be mostly used for gateway and sidecar outbound

networking/v1alpha3/envoy_filter.proto
```diff
@@ -429,6 +429,35 @@ import "networking/v1alpha3/sidecar.proto";
 //         name: "envoy.filters.listener.proxy_protocol"
 //         typed_config:
 //           "@type": "type.googleapis.com/envoy.extensions.filters.listener.proxy_protocol.v3.ProxyProtocol"
+//
+// The following example adds an `envoy.transport_sockets.upstream_proxy_protocol` transport socket match.
```
linsun, line 433: Could you explain what this achieve for users?

networking/v1alpha3/envoy_filter.proto
linsun: Any usage difference from gateway vs sidecar outbound that we should highlight?

hzxuzhonghu: no, juts the CTX

networking/v1alpha3/envoy_filter.proto
hzxuzhonghu: Yes.  This example is to add an upstream proxy protocol support, which can be used to keep the original ip of this pod.

On the server side, it can parse the proxy protocol to get the original src ip address.

hzxuzhonghu: BTW, proxy protocol is not the only transportsocket that can be added. https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/transport_socket/transport_socket listed all that supported by envoy

networking/v1alpha3/envoy_filter.proto
```diff
@@ -429,6 +429,37 @@ import "networking/v1alpha3/sidecar.proto";
...
+// spec:
+//   configPatches:
+//   - applyTo: TRANSPORT_SOCKET_MATCH
+//     match:
+//       context: SIDECAR_OUTBOUND
+//     patch:
+//       operation: INSERT_FIRST
+//       value:
+//        name: envoy.transport_sockets.upstream_proxy_protocol
+//        match: {}
```
howardjohn, line 451: Not sure how this would work, Istio will already have a `match: {}`?

howardjohn (changes requested):
Can you please explain why https://github.com/istio/istio/blob/bd9e52d05817b03b9811a761a818b2e76d356018/pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go#L85 doesn't work?

Is it because there is no way to remove an element, and they are matched in order?

cc @kyessenov

networking/v1alpha3/envoy_filter.proto
hzxuzhonghu: For eds cluster, we generated two transport socket by default. A match {} will hijack all the ts after it.
```
        "transportSocketMatches": [
            {
                "name": "tlsMode-istio",
                "match": {
                    "tlsMode": "istio"
                },

            {
                "name": "tlsMode-disabled",
                "match": {
                    "tlsMode": "istio"
                },
```

hzxuzhonghu:
>Is it because there is no way to remove an element, and they are matched in order?

There is no way to insert a upstream proxy tranport socket now. The existing hack ts merge is to handle tls/raw buffer

networking/v1alpha3/envoy_filter.proto
hzxuzhonghu: Actually the most graceful way would be  add two `upstream_proxy_protocol`, once with match `"tlsMode": "istio"`, and the other with `"tlsMode": "disabled"`,

kyessenov:
We have PROXY on downstream now in ProxyConfig? What does this feature look like as a first class API? Is this intended only on egress gateways?

hzxuzhonghu:
> We have PROXY on downstream now in ProxyConfig?

NO, the downstream proxy can be set on listener filter with envoyfilter. There are many usage in ingress gateway.

>What does this feature look like as a first class API? Is this intended only on egress gateways?

This is for upstream transport socket setting. Which means the outbound traffic will be sent with proxy protocol, so the upstream server could acquire the real IP address of the client.

TBH, it is not only for egress gateway, as the issues and also from https://discuss.istio.io/t/enable-proxy-protocol-from-istio-to-upstreams/10321, i can see user want to apply it to ingress gateway also.

kyessenov:
@hzxuzhonghu upstream proxy protocol is very different from downstream proxy protocol. Why do we need in-mesh PROXY? Isn't it the same as HBONE? You can fully implement HBONE with just PROXY.

hzxuzhonghu:
TBH, I can not distinguish in mesh or out mesh sometimes, For a sidecar, it may need to access an external service. I understand this case is still in mesh. So inoder to apply a proxy protocol TS,  need to specify `SIDECAR_OUTBOUND`

hzxuzhonghu:
i agree with you on simplicity.. Here i want to provide the general supporting

hzxuzhonghu:
@howardjohn any other question

howardjohn:
> Can you please explain why https://github.com/istio/istio/blob/bd9e52d05817b03b9811a761a818b2e76d356018/pilot/pkg/networking/core/v1alpha3/envoyfilter/cluster_patch.go#L85 doesn't work?
> 
> Is it because there is no way to remove an element, and they are matched in order?


still waiting to see the answer to this, might have gotten lost?

hzxuzhonghu:
>Is it because there is no way to remove an element, and they are matched in order?


This can be an important reason, we can not remove an element currently.
And we cannot insert a element in front/after of a specific one.
We can't also replace a specific element.

hzxuzhonghu:
>What if we just make it so you can set transportSocket?   Or we always merge TSMatch first instead of last?

If the transport socket match set, it can take precedence over transportSocket. 

What do you mean by ` merge TSMatch first instead of last`? Is this a hack merge, which is different from the protobuf merge, so how we can tell user this semantics.

**mergeTransportSocketCluster** is already a hack way, i am not sure what it does until after reading the whole code, and wonder how can normal users like me to know.

hzxuzhonghu:
> There are literally 100s of fields we can special case, the API will grow crazy (already is...).

This is an enum, not adding a new field under the EnvoyFilter. If it can resolve the upstream proxy protocol insert issues, i would rather not adding anything

howardjohn:
> > What if we just make it so you can set transportSocket?   Or we always merge TSMatch first instead of last?
> 
> If the transport socket match set, it can take precedence over transportSocket.
> 
> What do you mean by ` merge TSMatch first instead of last`? Is this a hack merge, which is different from the protobuf merge, so how we can tell user this semantics.
> 
> **mergeTransportSocketCluster** is already a hack way, i am not sure what it does until after reading the whole code, and wonder how can normal users like me to know.

Users already need to understand Istio and envoy internals for using envoy filter so I don't mind it being challenging to use really. plus we can always improve documentation 

> > There are literally 100s of fields we can special case, the API will grow crazy (already is...).
> 
> This is an enum, not adding a new field under the EnvoyFilter. If it can resolve the upstream proxy protocol insert issues, i would rather not adding anything

It has about the same cost, we still need a full bespoke implementation for this in istiod, etc. There are literally 100s of nested types within envoy, it's not sustainable to slowly add all of them one by one

hzxuzhonghu:
>Users already need to understand Istio and envoy internals for using envoy filter so I don't mind it being challenging to use really. plus we can always improve documentation

If you mean `mergeTransportSocketCluster ` looks good to you, I do not care much.  Come back to our target to make upstream proxy protocol work.

hzxuzhonghu:
@howardjohn Suggested to add native proxy protocol support in DestinationRule, rather than using envoyfilter. So would add new API design, and I've seen gateway API has protocol support too. Will align with it.

linsun:
> @howardjohn Suggested to add native proxy protocol support in DestinationRule, rather than using envoyfilter. So would add new API design, and I've seen gateway API has protocol support too. Will align with it.

I like it!  the upgrade to h2(https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-HTTPSettings-H2UpgradePolicy) is also defined in DR so this is quite intuitive for users to use.

hzxuzhonghu:
Thank you, will start to design it within DR in another pr.

PR #2998, opened 2023-11-20
OpenTelemetryTracingProvider : New configurations to export via HTTP

joaopgrassi:
Closes [#47835](https://github.com/istio/istio/issues/47835)

This PR is an initial iteration/attempt in extending the `OpenTelemetryTracingProvider` in MeshConfig with new configurations that allow exporting OpenTelemetry spans via HTTP instead of gRPC. 

The ability to export OTel spans via HTTP is a populate request in Envoy and support for it was added in this PR: https://github.com/envoyproxy/envoy/pull/29207 - starting from Envoy `v1.28.0`. This is also important due to the eminent OpenTracing deprecation in Envoy, starting from `v1.29.0`, since there's observability back-ends that don't support OTLP/gRPC and only OTLP/HTTP.

As discussed in Slack/WG meeting I'm opening this as a starting point to see what the community thinks and decide how we can move forward.

The new configs here can be used in the following way:

1. Add/change the OpenTelemetry extension provider in `MeshConfig`:

```yml
    - name: otel-tracing
      opentelemetry:
        port: 443
        service: my.olly-backend.com
        http: # >>> Here are new configs
          path: "/api/otlp/traces" # will be joined with service host name
          timeout: 10s
          headers:
          - name: "Some custom header"
            value: "some value"
```

2. Then would need to add a `ServiceEntry` for their observability back-end

```shell
k apply -n istio-system -f - <<EOF
kind: ServiceEntry
metadata:
  name: my-olly-backend
spec:
  hosts:
  - my.olly-backend.com
  ports:
  - number: 443
    name: https-port
    protocol: HTTPS
  resolution: DNS
  location: MESH_EXTERNAL
---
kind: DestinationRule
metadata:
  name: my-olly-backend
spec:
  host: my.olly-backend.com
  trafficPolicy:
    portLevelSettings:
    - port:
        number: 443
      tls:
        mode: SIMPLE
```

Then, redeploy/restart Envoys and then, they would be configured to send OpenTelemetry spans via HTTP to `my.olly-backend.com/api/otlp/traces`. In the end, this will configure the [http_service](https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/config/trace/v3/opentelemetry.proto#http_service) in Envoy.


mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1025,29 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. Specifies the configuration for exporting OTLP traces via HTTP.
```
howardjohn, line 1029: Should we just allow `service: otlp.default.svc.cluster.local` (gRPC), `service: http://otlp.default.svc.cluster.local/some/path` (tell it to use http and a path)?

timeout and headers can apply to gRPC as well I think?

mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1025,29 @@ message MeshConfig {
...
+      google.protobuf.Duration timeout = 2;
+
+      // Optional. Allows specifying custom HTTP headers that will be added
+      // to each OTLP/HTTP export request.
+      repeated HttpHeader headers = 3;
+    }
+
+    message HttpHeader {
+      optional string name = 1;
+      optional string value = 2;
```
howardjohn, line 1050:
```suggestion
      string name = 1;
      string value = 2;
```

mesh/v1alpha1/config.proto
joaopgrassi: > Should we just allow service: otlp.default.svc.cluster.local (gRPC), service: http://otlp.default.svc.cluster.local/some/path (tell it to use http and a path)?

The `service` today is used in `tracing.go` to find the cluster https://github.com/istio/istio/blob/master/pilot/pkg/networking/core/v1alpha3/tracing.go#L223 so I think we can't concatenate that in MeshConfig with the path. Unless we want to build something like, if HTTP service is configured, remove the path from `service` before looking up the cluster.


> timeout and headers can apply to gRPC as well I think?

The gRPC exporter in Envoy don't use/has configuration for those, so it's not going to be used.

mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1025,29 @@ message MeshConfig {
...
+    }
+
+    // Defines configuration for an OpenTelemetry tracing backend that accepts OTLP data via HTTP.
+    message OpenTelemetryTracingHttpService {
+      // REQUIRED. Specifies the path on the receiver for OTLP traces.
+      // The path is appended to the `service`
+      string traces_path = 1;
+
+      // REQUIRED. Specifies the timeout for the OTLP/HTTP export request.
+      google.protobuf.Duration timeout = 2;
```
zirain, line 1041: I think you can make this optional with a suitable default value, like `1s` or more?

joaopgrassi: You mean default in code, or more like `optional... [default = 1s]`?

zirain: make it optional, and comment what is the default value.

mesh/v1alpha1/config.proto
joaopgrassi: Fixed in [098dddd](https://github.com/istio/api/pull/2998/commits/098dddd3b123c936d44cbf6fd174db1900eb257b)

mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1025,31 @@ message MeshConfig {
...
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. Specifies the configuration for exporting OTLP traces via HTTP.
+      // When empty, traces will be exported via gRPC.
+      OpenTelemetryTracingHttpService otlp_http = 4;
+    }
+
+    // Defines configuration for an OpenTelemetry tracing backend that accepts OTLP data via HTTP.
+    message OpenTelemetryTracingHttpService {
```
hzxuzhonghu, line 1035: can we make it a generic HTTP options, so we can share it with other provider in future

mesh/v1alpha1/config.proto
howardjohn: Just remove the "Required" part of the comment and mention the default. The actual proto line `google.protobuf.Duration timeout = 2;` can stay

zirain: @joaopgrassi can you update this, otherwise LGTM.

joaopgrassi: Fixed in [3156a21](https://github.com/istio/api/pull/2998/commits/3156a21b9497ec2de3c14144f0ab0f89134d4f3d).

mesh/v1alpha1/config.proto
joaopgrassi: @hzxuzhonghu I changed it to `HttpService` now in [332c273](https://github.com/istio/api/pull/2998/commits/332c273cf0c0761f6f6f0a611d5d0ead3145cb76). Let me know what you think!. Thanks.

zirain (approved):
LGTM

hzxuzhonghu (approved):
We can move service and port in HTTPService, but not sure how to make it compatible

zirain:
> We can move service and port in HTTPService, but not sure how to make it compatible

you cannot assume it's HTTP.

mesh/v1alpha1/config.proto
joaopgrassi: @howardjohn  you're right, I was looking at the wrong place. It does support headers and timeout as well https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/core/v3/grpc_service.proto#L295-L302. I guess it then makes sense to move those out of the new message type?

joaopgrassi: So, if we move headers and timeout the only thing left is the `path`. @howardjohn do you think adding the path to the `service` makes sense given my comment above in how it is used?

howardjohn: I would prefer to use the `service` approach I think

joaopgrassi: @howardjohn Ok so you mean also adding the path to the `service` and then manipulating it so we can still use it as we do today to lookup the cluster?

joaopgrassi: The problem I see with this approach of appending the path to the `service`, is that users will have a different behavior depending on what exporter they use. If grpc, don't add path, if http add the path. Not sure if that's a good idea, now that I think about it.

mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1025,32 @@ message MeshConfig {
...
+
+    // Defines configuration for an HTTP service that can be used by an Extension Provider.
+    // that does communication via HTTP.
+    message HttpService {
+      // REQUIRED. Specifies the path on the service.
+      string path = 1;
+
+      // Optional. Specifies the timeout for the HTTP request.
+      // If not specified, the default is 3s.
+      google.protobuf.Duration timeout = 2;
```
linsun, line 1042: Would it be possible to provide some example? a little hard for me to envision how to use it.

linsun (commented):
would like to see some examples to understand how to use this

joaopgrassi:
> would like to see some examples to understand how to use this


@linsun  You mean how users would configure Istio with this? I can give some example:

1. Would need to add/change the OpenTelemetry extension provider in `MeshConfig`:

```yml
    - name: otel-tracing
      opentelemetry:
        port: 443
        service: my.olly-backend.com
        otlp_http: # >>> Here are new configs
          path: "/api/otlp/traces"
          timeout: 10s
          headers:
          - name: "Some custom header"
            value: "some value"
```

2. Then would need to add a `ServiceEntry` for their observability back-end

```shell
k apply -n istio-system -f - <<EOF
kind: ServiceEntry
metadata:
  name: dynatrace-https
spec:
  hosts:
  - my.olly-backend.com
  ports:
  - number: 443
    name: https-port
    protocol: HTTPS
  resolution: DNS
  location: MESH_EXTERNAL
---
kind: DestinationRule
metadata:
  name: my-olly-backend
spec:
  host: my.olly-backend.com
  trafficPolicy:
    portLevelSettings:
    - port:
        number: 443
      tls:
        mode: SIMPLE
```

Then, redeploy/restart Envoys and then, they would be configured to send OpenTelemetry spans via HTTP to `my.olly-backend.com/api/otlp/traces`.

mesh/v1alpha1/config.proto
joaopgrassi: Let me know if this helps: https://github.com/istio/api/pull/2998#issuecomment-1910674868

mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1025,32 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. Specifies the configuration for exporting OTLP traces via HTTP.
+      // When empty, traces will be exported via gRPC.
+      HttpService otlp_http = 4;
```
howardjohn, line 1031:
```suggestion
      HttpService http = 4;
```

howardjohn: otlp is redundant since we are already in otel section

mesh/v1alpha1/config.proto
howardjohn: ok I am fine with current approach

howardjohn (commented):
LGTM, just one minor name change

linsun:
> > would like to see some examples to understand how to use this
> 
> @linsun You mean how users would configure Istio with this? I can give some example:
> 
>     1. Would need to add/change the OpenTelemetry extension provider in `MeshConfig`:
> 
> 
> ```yaml
>     - name: otel-tracing
>       opentelemetry:
>         port: 443
>         service: my.olly-backend.com
>         otlp_http: # >>> Here are new configs
>           path: "/api/otlp/traces"
>           timeout: 10s
>           headers:
>           - name: "Some custom header"
>             value: "some value"
> ```
> 
>     2. Then would need to add a `ServiceEntry` for their observability back-end
> 
> 
> ```shell
> k apply -n istio-system -f - <<EOF
> kind: ServiceEntry
> metadata:
>   name: my-olly-backend
> spec:
>   hosts:
>   - my.olly-backend.com
>   ports:
>   - number: 443
>     name: https-port
>     protocol: HTTPS
>   resolution: DNS
>   location: MESH_EXTERNAL
> ---
> kind: DestinationRule
> metadata:
>   name: my-olly-backend
> spec:
>   host: my.olly-backend.com
>   trafficPolicy:
>     portLevelSettings:
>     - port:
>         number: 443
>       tls:
>         mode: SIMPLE
> ```
> 
> Then, redeploy/restart Envoys and then, they would be configured to send OpenTelemetry spans via HTTP to `my.olly-backend.com/api/otlp/traces`. In the end, this will configure the [http_service](https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/config/trace/v3/opentelemetry.proto#http_service) in Envoy.

Yep, could we have some of these in the API comment in this PR? Not everyone can ping you for usage like me in github :)

mesh/v1alpha1/config.proto
linsun: Agree with John here

hanxiaop (approved):
Looks like the right direction.

mesh/v1alpha1/config.proto
joaopgrassi: Done in [314d908](https://github.com/istio/api/pull/2998/commits/314d908750d72f541a6c456078d8457147ffb158)

joaopgrassi:
> Yep, could we have some of these in the API comment in this PR? Not everyone can ping you for usage like me in github :)

@linsun you mean add this to the PR description? If so, I added it now (also to the linked issue). My plan though is to add documentation with a full example when the feature is complete in Istio.

linsun:
> > Yep, could we have some of these in the API comment in this PR? Not everyone can ping you for usage like me in github :)
> 
> @linsun you mean add this to the PR description? If so, I added it now (also to the linked issue). My plan though is to add documentation with a full example when the feature is complete in Istio.
> 
> Edit: Never mind, I added now to the proto file, so the examples should show in the generated html file.

Thank you!  Yep most of times, I'm looking at proto and generated html to understand how to use things. The example you added was helpful!

linsun (approved):

linsun:
Per @zirain's req, we discussed the request to cherry pick this to Istio 1.21 branch in today's istio WG meeting. The general consensus is all api changes will require TOC approval to cherry pick if this misses the release branch cutting. Given there is still impl PR open on #3002  and we don't see a strong reason to make an exception for cherry pick this API change, we prefer to not cherry pick and wait till it lands in istio 1.22.

linsun:
cc @istio/release-managers-1-21

PR #2999, opened 2023-11-20
Add idle_timeout to DestinationRule.TcpSettings

jewertow:
This PR aims to enable setting idle timeout per outbound TcpProxy. Currently, this is only possible with `ISTIO_META_IDLE_TIMEOUT`, which affects all TCP services. This is problematic for users who want to disable idle timeout for a particular service, while keeping short timeout for others. Of course, another option is EnvoyFilter, but it's problematic for less advanced users.


networking/v1alpha3/destination_rule.proto
```diff
@@ -744,6 +744,12 @@ message ConnectionPoolSettings {
     // was established. If not set, there is no max duration. When max_connection_duration
     // is reached the connection will be closed. Duration must be at least 1ms.
     google.protobuf.Duration max_connection_duration = 4;
+
+    // The idle timeout for TCP connections.
+    // The idle timeout is defined as the period in which there are no bytes sent or received on either
+    // the upstream or downstream connection.
+    // If not set, the default idle timeout is 1 hour. If set to 0s, the timeout will be disabled.
+    google.protobuf.Duration idle_timeout = 5;
```
hzxuzhonghu, line 752: There is a limitation, how can we apply to inbound TCP proxy when there are many idle timeouts set for a single port

jewertow: Could you elaborate more on this case? When DR is applied to inbound listeners? Isn't it applied to outbound only?

howardjohn, line 752: Maybe same as https://github.com/istio/api/pull/2999/files#r1399954705, but isn't idle timeout applied on tcp_proxy? But DR is more specific.

Similar to tunnelsettings though. So its broken for weighted cluster I imagine? is that accurate?

If so, can we document it?

hzxuzhonghu: DR is not applied to either Inbound or Outbound listener now. It applies to outbound/inbound cluster, and the connectionPool is shared by `Sidecar`.

As you meant to apply idle timeout to listener's TCP proxy, so I think it is also for inbound listener. There can exist more than one service has the same target port.

ramaraochavali: I think DR seems like a bad choice for idle timeout of inbound listener. As @hzxuzhonghu mentioned, DR is applied mainly for upstream clusters

jewertow: > DR is not applied to either Inbound or Outbound listener now. It applies to outbound/inbound cluster

@hzxuzhonghu I think you are not 100% right here. DR applies to cluster setting in case of HTTP protocol. In TCP, some DR settings are applied to listener, because Envoy defines them in the [TcpProxy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto#extensions-filters-network-tcp-proxy-v3-tcpproxy) listener, e.g. `max_downstream_connection_duration` and `tunneling_config` are set in [buildOutboundNetworkFiltersWithSingleDestination](https://github.com/istio/istio/blob/682a0d820ee26f619f1791d0d8f55102fb28b0b1/pilot/pkg/networking/core/v1alpha3/networkfilter.go#L87). And `idle_timeout` is just another one.

> I think DR seems like a bad choice for idle timeout of inbound listener.

@ramaraochavali my intention was not to configure inbound listeners.

jewertow: I see that I didn't clarify in the PR description that it's intended for outbound purposes. I'm sorry for confusing you. I already updated the description.

jewertow: > Maybe same as https://github.com/istio/api/pull/2999/files#r1399954705, but isn't idle timeout applied on tcp_proxy?

Yes, it is applied on tcp_proxy.

> Similar to tunnelsettings though. So its broken for weighted cluster I imagine? is that accurate?

Yes, you're right. Another DR setting, which is not applied to weighted clusters is `maxConnectionDuration`.

> If so, can we document it?

Of course, I will do it.

hzxuzhonghu: From end-user's view, I wish with one config, can setup both inbound/outbound TCP proxy and clusters. This is a full path.

TBH i am not quite sure how we can design an API suitable for this

jewertow: Done

jewertow:
Do you still have any concerns to approve it? As I already said, idle timeout has the same semantics as already existing `ConnectionPoolSettings.TCPSettings.MaxConnections` or `ConnectionPoolSettings.TCPSettings.TunnelSettings`.

networking/v1alpha3/destination_rule.proto
```diff
@@ -760,6 +760,16 @@ message ConnectionPoolSettings {
     // was established. If not set, there is no max duration. When max_connection_duration
     // is reached the connection will be closed. Duration must be at least 1ms.
     google.protobuf.Duration max_connection_duration = 4;
+
+    // The idle timeout for TCP connections.
+    // The idle timeout is defined as the period in which there are no bytes sent or received on either
+    // the upstream or downstream connection.
+    // If not set, the default idle timeout is 1 hour. If set to 0s, the timeout will be disabled.
+    // Idle timeout is not configured per each cluster individually when weighted destinations are used,
+    // because idle_timeout is a property of a listener, not a cluster. In that case, idle_timeout
```
howardjohn, line 769:
```suggestion
    // because `idleTimeout` is a property of a listener, not a cluster. In that case, `idleTimeout`
```

howardjohn (approved):

networking/v1beta1/destination_rule.proto
```diff
@@ -709,6 +709,16 @@ message ConnectionPoolSettings {
...
+
+    // The idle timeout for TCP connections.
+    // The idle timeout is defined as the period in which there are no bytes sent or received on either
+    // the upstream or downstream connection.
+    // If not set, the default idle timeout is 1 hour. If set to 0s, the timeout will be disabled.
+    // Idle timeout is not configured per each cluster individually when weighted destinations are used,
+    // because idle_timeout is a property of a listener, not a cluster. In that case, idle_timeout
+    // specified in a destination rule for the first weighted route is configured in the listener,
+    // which means also for all weighted routes.
+    google.protobuf.Duration idle_timeout = 5;
```
hzxuzhonghu, line 721: So it also applies to http protocol options `HttpProtocolOptions,IdleTimeout`, right?

jewertow: No, this option is only for TCP. `HttpProtocolOptions` has its own `idleTimeout`, which is applied to a cluster.

networking/v1alpha3/destination_rule.proto
jewertow: Done

networking/v1beta1/destination_rule.proto
hzxuzhonghu: But max_connection_duration applies to http options

jewertow: Ok, right. I didn't think about applying TCP idle timeout to HTTP as well, because HttpProtocolOptions already has this setting, but I can see in the documentation of [TcpSettings](https://istio.io/latest/docs/reference/config/networking/destination-rule/#ConnectionPoolSettings-TCPSettings) that "Settings common to both HTTP and TCP upstream connections.", so it might make sense for the sake of consistency. And then TcpSettings.IdleTimeout would be applied to TCP as well as HTTP, while HttpProtocolOptions.IdleTimeout only to HTTP. Does it make sense?

hzxuzhonghu (approved):

PR #3000, opened 2023-11-24
Outlier detection support in DestinationRule for grpc proxyless

### LINKED ISSUE #3001: Support of OutlierDetection in DestinationRule for Grpc-Proxyless
istio grpc-proxyless support for outlierDetection

Currently istio supports outlierDetection feature for only envoy proxy and being one of the performant features of istio which we to use in our production setup, but this feature is not supported for grpc-xds proxyless setup.

Grpc-xds supports only 2 algorithms [xds-grfc](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md)

Failure Percent
Success Rate
Envoy's [outlierDetection](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier#outlier-detection) also the algorithm support but istio does not provide any way to configure these. In order to unblock we tried to use EnvoyFilter to configure those algorithm parameters. We were able to do this in presence of sidecar Envoy proxy as istio does provide the patching of EnvoyFilterConfig over the [clusterConfig](https://github.com/istio/istio/blob/141104caec71593378094d789a47e1da7979b57b/pilot/pkg/networking/core/v1alpha3/cluster_builder.go), but this patching was not provider for [grpc-proxyless](https://github.com/istio/istio/blob/141104caec71593378094d789a47e1da7979b57b/pilot/pkg/networking/grpcgen/cds.go)

hence a better solution would be if istio supports OutlierDetection for grpc-proxyless where we wont need to patch envoyFilter config as a workaround and configure outlierDetection from DestinationRule itself.

Describe alternatives you've considered
As an alternative we tried to use EnvoyFilter which patched the its config along the clusterConfig generated after applying destinationRule.
discussion mention here : https://github.com/istio/istio/issues/47927

[x] Docs
[ ] Installation
[x] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[x] User Experience

**Additional context**

aditya-32:
Closes #3001

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -963,12 +963,10 @@ message OutlierDetection {
   // applicable in k8s environments with few pods per service.
   int32 min_health_percent = 5;
 
-  // FailurePercent Algorithm for Grpc-xds proxyless for deciding
-  // if the host is an outlier or not.
+  // FailurePercent Algorithm for deciding if the host is an outlier or not.
   FailurePercentageEjection failure_percentage_ejection = 10;
 
-  // SuccessRateEjection Algorithm for Grpc-xds proxyless for deciding
-  // if the host is an outlier or not.
+  // SuccessRateEjection Algorithm for deciding if the host is an outlier or not.
   SuccessRateEjection success_rate_ejection = 11;
 }
 
@@ -992,8 +990,8 @@ message FailurePercentageEjection {
   // ejection will not be performed.
   google.protobuf.UInt32Value minimum_hosts = 3;
 
-  // The minimum number of total requests that must be collected in one interval (as defined by the
-  // interval duration above) to perform failure percentage-based ejection for this address. If the
+  // The minimum number of total requests that must be collected in one interval
+  // to perform failure percentage-based ejection for this address. If the
   // volume is lower than this setting, failure percentage-based ejection will not be performed for
   // this host.
   google.protobuf.UInt32Value request_volume = 4;
@@ -1025,10 +1023,9 @@ message SuccessRateEjection {
   google.protobuf.UInt32Value minimum_hosts = 3;
 
   // The minimum number of total requests that must be collected in one
-  // interval (as defined by the interval duration above) to include this address
-  // in success rate based outlier detection. If the volume is lower than this
-  // setting, outlier detection via success rate statistics is not performed
-  // for that address.
+  // interval to include this address in success rate based outlier detection.
+  // If the volume is lower than this setting, outlier detection via success
+  // rate statistics is not performed for that address.
   google.protobuf.UInt32Value request_volume = 4;
 }
 
@@ -1040,7 +1037,7 @@ message SuccessRateEjection {
 // for connections to upstream database cluster.
 //
 // {{<tabset category-name="example">}}
-// {{<tab name="v1alpha3" category-value="v1alpha3">}}
+// {{<tab name="v1alpha3" category-value="v1lpha3">}}
 // ```yaml
 // apiVersion: networking.istio.io/v1alpha3
 // kind: DestinationRule
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -542,8 +542,8 @@ spec:
                               nullable: true
                               type: integer
                             failurePercentageEjection:
-                              description: FailurePercent Algorithm for Grpc-xds proxyless
-                                for deciding if the host is an outlier or not.
+                              description: FailurePercent Algorithm for deciding if
+                                the host is an outlier or not.
                               properties:
                                 enforcementPercentage:
                                   description: The % chance that an address will be
@@ -558,9 +558,8 @@ spec:
                                   type: integer
                                 requestVolume:
                                   description: The minimum number of total requests
-                                    that must be collected in one interval (as defined
-                                    by the interval duration above) to perform failure
-                                    percentage-based ejection for this address.
+                                    that must be collected in one interval to perform
+                                    failure percentage-based ejection for this address.
                                   nullable: true
                                   type: integer
                                 threshold:
@@ -588,9 +587,8 @@ spec:
                                 origin failures from external errors.
                               type: boolean
                             successRateEjection:
-                              description: SuccessRateEjection Algorithm for Grpc-xds
-                                proxyless for deciding if the host is an outlier or
-                                not.
+                              description: SuccessRateEjection Algorithm for deciding
+                                if the host is an outlier or not.
                               properties:
                                 enforcementPercentage:
                                   description: The % chance that an address will be
@@ -605,9 +603,8 @@ spec:
                                   type: integer
                                 requestVolume:
                                   description: The minimum number of total requests
-                                    that must be collected in one interval (as defined
-                                    by the interval duration above) to include this
-                                    address in success rate based outlier detection.
+                                    that must be collected in one interval to include
+                                    this address in success rate based outlier detection.
                                   nullable: true
                                   type: integer
                                 stdevFactor:
@@ -889,9 +886,8 @@ spec:
                                     nullable: true
                                     type: integer
                                   failurePercentageEjection:
-                                    description: FailurePercent Algorithm for Grpc-xds
-                                      proxyless for deciding if the host is an outlier
-                                      or not.
+                                    description: FailurePercent Algorithm for deciding
+                                      if the host is an outlier or not.
                                     properties:
                                       enforcementPercentage:
                                         description: The % chance that an address
@@ -908,9 +904,8 @@ spec:
                                         type: integer
                                       requestVolume:
                                         description: The minimum number of total requests
-                                          that must be collected in one interval (as
-                                          defined by the interval duration above)
-                                          to perform failure percentage-based ejection
+                                          that must be collected in one interval to
+                                          perform failure percentage-based ejection
                                           for this address.
                                         nullable: true
                                         type: integer
@@ -943,8 +938,7 @@ spec:
                                     type: boolean
                                   successRateEjection:
                                     description: SuccessRateEjection Algorithm for
-                                      Grpc-xds proxyless for deciding if the host
-                                      is an outlier or not.
+                                      deciding if the host is an outlier or not.
                                     properties:
                                       enforcementPercentage:
                                         description: The % chance that an address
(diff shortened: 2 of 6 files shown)


networking/v1alpha3/destination_rule.proto
```diff
@@ -946,6 +946,74 @@ message OutlierDetection {
...
   // applicable in k8s environments with few pods per service.
   int32 min_health_percent = 5;
+
+  // FailurePercent Algorithm for Grpc-xds proxyless for deciding
+  // if the host is an outlier or not.
+  FailurePercentageEjection failure_percentage_ejection = 10;
+
+  // SuccessRateEjection Algorithm for Grpc-xds proxyless for deciding
+  // if the host is an outlier or not.
+  SuccessRateEjection success_rate_ejection = 11;
```
hzxuzhonghu, line 956: IIUC this is only for grpc not sidecar. 
How does it collaborate with the existing configs like `Consecutive_5Xx`? Does it have a limitation only some of these can be set.

networking/v1alpha3/destination_rule.proto
```diff
@@ -946,6 +946,74 @@ message OutlierDetection {
...
+  // The minimum number of addresses in order to perform failure percentage-based ejection.
+  // If the total number of addresses is less than this value, failure percentage-based
+  // ejection will not be performed.
+  google.protobuf.UInt32Value minimum_hosts = 3;
+
+  // The minimum number of total requests that must be collected in one interval (as defined by the
+  // interval duration above) to perform failure percentage-based ejection for this address. If the
+  // volume is lower than this setting, failure percentage-based ejection will not be performed for
+  // this host.
+  google.protobuf.UInt32Value request_volume = 4;
```
hzxuzhonghu, line 983: Not sure if it is a sliding window or not

networking/v1alpha3/destination_rule.proto
aditya-32: currently istio provides outlierDetection which is supported by envoy proxy only. For services using grpc with xds as proxyless dont have support for this. Grpc proxyless only supports 2 outlierDetection Algorithm ( [gRFC](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md) ). And currently we don't have any way to configure these properties (neither through DestinationRule nor EnvoyFilter). 

**Does it have a limitation only some of these can be set?**
regarding this , in order to trigger outlierDetection we need to set properties if either of these algorithm


reference for previous discussion:

-  [istio/istio/47927](https://github.com/istio/istio/issues/47927)
- [grpc-java/issue/10656](https://github.com/grpc/grpc-java/issues/10656)

networking/v1alpha3/destination_rule.proto
aditya-32: the request for the host should be atleast **_request_volume_** in the **_interval_** period in order to be eligible for outlierDetection

networking/v1alpha3/destination_rule.proto
```diff
@@ -946,6 +946,74 @@ message OutlierDetection {
...
+message FailurePercentageEjection {
+  // The failure percentage to use when determining failure percentage-based outlier detection. If
+  // the failure percentage of a given address is greater than or equal to this value, it will be
+  // ejected.
+  google.protobuf.UInt32Value threshold = 1;
+
+  // The % chance that an address will be actually ejected when an outlier status is detected through
+  // failure percentage statistics. This setting can be used to disable ejection or to ramp it up
+  // slowly.
+  google.protobuf.UInt32Value enforcement_percentage = 2;
```
howardjohn, line 972: Why do we need this? I am struggling to see the use case

howardjohn (changes requested):

networking/v1alpha3/destination_rule.proto
aditya-32: envoy has those parameters [envoy outlier detection](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier). We need to pass this information from istio to envoy in order configure and  trigger outlierDetection . And we are using grpc-proxyless (no envoy in sidecar) and currently istio does not support this for grpc-proxyless , and even with envoy we need to EnvoyFilter to set those parameters.

howardjohn: Ok but they do apply to envoy as well, right? if so, we shouldn't say its "for Grpc-xds proxyless", as its not. Its for any proxy type

networking/v1alpha3/destination_rule.proto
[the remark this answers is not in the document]
aditya-32: @howardjohn make sense will check and update regarding the fields priority and usecase!

networking/v1alpha3/destination_rule.proto
aditya-32: **_enforcement_percentage_** this is like what is the probability for a host to be removed from LB pool when detected as outlier.

networking/v1alpha3/destination_rule.proto
aditya-32: @howardjohn I hope i'm able to convey the use case along with the above mentioned previous discussions , As there is a gap in implementation for outlierDetection in envoy and grpx-proxyless. Since we are using grpc proxyless and use outlierDetection in it.

networking/v1alpha3/destination_rule.proto
howardjohn: I know _what_ it is, but I am not clear _why_ we need to offer it or why a user would want to set it. "It exists in the Envoy API" is not a reason why, either -- Envoy's API has different design principals than Istio

networking/v1alpha3/destination_rule.proto
aditya-32: envoy does support it but without envoy how can we configure and trigger outlierDetection for grpc based application ?
And even if envoy does support how can we configure it ?

hzxuzhonghu: I cannot get the answer, can you reply positively

aditya-32: @howardjohn envoy proto for outlierDetection has those parameters for FailurePercent and SuccessRate algorithm, but currently we don't have a way to configure it from istio. Only way to configure those for application using envoy-proxy is through EnvoyFilter. But our use case is for grpc proxyless application(no envoy sidecar proxy) , and istio does not support a way to configure those Algorithm parameters. Also currently there is no way to configure outlierDetection for grpc-proxyless application.

networking/v1alpha3/destination_rule.proto
anuragagarwal561994: @howardjohn I think I understand the use-case here, lets say that there are 500 servers and 50 of them started showing ejection behaviour. Then the outlier detection algorithm will eject all these 50 servers and distribute its load on rest of the 450 servers, this might be a big traffic shift for other servers to handle.

On the contrary, having this setting to lets say 50 will eliminate 25 and then 12 and so on, such that the traffic is ramped up in corresponding intervals.

networking/v1alpha3/destination_rule.proto
```diff
@@ -946,6 +946,74 @@ message OutlierDetection {
...
+
+  // The minimum number of addresses in order to perform failure percentage-based ejection.
+  // If the total number of addresses is less than this value, failure percentage-based
+  // ejection will not be performed.
+  google.protobuf.UInt32Value minimum_hosts = 3;
```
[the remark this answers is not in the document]
anuragagarwal561994: @howardjohn for this and the remaining fields, I am a little confused when we say in the context of istio, I would like to better understand how do we decide which properties do we need to translate to data plane like envoy / grpc-xds, because istio is itself might not be doing any action here excpet for providing the configurations and if a user would like to configure these things for their setup, how would one ensure that they are made available to them via the destination rule otherwise?

I would say these are these parameters are more or less tuning parameters and may not be left as well left to default values (which is 5), because they would otherwise not work for the cases where the service is having less than 5 hosts. They are better tuned as per the traffic patterns and the use cases of the user.

I would have given better answer if I would have used it on staging / production setup and tried and tested it, but to test it we would as well be needing these features in istio.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: Yes, my suggestion is to make them explicit, what does each option control and how will they collaborate.

aditya-32: @hzxuzhonghu can you please re-review, made the changes

networking/v1alpha3/destination_rule.proto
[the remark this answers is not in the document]
costinm: Normally users try it with EnvoyFilter - adding it to a beta API just to test it out and figure out that only 1 or 2 users understand how to use it - but we have to maintain it for a long time is not ideal. 

I know John has a different opinion on new CRDs - but given that Ambient will have particular troubles with DestinationRule ( per-node ztunnel won't apply them - and waypoints can't respect the full semantics ) - this sounds like something that would be much better off in a separate CRD, like CircuitBreaking, with the new attachment model.

DR is in particular tricky because it can be used both client side and server side. It is unreasonable to expect clients to know how many servers are running ( and all of them to update when this change). With auto-scaling - I doubt even the producer can know this. Some of those settings would be far better left to Istiod to compute.

networking/v1alpha3/destination_rule.proto
```diff
@@ -972,7 +1037,7 @@ message OutlierDetection {
 // for connections to upstream database cluster.
 //
 // {{<tabset category-name="example">}}
-// {{<tab name="v1alpha3" category-value="v1alpha3">}}
+// {{<tab name="v1alpha3" category-value="v1lpha3">}}
```
costinm, line 1040: I assume this is accidental.

costinm (commented):
At the very least this will need a LOT more docs and examples. Adding an API that even Istio reviewers have trouble understanding leaves little chance to regular users. 

I don't disagree that some of it may be useful - but also not clear if this is something the user should configure or Istiod should set, knowing the current number of endpoints. 

Another concern is that I don't understand how this relates to traffic shifting and versioning - if I deploy v2 and all the new endpoints are outliers, how would the setting affect this ? 

In any case - at the very least the new fields should be annotated with whatever 'experimental' we use to indicate they are NOT part of the beta API and are very experimental. Also a LOT of docs on how this will work with ambient ( we still share many APIs, and for new stuff we should document if it applies or not, where it applies and how the workload selector vs other composition rules are changed)

akamac:
I'd love to see it implemented as we are adopting proxyless setup.

linsun:
Not sure if @aditya-32 still working on it?  If not, should we close this PR? No activity for 10+ months.

martinbaillie:
I would like to be able to use these algorithms in my Envoy-fronted and
proxyless pods but reading the comments I would like to know if continuing work
on this likely to be accepted by the project? Or would it be best to manage this
configuration on the side?

If the former, which way would we like it to go? In the comments i'm seeing:
a. reduced exposure in new `DestinationRule` version.
b. new `CircuitBreaker` CRD.

PR #3002, opened 2023-11-27
OpenTelemetryTracingProvider: New configurations for resource detectors

joaopgrassi:
This PR is an alternative to #2998 to showcase "all" new MeshConfig properties we would need to have to support all the new OpenTelemetry tracing features in Envoy. I'm creating this so the Istio community can see it and maybe this influences the discussions. 

- HTTP Exporter
- Custom Sampler
- Resource Detectors

All of those features are defined by the OpenTelemetry project and were added to Envoy as a mechanism to enhance the traces produced by it. More specially, because with the eminent OpenTracing deprecation there, some observability back-ends will be affected and by improving the OTel support in Envoy users can switch to that. 

**EDIT**

I modified this PR to only include the [resource detectors](https://opentelemetry.io/docs/languages/js/resources/). Sampling will be added in a separated PR to avoid bloating. 

Closes istio/istio#48885


mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1026,53 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. Specifies the configuration for exporting OTLP traces via HTTP.
+      // When empty, traces will be exported via gRPC.
+      HttpService otlp_http = 4;
+
+       // Optional. Specifies a list of resource detectors to be used by the OpenTelemetry Tracer.
+       repeated google.protobuf.Any resource_detectors = 5;
```
hzxuzhonghu, line 1035: no sure what does it mean

zirain: why use any here?

joaopgrassi: I used any because I'm not sure how we can do it with the actual types, since they are different. For ex, here you can see the one for the environment https://github.com/istio/api/pull/3002/files#diff-b5c7ece76d5d50226c89f3a2b7a70b22a91f3f8fc0bf079dc0e7fcd2ecc98938R1044. Below there's one for Dynatrace. There can be many, for ex for AWS, Azure etc. They can have their own config options so we still need a type, even though most will be empty.

I'm all open to better approaches

howardjohn: I don't understand why we have both an Any _and the concrete messages_ (EnvironmentResourceDetector, DynatraceResourceDetector, etc)?

Seems like we would have one or the other...?

mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1026,53 @@ message MeshConfig {
...
+
+      // Optional. Specifies the sampler to be used by the OpenTelemetry Tracer
+      // When specified, it overrides the sampler configuration from Envoy.
+      oneof sampler {
+        AlwaysOnSampler always_on = 6;
```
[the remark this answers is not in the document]
samohte: > Do we need this?

There might be additional samplers (e.g. rate based, url based) for the OpenTelemetry tracer in the future. Therefore, it makes sense to allow to configure a sampler.

> What is the difference between setting this, and not setting any sampler?

It depends on the envoy telemetry setting https://istio.io/latest/docs/reference/config/telemetry/. These settings have an impact on every configured sampler. The AlwaysOnSampler can only be used with the OpenTelemetry tracer and can override the global decision.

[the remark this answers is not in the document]
howardjohn: I feel like sampling config here conflicts with the sampling config here (https://istio.io/latest/docs/reference/config/telemetry/#Tracing), no? And that replaced meshconfig.Tracing.Sample which replaced PILOT_TRACE_SAMPLING.

I am worried about so many different, conflicting, ways to configure the same thing.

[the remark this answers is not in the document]
joaopgrassi: The problem is that that sampling configuration is only a simple percentage AFAIK. Usually vendors offer more complex sampling decisions that allow for more control on the data (e.g. dynamic sampling). The OTel project also defines different sampling algorithms which is not possible to be covered today with the existing config. 

There is already two tracers in Envoy (AWS and Datadog) that do not use this sampling configuration and use their own algorithm/logic. For ex, AWS reads from a .json file. 

I get you concern, which is another counter argument for us to introduce a Dynatrace tracing config to Istio, which would hide all of this from the API surface in istio. This is the same as done for others. But if we want to make it generic and use OpenTelemetry, then we will need to have this extra config. I don't think it makes sense to change the config in [istio.io/latest/docs/reference/config/telemetry/#Tracing](https://istio.io/latest/docs/reference/config/telemetry/#Tracing) to use the new samplers, as those are not known by Envoy and only belongs to the tracer.

mesh/v1alpha1/config.proto
joaopgrassi: You can configure N resource detectors in Envoy. When this config is used by Istio, we need to know which type the user is using, so we can properly populate the `TypedConfig` in Envoy. You can see how this would work in the draft code I have

https://github.com/dynatrace-oss-contrib/istio/commit/190e0b0fe34726e3b6dbd62ed6211870afb431ae#diff-688eb642c95839e3154e407aee02f61329ae3bef4c74bc4c19290cacecbb4e2f

If we just use any, how will I know when transforming the Istio config to Envoy which resource detectors I need to use? I thought about using a simple enum, but some resource detectors may also have their own config (e.g. path of a file to load the attributes), so all in all having the concrete messages help in both things.

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
howardjohn: Hmm, interesting. All good points. I generally agree we probably cannot have the full range of sampling options in Telemetry inline.

Maybe we have something like

MeshConfig:
```
sampling = oneof {
  percentage {
    default_percentage 
  }
  some_smart_sampler {
  }
}
```

Telemetry remains the same

Logic is:
* If `sampling` is `percentage`, we have the following order of precedence: Telemetry randomSamplingPercentage > `provider.percentage.default_percentage` (maybe we can remove this, seems excessive) > proxyConfig.tracing.sampling > PILOT_TRACE_SAMPLING
* If its another sampler, use that logic. Ignore  Telemetry randomSamplingPercentage. If the custom sampler also has a percentage involved, maybe it could impact it (e.g. a sampler that is like "Random percent but ignore things on weekends" or something... not sure anything actually falls in this category).
* If no sampler set, default is `percentage{}` without a `default_percentage` set (i.e. current behavior)

WDYT?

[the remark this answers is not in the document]
joaopgrassi: I think that makes sense! One question I have is with the second bullet point, maybe you can clarify:

> If its another sampler, use that logic. Ignore Telemetry randomSamplingPercentage

In Envoy, for a trace to be initiated, the following must be true:

> By an external client via the [x-client-trace-id](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-client-trace-id) header.

> By an internal service via the [x-envoy-force-trace](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-envoy-force-trace) header.

> Randomly sampled via the [random_sampling](https://www.envoyproxy.io/docs/envoy/v1.28.0/configuration/http/http_conn_man/runtime#config-http-conn-man-runtime-random-sampling) runtime setting.

Let's ignore the first two ones - That means, for Envoy to start a trace, the `random_sampling` must be configured, and in a case of a custom "smart sampler" is configured, the random sampling must be 100%, so all spans are passed to the custom sampler so it can do its own decision. 

In the Envoy docs, it says `random_sampling` defaults to 100. So when you say: " Ignore Telemetry randomSamplingPercentage", would Istio set this field at all? If it would not set, then we are good with your suggestion. But if it would set to `0` then we would have a problem, as Envoy then would not start a trace.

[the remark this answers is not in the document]
joaopgrassi: You can see how this is the case, for ex Datadog instructs their customers to install Istio with the Pilot sampling to 100% https://docs.datadoghq.com/tracing/trace_collection/proxy_setup/?tab=istio#istio-configuration-and-installation - same deal, so their tracer in Envoy can perform the decision on all spans.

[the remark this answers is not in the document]
howardjohn: Thanks, really useful.

Do all samplers make sense with a percentage sampler in front? Like if you set `randomSamplingPercentage=50` and FooBarSampler, should we do both or ignore randomSamplingPercentage (set to 100%)?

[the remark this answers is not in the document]
joaopgrassi: I think the most correct is to ignore `randomSamplingPercentage` (so, make it 100%) so all spans go to the FooBarSampler where it will do its own logic. It could be again a percentage base, but it needs all spans to do it. If we mix them together then the sampling will be wrong, as it would "percent twice".

joaopgrassi:
Note: I have a smaller PR that just introduces the HTTP exporter #2998. This was just to show the "full" amount of new configuration we will need to support the new OTel tracing features in Envoy

zirain:
https://github.com/istio/api/pull/2998 merged, can we close this? cc @joaopgrassi

joaopgrassi:
@zirain #2998 added only the HTTP exporter configuration. We still need new configuration for the resource detectors part and samplers. I will rebase this and remove the http exporter, as that is already in, but the rest here still apply.

mesh/v1alpha1/config.proto
```diff
@@ -1025,6 +1026,53 @@ message MeshConfig {
       // Optional. Controls the overall path length allowed in a reported span.
       // NOTE: currently only controls max length of the path tag.
       uint32 max_tag_length = 3;
+
+      // Optional. Specifies the configuration for exporting OTLP traces via HTTP.
+      // When empty, traces will be exported via gRPC.
+      HttpService otlp_http = 4;
```
howardjohn, line 1032: nit: can you rebase?

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
howardjohn: Should we:

1. Remove AlwaysOn; if users want this they should set randomSamplingPercentage=100%. no need to have many ways to say "always on"
2. Document the behavior of these interaction with other sampling percentage options (i.e. percentage is ignored/set to 100%)
3

mesh/v1alpha1/config.proto
howardjohn: I guess I don't understand why for one we use concrete types and the other Any. Seems like we should be consistent in both area, unless there is a compelling reason to do them differently?

costinm:
Can you add a small comment with the OTLP env variables that set the
equivalent percentage ? Just for reference.

mesh/v1alpha1/config.proto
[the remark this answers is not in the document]
joaopgrassi: Let's do this: We are still working on a Sampler for Dynatrace in Envoy, so I will focus only on the resource detector part in this PR. My feeling is that we generally agree on the approach and just need to iron out details. Let's do that then in a separate PR.

joaopgrassi:
Hi all, I have rebased this PR on latest master, given #2998 was merged. I think we have some common understanding/agreement on how to go forward so I also removed the sampling parts of the PR. I will focus this on resource detectors and send a new PR for sampling once we have the code merged in Envoy for the Dynatrace sampler.

CC @howardjohn @zirain

mesh/v1alpha1/config.proto
joaopgrassi: I will try to explain. 

In OTel the user can configure `N` number of resource detectors. Those are going to load resource attributes on their own. For example, if you have the Dynatrace k8s operator deployed in your cluster, your telemetry can be enriched with things like k8s namespace, deployment name, host name etc. Dynatrace provides the resource detector package that _knows_ how to load these things.  

The same way, the user maybe want to add some attributes about their service via environment variable, so again, another resource detector is added that knows how to load things from the environment. In the end, both r[esources are merged](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#merge). 

Now, when a Istio user wants to have both Dynatrace+Env resource detectors, they will configure Istio like this:

```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  meshConfig:
    extensionProviders:
    - name: otel-tracing
      opentelemetry:
        port: 4317
        service: opentelemetry-collector.otel-collector.svc.cluster.local
        resource_detectors: # new config
        - dynatrace
        - environment
```

Which in the end, will configure the [resource_detectors](https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/config/trace/v3/opentelemetry.proto) in Envoy. Today in Envoy there are two resource detectors, [Dynatrace](https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/extensions/tracers/opentelemetry/resource_detectors/v3/dynatrace_resource_detector.proto#extension-envoy-tracers-opentelemetry-resource-detectors-dynatrace) and [Environment](https://www.envoyproxy.io/docs/envoy/v1.29.0/api-v3/extensions/tracers/opentelemetry/resource_detectors/v3/environment_resource_detector.proto#extension-envoy-tracers-opentelemetry-resource-detectors-environment).

Now, about the `Any` discussion: Since the user can add `N` number of resource detectors, how can we have a configuration that allows us to know later, when configuring Envoy which detector the user wanted? For example, when configuring Envoy, we need to do:

```
TypedConfig: protoconv.MessageToAny(&resourcedetectors.EnvironmentResourceDetectorConfig{})
```

How do I know what to pass inside `MessageToAny`? 

 So far I thought about:

1. Enum with fixed values
2. Simple string that has to match with some internal "mapping" (similar as a Enum..)
3. A strongly-typed value.

To me, option 3 is better, because:

- No mistake in typing a wrong name, we can easily `switch on TypeUrl`
- Resource detectors may have their own config, for ex, a path to where the attributes are, some URL to make a request, whatever. Having the types then is the only option. 

Then the "top-level" list property `resource_detectors` needs to be `any` so it can receive any type of resource detector message. 

Let me know if you have another idea, or if things are still unclear.

releasenotes/notes/otel-tracing-provider-resource-detectors.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: telemetry
+issue:
+  - https://github.com/istio/api/pull/2998
```
zirain, line 5: remove

mesh/v1alpha1/config.proto
zirain: What about:
```yaml
resource_detectors:
- dynatrace: {}
- environment: {}
- antoherDetector:
    detectorConfiguration:
    ...
```

joaopgrassi: But in the actual Istio code (`tracing.go`), how will we know the actual type I need to configure in Envoy? https://github.com/dynatrace-oss-contrib/istio/commit/190e0b0fe34726e3b6dbd62ed6211870afb431ae#diff-688eb642c95839e3154e407aee02f61329ae3bef4c74bc4c19290cacecbb4e2fR336-R352 

If it's just all `any`, how will we map properties from Istio > Envoy?

zirain:
```yaml
dynatrace -> envoy.tracers.opentelemetry.resource_detectors.dynatrace
environment -> envoy.tracers.opentelemetry.resource_detectors.environment
```
IMO, use `any` is fine to develop, but not a good UX to end users.

zirain: `'@type': type.googleapis.com/istio.mesh.v1alpha1.MeshConfig.ExtensionProvider.OpenTelemetryTracingProvider.DynatraceResourceDetector` looks terrible for API(except for EnvoyFilter)

joaopgrassi: Sure, I agree. But I'm not sure what other approach we can use? 

```
resource_detectors:
- dynatrace: {}
- environment: {}
- antoherDetector:
    detectorConfiguration:
```
In this example you gave, I assume those are all `any`? Since inheritance isn't a thing in protobuf, the `repeated` field as to be any, I don't see it another way. 

Then the argument how/what we put in the array, comes down to how they are later identified when configuring Envoy. If they themselves are also `any` we would still need the types so we can unmarshal them and access their properties (if they have any).

joaopgrassi: OK I tried something different, maybe it's what you @zirain  and @howardjohn were thinking? Please take a look at: [6309653](https://github.com/istio/api/pull/3002/commits/630965361a6acc552effa0bb7b0ee1b7260f1f88)

It would be used as this:

```
- name: otel-tracing
  opentelemetry:
    port: 443
    service: my.olly-backend.com
    resource_detectors:
      environment: {}
      dynatrace: {}
      other:
        prop: "value"
```

No more `any`.

releasenotes/notes/otel-tracing-provider-resource-detectors.yaml
joaopgrassi: I linked to a real issue that some other user opened: https://github.com/istio/istio/issues/48885

mesh/v1alpha1/config.proto
```diff
@@ -1096,6 +1096,23 @@ message MeshConfig {
...
+      //
+      // ```yaml
+      // - name: otel-tracing
+      //   opentelemetry:
+      //     port: 443
+      //     service: my.olly-backend.com
+      //     resource_detectors:
+      //       environment: {}
+      // ```
+      ResourceDetectors resource_detectors = 5;
```
zirain, line 1115: this LGTM, just want to make sure is threre any case need two same destector?

joaopgrassi: AFAIK the OTel specification doesn't say anything about it ([spec](https://opentelemetry.io/docs/specs/otel/resource/sdk/)) but from my experience, I don't think it would make sense to add the same detector twice. They would just be executed twice, one overwriting the other.

zirain: Thanks for clarification

zirain (approved):
LGTM, thanks!

howardjohn (approved):

linsun:
Per @zirain's req, we discussed the request to cherry pick this to Istio 1.21 branch in today's istio WG meeting. The general consensus is all api changes will require TOC approval to cherry pick if this misses the release branch cutting. Given there is still impl PR open on this and we don't see a strong reason to make an exception for cherry pick this API change, we prefer to not cherry pick and wait till it lands in istio 1.22.

linsun:
cc @istio/release-managers-1-21 ^^^

PR #3003, opened 2023-11-27
TracingProviders: Add Dynatrace trace provider

joaopgrassi:
In [#47453](https://github.com/istio/istio/issues/47453) I proposed to add a new tracing provider in Istio for Dynatrace. The idea was to add a provider for Dynatrace that "wraps" the OpenTelemetry one in Istio, offering easy of configuration to our customers. 

The community opinion was that this was not a good idea as there's a desire to decrease vendor code in MeshConfig. It was advised to simply instruct users to use the OTel provider already available today. 

I opened now initial PRs to extend the OTel provider in Istio to support the new Envoy features #3002. The goal of this PR is to offer a comparison to the community of how much config going with a Dynatrace extension would be compared to extending the OpenTelemetry one. Mainly for the concern of adding more things to MeshConfig. This route is much simpler, with the caveat that it's custom tailored to Dynatrace users, whereas the one in #3002 can benefit everyone.


hzxuzhonghu:
Instead of adding more and more providers, it seems OTEL is the generic way to do

zirain:
https://github.com/istio/api/pull/2998 merged, can we close this? cc @joaopgrassi

joaopgrassi:
@zirain yes I believe the direction is to add the opentelemetry configuration and not vendor specific. I will close this. Thanks!

PR #3007, opened 2023-11-28
Added upstream ProxyProtocol settings

hzxuzhonghu:
It is to add support for upstream proxy protocol. As a supersede of envoyfilter patching  https://github.com/istio/api/pull/2988.

Previously, we have supported downstream proxy protocol. ` "proxy.istio.io/config": '{"gatewayTopology" : { "proxyProtocol": {} }}'`

It is especially useful to be able to initiate ProxyProtocol to external service, to preserve the original source address.


hzxuzhonghu:
cc @linsun @howardjohn

networking/v1alpha3/destination_rule.proto
```diff
@@ -365,6 +365,22 @@ message TrafficPolicy {
...
   // Tunnel settings can be applied to TCP or TLS routes and can't be applied to HTTP routes.
   TunnelSettings tunnel = 6;
+
+  message ProxyProtocol {
+    enum VERSION {
+      // ⁣PROXY protocol version 1. Human readable format.
+      V1 = 0;
+
+      // ⁣PROXY protocol version 2. Binary format.
+      V2 = 1 [deprecated=true];
```
howardjohn, line 375: why is v2 deprecated?

hzxuzhonghu: I copied from other enum, and forgot to update

hzxuzhonghu: fixed now

networking/v1alpha3/destination_rule.proto
```diff
@@ -365,6 +365,22 @@ message TrafficPolicy {
...
+
+      // ⁣PROXY protocol version 2. Binary format.
+      V2 = 1;
+    };
+    // The PROXY protocol version to use. See https://www.haproxy.org/download/2.1/doc/proxy-protocol.txt for details.
+    // By default it is `V1`.
+    VERSION version = 1;
+  }
+
+  // The upstream PROXY protocol settings.
```
howardjohn, line 382: what is the semantics with tls?  does this replace TLS, or it's in addition to tls?

I am not a proxy expert but seems plausible that someone would want to use it with tls, as long as it's possible with envoy?

If not, we should add validation to o make sure it's not used with TLS config (and same with tunnel - I assume you cannot do both)

howardjohn (commented):
Overall makes sense. @jewertow @kyessenov @costinm may have opinions

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: I am gonna implement without influencing Existing transportsocket.

As you can see from the patch below, in the ProxyProtocolUpstreamTransport, we can set tls or rawbuffer transport_socket.

```
//     patch:
//       operation: INSERT_FIRST
//       value:
//        name: envoy.transport_sockets.upstream_proxy_protocol
//        match: {}
//        transport_socket:
//          name: envoy.transport_sockets.upstream_proxy_protocol
//          typed_config:
//            '@type': type.googleapis.com/envoy.extensions.transport_sockets.proxy_protocol.v3.ProxyProtocolUpstreamTransport
//            config:
//              version: V2
//            transport_socket:
//              name: "envoy.transport_sockets.raw_buffer"
//              typed_config:
//                '@type': type.googleapis.com/envoy.extensions.transport_sockets.raw_buffer.v3.RawBuffer
```

hzxuzhonghu: For proxy protocol, it requires both client and server side support it. Otherwise the communication will be broken.

If we specify tls transport socket under `ProxyProtocolUpstreamTransport`, it means the traffic will first be tls encripted, and then wrap a proxy protocol header. They donot conflict with each other.

howardjohn (approved):

networking/v1beta1/destination_rule.proto
```diff
@@ -314,6 +314,22 @@ message TrafficPolicy {
...
+
+      // ⁣PROXY protocol version 2. Binary format.
+      V2 = 1;
+    };
+    // The PROXY protocol version to use. See https://www.haproxy.org/download/2.1/doc/proxy-protocol.txt for details.
+    // By default it is `V1`.
+    VERSION version = 1;
+  }
+
+  // The upstream PROXY protocol settings.
```
kyessenov, line 331: How does this interact with TLS and Tunnel settings? Is PROXY inside or outside the other transports? Can we use them together?

zirain (approved):

networking/v1beta1/destination_rule.proto
hzxuzhonghu: @kyessenov see above https://github.com/istio/api/pull/3007#discussion_r1408656348

networking/v1alpha3/destination_rule.proto
kyessenov: OK, can you document that PROXY is outer to TLS? What about HTTP CONNECT tunneling? You can sequence them any way, depending on the tunneling mechanisms.

PR #3018, opened 2023-12-06
Add timeout field to JWTRule

leosarra:
This PR will add a timeout field to allow for per-JWTRule timeout customization.
Link to PR in istio repo is: coming soon


leosarra:
> You can try to use envoy to fetch jwks, it is more recommended.

Timeout is currently fixed regardless of the solver that is used.
I'm changing the istio code to no longer hardcode it for both Istiod and Envoy JWKS resolver via the following PR:
https://github.com/istio/istio/pull/48203

security/v1/jwt.proto
```diff
@@ -178,8 +179,11 @@ message JWTRule {
   // [Experimental] This feature is a experimental feature.
   repeated ClaimToHeader output_claim_to_headers = 11; // [TODO:Update the status whenever this feature is promoted.]
 
+// Sets the maximum duration that an OpenID request can take to arrive upon request. Default is 5s.
+  google.protobuf.Duration timeout = 13;
```
hzxuzhonghu, line 183: THe comment is not quite right. 
Timeout actually is the maximum duration before fetching the jwks

leosarra:
Adding a do-not-merge label until the status of https://github.com/istio/istio/pull/48203 is clarified

security/v1/jwt.proto
leosarra: Updated to "The maximum amount of time that will be spent waiting for the JWKS to be fetched. Default is 5s."

hzxuzhonghu (approved):

security/v1/jwt.proto
linsun: that -> what will be spent waiting?  sidecars? or proxies (if it covers waypoint or gw too)?

linsun (commented):
LGTM left a q on comment

security/v1/jwt.proto
leosarra: I remained pretty vague on the description because the entity that waits is controlled by PILOT_JWT_ENABLE_REMOTE_JWKS.

Assuming PILOT_JWT_ENABLE_REMOTE_JWKS=istiod|true then the controller will act as the resolver, and it will wait for the response.
If PILOT_JWT_ENABLE_REMOTE_JWKS=hybrid|false, proxies will wait if the jwks is internal, otherwise istiod will.
If PILOT_JWT_ENABLE_REMOTE_JWKS=envoy then proxies will wait.

Do you think I should go for:
"The maximum amount of time that istiod/proxies will spend waiting for the JWKS to be fetched. Default is 5s."?

linsun: Thanks! In addition to what you proposed, I think would be helpful to say "as configured per PILOT_JWT_ENABLE_REMOTE_JWKS env var".  Could you also clarify in the comment that this is only effective when PILOT_JWT_ENABLE_REMOTE_JWKS is configured?

howardjohn:
cc @aryan16 @liwenhao0810

security/v1/jwt.proto
leosarra: @linsun I updated the comment to
```
  // The maximum amount of time that the resolver, determined by the PILOT_JWT_ENABLE_REMOTE_JWKS environment variable,
  // will spend waiting for the JWKS to be fetched. Default is 5s.
```
I did not add the part about "it is only effective when PILOT_JWT_ENABLE_REMOTE_JWKS" because there were always be at least one resolver (the default for the env variable is false -> istiod will act as resolver)

PR #3022, opened 2023-12-08
layered future: add policy ancestor status

### LINKED ISSUE #3014: layered future: indicate whether a policy is implemented by the waypoint

Implementations with variance from stock MUST adopt the mechanism used by the Gateway API specification to indicate whether a policy is successfully implemented by a Waypoint. See [PolicyAncestorStatus](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1alpha2.PolicyAncestorStatus) and in particular [PolicyConditionReason](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.PolicyConditionReason). For clarity stock OSS Istio will adopt the same mechanism and use the ACCEPTED state to indicate support for its own APIs.

[ ] Docs
[ ] Installation
[ ] Networking
[X] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience

**Additional context**
[Istio’s Layered Future, Policy and Conformance](https://docs.google.com/document/d/1fg95r3C8Iqyivs0OO3z7diiaL6SkD6Ke9SqxaufAVBQ/edit?usp=sharing)

ilrudie:
adds minimal policy ancestor status

closes #3014


meta/v1alpha1/status.proto
```diff
@@ -44,6 +45,10 @@ message IstioStatus {
   // generation is still in progress.  See https://istio.io/latest/docs/reference/config/config-status/ for more info.
   // +optional
   int64 observed_generation = 3;
+
+  // Status from Ancestors
+  // +optional
+  repeated istio.type.v1beta1.PolicyAncestorStatus ancestors = 4;
```
howardjohn, line 51: IMO this should remain hidden for now, certainly until its implemented. However its still experimental in GW-API afaik, so we don't want to be ahead of the curve there (in the past, we have done similar and they changed their API. oops).

type/v1beta1/policy-ancestor-status.proto
```diff
@@ -0,0 +1,108 @@
...
+  // ControllerName is a domain/path string that indicates the name of the
+	// controller that wrote this status. This corresponds with the
+	// controllerName field on GatewayClass.
+  string controller_name = 2;
+
+  // repeated k8s.io.apimachinery.pkg.apis.meta.v1.Condition conditions = 3;
+  repeated Condition conditions = 3;
+}
+
+message Condition {
```
howardjohn, line 71: IstioCondition is basically identical. @therealmitchconnors do you remember why we have LastProbeTime but metav1.Condition has ObservedGeneration?

meta/v1alpha1/status.proto
ilrudie: I'm not sure what exactly hidden means. Do you mean just skip this implementation entirely for now and don't expose a status or do you mean leverage the allow unknown fields in status to write a status without actually adding the experimental fields to the API?

linsun: Maybe he meant something like `// $hide_from_docs`?

ilrudie: Will take a look at this today

type/v1beta1/policy-ancestor-status.proto
ilrudie: Maybe we could add the field as optional, possibly with hide_from_docs, into IstioCondition instead?

type/v1beta1/policy-ancestor-status.proto
```diff
@@ -0,0 +1,108 @@
...
+	// Port is the network port this Route targets. It can be interpreted
+	// differently based on the type of parent resource.
+	// +optional
+  int32 port = 6;
+}
+
+message PolicyAncestorStatus {
+  // AncestorRef corresponds with a ParentRef in the spec that this
+	// PolicyAncestorStatus struct describes the status of.
+  ParentReference ancestor_ref = 1;
```
hzxuzhonghu, line 60: Not sure what does this mean, can you give an concrete example

type/v1beta1/policy-ancestor-status.proto
```diff
@@ -0,0 +1,108 @@
...
+
+  // ControllerName is a domain/path string that indicates the name of the
+	// controller that wrote this status. This corresponds with the
+	// controllerName field on GatewayClass.
+  string controller_name = 2;
+
+  // repeated k8s.io.apimachinery.pkg.apis.meta.v1.Condition conditions = 3;
+  repeated Condition conditions = 3;
```
hzxuzhonghu, line 68: If the ancestor conditions copied here, does that mean the status will be stored in both ancestor and child resource status?

hzxuzhonghu (commented):
After looking at https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.PolicyAncestorStatus.
I donot think that's a good design, declaritive APIs are originally standalone APIs, if coupled with others, maintaining statuses of other dependencies. It would be very difficult for the controllers.

linsun:
> After looking at https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.PolicyAncestorStatus. I donot think that's a good design, declaritive APIs are originally standalone APIs, if coupled with others, maintaining statuses of other dependencies. It would be very difficult for the controllers.

cc @louiscryan to weigh in as he has been leading the design.  FYI @howardjohn @keithmattix

ilrudie:
We can re-open if/when needed

PR #3026, opened 2023-12-14
Update go.mod

jacob-delgado:


go.mod
```diff
@@ -1,32 +1,34 @@
 module istio.io/api
 
-go 1.18
+go 1.21
+
+toolchain go1.21.5
```
jacob-delgado, line 5: This is now required by make gen

ericvn: I thought we could remove the `toolchain` line added after the fact and we would be fine.

howardjohn (approved):

go.mod
howardjohn: `go 1.21` first then it will not add it (or remove toolchain after )

ericvn (approved):

PR #3033, opened 2024-01-03
docs: use Kubernetes instead of k8s in the proto field comments

peterj:
For the consistency/readability sake, updating all instances of "k8s" in the comments to "Kubernetes".


howardjohn (approved):
thanks!

hanxiaop (approved):

PR #3037, opened 2024-01-09
mark extraStatTags deprecated

zirain:
I think this's no more need, cc @kyessenov for confirmation.


linsun (commented):
Waiting for @kyessenov to confirm before approval.

Make sense to me, would need to remove associated reference when we remove it in istio/istio.

kyessenov (approved):
Thanks

annotation/annotations.yaml
```diff
@@ -476,7 +476,7 @@ annotations:
   - name: sidecar.istio.io/extraStatTags
     featureStatus: Alpha
     description: An additional list of tags to extract from the in-proxy Istio Wasm telemetry. Each additional tag needs to be present in this list.
```
howardjohn, line 478: nit: include in description what to use instead?

howardjohn (approved):

annotation/annotations.yaml
zirain: I recall I send an PR for that, let me bring it back later.

PR #3043, opened 2024-01-11
update annotations and labels page

zirain:
follow up: https://github.com/istio/common-files/pull/968/files
xref: https://github.com/istio/istio.io/issues/14453


linsun (approved):
Thanks!!

ericvn (approved):

PR #3044, opened 2024-01-12
add inbound traffic policy to mesh config

ramaraochavali:
As an alternative to this flag https://github.com/istio/istio/pull/48718


kfaseela:
+1 We were also using the flag which got removed downstream, and now confused how to proceed.

ramaraochavali:

mesh/v1alpha1/config.proto
```diff
@@ -191,6 +191,23 @@ message MeshConfig {
   // Default mode is `ALLOW_ANY` which means outbound traffic to unknown destinations will be allowed.
   OutboundTrafficPolicy outbound_traffic_policy = 17;
 
+  // Set the default behavior of the sidecar for handling inbound
+  // traffic to the application.  If your application listens on
+  // localhost, you will need to set this to `LOCALHOST`.
```
linsun, line 196: Does this impact ambient? In other words, do we expect ambient users may need this also? I am thinking no..

cc @bleggett

linsun (commented):
LG - added some discussion around ambient impact

mesh/v1alpha1/config.proto
bleggett: No, it shouldn't.

howardjohn: It actually _could_ work with inpod, but could not without inpod. And I don't think we want to make "inpod" a requirement for ambient...

bleggett: It could I guess, but for now I'd rather ambient stick to the non-legacy/current-default behavior of `PILOT_ENABLE_INBOUND_PASSTHROUGH`

linsun: Thanks for the comments, looks like not a concern for ambient atm.

linsun (approved):

ramaraochavali:
@hzxuzhonghu @howardjohn Can you PTAL when you get chance?

mesh/v1alpha1/config.proto
```diff
@@ -191,6 +191,23 @@ message MeshConfig {
...
+      // on Pod IP. This matches the behavior without Istio enabled at all 
+      // allowing proxy to be transparent.
+      PASSTHROUGH = 0;
+      // inbound traffic will be sent to the destinations listening on localhost.
+      LOCALHOST = 1;
+    }
+    Mode mode = 1;
+  }
+
+  InboundTrafficPolicy inbound_traffic_policy = 66;
```
hzxuzhonghu, line 209: nit: comments here

hzxuzhonghu (commented):
LGTM

hzxuzhonghu (approved):

PR #3045, opened 2024-01-12
add fallback support for PrivateKeyProvider

irisdingbj:
Envoy recently add support for  `fallback` in PrivateKeyProvider as in https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#extensions-transport-sockets-tls-v3-privatekeyprovider

Add the related support into Istio.


mesh/v1alpha1/proxy.proto
```diff
@@ -312,6 +312,9 @@ message PrivateKeyProvider {
     // In effect, this value controls the balance between latency and throughput.
     // The duration needs to be set to a value greater than or equal to 1 millisecond.
     google.protobuf.Duration poll_delay = 1;
+    // If the private key provider isn’t available (eg. the required hardware capability doesn’t existed)
+    // Envoy will fallback to the BoringSSL default implementation when the fallback is true.
```
linsun, line 316: Can you specify the default value in the comment?

mesh/v1alpha1/proxy.proto
```diff
@@ -321,6 +324,9 @@ message PrivateKeyProvider {
     // leading to potentially larger CPU usage.
     // The duration needs to be set to a value greater than or equal to 1 millisecond.
     google.protobuf.Duration poll_delay = 1;
+    // If the private key provider isn’t available (eg. the required hardware capability doesn’t existed)
+    // Envoy will fallback to the BoringSSL default implementation when the fallback is true.
```
linsun, line 328: same here - Can you specify the default value in the comment?

linsun: Also any reason why we have 2 places to configure this?

linsun (commented):
Thanks Iris, left some comment, would like to see them addressed before approval.

mesh/v1alpha1/proxy.proto
irisdingbj: fixed the comments, the reason is QAT and CryptoMB are two independent private key providers, see the same for original `poll_delay` fileld

irisdingbj:
> Thanks Iris, left some comment, would like to see them addressed before approval.

Thanks lin for the review! I have updated the PR to address your comments, PTAL, Thanks!

howardjohn (approved):

mesh/v1alpha1/proxy.proto
hzxuzhonghu: So what is the difference? QAT and CryptoMB are all the same except the name in API

irisdingbj: The different name will trigger different hardware capabilities accordingly.

mesh/v1alpha1/proxy.proto
linsun: Thanks!

releasenotes/notes/private-key-provider-fallback.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: security
+releaseNotes:
+  - |
+    **Added** an `fallback` filed for PrivateKeyProvider to support fallback to the BoringSSL default implementation if the private key provider isn’t available.
```
linsun, line 6:
```suggestion
    **Added** an `fallback` field for PrivateKeyProvider to support fallback to the BoringSSL default implementation if the private key provider isn’t available.
```

linsun (approved):

PR #3052, opened 2024-01-17
Support file mounted CRL

kfaseela:
Currently CRL is supported only as an additional argument in the `credentialName` parameter of [ClientTLSSettings](https://github.com/istio/api/blob/master/networking/v1alpha3/destination_rule.proto#L1146) and [ServerTLSSettings](https://github.com/istio/api/blob/master/networking/v1alpha3/gateway.proto#L696). Due to https://github.com/istio/istio/issues/43966 there are some limitation in using this credentialName support in all usecases, and hence would be nice if CRL is supported as a file mounted option as well.


networking/v1alpha3/destination_rule.proto
```diff
@@ -1174,6 +1174,11 @@ message ClientTLSSettings {
   // be `true` by default in a later version where, going forward, it will be
   // enabled by default.
   google.protobuf.BoolValue insecure_skip_verify = 8;
+
+  // OPTIONAL: The path to the file containing certificate revocation list (CRL)
```
ericvn, line 1178: multiple places:
```suggestion
  // OPTIONAL: The path to the file containing the certificate revocation list (CRL)
```

kfaseela: Done

ramaraochavali:
Thanks for adding this. LGTM

hzxuzhonghu:
file mounted certificate is not quite facilitated to use, would we still encourage users to use it?

ramaraochavali:
> file mounted certificate is not quite facilitated to use, would we still encourage users to use it?

Yes. We can not get away from it because lot of enterprises that have tools/processes that generate certs on file locations

hzxuzhonghu:
OK, i have no objections just to confirm the use case.

networking/v1alpha3/destination_rule.proto
```diff
@@ -1174,6 +1174,11 @@ message ClientTLSSettings {
   // be `true` by default in a later version where, going forward, it will be
   // enabled by default.
   google.protobuf.BoolValue insecure_skip_verify = 8;
+
+  // OPTIONAL: The path to the file containing the certificate revocation list (CRL)
+  // to use in verifying a presented server certificate. If
```
linsun, line 1179: Could you add a sentence for if enabled?  I think the behavior is if enabled, the certificates on the revocation list won't be allowed?

linsun: Also, is DR for sidecar here? do we expect users to mount the file to the pod?

networking/v1alpha3/gateway.proto
```diff
@@ -681,6 +681,11 @@ message ServerTLSSettings {
   // client side certificate.
   string ca_certificates = 5;
 
+  // OPTIONAL: The path to the file containing the certificate revocation list (CRL)
+  // to use in verifying a presented client side certificate. If
+  // omitted, the proxy will not verify the certificate against the `crl`.
```
linsun, line 686: do we expect admins to mount the file to the gw pod?

linsun (commented):
Left some clarification qs.

Wondering if this is best in mesh config.  Do we see users doing this differently for gw and apps?  And for apps, i'd expect to be mesh global, agree?

networking/v1alpha3/destination_rule.proto
kfaseela: this works as all other file mounted options in DR, eg: ca_certificates,client_certificates etc (see L1125) .. for gateway as well as sidecar..

networking/v1alpha3/gateway.proto
kfaseela: That is how it works for the existing options.. using userVolumeMount annotation.. https://istio.io/latest/docs/reference/config/annotations/

networking/v1alpha3/destination_rule.proto
kfaseela: will enhance the description to add how it will work if enabled

kfaseela:
> Left some clarification qs.
> 
> Wondering if this is best in mesh config or proxy config. Do we see users doing this differently for gw and apps? And for apps, i'd expect to be mesh global, agree?

CRL is a CA level property, so it would be ideal to have this configuration option at the same level where we allow ca_certificates is getting configured, which is currently at Client/ServerTLSSettings

ramaraochavali:
+1 to be at DR.

networking/v1alpha3/destination_rule.proto
linsun: Thx, SG, let me know when the desp is updated, I'll take another look

linsun:
> > Left some clarification qs.
> > Wondering if this is best in mesh config or proxy config. Do we see users doing this differently for gw and apps? And for apps, i'd expect to be mesh global, agree?
> 
> CRL is a CA level property, so it would be ideal to have this configuration option at the same level where we allow ca_certificates is getting configured, which is currently at Client/ServerTLSSettings

Ok, thanks for the explanation! I can see this being useful globally or per namespace or per source-destination pair, which DR can provide for either scope.

ramaraochavali:
@linsun Any objections to approve this?

kfaseela:
I think Lin has asked to just enhance the description and she will approve after that..I thought of finishing the implementation first and then rebasing this one.

ramaraochavali:
Thank you. Just want to ensure we have alignment to move forward with this

linsun:
@kfaseela @ramaraochavali sounds good - ping me when you want me to take another look.

networking/v1alpha3/destination_rule.proto
kfaseela: Done

linsun (approved):
LGTM thanks Faseela!

PR #3069, opened 2024-01-27
add shortname for authz

zirain:
I'm not sure is this raised before, but this will save a little life for me, `authorizationpolicies` is too long.


security/v1/authorization_policy.proto
```diff
@@ -449,7 +449,7 @@ option go_package="istio.io/api/security/v1";
 // +cue-gen:AuthorizationPolicy:labels:app=istio-pilot,chart=istio,istio=security,heritage=Tiller,release=istio
 // +cue-gen:AuthorizationPolicy:subresource:status
 // +cue-gen:AuthorizationPolicy:scope:Namespaced
-// +cue-gen:AuthorizationPolicy:resource:categories=istio-io,security-istio-io,plural=authorizationpolicies
+// +cue-gen:AuthorizationPolicy:resource:categories=istio-io,security-istio-io,shortNames=authz,plural=authorizationpolicies
```
hanxiaop, line 452: I think this can be `ap` or follow some similar patterns like other security resources.

hzxuzhonghu:
Yes , proposed before. https://github.com/istio/api/pull/1378

I am ok with this

hzxuzhonghu (approved):

linsun (commented):
probably need more TOC member to review. I think we also need a small release note for this.

security/v1/authorization_policy.proto
linsun: agree `ap` may be better

zirain: updated

linsun (commented):
LGTM, wait for other TOC members to review

nrjpoddar (approved):
LGTM

PR #3082, opened 2024-02-09
meshconfig: add compliance policy

kyessenov:
Ref: https://github.com/istio/istio/issues/49081

Change-Id: I43bb205b5b62e557f61852e08b5084b63453ba6e


mesh/v1alpha1/config.proto
```diff
@@ -1308,8 +1308,12 @@ message MeshConfig {
   // For ISTIO_MUTUAL TLS settings, use meshMTLS configuration.
   TLSConfig tls_defaults = 64;
 
+  // If set, applies policy-specific restrictions over all existing TLS settings, including in-mesh mTLS and external TLS.
+  // Valid values: "", "fips".
+  string compliance_policy = 67;
```
howardjohn, line 1313: What happens if you use compliancePolicy=fips with non-fips builds?

mesh/v1alpha1/config.proto
```diff
@@ -1308,8 +1308,12 @@ message MeshConfig {
   // For ISTIO_MUTUAL TLS settings, use meshMTLS configuration.
   TLSConfig tls_defaults = 64;
 
+  // If set, applies policy-specific restrictions over all existing TLS settings, including in-mesh mTLS and external TLS.
```
howardjohn, line 1311: What is the scope? Inbound, outbound, gateway? control plane communication?

mesh/v1alpha1/config.proto
kyessenov: No-op. FIPS requires that both validated modules are used, and that they are configured with compliance policy. If you do only one of those things, it's not compliant.

mesh/v1alpha1/config.proto
kyessenov: All TLS communication.

howardjohn: Just TLS or all crypto as well (JWT, etc)? We should clarify the scope in the docs. Especially if this about compliance, being verbose in the documentation is probably handy

mesh/v1alpha1/config.proto
```diff
@@ -1308,8 +1308,12 @@ message MeshConfig {
...
+  // Valid values: "", "fips".
```
keithmattix, line 1312: What version of FIPS? 140-2? 140-3 (which supersedes, 140-2)?

keithmattix (changes requested):
I think having an easy field to enable configuration that's consistent with FIPS guidelines is a great idea, but I'm concerned that the prescribed values may not be sufficient

mesh/v1alpha1/config.proto
kyessenov: It says "all TLS settings" in the doc. JWT is not in scope since JWT is not needed by default, and you can simply not use JWT features to remain compliant.

mesh/v1alpha1/config.proto
kyessenov: Istio almost entirely relies on this certificate: https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4407, which is 140-2. I left it a string precisely because the standards change, but I can make it explicitly call out that certificate.

howardjohn: _Istio_ doesn't rely on any certificate as it doesn't offer FIPS compliance or even builds with the FIPS crypto modules. Vendors sell Istio that has these properties.

Which to me makes this a bit awkward to have as an Istio configuration rather than a vendor configuration

keithmattix: Yeah I would make the string value more specific than just FIPS; maybe "fips-140-2" and then a comment above associating the string value to the certification 👍🏾

kyessenov: Vendors rely on the ability of Istio to configure FIPS compliance policies since they can substitute the module and satisfy 1 of 2 conditions. Without an API, a small fork is needed which makes it hard to audit compliance.

kyessenov: Applied your suggestion, @keithmattix .

mesh/v1alpha1/config.proto
keithmattix: Is it a no-op or ignored? That is, will the cryptographic algos be filtered even though the build is non-FIPS?

mesh/v1alpha1/config.proto
keithmattix: Yep completely agree. Thanks!@

mesh/v1alpha1/config.proto
kyessenov: It'll do what the comment does: apply the restrictions according to the compliance policy. Whether that is enough  for FIPS depends on whether the build is also using the CMVP TLS crypto module.

So it's not really no-op or ignored - it'll do the same thing regardless of what the build is.

keithmattix: Ack that's what I would expect. Probably worth a big warning sign in the documentation saying that this flag is not sufficient for FIPS

ramaraochavali: And with the current implementation if we set fips-140-2 - is the plan to disable TLS 1.3 and set Max version as 1.2?

kyessenov: @ramaraochavali Yes.

kyessenov:
@keithmattix @pmerrison @ZackButcher 

Please provide feedback. We'd like to restore compliance for all Istio deployments, and this is a necessary step to minimize the work for all vendors.

mesh/v1alpha1/config.proto
keithmattix: If we either:
1. Hide this from docs for now and add more warnings later or
2. Add a warning like Envoy has: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/security/ssl#fips-140-2

Then I'm good to approve

kyessenov:
@keithmattix @howardjohn  Thanks for reviewing. I expanded the explanation.

keithmattix (approved):
LGTM

pmerrison:
Had a chat with Kuat in Slack, and this looks good to me too.  Thanks for doing this Kuat, it saves us all work.

pmerrison (approved):

ramaraochavali:
@kyessenov just to clarify, in future based on FIPS compliance do you plan to configure cipher suites  also?  What else is configurable via this option?

howardjohn (approved):

kyessenov:
> @kyessenov just to clarify, in future based on FIPS compliance do you plan to configure cipher suites also? What else is configurable via this option?

Yes, ciphersuites and curves for TLSv1.2 must be configured for FIPS now and in the future with the updated boringssl.

> I am wondering if we enhance TLSSettings to have MaxVersion and provide guidance in docs on what values need to be set to be FIPSCompliant instead of having an additional API?

That doesn't work because there's more TLS configuration than xDS. gRPC C++, Go SDK, and gRPC Go SDKs are all in scope and don't use this setting.

> * While we have added docs about using FIPS image, I wonder if users might incorrectly think based on this mode, Istio would inject FIPS version of Envoy

I made it clear that this setting is not sufficient. At a minimum, certified builds must be used as well. However, it's necessary.

> * We need document what fips mode does exactly in the current version? Today it is just setting MaxVersion and tomorrow it might change some cipher suites.

We'll make restrictions clear in the implementation PR and put it into change notes. FIPS is evolving, Ed25519 was recently approved, so it's expected that we respond by allowing when appropriate.

howardjohn:
Is there somewhat a bootstrapping problem? To read meshconfig we need to connect to k8s api server (over TLS)

kyessenov:
@howardjohn k8s in FIPS mode should not be negotiating non-FIPS ciphers, it's not our responsibility. We're responsible when the server can choose non-FIPS ciphers, e.g. with GCP or Envoy/sidecars.

keithmattix:
+1 - for any real FIPS solution, k8s has to be FIPS as well otherwise the whole thing is pointless.

howardjohn:
The question is the clients responsibility to not pick a non-compliant
cipher, or the servers, or both?

I think in GCP the server will offer whatever. And that may also apply to
the GKE control plane

kyessenov:
@howardjohn We might need a startup variable to Istiod. This change is needed to control injection and proxy containers, but it's not the only thing that is needed, as I already mentioned. Compliance is complicated, and requires a thorough audit - but we have to provide the flags to make it possible since we don't have any way to set restrictions anywhere.

howardjohn:
If it needs a startup flag then this should probably be a startup flag only, not sure we need mesh config also?

kyessenov:
@howardjohn GCP uses mesh config as a way to provision a custom control plane. I haven't validated that _we need a startup flag_ - we might not need it if we wait to connect to k8s a bit.

kyessenov:
@howardjohn @keithmattix @pmerrison We got some feedback that suggests reverting this change. 1) MeshConfig implies dynamic response, which is not possible for this fields, 2) We use TLS to read MeshConfig which causes a bootstrapping problem. The replacement is startup flags.

PR #3088, opened 2024-02-15
Doc: Add note on DR workloadSelector

NaturezzZ:
Supplementary explanation based on https://github.com/istio/istio/issues/49111#issuecomment-1921674872.
The same field name WorkloadSelector has a different effect/behavior in DR and SE. WorkloadSelector in ServiceEntry indicates destination workloads of traffic, while in DR workloadSelector reveals which workloads the traffic emanating from uses the DR.

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -168,6 +168,12 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// Destination Rules reveals which workloads the traffic emanating from 
+// uses this Destination Rule, unlike the workloadSelector configuration 
+// in [ServiceEntries](https://istio.io/docs/reference/config/networking/service-entry/#ServiceEntry), 
+// which is used to reveal which workloads the traffic is routed to.
+// 
 // {{<tabset category-name="selector-example">}}
 // {{<tab name="v1alpha3" category-value="v1alpha3">}}
 // ```yaml
--- a/networking/v1alpha3/destination_rule.pb.go
+++ b/networking/v1alpha3/destination_rule.pb.go
@@ -167,6 +167,12 @@
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in
+// Destination Rules reveals which workloads the traffic emanating from
+// uses this Destination Rule, unlike the workloadSelector configuration
+// in [ServiceEntries](https://istio.io/docs/reference/config/networking/service-entry/#ServiceEntry),
+// which is used to reveal which workloads the traffic is routed to.
+//
 // {{<tabset category-name="selector-example">}}
 // {{<tab name="v1alpha3" category-value="v1alpha3">}}
 // ```yaml
--- a/networking/v1alpha3/destination_rule.pb.html
+++ b/networking/v1alpha3/destination_rule.pb.html
@@ -134,6 +134,11 @@ spec:
 <p>Destination Rules can be customized to specific workloads as well.
 The following example shows how a destination rule can be applied to a
 specific workload using the workloadSelector configuration.</p>
+<p><strong>Note:</strong> The workloadSelector configuration in
+Destination Rules reveals which workloads the traffic emanating from
+uses this Destination Rule, unlike the workloadSelector configuration
+in <a href="https://istio.io/docs/reference/config/networking/service-entry/#ServiceEntry">ServiceEntries</a>,
+which is used to reveal which workloads the traffic is routed to.</p>
 <p>{{<tabset category-name="selector-example">}}
 {{<tab name="v1alpha3" category-value="v1alpha3">}}</p>
 <pre><code class="language-yaml">apiVersion: networking.istio.io/v1alpha3


networking/v1alpha3/destination_rule.proto
```diff
@@ -168,6 +168,12 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// Destination Rules reveals which workloads the traffic emanating from
```
ilrudie, line 172:
```suggestion
// destination rule controls which consumer workloads
```

I think `reveals`, at least to my mind, seems like something a status field would do vs `controls` which is more spec.
I think "consumer workloads" is probably clearer. We could also use "client workloads".

networking/v1alpha3/destination_rule.proto
```diff
@@ -168,6 +168,12 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// Destination Rules reveals which workloads the traffic emanating from 
+// uses this Destination Rule, unlike the workloadSelector configuration
```
ilrudie, line 173:
```suggestion
// uses this destination rule, unlike the workloadSelector configuration 
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -168,6 +168,12 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// Destination Rules reveals which workloads the traffic emanating from 
+// uses this Destination Rule, unlike the workloadSelector configuration 
+// in [ServiceEntries](https://istio.io/docs/reference/config/networking/service-entry/#ServiceEntry),
```
ilrudie, line 174:
```suggestion
// in [ServiceEntry](https://istio.io/docs/reference/config/networking/service-entry/#ServiceEntry), 
```

networking/v1alpha3/destination_rule.proto
```diff
@@ -168,6 +168,12 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// Destination Rules reveals which workloads the traffic emanating from 
+// uses this Destination Rule, unlike the workloadSelector configuration 
+// in [ServiceEntries](https://istio.io/docs/reference/config/networking/service-entry/#ServiceEntry), 
+// which is used to reveal which workloads the traffic is routed to.
```
ilrudie, line 175:
```suggestion
// which is used to control which workloads the traffic is routed to.
```

ilrudie (commented):
I'm not mandating any of the suggested changes. The overall clarification you've written seems like a good one but I think we can workshop the wording a little bit to ensure the new note is clear and concise.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu, line 173: why do we need to compare with SE here

hzxuzhonghu: or why it is SE not other API that contains workloadSelector. IMO we donot have to do such explicit comparation

NaturezzZ:
Thank you @hzxuzhonghu, I removed the comparison with SE and clarified use case of `workloadSelector` and `subset` instead.

networking/v1alpha3/destination_rule.proto
NaturezzZ: removed and changed a better counter example in latest commit. thank you.

hzxuzhonghu (approved):

networking/v1alpha3/destination_rule.proto
```diff
@@ -168,6 +168,12 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// destination rule controls which client workloads 
+// use this destination rule, rather than which server workloads do.
+// [Subsets](https://istio.io/latest/docs/reference/config/networking/destination-rule/#Subset) 
+// can be used to select server workloads in fine granularity.
```
howardjohn, line 175: I don't think this is accurate. Subsets control load balancing on the clinet-side, it doesn't impact which servers implemnent which policies.

https://github.com/istio/istio/blob/388d5d1b9f37e7c3d15cf72967fad22f0b8e70fb/pilot/pkg/networking/core/v1alpha3/cluster_builder.go#L380 - no subset involved.

networking/v1alpha3/destination_rule.proto
```diff
@@ -168,6 +168,12 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// destination rule controls which client workloads 
+// use this destination rule, rather than which server workloads do.
```
howardjohn, line 173: https://github.com/istio/istio/blob/388d5d1b9f37e7c3d15cf72967fad22f0b8e70fb/pilot/pkg/networking/core/v1alpha3/cluster_builder.go#L380 - SidecarScope.DestinationRule does control the server workloads?

howardjohn (changes requested):

networking/v1alpha3/destination_rule.proto
NaturezzZ: Subsets are used to control which servers the traffic is routed to. I'll improve the wording here to eliminate ambiguity.

NaturezzZ:
Sorry for the late reply. I revised some wording to eliminate ambiguity. Would you mind take a look? Thank you. : )

networking/v1alpha3/destination_rule.proto
```diff
@@ -100,6 +100,14 @@ import "type/v1beta1/selector.proto";
 // The following example shows how a destination rule can be applied to a
 // specific workload using the workloadSelector configuration.
 //
+// **Note:** The workloadSelector configuration in 
+// destination rule controls which client workloads 
+// use this destination rule, rather than controlling 
+// which servers are selected as traffic endpoints.
```
Phil1602, line 106: I would argue vice-versa:

Most of the Istio resources (e.g. Sidecar, DestinationRule) use `workloadSelector` to target clients (==Envoys which apply/use this config).

`ServiceEntry` is IMO the only exception: It uses `workloadSelector` to select endpoints/targets for traffic flow.

Or did I miss something?

Phil1602 (commented):
Thanks for tackling this! 

Maybe we can revive this PR - I linked an issue which basically describes this issue you are documenting.

networking/v1alpha3/destination_rule.proto
NaturezzZ: I think so too.

PR #3091, opened 2024-02-15
[release-1.20] meshconfig: add compliance policy

kyessenov:
Change-Id: If15a8db11f03434a6daf52aad54360fe038abf2b

xref: https://github.com/istio/api/pull/3082


howardjohn (changes requested):
Lets hold until we validate the implementation so we do not prematurely commit to the API if it cannot be implemented

dhawton:
> what about 1.21?

https://github.com/istio/api/pull/3090 ;)

Issue #2994, opened 2023-11-16
Remove API version tabs


howardjohn:
https://istio.io/latest/docs/reference/config/networking/destination-rule/ for example.

There is no need for this. 100% of users should use v1beta1. We should pretend v1alpha3 does not exist.

bleggett:
Is there _any scenario where the docs _should_ show multiple versions of the API, rather than just the latest stable?

I guess the only place might be explicit migration guides or alpha/beta features?

+1 on default posture being "don't use them, at all".

AndreaM12345:
looks like this might be the full list of files to adjust:

https://istio.io/latest/docs/reference/config/networking/destination-rule/
https://istio.io/latest/docs/reference/config/networking/workload-group/
https://istio.io/latest/docs/reference/config/networking/gateway/
https://istio.io/latest/docs/reference/config/networking/service-entry/
https://istio.io/latest/docs/reference/config/networking/sidecar/
https://istio.io/latest/docs/reference/config/networking/virtual-service/
https://istio.io/latest/docs/reference/config/networking/workload-entry/
https://istio.io/latest/docs/reference/config/networking/workload-group/

https://istio.io/latest/docs/reference/config/security/request_authentication/
https://istio.io/latest/docs/reference/config/security/authorization-policy/

linsun:
+1 on this. multiple tabs were useful during transition but not needed atm.

PR #3100, opened 2024-02-27
docs: remove per-version API tabs

howardjohn:
Fixes https://github.com/istio/api/issues/2994

### DIFF
--- a/networking/v1alpha3/destination_rule.pb.go
+++ b/networking/v1alpha3/destination_rule.pb.go
@@ -31,22 +31,6 @@
 // balancing pool. For example, a simple load balancing policy for the
 // ratings service would look as follows:
 //
-// {{<tabset category-name="example">}}
-// {{<tab name="v1alpha3" category-value="v1alpha3">}}
-// ```yaml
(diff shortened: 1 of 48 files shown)


linsun (approved):
LGTM

ericvn:
One thing I want to investigate, which is likely a nit, is why the .html file is under an older version in the source tree (ex: networking/v1alpha3/service_entry.pb.html as opposed to being in v1beta1). I suspect we have logic to only create one in the older version (at least in the case of dupes). Overall, this doesn't make a change in what is shown in istio.io, it's just where the file is placed here.

howardjohn:
> One thing I want to investigate, which is likely a nit, is why the .html file is under an older version in the source tree (ex: networking/v1alpha3/service_entry.pb.html as opposed to being in v1beta1). I suspect we have logic to only create one in the older version (at least in the case of dupes). Overall, this doesn't make a change in what is shown in istio.io, it's just where the file is placed here.

v1beta1 ones have `$mode: none` to avoid generation to avoid duplicates

ericvn (approved):

PR #3104, opened 2024-02-28
Document node label in DR failover

howardjohn:


hzxuzhonghu (approved):

zirain (approved):

kfaseela (approved):

ericvn (approved):
My only concern looking at this is this mentioning `kubernetes.io/hostname` matching node, but reading https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/ I see:
```
Note: The value of these labels is cloud provider specific and is not guaranteed to be reliable. For example, the value of kubernetes.io/hostname may be the same as the node name in some environments and a different value in other environments.
```
Since we are matching label values and mention the mapped label to which it is matching, I think this is fine.

PR #3111, opened 2024-03-06
Networking APIs graduation to v1

whitneygriffith:
As discussed in [Release Channels RFC](https://docs.google.com/document/d/1onST4-swbZE1UPCDMQm1c7T5rzYEn7Wo4uFNShz0kNw/edit#heading=h.ev9kkx8ollvx), `v1beta1` networking APIs can be promoted to `v1`.


networking/v1/proxy_config.proto
```diff
@@ -0,0 +1,138 @@
...
+// +cue-gen:ProxyConfig:preserveUnknownFields:false
+// -->
+//
+// <!-- go code generation tags
+// +kubetype-gen
+// +kubetype-gen:groupVersion=networking.istio.io/v1
+// +genclient
+// +k8s:deepcopy-gen=true
+// -->
+message ProxyConfig {
```
hzxuzhonghu, line 110: ProxyConfig is not that mature as other apis

releasenotes/notes/graduate-networking-apis.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+releaseNotes:
+  - |
+    **Graduated** Networking APIs (DestinationRule, Gateway, ProxyConfig, ServiceEntry, Sidecar, VirtualService, WorkloadEntry, WorkloadGroup) are graduated to v1. API's storageVersion is v1beta1.
```
linsun, line 6: What does storageVersion mean here?

networking/v1/proxy_config.proto
linsun: +1 on proxyconfig

linsun:
Do we feel all network APIs can be promoted to v1? Feel some are newer and less used, for example workload entry and workload group.

costinm:
WorkloadEntry and group are newer - but that's because they are
improvements to ServiceEntry, and replace the 'mesh expansion'
use in ServiceEntry based on what we learned.

It would be pretty bad to promote ServiceEntry to v1 ( with the parts we
know are obsolete and should be replaced with WE ) and
not promote WE.

I'm pretty concerned with the user perception that everything in the API is
covered by v1 - we have plenty of fields that are obsolete
like the SE endpoints ( replaced by WE), and other fields that got added as
'experimental'.

Old doesn't (always) mean good, in particular in Istio APIs where we have
no mechanism to make changes after 'experimental/alpha'.

networking/v1/proxy_config.proto
keithmattix: This is fair; I'd be fine eliding ProxyConfig

keithmattix:
> It would be pretty bad to promote ServiceEntry to v1 ( with the parts we
know are obsolete and should be replaced with WE ) and
not promote WE.

IIUC based on the PR, WorkloadEntry/Group are being promoted as well?

releasenotes/notes/graduate-networking-apis.yaml
keithmattix: +1 a link to https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version would be nice, but I also think the last sentence is both more appropriate for an upgrade note and it needs more detail. For instance, this change only affects users who are applying v1alphaN APIs, and those users need to either use the storageVersion migrator or update their resources once when upgrading.

networking/v1/proxy_config.proto
whitneygriffith: Do we have enough information to improve its maturity so that it can be graduated to `v1`?

keithmattix: IMO, the community needs to have a conversation about whether this resource needs to exist at all. It's been brought up in the past, but not sure there's a firm commitment to actually making changes at all

costinm:
It would be great to remove some of the fields in ServiceEntry.

Re. ProxyConfig - usage is low, I don't mind keeping it at beta or even
deprecating it. However we should make sure that the
annotation and mesh config are also treated as alpha/beta and don't get
promoted. We have not discussed the status
of annotations and labels.

whitneygriffith:
> It would be great to remove some of the fields in ServiceEntry. Re. ProxyConfig - usage is low, I don't mind keeping it at beta or even deprecating it. However we should make sure that the annotation and mesh config are also treated as alpha/beta and don't get promoted. We have not discussed the status of annotations and labels.


Which fields will you propose removing from ServiceEntry?

networking/v1beta1/workload_group.proto
```diff
@@ -100,7 +100,7 @@ option go_package = "istio.io/api/networking/v1beta1";
 // +k8s:deepcopy-gen=true
 // -->
 // <!-- istio code generation tags
-// +istio.io/sync-from:networking/v1alpha3/workload_group.proto
```
howardjohn, line 103: We still have the alpha API.

We need 1 to be the source of truth with sync-start, and the rest have sync-from pointing to that.

networking/v1/proxy_config.proto
whitneygriffith: We agreed ProxyConfig should not be promoted to v1. I have made the relevant changes

networking/v1beta1/workload_group.proto
whitneygriffith: Can we only make the beta the sync-start if the alpha is removed? I am guessing yes now.

whitneygriffith: reverted sync changes

whitneygriffith:
From today's working group meeting, we have agreed to promote all Networking APIs as is excepting ProxyConfig. Those changes have been made. PTAL

hzxuzhonghu (approved):
I am ok without proxyConfig

ericvn:
These changes LGTM.

releasenotes/notes/promote-networking-apis-v1.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+releaseNotes:
+  - |
+    **Promoted** Networking APIs (DestinationRule, Gateway, ServiceEntry, Sidecar, VirtualService, WorkloadEntry, WorkloadGroup) to v1.
+upgradeNotes:
+  - title: Upgrade existing v1alpha3 networking APIs
```
howardjohn, line 8: I wonder if we should remove this. There is actually no need to do this unless we want to remove v1alpha3. We cannot do that for _years_ due to revision support.

If or when we do, K8s will have first class support to do this (in k8s 1.30), or we can add it to istioctl.

Attempting to follow that doc made me **extremely** frustrated, one of the worst UXs in the ecosystem - best to avoid telling users to do that when its not needed IMO

howardjohn (commented):
LGTM, we should get broad signoff like for the PA change though

releasenotes/notes/promote-networking-apis-v1.yaml
whitneygriffith: @howardjohnwe just to confirm you are saying, we can still keep v1beta1 as storageVersion and as long as v1alpha3 exists, users wouldn't be forced to migrate so we can remove this upgrade note.

howardjohn: Correct

whitneygriffith: made the changes

whitneygriffith:
> LGTM, we should get broad signoff like for the PA change though

PTAL @nrjpoddar @ericvn @therealmitchconnors @linsun

louiscryan:
LGTM. Thanks @whitneygriffith

whitneygriffith:
Thanks all! We are just waiting on one last approval to merge the changes.

therealmitchconnors (approved):

whitneygriffith:
Part of [173](https://github.com/istio/enhancements/issues/173)

PR #3114, opened 2024-03-11
add some annotations for ambient traffic redirection 

zengyuxing007:
ref: https://github.com/istio/istio/issues/49829

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -530,3 +530,66 @@ annotations:
     hidden: true
     resources:
       - Any
+
+  - name: traffic.istio.io/includeOutboundIPRanges
+    featureStatus: Alpha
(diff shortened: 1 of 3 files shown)


howardjohn (approved):

howardjohn (commented):
These LGTM as documentation in general, but not for ambient necessarily; that is seperate

kfaseela (approved):

annotation/annotations.yaml
```diff
@@ -530,3 +530,66 @@ annotations:
     hidden: true
     resources:
       - Any
+
+  - name: traffic.istio.io/includeOutboundIPRanges
```
zirain, line 534: Need to highlight these annotations are used for ambient

PR #3122, opened 2024-03-12
Update insecureSkipVerify comment

leosarra:
Starting with Istio 1.21 VERIFY_CERTIFICATE_AT_CLIENT is true by default.
A comment had to be updated


kfaseela (approved):

networking/v1alpha3/destination_rule.proto
```diff
@@ -1168,10 +1168,7 @@ message ClientTLSSettings {
   // without `VERIFY_CERTIFICATE_AT_CLIENT` enabled, verification of the CA signature and
   // SAN will be skipped.
   //
-  // `insecureSkipVerify` is `false` by default.
-  // `VERIFY_CERTIFICATE_AT_CLIENT` is `false` by default in Istio version 1.9 but will
-  // be `true` by default in a later version where, going forward, it will be
-  // enabled by default.
+  // `insecureSkipVerify` is `false` by default while `VERIFY_CERTIFICATE_AT_CLIENT` is `true` by default.
```
hzxuzhonghu, line 1171: the relationship with VERIFY_CERTIFICATE_AT_CLIENT looks very confusing, would cause usage trouble

leosarra: Yes I agree. Overall the VERIFY_CERTIFICATE_AT_CLIENT makes the doc a bit more confusing for the user.
On the other end we can look forward to the removal of VERIFY_CERTIFICATE_AT_CLIENT once the deprecation process is completed (given that it is already in the compatibility values starting with Istio 1.21, it is planned for removal in 1.24).

Perhaps this field can be simplified as follows:

```
// The `insecureSkipVerify` flag determines whether the proxy should skip verifying the
// CA signature and SAN (Subject Alternative Name) for the server certificate associated
// with a specific host. This field is ignored if the environment variable
// `VERIFY_CERTIFICATE_AT_CLIENT` is set to `false` (the default is `true`),
// as the verification of the CA signature and SAN will be bypassed regardless.
```

Do you think it is better now?

hzxuzhonghu: From the code, it only collaborate with VERIFY_CERTIFICATE_AT_CLIENT when deciding whether to enable AutoSanValidation.  
```
func (cb *ClusterBuilder) setAutoSniAndAutoSanValidation(mc *clusterWrapper, tls *networking.ClientTLSSettings) {
	if mc == nil || !features.EnableAutoSni {
		return
	}

	setAutoSni := false
	setAutoSanValidation := false
	if len(tls.Sni) == 0 {
		setAutoSni = true
	}
	if features.VerifyCertAtClient && setAutoSni && len(tls.SubjectAltNames) == 0 && !tls.GetInsecureSkipVerify().GetValue() {
		setAutoSanValidation = true
	}

```

So the saying is not right

hzxuzhonghu: SGTM without mentioning VERIFY_CERTIFICATE_AT_CLIENT

leosarra: I removed mentions of VERIFY_CERTIFICATE_AT_CLIENT and ENABLE_AUTO_SNI.
I think it is fine to remove the mentions of ENABLE_AUTO_SNI since it is going through the same deprecation process as VERIFY_CERTIFICATE_AT_CLIENT.

howardjohn (approved):

PR #3126, opened 2024-03-14
fix sidecar reference page

zirain:
followup: https://github.com/istio/api/pull/3100
xref: https://github.com/istio/istio.io/pull/14739

### DIFF
--- a/networking/v1alpha3/sidecar.proto
+++ b/networking/v1alpha3/sidecar.proto
@@ -255,6 +255,26 @@ import "networking/v1alpha3/virtual_service.proto";
 // This feature is currently experimental.
 //
 // ```yaml
+// apiVersion: networking.istio.io/v1beta1
+// kind: Sidecar
+// metadata:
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/sidecar.proto
```diff
@@ -255,6 +255,26 @@ import "networking/v1alpha3/virtual_service.proto";
 // This feature is currently experimental.
 //
 // ```yaml
+// apiVersion: networking.istio.io/v1beta1
```
hzxuzhonghu, line 258: Do i miss some ctx, why all the versions are vabeta1

hzxuzhonghu (approved):

networking/v1alpha3/sidecar.proto
zirain: acutally, I believe this was wrong for a long time.

ericvn (approved):

PR #3128, opened 2024-03-14
Update PeerAuth Documentation to indicate that API Dev is frozen

### LINKED ISSUE #3125: Update PeerAuth Documentation to indicate that API Dev is frozen
I think we should update the API docs to indicate that the API is also frozen and likely to be replaced in ambient

_Originally posted by @louiscryan in https://github.com/istio/api/issues/3112#issuecomment-1995142260_

whitneygriffith:
Resolves #3125


ericvn (approved):
Thanks

jaellio (approved):

PR #3130, opened 2024-03-15
Update DestinationRule api:  sync add the cookie attributes for consistentHash 

zengyuxing007:
xref: https://github.com/istio/istio/issues/49870

envoy cookie affinity support [cookie attributes](https://en.wikipedia.org/wiki/HTTP_cookie) now, ref doc:  https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#envoy-v3-api-msg-config-route-v3-routeaction-hashpolicy-cookie


evenh:
Looking forward to this 🤞🏻

Issue #173, opened 2017-09-12
statsdUdpAddress in MeshConfig


andraxylia:
This has been added to proxyConfig only. Investigate if it is needed in MeshConfig as well.

rshriram:
fixed..

PR #3133, opened 2024-03-18
Promote Telemetry API to v1

whitneygriffith:
Resolves [46720](https://github.com/istio/istio/issues/46720)

Telemetry Features not promoted: 

- [metrics.reportingInterval](https://istio.io/latest/docs/reference/config/telemetry/#Metrics)
- [accessLogging.filter](https://istio.io/latest/docs/reference/config/telemetry/#AccessLogging)
- [tracing.useRequestIdForTraceSampling](https://github.com/istio/api/blob/d29365cf33d2219bf397e82fdbb5cda8eb112f01/telemetry/v1alpha1/telemetry.proto#L384)

These features are annotated with `// +cue-gen:Telemetry:releaseChannel:standard` which will be functional with the release channel work [#173](https://github.com/istio/enhancements/issues/173)

See [Telemetry API RFC](https://docs.google.com/document/d/16A-E-30txN5Y2V_9qrDNpVC3lT7P6aa_3Qdg6_4t5Sg/edit?usp=sharing)

Next up: 
- [ ] Recognize Telemetry v1 APIs similar to [#50171](https://github.com/istio/istio/pull/50171) and [#50172](https://github.com/istio/istio/pull/50172)


telemetry/v1alpha1/telemetry.proto
```diff
@@ -675,5 +680,6 @@ message AccessLogging {
 
   // Optional. If specified, this filter will be used to select specific
   // requests/connections for logging.
+  // +cue-gen:Telemetry:releaseChannel:standard
```
jaellio, line 683: For my own understanding, could you clarify why only some fields are marked with `cue-gen:Telemetry:releaseChannel:standard`? Thanks!

whitneygriffith: Of course, the fields that are marked `cue-gen:Telemetry:releaseChannel:standard` are the fields that have not been graduated to `v1`. The fields have to be marked in both the `v1alpha` and `v1` versions as at this moment, `v1` is synced to always be a copy of `v1alpha1`. [Here's](https://docs.google.com/document/d/1onST4-swbZE1UPCDMQm1c7T5rzYEn7Wo4uFNShz0kNw/edit#heading=h.7xeg0ys0wzdw) the section that speaks on this in the design doc.

linsun:
@zirain @lei-tang @kyessenov what is your thought on this? Do you feel confident it should be v1 API now?

releasenotes/notes/promote-telemetry-v1.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: release-notes/v2
+kind: feature
+area: telemetry
+releaseNotes:
+  - |
+    **Promoted** Telemetry API to v1. All Telemetry fields excepting `metrics.reportingInterval` and `accessLogging.filter` are now stable and will not change in a backwards-incompatible way.
```
lei-tang, line 6: I think a doc listing the tests, the user guides, the usage data for each API field to promote will help reviewers to evaluate its v1 readiness. We probably do not want to promote untested/undocumented/rarely used API fields. It is engineering costly to maintain a v1 API field.

lei-tang (commented):
I think a doc listing the tests and the user guides for each API field to promote will help evaluate its v1 readiness. We probably do not want to promote untested/undocumented API fields.

releasenotes/notes/promote-telemetry-v1.yaml
whitneygriffith: @lei-tang does this [doc](https://docs.google.com/document/d/16A-E-30txN5Y2V_9qrDNpVC3lT7P6aa_3Qdg6_4t5Sg/edit) not suffice? How can it be improved? We have leveraged the customer usage data we do have access to.

whitneygriffith: I can amend the doc to point to the test and user guides. The data we got from Google was the only quantitative usage feedback. Beyond that usage data was mostly debated on the [doc](https://docs.google.com/document/d/16A-E-30txN5Y2V_9qrDNpVC3lT7P6aa_3Qdg6_4t5Sg/edit?usp=sharing).

whitneygriffith: @imnmo has also made several attempts to get usage data from other known vendors

lei-tang: Thanks for sharing a doc! Clicking the "See [Telemetry API RFC](https://docs.google.com/document/d/16A-E-30txN5Y2V_9qrDNpVC3lT7P6aa_3Qdg6_4t5Sg/share)" doc link in the PR description results in a "page not found" error.

I think the doc can be improved by listing the tests, the user guides, and the usage data of each API field. 
Something like:
API field 1: links to the tests, links to the user guides, usage data (if usage data is unknown, mark as unknown).
API field 2: links to the tests, links to the user guides, usage data (if usage data is unknown, mark as unknown).
....

whitneygriffith: Thanks for catching that! I hope it works now. And I will improve the doc accordingly.

whitneygriffith: PTAL, I have updated the [doc](https://docs.google.com/document/d/16A-E-30txN5Y2V_9qrDNpVC3lT7P6aa_3Qdg6_4t5Sg/edit?usp=sharing).

louiscryan:
LGTM

Thanks for driving

louiscryan (approved):

howardjohn (approved):

ericvn:
LGTM as well, although I thought I had put a comment in the PR previously. Thanks.

kyessenov:
Sorry, just seeing it. As I always bring it up - I'm not sure you can claim backwards compatibility for things that are not provided in Envoy. CEL and access log formatter syntax are not part of xDS API and are not subject to compatibility expectations, so that's not something that Istio can provide. I'm also not sure what other extensions are needed by this API and whether they are mature enough.

costinm:
Access log formatter or labels are not part of the v1 API.

CEL on the other hand - and the metric label opt-in/out - are part of the
API, I had my doubts as well but it looks like the use cases are strong and
replacing it wasn't feasible.

xDS API and Envoy may change - but the Istio project is committed to
support the v1 API. If Envoy makes changes or lacks a feature - it will
be our responsibility to implement  Istio-specific extensions or fix Envoy
and continue to support the API for our users.

Good news is that (AFAIK) the v1 API does not extend to ambient - in
ambient L4 and L7 are swappable and are not
required to be OSS Istio-Envoy based. I think that implies that telemetry
API and set of metrics may not be based on Istio,
but on whatever Gateway or OTel define.

howardjohn:
IMO: Swappable waypoints does not imply that Istio APIs change. An implementation can chose to implement the Istio waypoints or not. Probably we would have some conformance to say they are "Base Waypoint" conformant vs "Istio API Waypoint" conformant.

costinm:
On Tue, Apr 9, 2024 at 7:55 AM John Howard ***@***.***> wrote:

> IMO: Swappable waypoints does not imply that Istio APIs change. An
> implementation can chose to implement the Istio waypoints or not. Probably
> we would have some conformance to say they are "Base Waypoint" conformant
> vs "Istio API Waypoint" conformant.
>
Istio API does change between '????' and 'ambient' - even if the same CRDs
are used, the fields, usage and behavior are different,
workload selector versus target, client overrides, etc. (  don't know how
we're calling the 'original Istio API for sidecars with workload selectors
and all the features promoted to v1')

Telemetry in particular is quite different.

For Waypoints that implement the Gateway API - I agree passing the
Gateway/GAMMA conformance would be very important,
and they should list which 'vendor extensions' they support. I don't know
if it's reasonable to expect a non-Istiod Waypoint to
support Istio APIs, but if it happens they would include it in their docs,
including what subset of fields they do ( since the
CRDs we have are a mix of fields ). Do you have any expectations that
other, non Istiod-based implementations ( i.e. not
forks of Istio) will suddenly adopt Istio API ? That was the belief 5 years
ago, that other meshes and gateways will adopt our
API and telemetry.

We should discuss it today, around API versioning - each field in the CRD
that applies to ambient should be marked
somehow, as well as the docs need some update for the behaviors that
change. Telemetry v1 for example
has no mention of how it behaves in ambient or for K8S Gateways/GAMMA.

PR #3134, opened 2024-03-20
OpenTelemetryTracingProvider: Allow configuring samplers

joaopgrassi:
API part of istio/istio#50001

The OpenTelemetry project defines a [Sampler API](https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampler) and this API is available [now in Envoy](https://www.envoyproxy.io/docs/envoy/v1.29.2/api-v3/config/trace/opentelemetry/samplers). With the next Envoy release, it will also come with a [Dynatrace Sampler](https://github.com/envoyproxy/envoy/pull/32598). This API in Envoy allows anyone (vendors for ex) to add their own sampler.

This PR then adds this new configuration on Istio, containing for now the Dynatrace sampler config. 

Today, there's multiple ways to configure sampling, which works/have the following precedence:

```
telemetry.RandomSamplingPercentage > defaultConfig.tracing.sampling > PILOT_TRACE_SAMPLING
```

This PR adds another option, changing the order above to: 

```
provider.sampler > telemetry.RandomSamplingPercentage > defaultConfig.tracing.sampling > PILOT_TRACE_SAMPLING
```

Which means:

- If `OpenTelemetryTracingProvider` has a custom sampler configured (**what this PR adds**): 
  - Set `telemetry.RandomSamplingPercentage` to `100` as all spans should arrive on the sampler, so it can perform its algorithm. This is hardcoded and the user cannot change (doesn't make sense to). Without all spans a sampler can't make consistent sampling decisions. 

- Else, continue with the same precedence as today. Essentially nothing changes when users don't use a Sampler: 

```
 telemetry.RandomSamplingPercentage > defaultConfig.tracing.sampling > PILOT_TRACE_SAMPLING
```

Note: I didn't add the `AlwaysOn` since that is basically achieved the same way by setting the `telemetry.RandomSamplingPercentage` to `100`. 

CC @howardjohn @zirain


mesh/v1alpha1/config.proto
```diff
@@ -1113,6 +1113,75 @@ message MeshConfig {
...
+        //
+        // - `root_spans_per_minute` is unset
+        // - `root_spans_per_minute` is set to 0
+        uint32 root_spans_per_minute = 3;
+
+        // Optional. Dynatrace HTTP API to obtain sampling configuration.
+        //
+        // When not provided, the Dynatrace Sampler will re-use the configuration from the OpenTelemetryTracingProvider HTTP Exporter
+        // (`service`, `port` and `http`), including the access token.
+        DynatraceApi http_service = 4;
```
joaopgrassi, line 1168: The idea behind this optional field is: Dynatrace customers use the OpenTelemetry collector as a central place for processing the spans before sending it to Dynatrace (removing PII from attributes, adding attributes, etc). 

In this case, Istio/Envoy can be configured to export to the Collector via gRPC, but still use the Dynatrace Sampler to sample the spans properly before exporting to the collector. 

When the Dynatrace customer doesn't use the collector and prefers to send the spans directly to the Dynatrace API, then the Sampler simply re-use the config from the HTTP exporter (service/host, http headers and etc). This is so the configuration minimal as most will be just repeated.

mesh/v1alpha1/config.proto
```diff
@@ -1113,6 +1113,75 @@ message MeshConfig {
       //       environment: {}
       // ```
       ResourceDetectors resource_detectors = 5;
+
+      // Optional. Configure a [Sampler](https://opentelemetry.io/docs/specs/otel/trace/sdk/#sampler)
+      // to be used by the OpenTelemetry Tracer.
+      oneof sampling {
```
zirain, line 1119: so this's another way(IIRC, the fourth.) to config sampling, can you tell what will this Telemetry API?

joaopgrassi: Yes, correct, multiple ways to configure this unfortunately. This is what John and I discussed in a previous PR https://github.com/istio/api/pull/3002#discussion_r1445079905 

> can you tell what will this Telemetry API

Do you mean what this new way to configure a sampler interfere with the Telemetry API? If so, I explained it in the linked issue here: https://github.com/istio/istio/issues/50001

zirain: yes, I think we should point that in somwhere else, issue is not a good place to end-users.

joaopgrassi: Ah sure, I agree. The issue was more for PR reviews here. We can also decide where to put this instructions, maybe in the docs?

zirain: you should put brief comment here, and I recall you're also working on a new page about OpenTelemetry tracing, you can share the details then, WDYT?

joaopgrassi: Yep, sounds good! I will copy the contents of the issue to this PR as well, so at least there's something in here.

joaopgrassi: @zirain the docs PR I was working on should be ready to review now: https://github.com/istio/istio.io/pull/14566. Once that base is there, I can work to document the ways a user can configure sampling.

zirain (approved):
LGTM

mesh/v1alpha1/config.proto
hzxuzhonghu: I think it confuses if this is set under opentelemetry provider

mesh/v1alpha1/config.proto
```diff
@@ -1113,6 +1113,75 @@ message MeshConfig {
...
+        //       tenant: "{your-environment-id}"
+        //       cluster_id: 1234
+        DynatraceSampler dynatrace_sampler = 6;
+      }
+
+      message DynatraceSampler {
+        // REQUIRED. The Dynatrace tenant.
+        //
+        // The value can be obtained from the Istio deployment page in Dynatrace.
+        string tenant = 1;
```
hzxuzhonghu, line 1148: what's this mean

mesh/v1alpha1/config.proto
```diff
@@ -1113,6 +1113,75 @@ message MeshConfig {
...
+
+        // REQUIRED. The id of the Dynatrace cluster.
+        //
+        // The value can be obtained from the Istio deployment page in Dynatrace.
+        int32 cluster_id = 2;
```
hzxuzhonghu, line 1153: how could this value set when use multicluster setup

mesh/v1alpha1/config.proto
joaopgrassi: This is a Dynatrace specific field. Each Dynatrace customer receives a "Tenant Id" and that is what this field is for. It identifies the customer's tenant in our SaaS product.

mesh/v1alpha1/config.proto
joaopgrassi: This is also a Dynatrace specific field. It is related to the Dynatrace cluster the customer's tenant is assigned. It has no relation to Istio whatsoever :)

joaopgrassi: I could rephrase it a bit, to avoid mixing with Istio/Envoy terms, if you think that's better. Maybe `The id of the cluster in the Dynatrace product` or smt of the like that.

mesh/v1alpha1/config.proto
zirain: it's fine to place under `DynatraceSampler`, a vendor extension.

mesh/v1alpha1/config.proto
joaopgrassi: I updated the proto docs. I hope it's more clear now.

howardjohn (approved):

PR #3142, opened 2024-03-29
Add pathTemplates and notPathTemplates to AuthorizationPolicy

jaellio:
Increases templating support for positive and negative path mathching.

Part of https://github.com/istio/istio/issues/16585


hzxuzhonghu (commented):
should we deprecate paths and notPaths

howardjohn:
I don't think we should make a new field but rather us 'path'. there shouldn't be any compatibility concerns as {} is not a valid character in a real path. all the discussions about what we support in templates is constraint inside {}

jaellio:
Closing in favor of https://github.com/istio/istio/pull/50365

PR #3154, opened 2024-04-05
Update mesh config to not import k8s libraries

howardjohn:
Impl in https://github.com/istio/istio/pull/51041. Drops 11% of the agent binary size

--

This PR decouples our APIs from k8s.io, which are super bloated. This gives the same user facing interface. Mesh config is not a proto that is sent over the wire.


keithmattix:
What's the motivation for this?

howardjohn:
The motivation here is to not require API importers to depend on bloated k8s libraries. In particular, pilot-agent

mesh/v1alpha1/config.proto
```diff
@@ -1428,6 +1427,38 @@ message MeshConfig {
    SERVICE_REGISTRY = 0;
  }
 
+message LabelSelector {
```
jwendell, line 1430: worth to also copy upstream comment for this struct?

jwendell: and perhaps also mention that both these structs are copies from upstream repo?

jwendell (approved):

kyessenov (approved):

hzxuzhonghu (approved):

ericvn (approved):

PR #3155, opened 2024-04-08
unhide use_request_id_for_trace_sampling

zirain:
this's introduced for a long time, should unhide.


zirain:
the main concern from me, why this was introduced but hide form the begining?

Can we just remove it?

keithmattix:
FWIW, this field was not promoted in #3133

costinm:
I don't know if this feature is still needed - if you want to unhide it I assume you have a user who needs it, and that's
a good enough reason to keep it IMO. 

My concern is about doing a proper API review when we add something to the v1 API - I think we all agree that the
past process of not doing API ( or security, scalability, etc) review when adding experimental/hidden features and
then promoting them unchanged and without review to v1 is not ideal.

whitneygriffith:
We should decide on if this field should be deprecated and therefore move forward with removing it completely instead of unhiding it.

bleggett:
https://github.com/istio/api/pull/2104 was the original add, for context

bleggett:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/observability/tracing#trace-context-propagation

Envoy docs themselves suggest using this field. 

April 10 WG decision - follow Up or Out: https://docs.google.com/document/d/1wsa06GGiq1LEGwhkiPP0FKIZJqdAiue-VeBonWAzAyk/edit#heading=h.6fcb5py1h0hy

- 1 week for someone to say they still want this
- Then, we deprecate (from CRDs) + remove 

My $0.02 is that

- We should not unhide it in docs
- We should deprecate and remove it from CRDs
- If we want to hardcode it to true to keep from breaking existing trace sampling behaviors, that's fine
- If there's a need for this to be _configurable_, then we need a proper proposal

zirain:
let's delete it in the next step.

PR #3159, opened 2024-04-12
policy attachment: allow `targetRefs`

howardjohn:
Based on https://github.com/kubernetes-sigs/gateway-api/pull/2966. Note
that we do not HAVE to follow the GatewayAPI here; we can make our own
decision. There is, however, a general desire to allow multiple for
ergonomics.

In this proposal, I hide `targetRef`, but the API will remain + be
implemented forever. Implementation cost here is near zero, as we can
easily translate it to a single `targetRefs`; we just hide from docs to
push users toward the new ones.

PR #3160, opened 2024-04-12
EnvoyFilter: implement `targetRefs`

howardjohn:
This is the only API we have a `selector` without `targetRef`.

The motivation at the time was that waypoints don't official support
EnvoyFilter, and targetRef was primarily for waypoints.

However, targetRef can be used with all Kubernetes Gateway, including
for ingress, where EnvoyFilter is supported. Also, long term it will
support waypoint as well I assume; the earlier we add the field the less
migration pain there is.

This PR goes directly to `targetRefs` in line with
https://github.com/istio/api/pull/3159.


costinm:
Looks good to me, makes a lot of sense. 

BTW - worth taking a look at https://gateway.envoyproxy.io/v0.5.0/design/extending-envoy-gateway/ and https://gateway.envoyproxy.io/v0.5.0/design/envoy-patch-policy/ - while we are commited to never promote EnvoyFilter to v1, we can support APIs defined by other orgs or work with others to standardize common extensions for Envoy-based gateways.

costinm (approved):

networking/v1alpha3/envoy_filter.proto
```diff
@@ -847,6 +848,24 @@ message EnvoyFilter {
...
+  //
+  // If not set, the policy is applied as defined by the selector.
+  // At most one of the selector and targetRefs can be set.
+  //
+  // NOTE: If you are using the `targetRefs` field in a multi-revision environment with Istio versions prior to 1.22,
+  // it is highly recommended that you pin the policy to a revision running 1.22+ via the `istio.io/rev` label.
+  // This is to prevent proxies connected to older control planes (that don't know about the `targetRefs` field)
+  // from misinterpreting the policy as namespace-wide during the upgrade process.
+  //
+  // NOTE: Waypoint proxies are required to use this field for policies to apply; `selector` policies will be ignored.
```
keithmattix, line 866: nit: I don't think this is true yet; it's still behind a feature flag: https://github.com/istio/istio/blob/3ddb82391650bcff9808600c0af73b1101f88d0d/pilot/pkg/features/pilot.go#L216

keithmattix (approved):

ericvn (approved):

PR #3171, opened 2024-04-23
cors: add unmatched_preflights

zirain:
xref: https://github.com/envoyproxy/envoy/pull/33051

Introduce `forward_not_matching_preflights` allow to return local reply when preflight origin does not match allowed origins.


networking/v1/virtual_service.proto
```diff
@@ -1366,6 +1366,11 @@ message CorsPolicy {
   // (not the preflight) using credentials. Translates to
   // `Access-Control-Allow-Credentials` header.
   google.protobuf.BoolValue allow_credentials = 6;
+
+  // Indicates whether preflight requests not matching the configured 
+  // allowed origin should be forwarded to the upstream. 
+  // Default is true for compatibility.
```
hzxuzhonghu, line 1372: should be false? It is a little wried to make a bool default value true

zirain: yes, should rename it and make it default false.

ramaraochavali: Better to call it "donot_forward_unmatched_preflights" if we want to use Bool or use the original name but change it to bool like Envoy. Either of it works

ramaraochavali (approved):
LGTM

networking/v1/virtual_service.proto
```diff
@@ -1366,6 +1366,11 @@ message CorsPolicy {
   // (not the preflight) using credentials. Translates to
   // `Access-Control-Allow-Credentials` header.
   google.protobuf.BoolValue allow_credentials = 6;
+
+  // Indicates whether preflight requests not matching the configured 
+  // allowed origin shouldn't be forwarded to the upstream. 
+  // Default is false.
+  google.protobuf.BoolValue donot_forward_unmatched_preflights = 8;
```
hzxuzhonghu, line 1373: This is double negative, i donot think it is good. 

I mean  `forward_unmatched_preflights` default value false. 

Does this violate current behavior?

ramaraochavali: yes

zirain: if you want default false and keep current behavior, cannot avoid double negative.

howardjohn: You can make it an enum ?

`unmatched_preflights = UNSPECIFIED | FORWARD | IGNORE`

zirain: sounds good

networking/v1/virtual_service.proto
```diff
@@ -1366,6 +1366,20 @@ message CorsPolicy {
...
+
+  // Indicates whether preflight requests not matching the configured 
+  // allowed origin shouldn't be forwarded to the upstream. 
+  // Default is forward to upstream.
+  UnmatchedPreflights unmatched_preflights = 8;
+
+  enum UnmatchedPreflights {
+    // Preflight requests not matching the configured allowed origin
+    // will be forwarded to the upstream.
+    FORWARD = 0;
```
howardjohn, line 1378: I would set 0 to UNSPECIFIED. If we do `forward = 0`, we can never change our mind in the future and make 'ignore' the default. Even if we don't plan to now, its best to be precise

zirain: updated

howardjohn (approved):

PR #3173, opened 2024-04-26
Added support for configuring JWKS using local file

wulianglongrd:
fix https://github.com/istio/istio/issues/50196


howardjohn:
Is the real use case here "load from secret" and we are implementing it by "load from file" when its maybe not the best option? Or do we think file is a primary use case?

cc @louiscryan @costinm @aryan16

wulianglongrd:
JWKS is a public key set, so I don't think it is security sensitive.

howardjohn:
Isn't the use case in https://github.com/istio/istio/issues/50196 that is linked stating it is private for their use case?

wulianglongrd:
Oh yes. From what I understand, the `Secret` is just the way the file is stored, not how envoy perceives it? After the `Secret` is mounted to the container, is it just a normal text file to envoy? (I'm not familiar with kubernetes, I would be grateful if you could point out my mistake)

@ks-yim Can you describe your operation in detail? It would be nice to describe how envoy reads and uses mounted files. Thanks.

howardjohn:
> Oh yes. From what I understand, the Secret is just the way the file is stored, not how envoy perceives it? After the Secret is mounted to the container, is it just a normal text file to envoy? (I'm not familiar with kubernetes, I would be grateful if you could point out my mistake)

Correct. But Istio can also read Secrets directly and send them to envoy.

On that note - JWKS are generally considered public and shown directly in envoy config dumps, logs, etc. Secrets are not shown. This may be problematic

wulianglongrd:
No matter if the jwks is public or private, it should be public to envoy. We can use `Secret` or any other method to mount the file, as long as the end result for envoy is a readable `local file`.

I think this api just tells envoy to read a file from a certain address, and envoy doesn’t care why the file exists.

Also, like you said, if istio reads the `Secret` directly and sends it to envoy, then it will be directly shown in the envoy config dumps. But using files will not, so this api seems to have its purpose.

wulianglongrd:
I don't need this api right now, but I'm currently using jwt authentication. It seems to make sense to me and with this api it might be an option for me in the future.

hzxuzhonghu:
Loading jwks from local file is commonly used in microservice. But in istio managed no matter sidecar or gateway, mounting secret seems not that native.

ks-yim:
The motivation of https://github.com/istio/istio/issues/50196 was that there's no easy way to securely validate JWTs signed by symmetric key algorithms without leaking private jwks somewhere in plaintext and I wish istio has a native API to support this use case.

The original workflow that I had in mind was mounting a `Secret` that contains jwks as file onto istio-proxy container and let `jwt_authn` filter refer to the mounted file and I thought this can be achieved without excessive efforts because Envoy's `jwt_auhtn` has the feature built-in. jwks are not leaked because only the filename will be shown in the config dump.

This, though, requires users an extra work to mount the secret on their own to istio-proxy container and if there's still a room for improvement on how istio-proxy reads the secret, I am up for it.

wulianglongrd:
> > Also, like you said, if istio reads the `Secret` directly and sends it to envoy, then it will be directly shown in the envoy config dumps. But using files will not, so this api seems to have its purpose.
> 
> No, certificate private keys are sent from Secret to envoy and obscured in config dumps, logs, etc.

Send using SDS? Otherwise, how does it work?

wulianglongrd:
friendly ping @louiscryan @costinm @aryan16
 
Could you please take a look and provide some suggestions for improvement?

costinm:
How about just allowing jwksURI to use a file:// prefix ? I.e. just doc change in this PR.

wulianglongrd:
> How about just allowing jwksURI to use a file:// prefix ? I.e. just doc change in this PR.

I think it's good.

wulianglongrd:
Close, it looks like implementing it by extending jwksURI is possible.

vr:
@wulianglongrd How do you implement this by extending jwksURI ? i have just tried and i am blocked because jwksURI don't want file:// scheme

PR #3179, opened 2024-05-07
sort JWTRule after RequestAuthentication in the Istio docs.

craigbox:


howardjohn (approved):
Maybe outside of scope, but on the subject of docs here.. Jwtrule should not be its own page. Its something we only do for Jwtrule. JwtRule is basically just the spec of Request authentication

craigbox:
I agree but I didn't get as far as figuring out how to fix that 😁

hzxuzhonghu (approved):

PR #3186, opened 2024-05-09
Allow defining CRDs from a single version

howardjohn:
Part of https://github.com/istio/api/issues/3127. Goes with a
corresponding tools change; this will fail until that merges.

This just shows DR. The tool will support both the new and old way (we
can remove the old way if we want), so we don't have to move everything
at once. We will, though. I kept it to one so its easy to review first.

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -129,7 +129,7 @@ option go_package = "istio.io/api/networking/v1alpha3";
 //
 // <!-- crd generation tags
 // +cue-gen:DestinationRule:groupName:networking.istio.io
-// +cue-gen:DestinationRule:version:v1alpha3
+// +cue-gen:DestinationRule:versions:v1beta1,v1alpha3,v1
 // +cue-gen:DestinationRule:annotations:helm.sh/resource-policy=keep
 // +cue-gen:DestinationRule:labels:app=istio-pilot,chart=istio,heritage=Tiller,release=istio
 // +cue-gen:DestinationRule:subresource:status
--- a/extensions/v1alpha1/wasm.pb.go
+++ b/extensions/v1alpha1/wasm.pb.go
@@ -517,7 +517,7 @@ func (FailStrategy) EnumDescriptor() ([]byte, []int) {
 //
 // <!-- crd generation tags
 // +cue-gen:WasmPlugin:groupName:extensions.istio.io
-// +cue-gen:WasmPlugin:version:v1alpha1
+// +cue-gen:WasmPlugin:versions:v1alpha1
 // +cue-gen:WasmPlugin:storageVersion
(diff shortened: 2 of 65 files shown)


whitneygriffith (approved):
LGTM! I like that you focused on removing the CRD generation from each version for now. 

Once this is merged, do we want to wait on anything in particular before we remove the repeated protos and the old way of generation?

whitneygriffith:
Also does it make sense to keep a central source of each API and all its API version? And maybe even template the versions value in the protos based on the central source of APIs and their versions.

howardjohn:
> Also does it make sense to keep a central source of each API and all its API version? And maybe even template the versions value in the protos based on the central source of APIs and their versions.

Seems reasonable, or could be the other way and derived from the proto as the source of truth.



> Once this is merged, do we want to wait on anything in particular before we remove the repeated protos and the old way of generation?

I don't think in terms of time. I have it implemented in https://github.com/istio/api/pull/3188. I split it out for reviewing sake

howardjohn:
Discussed in TOC today; there were no objections that came up there. Feel free to comment here if you were not there/disagree/changed your mind/etc

costinm:
For my curiosity: why don't we use the client-go ( which I think is generated from the CRD) in the pilot code ? 
If I understand the code correctly, for k8s gateway we use their generated client, and most of k8s ecosystem does the same. If there are problems parsing in client-go - anyone else using  that package would be impacted, I can't see how our proto translations help. There is the (uncommon) case where we send the proto over XDS - but I suspect it's a far less used code path and if needed we can convert to proto only when that feature is used, before sending over XDS. 

The 'source of truth' is likely the CRD in K8S. 

+1 on the PR, I can see the tradeoffs and benefits - but I still think all the proto use is too complex and not worth it in this case.

linsun (approved):
LGTM

I don't see any objections from other reviewers either.

networking/v1alpha3/destination_rule.proto
```diff
@@ -129,7 +129,7 @@ option go_package = "istio.io/api/networking/v1alpha3";
 //
 // <!-- crd generation tags
 // +cue-gen:DestinationRule:groupName:networking.istio.io
-// +cue-gen:DestinationRule:version:v1alpha3
+// +cue-gen:DestinationRule:versions:v1beta1,v1alpha3,v1
```
keithmattix, line 132: To move v1 as the storage version (maybe in 3 releases or so), do we just change the order?

keithmattix (approved):
This LGTM - all of the CRDs are the same anyway, so this better reflects the current state

networking/v1alpha3/destination_rule.proto
howardjohn: Yep

howardjohn: TBH we might need something more complex, but its fairly straightforward to tweak the tooling if that comes

PR #3192, opened 2024-05-14
Show the latest API Version in Examples

whitneygriffith:
Part of #3127
Resolves [istio.io/14992](https://github.com/istio/istio.io/issues/14992)

Depends on [tools/2921]( https://github.com/istio/tools/pull/2921)
Builds on #3188, I will rebase once its merged

### DIFF
--- a/networking/v1alpha3/destination_rule.pb.go
+++ b/networking/v1alpha3/destination_rule.pb.go
@@ -32,7 +32,7 @@
 // ratings service would look as follows:
 //
 // ```yaml
-// apiVersion: networking.istio.io/v1beta1
+// apiVersion: networking.istio.io/v1
 // kind: DestinationRule
 // metadata:
 //   name: bookinfo-ratings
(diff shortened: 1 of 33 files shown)


whitneygriffith:
I noticed there are aliases based on the version. For example, `// $aliases: [/docs/reference/config/networking/v1alpha3/destination-rule]`. How is this used? Should we remove or update this?

howardjohn (approved):

whitneygriffith:
@ericvn PTAL

ericvn (approved):

ericvn:
> I noticed in the protos there are aliases based on the version. For example, `// $aliases: [/docs/reference/config/networking/v1alpha3/destination-rule]`. How is this used? Should we remove or update this?

If I understand your question, this is simply telling Hugo that there is another address that can point to this page (it's typically used when moving or renaming a page). ie, `/docs/reference/config/networking/v1alpha3/destination-rule` and `/docs/reference/config/networking/v1beta1`/destination-rule` could be resolved and point to a `/docs/reference/config/networking/v1/destination-rule` page.

PR #3193, opened 2024-05-14
proxy config: clarify concurrency docs

howardjohn:
Fixes https://github.com/istio/api/issues/1783

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -456,9 +456,9 @@ message ProxyConfig {
   int32 stat_name_length = 15;
 
   // The number of worker threads to run.
-  // If unset, this will be automatically determined based on CPU requests/limits.
-  // If set to 0, all cores on the machine will be used.
-  // Default is 2 worker threads.
+  // If unset, which is recommended, this will be automatically determined based on CPU requests/limits.
+  // If set to 0, all cores on the machine will be used, ignoring CPU requests or limits. This can lead to major performance
+  // issues if CPU limits are also set.
   google.protobuf.Int32Value concurrency = 16;
 
   // Path to the proxy bootstrap template file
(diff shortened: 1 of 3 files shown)


mesh/v1alpha1/proxy.proto
```diff
@@ -456,9 +456,9 @@ message ProxyConfig {
   int32 stat_name_length = 15;
 
   // The number of worker threads to run.
-  // If unset, this will be automatically determined based on CPU requests/limits.
-  // If set to 0, all cores on the machine will be used.
-  // Default is 2 worker threads.
```
craigbox, line 461: Do we want to lose this fact?
#1783 suggests it's still true

howardjohn: It is sorta true sort of not... https://github.com/istio/api/issues/1783#issuecomment-2111262691.

mesh/v1alpha1/proxy.proto
```diff
@@ -456,9 +456,9 @@ message ProxyConfig {
...
+  // If unset, which is recommended, this will be automatically determined based on CPU requests/limits.
```
hzxuzhonghu, line 459: Based on pod cpu

zirain (approved):

ericvn (approved):

mesh/v1alpha1/proxy.proto
rehevkor5: > based on CPU requests/limits

How, though?

rehevkor5: Based on https://github.com/istio/istio/blob/af58e9e09598f8181ae217ea886c747a288f0a2c/pilot/cmd/pilot-agent/config/config.go#L64 it looks like CPU Request doesn't have anything to do with it. If CPU Limit is set, it looks like they take the ceiling (round up) of the CPU Limit and use that as the concurrency value. It's unclear what happens if CPU Limit is not set.

howardjohn: If its not set it uses 1 thread per core on the machine

PR #3199, opened 2024-05-15
Adds Service Type to PolicyTargetReference API Docs

danehans:
Previously, only a Gateway resource was defined as a supported attachment type. This PR updates the API docs to include a Service as a supported type and also fixes an incorrect link to Gateway API documentation.

### DIFF
--- a/extensions/v1alpha1/wasm.proto
+++ b/extensions/v1alpha1/wasm.proto
@@ -250,12 +250,13 @@ message WasmPlugin {
   // $hide_from_docs
   istio.type.v1beta1.PolicyTargetReference targetRef = 15;
 
-  // Optional. The targetRef specifies the gateway the policy should be
-  // applied to. The targeted resource specified will determine which
-  // workloads the policy applies to.
+  // Optional. The targetRefs specifies a list of resources the policy should be
+  // applied to. The targeted resources specified will determine which workloads
+  // the policy applies to.
   //
   // Currently, the following resource attachment types are supported:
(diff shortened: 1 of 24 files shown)


danehans:
cc: @howardjohn @linsun

extensions/v1alpha1/wasm.proto
```diff
@@ -250,12 +250,13 @@ message WasmPlugin {
...
-  // Optional. The targetRef specifies the gateway the policy should be
-  // applied to. The targeted resource specified will determine which
-  // workloads the policy applies to.
+  // Optional. The targetRefs specifies a list of resources the policy should be
+  // applied to. The targeted resources specified will determine which workloads
+  // the policy applies to.
   //
   // Currently, the following resource attachment types are supported:
   // * `kind: Gateway` with `group: gateway.networking.k8s.io` in the same namespace.
+  // * `kind: Service` with `""` in the same namespace.
```
howardjohn, line 259: We might want to indicate this applies only for waypoints (not ingress/egress)?

howardjohn (commented):
LGTM!

extensions/v1alpha1/wasm.proto
danehans: Good point @howardjohn. I'll push an update shortly to provide additional clarity. Thanks for the quick review.

howardjohn (approved):

linsun (approved):
Thanks @danehans!!

linsun:
related: https://github.com/istio/istio/pull/51018

PR #3204, opened 2024-05-18
update authz comment

zirain:
Stackdriver is removed from OSS.


kfaseela (approved):

ericvn (approved):

howardjohn (commented):
Do we want to say "There is no plugin supporting this"? 

Which is weird... do we want to hide the docs?

Issue #3215, opened 2024-05-23
Support agression and min_weight_percent in DestinationRule


frgaudet:
(This is used to request new product features, please visit <https://discuss.istio.io> for questions on using Istio)

**Describe the feature request**
We are using Java micro-services, which usually needs a warmup phase to have the best performance. After testing `warmupDurationSecs` on production it was not concluant, because of the `min_weight_percent` that is by default 10%. It is already too much for our service.

Indeed currently DestinationRules only support envoy's `slow_start_window` :

```
spec:
  host:app.svc.cluster.local
  trafficPolicy:
    loadBalancer:
      warmupDurationSecs: 2
 ```

But the API is lacking `aggression` and `min_weight_percent` parameters (cf. [envoy doc](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start)) which are essentials in our context.

**Describe alternatives you've considered**

We considered using EnvoyFilter (cf. https://github.com/istio/istio/issues/41039#issuecomment-1979174285); It provides us good technical results but having so much EnvoyFilter to maintain adds a lot of toil.

**Affected product area (please put an X in all that apply)**

[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[X] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience

**Additional context**

This is a feature I'm ok to implement if this is a something the community really wants.

PR #3216, opened 2024-05-23
feat(DestinationRules): Adding aggression and min_weight_percent to DestinationRules API

frgaudet:
Adding envoy slowStartMode aggression and min_weight_percent parameters to DestinationRules API

Fixes https://github.com/istio/api/issues/3215

Next PR to come on the cluster_traffic_policy side

First time I contribute here, hope this is good :)


networking/v1alpha3/destination_rule.proto
```diff
@@ -513,6 +513,16 @@ message LoadBalancerSettings {
   // amount of requests.
   // Currently this is only supported for ROUND_ROBIN and LEAST_REQUEST load balancers.
   google.protobuf.Duration warmup_duration_secs = 4;
+  
+  // Aggression parameter non-linearly affects endpoint weight and represents the speed of ramp-up. 
+  // By tuning aggression parameter, one could achieve polynomial or exponential speed for traffic increase.
+  message aggression {
+    uint32 default_value = 5;
+    string runtime_key = 6;
```
hzxuzhonghu, line 521: what  do they mean and can you provide an demo

hzxuzhonghu (commented):
Can we wrap all slow star config together

networking/v1alpha3/destination_rule.proto
ramaraochavali: I do not think we should expose a runtime key here. Also for how many services do you have to configure this? And does it differ from service to service?

frgaudet: > I do not think we should expose a runtime key here. Also for how many services do you have to configure this? And does it differ from service to service?

As far as I remember runtime_key parameter is mandatory if we want to use an aggression parameter (which is the one we really need).

frgaudet: > what do they mean and can you provide an demo

We're using Java microservices and pods need a warmup phase in order to have full performance.

 In practice the goal is to avoid giving 100% of the traffic to a new READY pod. Leveraging a slow start allow us to give first a certain % of the traffic then ramp-up progressively to 100%. 

For this first attempt we tried to use the [LoadBalancerSettings](https://istio.io/v1.14/docs/reference/config/networking/destination-rule/#LoadBalancerSettings) feature from Istio. This allow us to specify the duration of the warmup. However in this config we can’t configure 2 important options because they are not exposed by Istio API :

`min_weight_percent` : specifies the initial percent of origin load, if not present, it is default to 10%.

`aggression` : will defined the evolution of the % of traffic sent to the pods from min_weight_percent to 100%, by default the the ramp-up curve is linear, but by customising it we can [achieve exponential type of curve](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start).

The result (sorry I don't have a picture to illustrate that) is that 10% of traffic still too much : our latency increase a lot and impact our users.

To check if the 2 parameters mentioned above impact our traffic, we used an `EnvoyFilter` that we applied to 3 clients of our app.

image

Deploying this config from only a portion of our traffic (roughly 75%) with a `slow_start_window` of 3 minutes and a `min_weight_percent` of 1% we have been able to observe an impact were we can see the progressive ramp-up of the traffic.

frgaudet:
> Can we wrap all slow star config together

Just to be sure to understand your request : you mean wrap all slowStart fields into a new message struct ? What is the best practice you would recommend dealing with such proto change ?

```
  message slowStart {
    google.protobuf.Duration warmup_duration_secs = 1;
    message aggression {
      uint32 default_value = 2;
      string runtime_key = 3;
    }
    uint32 min_weight_percent = 4;
  };
```

networking/v1alpha3/destination_rule.proto
ramaraochavali: I do not think runtime is must https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-runtimedouble

frgaudet: it is actually mandatory if we want to use the `agression` parameter. If I try this EnvoyFilter :

```
kind: EnvoyFilter
metadata:
  name: h2-control
spec:
  configPatches:
    - applyTo: CLUSTER
      match:
        cluster:
          name: "outbound|8080||http-echo.infra.svc.cluster.local"
      patch:
        operation: MERGE
        value:
          name: "outbound|8080||http-echo.infra.svc.cluster.local"
          lbPolicy: LEAST_REQUEST
          leastRequestLbConfig:
            slowStartConfig:
              min_weight_percent: { value: 99 }
              slow_start_window: "12s"
              aggression: { default_value: 2  }
  workloadSelector:
    labels:
      app: landing-f.gaudet
```

Then I have this warning in the logs :

```
landing-f.gaudet istio-proxy {"level":"warning","time":"2024-06-04T04:41:49.406026Z","scope":"envoy config","msg":"gRPC config for type.googleapis.com/envoy.config.cluster.v3.Cluster rejected: Proto constraint validation failed (ClusterValidationError.LeastRequestLbConfig: embedded message failed validation | caused by LeastRequestLbConfigValidationError.SlowStartConfig: embedded message failed validation | caused by SlowStartConfigValidationError.Aggression: embedded message failed validation | caused by RuntimeDoubleValidationError.RuntimeKey: value length must be at least 1 characters):
```
and the config is not applied. However, if I setup the runtime key 

```
          leastRequestLbConfig:
            slowStartConfig:
              min_weight_percent: { value: 99 }
              slow_start_window: "12s"
              aggression: { default_value: 2, runtime_key: "(" }
```
Then the config is successfully applied :
```
istioctl pc cluster landing-f.gaudet.infra --fqdn http-echo.infra.svc.cluster.local -ojson | jq ".[].leastRequestLbConfig"


{
  "slowStartConfig": {
    "slowStartWindow": "12s",
    "aggression": {
      "defaultValue": 2,
      "runtimeKey": "("
    },
    "minWeightPercent": {
      "value": 99
    }
  }
}
```

hzxuzhonghu:
Yes @frgaudet i mean something like this

ramaraochavali:
Do you need separate value of aggression for each service?

frgaudet:
> Do you need separate value of aggression for each service?

Potentially yes, depending on the Java code, the warmup could be tweaked differently

frgaudet:
@hzxuzhonghu @ramaraochavali do you need something else ?

frgaudet:
@howardjohn do you think this could be added in the next release to come ?

networking/v1alpha3/destination_rule.proto
```diff
@@ -514,6 +514,19 @@ message LoadBalancerSettings {
...
+  message aggression {
+    uint32 default_value = 1;
+    string runtime_key = 2;
+  }
+
+  // MinWeightPercent parameter specifies the minimum percent of origin weight to make sure the EDF scheduler has a reasonable deadline, default is 10%.
+  uint32 min_weight_percent = 3;
+  }
+
 }
```
howardjohn, line 531:
```suggestion
  // Deprecated: use `warmup` instead.
  google.protobuf.Duration warmup_duration_secs = 4;
  
  WarmupConfiguration warmup = 5;
}

message WarmupConfiguration {
  google.protobuf.Duration duration = 1;
  // +kubebuilder:validation:Maximum=100
  // +kubebuilder:validation:Minimum=0
  google.protobuf.Double minimum_percent = 2;
  // +kubebuilder:validation:Minimum=0
  google.protobuf.Double aggression = 3;
}
```

On `LoadBalancerSettings` add

```
// +kubebuilder:validation:XValidation:message="only one of warmupDurationSecs or warmup can be set",rule="(has(self.warmupDurationSecs)?1:0)+(has(self.warmup)?1:0)<=1"
```

This fixes a few concerns:
* Move single `warmup_duration_secs` into a nested stuct so we can have all the config here. Deprecate the old setting, and add validation to ensure both are not set
* Hide the aggression complexity w/ runtime key. Make it a float to match Envoy usage
* Make percent a float to match other apis, validate it is between 0 and 100


I left off the docs but we will want to have some description of them, probably linking to envoy docs since 'aggression' is super complex to understand and explain and the envoy page has some nice graphs, etc

networking/v1alpha3/destination_rule.proto
```diff
@@ -377,6 +377,7 @@ message Subset {
 // ```
 //
 message LoadBalancerSettings {
+  // +kubebuilder:validation:XValidation:message="only one of warmupDurationSecs or warmup can be set",rule="(has(self.warmupDurationSecs)?1:0)+(has(self.warmup)?1:0)<=1"
```
howardjohn, line 380: Not a blocker, but would be nice to add a test case like `tests/testdata/workloadentry-invalid.yaml` but for destinationrule. We don't yet have any for DestRule, so certainly not expecting to backfill all of them, but would be good to test the new rules out

networking/v1alpha3/destination_rule.proto
```diff
@@ -514,9 +515,20 @@ message LoadBalancerSettings {
   // endpoints are relatively new like new deployment, this is not very effective as all endpoints end up getting same
   // amount of requests.
   // Currently this is only supported for ROUND_ROBIN and LEAST_REQUEST load balancers.
+  // Deprecated: use `warmup` instead.
   google.protobuf.Duration warmup_duration_secs = 4;
+  
+  WarmupConfiguration warmup = 5;
```
howardjohn, line 521: Can we add some commentary here and on the message on what its for, how to use it, etc?

howardjohn (approved):
Thanks!

frgaudet:
@ramaraochavali or @hzxuzhonghu can you please have a look, your approval is needed to move forward, thanks !

hzxuzhonghu (approved):

PR #3219, opened 2024-05-30
Fix CEL validation on older k8s versions

howardjohn:
Also add testing so we can catch this (though its tied to a single k8s
version, so its not perfect -- but should be able to catch most cases).

CEL before k8s 1.30 did not estimate the cost correctly so its basically
impossible


howardjohn:
I have a better fix in progress (allowing us to keep the rule), happy to wait or merge as-is in the meantime

jwendell:
Will this fix this error?

```
CustomResourceDefinition.apiextensions.k8s.io "peerauthentications.security.istio.io" is invalid: [spec.validation.openAPIV3Schema.properties[spec].properties[selector].properties[matchLabels].x-kubernetes-validations[1].rule: Forbidden: estimated rule cost exceeds budget by factor of more than 100x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
```

howardjohn:
yes

ericvn (approved):

jwendell (approved):

PR #3223, opened 2024-05-31
CEL Validations for ProxyConfig, RequestAuthentication, and PeerAuthentication

howardjohn:
Depends on https://github.com/istio/tools/pull/2947
Manually testing+fuzzing in https://github.com/istio/istio/pull/51287 (manually because it isn't merged yet, it will be automated) in addition to the tests we have in this repo


networking/v1alpha3/destination_rule.proto
```diff
@@ -243,6 +243,7 @@ message TrafficPolicy {
   // settings specified at the destination-level will not be inherited when
   // overridden by port-level settings, i.e. default values will be applied
   // to fields omitted in port-level traffic policies.
+  // +kubebuilder:validation:MaxItems=4096
```
hzxuzhonghu, line 246: how is this value chosen, usually this is too large

type/v1beta1/selector.proto
```diff
@@ -34,6 +34,7 @@ message WorkloadSelector {
   // on which a policy should be applied. The scope of label search is restricted to
   // the configuration namespace in which the resource is present.
   // +kubebuilder:validation:XValidation:message="wildcard not allowed in label key match",rule="self.all(key, !key.contains('*'))"
+  // +kubebuilder:validation:XValidation:message="key must not be empty",rule="self.all(key, key.size() != 0)"
```
zirain, line 37: which version of kubernetes is used for test? I recall there's a issue like a valid cel on 1.28 may failed on 1.29.

howardjohn: We have many levels of testing.

Istio API unit tests: based on `tests/go.mod`, which is 1.29
Istio/Istio unit tests: based on go.mod, which is 1.30
Istio e2e tests in postsubmit: 1.23-1.30


The validity is generally about cost, btw. Each release gets smarter about cost.

networking/v1alpha3/destination_rule.proto
howardjohn: Chosen to be a very large value that is enough to make CEL work. The goal here is not to artificially constrain this too much -- but if we don't do this, k8s assumes it can be infinite and basically doesn't allow you to use CEL since it needs to have a bounded computational cost.

type/v1beta1/selector.proto
zirain: I call out this because I met it before once, glad to hear that istio have test it on different versions.

howardjohn (commented):
This is flaky until https://github.com/istio/tools/pull/2951

kfaseela (approved):

therealmitchconnors (approved):

PR #3227, opened 2024-06-03
Fixed link to the https://github.com/google/re2/wiki/Syntax

eabykov:
Removed `)` in the end of the link https://github.com/google/re2/wiki/Syntax

Fixes: https://github.com/istio/istio.io/pull/15231

PR #3228, opened 2024-06-03
Fixed link to the https://github.com/google/re2/wiki/Syntax

eabykov:
Removed `)` in the end of the link https://github.com/google/re2/wiki/Syntax

Fixes: https://github.com/istio/istio.io/pull/15231

Related to: https://github.com/istio/api/pull/3227

### DIFF
--- a/networking/v1alpha3/virtual_service.proto
+++ b/networking/v1alpha3/virtual_service.proto
@@ -738,7 +738,7 @@ message HTTPMatchRequest {
   //
   // - `prefix: "value"` for prefix-based match
   //
-  // - `regex: "value"` for RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
+  // - `regex: "value"` for RE2 style regex-based match https://github.com/google/re2/wiki/Syntax
   //
   // **Note:** Case-insensitive matching could be enabled via the
   // `ignoreUriCase` flag.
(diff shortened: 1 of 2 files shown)


eabykov:
@kfaseela finally: done

kfaseela:
i think there will be some more files. but anyways

networking/v1alpha3/virtual_service.proto
```diff
@@ -738,7 +738,7 @@ message HTTPMatchRequest {
   //
   // - `prefix: "value"` for prefix-based match
   //
-  // - `regex: "value"` for RE2 style regex-based match (https://github.com/google/re2/wiki/Syntax).
+  // - `regex: "value"` for RE2 style regex-based match https://github.com/google/re2/wiki/Syntax
```
kfaseela, line 741: we should retain the brackets by escaping the character I think

eabykov: If you believe that this is the best course of action, then please proceed. I am tired of attempting to assist you with this broken link (and I now understand why no one has attempted this before me).

kfaseela, line 741:
```suggestion
  // - `regex: "value"` for [RE2 style regex-based match](https://github.com/google/re2/wiki/Syntax)
```

kfaseela: If you would like to proceed with the PR, you can do like above at all occurrences of this link, and then do a `make gen` to generate all the required files including the html. Then only the gen check will pass on the PR. But if you do not want to do it, let me know, I can push a separate PR

eabykov:
thx

PR #3230, opened 2024-06-04
fix gen check

kfaseela:
make gen is generating this additional change, not sure how it was missed out


kfaseela:
cc @howardjohn

kfaseela:
already covered by https://github.com/istio/api/pull/3223/files

PR #3234, opened 2024-06-07
Add DNSDomain as a proxyConfig option, used to be values.yaml only

costinm:
See the comments - most of our configuration is now in MeshConfig/ProxyConfig or env variables - this currently is configurable via values.yaml and gets set via '--domain' flag in Istiod and agent. 

This is called 'clusterDomain' in kubelet and dnsDomain in kubeadmin - I used the later as name, don't mind using the first but it would be best to be one of the 2 to reduce confusion. 

It is a very old setting - only existing comment is "# CAUTION: It is important to ensure that all Istio helm charts specify the same clusterDomain value"

Use of K8S with a different suffix is rare, but some vendors and kubeadmin allows this to be set.
Despite the age - in Istio 1.0 was called proxyDomain, 1.1 changed it to clusterDomain - this is a rarely used option and test coverage is limited, but it is required if K8S is set to use a custom domain instead of cluster.local, or the generated config will break FQDNs.  

The goal is to improve documentation/testing and allow setting this via MeshConfig for consistency (and simplify a bit the related code)

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -691,6 +691,26 @@ message ProxyConfig {
     // If IN_MESH, these headers will not be appended to outbound requests from sidecars to services not in-mesh.
     MetadataExchangeHeaders metadata_exchange_headers = 6;
   }
+
+  // dnsDomain is the DNS domain used by Kubernetes Services. Defaults to "cluster.local"
+  //
+  // Must match the 'dnsDomain' setting in kubeadm config and clusterDomain setting in kubelet.
+  // Must match the custom domain set in the external DNS server, if CoreDNS is not used.
+  //
+  // Do not set unless you are using a custom DNS domain for your K8S cluster and K8S
+  // is already set to use this value.
+  //
+  // The helm charts use "default.global.proxy.clusterDomain".
+  //
+  // Istiod uses this value to generate FQDNs for services. Agent uses this value
+  // in setting its node ID.
+  //
+  // See:
+  // https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
+  // https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/#kubeadm-k8s-io-v1beta3-DNS
+  //
+  //  $hide_from_docs
+  string DNSDomain = 40;
 }
 
 message RemoteService {


mesh/v1alpha1/proxy.proto
```diff
@@ -691,6 +691,26 @@ message ProxyConfig {
...
+  //
+  // Istiod uses this value to generate FQDNs for services. Agent uses this value
+  // in setting its node ID.
+  //
+  // See:
+  // https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
+  // https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/#kubeadm-k8s-io-v1beta3-DNS
+  //
+  //  $hide_from_docs
+  string DNSDomain = 40;
```
howardjohn, line 713: This only controls istio-agent (?) but we need it in istiod as well. What does that mean for usage?

costinm:
No, it is used in istiod as well ( --domain ). Both must match.

Not sure what you mean with "What does that mean for usage" - the idea is
that if this is set, it will be used
instead of the old value.

And probably we'll need a bunch of fixes and more tests - because
.cluster.local is hardcoded in
a LOT of places, but need to have the proto updated first.

For context - GKE now allows creating clusters with a custom DNS name, and
OSS K8S had
this feature for a very long time - and so did Istio, we just never tested
it much.

costinm:
Right now it goes to kube/controller/Options.DomainSuffix - and
kube.ServiceHostname is using it, not completely broken.

howardjohn:
proxy config is a per-pod setting so I don't see how it can impact mesh-wide settings in istiod

costinm:
Right now Istiod gets "--domain" flag from
default.global.proxy.clusterDomain, just like the injection template.

Istiod is using quite a few of the 'default' proxy config settings - but
there are very few leftovers of 'helm only configs'
like this one - we normally use MeshConfig and the default ProxyConfig
plus the env variables.

This PR is just making it part of ProxyConfig - we'll obviously keep the
clusterDomain setting (no backward incompatible
changes). It remains configurable via "--domain" CLI as well - but I would
deprecate both.

costinm:
Well, it is arguable if it ever makes sense for a sidecar to use a
different value from Istiod - and I don't think
the code can correctly handle this. There is a use case - having a central
Istiod ( not sure how stable that
mode is ), and each cluster having a different DNS domain. I think we'll
need quite a few more code
changes to get that working - but the API and use case seem legitimate.

hzxuzhonghu:
And there is `TrustDomain` there, we usually set this same as DNSDomain, so what's the difference between them

costinm:
Very different things (unfortunately). Trust domain is mesh wide -
DNSDomain is cluster wide, i.e. each
cluster will have a different zone ( cluster1.prod.example.com ).

At least on GKE the trustdomain is project_id.svc.goog or cluster.local.

In an ideal world - yes, they would be related ( prod.example.com trust
domain, cluster1.prod.example.com
the DNSDomain for that cluster ) - and we could also issue DNS certs that
work with any client.

costinm:
I think there are better solutions - and ProxyConfig is indeed confusing since it's also used for client. We'll probably need multiple domains anyways.

PR #3235, opened 2024-06-11
validation: add CEL validation to workload entry

howardjohn:


zirain (approved):

networking/v1alpha3/service_entry.proto
```diff
@@ -554,6 +554,7 @@ message ServiceEntry {
 
   // One or more endpoints associated with the service. Only one of
   // `endpoints` or `workloadSelector` can be specified.
+  // +kubebuilder:validation:MaxItems=4096
```
hzxuzhonghu, line 557: If we allow 4096 endpoints, then the workloadEntry must exceed the max size limit of a object from etcdm, which is likely 1.5MB

howardjohn: right now we allow infinity. So this is only making it more restrictive, not less

hzxuzhonghu: I know, could we strict limit it.  In real scenario, can limit to 128. If more endpoints associated, we should encourage to use workloadentry separately

howardjohn: If I was building today I would just do not want to break existing users if they have a lot?

hzxuzhonghu (approved):

networking/v1alpha3/service_entry.proto
nrjpoddar, line 557: We should add documentation around it too. I don't expect folks to have so many entries but we should document our limits clearly.

nrjpoddar (commented):
LGTM

linsun (approved):
LGTM

PR #3240, opened 2024-06-14
add annotations

wulianglongrd:
fix https://github.com/istio/istio/issues/51575, move annotations from [istio/istio](https://github.com/istio/istio/blob/f30f63964573f93af0b4fc1acc005ac1ac84abc5/pilot/pkg/config/kube/gateway/model.go#L30) here.

I have not used the Gateway API and am not familiar with the usage scenarios of these annotations. I would be grateful if you could point out any inappropriate descriptions.

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -29,6 +29,7 @@ const (
 	Unknown ResourceTypes = iota
     Any
     AuthorizationPolicy
+    Gateway
     Ingress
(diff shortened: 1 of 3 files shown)


wulianglongrd:
@howardjohn can you please take a look ?

hzxuzhonghu (commented):
These maybe modified later, so would keep it in istio/istio

PR #3241, opened 2024-06-18
remove WIP annotations

zirain:
xref: https://github.com/istio/istio/issues/51611


hzxuzhonghu (approved):
These annotations may be changed

annotation/annotations.yaml
```diff
@@ -537,7 +537,7 @@ annotations:
       (optional). The wildcard character '*' can be used to redirect all outbound traffic.
       An empty list will disable all outbound redirection.
     deprecated: false
-    hidden: false
+    hidden: true
```
howardjohn, line 540: why hidden? this has been implemented and used for a long time

howardjohn: Oh it doesn't have `sidecar` in it...

I would just revert these entirely. I think they were accidentally merged (I was the approver) thinking they were just adding the sidecar ones

zirain: remove them first, we should add something to point out it related to ambient or ztunnel.

howardjohn (approved):
ty

PR #3243, opened 2024-06-20
Fix concurrency docs

timonwong:
Based on istio/istio#43865, concurrency will only be auto determined based on cpu limits (no requests)


howardjohn (approved):

howardjohn:
/ok-to-test

thanks!

zirain (approved):

PR #3244, opened 2024-06-26
New Status proto for ServiceEntry

ilrudie:
Adds a new proto based on IstioStatus to be used in ServiceEntry. This adds an addresses field to support the auto-allocation v2 controller's persistence requirement.


meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+// $title: Istio ServiceEntry Status
+// $description: Status field for Istio ServiceEntry.
+// $location: https://istio.io/docs/reference/config/meta/v1beta1/istio-serviceentry-status.html
+
+package istio.meta.v1alpha1;
+
+option go_package="istio.io/api/meta/v1alpha1";
+
+message IstioServiceEntryStatus {
+  // Current service state of pod.
```
hzxuzhonghu, line 34: ??? what pod

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+  // +optional
+  // +patchMergeKey=type
+  // +patchStrategy=merge
+  repeated analysis.v1alpha1.AnalysisMessageBase validation_messages = 2;
+
+  // Resource Generation to which the Reconciled Condition refers.
+  // When this value is not equal to the object's metadata generation, reconciled condition  calculation for the current
+  // generation is still in progress.  See https://istio.io/latest/docs/reference/config/config-status/ for more info.
+  // +optional
+  int64 observed_generation = 3;
```
hzxuzhonghu, line 51: Not sure the relationship with Generation in K8s Meta

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+
+  // Above this is just a copy of the common IstioStatus since proto cannot embed an anonymous message whole cloth
+
+  // List of addresses which were automatically allocated by istiod
+  // +optional
+  repeated IstioServiceEntryAddress addresses = 10;
+}
+
+// minor abstraction to allow for adding hostnames if relevant
```
hzxuzhonghu, line 60: why adding hostname?

ilrudie: We aren't adding hostname, and at present can't come up with a good reason to. The abstraction is there in case we want to add it at some point in the future (and to make it align better with gtwapi addresses field).

meta/v1alpha1/serviceentry-status.proto
ilrudie: fair question which I can't answer, it is verbatim copied from the existing istiostatus though

https://github.com/istio/api/blob/7c6732ba7766b7188bfaa69bcd244733da7f368e/meta/v1alpha1/status.proto#L29

meta/v1alpha1/serviceentry-status.proto
ilrudie: another verbatim copy from istiostatus

https://github.com/istio/api/blob/7c6732ba7766b7188bfaa69bcd244733da7f368e/meta/v1alpha1/status.proto#L46

meta/v1alpha1/serviceentry-status.proto
ilrudie: For some added context, This was something which was asked for in the WG where we discussed the api change. Costin wanted thins to align pretty close with an existing impl so that things which scrape status for addresses (namely external DNS) don't need to deal with us reinventing the wheel. gtwapi seemed as good as any though its not set in stone

costinm (changes requested):
I'm not sure Istio status - and adding all the baggage and unrelated API - is the right approach.

John had a good point that adding the full K8S Gateway status is too much - but this is a different level of too much.

How about adding just the IP - and a small comment with a point to someone else that uses the same structure. Istio status doesn't have IP, so it's not relevant.

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+  // +optional
+  // +patchMergeKey=type
+  // +patchStrategy=merge
+  repeated IstioCondition conditions = 1;
+
+  // Includes any errors or warnings detected by Istio's analyzers.
+  // +optional
+  // +patchMergeKey=type
+  // +patchStrategy=merge
+  repeated analysis.v1alpha1.AnalysisMessageBase validation_messages = 2;
```
costinm, line 45: If we decide to support validation_message in ServiceEntry - that should be a completely separate PR and discussion. I would rather follow the patterns for status in K8S Gateway - but I don't see how adding an IP is related to adding validation_messages.

meta/v1alpha1/serviceentry-status.proto
costinm: I don't think K8S Gateway is using 'value' as field name - how does the yaml look like ?

costinm: And I think in the WG the conclusion was to not add hostname now, but keep the structure in place so we can add hostname later.

If you look at Service - there are several cases where they had to add hostname ( usually for integrating with specific vendors and LBs), so there is a good chance we'll need it at some point.

meta/v1alpha1/serviceentry-status.proto
howardjohn: Its already there. Not including this would mean removing a field from an existing v1 API.

This is basically a wonky diff since we had a shared message before and now we don't

costinm: I see, so we no longer have a shared status - but we have to keep the old status.

Ok, nothing we can do about this. But at least we can use the same style and keywords for anything we add to match K8S Gateway - since I assume we want all our resources to also be consistent with the new API. 

How did we add the status for the other attached resources ? Can we share that ?

meta/v1alpha1/serviceentry-status.proto
ilrudie: https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.GatewayStatusAddress is what I based this on. It has optional field `type` which I did not include since for the time being the type is known to be ip address and required field `value`. My thought process was if we ever want to support hostname we can decide at that time if we include type to help denote hostname vs ip address. [Here is the golang type](https://github.com/kubernetes-sigs/gateway-api/blob/cb5bf1541fa70f0692aebde8c64bba434cf331b6/apis/v1/gateway_types.go#L610).

ilrudie:
> I'm not sure Istio status - and adding all the baggage and unrelated API - is the right approach.
> 
> John had a good point that adding the full K8S Gateway status is too much - but this is a different level of too much.
> 
> How about adding just the IP - and a small comment with a point to someone else that uses the same structure. Istio status doesn't have IP, so it's not relevant.

I think we should at least include conditions. I was thinking about this some more last night and currently in v1 we have a situation where everything is ephemeral so if something about an SE changes such that we no longer do auto allocation we quietly ignore that se and possibly reshuffle the state of the world 2-tier hashing... With persistence via status we may not wish to immediately remove the assigned IPs though. Perhaps the IPs stay assigned, at least for some time until we garbage collect, but we note in a condition that the IPs will not be used due to reason X. User maybe corrects that and then the IP go back into effect without needing a new assign.

Maybe this need is not sufficient to duplicate all the IstioStatus fields though.

meta/v1alpha1/serviceentry-status.proto
costinm: Ok, got it. 

Thanks for the explanation. the large copy confused me. 

Looks good to me.

costinm (approved):

meta/v1alpha1/serviceentry-status.proto
hzxuzhonghu: In the SE sceanario, We can change to state of the SE

ilrudie: we could do but it would have to be kept consistently slightly different from the status it is based on. Perhaps we just circle back and correct the comment everywhere to be "Istio resource"? I'm not really clear if `pod` is just a typo in the first place

meta/v1alpha1/serviceentry-status.proto
ilrudie: Yeah, I would not call the large copy ideal but as far as I can tell proto supports nothing like embedding so this manual copy is correct for proto. Happy to be proven wrong and learn a better way though.

howardjohn: This LGTM but we need to cleanup the comments; these are use facing API docs.

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+
+// IMPORTANT!!
+// IstioServiceEntryStatus IS LARGELY A COPY OF IstioStatus AND SHOULD BE KEPT IN SYNC
+// IMPORTANT!!
+
+// $title: Istio ServiceEntry Status
+// $description: Status field for Istio ServiceEntry.
+// $location: https://istio.io/docs/reference/config/meta/v1beta1/istio-serviceentry-status.html
+
+package istio.meta.v1alpha1;
```
howardjohn, line 29: should we put this in networking package?

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+
+option go_package="istio.io/api/meta/v1alpha1";
+
+message IstioServiceEntryStatus {
```
howardjohn, line 33: ServiceEntryStatus (drop Istio)

meta/v1alpha1/serviceentry-status.proto
howardjohn: Let's change to SE here and follow up to change the rest

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+  // +patchStrategy=merge
+  repeated analysis.v1alpha1.AnalysisMessageBase validation_messages = 2;
+
+  // Resource Generation to which the Reconciled Condition refers.
+  // When this value is not equal to the object's metadata generation, reconciled condition  calculation for the current
+  // generation is still in progress.  See https://istio.io/latest/docs/reference/config/config-status/ for more info.
+  // +optional
+  int64 observed_generation = 3;
+
+  // Above this is just a copy of the common IstioStatus since proto cannot embed an anonymous message whole cloth
```
howardjohn, line 53: make sure this doesn't show up in generated html

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+
+  // List of addresses which were automatically allocated by istiod
```
howardjohn, line 55:
```suggestion
  // List of addresses which were assigned to this ServiceEntry.
```

meta/v1alpha1/serviceentry-status.proto
```diff
@@ -0,0 +1,65 @@
...
+  // +optional
+  repeated IstioServiceEntryAddress addresses = 10;
+}
+
+// minor abstraction to allow for adding hostnames if relevant
+message IstioServiceEntryAddress{
```
howardjohn, line 61: Drop Istio prefix

howardjohn (commented):
Core lgtm just naming and comment issues

meta/v1alpha1/serviceentry-status.proto
ilrudie: sounds good. thanks for the feedback!

meta/v1alpha1/serviceentry-status.proto
ilrudie: fixed in service entry. could fix `pod` in status in this PR if folks don't mind but didn't want to make it larger than necessary unless folks want it

meta/v1alpha1/serviceentry-status.proto
ilrudie: moved it

meta/v1alpha1/serviceentry-status.proto
ilrudie: I don't see this or the other similar comments in the generated html, I think it is OK unless I misunderstand

howardjohn (approved):

networking/v1alpha3/serviceentry-status.proto
```diff
@@ -0,0 +1,64 @@
...
+
+import "analysis/v1alpha1/message.proto";
+
+import "meta/v1alpha1/status.proto";
+
+// IMPORTANT!!
+// IstioServiceEntryStatus IS LARGELY A COPY OF IstioStatus AND SHOULD BE KEPT IN SYNC
+// IMPORTANT!!
+
+// $title: ServiceEntry Status
```
howardjohn, line 24: nit: maybe we should just put this in servicentry.proto directly. I don't think we want a whole new docs page for it

networking/v1alpha3/serviceentry-status.proto
```diff
@@ -0,0 +1,64 @@
...
+  int64 observed_generation = 3;
+
+  // Above this is just a copy of the common IstioStatus since proto cannot embed an anonymous message whole cloth
+
+  // List of addresses which were assigned to this ServiceEntry.
+  // +optional
+  repeated ServiceEntryAddress addresses = 10;
+}
+
+// minor abstraction to allow for adding hostnames if relevant
```
howardjohn, line 59: Dev comment should not end up in public docs

PR #3265, opened 2024-07-15
add cluster domain to meshConfig

nicole-lihui:
later  needs to read meshconfig.custerDomain from https://github.com/istio/istio/pull/51064


howardjohn (commented):
Previous discussion in https://github.com/istio/api/pull/3234.

This needs some clarity on how this will be implemented. Currently this is read by Istiod and the agent. What is the plan to get these pieces of info there, how will it handle updates, what is the migration path, etc?

If we are adding this it cannot just be a hint to `analyze`, it needs to replace the existing ways to configure it

nicole-lihui:
> Previous discussion in #3234.
> 
> This needs some clarity on how this will be implemented. Currently this is read by Istiod and the agent. What is the plan to get these pieces of info there, how will it handle updates, what is the migration path, etc?
> 
> If we are adding this it cannot just be a hint to `analyze`, it needs to replace the existing ways to configure it

oh, we can reopen this discussion to clarify the usage of the `clusterDomain`

nicole-lihui:
Hi @costinm,What do you think about this?

 I am currently facing an issue with the usage of the cluster DNS domain. I want to define the cluster DNS domain as a mesh-level configuration, not just within `proxyConfig`. 

Here, I also have another thought: in a multi-cluster setup, theoretically, each cluster can have its own separate domain. I am wondering if we need to consider this, similar to how `meshConfig.networks` is structured.

nicole-lihui:
@howardjohn @costinm @hzxuzhonghu  
Do you have any ideas or suggestions for the cluster domain? I need your advice. Thank you 🌹

PR #3270, opened 2024-07-17
Improve OutboundTrafficPolicy docs

howardjohn:
This is a bit of a mess since its documented in 3 places * 2 resources.

Meaningful changes:
* Explain registry only != security
* Remove recommendation to use registry_only, but explain why someone
  might


zirain (approved):

linsun (approved):
LG!

PR #3271, opened 2024-07-18
Allow targetRef to SE

howardjohn:
This adds the ability to use SE as a targetRef in policies. We already have Service, so its natural to work with SE.

Implementation is in https://github.com/istio/istio/pull/52154

### DIFF
--- a/type/v1beta1/selector.proto
+++ b/type/v1beta1/selector.proto
@@ -106,7 +106,7 @@ enum WorkloadMode {
 //         methods: ["POST"]
 //         ports: ["8080"]
 // ```
-// +kubebuilder:validation:XValidation:message="Support kinds are core/Service and gateway.networking.k8s.io/Gateway",rule="[self.group, self.kind] in [['core','Service'], ['','Service'], ['gateway.networking.k8s.io','Gateway']]"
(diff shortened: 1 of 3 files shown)


type/v1beta1/selector.proto
```diff
@@ -106,7 +106,7 @@ enum WorkloadMode {
 //         methods: ["POST"]
 //         ports: ["8080"]
 // ```
-// +kubebuilder:validation:XValidation:message="Support kinds are core/Service and gateway.networking.k8s.io/Gateway",rule="[self.group, self.kind] in [['core','Service'], ['','Service'], ['gateway.networking.k8s.io','Gateway']]"
+// +kubebuilder:validation:XValidation:message="Support kinds are core/Service, networking.istio.io/ServiceEntry, gateway.networking.k8s.io/Gateway",rule="[self.group, self.kind] in [['core','Service'], ['','Service'], ['gateway.networking.k8s.io','Gateway'], ['networking.istio.io','ServiceEntry']]"
```
hzxuzhonghu, line 109: I think this kind of doc should be put on the policy `TargetRefs`

It is a shared structure, not sure any parent api support all these kinds

howardjohn: the selection logic for targetref is all generic, all types that support it support the same ones. You can see in my linked istio/istio PR I only need to change one function.

linsun (approved):
make sense to me to add SE also to PolicyTargetReference.  LGTM overall.

keithmattix (approved):

PR #3274, opened 2024-07-19
docs: clarify target ref to service

stevenctl:
> Service  with `""`

The wording doesn't explicitly mention that it's `group: ""`. This makes it very clear, as well as mentions you can write `"core"` as well. 

https://github.com/istio/istio/blob/d7a9700d5eaa4e9728274b408623670f48deadb5/pilot/pkg/model/policyattachment.go#L133-L140

### DIFF
--- a/security/v1beta1/request_authentication.proto
+++ b/security/v1beta1/request_authentication.proto
@@ -264,7 +264,7 @@ message RequestAuthentication {
   //
   // Currently, the following resource attachment types are supported:
   // * `kind: Gateway` with `group: gateway.networking.k8s.io` in the same namespace.
-  // * `kind: Service` with `""` in the same namespace. This type is only supported for waypoints.
+  // * `kind: Service` with `group: ""` or `group: "core"` in the same namespace. This type is only supported for waypoints.
   //
   // If not set, the policy is applied as defined by the selector.
   // At most one of the selector and targetRefs can be set.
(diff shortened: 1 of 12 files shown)


howardjohn (approved):

security/v1beta1/request_authentication.proto
```diff
@@ -264,7 +264,7 @@ message RequestAuthentication {
   //
   // Currently, the following resource attachment types are supported:
   // * `kind: Gateway` with `group: gateway.networking.k8s.io` in the same namespace.
-  // * `kind: Service` with `""` in the same namespace. This type is only supported for waypoints.
+  // * `kind: Service` with `group: ""` or `group: "core"` in the same namespace. This type is only supported for waypoints.
```
hzxuzhonghu, line 267: where is core defined, in k8s it is "" for legacy reason  https://github.com/kubernetes/api/blob/master/core/v1/register.go#L26

stevenctl: https://github.com/istio/istio/blob/406d465200b618bc2c787c92f057e4d1e53b5321/pkg/config/model.go#L426

I don't know if we want to encourage users to write "core" but it's technically supported. We don't have to document, this PR was more to make sure we menthion `group:`.

linsun (approved):

PR #3275, opened 2024-07-19
Multiversion CEL tests

howardjohn:


howardjohn:
This is a bit painful to maintain but still good for one-off tests at least..

PR #3285, opened 2024-07-30
Allow setting excludeNamespaces

hzxuzhonghu:
For https://github.com/istio/istio/issues/52343


mesh/v1alpha1/config.proto
```diff
@@ -1277,6 +1277,13 @@ message MeshConfig {
   // for additional detail on selector semantics.
   repeated LabelSelector discovery_selectors = 59;
 
+  // A list of namespaces that are skipped processing by istio. The number of entities (including services, pods, and endpoints) 
+  // in these namespaces are all ignorred. It works in collaboration with `discoverySelectors`
+  // A request to the service or pod can succeed or fail depends on the `outboundTrafficPolicy`.
+  // If `outboundTrafficPolicy` is ALLOW_ANY, it can succeed, otherwise it can fail.
+  repeated string exclude_namespaces = 67;
```
ramaraochavali, line 1284: which one takes precedence if a namespace is listed here but is selected by discovery selector?

hzxuzhonghu: will still be excluded

howardjohn: Why do we need another selector on top of discoverySelectors?

hzxuzhonghu: If use discoverySelectors, user should set all the namespaces they want to use. On the contract, it would be much simpler to exclude a small bunch of namespaces

howardjohn: https://preliminary.istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#LabelSelectorRequirement

You can do "metadata.name NotIn [kube-system, istio-system]"

hzxuzhonghu: That sounds ok

PR #3288, opened 2024-08-02
update destaintion rule api regarding locality settings.

incfly:
Signed-off-by: Jianfei Hu <hujianfei258@gmail.com>

https://github.com/istio/istio/issues/47804#issuecomment-2261523250

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -1065,7 +1065,10 @@ message LocalityLoadBalancerSetting{
     // And it should be used together with `OutlierDetection` to detect unhealthy endpoints, otherwise has no effect.
     repeated string failover_priority = 4;
 
-    // enable locality load balancing, this is DestinationRule-level and will override mesh wide settings in entirety.
+    // Enable locality load balancing, default to true.  This is DestinationRule-level and will override mesh wide settings in entirety.
     // e.g. true means that turn on locality load balancing for this DestinationRule no matter what mesh wide settings is.
+    // When this is enabled, the specified load balancer algorithm, such as LEAST_REQUEST, would be evaluted for a group
+    // of endpoints in the same locality. To achieve load balancing across all locality endpoints, you need to explicitly
+    // turn this off.
     google.protobuf.BoolValue enabled = 3;
 }
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -1065,10 +1065,10 @@ message LocalityLoadBalancerSetting{
     // And it should be used together with `OutlierDetection` to detect unhealthy endpoints, otherwise has no effect.
     repeated string failover_priority = 4;
 
-    // Enable locality load balancing, default to true.  This is DestinationRule-level and will override mesh wide settings in entirety.
+    // Enable locality load balancing, this is DestinationRule-level and will override mesh wide settings in entirety.
     // e.g. true means that turn on locality load balancing for this DestinationRule no matter what mesh wide settings is.
     // When this is enabled, the specified load balancer algorithm, such as LEAST_REQUEST, would be evaluted for a group
     // of endpoints in the same locality. To achieve load balancing across all locality endpoints, you need to explicitly
-    // turn this off.
+    // set this to false.
     google.protobuf.BoolValue enabled = 3;
 }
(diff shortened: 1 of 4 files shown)


incfly:
I suspect Istio bot is outdated. https://github.com/istio/community/blob/master/CLA.md says the EasyCLA (from CNCF?) would send me the link if no CLA is found, but the EasyCLA CI is passing.

networking/v1alpha3/destination_rule.proto
```diff
@@ -1065,7 +1065,10 @@ message LocalityLoadBalancerSetting{
     // And it should be used together with `OutlierDetection` to detect unhealthy endpoints, otherwise has no effect.
     repeated string failover_priority = 4;
 
-    // enable locality load balancing, this is DestinationRule-level and will override mesh wide settings in entirety.
+    // Enable locality load balancing, this is DestinationRule-level and will override mesh wide settings in entirety.
     // e.g. true means that turn on locality load balancing for this DestinationRule no matter what mesh wide settings is.
+    // When this is enabled, the specified load balancer algorithm, such as LEAST_REQUEST, would be evaluted for a group
+    // of endpoints in the same locality. To achieve load balancing across all locality endpoints, you need to explicitly
```
hzxuzhonghu, line 1071: Locality lb setting must be alongside with outlier detection. we donot apply it unless outlier detection is also set

incfly: Could you explain what does it mean not apply? 
To us, the issue is not locality lb itself is enabled or not. checked in 1.21 release, without any destinationrule we will have 

```
        "commonLbConfig": {                                                                                                                                                                                       
            "localityWeightedLbConfig": {}                                                                                                                                                                        
        },  
```

in upstream cluster.

When this is present, it would affects the least request load balancer, to make it load balancing within client envoy's locality,  (an envoy beahvior).  see https://github.com/istio/istio/issues/47804#issuecomment-2279157953

hzxuzhonghu: IC,  so should we remove setting `LocalityWeightedLbConfig` in this case

howardjohn: Are you saying with default config it is prefering local...?

Does not seem to reproduce.

```
$ kubectl locality
NAME                       REGION     ZONE
echo-66d88ff694-8wjzn      us
echo-66d88ff694-nxz7p      us
echo-66d88ff694-z5jv5      eu
shell-56bd5dbdbf-fpmjq     us
```

```
$ qexec shell -- curl -s echo/{0..100} | grep Hostname= | count
     33 Hostname=echo-66d88ff694-8wjzn
     34 Hostname=echo-66d88ff694-nxz7p
     34 Hostname=echo-66d88ff694-z5jv5
```

incfly: no. not preferring local, but least request load balancer is not working. for example:

1. make `echo-8wjzn` holds 2 long running connection, only for these two connection.
2. now all echo pods return the results immediately
3. make 100 connection. you will see each echo still returns 1/3 response, evenly distributed. this would be a surprise when the default lb algorithm is least request 

however, if you configure a destination rule to disable the locality settings as I mentioned above, result is expected:

`8wjzn` served much fewer requets (roughly 1/9, but that's another issue) than the rest two.

incfly: some sample (probably not complete) testing application, https://gist.github.com/incfly/c60885a610c9d648de8e1d655ad5b9d3 

do `k exec shell -- curl tcp-service` until 2 connection ends up with the pod with delay introduced.

then in separate terminal, still do a bunch of curl/request, will see 1/3 ends up on delayed pod (stuck no response).

howardjohn: Ahh thanks for the clarification

hzxuzhonghu (approved):

networking/v1alpha3/destination_rule.proto
incfly: @howardjohn thoughts? pending TOC approval.

howardjohn: Sorry, actually thought this was merged and forgot about it. I think with https://github.com/istio/istio/pull/52898 this is no longer true. But maybe we want to document in older release (change is not backported)?

incfly: Cool. It's sufficient to me if the behavior has been fixed! Will close the PR.

PR #3302, opened 2024-08-28
Add CEL validation to SE and WG

howardjohn:
This continues the trend of adding CEL validation by adding WG support and some of SE (it got too much, so wasn't able to complete it)

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -7336,7 +7336,7 @@ spec:
             - message: CIDR addresses are allowed only for NONE/STATIC resolution
                 types
               rule: '!(has(self.addresses) && self.addresses.exists(k, k.contains(''/''))
-                && (self.resolution != ''STATIC'' && self.resolution != ''STATIC''))'
+                && (self.resolution != ''STATIC'' && self.resolution != ''NONE''))'
             - message: NONE mode cannot set endpoints
               rule: '(!has(self.resolution) || self.resolution == ''NONE'') ? !has(self.endpoints)
                 : true'
@@ -7630,7 +7630,7 @@ spec:
             - message: CIDR addresses are allowed only for NONE/STATIC resolution
                 types
               rule: '!(has(self.addresses) && self.addresses.exists(k, k.contains(''/''))
-                && (self.resolution != ''STATIC'' && self.resolution != ''STATIC''))'
+                && (self.resolution != ''STATIC'' && self.resolution != ''NONE''))'
             - message: NONE mode cannot set endpoints
               rule: '(!has(self.resolution) || self.resolution == ''NONE'') ? !has(self.endpoints)
                 : true'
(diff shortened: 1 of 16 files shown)


linsun:
Hi @howardjohn i saw you have don't merge label, just checking if this is ready for review?

howardjohn:
yeah this is

networking/v1alpha3/service_entry.proto
```diff
@@ -436,6 +437,10 @@ option go_package = "istio.io/api/networking/v1alpha3";
 // +k8s:deepcopy-gen=true
 // istiostatus-override: ServiceEntryStatus: istio.io/api/networking/v1alpha3
 // -->
+// +kubebuilder:validation:XValidation:message="only one of WorkloadSelector or Endpoints can be set",rule="(has(self.workloadSelector)?1:0)+(has(self.endpoints)?1:0)<=1"
+// +kubebuilder:validation:XValidation:message="CIDR addresses are allowed only for NONE/STATIC resolution types",rule="!(has(self.addresses) && self.addresses.exists(k, k.contains('/')) && (self.resolution != 'STATIC' && self.resolution != 'STATIC'))"
```
linsun, line 441: You had twp checks on `STATIC`...Should one of the `self.resolution` check on `NONE` instead?

linsun (commented):
lgtm, just one small q.

networking/v1alpha3/service_entry.proto
howardjohn: nice find, will update

linsun (approved):
LGTM

linsun:
@ilrudie or @keithmattix do one of you would like to review also? Thanks!

networking/v1alpha3/service_entry.proto
```diff
@@ -600,19 +613,23 @@ message ServiceEntry {
...
+  // +kubebuilder:validation:MaxLength=256
   string protocol = 2;
 
   // Label assigned to the port.
+  // +kubebuilder:validation:MaxLength=256
   string name = 3 [(google.api.field_behavior) = REQUIRED];
 
   // The port number on the endpoint where the traffic will be
   // received. If unset, default to `number`.
+  // +kubebuilder:validation:XValidation:message="port must be between 1-65535",rule="0 < self && self <= 65535"
```
ilrudie, line 632: Would this have a potential bad interaction with the golang zero value since it is an int?

ilrudie: I suppose not since we have tests which cover leaving target_port empty

kubernetes/customresourcedefinitions.gen.yaml
```diff
@@ -7369,6 +7414,10 @@ spec:
                 type: array
             type: object
             x-kubernetes-preserve-unknown-fields: true
+        required:
+        - spec
+        - spec
+        - spec
```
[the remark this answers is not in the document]
howardjohn: This is probably a bug caused by that

ilrudie (approved):
LGTM

kubernetes/customresourcedefinitions.gen.yaml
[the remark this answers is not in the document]
howardjohn: We should fix but I don't think it has any impact so not blocking. Nice catch

PR #3304, opened 2024-09-05
clarify documentation around stringMatch and case-sensitivity

ilrudie:
seeks to add clarification around case-sensitive vs case-insensitive options for stringMatch in the VirtualService API


linsun (approved):
Thanks Ian!!

ilrudie:
1.22 has the same verbiage, we could consider cherry-picking the doc update for it as well

howardjohn (approved):

PR #3312, opened 2024-09-12
Improve validation for targetRefs

howardjohn:
Per
https://gateway-api.sigs.k8s.io/geps/gep-2648/?h=targetrefs#multiple,
only 16 max allowed -- which is quite reasonable.

Additionally, consistently allow only workloadSelector OR targetRef; we
had this only on some types


hzxuzhonghu (approved):

zirain (approved):

linsun (approved):
lgtm

PR #3319, opened 2024-10-03
Improve documentation for subjectAltNames

zioproto:
The user might be confused about the usage of `subjectAltNames`, trying to use it to add more hostnames to an endpoint.
Make clear that it requires TLS mode mutual, and the the alternative names are those presented by the client for authentication.

Originally proposed as a docs contribution:
https://github.com/istio/istio.io/pull/15733
but the docs are autogenerated from this repo.

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -393,12 +393,12 @@ message ServerTLSSettings {
     PASSTHROUGH = 0;
 
     // Secure connections with standard TLS semantics. In this mode
-    // client certificate is not requested during handshake. 
+    // client certificate is not requested during handshake.
     SIMPLE = 1;
 
     // Secure connections to the downstream using mutual TLS by
     // presenting server certificates for authentication.
-    // A client certificate will also be requested during the handshake and 
+    // A client certificate will also be requested during the handshake and
     // at least one valid certificate is required to be sent by the client.
     MUTUAL = 2;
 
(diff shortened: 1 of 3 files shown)


nshankar13 (approved):

MorrisLaw (approved):
nice!

networking/v1alpha3/gateway.proto
```diff
@@ -473,6 +473,7 @@ message ServerTLSSettings {
 
   // A list of alternate names to verify the subject identity in the
   // certificate presented by the client.
+  // Requires TLS mode to be set to `MUTUAL`.
```
howardjohn, line 476: I think it could also be ISTIO_MUTUAL?

zioproto: I went to check what `ISTIO_MUTUAL` means in the Gateway:
https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings-TLSmode

From the docs `ISTIO_MUTUAL` is to secure connections **from** the downstream. Not sure about "`TLSOptions` should be empty", because I dont see `TLSOptions` in the spec.
Does it mean you can set only `tls.mode` ?

However, the key idea of `ISTIO_MUTUAL` is to use certificates generated automatically by Istio for mTLS. Then why the client should present an alternate name if it would be using a certificate generated by Istio ?

Another comment on the documentation:
The `MUTUAL` is securing connections **to** the downstream ? I am not sure the terminology in the docs is correct here. I thought the `DestinationRule` configured the connections **to** the **upstream**, here we should have again a reference to connections **from** the **downstream**.

For upstream and downstream definitions:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/intro/terminology

zioproto: I will test if an istio ingressgateway configured with `ISTIO_MUTUAL` accepts the `tls.subjectAltNames` configuration and I will report back. The external client outside of the mesh will not have either `MUTUAL` or `ISTIO_MUTUAL`. :)

zioproto: I tested, with `ISTIO_MUTUAL` configured on the Gateway it does not work for me. It seems the Gateway is not using the certificate I expect.

howardjohn (approved):

PR #3323, opened 2024-10-04
add ability to configure xfcc header for inbound connection

ramaraochavali:
As discussed in https://github.com/istio/istio/pull/53337


mesh/v1alpha1/proxy.proto
```diff
@@ -662,10 +662,38 @@ message ProxyConfig {
...
       MetadataExchangeMode mode = 1;
     }
+    message SetCurrentClientCertDetails {
+      // Whether to forward the subject of the client cert. Defaults to true.
+      google.protobuf.BoolValue subject = 1;
+  
+      // Whether to forward the entire client cert in URL encoded PEM format. This will appear in the
+      // XFCC header comma separated from other values with the value Cert="PEM".
+      // Defaults to false.
+      bool cert = 2;
```
howardjohn, line 672: nit: would just make these all BoolValue, gives us more flexibiltiy to change the default in the future

howardjohn: and has better merging semantics

howardjohn (approved):

zirain (approved):

PR #3328, opened 2024-10-09
Add traffic distribution annotation

howardjohn:
Added in https://github.com/istio/istio/pull/53435

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -581,3 +581,21 @@ annotations:
     hidden: true
     resources:
       - Gateway
+
+  - name: networking.istio.io/traffic-distribution
+    featureStatus: Alpha
+    description: |
+      Controls how traffic is distributed across the set of available endpoints.
+      
+      At this time, this annotation only impacts routing done by Ztunnel.
+      
+      Accepted values:
+      * PreferClose: endpoints will be categorized by how "close" they are, consider network, region, zone, and subzone.
+        Traffic will only be sent to the set of closest healthy endpoints.
+        For example, if I have a client in zone `us-west` with endpoints: `us-west,us-west,us-east`, all traffic will go to
+        the two `us-west` backends. If those backends become unhealthy, traffic will send to `us-east`.
+    deprecated: false
+    hidden: false
+    resources:
+      - Service
+      - ServiceEntry
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -590,10 +590,12 @@ annotations:
       At this time, this annotation only impacts routing done by Ztunnel.
       
       Accepted values:
-      * PreferClose: endpoints will be categorized by how "close" they are, consider network, region, zone, and subzone.
-        Traffic will only be sent to the set of closest healthy endpoints.
-        For example, if I have a client in zone `us-west` with endpoints: `us-west,us-west,us-east`, all traffic will go to
-        the two `us-west` backends. If those backends become unhealthy, traffic will send to `us-east`.
+      * `PreferClose`: endpoints will be categorized by how "close" they are, consider network, region, zone, and subzone.
+        Traffic will be prioritized to the closest healthy endpoints.
+        For example, if I have a Service with `PreferClose` set, with endpoints in zones `us-west,us-west,us-east`. When 
+        sending traffic from a client in zone `us-west`, all traffic will go to the two `us-west` backends.
+        If one those backends become unhealthy, all traffic will go to the remaining endpoint in `us-west`.
+        If that backend becomes unhealthy, traffic will sent to `us-east`.
     deprecated: false
     hidden: false
     resources:
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -35,6 +35,7 @@ const (
     Namespace
     Pod
     Service
+    ServiceEntry
     WorkloadEntry
 )
 
@@ -57,6 +58,8 @@ func (r ResourceTypes) String() string {
 	case 8:
 		return "Service"
 	case 9:
+		return "ServiceEntry"
+	case 10:
 		return "WorkloadEntry"
 	}
 	return "Unknown"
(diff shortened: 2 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -581,3 +581,21 @@ annotations:
...
+    featureStatus: Alpha
+    description: |
+      Controls how traffic is distributed across the set of available endpoints.
+      
+      At this time, this annotation only impacts routing done by Ztunnel.
+      
+      Accepted values:
+      * PreferClose: endpoints will be categorized by how "close" they are, consider network, region, zone, and subzone.
+        Traffic will only be sent to the set of closest healthy endpoints.
+        For example, if I have a client in zone `us-west` with endpoints: `us-west,us-west,us-east`, all traffic will go to
```
linsun, line 595: Could you clarify if user should annotate on the client or server side?

howardjohn: This is a service annotation so it can really only be on the server side?

howardjohn: Below it says which resources it applies to

annotation/annotations.yaml
```diff
@@ -581,3 +581,21 @@ annotations:
     hidden: true
     resources:
       - Gateway
+
+  - name: networking.istio.io/traffic-distribution
```
keithmattix, line 585: If users are setting this, should it be a label?

annotation/annotations.yaml
```diff
@@ -581,3 +581,21 @@ annotations:
...
+    featureStatus: Alpha
+    description: |
+      Controls how traffic is distributed across the set of available endpoints.
+      
+      At this time, this annotation only impacts routing done by Ztunnel.
+      
+      Accepted values:
+      * PreferClose: endpoints will be categorized by how "close" they are, consider network, region, zone, and subzone.
+        Traffic will only be sent to the set of closest healthy endpoints.
```
hzxuzhonghu, line 594: `only be sent` seems not correct though below you mentioned the fall back

annotation/annotations.yaml
hzxuzhonghu: make some sense, labels can be used to fiter by user

annotation/annotations.yaml
howardjohn: It's to the "set of closest healthy endpoints". So even if it's the second closest (fail over)the statement seems accurate. 

Happy to reword it since I don't think it's particularly clear, but I don't think it's incorrect

hzxuzhonghu: I mean should we say prioritized sending to those endpoint.

annotation/annotations.yaml
linsun, line 595:
```suggestion
        For example, if I have a client in zone `us-west` with server endpoints: `us-west,us-west,us-east`, all traffic will go to
```

linsun: Is this correct, based on my understanding?

annotation/annotations.yaml
linsun, line 594:
```suggestion
        Traffic will be sent to the set of closest healthy endpoints when possible.
```

linsun: How about this?

annotation/annotations.yaml
linsun: We have labels that start with networking.istio.io (for example gatewayPort); we also have annotations that start with networking.istio.io, for example exportTo.

howardjohn: I have a design to set a guidance for this: https://docs.google.com/document/d/1q27KAgIzkmqDMWI1nTMOdlBvY-dRyGTldxe9Yu3pVq0/edit.

(the doc suggests this should be an annotation)

bleggett: I've added an alternate proposal to that doc, because I prefer a criteria that's based on end-user usage and not development internals. I would suggest everyone look at that and add anything that's been missed.

This isn't a technical question at all really. It's a simple type. It could quite comfortably be either a label or an annotation, even under the proposal doc in its current state. If you think the tiebreak should be annotations, it's an annotation. If you think the tiebreak should be labels, it's a label.

Which is kinda why I'd rather go with what is maximally logical and simple to the end user.

howardjohn: In the TOC meeting on monday there was some agreement to go with the proposal in the doc which would make this an annotation

annotation/annotations.yaml
```diff
@@ -581,3 +581,23 @@ annotations:
     hidden: true
     resources:
       - Gateway
+
+  - name: networking.istio.io/traffic-distribution
```
zirain, line 585: should we use same naming parttern like others, use `networking.istio.io/trafficDistribution`?

annotation/annotations.yaml
```diff
@@ -581,3 +581,23 @@ annotations:
...
+      Accepted values:
+      * `PreferClose`: endpoints will be categorized by how "close" they are, consider network, region, zone, and subzone.
+        Traffic will be prioritized to the closest healthy endpoints.
+        For example, if I have a Service with `PreferClose` set, with endpoints in zones `us-west,us-west,us-east`. When 
+        sending traffic from a client in zone `us-west`, all traffic will go to the two `us-west` backends.
+        If one those backends become unhealthy, all traffic will go to the remaining endpoint in `us-west`.
+        If that backend becomes unhealthy, traffic will sent to `us-east`.
+    deprecated: false
+    hidden: false
+    resources:
```
MorrisLaw, line 601: I don't have the full context but I am curious why we do not include `Workload` and `WorkloadEntry` for this traffic distribution PreferClose feature too?

howardjohn: It's a service concept. A workload entry is 1 single IP. there is no load balancing or endpoint selection. Service has many IPs we need to pick from

annotation/annotations.yaml
howardjohn: It seems we are split on which we use, with most new ones using `-` instead of the casing.

zirain: we should make a clear rule on this, we just introuduce some new casing annotations IIRC.

annotation/annotations.yaml
linsun: Thanks for that context, i caught up with the meeting mins and doc, SGTM

linsun (approved):
LGTM, don't see outstanding issue after the TOC discussion on Monday.

hzxuzhonghu (approved):

PR #3333, opened 2024-10-10
Validation: add documentation and use CEL pre-processor

howardjohn:
Depends on https://github.com/istio/tools/pull/3050.

This PR adds support for a new cel pre-processor: https://github.com/howardjohn/celpp.

This avoids the need to write incredibly verbose and error prone boilerplate to workaround limitations in the CEL language.

----

This PR has a few commits:
* Run with no changes. The pre-processor formats the CEL expression, so this has some trivial changes
* Incrementally migrate things to the new CEL macros where it makes sense to simplify
* Add documentation for validation (includes more than just the new stuff)

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -19,6 +19,7 @@ followed for Istio APIs.
 - [Proto Guidelines](#proto-guidelines)
     - [Style](#style)
     - [Basic Proto Versioning](#basic-proto-ersioning)
+- [Validation Guidelines](#validation-guidelines)
 - [CRD Guidelines](#crd-guidelines)
     - [Style](#crd-style)
     - [Basic CRD Versioning](#basic-crd-versioning)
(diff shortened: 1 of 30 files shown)


GUIDELINES.md
```diff
@@ -214,6 +215,75 @@ protos.
...
+Kubernetes changes version-to-version on how it estimates cost (usually getting more lenient) and what functions are available.
+We want to target the oldest version for compatibility purposes.
+Our tests do not currently cover this (a prototype of doing so can be found [here](https://github.com/istio/api/pull/3275)).
+A list of what features are in which versions can be found [here](https://kubernetes.io/docs/reference/using-api/cel/#cel-options-language-features-and-libraries).
+
+Istio has some custom macros that are expanded at compile time, driven by the [celpp](https://github.com/howardjohn/celpp) package.
+This extends CEL with these capabilities:
+* **default**. Usage: `default(self.x, 'DEF')`.
+* **oneof**. Usage: `oneof(self.x, self.y, self.z)`. This checks that 0 or 1 of these fields is set.
+* **index**. Usage: `self.index({}, x, z, b)`. This does `self.x.z.b` and returns `{}` if any of these is not set.
```
ilrudie, line 270: Is the difference 'DEF' must be a concrete value whereas {} could be an expression of some sort?

howardjohn: Both `'DEF'` and `{}` are just examples of the same thing -- the default value. In one example I return a string and the other an empty object, but there is not a difference.

networking/v1alpha3/service_entry.proto
```diff
@@ -437,10 +437,10 @@ option go_package = "istio.io/api/networking/v1alpha3";
...
-// +kubebuilder:validation:XValidation:message="NONE mode cannot set endpoints",rule="(!has(self.resolution) || self.resolution == 'NONE') ? !has(self.endpoints) : true"
-// +kubebuilder:validation:XValidation:message="DNS_ROUND_ROBIN mode cannot have multiple endpoints",rule="(has(self.resolution) && self.resolution == 'DNS_ROUND_ROBIN') ? (!has(self.endpoints) || size(self.endpoints) == 1) : true"
+// +kubebuilder:validation:XValidation:message="only one of WorkloadSelector or Endpoints can be set",rule="oneof(self.workloadSelector, self.endpoints)"
+// +kubebuilder:validation:XValidation:message="CIDR addresses are allowed only for NONE/STATIC resolution types",rule="!(default(self.addresses, []).exists(k, k.contains('/')) && !(default(self.resolution, 'NONE') in ['STATIC', 'NONE']))"
+// +kubebuilder:validation:XValidation:message="NONE mode cannot set endpoints",rule="default(self.resolution, 'NONE') == 'NONE' ? !has(self.endpoints) : true"
+// +kubebuilder:validation:XValidation:message="DNS_ROUND_ROBIN mode cannot have multiple endpoints",rule="default(self.resolution, '') == 'DNS_ROUND_ROBIN' ? default(self.endpoints, []).size() <= 1 : true"
```
ilrudie, line 443: **nit** it really doesn't matter for this logic but just looks a bit odd that we default to something which isn't actually the default 'NONE'

ilrudie (commented):
This really improves readability. It looks good to me

linsun (approved):
lgtm

ilrudie (approved):

PR #3335, opened 2024-10-14
Add ObservedGeneration field to IstioCondition

tjons:
This PR adds the `ObservedGeneration` field to the `IstioCondition` type, and marks it as deprecated in the `IstioStatus` type. Will be used by https://github.com/istio/istio/pull/53522/.

cc @howardjohn


meta/v1alpha1/status.proto
```diff
@@ -70,4 +72,9 @@ message IstioCondition {
   // Human-readable message indicating details about last transition.
   // +optional
   string message = 6;
+
+  // Resource Generation to which the Condition refers.
+  // +optional
+  // +protoc-gen-crd:validation:XIntOrString
```
howardjohn, line 78: I think we can drop this - it was only needed because an older version of the old field was incorrectly setting it to a string.

howardjohn (commented):
LGTM besides my one note

meta/v1alpha1/status.proto
tjons: @howardjohn I kept this as an `int64` for parity with the above, which will be encoded to a string in proto-JSON conversion: https://protobuf.dev/programming-guides/json/. we can change this to an `int32` and drop the validation, or keep `int64` and the validation hint. 

Personally, I'd recommend keeping it as is, because the `generation` type in k8s is an `int64`: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L177

howardjohn: Ahh... annoying. Ok, thanks

ilrudie (approved):
lgtm

PR #3339, opened 2024-10-18
Add clear_route_cache parameter to ExtAuthZ extensions

bvandewalle:
API change for https://github.com/istio/istio/issues/53555

- Adding a `clear_route_cache` proto parameter for both `EnvoyExternalAuthorizationHttpProvider` and `EnvoyExternalAuthorizationGrpcProvider`

API change used here: https://github.com/istio/istio/pull/53581


zirain (approved):
I'm not good at naming, but the idea LGTM.

bvandewalle:
Thanks @zirain . The name for this field is the same as the one used by Envoy on the ExtAuthZ proto:
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto

howardjohn (approved):

PR #3340, opened 2024-10-18
AuthorizationPolicy: add `serviceAccounts` field

howardjohn:
This is a minor implementation complexity in favor of a dramatic
simplification to usage of Istio authorization.

Today, if a user wants to dive into zero-trust 101, they are presented
with a requirement to set `principals`: `A list of peer identities
derived from the peer certificate`, and write
`<TRUST_DOMAIN>/ns/<NAMESPACE>/sa/<SERVICE_ACCOUNT>`.

This simple sentance is a huge cognitive overload for users in my
experience working with users, and unnecesarily pushes SPIFFE, trust
domains, and other unneccesary concepts onto users. Additionally, the
requirement to set 'trust domain', which is overwhelmingly not desired
by users who just want SA auth, leads to all sorts of wonky workarounds
in Istio like `cluster.local` being a magic value.

Instead, we just add a SA field directly. This takes the format `ns/sa`,
as you cannot safely reference a SA without a namespace field as well.
Note we do this, rather than just require you to set 'service account' and 'namespace'
as individual fields, since you could have `namespace=[a,b],sa=[d,e]`
which is ambiguous.

If this is directionally approved, I will add some more documentation
and CEL validation and testing.


security/v1beta1/authorization_policy.proto
```diff
@@ -428,6 +428,8 @@ message Source {
   // `"<TRUST_DOMAIN>/ns/<NAMESPACE>/sa/<SERVICE_ACCOUNT>"`, for example, `"cluster.local/ns/default/sa/productpage"`.
   // This field requires mTLS enabled and is the same as the `source.principal` attribute.
   //
+  // Usage of `serviceAccounts` is typically simpler and offers the same functionality.
```
hzxuzhonghu, line 431: nope, if this is a client from outside, sa is not known, in this case principals is still needed.

hzxuzhonghu (commented):
And you should document only one field can be specified

security/v1beta1/authorization_policy.proto
howardjohn: I don't get it. If I know to write `spiffe://cluster.local/ns/foo/sa/bar` then surely I can know to write `foo/bar`?

hzxuzhonghu: If the client is from external, the identity could be any format, nit limited to `spiffe://cluster.local/ns/foo/sa/bar`

howardjohn: In that case the user would not use this field then.

howardjohn: It's not deprecating `principals`,just making the 99.9999% use case easier

ilrudie:
> And you should document only one field can be specified

Is this decided for sure? It seems to me that mixing and matching is plausible even if likely not recommended and more error prone.

bleggett:
> And you should document only one field can be specified

You should be able to document as many fields as you want in the AuthPolicy - if they can be matched against the identity, we will match them.

security/v1beta1/authorization_policy.proto
bleggett: It's not even for complex scenarios - hardcoding principals is one reason we can't currently _support_ complex  scenarios at all (custom SPIFFE IDs, SPIRE etc) - so we should just say that:

```suggestion
  // Usage of `serviceAccounts` is typically simpler and offers the same functionality. Principals are still fully supported, it is not recommended to encode complete principal strings, as this leads to fragile policies.
```

security/v1beta1/authorization_policy.proto
```diff
@@ -453,6 +455,17 @@ message Source {
   // Optional. A list of negative match of namespaces.
   repeated string not_namespaces = 7;
 
+  // Optional. A list of service accounts derived from the peer certificate.
+  // This field requires mTLS enabled and is the same as the `source.serviceaccount` attribute.
+  //
+  // This takes the format `<namespace>/<serviceaccount>`.
+  //
+  // If not set, any service account is allowed.
```
bleggett, line 463:
```suggestion
  // If not set, any service account is allowed.
  // if both principal and this field are set, this field has precedence
```

If we are going to let people set both, we need to establish whether it's an AND or an OR match, and what the precedence is if both are set.

security/v1beta1/authorization_policy.proto
```diff
@@ -453,6 +455,17 @@ message Source {
...
+  repeated string service_accounts = 11;
+
+  // Optional. A list of negative match of service accounts.
```
bleggett, line 466: If we are going to introduce matching semantics, we should probably separate the boolean logic from the match strings - this feels a bit weird to me and doesn't scale particularly well if we ever add more things.


e.g. 

```
match:
  serviceAccount: foo
---
not-match:
  serviceAccount: foo
```
versus what you have here:

```
service_account: foo
---
not_service_account: foo
```

howardjohn: I agree but its a bit too late to change, we already do this with 5 fields, it would be quite odd for a 6th to work entirely differently.

bleggett: Alternatively, just drop `not_service_account` - I am not sure we really need it to replicate current AuthPolicy functionality around principal matching.

howardjohn: Every single field in auth policy has a `not` variant.. I don't really get why we would change that now?

I agree its fairly silly, as is "notMethod" and many others... but consistency silly is better than not

bleggett: > Every single field in auth policy has a not variant.. I don't really get why we would change that now?

Yeah - I didn't realize we were already doing this for everything else, so ignore.

security/v1beta1/authorization_policy.proto
ilrudie: I think setting both should be allowed. Presumably internally we can normalize the types and then can just append one list to the other, dedupe and move on so neither takes precedence.

bleggett: > I think setting both should be allowed. Presumably internally we can normalize the types and then can just append one list to the other, dedupe and move on so neither takes precedence.

The problem is I don't think you can actually do that/it would make the existing problem worse to do that.

Given an AuthPol

```
principal: spiffe://example.org/ns/default/sa/my-sa
service_account: default/my-sa
```

How do I evaluate the AuthPol if the presented workload principal is actually (best  case)

```
spiffe://example.org/ns/default/sa/my-sa/some/other/stuff
```

or (worst case)
```
spiffe://example.org/beep/boop/ns/default/sa/my-sa
```

?

ilrudie: maybe I misunderstand but my guess was internally we would use this as a shorthand for a spiffe ID in the istio format making the conversion from ns/sa to spiffe pretty straight forward. I do agree that spiffe -> ns/sa presumes all spiffe IDs are in the istio format which we likely don't want but spiffe -> ns/sa is lossy so we probably don't want to do that conversion even if we were ok mandating the istio format

ilrudie: I think the use case for a mix and match would, in practice, be limited. If you want to require some advanced ID format which includes `some/other/stuff` then I don't think relying on ns/sa is going to work for you really at all and in that case you should NOT try to specify things that way. I just don't think our API can really track the user's intent in that way. If this is your scenario then you probably need additional policy to enforce it

bleggett: > internally we would use this as a shorthand for a spiffe ID in the istio format making the conversion from ns/sa to spiffe pretty straight forward.

If we are going to change the API to allow people to stop specifying the Istio format, why would we underdesign the API to maintain an implicit relationship between the Istio SPIFFE ID format we want to stop surfacing, and the thing which has nothing to do with SPIFFE IDs or the particular Istio variant of them?

> I do agree that spiffe -> ns/sa presumes all spiffe IDs are in the istio format which we likely don't want but spiffe -> ns/sa is lossy

It's really not lossy - the ns/sa is all Istio uses out of the current Istio-only SPIFFE ID format anyway. The trust domain may be checked in a handful of places - and if that's something we need, it seems pretty trivial to add an optional `trust_domain` field to match on, right next to the SA we're adding here?

What we are doing here is trying to move people writing AuthPolicies from the Old Way (raw/fixed principal - which is effectively always wrong and is the primary reason we're stuck forcing people to use a local format, and forcing people writing AuthPolicies to worry about what a SPIFFE ID is) to the New Way (just use SA/TD because that's all Istio cares about anyway, and we match those against the principal opaquely).

bleggett: > I think the use case for a mix and match would, in practice, be limited. If you want to require some advanced ID format which includes `some/other/stuff` then I don't think relying on ns/sa is going to work for you really at all and in that case you should NOT try to specify things that way. I just don't think our API can really track the user's intent in that way. If this is your scenario then you probably need additional policy to enforce it

Sure - and I can have additional policy to enforce it outside of Istio _if_ Istio doesn't implicitly forbid me from using a non-Istio format - which is the problem today, it does. And this is where it forbids that. Needlessly. There's no benefit to its doing so, especially not with this change - we don't have any reason to require a specific fixed format (or assume it, just so we can have lax combinatorics rules).

AuthPolicy - and implentations of it - can only assume that a given principal/identity (whatever the format) has AT-LEAST the fields supported in the AuthPolicy. Not AT-MOST, which is the current state, and what would be implicitly required if we supported both in an AND clause here.

Is there a concrete use case for the `AND` clause that I'm missing? That seems even more unlikely to be useful then hitting problems because Istio's APIs make bad assumptions about the principal format, forcing you to stick to its own.

Also - forget SPIFFE - if we need to add other kinds of identifiers in the auth policy beyond ns/sa in the future because we adopt some other principal format - those are easy to append _if_ the API has clear semantics. Those are not easy to append if we design the API semantics around an implicit assumption that Istio's SPIFFE ID format is the only one people can use here.

ilrudie: Maybe I just fully misunderstand this. What do we expect the SANs in our certs will look like if not a spiffe ID?

bleggett: > Maybe I just fully misunderstand this. What do we expect the SANs in our certs will look like if not a spiffe ID?

We _should_ expect it will look like a SPIFFE ID (it follows the spec, has a trust domain, and some fields we care about).

We _should not_ require it to look like an _Istio_ SPIFFE ID (it follows the spec, has a trust domain, and ONLY contains the fields we care about, or we barf)

There's no reason for us to design a (very needed) net new API around implicitly requiring the _istio-specific_ SPIFFE ID format. There's no particular reason they need to be "our" certs. They could be SPIREs, or anybodys. Istio (especially ambient) really doesn't care what grants workload identities, as long as those identities have 

1. A SAN.
2. Which is in the SPIFFE ID format.
3. Which has AT-LEAST certain Istio-specific fields.

The problem is the current AuthPol API practically requires that to be AT-MOST certain Istio-specific fields, because it only supports an exact-match `principal` - _Istio_ does not require this. Just our (not great) AuthPol API, which this seeks to change with net-new fields (which is great).

Imagine the SAN is a SPIFFE ID, but you can't make assumptions about its complete format. You can assume it will have ns/sa/td _parts_ - but you can't assume you will always have an _exact_ string match against the Istio-only format.

- for an AuthPol that is `match exactPrincipal || ns/sa` -> policy resolution is always unambiguous, great.
- for an AuthPol that is `match exactPrincipal` -> policy resolution is always unambiguous, great.
- for an AuthPol that is `match ns/sa` -> -> policy resolution is always unambiguous, great.
- for an AuthPol that is `match exactPrincipal && ns/sa` -> this is effectively impossible to resolve unambiguously, unless the API excludes this condition or introduces an explicit precedence (or we make implicit assumptions elsewhere which will confuse people looking at the under-specified API - which I would like to avoid here).


Current state:
1. AuthPol has the `principal` field. 
2. If specified, this must be a SPIFFE ID. 
3. If specified, it also (at this time) must be a SPIFFE ID (exact/strict match) in the Istio-specific format. Which is not desirable because it effectively singlehandledly forces you to use Istio's workload CA - or use a different workload CA and retool it to use Istio's format for SANs. Which is a pain if you already have a workload CA in your env and you already have a format - we have had multiple bugs opened around this as well as user and customer complaints. 
4. Logically/codewise, Istio (especially Ambient) doesn't need to force you to use the _exact_ Istio format, and we could fix that relatively easily. 
5. However, just fixing that in the code doesn't solve the problem - you still can't use a different SPIFFE ID format because now all people's extant AuthPols will break - because they all force you to encode the _entire_ SPIFFE ID in the strict Istio format. We have talked about this before actually, it's come up repeatedly in SPIRE discussions as the main blocker for fully supporting SPIRE or other workload CAs that follow the SPIFFE ID standard - there are no good workarounds, just hard-to-maintain hacks that create operational burden.

This PR:

1. Ditches the requirement to specify a `principal` field (great - I agree, nobody should put a raw SPIFFE ID in their AuthPol, and we don't need the full/exact SPIFFE ID to resolve Istio policy)
2. Instead we can ignore the underlying format of the principal _as a first-class API concern_ (also great) and just let people specify matchers like (minimally) `serviceAccount`+`namespace`. Maybe `trust domain` as an optional additional specifier later but that's out of scope as I think we can easily assume that if only the `SA/NS` is defined. Maybe other things later, whatever.
3. Great, now we have removed the explict and implicit assumption of a specific, fixed SPIFFE ID format from the AuthPol API - perfect. 
4. Except if we allow `exactPrincipal` and `<other stuff>`  to be AND-able in AuthPol, we reintroduce the implicit assumption again because the API is underspecified (not to mention we'll confuse the hell out of people reading the AuthPol docs - what's the operational point of ANDing those?)

ilrudie: So then the expectation is that we change things and configure our proxies to do a match on both "ns/specified-ns" and "sa/specified-sa" being present in the SANs if we are using this field?

bleggett: Ah nevermind, I had forgotten we support N principal strings.

Principals:
`spiffe://td/ns/<ns>/sa/<sa>`
`spiffe://td/ns/<ns>/sa/<sa>/foo/bar/baz`

SA:
`ns/sa`

So the above would OR the principal(s) and AND the SA.

That's fine, then. The problem I was thinking of was

Principals:
`spiffe://td/ns/<ns>/sa/<sa>`

SA:
`ns/sa`

but my actual principal is `spiffe://td/ns/<ns>/sa/<sa>/foo/bar/ba`

Here this would _still_ fail, but we could just say `either define all possible principals, or none`

bleggett (approved):
This LGTM and makes https://github.com/istio/istio/issues/43105 much easier to boot.

It may be necessary to add something like a trust domain later, but I do think it's much better to have an API that looks like

- service_account
- trust_domain
- ...etc

versus 
- exactPrincipal

or 

- matcherTupleWhichIsAKindOfIdentity (random fixed fields)

security/v1beta1/authorization_policy.proto
bleggett: > So then the expectation is that we change things and configure our proxies to do a match on both "ns/specified-ns" and "sa/specified-sa" being present in the SANs if we are using this field?

The API as I read it doesn't require the new `service_account` field to be a strict substring of the SPIFFE ID.

So that means if we have an AuthPol with

```
service_account: default/bar
```

we can internally map that against an identity principal of `spiffe://td/ns/default/sa/bar` OR ` spiffe://td/ns/default/sa/bar/baz/beep/boop` pretty trivially with one AuthPolicy.

howardjohn:
I've added some tests and validation. I have blocked usage of SA with principals, in the same `from`, since they are ANDed -- it does not make sense to say "from ns=a AND sa=foo/bar".

bleggett:
> I've added some tests and validation. I have blocked usage of SA with principals, in the same `from`, since they are ANDed -- it does not make sense to say "from ns=a AND sa=foo/bar".

We have historically supported (in the API validation sense, not the logical sense) 

`from ns=a AND principal=spiffe://td/ns/b/sa/c`

which is, by the same token, also ~always a validation error we should probably check for (but that's OOS)

security/v1beta1/authorization_policy.proto
```diff
@@ -423,11 +425,14 @@ message Rule {
 // namespaces: ["prod", "test"]
 // notIpBlocks: ["203.0.113.4"]
 // ```
+// +kubebuilder:validation:XValidation:message="Cannot set serviceAccounts with namespaces or principals",rule="(has(self.serviceAccounts) || has(self.notServiceAccounts)) ? (!has(self.principals) && !has(self.notPrincipals) && !has(self.namespaces) && !has(self.notNamespaces)) : true"
```
hzxuzhonghu, line 428: https://github.com/istio/api/pull/3340#issuecomment-2429533214  

please document on the Serviceaccount field, writing here it is not exposed to user.

howardjohn: This is kind of a broader question about if we need to comment every single validation rule. I think no - think about the validation webhook, we have 1000s of rules that are not in the API comments. It's just excessively verbose to list each one; if a user tries something invalid it will fail and they can easily see why

hzxuzhonghu: I donot fully agree, if a large number of fields has strict requirements. and we donot comment. How many error tries a user should do? It is a UX problem.

ilrudie: It would be nice to be able to list the validation requirements in doc (maybe hidden by default but able to be expanded) but this also creates some issues:
1. opportunity for deviation if CEL and/or VWH gets updated but docs are missed
2. confusion if CEL and/or VWH differs slightly from what was documented
3. confusion if/when only _some_ requirements are documented but others are still enforced

I don't have solution that doesn't require a lot of manual, error prone work

howardjohn: > I donot fully agree, if a large number of fields has strict requirements. and we donot comment. How many error tries a user should do? It is a UX problem.

And if I did the "lazy" thing and only put this validation in the mutating webhook instead of in the API -- where its far more safe, efficient, and visible -- no one would have any objections! Our validations are [5,000 lines long](https://github.com/istio/istio/blob/dd2c7486262208565bcda07dd60c480100d284a8/pkg/config/validation/validation.go#L3270-L3269). I think it would be a determinant to users and developers to try to document all of these (nor have I seen any other project do so).

Its not like we are rejecting some common use case -- we are rejecting configs that don't even make sense (`A && B` where A != B is never true).

hzxuzhonghu: I always feel it should be encouraged to document valid values for a field and conflicts cases if there is any.   Istio has many places saying `only one of ` `cannot specify .... with ...`, and k8s also does the same thing.

This is not complex case like cross CRDs, which can not be described clearly

howardjohn: Added it to the doc

hzxuzhonghu (approved):
Thank you

security/v1beta1/authorization_policy.proto
louiscryan: I largely agree with Ian's comment about semantic assumptions in this thread.

We clearly have limitations in how we interpret SPIFFE URIs when the principal is a multi-path segment but leaving those aside for a moment these are both ways to define principals and should be ORed

After all there's not a lot of logical difference between the new fields and allowing a new URI type in this field to reference SAs. E.g

k8s://ServiceAccount/{namespace}/{name}

This is also the same damn thing as targetRef which has the nice property of allowing for reference expansion without requiring the API to evolve. E.g. this would allow for the introduction of types which represented principal groups to be referenced if they were added to the system

linsun (commented):
Felt a little odd to have namespace inside the value for serviceAccount while we also have namespace as a separate field under `source` but does make sense in this context.  

LGTM

howardjohn:
> Felt a little odd to have namespace inside the value for serviceAccount while we also have namespace as a separate field under `source` but does make sense in this context.
> 
> LGTM

I thought the same initially and I would not have done it, but the two are lists and NS supports globbing, so we cannot meaningfully re-use the field IMO

security/v1beta1/authorization_policy.proto
bleggett: > I largely agree with Ian's comment about semantic assumptions in this thread.
> 
> We clearly have limitations in how we interpret SPIFFE URIs when the principal is a multi-path segment but leaving those aside for a moment these are both ways to define principals and should be ORed


Yeah, I'm fine with `OR` - `AND` is what I originally thought made it a bit tricky given current assumptions Istio makes about SPIFFE ID formats (but even that is work-around-able, so I'm not worried about it)

> After all there's not a lot of logical difference between the new fields and allowing a new URI type in this field to reference SAs. E.g
> 
> k8s://ServiceAccount/{namespace}/{name}

Do our APIs need *any* assumptions about principal formats at all? I kind of believe they do not. 

e.g. istio cares about SA/NS/TD. How those are encoded in a specific principal format shouldn't be an API assumption, and doesn't need to be. The code behind the API just needs to know how to extract those, to match them.

The API doesn't need to care if the principal format is


- k8s://ServiceAccount/{namespace}/{name}
- spiffe://foo/bar/baz/ServiceAccount/{namespace}/{name}
- some x509 custom fields

as long as the policy encodes the properties, it doesn't need to explicitly encode the format that contains them.

> This is also the same damn thing as targetRef which has the nice property of allowing for reference expansion without requiring the API to evolve. E.g. this would allow for the introduction of types which represented principal groups to be referenced if they were added to the system

something like

```
targetRefs:
   - principalTypes
```

?

linsun:
Hi @louiscryan a friendly ping on this, any objections?

costinm (commented):
Sorry for the late comment - when I read it first it sounded like a great idea, but I recently changed my mind a bit.

The fundamental problem with principal is the documentation stating "it must be in the weird spiffe format that only Istio is using". A service account in another weird format only used by Istio is not making things much better. 

Keep in mind a service_account has the built-in assumption that multi-cluster is a flat trust model (which is not entirely true if you use principal fully).  It also has a built in assumption that the identity is expressed a K8S service account ( there are many other service accounts ). 

And while namespace is very important - the service account in the client namespace is quite less important and very hard to track if others use the model of one service account per revision.

In the end - the service_account is still a principal, just a very narrow kind.

How about just relaxing the docs and validations on principal, and allow either a url or 'did' like format - with a prefix that indicate what kind of principal it is.

Like did - we can have did:k8s:namespace or did:k8sprefix:namespace:sa_prefix (to deal with what I think is quite common case of versioned service account - see Istiod own use). And we can also support non-Istio SANs ( plain old service accounts using an email format), or DNS ( now that using did:web is getting popular ).

security/v1beta1/authorization_policy.proto
costinm: I suspect this comment thread is close to my last comment - to be clear my suggestion is to just use the existing principal field, with a URL or DID syntax, and to not emphasize the service account but namespace and a service account prefix.

howardjohn:
> A service account in another weird format only used by Istio is not making things much better.

I have a hard time aligning this with the current state of Istio. Service Account is **the core identity primitive** in Istio -- full stop. Yet we make people express their polices in a weird Istio internal format (spiffe) instead of letting them directly reference the primitive.

It would be like an IAM engine that, instead of letting you say `email: john@mail.com` makes me say `identity: auth://internal/email/john@mail.com`. The former is a lot more ergonomic for a user to deal with than trying to understand yet another mapping of one concept to another

> Keep in mind a service_account has the built-in assumption that multi-cluster is a flat trust model (which is not entirely true if you use principal fully)

I agree but this is a feature -- if a user wants to express the full identity (taking into account trust domain) they are free to do so! But in many cases, they actually want to express just the SA (namespace sameness is a core tenant of Istio and basically every multi-cluster identity scheme from cloud k8s providers). The existance of the `serviceAccounts` field doesn't preclude a user from using `principals`.

>  It also has a built in assumption that the identity is expressed a K8S service account ( there are many other service accounts ).

The existence of the `serviceAccounts` field doesn't preclude a user from using `principals`. Just because we support many formats doesn't mean we cannot offer helpers to easily and safely reference the common use cases.

> And while namespace is very important - the service account in the client namespace is quite less important and very hard to track if others use the model of one service account per revision.

The existence of the `serviceAccounts` field doesn't preclude a user from using `namespaces`...

> In the end - the service_account is still a principal, just a very narrow kind.

It is narrow, perhaps, in the broad sense. In Istio, it covers >99% of use cases. Its also likely to become the core primitive for NetworkPolicy in Kubernetes (https://github.com/kubernetes-sigs/network-policy-api/pull/274). 

> How about just relaxing the docs and validations on principal, and allow either a url or 'did' like format - with a prefix that indicate what kind of principal it is.
> Like did - we can have did:k8s:namespace or did:k8sprefix:namespace:sa_prefix (to deal with what I think is quite common case of versioned service account - see Istiod own use). And we can also support non-Istio SANs ( plain old service accounts using an email format), or DNS ( now that using did:web is getting popular ).

This generally defeats the purpose of this proposal IMO... The issue is not the lack of ability to specify things, but the lack of ability to **understand how to use** the API. Adding yet another obscure format doesn't seem to meaningfully help this situation.

costinm:
On Tue, Dec 17, 2024 at 11:31 AM John Howard ***@***.***>
wrote:

> A service account in another weird format only used by Istio is not making
> things much better.
>
> I have a hard time aligning this with the current state of Istio. Service
> Account is *the core identity primitive* in Istio -- full stop. Yet we
> make people express their polices in a weird Istio internal format (spiffe)
> instead of letting them directly reference the primitive.
>

Well - for client identification, the service account is the core identity
for K8S clients and in Istio, but I would argue the namespace is really the
'core' boundary in both Istio and K8S.

I don't disagree that the current spiffe format in principal is bad - my
point is that 'principal' is  the core concept for authentication in
general, and it is in no way restricted to spiffe or istio format.
We added it precisely to allow non-Istio (and non-K8S) identities to be
expressed.



> It would be like an IAM engine that, instead of letting you say email:
> ***@***.*** makes me say identity: ***@***.***
> The former is a lot more ergonomic for a user to deal with than trying to
> understand yet another mapping of one concept to another
>

Yes, 100% agree. So

  principal: ***@***.***





> Keep in mind a service_account has the built-in assumption that
> multi-cluster is a flat trust model (which is not entirely true if you use
> principal fully)
>
> I agree but this is a feature -- if a user wants to express the full
> identity (taking into account trust domain) they are free to do so! But in
> many cases, they actually want to express just the SA (namespace sameness
> is a core tenant of Istio and basically every multi-cluster identity scheme
> from cloud k8s providers). The existance of the serviceAccounts field
> doesn't preclude a user from using principals.
>

I'm not saying it precludes - just that another istio-invented syntax is
not the best option.
See also https://kubernetes.io/docs/setup/best-practices/certificates/*
 - *another
syntax/format for service accounts used by K8S, or the syntax used in K8S
issued JWT tokens.


If it was 'namespaces' - I wouldn't mind - it is a clear concept and
doesn't need another mangling.



> It also has a built in assumption that the identity is expressed a K8S
> service account ( there are many other service accounts ).
>
> The existence of the serviceAccounts field doesn't preclude a user from
> using principals. Just because we support many formats doesn't mean we
> cannot offer helpers to easily and safely reference the common use cases.
>

I meant: "one" service account. Istiod identity is not expressed as a
single service account - it's a set of service accounts that change with
each version.


> And while namespace is very important - the service account in the client
> namespace is quite less important and very hard to track if others use the
> model of one service account per revision.
>
> The existence of the serviceAccounts field doesn't preclude a user from
> using namespaces...
>
Or using a `principal: k8s:namespace:serviceaccount` ( or whatever the
format K8S is already using in JWTs and client certs) - instead of an Istio
namespace/serviceaccount



> In the end - the service_account is still a principal, just a very narrow
> kind.
>
> It is narrow, perhaps, in the broad sense. In Istio, it covers >99% of use
> cases. Its also likely to become the core primitive for NetworkPolicy in
> Kubernetes (kubernetes-sigs/network-policy-api#274
> <https://github.com/kubernetes-sigs/network-policy-api/pull/274>).
>

It doesn't even cover Istiod use case (versioned SA).

If the NetworkPolicy is adopted and uses ns/sa syntax - I don't have any
objections. My concern is defining another Istio-only syntax for KSA.




> How about just relaxing the docs and validations on principal, and allow
> either a url or 'did' like format - with a prefix that indicate what kind
> of principal it is.
> Like did - we can have did:k8s:namespace or
> did:k8sprefix:namespace:sa_prefix (to deal with what I think is quite
> common case of versioned service account - see Istiod own use). And we can
> also support non-Istio SANs ( plain old service accounts using an email
> format), or DNS ( now that using did:web is getting popular ).
>
> This generally defeats the purpose of this proposal IMO... The issue is
> not the lack of ability to specify things, but the lack of ability to *understand
> how to use* the API. Adding yet another obscure format doesn't seem to
> meaningfully help this situation.
>

So adding an obscure `serviceAccount: namespace/serviceAccount` is
significantly better than `principal: k8s:namespace:sa` ?

See my previous comment: if it is a syntax that is shared with
NetworkPolicy and not unique to Istio - no concerns, looks good.

I have not seen you address the versioned service accounts - which is IMO
one of the biggest issues.

Costin

costinm:
To be clear: I'm 100% for replacing the spiffe URL in principal, I loved
the proposal initially and I fully support using a more explicit service
account.

The details of the syntax - and limitations on expressing versioned SA -
are the problem, not preserving the use of the current (bad) syntax.

bleggett:
> How about just relaxing the docs and validations on principal, and allow either a url or 'did' like format - with a prefix that indicate what kind of principal it is.
> 
> Like did - we can have did:k8s:namespace or did:k8sprefix:namespace:sa_prefix (to deal with what I think is quite common case of versioned service account - see Istiod own use). And we can also support non-Istio SANs ( plain old service accounts using an email format), or DNS ( now that using did:web is getting popular ).

This is kinda why I mentioned above I would ultimately rather the API not support specifying particular principal formats _at all_ - there are many, and the Istio API, strictly speaking, does not need to care about them all.

In an ideal world no one would put principals in their AuthPols at all, they'd simply define one or more "principal matchers" in the API, and then the same AuthPol can be applied to any principal format Istio happens to support internally (`did`, x509 fields, SPIFFE, JWT, whatever)

howardjohn:
> is significantly better than `principal: k8s:namespace:sa`

Setting aside what is a better UX, which is a higher opinionated topic... I functionally don't think we can do this. The contract today is that principal is direct mapping to the (SPIFFE specific) URI SAN. We cannot really change that -- how would we know if we were supposed to be treating that as a service account reference, or a strange SPIFFE url? There is also all sorts of complexities in the `principal` field like wildcards -- what if a user does `principal: *k8s:namespace:sa`?

I worry this would bring even more confusion to users, not less.

Keep in mind most Istio users are far from authentication experts. You (and probably everyone on this PR) are easily in the 99.999% percentile in that regard -- most people probably don't even know what a `principal` means in the first place!

>  I have not seen you address the versioned service accounts - which is IMO one of the biggest issues.

Please do not scope creep this... The goal is not to allow **more** functionality, its to make core use cases easier which has been the consistent goal of the project for the last 3 years.

howardjohn:
To be super clear -- the use case I am trying to solve is this: https://istio.io/latest/blog/2022/get-started-ambient/#l4-authorization-policies.

The **vast** majority of policies are of this form, and its such a huge cliff to understanding Istio. Its **the** top motivator for Istio adoption, we have it as literally step 1 in our (and others) getting started walkthroughs, and its 10x more complicated than it needs to be!

costinm:
On Tue, Dec 17, 2024 at 12:07 PM John Howard ***@***.***>
wrote:

> is significantly better than principal: k8s:namespace:sa
>
> Setting aside what is a better UX, which is a higher opinionated topic...
> I functionally don't think we can do this. The contract today is that
> principal is direct mapping to the (SPIFFE specific) URI SAN. We cannot
> really change that -- how would we know if we were supposed to be treating
> that as a service account reference, or a strange SPIFFE url? There is also
> all sorts of complexities in the principal field like wildcards -- what
> if a user does principal: *k8s:namespace:sa?
>
> I worry this would bring even more confusion to users, not less.
>

SIPFFE is a URI SAN with spiffe:// scheme.

We currently don't require the scheme - but the actual principal must have
it, so that's already a big part of the confusion.

I don't think changing the docs (and validation) to allow users to specify
the scheme - and exactly match the principal and the SAN - is confusing or
problematic, and it is possible to do it in a backwards-compatible
way.

If we allow scheme: I don't see the confusion in allowing other schemes,
like k8s.

Taking up words like 'principal' or 'serviceAccount' - and using them in a
super narrow and 'original' sense - as 'istio mangling of a service account
to a URL without scheme' and
'k8s service account in a particular istio specific format' is far more
confusing than the normal semantics that people outside istio use.

Allowing wildcards in principal is a slightly different topic - if we do,
we should probably seriously consider deprecating it. Prefix may be ok,
different field may be ok too - but regex and arbitrary widlcards are a
huge problem.

Using scheme can solve this in a more elegant way:

  principal 'namespace:example'
or
 principal 'k8sprefix:namespace:istiod-'

could use the schema to clearly indicate which principal formats can be
interpreted as a matcher.




> Keep in mind most Istio users are far from authentication experts. You
> (and probably everyone on this PR) are easily in the 99.999% percentile in
> that regard -- most people probably don't even know what a principal
> means in the first place!
>

Most people don't know 'serviceAccount' means a K8S service account in a
new syntax we just invented ( in particular if they use any cloud - where
service account can be a vendor service account, or use enterprise
SSO/LDAP/etc which has its own service accounts)

And I suspect the people who are not security experts are better using
'principal: k8snamespace:example' and not deal with (versioned) service
accounts or matchers at all.


> I have not seen you address the versioned service accounts - which is IMO
> one of the biggest issues.
>
> Please do not scope creep this... The goal is not to allow *more*
> functionality, its to make core use cases easier which has been the
> consistent goal of the project for the last 3 years.
>
I believe Istiod has used versioned accounts for >3years, and having a
separate SA per version ( allowing the RBAC to evolve ) is not 'scope
creep' but something we ourselves are doing.

We can't pretend versioned service accounts don't exists or users are not
allowed to use them ( but we are ).

We can make it easier by allowing the user to specify just a namespace -
hard to make it easier than that.

costinm:
On Tue, Dec 17, 2024 at 12:10 PM John Howard ***@***.***>
wrote:

> To be super clear -- the use case I am trying to solve is this:
> https://istio.io/latest/blog/2022/get-started-ambient/#l4-authorization-policies
> .
>
> The *vast* majority of policies are of this form, and its such a huge
> cliff to understanding Istio. Its *the* top motivator for Istio adoption,
> we have it as literally step 1 in our (and others) getting started
> walkthroughs, and its 10x more complicated than it needs to be!
>
>
100% agree with the use case and goal.

My concern is with repeating the same mistake by adding another 'unique'
format and more confusion.

At the very least call it k8sServiceAccount, not 'serviceAccount' - and use
the same scheme that k8s is using in NetworkPolicy.

But I fail to see how allowing principal to be a URL is a problem - and
adding a new field is better.

howardjohn:
> We can make it easier by allowing the user to specify just a namespace - hard to make it easier than that.

This is already supported in the API and is a great feature! But often users want to segment further than namespace. I know you personally do not think that is a good idea, but it is the reality of how users use Istio today. I have lost track of the number of users I have seen that have massive namespaces (or even a single namespace!) with desperate apps. Namespace is the **best** boundary but not the only boundary.

costinm:
I think it is a very good idea to support service accounts - but taking
into account that service accounts can be versioned (like we do), that the
syntax for representing K8S Service accounts needs to be consistent with
K8S -
and that not all service accounts are K8S ( and will be very confusing to
many users who also use non-K8S service accounts - Istio is still supposed
to support more than K8S).

costinm:
John - I don't think we disagree on the goals or why this is a good idea,
my comments are just about syntax and consistency, i.e. using
'serviceAccount' keyword and  namespace/serviceAccount.

For the latter - I am ok with it if that's also what K8S is using in
NetworkPolicy.

Regardless of the presence of serviceAccount ( or k8sServiceAccount as I
would suggest ) - I think allowing/supporting proper scheme:// in the
principal would be extremely useful,
and a k8s:// scheme should be possible.

therealmitchconnors:
Istio has operated for as long as I can remember with the mantra that "Easy things should be easy, and hard things should be possible" (shamelessly stolen from Larry Wall).  Using a K8s Service Account as a principal should be an easy thing, as it is what is used in all but the most complex Istio environments.  Using a DID for principal, related to an arbitrary identity provider, is not something we've heard users ask for, but seems to fall into the category of hard things which should be possible (assuming we can safely handle migrating the field to support scheme, etc.)

With that principle in mind, it seems to me that both Service Account fields and an expanded support for DID principals would be valuable additions to the project, and do not appear to be incompatible with one another.  I think it makes sense to merge this PR, and to allow @costinm to write a formal proposal for supporting arbitrary identity providers...

From a UX perspective, I don't think it makes sense to support policies which explicitly set the service account field and the principal field.  We can sort of consider the Service Account field to be a structured projection of the principal field.  In the case that both are specified, I think we should mark a status.condition to indicate that the policy is invalid.

costinm:
On Tue, Dec 17, 2024 at 3:30 PM Mitch Connors ***@***.***>
wrote:

> Istio has operated for as long as I can remember with the mantra that
> "Easy things should be easy, and hard things should be possible"
> (shamelessly stolen from Larry Wall). Using a K8s Service Account as a
> principal should be an easy thing, as it is what is used in all but the
> most complex Istio environments. Using a DID for principal, related to an
> arbitrary identity provider, is not something we've heard users ask for,
> but seems to fall into the category of hard things which should be possible
> (assuming we can safely handle migrating the field to support scheme, etc.)
>
> With that principle in mind, it seems to me that both Service Account
> fields and an expanded support for DID principals would be valuable
> additions to the project, and do not appear to be incompatible with one
> another. I think it makes sense to merge this PR, and to allow @costinm
> <https://github.com/costinm> to write a formal proposal for supporting
> arbitrary identity providers...
>

I was not actually suggesting we support DID - it was just an example.

spiffe:// ( as required by the spiffe specification) - and using a scheme
for k8s instead of foo/bar ( or some other way to make it clear it is a k8s
account in an Istio specific syntax  - not taking over the 'service
account' concept).

We have the 'easy' part with namespace - I don't think the
namespace/serviceaccount is really easy if you consider the best practices
(that we follow) of not sharing
the same RBAC and service account across versions. Making it easy to do the
wrong thing (sharing a SA across multiple versions) is not ideal, even if
it is common.



> From a UX perspective, I don't think it makes sense to support policies
> which explicitly set the service account field and the principal field. We
> can sort of consider the Service Account field to be a structured
> projection of the principal field. In the case that both are specified, I
> think we should mark a status.condition to indicate that the policy is
> invalid.
>
Not sure - it's an 'OR', a service may accept both 'easy' clients in other
namespaces, service accounts - and some common non-Istio certificates ( in
spiffe: or dns: or email: format)


BTW - are we still agreeing that 'least privilege' and not having a shared
account ( like we had in the old Istio) for all versions/revisions is the
best practice ?

nmnellis:
I think we are missing the point of this feature and the end user it serves. 95% of Istio users just want to enable Service A to talk to Service B and that is all they know. Being able to express that intent easily helps prevent these users from making mistakes and having to learn about the identity system Istio uses. `<namespace>/<service>` is simple enough for all users to understand but making it some text based format like `principal 'k8sprefix:namespace:istiod-'` would not serve these users any better than today.

This feature request was in response to a NEW Istio end user trying it and getting frustrated trying to understand the complicated nature of Spiffe. This user ended up in a situation where a messed up AuthorizationPolicy took 2 weeks to find.

@costinm If we call it `k8sServiceAccount` would that be sufficient in moving forward?

costinm:
On Wed, Dec 18, 2024, 07:44 Nick ***@***.***> wrote:

> I think we are missing the point of this feature and the end user it
> serves. 95% of Istio users just want to enable Service A to talk to Service
> B and that is all they know. Being able to express that intent easily helps
> prevent these users from making mistakes and having to learn about the
> identity system Istio uses. <namespace>/<service> is simple enough for
> all users to understand but making it some text based format like principal
> 'k8sprefix:namespace:istiod-' would not serve these users any better than
> today.
>
Very interesting point. I wouldn't mind a k8sService: scheme - it doesn't
have the problem with versioned accounts and indeed matches what people
understand.

Keep in mind Service A may use multiple service accounts ( like Istio) and
a service account can be shared by multiple services. Miss one - either
breakage or security issue

Namespace/service would be easy to understand if that was the syntax used
by k8s - but it is not, k8s is using service.namespace

For service accounts - it is using the system:... - in the JWTs we use to
get the cert.






This feature request was in response to a NEW Istio end user trying it and
> getting frustrated trying to understand the complicated nature of Spiffe.
> This user ended up in a situation where a messed up AuthorizationPolicy
> took 2 weeks to find.
>
> @costinm <https://github.com/costinm> If we call it k8sServiceAccount
> would that be sufficient in moving forward?
>

I can live with that - if we can confirm the namespace/ksa scheme is what
NetworkPolicy will adopt, or if we use a scheme that is in use with k8s.

It still has the problem of versioned service account - I would much rather
have what you suggested in this comment, k8sService - which is also the
typical user intent and solves both versioned accounts and shared accounts.

But it's not as bad as what we have.

costinm:
BTW - to add on the idea in Nick's email - "Service A talk to Service B" is
what we use for the client to server authentication ( 'secure naming' ) -
so it would make the security symmetrical and easier to understand.

At the same time - it has its own problems ( the set of identities
associated with service changes - in particular with versioned accounts ).
But as a concept it is easiest to understand by a user.

linsun (approved):
Based on the discussion in today's upstream WG mtg, this PR does help users to simplify authz policies from the required principal today to k8s service account.  Regarding Costin's concern about cloud provider also has service account concept, the general consensus is to clarify in the comment that the service account means k8s service account here and not asking users to use k8s as part of the `serviceAccounts` field.

/lgtm 

/hold

adding hold so John can remove after he clarified that in the comment for that.

PR #3344, opened 2024-10-29
mesh: support configuring XFF trusted cidrs

rudrakhp:
Ref istio/istio#53185
Related feature in envoy [here](https://github.com/envoyproxy/envoy/blob/62be4be62fab3f56490dcbb3cdaf716040823746/api/envoy/extensions/http/original_ip_detection/xff/v3/xff.proto#L48).


mesh/v1alpha1/proxy.proto
```diff
@@ -256,7 +256,7 @@ message Topology {
   // forwarding it to the upstream services in the cluster.
   // The default value of numTrustedProxies is 0.
   // See [Envoy XFF](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-forwarded-for)
-  // header handling for more details.
+  // header handling for more details. Only one of num_trusted_proxies and trusted_cidrs must be set.
```
howardjohn, line 259:
```suggestion
  // header handling for more details. Only one of `numTrustedProxies and `trustedCidrs` may be set.
```

Users interact with YAML which is camel case

mesh/v1alpha1/proxy.proto
```diff
@@ -274,6 +274,15 @@ message Topology {
   // Enables [PROXY protocol](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) for
   // downstream connections on a gateway.
   ProxyProtocolConfiguration proxy_protocol = 3;
+
+  // List of CIDRs to trust when evaluating the remote IP address to determine the original client’s IP address.
+  // When the remote IP address matches a trusted CIDR and the x-forwarded-for header was sent,
```
hzxuzhonghu, line 279:
```suggestion
  // When the remote IP address matches a trusted CIDR and the `x-forwarded-for` header was sent,
```

mesh/v1alpha1/proxy.proto
```diff
@@ -274,6 +274,15 @@ message Topology {
...
+  // each entry in the x-forwarded-for header is evaluated from right to left
```
hzxuzhonghu, line 280:
```suggestion
  // each entry in the `x-forwarded-for` header is evaluated from right to left
```

mesh/v1alpha1/proxy.proto
```diff
@@ -274,6 +274,15 @@ message Topology {
...
+  // and the first public non-trusted address is used as the original client address.
+  // If all addresses in x-forwarded-for are within the trusted list, the first (leftmost) entry is used.
```
hzxuzhonghu, line 282:
```suggestion
  // If all addresses in `x-forwarded-for` are within the trusted list, the first (leftmost) entry is used.
```

mesh/v1alpha1/proxy.proto
```diff
@@ -274,6 +274,15 @@ message Topology {
...
+  // See [Envoy XFF](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-forwarded-for)
+  // header handling for more details. Only one of numTrustedProxies and trustedCidrs may be set.
```
hzxuzhonghu, line 284:
```suggestion
  // header handling for more details. Only one of `numTrustedProxies` and `trustedCidrs` may be set.
```

hzxuzhonghu (approved):

rudrakhp:
@istio/technical-oversight-committee bumping this up. Thanks!

mesh/v1alpha1/proxy.proto
```diff
@@ -274,6 +274,15 @@ message Topology {
...
   ProxyProtocolConfiguration proxy_protocol = 3;
+
+  // List of CIDRs to trust when evaluating the remote IP address to determine the original client’s IP address.
+  // When the remote IP address matches a trusted CIDR and the X-Forwarded-For (XFF) header was sent,
+  // each entry in the X-Forwarded-For (XFF) header is evaluated from right to left
+  // and the first public non-trusted address is used as the original client address.
+  // If all addresses in X-Forwarded-For (XFF) header are within the trusted list, the first (leftmost) entry is used.
+  // See [Envoy XFF](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#config-http-conn-man-headers-x-forwarded-for)
+  // header handling for more details. Only one of `numTrustedProxies` and `trustedCidrs` may be set.
+  repeated string trusted_cidrs = 4;
```
howardjohn, line 285: The envoy docs say we cannot use it with `UseRemoteAddress`. `UseRemoteAddress` is currently always set to 'true' for gateways.

What will be the behavior if a user sets this? Will we set useRemoteAddress=false? What side effects does this have beyond deciding how to parse the XFF header?

howardjohn (changes requested):

mesh/v1alpha1/proxy.proto
rudrakhp: @howardjohn Yes we will have to always set `UseRemoteAddress` to `false` if we decide to use the XFF original IP detection extension (which is recommended over the alternative we use today, might soon be on deprecation path).
>What side effects does this have beyond deciding how to parse the XFF header?

From the docs it does look like the only thing that changes is how the original IP is determined, HCM does it natively today, we will be moving to original IP detection extension.

Meanwhile waiting on a probable fix for difference in behaviour of HCM and the XFF origin detection extension: https://github.com/envoyproxy/envoy/pull/37780

rudrakhp: @howardjohn please refer https://github.com/envoyproxy/envoy/issues/34241#issuecomment-2585123169 for moving to original IP detection extension with backward compatibility. Corresponding implementation in [envoyproxy/gateway](https://github.com/envoyproxy/gateway/blob/a58021e6b4ae79fc268249cffcab8e342b04171b/internal/xds/translator/listener.go#L114-L123). WDYT?

jaellio: Is your goal to use a [detection extension](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-original-ip-detection-extensions) and set `xff_trusted_cidrs`?

mesh/v1alpha1/proxy.proto
```diff
@@ -274,6 +274,15 @@ message Topology {
   // Enables [PROXY protocol](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) for
   // downstream connections on a gateway.
   ProxyProtocolConfiguration proxy_protocol = 3;
+
+  // List of CIDRs to trust when evaluating the remote IP address to determine the original client’s IP address.
+  // When the remote IP address matches a trusted CIDR and the X-Forwarded-For (XFF) header was sent,
+  // each entry in the X-Forwarded-For (XFF) header is evaluated from right to left
+  // and the first public non-trusted address is used as the original client address.
```
jaellio, line 281: This differs from the current envoy documentation. https://github.com/envoyproxy/envoy/blob/main/docs/root/configuration/http/http_conn_man/headers.rst?plain=1#L274-L276
> * If the remote address is contained by an entry in ``xff_trusted_cidrs``, and the *last*
   (rightmost) entry is also contained by an entry in ``xff_trusted_cidrs``, the trusted client
   address is *second-last* IP address in XFF.

Based on the docs I'd expect the second to last IP address in the XFF header to be used as the original client address. Not sure if this documentation is correct though. The docs conflict with a later example https://github.com/envoyproxy/envoy/blob/main/docs/root/configuration/http/http_conn_man/headers.rst?plain=1#L370-L383
   ```
    Example 7: Envoy as edge proxy, with one trusted CIDR
    Settings:
      | use_remote_address = false
      | xff_trusted_cidrs = 192.0.2.0/24

    Request details:
      | Downstream IP address = 192.0.2.5
      | XFF = "203.0.113.128, 203.0.113.10, 192.0.2.1"

    Result:
      | Trusted client address = 192.0.2.1
      | X-Envoy-External-Address is set to 192.0.2.1
      | XFF is changed to "203.0.113.128, 203.0.113.10, 192.0.2.1, 192.0.2.5"
      | X-Envoy-Internal is removed (if it was present in the incoming request)```

Note the trusted client address is set to `192.0.2.1`. Based on the documentation I would have expected the trusted client address to be set to `203.0.113.10`. I thought maybe the `second-last` address was evaluated after appending the downstream IP address to the XFF but this also conflicts with a later example https://github.com/envoyproxy/envoy/blob/main/docs/root/configuration/http/http_conn_man/headers.rst?plain=1#L385-398
```
    Example 8: Envoy as edge proxy, with two trusted CIDRs
    Settings:
      | use_remote_address = false
      | xff_trusted_cidrs = 192.0.2.0/24, 198.51.100.0/24

    Request details:
      | Downstream IP address = 192.0.2.5
      | XFF = "203.0.113.128, 203.0.113.10, 198.51.100.1"

    Result:
      | Trusted client address = 203.0.113.10
      | X-Envoy-External-Address is set to 203.0.113.10
      | XFF is changed to "203.0.113.128, 203.0.113.10, 198.51.100.1, 192.0.2.5"
      | X-Envoy-Internal is removed (if it was present in the incoming request)```
In example 8, the trusted client address is set to 203.0.113.10 which is the expected second-last XFF IP. 

@rudrakhp Could you please verify the expected behavior?

jaellio: Create an envoy issue to track this https://github.com/envoyproxy/envoy/issues/38462

jaellio: Based on my investigation and digging into the envoy code, your comment is correct. The envoy documentation is incorrect. Will update the envoy issue I created.

rudrakhp: @jaellio thanks for confirming!

mesh/v1alpha1/proxy.proto
rudrakhp: >Is your goal to use a [detection extension](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-original-ip-detection-extensions) and set xff_trusted_cidrs

Yes

>Is this change blocked on a similar change as above?

It's technically not blocked. According to discussion in https://github.com/envoyproxy/envoy/issues/34241#issuecomment-2585123169, setting `UseRemoteAddress` to `false` will unblock setting trusted CIDRs, but will change the behaviour of `XFFNumTrustedHops` set today in HCM. So for `XFFNumTrustedHops` to remain backward compatible after unsetting `USeRemoteAddress`, we will have to decrement the user configured num hops before setting it in the detection extension. Please refer to comment referred above with an example of equivalent configurations.

chlunde: Could we move forward with this now @howardjohn ? We've tried working around this with EnvoyFilter but realize that with different gateway topologies this PR is the only proper fix.

The issue in the other comment seems to be resolved now.

PR #3350, opened 2024-11-06
Add StatefulFormatter to mesh.ProxyConfig

whitneygriffith:
Part of https://github.com/istio/istio/issues/53680


networking/v1beta1/proxy_config.proto
```diff
@@ -80,6 +80,18 @@ import "type/v1beta1/selector.proto";
 //     imageType: debug
 // ```
 //
+// To preserve the header case for HTTP 1.x requests, set the `statefulFormatter.preserveCase` field on the `ProxyConfig` resource:
+// ```yaml
+// apiVersion: networking.istio.io/v1beta1
+// kind: ProxyConfig
```
keithmattix, line 86: I don't think we want to add this to the ProxyConfig resource; only the proxyConfig annotation

mesh/v1alpha1/proxy.proto
```diff
@@ -603,6 +603,9 @@ message ProxyConfig {
   // Specifies the details of the proxy image.
   istio.networking.v1beta1.ProxyImage image = 35;
 
+  // Defines options for stateful header formatters that maintain context or state across the processing of a request.
+  istio.networking.v1beta1.StatefulFormatter stateful_formatter = 40;
```
keithmattix, line 607: I'm not sure I like the idea of passing the complexity of "stateful formatters" to the users who just want to preserve their header casing. Can we utilize the proxyHeaders struct in some way to make this more intuitive?

keithmattix (changes requested):

mesh/v1alpha1/proxy.proto
whitneygriffith: I agree that we should avoid exposing the complexity of stateful formatters directly to users who are using the higher level `meshConfig` to enable header case preservation. I will look into adding `preserveCase` or `preserveHeaderCase`  to the `proxyHeaders` struct.

networking/v1beta1/proxy_config.proto
whitneygriffith: Can you point me to the `proxyConfig` annotation? From what I understand to enable `preserveCase` at the namespace or workload level this feature should be apart of the `proxyConfig` resource. Where the annotation is used to apply a specific `proxyConfig` resource to a workload.

keithmattix: If you check the recording of the community meeting, the consensus was that we didn't want to make this too flexible/easy to use since we want to encourage users to adopt http2. Adding the field to the annotation and not the resource allows us to do so. Notice that the networking/v1beta1/proxy_config.proto file and the mesh/v1alpha1/proxy.proto files have different fields. The former is the CRD and the latter is the annotation struct; please add the new fields there

keithmattix (approved):
This shape LGTM, but the PR may need a rebase to include some reason changes around deeplinks

mesh/v1alpha1/proxy.proto
```diff
@@ -722,6 +729,17 @@ message ProxyConfig {
...
+    // When true, the original case of HTTP/1.x headers will be preserved
+    // as they pass through the proxy, rather than normalizing them to lowercase.
+    // This field is particularly useful for applications that require case-sensitive 
+    // headers for interoperability with downstream systems or APIs that expect specific 
+    // casing.
+    // The preserve_http1_header_case option only applies to HTTP/1.x traffic, as HTTP/2 requires all headers 
+    // to be lowercase per the protocol specification. Envoy will ignore this field for HTTP/2 
+    // requests and automatically normalize headers to lowercase, ensuring compliance with HTTP/2 
+    // standards. 
+    bool preserve_http1_header_case = 40;
```
hzxuzhonghu, line 741: should we use `google.protobuf.BoolValue` type

hzxuzhonghu (approved):

howardjohn (approved):

PR #3351, opened 2024-11-06
Clarify that while this annotation has been deprecated, the label of the same name is still available.

rsalmond:
The existing wording made it appear (to me at least) that the annotation was deprecated in favour of itself.


howardjohn (approved):

kfaseela:
@rsalmond : I think you should join the Istio org following the instructions in the above comment . 🙂

hzxuzhonghu (approved):

PR #3354, opened 2024-11-07
Add cares max queries to MeshConfig

Stevenjin8:
For https://github.com/istio/istio/pull/53806 which has release notes


howardjohn (commented):
I really don't want a user to ever need to set this. We should just work out of the box and not DOS the cluster DNS server.

Stevenjin8:
@howardjohn 
> I really don't want a user to ever need to set this. We should just work out of the box and not DOS the cluster DNS server.

I don't think it will ever DOS the cluster DNS server since by default, there are an unlimited number of queries. I just think that its weird to have a configuration like dns_refresh_rate in meshconfig, but have this one as an environment variable.

howardjohn:
> @howardjohn
> 
> > I really don't want a user to ever need to set this. We should just work out of the box and not DOS the cluster DNS server.
> 
> I don't think it will ever DOS the cluster DNS server since by default, there are an unlimited number of queries. I just think that its weird to have a configuration like dns_refresh_rate in meshconfig, but have this one as an environment variable.

Just because we made poor decisions in the past doesn't mean we need to continue :-). `dns_refresh_rate` was effectively made obsolete as well by making 'respect TTL' the default

Stevenjin8:
@howardjohn Fair enough. I didn't know we felt this away about dns_refresh_rate. Generally, what things are we looking to add to meshconfig?

mesh/v1alpha1/config.proto
```diff
@@ -336,6 +336,13 @@ message MeshConfig {
   // Default refresh rate is `60s`.
   google.protobuf.Duration dns_refresh_rate = 36;
 
+  // Sets the `max_udp_queries` option in Envoy for the Cares DNS resolver.
+  // Defaults to 0, an unlimited number of connections.
+  // See `extensions.network.dns_resolver.cares.v3.CaresDnsResolverConfig` in
+  // https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/network/dns_resolver/cares/v3/cares_dns_resolver.proto
+  // and `ARES_OPT_UDP_MAX_QUERIES` in https://c-ares.org/docs/ares_init.html
+  google.protobuf.UInt32Value cares_dns_resolver_udp_max_queries = 67;
```
hzxuzhonghu, line 344: how about calling `max_dns_queries_per_connection` or add udp

hzxuzhonghu:
We can keep a option here, just make the default value what we want to set

ramaraochavali:
based on https://github.com/istio/istio/issues/53577, it felt like we can not come up with a decent default

howardjohn:
I am not sure why we cannot come up with a default value. We don't need to let perfect be the enemy of the good... we **know** that the current default (infinite) is bad. So what is better? 

I would rather pick a random high number than do nothing. But we can still do better by just trying out a few values (10/100/1k/10k?) and comparing the result.

I don't mind keeping an env var to tune it, as long as its an escape hatch for users. If we find we picked a poor default and users are needing to tune it, we can either change the default (if there is a clearly better option for general usage) or accept maybe it legitimately needs tuning and promote it to a first class API.

There are thousands of obscure fields we need to set in Istio -- I think our strategy should be:
1. Pick a sane default
2. Expose a temporary flag to tune it
3. Remove the flag/change the default/promote to a real API as we get feedback

If we do not do (1), our users are hurt and our product is more complex ("I need to tune 50 flags to get a stable mesh"). If we jump straight to a stable API, we end up with an infinitely growing list of obscure options users need to wade through -- when most are not needed!

Stevenjin8:
https://github.com/c-ares/c-ares/issues/913

ramaraochavali:
> I am not sure why we cannot come up with a default value. We don't need to let perfect be the enemy of the good... we know that the current default (infinite) is bad. So what is better?

I would rather pick a random high number than do nothing. But we can still do better by just trying out a few values (10/100/1k/10k?) and comparing the result.

Sorry - I should clarify. I was n't saying we can not come up with default - my intention was a sane default for all cases - We do not know what it is unless we test in various envs.

As long as there is path forward for removing the env var - I think we are good

howardjohn:
> Sorry - I should clarify. I was n't saying we can not come up with default - my intention was a sane default for all cases - We do not know what it is unless we test in various envs.
> 
> As long as there is path forward for removing the env var - I think we are good

Makes sense. My goal was really just to try to get to an end goal where we have zero config for this. And wait and see if we need it before making it a permanent API

Stevenjin8:
Closing

PR #3362, opened 2024-11-14
Add ambient virtual interface annotation for pods

bleggett:
This is effectively an analog to https://istio.io/latest/docs/reference/config/annotations/#SidecarTrafficKubevirtInterfaces but for ambient.

Since [we know it's more general](https://istio.slack.com/archives/C041EQL1XMY/p1730147675490669) than kubeVirt (it's useful for any kind of virtualized interface), it has been renamed slightly.

Relates to https://github.com/istio/istio/issues/49829

This is the API side, putting it up for comment/TOC approval with a hold, ~impl pending.~ impl side here: https://github.com/istio/istio/pull/53933


howardjohn (approved):

annotation/annotations.yaml
```diff
@@ -573,3 +573,12 @@ annotations:
     hidden: true
     resources:
       - Pod
+
+  - name: ambient.istio.io/reroute-virtual-interfaces
+    featureStatus: Alpha
+    description: |
+      A comma separated list of virtual interfaces whose inbound traffic will be unconditionally treated as outbound. This allows workloads using virtualized networking (kubeVirt, VMs, docker-in-docker, etc) to function correctly with ambient mesh traffic capture.
```
hzxuzhonghu, line 580: IIUC, this requires user to be aware of the nics clearly.  So It is a very bad ux, how does a user know the underlying implement details, they are not platform operator, but just application developer or deployer.

howardjohn: The UX of this is basically "if you use docker in docker, set to docker0. If you use kubevirt, set to <device name>". 

I don't think its that bad... I would be open to suggestions to make it better though, but I cannot come up with any - please share if you have one in mind! I would rather have a slightly mediocre UX than to be broken in these environments, though.

bleggett: > IIUC, this requires user to be aware of the nics clearly. So It is a very bad ux, how does a user know the underlying implement details, they are not platform operator, but just application developer or deployer.

I agree, you can't use this unless you know exactly what you're doing, which was a flaw shared by the sidecar annotation this is based off of.

We can't effectively autodetect this tho, I don't think. While we could plausibly autodetect whether an interface was virtual, we probably cannot unconditionally/safely do this for every virtual interface we find.

It's a workaround for specific 3rd-party tooling.

bleggett: (also worth noting - we can mitigate some of the UX in some cases with `platform profiles`, like we do for CNI dir paths and such).

bleggett: > Not sure about this, the use cases are _specific pods_ that have additional nested interfaces. I don't know of any platform that unconditionally has this setup across all pods (though perhapsd they exist)

Yeah - I'm mostly spitballing. But I agree - base case must be per-pod.

hzxuzhonghu: Sorry i cannot agree. It is good to support more platforms and more CNI, but please make sure application developer not aware of this. 

Not only me, i believe most developers outside of solo.io cannot understand correctly what does this do

howardjohn: This setting existed on sidecars for like 5 years - this is just extending it to ambient and with a new name. 

It has *nothing* to do with platforms or CNI, and is not specific to ambient either. 

I was very confused when I first stumbled across it as well, but once you try an application where it's needed it's pretty clear. I suggest you go run a docker-in-docker pod. You will see outbound requests incorrectly get considered inbound. Then apply this/the sidecar one ("kubevirtinterfaces") and it is fixed.

hzxuzhonghu: I run kind cluster for years, how to understand `You will see outbound requests incorrectly get considered inbound. `

howardjohn: It's not about kind, it's the Pod itself has docker running inside of it. Or tail scale, Or kube-virt, or anything that creates network devices without a Pod object

howardjohn: That is why I also say it's nothing to do with platform - it's specific to the application the user is running in a pod

hzxuzhonghu: That's really tricky

bleggett: @hzxuzhonghu I've added this to the Wednesday agenda for a go/no go: https://docs.google.com/document/d/1SMlwliEnthgq7r2PjpLl1kCq3t8rAMbgu6r_lDAXJ0w/edit?tab=t.0#heading=h.yizcbnt4yh96

If you have suggestions or concerns, comments there would be welcome.

bleggett: > Eventually `traffic.sidecar.istio.io/kubevirtInterfaces` should be deprecated.

That one is alpha, so not hard to do, yes.

This label would have to become one of

- `istio.io/reroute-virtual-interfaces`
- `traffic.istio.io/reroute-virtual-interfaces`

as a consequence, but I'm not against that.

bleggett: @hzxuzhonghu - consensus was 

`istio.io/reroute-virtual-interfaces` and we deprecate `traffic.sidecar.istio.io/kubevirtInterfaces` as well.

I'll rework accordingly, thanks!

hzxuzhonghu (approved):

bleggett:
This is good to merge from my perspective, whenever someone with merge perms is ready.

PR #3366, opened 2024-11-20
Update telemetry docs to reflect GAMMA-style attachment doesn't work

craigbox:
This works:

```
apiVersion: telemetry.istio.io/v1
kind: Telemetry
metadata:
  name: logs
  namespace: default
spec:
  targetRefs:
  - kind: Gateway
    name: waypoint
    group: gateway.networking.k8s.io
  accessLogging:
  - providers:
    - name: envoy
```

This does not:

``` 
apiVersion: telemetry.istio.io/v1
kind: Telemetry
metadata:
  name: logs
  namespace: default
spec:
  targetRefs:
  - kind: Service
    name: ratings
    group: ""
  accessLogging:
  - providers:
    - name: envoy
```

### DIFF
--- a/telemetry/v1alpha1/telemetry.pb.go
+++ b/telemetry/v1alpha1/telemetry.pb.go
@@ -562,9 +562,8 @@ type Telemetry struct {
 	// applied to. The targeted resources specified will determine which workloads
 	// the policy applies to.
 	//
-	// Currently, the following resource attachment types are supported:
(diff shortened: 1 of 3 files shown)


keithmattix (approved):

howardjohn (changes requested):
This is simply a bug. The fix is easier than the docs change! I'll send it

PR #3372, opened 2024-12-02
Support grpc probes for workload group health checks

rudrakhp:
Related istio/istio#53085


networking/v1alpha3/workload_group.proto
```diff
@@ -190,6 +192,17 @@ message HTTPHealthCheckConfig {
...
 }
 
+message GrpcHealthCheckConfig {
+    // Port on which the endpoint lives.
+    // +kubebuilder:validation:XValidation:message="port must be between 1-65535",rule="0 < self && self <= 65535"
+    uint32 port = 1;
+
+    // Service is the fully qualified name of the service to send the grpc health check request
+    // +optional
+    // +default=""
```
howardjohn, line 202:
```suggestion
```

We don't use these in Istio and they aren't needed I think

howardjohn (commented):
Looks good! just one small comment

networking/v1alpha3/workload_group.proto
rudrakhp: @howardjohn done!

howardjohn (approved):

zirain (approved):

PR #3374, opened 2024-12-02
Add a new 'workload name' label override

howardjohn:
Today, we have a "workload name" concept that ends up in metrics.
This is defaulted to deriving from deployment metadata (Deployment name,
etc).

This approach is not perfect. We cannot always derive an appropriate
name (for instance, a Pod can be created directly!). WorkloadEntry,
which often is equvilent to Pod, is also using the WE name -- generally
we would want a higher order name there.

To fix this, I propose we add a label to explicitly specify the workload
name. This format mirrors the existing canonical-{service,revision}
label.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -133,14 +133,14 @@ spec:
                       type: string
                       x-kubernetes-validations:
                       - message: wildcard not allowed in label value match
-                        rule: '!self.contains(''*'')'
+                        rule: '!self.contains("*")'
                     description: One or more labels that indicate a specific set of
                       pods/VMs on which a policy should be applied.
                     maxProperties: 4096
                     type: object
                     x-kubernetes-validations:
                     - message: wildcard not allowed in label key match
-                      rule: self.all(key, !key.contains('*'))
+                      rule: self.all(key, !key.contains("*"))
                     - message: key must not be empty
                       rule: self.all(key, key.size() != 0)
                 type: object
@@ -180,8 +180,9 @@ spec:
                 x-kubernetes-validations:
                 - message: Support kinds are core/Service, networking.istio.io/ServiceEntry,
                     gateway.networking.k8s.io/Gateway
-                  rule: '[self.group, self.kind] in [[''core'',''Service''], ['''',''Service''],
-                    [''gateway.networking.k8s.io'',''Gateway''], [''networking.istio.io'',''ServiceEntry'']]'
+                  rule: '[self.group, self.kind] in [["core", "Service"], ["", "Service"],
+                    ["gateway.networking.k8s.io", "Gateway"], ["networking.istio.io",
+                    "ServiceEntry"]]'
               targetRefs:
                 description: Optional.
                 items:
@@ -215,8 +216,9 @@ spec:
                   x-kubernetes-validations:
                   - message: Support kinds are core/Service, networking.istio.io/ServiceEntry,
                       gateway.networking.k8s.io/Gateway
-                    rule: '[self.group, self.kind] in [[''core'',''Service''], ['''',''Service''],
-                      [''gateway.networking.k8s.io'',''Gateway''], [''networking.istio.io'',''ServiceEntry'']]'
+                    rule: '[self.group, self.kind] in [["core", "Service"], ["", "Service"],
+                      ["gateway.networking.k8s.io", "Gateway"], ["networking.istio.io",
+                      "ServiceEntry"]]'
                 maxItems: 16
                 type: array
               type:
@@ -235,10 +237,9 @@ spec:
                 type: string
                 x-kubernetes-validations:
                 - message: url must have schema one of [http, https, file, oci]
-                  rule: 'isURL(self) ? (url(self).getScheme() in ['''', ''http'',
-                    ''https'', ''oci'', ''file'']) : (isURL(''http://'' + self) &&
-                    url(''http://'' +self).getScheme() in ['''', ''http'', ''https'',
-                    ''oci'', ''file''])'
+                  rule: |-
+                    isURL(self) ? (url(self).getScheme() in ["", "http", "https", "oci", "file"]) : (isURL("http://" + self) &&
+                    url("http://" + self).getScheme() in ["", "http", "https", "oci", "file"])
               verificationKey:
                 type: string
               vmConfig:
@@ -272,7 +273,7 @@ spec:
                       type: object
                       x-kubernetes-validations:
                       - message: value may only be set when valueFrom is INLINE
-                        rule: '(has(self.valueFrom) ? self.valueFrom : '''') != ''HOST''
+                        rule: '(has(self.valueFrom) ? self.valueFrom : "") != "HOST"
                           || !has(self.value)'
                     maxItems: 256
                     type: array
@@ -285,7 +286,8 @@ spec:
             type: object
             x-kubernetes-validations:
             - message: only one of targetRefs or selector can be set
-              rule: (has(self.selector)?1:0)+(has(self.targetRef)?1:0)+(has(self.targetRefs)?1:0)<=1
+              rule: '(has(self.selector) ? 1 : 0) + (has(self.targetRef) ? 1 : 0)
+                + (has(self.targetRefs) ? 1 : 0) <= 1'
           status:
             properties:
               conditions:
@@ -2214,14 +2216,14 @@ spec:
                       type: string
                       x-kubernetes-validations:
                       - message: wildcard not allowed in label value match
-                        rule: '!self.contains(''*'')'
+                        rule: '!self.contains("*")'
                     description: One or more labels that indicate a specific set of
                       pods/VMs on which a policy should be applied.
                     maxProperties: 4096
                     type: object
                     x-kubernetes-validations:
                     - message: wildcard not allowed in label key match
-                      rule: self.all(key, !key.contains('*'))
+                      rule: self.all(key, !key.contains("*"))
                     - message: key must not be empty
                       rule: self.all(key, key.size() != 0)
                 type: object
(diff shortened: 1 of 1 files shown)


label/labels.yaml
```diff
@@ -49,6 +49,17 @@ labels:
     resources:
       - Pod
 
+  - name: service.istio.io/workload-name
```
keithmattix, line 52: Need to think about this more, but we probably need a warning to advise users to consider label cardinality

howardjohn: Probably so, same applies to the `canonical-{service,revision}` which I can add. Interestingly, usage of this label is often _in order to_ reduce cardinality (where we do not correctly pick an abstract-enough name and end up with random suffix in the default derivation)

costinm:
On Mon, Dec 2, 2024 at 10:49 AM John Howard ***@***.***>
wrote:

> Today, we have a "workload name" concept that ends up in metrics.
> This is defaulted to deriving from deployment metadata (Deployment name,
> etc).
>
> This approach is not perfect. We cannot always derive an appropriate
> name (for instance, a Pod can be created directly!). WorkloadEntry,
> which often is equvilent to Pod, is also using the WE name -- generally
> we would want a higher order name there.
>
Not sure what is the problem - yes, a pod can be created directly and like
WE it has a name.

Why would a user set a label that is different then the WE name ?

I don't think the canonical-{} stuff is ideal solution either - I don't
know the usage number, but besides the auto-added
stuff it is not as common.

Having a separate resource with a selector - without requiring pods to be
injected or deal with the various ways to
set labels in helm - is a better option.

And if the use case is telemetry - using the OTel mechanisms instead of
inventing our own would likely be better too.

howardjohn:
> Not sure what is the problem - yes, a pod can be created directly and like WE it has a name.
> Why would a user set a label that is different then the WE name ?

My pod/workload entry name is `foobar-j871d2-da88a`. Workload name is a metric label. Prometheus explodes.

We have **heuristics** to prevent this, but they are only that -- heuristics. And there is not (unless this PR lands) a way to override the heuristics. The intent here is to provide a reliable way to get sane behavior when the heuristic fails.


> Having a separate resource with a selector is better

I disagree:
* I haven't seen any data/anecdotes/etc indicating this, aside from you saying it, which doesn't match my intuition on what users prefer
* Its >10x harder to implement that way
* Its inconsistent with how the other fields, which have extremely similar functionality, are implemented

> using the OTel mechanisms instead of inventing our own would likely be better too.

There is not one that I know of.

howardjohn:
> For this particular case, i would think if no deployment or job found, we can fallback to service name. In most cases, they may have same name.

this is injected into the pod so getting service name is impossible

label/labels.yaml
```diff
@@ -49,6 +49,17 @@ labels:
...
+    featureStatus: Alpha
+    description: |
+      The workload name of the application a workload belongs to. If unset, defaults to the detect parent resource.
+      For example, a `Pod` resource may default to the `Deployment` name.
```
linsun, line 56: worth clarify this is used for metrics only?

costinm:
On Mon, Dec 2, 2024 at 5:43 PM John Howard ***@***.***> wrote:

> For this particular case, i would think if no deployment or job found, we
> can fallback to service name. In most cases, they may have same name.
>
> this is injected into the pod so getting service name is impossible
>
Getting the service name is not impossible - pod has labels, service has
label selectors.
Yes, a pod may be selected by multiple services - but it's telemetry, not
authorization or rocket science.
And if it doesn't match a service - it is also fine to just not set that
particular dimension - we are not
breaking anything since the attribute doesn't exist yet.

The proper way to handle that is to work with open telemetry and figure out
an API that is independent
of Istio - ideally using more precise label selectors in a separate CRD, if
having such dimensions in corner cases is
that critical, which I personally doubt. I don't mind an API to
group/organize pods without deployments - my
concerns are with even more annotations on the pod (that is a pretty bad
API - even if we use it extensively)
and is another istio-only telemetry.

We are talking about pods without deployment/job (not so common) - we
already track the service account
for client metrics, for servers we should be able to track the VIP and/or
service name in http case.

howardjohn:
> And if it doesn't match a service - it is also fine to just not set that particular dimension - we are not breaking anything since the attribute doesn't exist yet.

The attribute does exist. I am not proposing adding a new attribute, I would be strongly against that. The problem is users today get a label set to WorkloadEntry/Pod name, or something else that is dynamic, which is extremely unsafe. This is trivial to reproduce, try to use an WorkloadEntry and you will see its name show up in the Prometheus labels. 

> Getting the service name is not impossible - pod has labels, service has label selectors.

Even if we hypothetically could, this attribute is something that is derived at startup/injection time so its practically infeasible. Not to mention the value is **workload** which is intentionally distinct from "service" as a concept in Istio.

louiscryan:
Its pretty clear the OTel can't be a solution for this issue since it can only operate on data that is in-band to the pipeline and an alternate semantic name could never be.

The proposal here mimics what kubernetes has done for itself with the app.kubernetes.io/name label per https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Could we use that instead?

howardjohn:
> The proposal here mimics what kubernetes has done for itself with the app.kubernetes.io/name label per https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

We currently have these labels in our metadata:

- workload: what this PR is about. Typically this is the Deployment name, but it is a heuristic. Falls back to WE name or Pod name if we cannot derive a "parent" resource. 
- app: Logical "application" name. Made obsolete by 'canonical-service'
- canonical_service: `service.istio.io/canonical-name` | `app.kubernetes.io/name`| `app` | workload name. Meant to be the new more modern way to reference an "application".
- version: logical "application" version. Made obsolete by 'canonical-revision'
- canonical_revision: `service.istio.io/canonical-revision` |  `app.kubernetes.io/version` |  `version` | "latest" . Meant to be the more modern way to reference a "version"
- workload_namespace: namespace
- principal: spiffe
- cluster: cluster

So effectively we have 2 concepts:
* Application - a name + a version, encoded as a legacy and modern label
* Workload - an infrastructure thing, not the logical application

One may argue we should just have one concept, application. Maybe so, but IMO its pretty much too late to change it.

Notably, if we used `app.kubernetes.io/name` we are making a backwards incompatible change - users would no longer have a distinction between "application" and "workload". We **know** users are using this label as well, so they would unexpectedly get this new behavior on upgrade.

Using a new label doesn't have this issue since (1) its clearly scoped to just workload ( `app.kubernetes.io/name`  impacts canonical service) and (2) its a new label so no one is using it yet.

howardjohn:
The expectation here is that a user will set this to point to the object type of the highest level of workload orchestration they have.

This could be a WorkloadGroup, a Deployment, an ArgoCD application, etc.

In many of these cases we can auto detect it. In fact, with WorkloadGroup we could (there is a "TODO" in the codebase to do this, which I will also address). But not all -- sometimes Pods and WorkloadEntry are created in a such a way that there is no way to determine their parent automatically. In these cases, it is imperative that we allow users a way to not place an unstable ID into a Prometheus label.

louiscryan:
I generally agree with what's said above. 

We use the ability of Deployment to copy labels down onto pods to handle the "provide a semantic name" thing and we support a plethora of labels via that mechanism to align with common usage patterns.

WorkloadGroup plays the same role as Deployment here but we don't have an auto-copying mechanism and WorkloadEntry is allowed to exist without a WorkloadGroup. Assuming we keep using the same label definitions and derivation mechanisms I see no reason not support reading those labels of WorkloadEntry just as we would for a directly scheduled Job or Pod

costinm:
Can we set it to the workload group instead of workload entry ? I don't
mind changes to Istio APIs, my concern was Pod, and dependency on injection.

Also: are we creating more difficulty on adoption of ambient ? At some
point we may need to reduce the set of labels that only work with
sidecar/injection.

I'm not trying to block this or make it too complicated - just to make sure
we understand the implications.

costinm:
And to be clear - WorkloadEntry labels are sent in CDS/EDS and computed by
Istiod, I don't think they depend on injection, so not sure what is the
connection. For dest labels - almost everything is under Istiod control.

howardjohn:
I think it will answer a lot of questions to just send a draft PR showing
exactly what changes will be made. I'll send one next week.

howardjohn:
Here is the implementation: https://github.com/istio/istio/pull/54340.

WorkloadName is used in 3 places:
1. Bootstrap for envoy, so the proxy knows how to emit source_workload and send it over metadata exchange
2. WDS for Ztunnel, so it knows how to emit {source,destination}_workload
3. EDS for envoy proxies, so it knows how to emit destination_workload **as a fallback if metadata exchange does not have it**

> Can we set it to the workload group instead of workload entry 

The above PR is adding this. It **sort of** already exists because `istioctl` will read the WorkloadGroup and generate a bootstrap config that has `ISTIO_META_WORKLOAD_NAME` which hits case (1) and sends over MX. This doesn't work with HBONE though. The above PR makes sure all 3 cases fallback to WorkloadGroup name

costinm:
Good, that works, my concern was on depending on injection or Pod
modifications. As long as it's in WDS/EDS - no problem.
Not so much concerned about sidecar behavior, it already depends on
injection and lots of other annotations - just want to make sure a pure
ambient
can work 100% without requiring pod annotation.

costinm:
Added a comment on your PR - 99% good.

label/labels.yaml
linsun: @howardjohn any AI from @keithmattix's comment? Wasn't sure if the warning should be added from your comment.

label/labels.yaml
linsun: Saw other labels don't say metrics only, so going to mark this as not needed

linsun (approved):
lgtm - don't see outstanding comments unresolved, except Keith's comment which I am not exactly sure if needs any resolution.

costinm (approved):
No concerns with having the label itself, we have plenty of others.

PR #3376, opened 2024-12-04
add domain name for virtual host match in envoy filters

ramaraochavali:


zirain (approved):

networking/v1alpha3/envoy_filter.proto
```diff
@@ -569,6 +569,9 @@ message EnvoyFilter {
       // registry.
       string name = 1;
 
+      // Match a specific domain name in a virtual host.
+      string domain_name=3;
```
howardjohn, line 573: what is the use case for using this vs name? Does it do wildcards or only exact match (like can I have a wildcard here to match a wildcard domain)? Does it consider port (I think we configure envoy to ignore port?)?

ramaraochavali: The use case is - if we enable route collapse optimization with different domain names using same routes, we pick one route name and merge. 

Yes. It can match wildcard names and port is not included.

hzxuzhonghu: When is the name not able to distinguish the VHost?

ramaraochavali: Actual Routes"

```
 {
     "version_info": "2024-11-01T23:45:05Z/17931",
     "route_config": {
      "@type": "type.googleapis.com/envoy.config.route.v3.RouteConfiguration",
      "name": "https.10443.xxxx",
      "virtual_hosts": [
       {
        "name": "*.au:10443",
        "domains": [
         "*.au"
        ],
        "routes": [..]
       },
       {
        "name": "*.com:10443",
        "domains": [
         "*.com",
         "*"
        ],
        "routes": [..]
       },
      ]
     }
}
```
after merge
```
     {
      "version_info": "2024-11-01T23:45:05Z/17931",
      "route_config": {
       "@type": "type.googleapis.com/envoy.config.route.v3.RouteConfiguration",
       "name": "https.10443.xxxx",
       "virtual_hosts": [
        {
         "name": "*.com:10443",
         "domains": [
          "*.au",
          "*.com",
         ],
         "routes": [..]
        }
       ]
      }
```

howardjohn: So in this case, i put `match.domainName = "*.au"`? And it will apply to `*.com` as well though?

ramaraochavali: Yes. I think the semantics here should be domain list contains this "*.au" as they happen to share the same route lis t not "exactly" match this domain.

I think one thing we can do without any API change is add the list of virtual hosts that gets merged to in metadata field here https://www.envoyproxy.io/docs/envoy/v1.32.1/api-v3/config/route/v3/route_components.proto.html - like merged_vhosts and use that list in Envoy filters in the match criteria - so basically enhance vhost match.

howardjohn: OK i think the semantics are fine we just need to document it

ramaraochavali: Updated . PTAL

linsun: Hi @ramaraochavali can you please add an example on how to use this in the API doc?  Also probably needs a release note as well.

ramaraochavali: I will add example. Release notes planning to add it for the implementation PR

hzxuzhonghu: Seems only me is still confusing, Isn't `name` already unique to distinguish the VirtualHost

ramaraochavali: it is not. When we have routes collapsed for different domains, we pick one virtual host and combine domains - So it is not possible to uniquely identify a virtual host by name for a domain.

ramaraochavali: @linsun  added example. PTAL

ramaraochavali: @linsun can you PTAL when you get chance?

ramaraochavali:

ramaraochavali:
@istio/technical-oversight-committee @linsun Can you PTAL?

howardjohn (approved):

PR #3384, opened 2024-12-11
Fix some proto descriptions

craigbox:
A bunch of tidying up:

* put Telemetry API in its place; remove a heading that caused rendering errors
* `s/Pilot/istiod/`
* fix a comment that caused page to wrap
* made some fields be `code` in some descriptions

 Generated files are using the old HTML generator at this point.


networking/v1alpha3/destination_rule.proto
```diff
@@ -717,7 +719,7 @@ message OutlierDetection {
   int32 consecutive_errors = 1 [deprecated=true];
 
   // Determines whether to distinguish local origin failures from external errors. If set to true
-  // consecutive_local_origin_failure is taken into account for outlier detection calculations.
+  // `consecutive_local_origin_failure` is taken into account for outlier detection calculations.
```
hzxuzhonghu, line 722: Maybe change to consecutiveLocalOriginFailure, which is used by users

networking/v1alpha3/destination_rule.proto
```diff
@@ -738,11 +740,11 @@ message OutlierDetection {
   // events qualify as a gateway error.
   // This feature is disabled by default or when set to the value 0.
   //
-  // Note that consecutive_gateway_errors and consecutive_5xx_errors can be
+  // Note that `consecutive_gateway_errors` and `consecutive_5xx_errors` can be
```
hzxuzhonghu, line 743: same for the below fields

networking/v1alpha3/destination_rule.proto
craigbox: Thanks, picked up a few more too. 
I would recommend someone goes through each doc and does a proper revision, but I'm just fixing obvious things now.

hzxuzhonghu (approved):

howardjohn (approved):
nice!

PR #3389, opened 2024-12-14
New format docs for 1.24

craigbox:
#3338 but for 1.24.

I think this one has an updated container in prow already?


craigbox:
These changes were rolled into https://github.com/istio/api/pull/3385.

PR #3397, opened 2024-12-27
add label selector to DestinationRule subset config

shulin-sq:


networking/v1alpha3/destination_rule.proto
```diff
@@ -337,6 +337,8 @@ message Subset {
   // specified at the subset level will override the corresponding settings
   // specified at the DestinationRule level.
   TrafficPolicy traffic_policy = 3;
+
+  istio.type.v1beta1.LabelSelector selector = 4;
```
hzxuzhonghu, line 341: add some ducuments, what is the relationship with `labels` field

PR #3414, opened 2025-01-16
Validation: remove CEL for PolicyTargetRef to allow vendor extensions

ilrudie:
This PR removes CEL validation from PolicyTargetRef for two primary reasons:

1. Allow vendor extensions by allowing Istio policy to target non-Istio references
2. Allow Istio policy to target a GatewayClass

### DIFF
--- a/releasenotes/notes/3412.yaml
+++ b/releasenotes/notes/3412.yaml
@@ -5,4 +5,4 @@ issue:
   - https://github.com/istio/istio/issues/54696
 releaseNotes:
 - |
-  **Added** GatewayClass.gateway.networking.k8s.io as a valid PolicyTargetReference
+  **Removed** CEL validation of group/kind for PolicyTargetReference to enable vendor extensions
--- a/releasenotes/notes/3412.yaml
+++ b/releasenotes/notes/3412.yaml
@@ -1,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+issue:
+  - https://github.com/istio/istio/issues/54696
+releaseNotes:
+- |
+  **Added** GatewayClass.gateway.networking.k8s.io as a valid PolicyTargetReference
--- a/type/v1beta1/selector.proto
+++ b/type/v1beta1/selector.proto
@@ -106,7 +106,9 @@ enum WorkloadMode {
 //         methods: ["POST"]
 //         ports: ["8080"]
 // ```
-// +kubebuilder:validation:XValidation:message="Support kinds are core/Service, networking.istio.io/ServiceEntry, gateway.networking.k8s.io/Gateway",rule="[self.group, self.kind] in [['core','Service'], ['','Service'], ['gateway.networking.k8s.io','Gateway'], ['networking.istio.io','ServiceEntry']]"
+// 
+// When binding to a GatewayClass resource using PolicyTargetReference, your policy must be in the root namespace.
+// +kubebuilder:validation:XValidation:message="Support kinds are core/Service, networking.istio.io/ServiceEntry, gateway.networking.k8s.io/Gateway, gateway.networking.k8s.io/GatewayClass",rule="[self.group, self.kind] in [['core','Service'], ['','Service'], ['gateway.networking.k8s.io','Gateway'], ['networking.istio.io','ServiceEntry'], ['gateway.networking.k8s.io','GatewayClass']]"
 message PolicyTargetReference {
   // group is the group of the target resource.
   // +kubebuilder:validation:MaxLength=253
--- a/type/v1beta1/selector.proto
+++ b/type/v1beta1/selector.proto
@@ -108,8 +108,6 @@ enum WorkloadMode {
 // ```
 // 
 // When binding to a GatewayClass resource using PolicyTargetReference, your policy must be in the root namespace.
-//
-// Supported kinds are core/Service, networking.istio.io/ServiceEntry, gateway.networking.k8s.io/Gateway, and gateway.networking.k8s.io/GatewayClass
 message PolicyTargetReference {
   // group is the group of the target resource.
   // +kubebuilder:validation:MaxLength=253
--- a/type/v1beta1/selector.proto
+++ b/type/v1beta1/selector.proto
@@ -108,7 +108,8 @@ enum WorkloadMode {
 // ```
 // 
 // When binding to a GatewayClass resource using PolicyTargetReference, your policy must be in the root namespace.
-// +kubebuilder:validation:XValidation:message="Support kinds are core/Service, networking.istio.io/ServiceEntry, gateway.networking.k8s.io/Gateway, gateway.networking.k8s.io/GatewayClass",rule="[self.group, self.kind] in [['core','Service'], ['','Service'], ['gateway.networking.k8s.io','Gateway'], ['networking.istio.io','ServiceEntry'], ['gateway.networking.k8s.io','GatewayClass']]"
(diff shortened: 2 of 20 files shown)


type/v1beta1/selector.proto
```diff
@@ -106,7 +106,10 @@ enum WorkloadMode {
 //         methods: ["POST"]
 //         ports: ["8080"]
 // ```
-// +kubebuilder:validation:XValidation:message="Support kinds are core/Service, networking.istio.io/ServiceEntry, gateway.networking.k8s.io/Gateway",rule="[self.group, self.kind] in [['core','Service'], ['','Service'], ['gateway.networking.k8s.io','Gateway'], ['networking.istio.io','ServiceEntry']]"
+// 
+// When binding to a GatewayClass resource using PolicyTargetReference, your policy must be in the root namespace.
+//
+// Supported kinds are core/Service, networking.istio.io/ServiceEntry, gateway.networking.k8s.io/Gateway, and gateway.networking.k8s.io/GatewayClass
```
howardjohn, line 112: This is documented on the individual usages. Like

```
networking/v1alpha3/envoy_filter.proto
871:  repeated istio.type.v1beta1.PolicyTargetReference targetRefs = 6;
```

We should update those. We might want to only include it there, in case there are per-policy supported types? If we know all will be the same then here is probably good

ilrudie: Good shout. Actually, looking at this more closely it's probably most clear now with no CEL to simply not mention supported group/kind of the PolicyTargetRef message at all. Support likely varies by where it is used and should be mentioned where this message is used going forward.

releasenotes/notes/3412.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: release-notes/v2
+kind: feature
+area: traffic-management
+issue:
+  - https://github.com/istio/istio/issues/54696
+releaseNotes:
+- |
+  **Removed** CEL validation of group/kind for PolicyTargetReference to enable vendor extensions
```
keithmattix, line 8: If we're removing the CEL validation, should we add this to the OSS ValidatingWebhook?

ilrudie: Maybe status, "you tried to bind to a beep.boop.bop.io/beepboop and Istiod has not honored this"

keithmattix: Yeah that would work; just some indication since CEL isn't going to be enforcing it now

type/v1beta1/selector.proto
ilrudie: Would we want to allow binding resources which are not Authorization Policy to a GatewayClass?

ilrudie: I moved supported group/kind from being centralized on the message to being in our higher lever resources where the PolicyTargetRef is used. TBH, after the move I'm feeling less certain about doing it this way but it is an improvement over having inconsistent group/kind information in different contexts.

releasenotes/notes/3412.yaml
ilrudie: Got this in the VWH in https://github.com/istio/istio/pull/54780

howardjohn (approved):

LiorLieberman:
@ilrudie I followed up internally with a couple more Gateway folks and I actually share some of the concerns. The main one being - that AuthorizationPolicy lives in a namespace and it will have a global effect if you target a gatewayclass - which sounds very odd. Ideally, we want to encourage cluster-scope resources triggering global effect.

I know we have prior art for it in istio with the root namespace, but was hoping to avoid it with gateway apis if we can.

howardjohn:
@LiorLieberman the only way to achieve that is to duplicate every single CRD we have to add a `ClusterXXX` which is pretty untenable.

LiorLieberman:
The conversation during last week's meeting branched into talking about how to achieve default deny and the current implementation (what Louis brought up). If we come back to the specific usecase that led to this discussion, I think if we were to implement what Louis said (default deny is presented when introducing waypoint) - it should be enough to satisfy that usecase.

LiorLieberman:
I agree, but I think it deserves more thought. And perhaps to be solved in gateway. If there was a concrete need to implement it right away, and we (istio) did not want or did not have the time to wait for gateway to catch up - this would have make sense. But - I think the correct way to solve the specific problem that brought this up - is not by doing this. (Plus, as said - this will have wider implication which we may be able to avoid - or design differently)

howardjohn:
Are there other thoughts on how to achieve this? We have been thinking about this for like 3 years, I haven't actually heard any proposal _except_ this.

keithmattix:
> Are there other thoughts on how to achieve this? We have been thinking about this for like 3 years, I haven't actually heard any proposal _except_ this.

+1 - regardless of what Gateway decides specifically, the only way to get around duplicating all of your CRDs is to limit GatewayClass binding to a privileged namespace. That's what we're doing here.

LiorLieberman:
I am not trying to present a strong opposition to this idea. Instead, I think this is not the right solution for the problem. 

As a user, when I want to have default deny - I want to have one policy - I dont want to worry about creating new policies when presenting waypoint in the middle. So ideally I will have one authorization policy that denies everything (maintaining the status-quo today - in the istio-system namespace, not trying to change that). This is much a simpler UX, and a step forward to resolve current confusions around authzpolicies today in ambient.

I definitely dont think thats the only way, for example, putting it in meshConfig is another option - I dont advocate for it, and it has probably been suggested in the past - this is just the first alternative came to mind.

LiorLieberman:
@howardjohn  The two objections that were presented in the meeting for ParametersRef are:

> 1. The GatewayClass is something that the implementation is managing and the user should not touch it. 

I think thats also somehow valid for the inverse (whats being proposed in this PR). I agree it is slightly better that the user should not edit it, but only reference it for this concern. But in anycase the user has to be fully aware of this.

2. The need to bind multiple things in ParametersRef. If thats the concern, we can argue to make this a list upstream. I dont see any problem with this.

I think the paramsref approach maintains cluster-scope object triggers cluster-wide effects, which is more straightforward and has better UX IMO.

howardjohn:
> As a user, when I want to have default deny - I want to have one policy - I dont want to worry about creating new policies when presenting waypoint in the middle. So ideally I will have one authorization policy that denies everything (maintaining the status-quo today - in the istio-system namespace, not trying to change that). This is much a simpler UX, and a step forward to resolve current confusions around authzpolicies today in ambient.

Ah I see and agree. This is not the purpose of this proposal, we are also planning to have a different change where waypoint applies L4 policies like ztunnel _and_ then L7 policies from targetRef. More details coming here very soon

> I think thats also somehow valid for the inverse (whats being proposed in this PR).

If so, users shouldn't reference a GatewayClass in Gateway, but we force them to.

LiorLieberman:
>If so, users shouldn't reference a GatewayClass in Gateway, but we force them to.

Right, not saying its perfect, I am saying that this happens today, and it is not great. The fact that it exists does not mean its a great to build on it. I was convinced that ParametersRef are better approach here. 

Can you relate to why you against it? I referred to two of the concerns you brought up in the meeting. Anything else I did not consider?

LiorLieberman:
>Ah I see and agree. This is not the purpose of this proposal, we are also planning to have a different change where waypoint applies L4 policies like ztunnel and then L7 policies from targetRef. More details coming here very soon

Thats great, and again, I think this change is the right way to solve the usecase @ilrudie brought up.

keithmattix (approved):

PR #3417, opened 2025-01-17
AuthorizationPolicy serviceAccount: allow same namespace

howardjohn:
This allows applying a policy like

```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
 name: allow-waypoint
spec:
  selector:
    matchLabels:
      app: echo
  action: ALLOW
  rules:
  - from:
    - source:
        serviceAccounts: ["waypoint"]
```

To get the semantics of "allow the waypoint in the same namespace". 

It is common in k8s to allow configurations to be written without a namespace specified and have the namespace of references implied by the namespace it is eventually written to. This allows taking the same policy and applying it to many places without the need for modification/templating. As far as I know, this is allowed with virtually every k8s core type and Istio type today.

Note this just changes docs to match the implementation in https://github.com/istio/istio/pull/54745 (which has a "hold" until this is approved)

Note: a release note is skipped since we have not shipped this feature

### DIFF
--- a/security/v1beta1/authorization_policy.pb.go
+++ b/security/v1beta1/authorization_policy.pb.go
@@ -640,6 +640,7 @@ type Source struct {
 	// This field requires mTLS enabled and is the same as the `source.serviceaccount` attribute.
 	//
 	// This takes the format `<namespace>/<serviceaccount>`.
+	// `<serviceaccount>` may also be used to use the same namespace as the `AuthorizationPolicy`.
 	//
 	// If not set, any service account is allowed.
 	//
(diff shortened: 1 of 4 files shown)


keithmattix (commented):
LGTM, but I'm curious why I don't see any v1 changes generated?

howardjohn:
> LGTM, but I'm curious why I don't see any v1 changes generated?

v1 is just type aliases. These are done struct level. This is changing a field _within_ a struct, so there is no change

keithmattix (approved):

hzxuzhonghu (approved):

PR #3418, opened 2025-01-17
improve documentation for tag and rev label

therealmitchconnors:
As far as I know, this reflects the current state of these APIs.  I'm not sure how this got so out of date...

### DIFF
--- a/label/labels.yaml
+++ b/label/labels.yaml
@@ -61,20 +61,22 @@ labels:
       - WorkloadEntry
 
   - name: istio.io/rev
-    featureStatus: Alpha
-    description: Istio control plane revision associated with the resource; e.g. `canary`
+    featureStatus: Beta
+    description: Istio control plane revision or tag associated with the resource; e.g. `canary`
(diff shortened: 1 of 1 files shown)


label/labels.yaml
```diff
@@ -61,20 +61,22 @@ labels:
...
   - name: istio.io/rev
-    featureStatus: Alpha
-    description: Istio control plane revision associated with the resource; e.g. `canary`
+    featureStatus: Beta
+    description: Istio control plane revision or tag associated with the resource; e.g. `canary`
     hidden: false
     deprecated: false
     resources:
       - Namespace
+      - Gateway
```
howardjohn, line 70: Hmm, we may want to start disambiguating between k8s GW and Istio GW

costinm (approved):

label/labels.yaml
therealmitchconnors: True, but in this case, I believe this works with istio or k8s gateways.

PR #3420, opened 2025-01-21
Add missing annotation

avadhmr:
added the annotation as every other CRD has that.


zirain:
please fix CLA, you may need run `make gen`.

howardjohn (approved):

PR #3426, opened 2025-01-31
envoy file provider support `omit_empty_values`

zirain:
xref: https://github.com/istio/istio/issues/54930


mesh/v1alpha1/config.proto
```diff
@@ -942,6 +942,11 @@ message MeshConfig {
 
       // Optional. Allows overriding of the default access log format.
       LogFormat log_format = 2;
+
+      // Optional. If set to true, when command operators are evaluated to null,
+      // For text format, the output of the empty operator is changed from “-“ to an empty string.
+      // For json format, the keys with null values are omitted in the output structure.
+      bool omit_empty_values = 3;
```
hzxuzhonghu, line 949: by default is it is empty, isn't it difficult to distinguish which field is empty?

hzxuzhonghu: So can we only make it work for json

zirain: I think it's fine to give user the option, instead of make an exception.
User can use customized separator, e.g. `[%CEL(xxxxxx)%]`.

zirain:
kindly ping @istio/technical-oversight-committee

keithmattix (commented):
A couple of things:
1: I'm not totally sure what the use case is even after reading the linked issue.
2. Why MeshConfig over Telemetry API?

zirain:
> I'm not totally sure what the use case is even after reading the linked issue.

Before, envoy will send `"key": ""` in json
After, envoy will remove the field if the value is empty.

> Why MeshConfig over Telemetry API?

`Provider` (in MeshConfig) is part of Telemetry API.
In the past, we talked about make it a separated CRD or else, but didn't make a aggrement.

keithmattix (approved):

PR #3430, opened 2025-02-04
Update refdocs to use markdown links and standardize HTTP/2 notation in master

AdeshGhadage:
Converted inline URLs to markdown format in refdocs. Standardized "HTTP2" to "HTTP/2". Changes made in the master branch of istio/api. This aligns with issue [#16102](https://github.com/istio/istio.io/issues/16102).

PR #3431, opened 2025-02-04
Update refdocs 1.24  to use markdown links and standardize HTTP/2 notation 

AdeshGhadage:
Converted inline URLs to markdown format in refdocs. Standardized "HTTP2" to "HTTP/2". Changes made in the release branch of istio/api. This aligns with [issue](https://github.com/istio/istio.io/issues/16102).


mikemorris:
@craigbox Is separate PRs the expected approach here, or should #3430 be backported?

craigbox:
Yes; @AdeshGhadage, this should be fixed in #3430, pass all the tests, etc.
We may not need to cherry pick if this lands before 1.25.

PR #3432, opened 2025-02-04
add retry host predicate api

ramaraochavali:
We removed retry host predicate via this PR https://github.com/istio/istio/commit/3727b5719435c0cb4146413ac02ac16f8378cd36 to fix a bug that is breaking consistent hash on retry. But there are some use cases which rely on this behaviour. So this PR is to propose an API for it with default to current behaviour and people can change if needed. This is similar to other retry plugin i.e. retryRemoteLocaliities.


networking/v1alpha3/virtual_service.proto
```diff
@@ -1313,6 +1313,10 @@ message HTTPRetry {
   // Flag to specify whether the retries should retry to other localities.
   // See the [retry plugin configuration](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#retry-plugin-configuration) for more details.
   google.protobuf.BoolValue retry_remote_localities = 4;
+
+  // Flag to specify whether the retries should ignore previously tried hosts during retry.
+  // Defaults to true.
+  google.protobuf.BoolValue retry_ignore_previous_hosts = 5;
```
hzxuzhonghu, line 1319: how about `retry_previous_hosts`, and make it false by default

ramaraochavali: We can. It seems little confusing to me. But fine either way

ramaraochavali:

ramaraochavali:
Done

keithmattix (approved):

zirain (approved):

PR #3433, opened 2025-02-06
Clarify relationship btw MeshNetworks and ENABLE_HCM_INTERNAL_NET

jaellio:
Add comments explaining the ability to use MeshNetworks to configure Envoy's internal_address_config via
ENABLE_HCM_INTERNAL_NETWORK

https://github.com/istio/istio/issues/53402

Related: https://github.com/istio/istio/issues/52037

### DIFF
--- a/mesh/v1alpha1/network.proto
+++ b/mesh/v1alpha1/network.proto
@@ -116,6 +116,18 @@ message Network {
 //
 // If `ENABLE_HCM_INTERNAL_NETWORKS` is set to true, MeshNetworks can be used to
 // to explicitly define the networks in Envoy's internal address configuration.
+// Envoy uses the IPs in the `internalAddressConfig` to decide whether or not to sanitize
+// Envoy headers. If the IP address is listed an internal, the Envoy headers are not
+// sanitized. As of Envoy 1.33, the default value for `internalAddressConfig` is set to
+// an empty set. Previously, the default value was the set of all private IPs. Setting
+// the `internalAddressConfig` to all private IPs (via Envoy's previous default behavior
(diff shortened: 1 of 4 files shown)


mesh/v1alpha1/network.proto
```diff
@@ -114,6 +114,9 @@ message Network {
 //       locality: us-east-1a
 // ```
 //
+// If `ENABLE_HCM_INTERNAL_NETWORKS` is set to true, MeshNetworks can be used to
```
keithmattix, line 117: Do you think we need an example of how to do this?

jaellio: If `Networks` defaults to all mesh IPs then maybe not, but I need to clarify the default value of `Networks`

ramaraochavali: I the documentation is correct. But I would hesitate to add the temporary feature flag reference here.

jaellio: I liked having the information here so it would also be clear in our api docs on istio.io. Otherwise, the only documentation on how to utilize MeshNetworks to preserve envoy internal headers. I could modify/simplify the description and link to the upgrade note?

ramaraochavali: I think specifying MeshNetworks can also be used to define "internal" addresses of mesh and provide an example is fine.

keithmattix: I agree with @jaellio that the feature flag + example is useful//necessary here. Users who run into the Envoy behavior change are going to come to Istio documentation to tell them what to do. We should have a clear answer for them

howardjohn (approved):

keithmattix (approved):

PR #3435, opened 2025-02-10
[release-1.25] #3361 manual cherry

bleggett:
https://github.com/istio/api/pull/3427 seems stuck for whatever reason, so let's test a manual cherry.


bleggett:
Okay there's some prow misconfig here that's causing both this and https://github.com/istio/api/pull/3427 to fail, looking.

bleggett:
I think both of these will fail CI until https://github.com/istio/api/pull/3434/files goes in, so closing this one, once that goes in I'll resync.

PR #3438, opened 2025-02-10
Added traffic.istio.io/ingress-public-address annotation

schurcht:
Adding a service annotation that allows users to override the IP Istio will use to reach out to gateways.
This enables you to use a private load balancer with a public IP that DNATs to it.

### DIFF
--- a/annotation/annotations.gen.go
+++ b/annotation/annotations.gen.go
@@ -784,6 +784,21 @@ Accepted values:
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


remiminnebo (approved):

howardjohn (commented):
What is this actually intended to be used by?

keithmattix:
I strongly suspect this is with respect to #55041

hzxuzhonghu:
not needed

PR #3452, opened 2025-03-04
Add service_protocol field to Sidecar API

wtzhang23:
Istio supports [Ingress Sidecar Termination](https://www.google.com/search?q=tls+termination+sidecar&rlz=1C5GCEM_en&oq=tls+termination+sidec&gs_lcrp=EgZjaHJvbWUqBwgAEAAYgAQyBwgAEAAYgAQyBggBEEUYOTIGCAIQRRhAMg0IAxAAGIYDGIAEGIoFMgYIBBBFGDwyBggFEEUYPTIGCAYQRRg9MgYIBxBFGEDSAQg0MjgwajBqMagCALACAA&sourceid=chrome&ie=UTF-8), which allows configuring sidecars to terminate using certs provisioned elsewhere given that the `ENABLE_TLS_ON_SIDECAR_INGRESS` feature flag is set to true. This helps new users of Istio take advantage of Istio features such as inbound stats collection and EnvoyFilters injecting rate limiting while migrating their infrastructure stacks over the span of multiple years. Note that migration cannot be accelerated as it requires onboarding every client, and we have many services that have a high fan-in count (e.g. configuration services, facades over user databases).

However, the current implementation **only supports passing traffic to the service over HTTP/1.1**. As gRPC and other HTTP2-based protocols gain popularity, this means that users cannot use this feature out-of-the-box. EnvoyFilters can be used to patch this.

This PR proposes an ergonomic change which allows specifying the `service_protocol` that differs from the protocol specified in the sidecar port. The following [changes](https://github.com/istio/istio/compare/master...wtzhang23:istio:sidecar-explicit-after-termination-protocol) (original closed [PR](https://github.com/istio/istio/pull/53396)) show how this would be implemented:
* If blank, continue as was before
* If the sidecar protocol is HTTPS, allow `service_protocol` to be either http or http2
* If the sidecar protocol is TLS, allow `service_protocol` to be a TCP protocol (e.g. Redis)
* Else, the sidecar protocol must match the sidecar port protocol if specified.


networking/v1alpha3/sidecar.proto
```diff
@@ -466,6 +466,10 @@ message IstioIngressListener {
   // sidecar for requests originating from outside the mesh.
   // Currently supports only SIMPLE and MUTUAL TLS modes.
   ServerTLSSettings tls = 7;
+  // The protocol to use when proxying traffic to the service. This is useful
+  // when the incoming connection's protocol differs from the service, such as
+  // when `tls` is set.
```
costinm, line 471: Not sure I understand the comment. Usually service protocol is FTP or SMTP.

wtzhang23: Ya, I think the wording is improper. By saying "protocol" I'm coupling the transport (plaintext/TLS) and the application protocol (HTTP1/2), which as far as I can tell is the only reason why this field is desirable. This PR was proposed simply to address the case:

`port protocol: HTTPS -> service protocol: HTTP`. Maybe it is clearer if I just delete the second sentence?

costinm:
How about using 'http_version' - if all you want to identify is the
version of http to use ?

My concern is with the field name - 'service protocol' can have many
meanings, there are many protocols that support proxy and even
'service' is
an overloaded term ( there are non http services - and k8s/Istio
'services' that implement their own protocols like WebDAV over http ).

HTTP/2 is not quite a different application protocol - just a
different version of the http protocol with different encoding.

Would also help a lot to include some examples of the naming used in
other APIs/servers.

No concern with the feature, seems a good thing - just naming and docs.

wtzhang23:
I would like something a little more generic than `http_version` to "future-proof" the api. I agree `service` is overloaded. What do you think about something like `post_terminated_protocol` (somewhat verbose though) or `unterminated_protocol`, mirroring [AfterTLSTermination](https://github.com/istio/istio/blob/02330c24a33fd453bedabd4746aca8313b9be965/pkg/config/protocol/instance.go#L169-L178).

costinm:
I don't know what future you expect - but at least near term we either have HTTP based protocols ( and plenty of protocols are now on top of HTTP) and all other TCP based protocols. There are other places to specify the top service protocol - in Service object itself, the existing protcol fields, etc.

It may be better for the users to have a clear API to solve the narrow problem of HTTP/1.1 ( H3 is UDP, and H4 is not yet a concern), and deal with future problems in the future :-)

wtzhang23:
I was thinking more in lines of TCP protocols. E.g. if anyone has a (somewhat weird) request to forward Redis, in which case adding Redis listener filters may be desirable. Though agree with you we can keep the scope small for now and we can always deprecate/introduce a more general field when the time comes.

PR #3461, opened 2025-03-08
add retry backoff interval

ramaraochavali:
Configures retry backoff interval. Needed for gateway api also


networking/v1alpha3/virtual_service.proto
```diff
@@ -1317,6 +1317,9 @@ message HTTPRetry {
   // Flag to specify whether the retries should ignore previously tried hosts during retry.
   // Defaults to true.
   google.protobuf.BoolValue retry_ignore_previous_hosts = 5;
+
+  // Specifies the minimum duration between retry attempts.
```
howardjohn, line 1321: Can we indicate something along the lines of "if unset, retries will be instant"? And maybe clarify the interaction with a timeout (e.g. what if my `backoff=2s, timeout=3s, retries=3` - what happens? I assume we can retry at most 1x?)

ramaraochavali: Ok. Updated. PTAL

ramaraochavali:
@howardjohn can PTAL when you get chance?

networking/v1alpha3/virtual_service.proto
```diff
@@ -1317,6 +1317,13 @@ message HTTPRetry {
...
+  // If unset, default minimum duration of 25ms is used as base interval for exponetial backoff.
```
howardjohn, line 1322: Isn't the default no backoff?

howardjohn (commented):
LGTM aside from default comment

networking/v1alpha3/virtual_service.proto
ramaraochavali: No. Envoy's default is 25ms https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto.html#envoy-v3-api-msg-config-route-v3-retrypolicy

howardjohn: Is it valid to set `0s`? I think our standard valdiation will reject it unless we add `protoc-gen-crd:duration-validation:none`. 

Note the GW API says "if unset its implementation specific" so we can keep that at 25ms, but they do allow 0s i think

ramaraochavali: it should be greater than 0s in Envoy. If gateway allows that, we will have to tweak that in Envoy.

zirain (approved):

PR #3466, opened 2025-03-12
add ecdsa certificate support at gateways

ramaraochavali:
For https://github.com/istio/istio/issues/36181.

Also needed for Gateway API

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -380,9 +380,9 @@ message Port {
   uint32 target_port = 4 [deprecated=true];
 }
 
-// +kubebuilder:validation:XValidation:message="only one of credential_names or tls_certificates can be set",rule="oneof(self.tlsCertificates, self.credentialNames)"
-// +kubebuilder:validation:XValidation:message="only one of credential_name or credential_names can be set",rule="oneof(self.credentialName, self.credentialNames)"
-// +kubebuilder:validation:XValidation:message="only one of credential_name or tls_certificates can be set",rule="oneof(self.credentialNames, self.tlsCertificates)"
+// +kubebuilder:validation:XValidation:message="only one of credentialNames or tlsCertificates can be set",rule="oneof(self.tlsCertificates, self.credentialNames)"
+// +kubebuilder:validation:XValidation:message="only one of credentialName or credentialNames can be set",rule="oneof(self.credentialName, self.credentialNames)"
+// +kubebuilder:validation:XValidation:message="only one of credentialName or tlsCertificates can be set",rule="oneof(self.credentialNames, self.tlsCertificates)"
 message ServerTLSSettings {
   // If set to true, the load balancer will send a 301 redirect for
   // all http connections, asking the clients to use HTTPS.
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -6688,15 +6688,15 @@ spec:
                           type: array
                       type: object
                       x-kubernetes-validations:
-                      - message: only one of credential_names or tls_certificates
-                          can be set
+                      - message: only one of credentialNames or tlsCertificates can
+                          be set
                         rule: '(has(self.tlsCertificates) ? 1 : 0) + (has(self.credentialNames)
                           ? 1 : 0) <= 1'
-                      - message: only one of credential_name or credential_names can
+                      - message: only one of credentialName or credentialNames can
                           be set
                         rule: '(has(self.credentialName) ? 1 : 0) + (has(self.credentialNames)
                           ? 1 : 0) <= 1'
-                      - message: only one of credential_name or tls_certificates can
+                      - message: only one of credentialName or tlsCertificates can
                           be set
                         rule: '(has(self.credentialNames) ? 1 : 0) + (has(self.tlsCertificates)
                           ? 1 : 0) <= 1'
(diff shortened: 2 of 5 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -471,6 +471,31 @@ message ServerTLSSettings {
   // or credentialName can be specified.
   string credential_name = 10;
 
+  // Same as CredentialName but for multiple certificates. Mainly used for specifying
+  // RSA and ECDSA certificates for the same server.
+  repeated string credential_names = 14;
```
howardjohn, line 476: Are we going to support an arbitrary number or only 2? If a limit, can we add CEL validation?

Do we expect the certificates to be identical except in algorithm? Or could they, for instance, be multiple distinct SANs

ramaraochavali: Only 2. I will add CEL validation.

> Do we expect the certificates to be identical except in algorithm? Or could they, for instance, be multiple distinct SANs

Yes. I was n't expecting/not thinking of use case for different SANs

howardjohn: Can we also make sure you set either credential_name or credential_names but not both  (and same for tls_certificate(s) I think?)

> Yes. I was n't expecting/not thinking of use case for different SANs

ramaraochavali: Added validations. PTAL. one thing I could not add is 
only one of cred*, (server_certificate, private_key),tls_certificates can be specified. Do you know if it is possible?

ramaraochavali:

networking/v1alpha3/gateway.proto
```diff
@@ -381,6 +381,11 @@ message Port {
 }
 
 message ServerTLSSettings {
+// +kubebuilder:validation:XValidation:message="credential_names cannot have more than two credentials",rule="default(self.credential_names, []).size() <= 2"
```
howardjohn, line 384:
```suggestion
// +kubebuilder:validation:XValidation:message="credential_names cannot have more than two credentials",rule="default(self.credentialName, []).size() <= 2"
```

The CEL works on the CRD naming, not the proto, which is camelCase. Might be worth adding that note into GUIDELINES.md.

We could add a test under tests/testdata/ to catch this

howardjohn: Also this would need to be over `message ServerTLSSettings` to work, not under it.

But even better, this could be only the specific field itself (ex over `tls_certificates`).

or even better, we do not need CEL for this; can use `kubebuilder:validation:items:MaxItems`. We should also probably put MinItems=1

howardjohn: So to clarify:

* The two oneof checks need to be on `ServerTLSSettings`, since it needs access to the fields. It needs to change to camel case
* The two checks of size should move to the field level and use MinItems/MaxItems

networking/v1alpha3/gateway.proto
```diff
@@ -381,6 +381,11 @@ message Port {
...
+// +kubebuilder:validation:XValidation:message="cannot have more than two tls certificates",rule="default(self.tls_certificates, []).size() <= 2"
+// +kubebuilder:validation:XValidation:message="only one of credential_names or tls_certificates can be set",rule="oneof(self.tls_certificates, self.credential_names)"
+// +kubebuilder:validation:XValidation:message="only one of credential_name or credential_names can be set",rule="oneof(self.credential_name, self.credential_names)"
```
howardjohn, line 387: Do we intentionally allow tls_certificate + credential_names?

keithmattix: My vote would be no

networking/v1alpha3/gateway.proto
```diff
@@ -471,9 +476,36 @@ message ServerTLSSettings {
...
+
+  // Only one of `server_certificate`, `private_key`, `ca_certificates` or `credential_name`
+  // or `credential_names` or `tls_certificates` should be specified.
+  // This is mainly used for specifying RSA and ECDSA certificates for the same server.
+  repeated TLSCertificate tls_certificates = 15;
+
   // A list of alternate names to verify the subject identity in the
   // certificate presented by the client.
   // Requires TLS mode to be set to `MUTUAL`.
+  // When multiple certificates are provided via `credential_names` or `tls_certificates`,
```
keithmattix, line 507: What's doing the validating? If the validation fails, what errors?

networking/v1alpha3/gateway.proto
ramaraochavali: > Do we intentionally allow tls_certificate + credential_names?

No. only one of them. That was the oneof I added. Any issue with that?

ramaraochavali: nvm. I did not add that combination. Fixed now

networking/v1alpha3/gateway.proto
ramaraochavali: This is the existing SAN validation and you get certificate verify failed message.

networking/v1alpha3/gateway.proto
```diff
@@ -380,6 +380,9 @@ message Port {
   uint32 target_port = 4 [deprecated=true];
 }
 
+// +kubebuilder:validation:XValidation:message="only one of credential_names or tls_certificates can be set",rule="oneof(self.tlsCertificates, self.credentialNames)"
+// +kubebuilder:validation:XValidation:message="only one of credential_name or credential_names can be set",rule="oneof(self.credentialName, self.credentialNames)"
+// +kubebuilder:validation:XValidation:message="only one of credential_name or tls_certificates can be set",rule="oneof(self.credentialNames, self.tlsCertificates)"
```
howardjohn, line 385:
```suggestion
// +kubebuilder:validation:XValidation:message="only one of credential_name or tls_certificates can be set",rule="oneof(self.credentialName, self.tlsCertificates)"
```

nit: the error message probably should also be camel case, though that matters less

howardjohn (commented):
LGTM just one minor comment

ramaraochavali:

howardjohn (approved):

howardjohn:
When you implement this can you make sure to either add support in gateway-api as well (they have the API but we just allow 1 entry) or open an issue to track the TODO?

ramaraochavali:
> When you implement this can you make sure to either add support in gateway-api as well (they have the API but we just allow 1 entry) or open an issue to track the TODO?

Will do

keithmattix (approved):

PR #3469, opened 2025-03-20
add annotation for statsCompression

zirain:
xref: https://github.com/istio/istio/issues/52082


howardjohn (approved):
LGTM. This is already in the code so this is just documenting existing state.

hanxiaop (approved):

PR #3474, opened 2025-03-26
support read header value from env variable

zirain:
this's a proposal for https://github.com/istio/istio/issues/53408

With this patch, users will be able mount a secret into pilot discovery pod as env variable, then pilot will read from it and send to all proxies.

### DIFF
--- a/mesh/v1alpha1/config.pb.go
+++ b/mesh/v1alpha1/config.pb.go
@@ -4126,8 +4126,13 @@ type MeshConfig_ExtensionProvider_HttpHeader struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
 	// REQUIRED. The HTTP header name.
 	Name string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
-	// REQUIRED. The HTTP header value.
-	Value         string `protobuf:"bytes,2,opt,name=value,proto3" json:"value,omitempty"`
+	// The HTTP header value.
+	//
+	// Types that are valid to be assigned to HeaderValue:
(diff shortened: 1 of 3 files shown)


zirain:
cc @joaopgrassi @keithmattix, looks like you're interesting in original issue.

howardjohn:
I see. This absolutely needs to specify it's the environment variable of the control plane, which is not intuitive. 

This is a possible security risk, as someone can now trigger the control plane to send out arbitrary env vars down to proxies, which may include confidential information. Given this is in mesh config, not proxy config, that is likely ok as they are the same trust domain. 

However we are still storing the credentials in envoy, accessible by config_dump,etc. We do similar for TLS, but that is all in the "secret" code paths which obfuscate the contents.  I worry this doesn't really meet the goal of the original issue, as we don't actually have end to end security of the header

zirain:
> However we are still storing the credentials in envoy, accessible by config_dump,etc. We do similar for TLS, but that is all in the "secret" code paths which obfuscate the contents. I worry this doesn't really meet the goal of the original issue, as we don't actually have end to end security of the header

you can avoid this happen, users who can access config_dump will be able to read nearly everything.

In my experience, most people don't have direct access to the product environment (or need permission).
It still makes sense to decouple the mesh configuration(with argo or else) from the creation of secret(manually by ops team).

howardjohn:
Yeah good call. Maybe we just specify in the comments that:
* Its istiod env var
* Its not end to end secure

howardjohn (approved):

keithmattix (approved):

PR #3480, opened 2025-04-07
Split root_ca in CSR API

howardjohn:
Opening this for discussion. 

Good background reading: https://github.com/istio/istio/issues/55793. Note we do NOT need this API to fix the bug. This PR is about making the API more clear and less error-prone.

This API still is not ideal. A better approach would be to have a separate stream of root CAs and perhaps some way to tell the client they should send a new CSR. This is a much bigger change than is present here, though.


keithmattix (commented):
/cc @jaellio to weigh in, but this seems reasonable to me

hzxuzhonghu (commented):
how can you make sure during upgrade, it is capatible for both client and server iwith different versions

howardjohn:
> how can you make sure during upgrade, it is capatible for both client and server iwith different versions

that's a good call, I only really consider older control plane than data plane, not newer control plane... but that's an important issue

security/v1alpha1/ca.proto
```diff
@@ -46,6 +46,15 @@ message IstioCertificateResponse {
   // PEM-encoded certificate chain.
   // The leaf cert is the first element, and the root cert is the last element.
   repeated string cert_chain = 1;
+  // Root certificates. This field is newer, and therefor is optional with the following semantics:
+  // * if root_cert is specified, cert_chain contains [leaf, intermediate1, intermediate2] (intermediates are optional).
+  //   root_cert contains [root1, root2].
+  // * if root_cert is not specified, cert_chain contains [leaf, intermediate1, intermediate2, root1+root2] concatenated into one entry.
+  // Note that the individual cert_chain is only signed by a single root. The roots provided here is the full bundle of trusted roots.
+  Roots root_cert = 2;
```
ilrudie, line 54: I wonder if having a more significant addition which introduces a new structured type containing a full replacement for cert_chanin would be better/cleaner than making a small change like this. Don't make them mutually exclusive and then we could have an interop mode where `cert_chain` and `name_tbd` are both populated with the ~same information.

dgn: +1 to that, let's introduce a real replacement struct, potentially with arrays for intermediates and roots and a separate leaf variable

keithmattix: I could get behind that

costinm: How about changing the name to "trust_bundle" (the name used in ClusterTrustBundle which serves the same purpose ) ? 

And not sure if repeated ca_cert is necessary (and not confusing) since usually roots are multiple PEMs concatenated - is the user suppose to split ? 

IMO it may be better to wait for ClusterTrustBundle to make its slow way - and if we really want to do it sooner just create a message with exactly the same field names and content with ClusterTrustBundle, so it's easier to move without getting confused by different naming.

But cleaning this up is a great idea.

PR #3494, opened 2025-04-19
Add note for reroute-virtual-interfaces usage with dind

sridhargaddam:


sridhargaddam:
Motivation behind the PR: https://istio.slack.com/archives/C37A4KAAD/p1745070788535839?thread_ts=1744990810.376149&cid=C37A4KAAD

annotation/annotations.yaml
```diff
@@ -578,6 +578,7 @@ annotations:
     featureStatus: Alpha
     description: |
       A comma separated list of virtual interfaces whose inbound traffic will be unconditionally treated as outbound. This allows workloads using virtualized networking (kubeVirt, VMs, docker-in-docker, etc) to function correctly with mesh traffic capture.
+      Note: In the case of docker-in-docker containers, the default Docker bridge name is not fixed. To have a predictable name, you can configure the Docker option `com.docker.network.bridge.name` with a fixed value and use that name in the annotation.
```
howardjohn, line 581:
```suggestion
      Note: When using docker-in-docker container, the default bridge interface name is typically `docker0`. However, custom networks (often used with docker compose) are assigned a randomized interface name. To have a predictable name, you can configure the Docker option `com.docker.network.bridge.name` with a fixed value and use that name in the annotation.
```

howardjohn (approved):
LGTM! thanks

annotation/annotations.yaml
diranged: I'd suggest putting in a full docker-compose example... eg:
```yaml
networks:
  custom_bridge:
    driver: bridge
    driver_opts:
      com.docker.network.bridge.name: "istio-docker0" <<  point out that this needs to be done in conjunction with the annotation
```

sridhargaddam: Thank you @howardjohn.

@diranged it looks a bit too verbose for a note. But if John/others, feel that including the full `networks` section is fine, I'm happy to update the PR.

linsun: I agree it is a bit verbose, could we host the special instruction as the example in the doc instead?

linsun:
lgtm - proposed change from John.

annotation/annotations.yaml
sridhargaddam: I tried searching the current docs but couldn't find an appropriate section where this note could be added.

costinm (approved):
Seems reasonable - since we mention already DIND.

PR #3495, opened 2025-04-23
hide ca cert from tls certificates

ramaraochavali:
Added for dual cert support but realized later that envoy supports single cert for validation. So decided to uses ca_cert in ServerTLSSettings. Hiding this for now

Ref
#3466  https://github.com/istio/istio/pull/55760


hzxuzhonghu (commented):
you should also add 


`$hide_from_docs` above the ca cert

hzxuzhonghu (approved):

costinm (approved):

PR #3498, opened 2025-04-28
Migrate storage version of networking APIs to v1

kfaseela:
Fixes https://github.com/istio/istio/issues/55302

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -2280,7 +2280,7 @@ spec:
             x-kubernetes-preserve-unknown-fields: true
         type: object
     served: true
-    storage: false
+    storage: true
     subresources:
       status: {}
   - additionalPrinterColumns:
@@ -6078,7 +6078,7 @@ spec:
             x-kubernetes-preserve-unknown-fields: true
         type: object
     served: true
-    storage: true
+    storage: false
     subresources:
       status: {}
 ---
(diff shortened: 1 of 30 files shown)


kfaseela:
@keithmattix : wondering why `make gen` did not set storageVersion: true for these CRDs in crd.gen.yaml

keithmattix:
Hmm looks like this doesn't work the way I remember it working. Will take a look

keithmattix:
Ah, I remember now @kfaseela. We need to manually edit the versions that we need to make the storage version. So e.g. nothing under v1alpha3 should have storage version; remove that and manually add the comment to the v1 copies

keithmattix (approved):
This looks right; let's try and get this merged soon so automator will pull it into istio/istio and we can confirm that the tests work

dhawton:
> @dhawton do we need to cherrypick the api so this is available for 1.26?

Yes. We're outside of the window for merge so it'd need TOC approval if it were to be brought into 1.26. So if you/TOC desires it, then it will need the label to be in there.

keithmattix:
Polling @istio/technical-oversight-committee

costinm (approved):

keithmattix (approved):

keithmattix:
We're really close to 1.26 release and I'd prefer not to throw a wrench in it. Let's target 1.27 with this

kfaseela:
1.27 is fine!

PR #3502, opened 2025-05-10
add session persistence support in destination rule

ramaraochavali:
API for https://github.com/istio/istio/issues/55839


networking/v1alpha3/destination_rule.proto
```diff
@@ -495,6 +495,35 @@ message LoadBalancerSettings {
     uint64 minimum_ring_size = 4 [deprecated=true];
   };
 
+  // Session persistence settings for the destination. Use this for hard session affinity.
```
howardjohn, line 498: we need very clear docs on the difference between this and consist hash

networking/v1alpha3/destination_rule.proto
```diff
@@ -495,6 +495,35 @@ message LoadBalancerSettings {
...
+  message SessionPersistence {
+    // Type defines the type of session persistence such as through
+    // the use of a header or cookie. Default is COOKIE.
+    enum Type {
+      // Use a cookie to store the session affinity information.
+      COOKIE  = 0;
+      // Use a header to store the session affinity information.
+      HEADER = 1;
```
howardjohn, line 506: how can this possibly be used? how would a client know the opaque value to put here?

networking/v1alpha3/destination_rule.proto
costinm: I think both 'hard affinity' and 'consistent hash' are well defined - but what we do ( 'by encoding the IP:port of the service in a cookie/header' ) needs to be clearly understood by users.

networking/v1alpha3/destination_rule.proto
costinm: It's not that opaque unfortunately - just an IP:port in a proto. I believe Envoy returns the header on the first request, and user is expected to set it in the next request.

Agree with your comment: it needs to be super clear in the comments, and also the case where we can't connect to the IP:port - don't remember if we return an error or just go to another endpoint and return the new updated header.

networking/v1alpha3/destination_rule.proto
```diff
@@ -495,6 +495,35 @@ message LoadBalancerSettings {
...
+
+    // The name of the header or cookie to use for session persistence.
+    string name = 2;
+
+    // The cookie settings for session persistence.
+    Cookie cookie = 3;
+
+    message Cookie {
+      // Lifetime of the cookie. If specified, a cookie with the TTL will be
+      // generated if the cookie is not present. If the TTL is present and zero,
```
costinm, line 520: I don't remember exactly - but there were some concerns with non-session cookies ( around privacy regulations - I think you are required to show the consent screen and all kind of other things ). 

I would very much prefer if we start without this setting and only with session cookies, it's a very complicated problem, best to avoid Istio starting to set persistent cookies and dealing with privacy.

costinm (commented):
Good in general - but needs more docs.

costinm:
Not blocking for this CL, but few concerns I have with using DR:

- normally DR can be used both by client (consumer) and producer. Envoy implements this in a strange way where the client can actually set cookies and implement persistence - because the client-sidecar can encode the endpoint IP, and server side can also do this by encoding their own IP, but AFAIK our implementation is client side. This may have changed - but needs to be clearly explained. 

- ambient with Waypoints may support this - but I think it should be configured with the Gateway API equivalent. Either way - we should document if it is supported/used/allowed in Waypoint.

- same for Gateways (of both kinds).

networking/v1alpha3/destination_rule.proto
ramaraochavali: Envoy returns the header with ip:port encoded and the client has to repeatedly send the header for every request.

The behaviour of Envoy when the ip:port is not found is dependent on the [strict](https://www.envoyproxy.io/docs/envoy/v1.34.1/api-v3/extensions/filters/http/stateful_session/v3/stateful_session.proto.html#extensions-filters-http-stateful-session-v3-statefulsession) flag.

ramaraochavali: I will update the docs to make all of this clear.

costinm:
I think Gateways and Waypoints should use the Gateway API - and the doc
needs to be clear on the behavior (the endpoint IP is
generated by the gateway/waypoint, behave like normal / standard persistent
sessions ).

For Sidecars - that's where things are strange, because the persistent
session is actually handled on the client side, by client
sidecar encoding the picked endpoint - server has nothing to do with it
(unless it's a waypoint or gateway - and then things
get very weird because there client will stick to the gateways, which is
usually not what user wants - and a second header or cookie
is needed for sticking with the final backend ). Multicluster or egress
gateways in the path have this problem.

My point really is that it's not obvious and the docs should be detailed -
and it is better to keep the 'normal' cases separate
and handled with Gateway API, and not attempt to pretend it's the same or
mix them too much.

networking/v1alpha3/destination_rule.proto
hzxuzhonghu: This is weak feature, mostly we istio should be transparent of applications

networking/v1alpha3/destination_rule.proto
```diff
@@ -495,6 +495,35 @@ message LoadBalancerSettings {
...
+    // The type of session persistence to use.
+    Type type = 1;
+
+    // The name of the header or cookie to use for session persistence.
+    string name = 2;
+
+    // The cookie settings for session persistence.
+    Cookie cookie = 3;
+
+    message Cookie {
```
hzxuzhonghu, line 518: no cookie name needed?

networking/v1alpha3/destination_rule.proto
```diff
@@ -495,6 +495,35 @@ message LoadBalancerSettings {
...
+      // Use a cookie to store the session affinity information.
+      COOKIE  = 0;
+      // Use a header to store the session affinity information.
+      HEADER = 1;
+    }
+
+    // The type of session persistence to use.
+    Type type = 1;
+
+    // The name of the header or cookie to use for session persistence.
```
hzxuzhonghu, line 512: Ah, see it. It is reusable field. Why donot we make header and cookie oneof

ramaraochavali:
rebased. working on it now

ramaraochavali:
@istio/technical-oversight-committee can you PTAL?  This is needed for gateway APIs also. If we do not have consensus on the API, the other option is to smuggle the gateway api properties as annotation on DR and generate the config - which is sub optimal IMO

keithmattix:
Can you add this to the WG meeting agenda or start a slack thread?

ramaraochavali:
sure

PR #3512, opened 2025-06-05
add RetryBudget in DestinationRule

zirain:
xref: https://github.com/istio/istio/issues/55838

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -302,7 +302,7 @@ message TrafficPolicy {
     //
     // Defaults to 3.
     //
-    uint32  min_retry_concurrency = 2;
+    uint32 min_retry_concurrency = 2;
   }
 
   // Specifies a limit on concurrent retries in relation to the number of active requests.


networking/v1alpha3/destination_rule.proto
```diff
@@ -285,6 +285,23 @@ message TrafficPolicy {
...
+    // the sum of active requests and active pending requests.
+    // Defaults to 20%.
+    //
+    // +kubebuilder:validation:Maximum=100
+    // +kubebuilder:validation:Minimum=0
+    google.protobuf.DoubleValue percent = 1;
+    // Specifies the minimum retry concurrency allowed for the retry budget.
+    // Defaults to 3.
+    //
+    uint32  min_retry_concurrency = 2;
```
ramaraochavali: So at any time, 3 concurrent retries are allowed even if retry budget is less than 3 based on request count? Is it used in low traffic scenarios?  Can you expand more on when would user configure this?

ramaraochavali (commented):
LGTM, one minor doc request

@istio/technical-oversight-committee Can you please take a look as well? Without this it is really hard to use retry budgets.

networking/v1alpha3/destination_rule.proto
zirain: updated

PR #3517, opened 2025-06-09
Add ambient mode caveats

Stevenjin8:
`Sidecar`, `ServiceEntry.exportTo` and `VirtualService.*.source_*` are not used at all. While ServiceEntry wildcard hostnames also aren't supported.

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -174,8 +174,6 @@ message DestinationRule {
 
   // One or more named sets that represent individual versions of a
   // service. Traffic policies can be overridden at subset level.
-  // Note that subsets are not support in ambient mode.
-  // Define multiple services instead.
   repeated Subset subsets = 3;
 
   // A list of namespaces to which this destination rule is exported.
@@ -319,8 +317,6 @@ message TrafficPolicy {
 //
 // **Note:** Policies specified for subsets will not take effect until
 // a route rule explicitly sends traffic to this subset.
-//
-// **Note:** Subsets are not supported in ambient mode. Define multiple services instead.
 //
 // One or more labels are typically required to identify the subset destination,
 // however, when the corresponding DestinationRule represents a host that
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -174,6 +174,8 @@ message DestinationRule {
 
   // One or more named sets that represent individual versions of a
   // service. Traffic policies can be overridden at subset level.
+  // Note that subsets are not support in ambient mode.
+  // Define multiple services instead.
   repeated Subset subsets = 3;
 
   // A list of namespaces to which this destination rule is exported.
@@ -317,6 +319,8 @@ message TrafficPolicy {
 //
 // **Note:** Policies specified for subsets will not take effect until
 // a route rule explicitly sends traffic to this subset.
+//
+// **Note:** Subsets are not supported in ambient mode. Define multiple services instead.
 //
 // One or more labels are typically required to identify the subset destination,
 // however, when the corresponding DestinationRule represents a host that
--- a/networking/v1alpha3/service_entry.proto
+++ b/networking/v1alpha3/service_entry.proto
@@ -464,6 +464,9 @@ message ServiceEntry {
   // 1. subjectAltNames: In addition to verifying the SANs of the
   //    service accounts associated with the pods of the service, the
   //    SANs specified here will also be verified.
+  //
+  // **NOTE 3:** Wildcard hostnames are not support in ambient mode.
+  //
   // +kubebuilder:validation:MinItems=1
   // +kubebuilder:validation:MaxItems=256
   // +protoc-gen-crd:list-value-validation:XValidation:message="hostname cannot be wildcard",rule="self != '*'"
--- a/networking/v1alpha3/service_entry.proto
+++ b/networking/v1alpha3/service_entry.proto
@@ -465,7 +465,7 @@ message ServiceEntry {
   //    service accounts associated with the pods of the service, the
   //    SANs specified here will also be verified.
   //
-  // **NOTE 3:** Ambient mode does not support wildcard hostsnames.
+  // **NOTE 3:** Ztunnel and Waypoint proxies does not support wildcard hostnames.
   //
   // +kubebuilder:validation:MinItems=1
   // +kubebuilder:validation:MaxItems=256
@@ -599,8 +599,7 @@ message ServiceEntry {
   // the annotation "networking.istio.io/exportTo" to a comma-separated list
   // of namespace names.
   //
-  // **Note:** Ambient mode does not support this field. Service entries will
-  // be exported to all namespaces.
+  // **Note:** Ztunnel and Waypoint proxies not support this field and will read it at "*".
   repeated string export_to = 7;
 
   // If specified, the proxy will verify that the server certificate's
--- a/networking/v1alpha3/service_entry.proto
+++ b/networking/v1alpha3/service_entry.proto
@@ -465,7 +465,7 @@ message ServiceEntry {
   //    service accounts associated with the pods of the service, the
   //    SANs specified here will also be verified.
   //
-  // **NOTE 3:** Ztunnel and Waypoint proxies does not support wildcard hostnames.
+  // **NOTE 3:** Ztunnel and Waypoint proxies do not support wildcard hosts.
   //
   // +kubebuilder:validation:MinItems=1
   // +kubebuilder:validation:MaxItems=256
--- a/networking/v1alpha3/service_entry.proto
+++ b/networking/v1alpha3/service_entry.proto
@@ -465,7 +465,7 @@ message ServiceEntry {
   //    service accounts associated with the pods of the service, the
   //    SANs specified here will also be verified.
   //
-  // **NOTE 3:** Wildcard hostnames are not support in ambient mode.
+  // **NOTE 3:** Ambient mode does not support wildcard hostsnames.
   //
   // +kubebuilder:validation:MinItems=1
   // +kubebuilder:validation:MaxItems=256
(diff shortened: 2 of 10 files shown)


networking/v1alpha3/destination_rule.proto
```diff
@@ -174,6 +174,8 @@ message DestinationRule {
 
   // One or more named sets that represent individual versions of a
   // service. Traffic policies can be overridden at subset level.
+  // Note that subsets are not support in ambient mode.
```
howardjohn: This is not really true. Its not supported in HTTPRoute. Its supported in waypoints though

howardjohn (changes requested):

networking/v1alpha3/destination_rule.proto
Stevenjin8: Ok I'm hearing contrasting things. It seems to me like most if not all of VirtualService works with waypoints. it's just alpha. So depending on who you ask, it might not be "supported." Where it gets tricky is when we start mixing Istio APIs and GWAPI. Like httproute and virtual service don't work together, and DR subsets aren't compatible with httproute.

Do most fields in VirtualService do what they say they do? Are there any fields that don't work?

howardjohn: Everything but sourceLabels and sourceVersion works.

Even ignoring VS, its not really that "subsets are not supported". At worst, there is no way to _reference_ a subset in ambient. Which is maybe practically the same thing it just seems odd since you can absolutely create subsets in ambient

Stevenjin8: @howardjohn are you saying that the following config wouldn't do anything?

```
kind: VirtualService
metadata:
  name: reviews
spec:
  hosts:
  - reviews
  http:
  - match:
    - headers:
        end-user:
          exact: jason
    route:
    - destination:
        host: reviews
        subset: v2
  - route:
    - destination:
        host: reviews
        subset: v3
```

Stevenjin8: also, do you mean sourceNamespace instead of sourceVersion?

howardjohn: sourceNamespace yeah sorry.

And that VS is 100% fine in ambient. What I meant was "At worst [if you exclude VirtualServices due to them being alpha], there is no way to reference a subset". if you include VS, subset is perfectly fine.

Stevenjin8:
no stale

networking/v1alpha3/service_entry.proto
```diff
@@ -595,6 +598,9 @@ message ServiceEntry {
   // For a Kubernetes Service, the equivalent effect can be achieved by setting
   // the annotation "networking.istio.io/exportTo" to a comma-separated list
   // of namespace names.
+  //
+  // **Note:** Ambient mode does not support this field. Service entries will
+  // be exported to all namespaces.
```
hzxuzhonghu: Is this the temporary limitation?

Stevenjin8: No, IIRC one of the big changes in ambient is that we want all workloads to have the same view of the mesh. So the same host corresponds to the same service/serviceentry regardless of the workload namespace

keithmattix (commented):
LGTM - these caveats are really important to call out. Thanks for doing this!

networking/v1alpha3/virtual_service.proto
```diff
@@ -57,6 +57,11 @@ syntax = "proto3";
 // The source of traffic can also be matched in a routing rule. This allows routing
 // to be customized for specific client contexts.
 //
+// *Note for Ambient Users*: Support for `VirtualService` in Ambient mode is alpha, and
+// there are no plans to increase support.
+// Use [Gateway API](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#mesh-traffic)
+// for a stable traffic management API.
+//
```
craigbox:
```suggestion
// *Note for ambient mode users*: Waypoints are configured using Gateway API. [`VirtualService` is not supported for configuring waypoint traffic routing](https://istio.io/latest/docs/ambient/usage/l7-features/#virtualservice).
// Any use of VirtualService with waypoints is considered Alpha, and may be subject to change in future releases.
// Istio’s maintainers do not intend to remove this support, but will not be progressing it to [any further feature phase](https://istio.io/latest/docs/releases/feature-stages/).
```

perhaps?

networking/v1alpha3/virtual_service.proto
```diff
@@ -818,6 +823,8 @@ message HTTPMatchRequest {
   //
   // **Note:** This is not a runtime match, but is a selector; it filters which workloads the
   // VirtualService applies to.
+  //
+  // **Note:** Ambient mode does not support this field.
```
craigbox: We say above that ambient mode doesn't support anything.

How about:

```suggestion
  // **Note:** This feature has no effect in ambient mode.
```

?

networking/v1alpha3/service_entry.proto
```diff
@@ -464,6 +464,9 @@ message ServiceEntry {
   // 1. subjectAltNames: In addition to verifying the SANs of the
   //    service accounts associated with the pods of the service, the
   //    SANs specified here will also be verified.
+  //
+  // **NOTE 3:** Ambient mode does not support wildcard hostsnames.
```
craigbox:
```suggestion
  // **NOTE 3:** Wildcard hostnames are not supported in ambient mode.
```

networking/v1alpha3/service_entry.proto
craigbox:
```suggestion
  // **Note:** In ambient mode, service entries are exported to all namespaces,
  // and this field has no effect.
```

networking/v1alpha3/sidecar.proto
```diff
@@ -21,8 +21,9 @@ syntax = "proto3";
 // $aliases: [/docs/reference/config/networking/v1alpha3/sidecar]
 
 // `Sidecar` describes the configuration of the sidecar proxy that mediates
```
craigbox:
```suggestion
// In Istio's sidecar mode, the `Sidecar` resource describes the configuration of the proxy that mediates
```

networking/v1alpha3/sidecar.proto
```diff
@@ -21,8 +21,9 @@ syntax = "proto3";
...
-// inbound and outbound communication to the workload instance it is attached to. By
-// default, Istio will program all sidecar proxies in the mesh with the
+// inbound and outbound communication to the workload instance it is attached to.
+// `Sidecar` only applies to Sidecar mode.
```
craigbox:
```suggestion
```

networking/v1alpha3/sidecar.proto
craigbox: and possibly a "Note 4" that says "Sidecar configuration is not required in ambient mode."?

networking/v1alpha3/virtual_service.proto
Stevenjin8: What about

```
// *Note for ambient mode users*: Waypoints should be configured Gateway API and
// any use of [VirtualService with waypoints is alpha](https://istio.io/latest/docs/ambient/usage/l7-features/#virtualservice).
// Istio’s maintainers do not intend to remove virtual service support, but will not be progressing it to [any further feature phase](https://istio.io/latest/docs/releases/feature-stages/).
```

I think "waypoints are configured using gateway api" is confusing since it suggests that they can **only** be configured using gwapi.

Similarly "Alpha, and may be subject to change in future releases" suggests that support could become beta, or even unsupport, however "but will not be progressing it to" suggests that it will be alpha forever (which is what we want to say)

networking/v1alpha3/service_entry.proto
Stevenjin8: @craigbox I imagine we get into "this works for ingress but not ztunnel/waypoint" too right?

networking/v1alpha3/service_entry.proto
Stevenjin8: @craigbox same question about ingress

networking/v1alpha3/sidecar.proto
Stevenjin8: Its required in sidecar mode either. Its more like, it has no effect in ambient mode

networking/v1alpha3/virtual_service.proto
Stevenjin8: Agreed. "support" is ambiguous

Stevenjin8: What about ` // **Note:** This field has no effect on Waypoint proxies`... mm still doesn't feel too great.

craigbox:
https://docs.google.com/document/d/1JjsPzOMJfu_evzgiRp-aJf1KDpa9ImyYrew__y5CTao/edit?tab=t.0#heading=h.8vlgwrna7yow for the broader terminology discussion

networking/v1alpha3/sidecar.proto
craigbox: I so wish that this object had been named "`SidecarScope`" (and the same for legacy `Gateway` as `GatewayConfig` or something too)

keithmattix (commented):
I think we still need this PR. Can we be more specific and say these fields aren't respected by the ambient data plane (e.g. ztunnel and waypoints)?

networking/v1alpha3/service_entry.proto
```diff
@@ -464,6 +464,9 @@ message ServiceEntry {
   // 1. subjectAltNames: In addition to verifying the SANs of the
   //    service accounts associated with the pods of the service, the
   //    SANs specified here will also be verified.
+  //
+  // **NOTE 3:** Ztunnel and Waypoint proxies does not support wildcard hostnames.
```
keithmattix: What does "NOTE 3" refer to?

Stevenjin8: @keithmattix I'm not sure what the question is. There are two other notes above (hence the 3). It refers to the `hosts` field.git status

keithmattix: Oh sorry didn't scroll up enough

hzxuzhonghu (approved):
Ambient does not support exportTo, it seems like a security risk, ztunnel/waypoint can make acccess escalation

Stevenjin8:
> Ambient does not support exportTo, it seems like a security risk, ztunnek/waypoint can make acccess escalation

@hzxuzhonghu could you speak a bit more? To me exportto is about discoverability, not security, so I'm having trouble seeing the connection

howardjohn (approved):

PR #3522, opened 2025-06-16
format protobuf files

zirain:


zirain:
> If we are going to format (which seems good to me) we need automation IMO. What did you use to format? I think `buf`, which we use, can do formatting so we would just need to make sure its automated

it's part of `gen.sh`, do you want a seperated make target?

howardjohn (approved):
My bad, I just missed the gen.sh! Thanks

PR #3523, opened 2025-06-17
add support for additional cookie attributes

ramaraochavali:
For https://github.com/istio/istio/issues/56468 and https://github.com/istio/istio/issues/49870


ramaraochavali:
@istio/technical-oversight-committee PTAL

ramaraochavali:
@keithmattix @howardjohn can you PTAL when you get chance?

networking/v1alpha3/destination_rule.proto
```diff
@@ -482,6 +491,8 @@ message LoadBalancerSettings {
       // the generated cookie will be a session cookie.
       // +protoc-gen-crd:duration-validation:none
       google.protobuf.Duration ttl = 3;
+      // Additional attributes for the cookie. They will be used when generating a new cookie.
+      repeated Attribute attributes = 4;
```
zirain, line 495: what will happen if we have multiple attributes with same name/key?

ramaraochavali: In Envoy, they will get just appended. But nothing should break. We can add additional validations in Istio and reject the configuration if that is a concern

zirain (approved):

PR #3529, opened 2025-06-22
add gateway sds extension provider

ramaraochavali:
API for https://github.com/istio/istio/issues/56522

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -604,9 +604,14 @@ message MeshConfig {
       EnvoyTcpGrpcV3LogProvider envoy_tcp_als = 13;
       // Configures an Envoy Open Telemetry Access Logging Service provider.
       EnvoyOpenTelemetryLogProvider envoy_otel_als = 14;
+      // Configures an Extension Provider for Gateway SDS. This can be used to
+      // configure an external SDS service to supply secrets for certain Gateways.
+      // This is useful for scenarios where the secrets are stored in an external secret store like Vault.
+      // The secret should be configured with sds://provider-name format.
+      GatewaySDSProvider gateway_sds = 16;
 
       // $hide_from_docs
-      // next id: 16
+      // next id: 17
     }
 
     message EnvoyExternalAuthorizationRequestBody {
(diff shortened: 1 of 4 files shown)


ramaraochavali:
@howardjohn PTAL

mesh/v1alpha1/config.proto
```diff
@@ -604,9 +604,14 @@ message MeshConfig {
       EnvoyTcpGrpcV3LogProvider envoy_tcp_als = 13;
       // Configures an Envoy Open Telemetry Access Logging Service provider.
       EnvoyOpenTelemetryLogProvider envoy_otel_als = 14;
+      // Configures an Extension Provider for Gateway SDS. This can be used to
+      // configure an external SDS service to supply secrets for certain Gateways.
+      // This is useful for scenarios where the secrets are stored in an external secret store like Vault.
+      // The secret should be configured with sds://provider-name format.
+      GatewaySDSProvider gateway_sds = 16;
```
howardjohn: I am wondering if it should just be `SDSProvider`? Since it could be used for non-gateway (DestinationRule)? Or at least could in the future if its not today

mesh/v1alpha1/config.proto
```diff
@@ -1301,6 +1306,23 @@ message MeshConfig {
...
+      // REQUIRED. Specifies the service that implements the  SDS service.
+      // The format is `[<Namespace>/]<Hostname>`. The specification of `<Namespace>` is required only when it is insufficient
+      // to unambiguously resolve a service in the service registry. The `<Hostname>` is a fully qualified host name of a
+      // service defined by the Kubernetes service or ServiceEntry.
+      //
+      // Example: "gateway-sds.foo.svc.cluster.local" or "bar/gateway-sds.example.com".
+      string service = 2;
+
+      // REQUIRED. Specifies the port of the service.
+      uint32 port = 3;
```
howardjohn, line 1323: I think we have this same REQUIRED for every other provider but curious if this works with a Service with a UDS endpoint? Which _may_ be common with SDS

mesh/v1alpha1/config.proto
ramaraochavali: Good point. Will change

mesh/v1alpha1/config.proto
ramaraochavali: we can add uds_path and support both. Shall I change it to oneof and expose both?

howardjohn: Service is fine for UDS -- since it can be a ServiceEntry with UDS endpoint. I don't think we want to directly expose UDS.

I am just not sure what happens with 'port'. It looks like
```
kind: ServiceEntry
metadata:
  name: unix-domain-socket-example
spec:
  hosts:
  - "example.unix.local"
  location: MESH_EXTERNAL
  ports:
  - number: 80
    name: http
    protocol: HTTP
  resolution: STATIC
  endpoints:
  - address: unix:///var/run/example/socket
```
is valid so maybe the port is just essentially a dummy port but still works. So I think no change needed really

ramaraochavali:
@howardjohn  is OK with this change. @istio/technical-oversight-committee can you please review and approve this?

howardjohn (approved):

mesh/v1alpha1/config.proto
keithmattix: Should the docs specify that it's a dummy port?

keithmattix (approved):
LGTM

mesh/v1alpha1/config.proto
ramaraochavali: For Services, it is valid port - only for UDS it is dummy

PR #3531, opened 2025-06-26
[fixes #56772] Fix WasmPlugin .spec.url validation

DerekTBrown:
## What does this PR do?
- Per the Istio documentation, if the `oci://` prefix is missing in `.spec.url`, it is assumed that the URL is a `oci://` image reference[^1].
- Things like `foo.bar:5000/baz:latest` are valid image references, but do not pass validation, because this treats `foo.bar` as the schema.
- I added an extensive set of test cases, then modified the validation logic to support each.

[^1]: https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/#WasmPlugin-url.

### DIFF
--- a/extensions/v1alpha1/wasm.proto
+++ b/extensions/v1alpha1/wasm.proto
@@ -279,7 +279,7 @@ message WasmPlugin {
   // within the proxy container, and `http[s]://` for `.wasm` module files
   // hosted remotely.
   // +kubebuilder:validation:MinLength=1
-  // +kubebuilder:validation:XValidation:message="url must have schema one of [http, https, file, oci]",rule="isURL(self) ? (url(self).getScheme() in ['', 'http', 'https', 'oci', 'file']) : (isURL('http://' + self) && url('http://' +self).getScheme() in ['', 'http', 'https', 'oci', 'file'])"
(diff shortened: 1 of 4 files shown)


extensions/v1alpha1/wasm.proto
```diff
@@ -279,7 +279,7 @@ message WasmPlugin {
   // within the proxy container, and `http[s]://` for `.wasm` module files
   // hosted remotely.
   // +kubebuilder:validation:MinLength=1
-  // +kubebuilder:validation:XValidation:message="url must have schema one of [http, https, file, oci]",rule="isURL(self) ? (url(self).getScheme() in ['', 'http', 'https', 'oci', 'file']) : (isURL('http://' + self) && url('http://' +self).getScheme() in ['', 'http', 'https', 'oci', 'file'])"
```
howardjohn, line 282: what about:

```
self.contains("://") ? (url(self).getScheme() in ['', 'http', 'https', 'oci', 'file']) : (isURL('http://' + self) && url('http://' +self).getScheme() in ['', 'http', 'https', 'oci', 'file'])
```

replacing the first isURL() with `self.contains("://")`

DerekTBrown: Makes sense to me (to make the logic simpler / easier to follow).

I am going to implement this as  `self.contains("://") && isURL(self) && url(self).getScheme() in ... `, just because `isURL` guards `url(self).getScheme()` in my head.

PR #3532, opened 2025-06-27
add x-forwarded-host and x-forwarded-port in proxy headers

ramaraochavali:


mesh/v1alpha1/proxy.proto
```diff
@@ -740,6 +746,13 @@ message ProxyConfig {
     // requests and automatically normalize headers to lowercase, ensuring compliance with HTTP/2
     // standards.
     google.protobuf.BoolValue preserve_http1_header_case = 40;
+    // Controls the `X-Forwarded-Host` header. If enabled, the `X-Forwarded-Host` header is appended
```
howardjohn, line 749: This defaults to OFF, right?

If so, we should probably mention that (since most of these are default ON), and _maybe_ flip `disabled` to `enabled` so you don't need to do `disabled: false`?

ramaraochavali: Yes. Changed

ramaraochavali:

howardjohn (approved):

ramaraochavali:

hzxuzhonghu (approved):

PR #3543, opened 2025-07-22
unhide ambient.istio.io/bypass-inbound-capture annotation

mjnagel:
Per discussion in Slack: https://istio.slack.com/archives/C50V5EATT/p1753200025888989

Unhiding `ambient.istio.io/bypass-inbound-capture` to make it more visible.


howardjohn (approved):
thanks!

mjnagel:
I don't think I have permissions to add a label here, but this shouldn't require a release note 😃

PR #3547, opened 2025-08-07
Add spaceDelimitedClaims field to RequestAuthentication API

fjglira:
Related Issues
#[56873](https://github.com/istio/istio/issues/56873)

Following a change in Istio 1.21, JWT claims other than `scope` and `permissions` are treated as exact string values. This created an issue for users with custom, space-delimited claims (for example: a roles claim with the value "editor admin"), as they could no longer match individual values like editor or admin in their Authorization Policies.

This PR introduces a new field, `spaceDelimitedClaims`, to the `RequestAuthentication` API. This field allows users to explicitly specify a list of custom claims that Istio should parse as space-delimited strings. This restores the previous, more flexible behavior for users who depend on it, without changing the new default for other claims.

  ### API Changes

  ```yaml
  kind: RequestAuthentication
  spec:
    jwtRules:
    - issuer: "https://example.com"
      jwksUri: "https://example.com/.well-known/jwks.json"
      spaceDelimitedClaims:
      - "custom_scope"
      - "provider.login.scope"
      - "roles"
```

This change is fully backward compatible. The default behavior for the standard scope and permissions claims remains unchanged; they will always be treated as space-delimited lists, regardless of whether they are included in the new field.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -16174,6 +16174,14 @@ spec:
                       description: This field specifies the header name to output
                         a successfully verified JWT payload to the backend.
                       type: string
+                    spaceDelimitedClaims:
+                      description: List of JWT claim names that should be treated
+                        as space-delimited strings.
+                      items:
+                        minLength: 1
+                        type: string
+                      maxItems: 64
+                      type: array
                     timeout:
                       description: The maximum amount of time that the resolver, determined
                         by the PILOT_JWT_ENABLE_REMOTE_JWKS environment variable,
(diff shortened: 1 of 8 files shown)


fjglira:
> Do we know if space is the only delimiter we care about ? Would we be better off with something a little more long-term flexible
> 
> ```yaml
> apiVersion: security.istio.io/v1beta1
> kind: RequestAuthentication
> spec:
>   jwtRules:
>   - issuer: "https://example.com"
>     jwksUri: "https://example.com/.well-known/jwks.json"
>     formattedClaims:
>        - format: SPACE_DELIMITED
>          claims:
>             - "custom_scope"
>             - "provider.login.scope"
>             - "roles"
> ```
> 
> where 'format' likely starts with just SPACE_DELIMITED and CSV as the available options but can evolve without requiring schema change over time to handle more cases ?

Hi @louiscryan, that's a great point about ensuring long-term flexibility.

The primary motivation for this feature is to handle custom claims that adopt the same format as the standard scope claim. For instance, the original issue highlights the need to parse a space-delimited string from a custom claim like provider.oauth.scope used by Oracle OAuth.

As defined in [RFC 6749, Section 3.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3), the scope value is explicitly a space-delimited list of strings. Given that this is the standardized and most prevalent format for this kind of value, I opted for a solution that applies this same parsing logic to other claims that follow this convention.

I agree that your proposed `formattedClaims` structure is more extensible, and we can certainly adopt it if we find concrete use cases for other delimiters (like CSV) in the future. For now, would you be comfortable merging this targeted solution to solve the immediate issue, and we can iterate with a more generic approach later if we need to?

PR #3549, opened 2025-08-13
Clarify Service Entry addresses behavior limitations

kfaseela:
Ref: https://github.com/istio/istio/issues/34577

### DIFF
--- a/networking/v1alpha3/service_entry.proto
+++ b/networking/v1alpha3/service_entry.proto
@@ -482,7 +482,9 @@ message ServiceEntry {
   // service in the mesh. In other words, the sidecar will behave as a
   // simple TCP proxy, forwarding incoming traffic on a specified port to
   // the specified destination endpoint IP/host. Unix domain socket
-  // addresses are not supported in this field.
+  // addresses are not supported in this field. Although the API allows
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/service_entry.proto
```diff
@@ -482,7 +482,9 @@ message ServiceEntry {
   // service in the mesh. In other words, the sidecar will behave as a
   // simple TCP proxy, forwarding incoming traffic on a specified port to
   // the specified destination endpoint IP/host. Unix domain socket
-  // addresses are not supported in this field.
+  // addresses are not supported in this field. Note that the `addresses`
+  // API implies you can provide multiple addresses but there are limitations
+  // on it (e.g.: you cannot set VIP addresses regardless of the combination of ip families).
```
ilrudie: I have a hard time understanding what this means. I think the clarification we wish to convey is that multiple VIPs will not be assigned to a single host, even though having a list suggests that might occur. As a result of this, you are not able to specify dual-stack VIPs using this field.

ilrudie (approved):
This wording is clearer to my mind. Thanks!

howardjohn (approved):

PR #3550, opened 2025-08-14
Add new EnvoyFilter type UPSTREAM_HTTP_FILTER

wtzhang23:
Issue: https://github.com/istio/istio/issues/57342

This API change allows patching the array of http filters in clusters similar to how its downstream variant HTTP_FILTER patches the array of filters in the HCM.

### DIFF
--- a/networking/v1alpha3/envoy_filter.proto
+++ b/networking/v1alpha3/envoy_filter.proto
@@ -420,6 +420,47 @@ syntax = "proto3";
 //             - request_headers:
 //                 header_name: ":path"
 //                 descriptor_key: "path"
+// ```
+//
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/envoy_filter.proto
```diff
@@ -538,6 +541,14 @@ message EnvoyFilter {
   // Conditions specified in `ClusterMatch` must be met for the patch
   // to be applied to a cluster.
   message ClusterMatch {
```
ramaraochavali, line 584: Can you please provide an example usage in the docs?

wtzhang23: Added example to the top-level package comments

PR #3562, opened 2025-09-04
stats: add bootstrap annotations for stats flush and eviction

kyessenov:
Change-Id: I6c64ddd71aee82b570c7df25f13e2d710a046558

### DIFF
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -210,6 +210,16 @@ annotations:
     resources:
       - Pod
 
+  - name: sidecar.istio.io/statsHistogramBins
+    featureStatus: Alpha
+    description: Specifies the bin size per time series for the Istio standard
+      metrics histograms. Reducing this value from the default `100` decreases
+      overall memory usage for sparse and/or high cardinality histograms.
+    deprecated: false
+    hidden: false
+    resources:
+      - Pod
+
   - name: sidecar.istio.io/userVolume
     featureStatus: Alpha
     description: Specifies one or more user volumes (as a JSON array) to be added to
--- a/annotation/annotations.yaml
+++ b/annotation/annotations.yaml
@@ -191,6 +191,25 @@ annotations:
     resources:
       - Pod
 
+  - name: sidecar.istio.io/statsFlushInterval
+    featureStatus: Alpha
+    description: Specifies the flush interval for push-based stat sinks, e.g. OTLP. Default interval is `5s`.
+    deprecated: false
+    hidden: false
+    resources:
+      - Pod
+
+  - name: sidecar.istio.io/statsEvictionInterval
+    featureStatus: Alpha
+    description: Specifies the expiration interval for the Istio standard
+      metrics. This gets rounded to a multiple of the flush interval. A time
+      series is expected to be evicted after 2 iterations of this interval from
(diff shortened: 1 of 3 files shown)


annotation/annotations.yaml
```diff
@@ -191,6 +191,25 @@ annotations:
     resources:
       - Pod
 
+  - name: sidecar.istio.io/statsFlushInterval
+    featureStatus: Alpha
+    description: Specifies the flush interval for push-based stat sinks, e.g. OTLP. Default interval is `5s`.
```
keithmattix, line 196: qq: if this is only for push-based stats sinks, what about prometheus (pull based)?

kyessenov: Flush interval doesn't matter for prometheus. However, eviction piggy backs on flush timer, and eviction works for prometheus.

howardjohn: I don't get it. if it doesn't matter, but we evict when we flush, then it does matter...? And if the default is 5s, isn't that going to break prometheus?

kyessenov: Envoy always runs a stats flush timer, regardless of prometheus. During that timer, server stats are published (e.g. memory allocations, etc), and some other work is done, e.g. eviction.

Prometheus endpoint can snip stats are any point through, in between flushes.

So flush impacts prometheus only indirectly, e.g. by updating/removing some metrics.

howardjohn: Maybe more concretely... I send 1 request to `foo.com`. I get an istio_request_total metric for it.

When, with prometheus, will this metric stop showing up in /stats/prometheus with/without this annotation set?

kyessenov: For flush annotation: it will show up immediately and stay forever.
For eviction annotation: same, but it will disappear after 2*eviction interval unless the metric gets another hit.

howardjohn: Thanks! Sorry I was being a bit dense, github UI was hiding part of the PR

zirain (approved):
LGTM, Istio don't have first class API for bootstrap, this's the best option for now.

annotation/annotations.yaml
zirain: @howardjohn @keithmattix can we move this forward?

annotation/annotations.yaml
```diff
@@ -191,6 +191,35 @@ annotations:
     resources:
       - Pod
 
+  - name: sidecar.istio.io/statsFlushInterval
```
keithmattix, line 194: Does the sidecar prefix mean this doesn't work for gateways/waypoints? I would guess no (that this is just convention) but wanted to confirm

kyessenov: No, it's interpreted by pilot-agent which is the same for gateways and sidecars.

ringerc:
Is there any way to control defaults for `sidecar.istio.io/statsEvictionInterval` settings via the Istio `Telemetry` resource, rather than having to annotate every single workload?

zirain:
> Is there any way to control defaults for `sidecar.istio.io/statsEvictionInterval` settings via the Istio `Telemetry` resource, rather than having to annotate every single workload?

this's part of bootstrap, which is not dynamiced.

PR #3565, opened 2025-09-15
Add DYNAMIC_DNS resolution type for wildcard hosts

jaellio:
Part of https://github.com/istio/istio/issues/54540

### DIFF
--- a/networking/v1/service_entry_alias.gen.go
+++ b/networking/v1/service_entry_alias.gen.go
@@ -102,16 +102,16 @@ const ServiceEntry_DNS_ROUND_ROBIN ServiceEntry_Resolution = v1alpha3.ServiceEnt
 
 // DYNAMIC_DNS will attempt to resolve the host name specified in
 // the Host header or SNI to an IP address when handling traffic. This
-// is particularly useful when multiple dns addresses can be represented
-// by a single wildcard `host` entry without having to explicitly
-// enumerate all possible endpoints. During DNS proxying, ztunnel will
-// resolve all subdomains matching the wildcard host name to a VIP which
-// isn't used for routing outside the mesh. `DYNAMIC_DNS` will provide
-// configuration to a waypoint proxy to recover the original host name
-// using information from SNI or a Host header in an HTTP Request. This
-// original host name will then be resolved so that traffic can be routed
-// to the intended IP address. This method of handling wildcard traffic
-// is not compatible with raw TCP traffic where the original host cannot
+// allows multiple DNS addresses to be represented by a single wildcard
+// `host` entry without having to explicitly enumerate all possible
+// endpoints. During DNS proxying, ztunnel will resolve all subdomains
+// matching the wildcard host name to a VIP which isn't used for routing
+// outside the mesh. `DYNAMIC_DNS` will provide configuration to a
+// waypoint proxy to recover the original host name using information
+// from SNI or a Host header in an HTTP Request. This original host name
+// will then be resolved so that traffic can be routed to the intended
+// IP address. This method of handling wildcard traffic is not
+// compatible with raw TCP traffic where the original host cannot
 // be recovered. `DYNAMIC_DNS` is only supported for wildcard hosts,
 // `MESH_EXTERNAL` location and in ambient mode. The ServiceEntry must
 // be bound to a waypoint. Specified endpoints will be ignored.
--- a/networking/v1alpha3/service_entry.proto
+++ b/networking/v1alpha3/service_entry.proto
@@ -563,16 +563,16 @@ message ServiceEntry {
 
     // DYNAMIC_DNS will attempt to resolve the host name specified in
     // the Host header or SNI to an IP address when handling traffic. This
-    // is particularly useful when multiple dns addresses can be represented
-    // by a single wildcard `host` entry without having to explicitly
-    // enumerate all possible endpoints. During DNS proxying, ztunnel will
-    // resolve all subdomains matching the wildcard host name to a VIP which
-    // isn't used for routing outside the mesh. `DYNAMIC_DNS` will provide
-    // configuration to a waypoint proxy to recover the original host name
-    // using information from SNI or a Host header in an HTTP Request. This
-    // original host name will then be resolved so that traffic can be routed
-    // to the intended IP address. This method of handling wildcard traffic
-    // is not compatible with raw TCP traffic where the original host cannot
+    // allows multiple DNS addresses to be represented by a single wildcard
+    // `host` entry without having to explicitly enumerate all possible
+    // endpoints. During DNS proxying, ztunnel will resolve all subdomains
+    // matching the wildcard host name to a VIP which isn't used for routing
+    // outside the mesh. `DYNAMIC_DNS` will provide configuration to a
+    // waypoint proxy to recover the original host name using information
+    // from SNI or a Host header in an HTTP Request. This original host name
+    // will then be resolved so that traffic can be routed to the intended
+    // IP address. This method of handling wildcard traffic is not
+    // compatible with raw TCP traffic where the original host cannot
     // be recovered. `DYNAMIC_DNS` is only supported for wildcard hosts,
     // `MESH_EXTERNAL` location and in ambient mode. The ServiceEntry must
     // be bound to a waypoint. Specified endpoints will be ignored.
--- a/networking/v1alpha3/service_entry.pb.go
+++ b/networking/v1alpha3/service_entry.pb.go
@@ -519,16 +519,16 @@ const (
 	ServiceEntry_DNS_ROUND_ROBIN ServiceEntry_Resolution = 3
 	// DYNAMIC_DNS will attempt to resolve the host name specified in
 	// the Host header or SNI to an IP address when handling traffic. This
-	// is particularly useful when multiple dns addresses can be represented
-	// by a single wildcard `host` entry without having to explicitly
-	// enumerate all possible endpoints. During DNS proxying, ztunnel will
-	// resolve all subdomains matching the wildcard host name to a VIP which
-	// isn't used for routing outside the mesh. `DYNAMIC_DNS` will provide
-	// configuration to a waypoint proxy to recover the original host name
-	// using information from SNI or a Host header in an HTTP Request. This
-	// original host name will then be resolved so that traffic can be routed
-	// to the intended IP address. This method of handling wildcard traffic
-	// is not compatible with raw TCP traffic where the original host cannot
+	// allows multiple DNS addresses to be represented by a single wildcard
+	// `host` entry without having to explicitly enumerate all possible
+	// endpoints. During DNS proxying, ztunnel will resolve all subdomains
+	// matching the wildcard host name to a VIP which isn't used for routing
+	// outside the mesh. `DYNAMIC_DNS` will provide configuration to a
+	// waypoint proxy to recover the original host name using information
+	// from SNI or a Host header in an HTTP Request. This original host name
+	// will then be resolved so that traffic can be routed to the intended
+	// IP address. This method of handling wildcard traffic is not
+	// compatible with raw TCP traffic where the original host cannot
 	// be recovered. `DYNAMIC_DNS` is only supported for wildcard hosts,
 	// `MESH_EXTERNAL` location and in ambient mode. The ServiceEntry must
 	// be bound to a waypoint. Specified endpoints will be ignored.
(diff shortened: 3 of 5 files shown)


networking/v1alpha3/service_entry.proto
```diff
@@ -560,6 +560,16 @@ message ServiceEntry {
...
+
+    // Similar to DNS, by obtains the FQDN from the Host header or SNI to
+    // perform DNS resolution when the matching Host is a wildcard. Initial
+    // DNS resolution returns the allocated VIP for the matching wildcard
+    // hosts specified in the ServiceEntry. `DYNAMIC_DNS` must be used with
+    // wildcard hosts. Depending on the traffic type, the proxy will resolve
+    // the DNS address specified in the Hosts header or SNI of the proxied
+    // request. Specified endpoints will be ignored. Only supported for
+    // `MESH_EXTERNAL` ServiceEntries.
+    DYNAMIC_DNS = 4;
```
keithmattix, line 579: Should this only be implemented for ambient waypoints? I don't know that we want/need this for sidecar

jaellio: For now, I think it should be ambient only but we could expand it to sidecar. This would allow users to no longer have to configure dfp manually with envoy filters. I'll clarify this in the comment

jaellio:
Our current kubebuilder directive for wildcard hosts disallows setting any host to `*`. It does not check if a defined host contains a `*`. Is the intended behavior? Do we want to be more restrictive here? 

https://github.com/istio/api/pull/3565/files#diff-9220e0fa673c4bf811c9e82e0095c18e1306374c66a2a39e51fabd456034c217R469

networking/v1alpha3/service_entry.proto
```diff
@@ -560,6 +560,17 @@ message ServiceEntry {
...
+    // Similar to DNS, by obtains the FQDN from the Host header or SNI to
+    // perform DNS resolution when the matching Host is a wildcard. Initial
+    // DNS resolution returns the allocated VIP for the matching wildcard
+    // hosts specified in the ServiceEntry. `DYNAMIC_DNS` must be used with
+    // wildcard hosts. Depending on the traffic type, the proxy will resolve
+    // the DNS address specified in the Hosts header or SNI of the proxied
+    // request. Specified endpoints will be ignored. Only supported for
+    // `MESH_EXTERNAL` ServiceEntries and in ambient mode. The ServiceEntry
+    // must be bound to a waypoint.
+    DYNAMIC_DNS = 4;
```
jaellio, line 579: We could add kubebuilder validation to verify a wildcard hostname is set if `DYNAMIC_DNS` is set

keithmattix: Yeah either kubebuilder/OpenAPI or istiod validation

networking/v1/service_entry_alias.gen.go
```diff
@@ -100,6 +100,17 @@ const ServiceEntry_DNS ServiceEntry_Resolution = v1alpha3.ServiceEntry_DNS
 // cannot be used with Unix domain socket endpoints.
 const ServiceEntry_DNS_ROUND_ROBIN ServiceEntry_Resolution = v1alpha3.ServiceEntry_DNS_ROUND_ROBIN
 
+// Similar to DNS, by obtains the FQDN from the Host header or SNI to
```
grnmeira:
```suggestion
// Similar to DNS, but obtains the FQDN from the Host header or SNI to
```

networking/v1/service_entry_alias.gen.go
```diff
@@ -100,6 +100,17 @@ const ServiceEntry_DNS ServiceEntry_Resolution = v1alpha3.ServiceEntry_DNS
...
+// perform DNS resolution when the matching Host is a wildcard. Initial
+// DNS resolution returns the allocated VIP for the matching wildcard
```
grnmeira: Instead of "initial DNS resolution", can we be more specific using ztunnel? Unless "initial DNS resolution" is a common jargon in the API meaning the ambient per-node proxy.

networking/v1/service_entry_alias.gen.go
```diff
@@ -100,6 +100,17 @@ const ServiceEntry_DNS ServiceEntry_Resolution = v1alpha3.ServiceEntry_DNS
...
+// hosts specified in the ServiceEntry. `DYNAMIC_DNS` must be used with
+// wildcard hosts. Depending on the traffic type, the proxy will resolve
```
grnmeira: We can also use it with FQDNs, but I'm not sure if that's really useful.

keithmattix: I would advocate for using dynamic DNS ONLY for wildcards at this point. There are alternatives for FQDNs that are more in keeping with existing istio behavior

networking/v1/service_entry_alias.gen.go
jaellio: I think "client-initiated DNS" is better. From the client's perspective there is only one dns resolution.

grnmeira: Yeah, actually thinking from the client's perspective both work fine. I noticed though that you mention `Host` twice. First time it's the Host header, and then it's the ServiceEntry's hostname field. Maybe that could be ambiguous in a sense that someone might expect wildcards in the Host header?

ilrudie: Maybe it would make sense to describe it as "DNS resolution when egressing the mesh" or something similar. Would this type of concept help distinguish what envoy DFP is doing vs what client's using DNS capture are seeing?

networking/v1alpha3/service_entry.proto
```diff
@@ -560,6 +560,17 @@ message ServiceEntry {
     // specified in the hosts field, if wildcards are not used. DNS resolution
     // cannot be used with Unix domain socket endpoints.
     DNS_ROUND_ROBIN = 3;
+
+    // Similar to DNS, but obtains the FQDN from the Host header or SNI to
+    // perform DNS resolution when the matching Host is a wildcard. Client
```
ilrudie: I sort of dislike defining one option in terms of another. Should we adjust it so it stands on its own?

```suggestion
    // DYNAMIC_DNS will attempt to resolve the host name specified in the Host header or SNI to an IP
    // address when handling traffic. This is particularly useful when the front-end host name contains a
    // wildcard which prevents our DNS and DNS_ROUND_ROBIN types from being able to resolve the
    // host name specified in the `hosts` field to a concrete back end.
```

networking/v1alpha3/service_entry.proto
```diff
@@ -560,6 +560,17 @@ message ServiceEntry {
...
+    // initiated DNS resolution returns the allocated VIP for the matching
```
ilrudie: Talking here about client-initiated seems a bit strange to me. Maybe we should just clarify that these setting concern how hosts are resolve to concrete back-end IPs and unrelated to how mesh clients resolve addresses.

keithmattix (commented):
LGTM; will approve after:
1. Ian's comments/suggestions are addressed
2. We have clarity on if we're using webhook validation or kubebuilder validation how this field is going to be used (I recommend webhook validation if we want to support this mode in sidecars later)

networking/v1alpha3/service_entry.proto
jaellio: I agree with using webhook validation instead of kubebuilder validation for the reasonings you noted below - allows more flexibility to extend capabilities and support in the future.

networking/v1alpha3/service_entry.proto
keithmattix: +1

networking/v1alpha3/service_entry.proto
jaellio: Could you clarify what you mean by "front-end host name"?

jaellio: @keithmattix and @ilrudie took a shot at clarifying the description

ilrudie: What I was trying to convey, and maybe it's just confusing, is there is the wildcard that we're matching on which is using some sort of Auto VIP most likely. This is sort of the "front" of the egress gateway. There is also the back-end host which in this case we derive from SNI or a header and actually resolve to a real address.

networking/v1alpha3/service_entry.proto
```diff
@@ -561,15 +561,18 @@ message ServiceEntry {
...
-    // used with wildcard hosts. Depending on the traffic type, the proxy
-    // will resolve the DNS address specified in the Hosts header or SNI of
-    // the proxied request. Specified endpoints will be ignored. Only
-    // supported for `MESH_EXTERNAL` ServiceEntries and in ambient mode.
-    // The ServiceEntry must be bound to a waypoint.
+    // DYNAMIC_DNS will attempt to resolve the host name specified in
+    // the Host header or SNI to an IP address when handling traffic. This
+    // is particularly useful when multiple dns addresses can be represented
+    // by a single wildcard `host` entry without having to explicitly
+    // enumerate all possible endpoints. DYNAMIC_DNS resolution when
```
ilrudie: I like this, it's more clear than my "front-end" wording while still conveying the important point.

networking/v1alpha3/service_entry.proto
```diff
@@ -561,15 +561,18 @@ message ServiceEntry {
...
+    // egressing the mesh returns the allocated VIP for the matching
+    // wildcard `hosts` specified in the ServiceEntry. Depending on the
```
ilrudie: I don't think this is accurate. DYNAMIC_DNS has no effect on how captured DNS requests are handled by istio/ztunnel. I'm not sure I'd suggest using this verbatim but here's an idea:

```suggestion
    // enumerate all possible endpoints. During DNS proxying, ztunnel 
    // will resolve all subdomains matching the wildcard host name to 
    // a VIP, which could be auto-assigned and won't be routed outside
    // the mesh. `DYNAMIC_DNS` will configure envoy to recover the 
    // original host name using information from SNI or from headers
    // in an HTTP Request. This original host name will then be resolved
    // so that traffic can be routed to the intended IP address.  This method
    // of handling wildcard traffic is not compatible with raw TCP traffic 
    // where the original host cannot be recovered.
```

jaellio: I think I was trying to clarify that before we wouldn't have allocated a VIP for a wildcard ServiceEntry

jaellio: @ilrudie How about this? - (I removed the part about "could be auto assigned" since I wasn't planning on initially allowing users to specify a VIP since it is a noop and isn't used for routing beyond the mesh)

```
    //... During DNS proxying, ztunnel 
    // will resolve all subdomains matching the wildcard host name to 
    // a VIP which isn't used for routing outside
    // the mesh. `DYNAMIC_DNS` will provide configuration to a waypoint proxy to recover the 
    // original host name using information from SNI or a Host header
    // in an HTTP Request. This original host name will then be resolved
    // so that traffic can be routed to the intended IP address.  This method
    // of handling wildcard traffic is not compatible with raw TCP traffic 
    // where the original host cannot be recovered.
```

ilrudie: Sounds good.

networking/v1alpha3/service_entry.proto
```diff
@@ -560,6 +560,23 @@ message ServiceEntry {
     // specified in the hosts field, if wildcards are not used. DNS resolution
     // cannot be used with Unix domain socket endpoints.
     DNS_ROUND_ROBIN = 3;
+
+    // DYNAMIC_DNS will attempt to resolve the host name specified in
+    // the Host header or SNI to an IP address when handling traffic. This
+    // is particularly useful when multiple dns addresses can be represented
+    // by a single wildcard `host` entry without having to explicitly
```
ilrudie: Did we decide we only allow configuring this when using wildcards? The wording suggests it could be used elsewhere but I'm not sure if we strongly disallow that, or only don't recommend it.

ilrudie (commented):
This seems pretty correct now. One question about when this is permitted though. Depending on the answer it might not be that relevant but if we won't configure DYNAMIC_DNS without wildcard we probably should say that more strongly here.

networking/v1alpha3/service_entry.proto
keithmattix: Personally, I'm in favor of a more narrow application and then expand it as we test

jaellio: I agree with being stricter in our official support. I believe in the POC I explicitly disallow non wildcard hosts with `DYNAMIC_DNS` resolution.

jaellio: I have this right now later in the comment

```
`DYNAMIC_DNS` is only supported for wildcard hosts
```

ilrudie: My thinking is that "particularly useful" suggests there are other applications allowed, just we didn't think they would be so useful. We could change it to something like "This is required when using a wildcard host name..." which to my mind doesn't suggest other applications.

I may just be splitting hairs here though. It's likely clear enough and just my mistake reading sentences individually versus in full context.

jaellio: Removed "Particularly useful"

PR #3567, opened 2025-09-22
Migrate storage version of security APIs to v1

kfaseela:
Discussion: https://istio.slack.com/archives/C37A4KAAD/p1758190615806769

Related: https://github.com/istio/istio/issues/55302


evenh (approved):
Thanks for the quick PR 🙌🏻

PR #3577, opened 2025-10-09
Support EnvoyFilter in Waypoint

zirain:
RFC: https://docs.google.com/document/d/1hnYJ1P3hYNBALR_2U2h03xJYhyKtcDHLzBAEVEl_pDc/edit?tab=t.0

`Protocol` and `ClusterMatch(subset)` is not implemented in the first iterate, we can add them back if needed.

### DIFF
--- a/networking/v1alpha3/envoy_filter.proto
+++ b/networking/v1alpha3/envoy_filter.proto
@@ -736,6 +736,8 @@ message EnvoyFilter {
   message WaypointMatch {
     // The service port to match on.
     // If not specified, matches all ports.
+    //
+    // +kubebuilder:validation:XValidation:message="port must be between 1-65535",rule="0 < self && self <= 65535
     uint32 port_number = 1;
 
     // Conditions specified in `RouteMatch` must be met for the patch
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -6649,6 +6649,9 @@ spec:
                               maximum: 4294967295
                               minimum: 0
                               type: integer
+                              x-kubernetes-validations:
+                              - message: port must be between 1-65535
+                                rule: 0 < self && self <= 6553
                             route:
                               description: Match a specific route.
                               properties:
--- a/networking/v1alpha3/envoy_filter.pb.go
+++ b/networking/v1alpha3/envoy_filter.pb.go
@@ -1314,6 +1314,8 @@ type EnvoyFilter_WaypointMatch struct {
 	state protoimpl.MessageState `protogen:"open.v1"`
 	// The service port to match on.
 	// If not specified, matches all ports.
+	//
+	// +kubebuilder:validation:XValidation:message="port must be between 1-65535",rule="0 < self && self <= 65535
 	PortNumber uint32 `protobuf:"varint,1,opt,name=port_number,json=portNumber,proto3" json:"port_number,omitempty"`
 	// Match a specific route.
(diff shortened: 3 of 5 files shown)


networking/v1alpha3/envoy_filter.proto
```diff
@@ -730,6 +733,49 @@ message EnvoyFilter {
     string name = 4;
   }
 
+  message WaypointMatch {
```
howardjohn, line 736: I think we will need CEL validation in place to ensure that you only use waypoint match when mode is WAYPOINT, and that you don't use other matchers when mode is WAYPOINT.

Also for targetRef service, must be WAYPOINT. For targetRef gateway, must be waypoint or gateway or empty?

networking/v1alpha3/envoy_filter.proto
```diff
@@ -832,6 +878,7 @@ message EnvoyFilter {
 
   // One or more match conditions to be met before a patch is applied
   // to the generated configuration for a given proxy.
+  // +kubebuilder:validation:XValidation:message="only support waypointMatch when context is WAYPOINT",rule="(self.context == 'WAYPOINT' && has(self.waypoint)) || (self.context != 'WAYPOINT' && !has(self.waypoint))"
```
howardjohn, line 883: I think this may fail if `self.context` is unset. The `default` function may be handy here

networking/v1alpha3/envoy_filter.proto
howardjohn: nvm I see you have some of these already, nice

networking/v1alpha3/envoy_filter.proto
```diff
@@ -730,6 +733,49 @@ message EnvoyFilter {
     string name = 4;
   }
 
+  message WaypointMatch {
+    // The service port to match on.
+    // If not specified, matches all ports.
+    uint32 port_number = 1;
```
howardjohn, line 741: `// +kubebuilder:validation:XValidation:message="port must be between 1-65535",rule="0 < self && self <= 65535`

networking/v1alpha3/envoy_filter.proto
zirain: add default will execced the limit cost of CEL, add a test case for it.

networking/v1alpha3/envoy_filter.proto
zirain: done

howardjohn (approved):

keithmattix (approved):

networking/v1alpha3/envoy_filter.proto
keithmattix: @howardjohn was right here; this EnvoyFilter currently fails: https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/config/charts/body-based-routing/templates/istio.yaml. Maybe we should add another || expression: `self.context == null && !has(self.waypoint)`

zirain: let me take a look.

zirain: Unfortunately, we have exceeded the limit:
```
 estimated rule cost exceeds budget by factor of 1.048576x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
```

zirain: raised https://github.com/istio/api/pull/3589

PR #3597, opened 2025-11-03
add requireJwt field in RequestAuthentication

tanujd11:
Problem:
Istio's RequestAuthentication has hardcoded allowMissing behavior, requiring users to combine it with AuthorizationPolicy to enforce JWT presence. This results in:
- HTTP status 403 Forbidden instead of 401 Unauthorized for missing JWT which could be a requirement.
- Missing WWW-Authenticate header ([RFC 7235](https://datatracker.ietf.org/doc/html/rfc7235)).

Solution:
- Add bool require_jwt = 15; field to JWTRule message.
- When true: Missing JWT → 401 Unauthorized + WWW-Authenticate: Bearer header from envoy proxy.
- When false (default): Current behavior (JWT optional).

Istio PR will be taken up if this is acceptable.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -16410,6 +16410,12 @@ spec:
                       description: This field specifies the header name to output
                         a successfully verified JWT payload to the backend.
                       type: string
+                    requireJwt:
+                      description: If set to true, requests without a valid JWT token
+                        will be rejected with a 401 Unauthorized status code along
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
IIUC this is not really accidentally missing, but an intentional decision to decouple authentication and authorization. cc @louiscryan @costinm

tanujd11:
With that being said is there anything stopping us from adding this as a feature as this could be a requirement and it is also a feature in envoy where no header mean unauthorised instead of denied which is hardcoded in Istio?

PR #3619, opened 2025-12-26
add priority to workload entry

ramaraochavali:
Adds priority to workload entry. This is especially useful to setup active/passive between two clusters defined by two different DNS based clusters. When endpoints in priority(0), becomes unhealthy -  traffic automatically moves to priority1.

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -7951,6 +7951,11 @@ spec:
                       x-kubernetes-validations:
                       - message: port name must be valid
                         rule: self.all(key, size(key) < 63 && key.matches("^[a-zA-Z0-9](?:[-a-zA-Z0-9]*[a-zA-Z0-9])?$"))
+                    priority:
+                      description: The priority for this Endpoint.
+                      maximum: 4294967295
+                      minimum: 0
+                      type: integer
                     serviceAccount:
                       description: The service account associated with the workload
                         if a sidecar is present in the workload.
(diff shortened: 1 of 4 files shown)


zirain (approved):

howardjohn (changes requested):
How does this relate to the many many other 'priority' mechanisms we have? what if its `priority=3` but in another region with locality LB enabled?

This seems like it could be done today with DR failoverPriority? https://istio.io/latest/docs/reference/config/networking/destination-rule/#LocalityLoadBalancerSetting-failover_priority

ramaraochavali:
It is the same workload entry that has different DNS as endpoints. It is used in the context of ServiceEntry as two different endpoints

ramaraochavali:
Sorry. I meant two workload entries used in the context of ServiceEntry as two different DNS end points. We could create synthetic labels for each to identify priority but that sounds very indirect way of doing it for external endpoints. 
Will write an example and share

ramaraochavali:
>. This is because we donot suport DNS cluster failover yet. Haven't tried if envoy works in this case

It works. I have validated it if we set the correct priority to each LocalityLbEndpoint.

ramaraochavali:
> IMO this is a bug. There is no reason for this other than the implementation in the control plane is wrong because we pass nil here (https://github.com/istio/istio/blob/2eae57b36d1740f8e44e54c0efde07bfb9c706c5/pilot/pkg/networking/core/cluster_traffic_policy.go#L329) but depend on it in the locality code

Yes. This can be fixed. But the point I am making is, for DNS type of Endpoints defined via Service Entries, it seems creating synthetic labels to specify priority seems not intuitive (or indirect way) than specifying the priority directly when defining the WLE.  We can argue if we need a different way to set priority than relying on failoverPriority in DR but IMO it seems simpler to specify priority directly instead of having manage labels especially when there are more than two endpoints in the SE.

ramaraochavali:
> The main concern is that WorkloadEntry can also be used separate object other than endpoints inlined in SE. If we support priority in the spec, it seems lack of flexibility when used separately. 

I agree with this. I think it has partly to do with how Resolution is there in SE and reusing WorkloadEntry as is there. We can say if priority is set for WLE, it is considered and rest of the priority calculations are not done for DNS type of SEs. Other option is to have `repeated Priorities` in SE that maps to endpoints in the same order or introduce name in WLE and have named priorities. None of these are great options but may be better than synthetic label approach.

If we decide not to add any api option, we can always suggest DR option after fixing the bug (But I need to check how complicated the config will be if there are more than two endpoints)

PR #3661, opened 2026-03-06
Add adaptive concurrency filter configuration to DestinationRule

prashanthjos:
Adds support for configuring Envoy's [adaptive concurrency filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/adaptive_concurrency_filter) through Istio's DestinationRule TrafficPolicy.

The adaptive concurrency filter dynamically adjusts the allowed number of outstanding requests to upstream hosts based on sampled latency measurements, providing automatic concurrency limiting without manual tuning.

**Example usage:**

```yaml
kind: DestinationRule
metadata:
  name: my-service-adaptive
spec:
  host: my-service.default.svc.cluster.local
  trafficPolicy:
    adaptiveConcurrency:
      enabled: true
      gradientControllerConfig:
        sampleAggregatePercentile: 50
        concurrencyLimitParams:
          maxConcurrencyLimit: 1000
          concurrencyUpdateInterval: 100ms
        minRttCalcParams:
          interval: 60s
          requestCount: 50
          jitter: 10
          minConcurrency: 3
          buffer: 25
      concurrencyLimitExceededStatus: 503
```


prashanthjos:
[This](https://github.com/istio/istio/issues/25991) is the original issue, and [this](https://docs.google.com/document/d/1pT_bfVf7jnYFhbsCMNDe29OqXK3wUrW5/edit?pli=1) is the proposal document.

networking/v1alpha3/destination_rule.proto
```diff
@@ -307,6 +307,11 @@ message TrafficPolicy {
 
   // Specifies a limit on concurrent retries in relation to the number of active requests.
   RetryBudget retry_budget = 8;
+
+  // Adaptive concurrency settings for dynamically adjusting the allowed number of
+  // outstanding requests based on sampled latencies. This enables the Envoy
+  // adaptive concurrency filter for the destination.
+  AdaptiveConcurrency adaptive_concurrency = 9;
```
ericdbishop, line 337: * Does this completely override the static `max_connections` circuit breaker threshold? Should we describe that here so that its more clear what to expect if both are configured?

* Is it also worth noting that new connections that exceed the adaptive concurrency threshold won't be reflected in [circuit breaker statistics](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#circuit-breakers-statistics), the way they would when connections exceed the `max_connections` threshold? (At least that is how I understand it).

ericdbishop: Normally when the circuit breaker threshold is exceeded in Envoy, you also have the `x-envoy-overloaded` header returned as part of the response. Some users might rely on that header, or maybe the client-side Envoy proxy relies on that in some way.

Should we note that the `x-envoy-overloaded` header won't be present when the adaptive concurrency threshold is exceeded?

prashanthjos: > * Does this completely override the static `max_connections` circuit breaker threshold? Should we describe that here so that its more clear what to expect if both are configured?
> * Is it also worth noting that new connections that exceed the adaptive concurrency threshold won't be reflected in [circuit breaker statistics](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#circuit-breakers-statistics), the way they would when connections exceed the `max_connections` threshold? (At least that is how I understand it).

@ericdbishop these are good points, I digged a little bit more into Envoy code, here is their adaptive filter. When the adaptive concurrency filter blocks a request, it does so here , the request never reaches the router at all:

```
  if (controller_->forwardingDecision() == Controller::RequestForwardingAction::Block) {
    decoder_callbacks_->sendLocalReply(config_->concurrencyLimitExceededStatus(),
                                       "reached concurrency limit", nullptr, absl::nullopt,
                                       "reached_concurrency_limit");
    return Http::FilterHeadersStatus::StopIteration;
```

Whereas circuit breakers are checked much later, inside the connection pool when the router tries to establish an upstream connection:

```
  const bool can_create_connection = host_->canCreateConnection(priority_);
  if (!can_create_connection) {
    host_->cluster().trafficStats()->upstream_cx_overflow_.inc();
  }
```
```
  if (!host_->cluster().resourceManager(priority_).pendingRequests().canCreate()) {
    ENVOY_LOG(debug, "max pending streams overflow");
    onPoolFailure(nullptr, absl::string_view(), ConnectionPool::PoolFailureReason::Overflow,
                  context);
    host_->cluster().trafficStats()->upstream_rq_pending_overflow_.inc();
    return nullptr;
  }
```
So, No, Adaptive Concurrency Does NOT Override Circuit Breakers. If both are configured, a request must pass both checks independently:
- First, the adaptive concurrency filter checks if num_rq_outstanding < concurrency_limit. If not, the request is rejected immediately with a 503.
- If it passes, the request continues through the filter chain to the router, where circuit breaker thresholds (max_connections, max_pending_requests, max_requests) are checked independently at the connection pool level.

In practice, whichever limit is more restrictive will be the binding constraint. If the adaptive concurrency limit is lower than max_connections, most requests will be shed by the filter before circuit breakers ever see them.

When the adaptive concurrency filter blocks a request, it increments only its own stat:
- `adaptive_concurrency.gradient_controller.rq_blocked`

Since the request never reaches the router or connection pool, none of the circuit breaker stats are affected:
- No `upstream_cx_overflow_`
- No `upstream_rq_pending_overflow_`
- No `upstream_rq_retry_overflow_`

Someone monitoring only circuit breaker stats would not see that requests are being shed.This makes the three key facts explicit:

- They don't override each other, they're layered
- They limit different things at different points
- They have separate stats, so operators need to monitor both

This is part of Envoy and little we can do with Istio, In any case I have updated the comment.

prashanthjos: > Normally when the circuit breaker threshold is exceeded in Envoy, you also have the `x-envoy-overloaded` header returned as part of the response. Some users might rely on that header, or maybe the client-side Envoy proxy relies on that in some way.
> 
> Should we note that the `x-envoy-overloaded` header won't be present when the adaptive concurrency threshold is exceeded?

This is true, the header is not added, I have updated the comment !

ericdbishop: The notes you added to the comment look good to me, so at least users are aware of the differences. Thanks!

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,115 @@ message OutlierDetection {
...
+    }
+
+    // Parameters controlling the periodic minRTT recalculation.
+    // The minRTT is periodically measured by allowing only a very low outstanding
+    // request count to an upstream cluster and measuring the latency under
+    // these ideal conditions.
+    message MinimumRTTCalculationParams {
+      // The time interval between recalculating the minimum request round-trip time.
+      // Has to be positive. If set to zero, dynamic sampling of the minRTT is disabled.
+      google.protobuf.Duration interval = 1;
```
keithmattix: What is the default?

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,115 @@ message OutlierDetection {
...
+
+      // The fixed value for the minRTT. This value is used when minRTT is not sampled
+      // dynamically. If dynamic sampling of the minRTT is disabled (i.e. `interval` is
+      // set to zero), this field must be set.
+      google.protobuf.Duration fixed_value = 6;
```
keithmattix: So at least one of fixed_value and or interval must be set? Should it be a oneof?

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,115 @@ message OutlierDetection {
...
+
+      // The number of requests to aggregate/sample during the minRTT recalculation
+      // window before updating. Defaults to 50.
+      google.protobuf.UInt32Value request_count = 2;
+
+      // Randomized time delta that will be introduced to the start of the minRTT
+      // calculation window. This is represented as a percentage of the interval duration.
+      // Defaults to 15%. This is recommended to prevent all hosts in a cluster from
+      // being in a minRTT calculation window at the same time.
+      google.protobuf.DoubleValue jitter = 3;
```
keithmattix: Does this really need to be a double if, as in the example, folks are setting whole number percentages? Same goes for buffer

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,115 @@ message OutlierDetection {
...
+    google.protobuf.DoubleValue sample_aggregate_percentile = 1;
+
+    // Parameters for periodic recalculation of the concurrency limit.
+    ConcurrencyLimitCalculationParams concurrency_limit_params = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // Parameters for periodic minRTT recalculation.
+    MinimumRTTCalculationParams min_rtt_calc_params = 3 [(google.api.field_behavior) = REQUIRED];
+  }
+
+  oneof concurrency_controller_config {
```
keithmattix, line 1032: Why is this a oneof?

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,115 @@ message OutlierDetection {
...
+    // Gradient concurrency control will be used.
+    GradientControllerConfig gradient_controller_config = 1;
+  }
+
+  // If set to false, the adaptive concurrency filter will operate as a pass-through
+  // filter. If unspecified, the filter will be enabled.
```
keithmattix: This is going to change every single Envoy deployment to use adaptive concurrency right? That is very likely to throttle people's production traffic

keithmattix (changes requested):

networking/v1alpha3/destination_rule.proto
sschepens: i have another question, does this field even make sense? If I wanted to disable it, wouldn't I not specify `adaptiveConcurrency` directly?

ramaraochavali: +1. I have asked the same question in the design doc also

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,115 @@ message OutlierDetection {
...
+message AdaptiveConcurrency {
+  // Configuration parameters for the gradient concurrency controller.
+  // The gradient controller makes forwarding decisions based on a periodically
+  // measured ideal round-trip time (minRTT) for an upstream.
+  message GradientControllerConfig {
+    // Parameters controlling the periodic recalculation of the concurrency limit
+    // from sampled request latencies.
+    message ConcurrencyLimitCalculationParams {
+      // The allowed upper-bound on the calculated concurrency limit. Defaults to 1000.
+      google.protobuf.UInt32Value max_concurrency_limit = 2;
```
ramaraochavali, line 932: How would service owners arrive at what is the correct value for this?

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,115 @@ message OutlierDetection {
...
+      google.protobuf.UInt32Value request_count = 2;
+
+      // Randomized time delta that will be introduced to the start of the minRTT
+      // calculation window. This is represented as a percentage of the interval duration.
+      // Defaults to 15%. This is recommended to prevent all hosts in a cluster from
+      // being in a minRTT calculation window at the same time.
+      google.protobuf.DoubleValue jitter = 3;
+
+      // The concurrency limit set while measuring the minRTT. Defaults to 3.
+      google.protobuf.UInt32Value min_concurrency = 4;
```
ramaraochavali: What does this indicate and how would service owner know what to set?

ramaraochavali (changes requested):
We need to give guidance on how would service owner sets some of the values what are the implications of setting a specific value with proper examples. This is very hard to configure and get it right unless we provide proper documentation and guidance.

networking/v1alpha3/destination_rule.proto
ericdbishop: This is the lower bound for concurrency (i.e. ensure the adaptive concurrency filter doesn't throttle the concurrency below this threshold). I think most of these descriptions are the same as the [Envoy API reference](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/adaptive_concurrency/v3/adaptive_concurrency.proto#envoy-v3-api-msg-extensions-filters-http-adaptive-concurrency-v3-gradientcontrollerconfig). 

The [main docs page](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/adaptive_concurrency_filter) explains the calculation with more detail, but it is a lot to interpret for the user. Even the defaults for adaptive concurrency [used at Lyft were pretty different than the Envoy defaults](https://www.youtube.com/watch?v=CQvmSXlnyeQ), and some settings might have to tuned at the per-service level.

With all of the tuning available, this is overall an advanced feature, maybe this PR should be accompanied by an istio.io docs user guide?

networking/v1alpha3/destination_rule.proto
prashanthjos: I think its a good suggestion but these fields are not a oneof in the [upstream Envoy v3 proto](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/adaptive_concurrency/v3/adaptive_concurrency.proto) also. I've kept the structure aligned with Envoy's proto to maintain a clear mapping for the control plane translation layer and updated the field comments instead, if interval is set, fixed_value is ignored.

Happy to switch to a oneof if we prefer a stricter Istio-native API over Envoy alignment. Let me know your preference.

networking/v1alpha3/destination_rule.proto
prashanthjos: interval has no default, users must either set interval (to enable dynamic minRTT sampling) or set fixed_value (to use a static minRTT). If you set neither, config validation throws an exception.

networking/v1alpha3/destination_rule.proto
prashanthjos: I updated the comments to the best of my knowledge :) , unfortunately this isn't documented well enough in Envoy also. I would let any experts correct me here and I would be happy to change the comments.

networking/v1alpha3/destination_rule.proto
prashanthjos: hmm fair point, changed jitter, buffer, and sample_aggregate_percentile from DoubleValue to UInt32Value. These are whole-number percentages (0–100) in practice, and UInt32Value still preserves nullability so unset values correctly fall back to Envoy's defaults (15%, 25%, p50 respectively).

Note: Envoy's proto uses type.v3.Percent which wraps a double, but since Istio doesn't import Envoy types and fractional percentages aren't practically useful here, UInt32Value is a better fit and consistent with other fields in this message (request_count, min_concurrency). Let me know what you think.

networking/v1alpha3/destination_rule.proto
prashanthjos: Yes we can do that, while merging the implementation PR, I will try my best to do justice to the documentation :) .

networking/v1alpha3/destination_rule.proto
prashanthjos: It's a oneof in the [Envoy v3 proto](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/adaptive_concurrency/v3/adaptive_concurrency.proto) to allow for future concurrency controller types beyond the gradient controller, but from what I can understand, Envoy has only ever had the gradient controller (since the filter was introduced), so in practice there's only one option. We can still drop OneOf but if we have to align with Envoy Proto we can let it stay. 

let me know your thoughts I am open to change to plain field.

networking/v1alpha3/destination_rule.proto
prashanthjos: +1, agreed. Removed the enabled field. I was dedicatedly following the Envoy's spec. Diving a little deep into Envoy I understood in Envoy's filter chain model, enabled (as a RuntimeFeatureFlag) allows toggling the filter at runtime without removing the config. Since Istio doesn't expose Envoy's runtime feature flags, the field add's no value here.

Apologies @ramaraochavali I missed your comment in the design doc.

networking/v1alpha3/destination_rule.proto
keithmattix: There's not a practical difference for users since none of this is user facing, but I would prefer a oneOf personally for developers because it makes the semantics clear without having to read other fields' documentation

networking/v1alpha3/destination_rule.proto
keithmattix: Yep, this is what I was looking for; it should help prevent bugs in the future

networking/v1alpha3/destination_rule.proto
keithmattix: Ack, future proofing SGTM

networking/v1alpha3/destination_rule.proto
prashanthjos: @keithmattix I changed to oneOf

networking/v1alpha3/destination_rule.proto
```diff
@@ -307,6 +307,30 @@ message TrafficPolicy {
 
   // Specifies a limit on concurrent retries in relation to the number of active requests.
   RetryBudget retry_budget = 8;
+
+  // Adaptive concurrency settings for dynamically adjusting the allowed number of
+  // outstanding requests based on sampled latencies. This enables the Envoy
```
ramaraochavali: nit: This is an implementation detail that it enables Envoy's adaptive concurrency filter. Need not be captured in the api docs

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,140 @@ message OutlierDetection {
   int32 min_health_percent = 5;
 }
 
+// Adaptive concurrency settings for dynamically adjusting the allowed number of
+// outstanding requests based on sampled latencies. See Envoy's
+// [adaptive concurrency filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/adaptive_concurrency_filter)
+// for more details.
+//
+// The following example configures adaptive concurrency for the my-service
+// service with a gradient controller:
```
ramaraochavali: Can you please explain here what this configuration will do if a service configures this instead of just saying "configures adaptive concurrency"?

networking/v1alpha3/destination_rule.proto
ramaraochavali: Does this mean never send more requests than the limit configured here?
Can this be understood as "When deciding to cap the concurrency, do not ever allow more than this number irrespective of what gradient controllers determines based on observed latency" ? Going back to my question - would services have to test their service under some load and come up with this number?

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +841,140 @@ message OutlierDetection {
...
+        // When set, the controller periodically enters a low-concurrency measurement
+        // window (pinned to `min_concurrency`) to re-measure the baseline latency.
+        // During this window, throughput is temporarily reduced, so longer intervals
+        // reduce overhead but may react slower to changes in upstream performance.
+        google.protobuf.Duration interval = 1;
+
+        // A fixed value for the minRTT. Must be positive.
+        // Use this when the baseline latency of the upstream is well-known and
+        // stable, to avoid the periodic low-concurrency measurement windows.
+        google.protobuf.Duration fixed_value = 6;
```
ramaraochavali: Don't understand what is this duration corresponds to? What does it translate to in Envoy?

networking/v1alpha3/destination_rule.proto
ramaraochavali: But the key point about this is, during the minRTT calculation window, the filter allows only "3" requests the default - So depending on the interval, we may see more requests failing with 503.

Also what is our recommendation on how often this should be calculated to avoid these errors? So the interplay between min_concurrency and interval - What is user supposed to do to avoid these errors during recalculation window.

ramaraochavali (commented):
Thank you for addressing comments. Added few more comments.

~~Currently this is Alpha in Envoy - Should we check on what is their plan to move this out of Alpha?~~ . Sorry was looking at wrong proto.

Also asked [here](https://github.com/envoyproxy/envoy/issues/17647#issuecomment-4111117306) production usage feedback

networking/v1alpha3/destination_rule.proto
prashanthjos: Updated to the best of my knowledge.

networking/v1alpha3/destination_rule.proto
prashanthjos: Yes, that's right, `max_concurrency_limit` is a hard ceiling that means "never allow more outstanding requests than this, regardless of what the gradient algorithm calculates." The algorithm dynamically computes a limit each cycle using `new_limit = old_limit × gradient + headroom`, where g`radient = (minRTT + buffer) / sampleRTT` and headroom = `sqrt(limit)`. The result will never exceed max_concurrency_limit. This cap doesn't influence the algorithm's calculation, it only limits the final value.

In most cases, service owners don't need to load test to set this. The default of 1000 is intentionally high so the algorithm can self-regulate based on observed latency without hitting the cap, it will typically settle well below 1000 on its own, which you can verify via the `concurrency_limit` gauge. You'd only want to lower it if you know from load testing or production observation that the service (or a downstream dependency) degrades above a specific concurrency and you want a hard safety net beyond the algorithm's own backoff behavior.

networking/v1alpha3/destination_rule.proto
prashanthjos: I have updated the comments on the field, I tried to derive this from envoy documentation, 

fixed_value is the assumed baseline round-trip time (minRTT) of the upstream, expressed as a duration (e.g., 5ms, 10ms). It maps directly to fixed_value in Envoy's MinimumRTTCalculationParams.

The gradient controller uses the minRTT as the baseline in its core formula: gradient = (minRTT + buffer) / sampleRTT. Normally, the minRTT is dynamically sampled by periodically dropping to low concurrency and measuring the latency. fixed_value lets you hardcode this baseline instead, e.g., fixed_value: 5ms means "assume my service responds in 5ms under ideal conditions." This skips the periodic measurement windows that temporarily reduce throughput, making it useful when the upstream's baseline latency is well-known and stable. I've updated the comment to clarify this

networking/v1alpha3/destination_rule.proto
prashanthjos: I updated in the comments

prashanthjos:
> Thank you for addressing comments. Added few more comments.
> 
> ~Currently this is Alpha in Envoy - Should we check on what is their plan to move this out of Alpha?~ . Sorry was looking at wrong proto.
> 
> Also asked [here](https://github.com/envoyproxy/envoy/issues/17647#issuecomment-4111117306) production usage feedback

Looks like they are no longer work at Lyft to provide production data.

networking/v1alpha3/destination_rule.proto
ramaraochavali: Another difference that may have call out is this is HTTP specific and only applied on the server side

networking/v1alpha3/destination_rule.proto
ramaraochavali: So when this is set, none of the minRTT calculation params like request_count etc are needed right? It just uses this value as the latency? If that is the case, the current one of state is not correct?

May be we should segregate DynamicSampling and FixedLatency as two different messages and have oneof between them? 

@keithmattix WDYT? Previous discussion around this https://github.com/istio/api/pull/3661#discussion_r2921372106

ramaraochavali (commented):
Thank you for your patience and iterating on this. Left few more comments.

Also I did not see mention of possible 503s during minRTT calculation windows and having client side retries.  This is a very important consideration that service owner need for enabling this and we should highlight it.  See https://github.com/envoyproxy/envoy/issues/18152#issuecomment-921261981. 

Also we need to mention this is not supported for gRPC streaming and WebSockets


Overall I think this is looking good.  There are too many knobs for users to configure and reason about.  The other concern is we are copying Envoy's schema as is without abstracting/simplifying it for user. Don't have better ideas here but @louiscryan or other @istio/technical-oversight-committee members may have some ideas.

networking/v1alpha3/destination_rule.proto
ericdbishop: I don't think increased 503s are avoidable. But since the configured jitter ensures that the minRTT calculation does not occur for all hosts at once, as long as the user has configured static count retries on the VirtualService / HTTPRoute the request should be retried to another endpoint and ultimately succeed.

ramaraochavali: Yes. I want to document that and also keep in mind that 503s are not safe for retry in all cases - so users have to make a conscious choice.

networking/v1alpha3/destination_rule.proto
prashanthjos: @ramaraochavali actually this is a good point, the filter should be applied on the server side (inbound), so that the filter has full view of the requests that are coming in, below is what the envoy documentation says 

> The filter works as intended in the filter chain for a local cluster.
The filter must be able to limit the concurrency for a cluster. This means there must not be requests destined for a cluster that are not decoded by the adaptive concurrency filter.

From my knowledge of DestinationRules, traffic policies are traditionally applied on the client-side (outbound) proxy.  I wanted to highlight this important deviation since I never called this out explicitly.

I have updated the comments accordingly, thanks for bringing this up.

networking/v1alpha3/destination_rule.proto
prashanthjos: Rama, I agree but wanted to see if anyone else feels if this is necessary, Envoy doesn't have this oneOf and has all fields flat in MinimumRTTCalculationParams.

I updated the code, but if others have a different opinion, please do let me know.

ramaraochavali: I think either what Envoy has (no oneof) or have them group together as I suggested above works better. But in the current state IMO, it is partial - the oneof exists at one place and rest of configuration is not dependant on that oneof but only applies if user selects one in that oneof. It seems confusing to me.

prashanthjos: @ramaraochavali I updated the code yesterday to use oneOf.

prashanthjos:
> Thank you for your patience and iterating on this. Left few more comments.
> 
> Also I did not see mention of possible 503s during minRTT calculation windows and having client side retries. This is a very important consideration that service owner need for enabling this and we should highlight it. See [envoyproxy/envoy#18152 (comment)](https://github.com/envoyproxy/envoy/issues/18152#issuecomment-921261981).
> 
> Also we need to mention this is not supported for gRPC streaming and WebSockets
> 
> Overall I think this is looking good. There are too many knobs for users to configure and reason about. The other concern is we are copying Envoy's schema as is without abstracting/simplifying it for user. Don't have better ideas here but @louiscryan or other @istio/technical-oversight-committee members may have some ideas.

@ramaraochavali @keithmattix @louiscryan what do you guys think overall about the PR, I know there are many questions about the number of knobs and the tuning involved, should we release this feature as Experimental (Alpha with feature flags) where it is disabled by default and based on usage we can either modify or upgrade.

networking/v1alpha3/destination_rule.proto
keithmattix: Wait this filter is applied on the server side?? Why is it on destination rule then? I thought the idea was to limit requests to upstream

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +845,202 @@ message OutlierDetection {
...
+        // The number of requests to aggregate/sample during the minRTT
+        // recalculation window before updating. Defaults to 50.
+        google.protobuf.UInt32Value request_count = 2;
+
+        // Randomized time delta that will be introduced to the start of the minRTT
+        // calculation window. This is represented as a percentage of the interval
+        // duration. Defaults to 15%. Value must be in the range [0, 100].
+        // This is recommended to prevent all hosts in a cluster from
+        // being in a minRTT calculation window at the same time.
+        google.protobuf.UInt32Value jitter = 3;
```
keithmattix, line 965: Can we just pick a default here? We can always add if needed but can't take away

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +845,202 @@ message OutlierDetection {
...
+      // a low-concurrency measurement window.
+      message DynamicSampling {
+        // The time interval between recalculating the minimum request round-trip
+        // time. Must be >= 1ms.
+        //
+        // When set, the controller periodically enters a low-concurrency measurement
+        // window (pinned to `min_concurrency`) to re-measure the baseline latency.
+        // During this window, throughput is temporarily reduced, so longer intervals
+        // reduce overhead but may react slower to changes in upstream performance.
+        google.protobuf.Duration interval = 1;
```
keithmattix, line 954: I know envoy requires this, but I think we can probably do better and set a default here.

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +845,202 @@ message OutlierDetection {
...
+      // Setting this too low may prevent the algorithm from reaching the upstream's
+      // true capacity.
+      //
+      // Monitor the `concurrency_limit` gauge to observe where the algorithm settles
+      // under normal load.
+      google.protobuf.UInt32Value max_concurrency_limit = 2;
+
+      // The period of time samples are taken to recalculate the concurrency limit.
+      // This is required.
+      google.protobuf.Duration concurrency_update_interval = 3 [(google.api.field_behavior) = REQUIRED];
```
keithmattix, line 936: Again let's pick a default here

networking/v1alpha3/destination_rule.proto
keithmattix: Then let's not add this at first. Let's stick with envoy's default for now

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +845,202 @@ message OutlierDetection {
...
+      // In most cases the default of 1000 is sufficient, since the algorithm will
+      // self-regulate well below this ceiling based on observed latencies. Consider
+      // lowering it only if you know the upstream's maximum sustainable concurrency
+      // and want a hard safety net beyond the algorithm's own backoff behavior.
+      // Setting this too low may prevent the algorithm from reaching the upstream's
+      // true capacity.
+      //
+      // Monitor the `concurrency_limit` gauge to observe where the algorithm settles
+      // under normal load.
+      google.protobuf.UInt32Value max_concurrency_limit = 2;
```
keithmattix, line 932: nit: why not use field 1?

networking/v1alpha3/destination_rule.proto
```diff
@@ -817,6 +845,202 @@ message OutlierDetection {
...
+        google.protobuf.Duration fixed_latency = 2;
+      }
+
+      // Amount added to the measured minRTT to add stability to the concurrency
+      // limit during natural variability in latency. This is expressed as a
+      // percentage of the measured value and can be adjusted to allow more or
+      // less tolerance to the sampled latency values. Defaults to 25%.
+      // Value must be in the range [0, 100].
+      // This applies regardless of whether minRTT is dynamically sampled or fixed.
+      google.protobuf.UInt32Value buffer = 3;
```
keithmattix, line 1018: Let's stick to envoy defaults here and not expose this until a user really needs it

keithmattix (commented):
After coming back around to this proposal and seeing the discussion, I'm even more concerned about the sheer complexity of this feature. I think we should start with far fewer knobs and allow users to tell us when they need more control

networking/v1alpha3/destination_rule.proto
sschepens: Yes, adaptativeConcurrency must be configured server side, it needs to be on a Listener that calls the local cluster.

TBH DestinationRule currently is a mixed bag of client and server related configs. Circuit Breakers are applied client and server side for example. And I think there are other examples. Though applying on server side is trickier because multiple services can match the same workload.

I don't think DestinationRule is a bad place for putting adaptativeConcurrency, might be the best option we have. But we have to decide if we want to commit on a stable API for this or just rely on EnvoyFilters, configuring EnvoyFilter for client side behaviors is a pain, but configuring it for server side is pretty straightforward.

ramaraochavali: +1 for Destination Rule. That is the best possible place  if at all we want to support this stable API. But as I mentioned else where, the sheer complexity of the features, knobs that it exposes are hard to get right unless we have super clear docs and examples. Another point is about the production usage of it in real world. 

> configuring EnvoyFilter for client side behaviors is a pain, but configuring it for server side is pretty straightforward.

Yes. Specially since this is a new filter - it is easy to configure via Envoy filter

Another approach from Istio side we can take is document it like how we do [Global Ratelimit feature](https://istio.io/latest/docs/tasks/policy-enforcement/rate-limit/)  with sample Envoy filters. The downside of it is we do not have stable and easy to use API.

prashanthjos:
@ramaraochavali any final thoughts on this PR, this is open for a while.

frittentheke:
@keithmattix @ramaraochavali Would also love to know if this PR could be landing in 1.30 by any chance?

prashanthjos:
@frittentheke I am waiting for @ramaraochavali final comments, so that I can address all the comments together including the ones asked by @keithmattix

ramaraochavali:
My main concerns were complexity of configuration and too many knobs which if we expose fewer knobs a @keithmattix suggested should mitigate - The question I still have is would those defaults make sense for all use cases and people would still be to use this feature effectively?

@louiscryan @howardjohn any ideas on simplifying this?

frittentheke:
I totally agree with @keithmattix's remarks about setting sane defaults.
This feature should not remain some hidden, expert-level special config, but be widely applicable.
To me, the machine (load-balancing) should become "adaptive" to auto-adjust to the elasticity / concurrency behavior of applications. The values required / used should not be something that the user should require to change for MOST use cases. Because even if a suitable set of values could be determined via e.g. some benchmark by the users, this very unlikely is then not done broadly or even repeated if the software stack or load pattern should change.

To me the whole idea is to use Envoys existing capabilities better -- in the context of Istio. I'd imagine a mesh wide toggle to enable "adaptive" concurrency by default at some point, just like a change to the default load balancing to something latency sensitive like PeakEWMA / Prequel to have all services benefit automatically.
This both is not about taking control away from the user, but to actually accept "efficiency" and "smart load balancing" as fundamental features to the mesh's role in an installation.

PR #3665, opened 2026-03-09
Add trustDomains and notTrustDomains to AuthzPolicy source

ymesika:
This PR introduces `trustDomains` and `notTrustDomains` fields to the `AuthorizationPolicy` Source rule.
This allows users to accurately and safely match against the identity's trust domain without resorting to potentially error-prone glob-style matching on the full principal URI.

This proposal is fully backward compatible. The addition of optional fields will not impact existing AuthorizationPolicy resources that do not use the new trustDomains fields.

Example:
```yaml
kind: AuthorizationPolicy
metadata:
  name: allow-some-trust-domains
  namespace: default
spec:
  action: ALLOW
  rules:
  - from:
    - source:
        trustDomains: ["cluster1.local", "cluster2.local"]
```

Or:

```yaml
kind: AuthorizationPolicy
metadata:
  name: allow-all-but-specific
  namespace: default
spec:
  action: ALLOW
  rules:
  - from:
    - source:
        notTrustDomains: ["not.trust.domain"]
```


dgn (approved):

security/v1beta1/authorization_policy.proto
```diff
@@ -506,6 +506,21 @@ message Source {
...
+  // Optional. A list of trust domains derived from the peer certificate.
+  // Can be exact, prefix, suffix and presence.
+  // This field requires mTLS enabled and is the same as the `source.trustDomain` attribute.
+  //
+  // If not set, any trust domain is allowed.
+  //
+  // +cue-gen:AuthorizationPolicy:releaseChannel:extended
+  repeated string trust_domains = 13;
+
+  // Optional. A list of negative match of trust domains.
```
keithmattix, line 519: What are the semantics of notTrustDomains is set but the connection doesn't have mTLS enabled?

security/v1beta1/authorization_policy.proto
```diff
@@ -506,6 +506,21 @@ message Source {
 
   // Optional. A list of negative match of remote IP blocks.
   repeated string not_remote_ip_blocks = 10;
+
+  // Optional. A list of trust domains derived from the peer certificate.
+  // Can be exact, prefix, suffix and presence.
```
keithmattix, line 511: Based on the example, it seems like it's more global based with * vs an envoy string matcher type of API. Does it only match whole components of a SPIFFE URI? Can you do, eg "fo*-bar"?

ymesika: The matching semantics are clearly [documented above](https://github.com/ymesika/api/blob/a23a5bc7d3b28a0ca57056462867817b380479c7/security/v1beta1/authorization_policy.proto#L383-L388). The trustdomains field supports the same pattern matching as all other string fields in the rule (suffix, prefix, exact or presence).
So no, you cannot do `fo*-bar`. But that's something to keep in mind to validate in the impl.

security/v1beta1/authorization_policy.proto
ymesika: I think it's the same trap as `notPrinicipals`/`notNamespaces`.
For those, in positive values we have a comment that it requires mTLS and I also added it (https://github.com/ymesika/api/blob/a23a5bc7d3b28a0ca57056462867817b380479c7/security/v1beta1/authorization_policy.proto#L512) but we lack this comment on the negative ones.
I tried to be consistent with them.

ymesika: BTW, for all those negative forms when mTLS is absent, not_* will trivially pass (empty string doesn't match any specified value), potentially allowing unauthenticated traffic through.

security/v1beta1/authorization_policy.proto
keithmattix: Ack, thanks for the pointer

keithmattix (approved):

PR #3671, opened 2026-03-18
add compliance policy to gateway API

ramaraochavali:
There is no easy way to ensure the compliance in some secure envs, that blocks *CHACHA* cipher for TLS 1.3. Envoy supports this compliance policy now and is a common requirement for gateways. For gateway API, we can support it as one of the options.

### DIFF
--- a/networking/v1alpha3/gateway.proto
+++ b/networking/v1alpha3/gateway.proto
@@ -579,4 +579,26 @@ message ServerTLSSettings {
   // * `AES256-SHA`
   // * `DES-CBC3-SHA`
   repeated string cipher_suites = 9;
+
+  enum CompliancePolicy {
+    // FIPS_202205 configures a TLS connection to use:
+    //
+    //   * TLS 1.2 or 1.3
+    //   * For TLS 1.2, only ECDHE_[RSA|ECDSA]_WITH_AES_*_GCM_SHA*.
+    //   * For TLS 1.3, only AES-GCM
+    //   * P-256 or P-384 for key agreement.
+    //   * For server signatures, only ``PKCS#1/PSS`` with ``SHA256/384/512``, or ECDSA
+    //     with P-256 or P-384.
+    //
+    // .. attention::
+    //
+    //   Please refer to `BoringSSL policies <https://boringssl.googlesource.com/boringssl/+/refs/tags/0.20240913.0/include/openssl/ssl.h#5608>`_
+    //   for details.
+    FIPS_202205 = 0;
+  }
+
+  // Optional list of compliance policies. Compliance policies configure various aspects of the TLS
+  // based on the given policy. The policies are applied last during configuration and may override
+  // the other TLS parameters, or any previous policy.
+  repeated CompliancePolicy compliance_policies = 17;
 }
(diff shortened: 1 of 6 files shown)


networking/v1alpha3/gateway.proto
```diff
@@ -579,4 +579,26 @@ message ServerTLSSettings {
...
+    //
+    //   Please refer to `BoringSSL policies <https://boringssl.googlesource.com/boringssl/+/refs/tags/0.20240913.0/include/openssl/ssl.h#5608>`_
+    //   for details.
+    FIPS_202205 = 0;
+  }
+
+  // Optional list of compliance policies. Compliance policies configure various aspects of the TLS
+  // based on the given policy. The policies are applied last during configuration and may override
+  // the other TLS parameters, or any previous policy.
+  repeated CompliancePolicy compliance_policies = 17;
```
howardjohn, line 603: I think this is problematic on Gateway... I think the idea was this was a per-proxy setting or even global istiod setting. 1 pod can have multiple Gateways which would make this tricky. @keithmattix wdyt

ramaraochavali: Are you referring to the new Envoy compliance policy added to [CommonTLSContext](https://www.envoyproxy.io/docs/envoy/v1.37.1/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-msg-extensions-transport-sockets-tls-v3-tlsparameters) or compliance policy that Istio added?

I was thinking of supporting Envoy one and it can be supported per gateway right - this is similar to Cipher suites for example?

howardjohn: ah. Istio already has this concept though. I think the istio concept should translate to the envoy one probably?

I (now) agree this could work as you implemented, just not sure its the best or just expand our existing compliance policy

keithmattix: Selfishly, my concern is that this compliance policy in envoy is boringSSL only and has dubious/non existent support for other crypto implementations. I do agree it could work as described though.

As to our existent compliance policy, the lines are kinda blurry now that envoy/boring supports this first class. Should we have separate control plane/data plane compliance policy fields?

ramaraochavali: > I think the istio concept should translate to the envoy one probably?

That is a possibility.  But AFAIK, the Istio compliance policy is not pod/proxy specific. If we set at Istiod it works for all components and all proxies. If we want to enable only at gateways it is not possible and also today's compliance policy does not allow 1.3. If we want to do pod/proxy specific, we have to introduce ISTIO_META and drive it through that. Should we do that or alternatively expose this API and if this is specified, for gateways it takes precedence that way it can be enabled for gateways only

> Should we have separate control plane/data plane compliance policy fields?

One challenge with this is, this envoy policy applies to Envoy and if we broadly classify it as data plane, we can not apply to pilot-agent - which probably is fine because pilot-agent though part of dataplane mostly plays control plane component role?

ramaraochavali: @keithmattix @howardjohn WDYT ^^?

ramaraochavali:
closing this infavour of https://github.com/istio/istio/pull/59847

PR #3678, opened 2026-03-20
Expose dns_respect_ttl option in MeshConfig

leosarra:
See https://github.com/istio/istio/pull/59555 for context


leosarra:
> [istio/istio#16173](https://github.com/istio/istio/pull/16173) - context on why we switched to respect_dns_ttl. If you do this as Mesh Config property, we would reverse that issue - May be you need it configurable per SE where you have problem? BTW, why is DNS TTL so low?

Hey,
Apologies for the delay, I was on vacation over the past month.

If we are to expose dns_respect_ttl in SE then I think it would also make sense to expose also the DNS refresh rate as a configurable option inside ServiceEntries, instead of keeping it only at a global level. Also, maybe we should support both a global setting and a ServiceEntry specific option for this. WDYT?

As for the low DNS TTL, in our setup we run a cluster with CoreDNS and NodeLocal DNS, together with Istio (with SmartDNS enabled). Pods resolve names through NodeLocal DNS, where the TTL is intentionally kept low, as this is required for cluster.local services.

howardjohn:
you could configure coredns to return a low ttl only for cluster.local if
you wanted I think

PR #3688, opened 2026-04-10
Fix EnvoyFilter port validation

### LINKED ISSUE #59832: truncated upper port value '6553' in validation for EnvoyFilter v1alpha3
### Is this the right place to submit this?

- [x] This is not a security vulnerability or a crashing bug
- [x] This is not a question about how to use Istio

### Bug Description

EnvoyFilter v1alpha3 has a typo in the port validation in the 1.29.1 chart. I cannot find it in the code because it appears to be fully autogenerated, so really should not happen. 

### Version

```Text
1.29.1 of the chart.
```

### Additional Information

You can see the [problem line here](https://github.com/istio/istio/blob/189832a289c1aaf0e868932b4c46e0059638ea74/manifests/charts/base/files/crd-all.gen.yaml#L7014)

or in the chart:

```
$ helm pull istio/base --version 1.28.9 --untar
$ grep -w "6553" base/files/crd-all.gen.yaml
                                rule: 0 < self && self <= 6553
```

Also present in 1.29.0
Not present in 1.28.5

keithmattix:
Apparently the missing quotation causes trunctation. Fixes istio/istio#59832

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -7011,7 +7011,7 @@ spec:
                               type: integer
                               x-kubernetes-validations:
                               - message: port must be between 1-65535
-                                rule: 0 < self && self <= 6553
+                                rule: 0 < self && self <= 65535
                             route:
                               description: Match a specific route.
                               properties:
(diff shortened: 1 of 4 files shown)


howardjohn (approved):
That is... fun behavior

zirain (approved):

PR #3696, opened 2026-04-21
feat: add HBONE window sizing to MeshConfig API

goosmurf:
To allow tuning of HBONE window sizing, adds the following `meshConfig` tunables:
- hbone_initial_stream_window_size
- hbone_initial_connection_window_size

Currently the default Envoy values of stream=16MB and connection=24MB are used. This results in significant buffering within Gateways/Waypoints, resulting in loss of feedback for applications and head-of-line blocking in multiplexed HBONE connections.

Companion istio PR: https://github.com/istio/istio/pull/59943

See also: https://github.com/istio/istio/issues/59887 (Excessive buffering / insufficient backpressure when clients have low bandwidth)

Note: I am very new to Istio and used Claude to help write these PRs. Please excuse any silliness, I have reviewed the code and comments and think it makes sense.

### DIFF
--- a/mesh/v1alpha1/config.proto
+++ b/mesh/v1alpha1/config.proto
@@ -68,18 +68,6 @@ message MeshConfig {
   // When unset, the field is not set on Envoy and Envoy's built-in default is used.
   google.protobuf.UInt32Value hbone_initial_connection_window_size = 72;
 
-  // HTTP/2 `initial_stream_window_size` applied to the HBONE CONNECT upstream clusters
-  // used for double-HBONE tunneling in ambient multi-network deployments (inner and outer
-  // connect-originate clusters on east-west gateways).
-  // When unset, the field is not set on Envoy and Envoy's built-in default is used.
-  google.protobuf.UInt32Value hbone_no_pooling_initial_stream_window_size = 73;
-
-  // HTTP/2 `initial_connection_window_size` applied to the HBONE CONNECT upstream clusters
-  // used for double-HBONE tunneling in ambient multi-network deployments (inner and outer
-  // connect-originate clusters on east-west gateways).
-  // When unset, the field is not set on Envoy and Envoy's built-in default is used.
-  google.protobuf.UInt32Value hbone_no_pooling_initial_connection_window_size = 74;
-
   // $hide_from_docs
   // Automatic protocol detection uses a set of heuristics to
   // determine whether the connection is using TLS or not (on the
@@ -1656,7 +1644,7 @@ message MeshConfig {
   TLSConfig tls_defaults = 64;
 
   // $hide_from_docs
-  // Next available field number: 75
+  // Next available field number: 73
   reserved 1;
   reserved "mixer_check_server";
   reserved 2;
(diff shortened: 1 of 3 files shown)


mesh/v1alpha1/config.proto
```diff
@@ -58,6 +58,28 @@ message MeshConfig {
   // For example, if your CNI has a 30s cooldown period, setting this to 15s is recommended.
   google.protobuf.Duration hbone_idle_timeout = 70;
 
+  // HTTP/2 `initial_stream_window_size` applied to the HBONE CONNECT upstream cluster
+  // used by Envoy proxies (waypoints, ingress/east-west gateways) when tunneling to ztunnel.
+  // When unset, the field is not set on Envoy and Envoy's built-in default is used.
```
goosmurf, line 63: It would be nice to suggest values for all 4 settings that would reduce excessive buffering.

Is it reasonable to assume most Istio users are running clusters connected by high bandwidth, low latency links?

Alternatively, should we say "assuming ..., these values are recommended"?

mesh/v1alpha1/config.proto
```diff
@@ -58,6 +58,28 @@ message MeshConfig {
...
+  // used for double-HBONE tunneling in ambient multi-network deployments (inner and outer
+  // connect-originate clusters on east-west gateways).
+  // When unset, the field is not set on Envoy and Envoy's built-in default is used.
+  google.protobuf.UInt32Value hbone_no_pooling_initial_stream_window_size = 73;
+
+  // HTTP/2 `initial_connection_window_size` applied to the HBONE CONNECT upstream clusters
+  // used for double-HBONE tunneling in ambient multi-network deployments (inner and outer
+  // connect-originate clusters on east-west gateways).
+  // When unset, the field is not set on Envoy and Envoy's built-in default is used.
+  google.protobuf.UInt32Value hbone_no_pooling_initial_connection_window_size = 74;
```
sschepens: does it really make sense to have separate configurations here than the pooled version?

mesh/v1alpha1/config.proto
sschepens: > It would be nice to suggest values for all 4 settings that would reduce excessive buffering.
> 
> Is it reasonable to assume most Istio users are running clusters connected by high bandwidth, low latency links?
> 
> Alternatively, should we say "assuming ..., these values are recommended"?

IMO recommending values is out of the scope of this and depends of a lot of things, we should also probably try to devise a way to avoid users using these settings altogether.
As I said in the issue, Envoy default settings are good for internal service to service communication, but are not good for external untrusted traffic, we should probably revise how ambient fits here and how we can standardize edge proxy and edge traffic configurations in Istio.

mesh/v1alpha1/config.proto
goosmurf: I think you are right that it's unnecessary, removed in https://github.com/istio/api/pull/3696/commits/4dc790d9fdbcda79ca51adf523d1757c29b8abf6

Updated the PR description to remove these also

sschepens:
@goosmurf we can probably close this since we've switched to env variables

goosmurf:
👍 
Closing this PR and its companion https://github.com/istio/istio/pull/59943 as we opted to implement the same behaviour as an env var in https://github.com/istio/istio/pull/59961

PR #3700, opened 2026-05-03
add ecdh curves to gateway server TLS settings

ramaraochavali:


ramaraochavali:

networking/v1alpha3/gateway.proto
```diff
@@ -577,4 +577,9 @@ message ServerTLSSettings {
   // * `AES256-SHA`
   // * `DES-CBC3-SHA`
   repeated string cipher_suites = 9;
+
+  // Optional: If specified, only support the specified ecdh curves.
+  // Otherwise default to the default ecdh list supported by Envoy
+  // as specified [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto).
+  repeated string ecdh_curves = 17;
```
hzxuzhonghu: Curious what will happen if the proxy does not support these curves

ramaraochavali: The config is rejected at run time - same as cipher suites

ramaraochavali:

keithmattix (approved):

keithmattix:
Seems reasonable for 1.32

kfaseela (approved):

PR #3701, opened 2026-05-03
cleariy idle timeout docs

### LINKED ISSUE #60009: DestinationRule connectionPool.tcp.idleTimeout appears to be ignored in Istio 1.27.x (sidecar TCP)
### Is this the right place to submit this?

- [x] This is not a security vulnerability or a crashing bug
- [x] This is not a question about how to use Istio

### Bug Description

In Istio 1.27.x, setting `connectionPool.tcp.idleTimeout` in a DestinationRule has no effect on actual TCP connections.

The timeout is not propagated to Envoy's `tcp_proxy` filter, so the default behavior is used instead.

This happens consistently, regardless of whether the DestinationRule is namespace-scoped or global (e.g. `*.local` in istio-system).

### Version

```Text
### Version

Istio client: 1.27.4  
Control plane: mixed (1.25.5, 1.26.8, 1.27.5)  
Data plane: mixed (1.26.8, 1.27.5)

Note: I am aware that 1.27 is EOL. This issue is reported based on this environment. I have not yet verified whether it persists in newer Istio versions.
```

### Additional Information

Example DestinationRule (not working):
```
kind: DestinationRule  
metadata:  
  name: example-tcp-idle-dr  
  namespace: example-ns  
spec:  
  exportTo:  
  - "*"
  host: example-service.example-ns.svc.cluster.local  
  trafficPolicy:  
    connectionPool:  
      tcp:  
        idleTimeout: 3700s  
    tls:  
      mode: ISTIO_MUTUAL
```

Notes:
- Changing exportTo between `"*"` and `"."` does not change behavior
- The DestinationRule is applied and visible via `kubectl get dr`
- However, no idleTimeout appears in the generated Envoy tcp_proxy configuration ( `istioctl pc listener <pod> -ojson | grep idle` )

Verification:
- With DestinationRule only:
  - No `idleTimeout` present in `istioctl proxy-config listener`
- With EnvoyFilter:
  - idleTimeout is present (e.g. `5400s`) and works as expected

Workaround using EnvoyFilter:

```
applyTo: NETWORK_FILTER  
match:  
  context: SIDECAR_OUTBOUND  # and another for SIDECAR_INBOUND
  listener:  
    filterChain:  
      filter:  
        name: envoy.filters.network.tcp_proxy  
patch:  
  operation: MERGE  
  value:  
    name: envoy.filters.network.tcp_proxy  
    typed_config:  
      "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy  
      idle_timeout: 90m  
```

This indicates that the issue is in Istio's config generation rather than Envoy itself.

I reviewed recent Istio release notes and changelogs (including versions after 1.27), but could not find any mention of a fix related to DestinationRule `connectionPool.tcp.idleTimeout` being propagated to Envoy `tcp_proxy.idle_timeout`.

Based on this, it is unclear whether this behavior has been addressed in newer versions.

ramaraochavali:
For https://github.com/istio/istio/issues/60009#issuecomment-4364048012

Fixes https://github.com/istio/istio/issues/60009

### DIFF
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -630,14 +630,19 @@ message ConnectionPoolSettings {
     // is reached the connection will be closed. Duration must be at least 1ms.
     google.protobuf.Duration max_connection_duration = 4;
 
-    // The idle timeout for TCP connections.
-    // The idle timeout is defined as the period in which there are no bytes sent or received on either
-    // the upstream or downstream connection.
+    // The idle timeout for TCP connections. This is applied to the outbound TCP proxy
+    // filter on the listener for the upstream cluster.
+    // The idle timeout is defined as the period in which there are no bytes sent or received on the
+    // upstream connection.
     // If not set, the default idle timeout is 1 hour. If set to 0s, the timeout will be disabled.
     // Idle timeout is not configured per each cluster individually when weighted destinations are used,
     // because idleTimeout is a property of a listener, not a cluster. In that case, idleTimeout
     // specified in a destination rule for the first weighted route is configured in the listener,
     // which means also for all weighted routes.
+    // To set the idle timeout for downstream (inbound) TCP connections, use the
+    // `idleTimeout` field in the proxy configuration (e.g., via the `proxy.istio.io/config`
+    // annotation) or use an EnvoyFilter to patch the inbound listener's
+    // `envoy.filters.network.tcp_proxy` network filter.
     // +protoc-gen-crd:duration-validation:none
     google.protobuf.Duration idle_timeout = 5;
   }
--- a/networking/v1alpha3/destination_rule.proto
+++ b/networking/v1alpha3/destination_rule.proto
@@ -630,8 +630,7 @@ message ConnectionPoolSettings {
     // is reached the connection will be closed. Duration must be at least 1ms.
     google.protobuf.Duration max_connection_duration = 4;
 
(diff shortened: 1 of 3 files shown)


networking/v1alpha3/destination_rule.proto
```diff
@@ -630,14 +630,17 @@ message ConnectionPoolSettings {
...
+    // The idle timeout for TCP connections. This is applied to the outbound connections to the upstream service.
+    // The idle timeout is defined as the period in which there are no bytes sent or received on the
+    // upstream connection.
     // If not set, the default idle timeout is 1 hour. If set to 0s, the timeout will be disabled.
     // Idle timeout is not configured per each cluster individually when weighted destinations are used,
     // because idleTimeout is a property of a listener, not a cluster. In that case, idleTimeout
     // specified in a destination rule for the first weighted route is configured in the listener,
     // which means also for all weighted routes.
+    // To set the idle timeout for downstream (inbound) connections, use the
+    // `idleTimeout` field in the proxy configuration (e.g., via the `proxy.istio.io/config`
```
hzxuzhonghu: is this correct? I thinbk it should be from `ISTIO_META_IDLE_TIMEOUT`

ramaraochavali: it actually has to be under metadata. I updated

keithmattix (approved):

ramaraochavali:
@istio/wg-networking-maintainers can you PTAL?

sridhargaddam (approved):

PR #3706, opened 2026-05-12
Add HTTP/2 ping options to DestinationRule

mpuncel:
## Summary
- Adds `connectionPool.http.http2ProtocolOptions.connectionKeepalive.interval/timeout` to `DestinationRule`.
- Regenerates protobuf, JSON/deepcopy/alias, docs, and CRD outputs.
- Adds a CRD validation fixture covering the new YAML shape.

## Related
- Implements the API surface proposed in istio/istio#55640.
- Consumed by the companion Istio control-plane draft: https://github.com/mpuncel/istio/pull/1

## Tests
- `buf lint`
- `go test ./...` from `tests/`
- `git diff --check`


networking/v1alpha3/destination_rule.proto
```diff
@@ -703,6 +695,26 @@ message ConnectionPoolSettings {
     // The maximum number of concurrent streams allowed for a peer on one HTTP/2 connection.
     // Defaults to 2^31-1.
     int32 max_concurrent_streams = 8;
+
+    // Settings for HTTP/2 protocol options.
+    message Http2ProtocolOptions {
```
ramaraochavali: instead of introducing a new Http2ProtocolOptions now (with concurrent streams and other exposed directly), can we add a new field directly like http2_keep_alive of type ConnectionKeepalive at the top level?

ramaraochavali (commented):
One question on the proto.

Can you explain your use case more - where you want to use this?

mpuncel:
@ramaraochavali the use case is that sometimes in our service mesh we see "stuck" TCP connections, where the request appears to be written to the connection successfully but no response ever comes back, all requests time out for a period of time until the connection is torn down. Envoy HTTP/2 ping can detect this faster and help recover. We have seen this resolve the issue with HTTP/2 ping directly from gRPC client libraries, but want to make it possible to enable via Istio.

networking/v1alpha3/destination_rule.proto
mpuncel: sounds good, making that change

mpuncel: addressed

networking/v1alpha3/destination_rule.proto
```diff
@@ -703,6 +695,20 @@ message ConnectionPoolSettings {
     // The maximum number of concurrent streams allowed for a peer on one HTTP/2 connection.
     // Defaults to 2^31-1.
     int32 max_concurrent_streams = 8;
+
+    // Settings for HTTP/2 PING frames.
+    message ConnectionKeepalive {
```
ramaraochavali, line 700: are you planning to default the jitter to envoy's default?

mpuncel: yeah, should I make that configurable here?

ramaraochavali: No. Just checking. Envoy default is fine.

ramaraochavali (commented):
Looks good to me
@istio/technical-oversight-committee any objections to this?

mpuncel:
@ramaraochavali have there been any conversations about this?

ramaraochavali (approved):

ramaraochavali:
Sorry. I missed this. @istio/wg-networking-maintainers can you PTAL?

hzxuzhonghu (approved):

mpuncel:
I (hopefully) addressed the build failures

PR #3627, opened 2026-01-09
edge proxy config API

skriss:
This is a draft API proposal for feedback/discussion, related to the [Edge Proxy Config RFC](https://docs.google.com/document/d/1Wi9pxnL2CZllk_cEOxd_1NQB_HOsZwxcVcGbMjn_l-c/edit?tab=t.0#heading=h.xw1gqgyqs5b).

One minor change from the RFC is replacing the `enableEdgeProxyConfig` boolean field with a `profile` field, per suggestion in the WG meeting. This is more extensible and consistent with other parts of the Istio config API. 

cc @keithmattix @dgn 

ref. https://github.com/istio/istio/issues/57973

PR #3707, opened 2026-05-15
edge proxy config API

cam-garrison:
Picks up and rebases #3627 (by @skriss) (addt'l adding release note).

Adds edge proxy configuration fields to `ProxyConfig` — a `profile` enum (`SIDECAR`/`EDGE`), connection buffer limits, HTTP timeouts, HTTP/2 tuning, header/path normalization, and connection limits. Enables native Istio support for [Envoy
edge proxy best practices](https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge.html).

Design per WG feedback: uses `profile` enum instead of boolean. Overload manager out of scope for initial implementation.

Related RFC: [Edge Proxy Config](https://docs.google.com/document/d/1Wi9pxnL2CZllk_cEOxd_1NQB_HOsZwxcVcGbMjn_l-c/edit?tab=t.0#heading=h.xw1gqgyqs5b)

Supersedes #3627.
cc @skriss @keithmattix @dgn @frittentheke

ref. https://github.com/istio/istio/issues/57973
ref. https://github.com/istio/istio/issues/24715

### DIFF
--- a/mesh/v1alpha1/proxy.proto
+++ b/mesh/v1alpha1/proxy.proto
@@ -773,6 +773,134 @@ message ProxyConfig {
   // Defaults to true.
   // Optional.
   google.protobuf.BoolValue stats_compression = 42;
+
+  // ProxyConfigProfile defines the configuration profile for the proxy.
+  // Different profiles optimize the proxy's behavior for specific deployment patterns.
+  // The profile determines which configuration settings are applied by default.
+  enum ProxyConfigProfile {
+    // SIDECAR profile is optimized for sidecar deployments.
+    // This is the default profile and is suitable for proxies running alongside application containers.
+    // Sidecar proxies typically handle lower connection volumes and shorter-lived connections.
+    SIDECAR = 0;
+
+    // EDGE profile is optimized for edge gateway deployments.
+    // This profile is suitable for proxies that serve as ingress or egress gateways.
+    // Edge proxies typically handle higher connection volumes, longer-lived connections,
+    // and require more robust buffer and timeout configurations.
+    EDGE = 1;
+  }
+
+  // The config profile to use for this proxy.
+  ProxyConfigProfile profile = 43;
+
+  // Soft limit on size of the listener's new connection read and write buffers in bytes.
+  // See Envoy's [per_connection_buffer_limit_bytes](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#envoy-v3-api-field-config-listener-v3-listener-per-connection-buffer-limit-bytes).
+  int32 listener_per_connection_buffer_limit_bytes = 44;
+
+  // Soft limit on size of the cluster's new connection read and write buffers in bytes.
+  // See Envoy's [per_connection_buffer_limit_bytes](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-per-connection-buffer-limit-bytes).
+  int32 cluster_per_connection_buffer_limit_bytes = 45;
+
+  // The idle timeout for HTTP connections. The idle timeout is defined as the period in which there are no active requests.
+  // When the idle timeout is reached, the connection will be closed.
+  // Note that request-based timeouts mean that HTTP/2 PINGs will not keep the connection alive.
+  // See Envoy's [idle_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-httpprotocoloptions-idle-timeout).
+  google.protobuf.Duration http_idle_timeout = 46;
+
+  // The maximum duration of a connection.
+  // When this timeout is reached, the connection will be closed.
+  // See Envoy's [max_connection_duration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-httpprotocoloptions-max-connection-duration).
+  google.protobuf.Duration http_max_connection_duration = 47;
+
+  // The time that Envoy will wait between sending an HTTP/2 shutdown notification (GOAWAY frame with max stream ID)
+  // and a final GOAWAY frame. This is used so that Envoy can drain in-flight requests.
+  // See Envoy's [drain_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-httpprotocoloptions-drain-timeout).
+  google.protobuf.Duration http_drain_timeout = 48;
+
+  // The amount of time that Envoy will wait for the entire request to be received.
+  // The timer is activated when the request is initiated, and is reset each time new data arrives.
+  // See Envoy's [request_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-request-timeout).
+  google.protobuf.Duration http_request_timeout = 49;
+
+  // The amount of time Envoy will wait for the request headers to be received.
+  // The timer is activated when the first byte of the headers is received and is disarmed when the last byte of the headers has been received.
+  // See Envoy's [request_headers_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-request-headers-timeout).
+  google.protobuf.Duration http_request_headers_timeout = 50;
+
+  // The amount of time that Envoy will allow a stream to exist with no upstream or downstream activity.
+  // The timer is activated when the downstream connection sends the request and is reset on any frame from the upstream or downstream for the stream.
+  // See Envoy's [stream_idle_timeout](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-field-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-stream-idle-timeout).
+  google.protobuf.Duration http_stream_idle_timeout = 51;
+
+  // The maximum duration of a stream.
+  // When this timeout is reached, the stream will be closed.
+  // See Envoy's [max_stream_duration](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-httpprotocoloptions-max-stream-duration).
+  google.protobuf.Duration http_max_stream_duration = 52;
+
+  // Maximum number of concurrent streams allowed for HTTP/2 and HTTP/3 connections.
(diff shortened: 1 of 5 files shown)


mesh/v1alpha1/proxy.proto
```diff
@@ -773,6 +773,134 @@ message ProxyConfig {
...
+
+    // EDGE profile is optimized for edge gateway deployments.
+    // This profile is suitable for proxies that serve as ingress or egress gateways.
+    // Edge proxies typically handle higher connection volumes, longer-lived connections,
+    // and require more robust buffer and timeout configurations.
+    EDGE = 1;
+  }
+
+  // The config profile to use for this proxy.
+  ProxyConfigProfile profile = 43;
```
ramaraochavali: How will this  and below fields work together? When I select Edge profile, do I get some default values for the below configurations? I think it would be good to tie this profile and the below fields together or explain clearly how they work together?

cam-garrison: Good question, thx - updated the proto comments to help clarify https://github.com/istio/api/pull/3707/commits/6c3c4d84e7e50acfabd197f9ee9a4bf3d1f405b9.

The profile sets defaults for the fields below, explicitly setting any field overrides the profile default. SIDECAR (the zero value) preserves existing behavior. EDGE applies Envoy's recommended edge defaults. The actual default values will be applied in the istio/istio implementation.

mesh/v1alpha1/proxy.proto
```diff
@@ -773,6 +773,135 @@ message ProxyConfig {
   // Defaults to true.
   // Optional.
   google.protobuf.BoolValue stats_compression = 42;
+
+  // ProxyConfigProfile defines the configuration profile for the proxy.
+  // The profile determines default values for the fields below (buffer limits,
+  // timeouts, HTTP/2 tuning, header/path normalization, and connection limits).
+  // Explicitly setting any field always takes precedence over profile defaults.
+  enum ProxyConfigProfile {
```
ramaraochavali: it is hard to figure out which fields impacted by this profile. I would suggest adding a new Message like ProxyDefaults or some thing like that , that has all the fields that the profile controls and then select the profile ?

cam-garrison: That's fair, I understand. I think it is better to not have them be nested like `proxyConfig.proxyDefaults.httpIdleTimeout` and instead be flat on proxyConfig like `proxyConfig.httpIdleTimeout`. These fields are individually useful without a profile and the fields are not one conceptual unit, they cover different envoy scopes. 

I agree it should be more clear which fields the profile effects. I can add a comment to each field noting that it's subject to profile defaults, would that address your concern?

ramaraochavali: Stepping back, why do we need an Enum here? Can we just use a bool like `enable_edge_defaults` (default to false) and when enabled Edge defaults are enabled? If we do that, can we group all these configurations under some new message like `GatewayProxySettings` or `EdgeProxySettings` ?

cam-garrison: The original proposal from @skriss used a bool, and it was changed to an enum per the working group's review. "One minor change from the RFC is replacing the enableEdgeProxyConfig boolean field with a profile field, per suggestion in the WG meeting. This is more extensible and consistent with other parts of the Istio config API." from #3627. 

Re: grouping, I still feel like flat fields are best. These are general-purpose proxy settings that are useful independently of the profile. A user should be able to set http_idle_timeout on a sidecar without any
edge/gateway connotation. Wrapping them in `EdgeProxySettings` would imply they're edge-specific when they're not, the profile just provides convenient defaults.

ramaraochavali: > http_idle_timeout on a sidecar without any edge/gateway connotation

The problem of applying this field to sidecar is - it is not clear to which side of hop/service we are setting this. It interfers with existing mechanisms like DR and ProxyConfig

cam-garrison: Yeah, that's fair.  Tthere is overlap with DestinationRule at the cluster level. A few thoughts:

- The EDGE profile defaults will only be applied to gateway (Router-type) proxies in the implementation. Sidecars aren't affected unless a user explicitly sets individual fields.
 
- For explicit overrides, these are proxy-wide settings at the Envoy listener/HCM/cluster level, the same scope as existing ProxyConfig fields like concurrency. The sidecar inbound/outbound ambiguity exists there too.
 
- The interaction with a DR is at different Envoy layers (HCM vs cluster), so they don't directly conflict, but I agree it's worth documenting.

I don't think that grouping these under a separate message wouldn't resolve the sidecar scope question the ambiguity comes from ProxyConfig being proxy-wide, not from where the fields sit in the proto. Would some more documentation on scope and the DR interaction address what you're getting at here?

ramaraochavali: >  the same scope as existing ProxyConfig fields like concurrency. The sidecar inbound/outbound ambiguity exists there too.

The concurrency is not at listener/HCM/cluster level. It applies to entire proxy so it makes sense to apply at proxy level. There is no confusion there.

> The interaction with a DR is at different Envoy layers (HCM vs cluster), so they don't directly conflict, but I agree it's worth documenting.

What does specifying `http_idle_timeout` mean on a sidecar? Does it apply to the inbound listener, outbound clusters, or both?  For outbound, which upstream services are affected? This conflicts with ConnectionPoolSettings.tcp.idle_timeout in DestinationRule, which already provides per-service outbound idle timeout control.  If it applies to all outbound and all inbound, we need to clearly document that and also specify what will have preference ProxyConfig or DR? Is n't that additional confusion? 

> I don't think that grouping these under a separate message wouldn't resolve the sidecar scope question the ambiguity comes from ProxyConfig being proxy-wide, not from where the fields sit in the proto. 

What is the motivation to expose only these fields when Envoy provides 100 other knobs?  I believe the reason we selected these fields is because Envoy documentation says it is best practice to set defaults to these fields.    There are now 18 new fields in ProxyConfig with only a comment tying them to the profile. A future reader (or contributor adding field 61) has no structural indication of which fields the profile governs - I have to read the entire proxy config proto to understand what the profile configures.

Yes. They can be still be configured individually outside profile if we want to allow that.

I'm supportive of the feature — edge hardening is valuable. My concern is purely about API structure, clarity of scope (the issues and the PR description says "edge proxy config API") and future maintainability.

cam-garrison: Thanks for clarifying with such a detailed review, I appreciate it (and thanks for correcting my misunderstanding of concurrency). I took some time to look things over with the long weekend and I think your perspective on scope and structure for this is right. I made some changes to the structure and added much more documentation as well, changes are as follows:

Nested fields into ConnectionSettings: The 18 fields and the profile enum now live in a new ConnectionSettings message on ProxyConfig (proxyConfig.connectionSettings.*). This bind the profile's scope to a single message rather than relying on per-field comments.   

Scope documented at the message level: Added a doc comment clarifying that these primarily configure the downstream side (listeners + HCM), naming cluster_per_connection_buffer_limit_bytes as the cluster level exception, and explaining the DR interaction, overlapping fields like connectionPoolSettings.tcp.idleTimeout apply independently at different hops rather than one overriding the other.

Section dividers within the message group fields by category for readability.  

Connection-limit comments now reference the actual Envoy runtime config keys      (envoy.resource_limits.listener.<name>.connection_limit, overload.global_downstream_max_connections) instead of the original API links that pointed at adjacent fields.

cc @skriss @keithmattix @dgn, this is a structural change beyond the original API PR (#3627); the field set and profile enum are unchanged, the nesting is new.

ramaraochavali (commented):
Thank you for iterating on this. The direction looks good to me now. I would let other @istio/technical-oversight-committee members review the approach. I will take a detailed look.

mesh/v1alpha1/proxy.proto
```diff
@@ -773,6 +773,172 @@ message ProxyConfig {
...
+  // (notably `connectionPoolSettings.tcp.idleTimeout`), both apply
+  // independently at different hops rather than one overriding the other:
+  // DestinationRule governs Envoy → upstream connections, while these
+  // settings govern downstream → Envoy connections. For per-destination
+  // connection pool configuration, use DestinationRule's
+  // `connectionPoolSettings`.
+  message ConnectionSettings {
+    // ProxyConfigProfile selects a default value set for the fields in this message.
+    // Explicitly setting any field always takes precedence over profile defaults.
+    enum ProxyConfigProfile {
```
ramaraochavali: nit: We can just call Profile as the outer layer gives the scope. "ProxyConfig" IMO is little too broad

mesh/v1alpha1/proxy.proto
```diff
@@ -773,6 +773,172 @@ message ProxyConfig {
...
+      // SIDECAR profile preserves existing Istio behavior.
+      // This is the default profile. No additional defaults are applied.
+      SIDECAR = 0;
+
+      // EDGE profile applies Envoy's recommended defaults for edge gateway deployments.
+      // See https://www.envoyproxy.io/docs/envoy/latest/configuration/best_practices/edge
+      // Explicitly setting any field overrides the corresponding profile default.
+      EDGE = 1;
```
ramaraochavali, line 819: Please document the defaults of each of these fields when this is chosen.

mesh/v1alpha1/proxy.proto
```diff
@@ -773,6 +773,172 @@ message ProxyConfig {
...
+
+    // Initial stream-level flow-control window size for HTTP/2 connections.
+    // Valid values range from 65535 (2^16 - 1, HTTP/2 default) to 2147483647 (2^31 - 1, HTTP/2 maximum).
+    // See Envoy's [initial_stream_window_size](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-initial-stream-window-size).
+    int32 http2_initial_stream_window_size = 12;
+
+    // Initial connection-level flow-control window size for HTTP/2 connections.
+    // Valid values range from 65535 (2^16 - 1, HTTP/2 default) to 2147483647 (2^31 - 1, HTTP/2 maximum).
+    // See Envoy's [initial_connection_window_size](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-http2protocoloptions-initial-connection-window-size).
+    int32 http2_initial_connection_window_size = 13;
```
ramaraochavali: Should we use wrappers here instead of int32 to distinguish between set and unset? Same comment for others as well

mesh/v1alpha1/proxy.proto
```diff
@@ -773,6 +773,172 @@ message ProxyConfig {
...
+
+    // --- Header and path normalization ---
```
ramaraochavali: nit: Generally we do not segregate sections like this in Istio

ramaraochavali (commented):
some comments

mesh/v1alpha1/proxy.proto
cam-garrison: Thx, removed these dividers in 14ad5b3

mesh/v1alpha1/proxy.proto
cam-garrison: good point, took a look and converted to `google.protobuf.Int32Value` for seven of the fields so we can distinguish from 0 and unset, addressed in 14ad5b3

mesh/v1alpha1/proxy.proto
cam-garrison: Listed the concrete envoy default fields/values from the best practices document, 14ad5b3

mesh/v1alpha1/proxy.proto
cam-garrison: agreed, renamed 14ad5b3

cam-garrison:
Hey @keithmattix, could you please take a look at this when you have a chance? I know had input on the original RFC and I'd appreciate any thoughts you have on the state of this PR. TYSM!

keithmattix (approved):
LGTM - I'd like @ramaraochavali to be the final say here though

ramaraochavali (approved):
Thank you for the all work here.

PR #3708, opened 2026-05-16
networking: add prefix_rewrite to HTTPRedirect

liamawhite:
## Summary

`HTTPRedirect.uri` performs a full path replacement and has no prefix-aware variant. Envoy's `RedirectAction` has supported `prefix_rewrite` for years, but Istio has never exposed it, making it impossible to perform redirects that preserve the path suffix (e.g. `example.com/foo/bar` → `foo.example.com/bar`).

This was requested in istio/istio#47500, istio/istio#47777, and istio/istio#52521, all of which were closed as stale.

This PR adds a `prefix_rewrite` field to `HTTPRedirect`, mutually exclusive with the existing `uri` field. The matched route prefix is stripped and replaced with the new value. The legacy `uri` field is unchanged.

Regex rewrite support for redirects will be proposed separately via the Gateway API GEP process.

## Examples

Strip a prefix and change host (`example.com/foo/bar` → `foo.example.com/bar`):
```yaml
- match:
  - uri:
      prefix: /foo/
  redirect:
    authority: foo.example.com
    prefix_rewrite: /
    redirectCode: 301
```

Replace a prefix (`example.com/foo/bar` → `example.com/baz/bar`):
```yaml
- match:
  - uri:
      prefix: /foo
  redirect:
    prefix_rewrite: /baz
    redirectCode: 301
```

Full path replacement (existing behaviour, unchanged):
```yaml
- match:
  - uri:
      prefix: /foo
  redirect:
    uri: /bar
    redirectCode: 301
```

## Related

- istio/istio#47500
- istio/istio#47777
- istio/istio#52521

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -11523,19 +11523,31 @@ spec:
                       description: The name assigned to the route for debugging purposes.
                       type: string
                     redirect:
+                      allOf:
+                      - oneOf:
+                        - not:
+                            anyOf:
+                            - required:
+                              - port
+                            - required:
+                              - derivePort
+                        - required:
+                          - port
+                        - required:
+                          - derivePort
+                      - oneOf:
+                        - not:
+                            anyOf:
+                            - required:
+                              - prefixRewrite
+                            - required:
+                              - uriRegexRewrite
+                        - required:
+                          - prefixRewrite
+                        - required:
+                          - uriRegexRewrite
                       description: A HTTP rule can either return a direct_response,
                         redirect or forward (default) traffic.
-                      oneOf:
-                      - not:
-                          anyOf:
-                          - required:
-                            - port
-                          - required:
-                            - derivePort
-                      - required:
-                        - port
-                      - required:
-                        - derivePort
                       properties:
                         authority:
                           description: On a redirect, overwrite the Authority/Host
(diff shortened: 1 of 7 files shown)


keithmattix:
I'm not sure why we shouldn't do this under the banner of gateway API (/cc @Stevenjin8). The fields will likely exist on VS since that's how we translate, but ideally Gateway API is the scope

Stevenjin8:
Fromt the title though, this is for http redirect, not path rewrites, so they are slightly different things.

liamawhite:
IIRC prefix rewrite for redirect exists in the gateway API but Istio implements it using a hack to get round the fact its not in the virtual service API. 

i.e. Some metadata gets added during translation that allows the gateway API to pass this through to the thing generating the XDS configuration without it existing in the VirtualService API.

Stevenjin8:
@liamawhite you're right; https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httppathmodifier

liamawhite:
So I guess my question is can I get this approved with just a prefix rewrite whilst the regex part goes through the gateway api process? Or are we ok adding support for both in the VS whilst we go through the gateway api process?

keithmattix:
I don't have any opposition towards moving prefix rewrite to first-class on VS; it'll remove the hack from gateway api and set us up for a more consistent API in the meantime. I think the regex piece is more interesting; I really don't want to go down a path before we have a solid GEP upstream that tells us what the constraints will be

keithmattix (commented):
LGTM - let's discuss in Slack/the upcoming working group meeting.

/cc @istio/technical-oversight-committee

keithmattix (approved):

zirain (approved):

PR #3710, opened 2026-05-19
Add the support on clock_skew_seconds in request authentication

freedomljc:
It's to mimic the support of `clock_skew_seconds` on jwt_authn https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto

### DIFF
--- a/kubernetes/customresourcedefinitions.gen.yaml
+++ b/kubernetes/customresourcedefinitions.gen.yaml
@@ -16709,6 +16709,12 @@ spec:
                         minLength: 1
                         type: string
                       type: array
+                    clockSkewSeconds:
+                      description: The clock skew, in seconds, allowed when verifying
+                        the JWT `exp` and `nbf` claims.
+                      maximum: 4294967295
+                      minimum: 0
+                      type: integer
                     forwardOriginalToken:
                       description: If set to true, the original token will be kept
                         for the upstream request.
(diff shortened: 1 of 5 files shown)


howardjohn (commented):
In what scenario do you need more than a 60s clock skew?

freedomljc:
> In what scenario do you need more than a 60s clock skew?

I guess there're two scenarios we might want to adjust clock_skew_seconds:
- Short-lived JWT token: There's an example mentioned in https://github.com/envoyproxy/envoy/pull/13872#issuecomment-721435296: By having short-lived JWT token that expires in ~5 minutes (or even shorter), 1 minute skew means at least 20% difference.
- Long-lived JWT token with more flexible handling on expired token: Instead of letting requestAuthentication directly reject the expired token, there could be a grey area regarding whether to handle the expired token: we can have subsequent envoy filter or domain app to decide based on business logic.

howardjohn:
The short lived case is a pure hypothetical from 6 years ago not a real use case. It sounds more like what you want is not "clock skew" but to disable expiry validation?

freedomljc:
> The short lived case is a pure hypothetical from 6 years ago not a real use case. It sounds more like what you want is not "clock skew" but to disable expiry validation?

I'd like to disable the expiry validation within requestAuthentication and let subsequent components determine how to handle expired tokens instead. 
Also, rather than hard-coding a 60-second clock skew, it would make more sense to make that value configurable.

freedomljc:
@howardjohn , would you mind to follow-up and share your thoughts? 

Having envoyfilter to override jwt_authn configuration is quite inconvenient, due to the  `MERGE` operation in istio EnvoyFilter at this level requires the entire provider proto message, so all fields that istiod would normally generate must be re-declared. Additionally, to avoid configuring inlined jwks key material in envoyfilter, we'd need to have Envoyfilter with [remote_jwks](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto#envoy-v3-api-msg-extensions-filters-http-jwt-authn-v3-remotejwks) specified, which requires to create istio destination_rule/service_entry for generating envoy cluster.

PR #3722, opened 2026-06-09
feat(tracing): add otel always_on sampler

### LINKED ISSUE #60500: Support more samplers in the OpenTelemetry ExtensionProvider config

#50001 has added support for samplers with the OpenTelemetry tracing provider, but as far as I understand only the Dynatrace sampler was implemented. However, the [Envoy docs list multiple other samplers which are simple but useful](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/trace/v3/opentelemetry.proto.html#envoy-v3-api-field-config-trace-v3-opentelemetryconfig-sampler).
In my case, I would like to just use an `always_on` sampler to ignore any trace flags on incoming requests.

I managed to create this EnvoyFilter to apply the setting, but to do this I have to repeat config that was generated from the provider settings (otherwise grpc_service, resource_detectors, service_name are simply removed by the merge) which is less than ideal.
<details>
<summary>Example Filter YAML</summary>

```yaml
kind: EnvoyFilter
metadata:
  name: gateway-public-tracing-always-on
  namespace: istio-system
spec:
  workloadSelector:
    labels:
      gateway.networking.k8s.io/gateway-name: public
  configPatches:
    - applyTo: NETWORK_FILTER
      match:
        context: GATEWAY
        listener:
          filterChain:
            filter:
              name: envoy.filters.network.http_connection_manager
      patch:
        operation: MERGE
        value:
          typed_config:
            "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
            tracing:
              provider:
                typed_config:
                  "@type": type.googleapis.com/envoy.config.trace.v3.OpenTelemetryConfig
                  grpc_service:
                    envoy_grpc:
                      authority: opentelemetry-collector.infrastructure.svc.cluster.local
                      cluster_name: outbound|4317||opentelemetry-collector.infrastructure.svc.cluster.local
                  resource_detectors:
                    - name: envoy.tracers.opentelemetry.resource_detectors.environment
                      typed_config:
                        '@type': type.googleapis.com/envoy.extensions.tracers.opentelemetry.resource_detectors.v3.EnvironmentResourceDetectorConfig
                  service_name: istio-public.istio-system
                  sampler:
                    name: envoy.tracers.opentelemetry.samplers.always_on
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.tracers.opentelemetry.samplers.v3.AlwaysOnSamplerConfig
```
</details>

[ ] Ambient
[ ] Docs
[ ] Dual Stack
[ ] Installation
[ ] Networking
[x] Extensions and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Affected features (please put an X in all that apply)**

[ ] Multi Cluster
[ ] Virtual Machine
[ ] Multi Control Plane

**Additional context**

davidspek:
Fixes: https://github.com/istio/istio/issues/60500

This PR adds configuration support for the OpenTelemetry `always_on` trace sampler. This sampler has been part of the Istio [proxy build](https://github.com/istio/proxy/blob/master/bazel/extension_config/extensions_build_config.bzl#L264) for 2 years already, but has never been exposed in the configuration.

### DIFF
--- a/mesh/v1alpha1/config.pb.go
+++ b/mesh/v1alpha1/config.pb.go
@@ -4389,6 +4389,7 @@ type MeshConfig_ExtensionProvider_OpenTelemetryTracingProvider struct {
 	// Types that are valid to be assigned to Sampling:
 	//
 	//	*MeshConfig_ExtensionProvider_OpenTelemetryTracingProvider_DynatraceSampler_
+	//	*MeshConfig_ExtensionProvider_OpenTelemetryTracingProvider_AlwaysOnSampler_
 	Sampling      isMeshConfig_ExtensionProvider_OpenTelemetryTracingProvider_Sampling `protobuf_oneof:"sampling"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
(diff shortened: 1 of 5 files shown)


zirain:
what's the different with 100% sampling rate?

davidspek:
@zirain The difference is that the always_on sampler will change the propagated trace header to indicate that it's sampled. We have issues with an external party passing trace headers that aren't sampled on their side, causing our internal spans to also not be sampled resulting in loss of observability data.

zirain (approved):

davidspek:
@zirain Just rebased the PR so should be good now.

PR #3724, opened 2026-06-10
EF: add new MERGE_AND_REPLACE_LIST to the EnvoyFilter patch operation

wbpcode:
- Added new `MERGE_AND_REPLACE_LIST` option make it's possible to replace existing list rather than to append new list to existing list. This make it's pretty easy to patch `repeated` proto field, like ALPN list and so on.


wbpcode:
cc @ramaraochavali

ramaraochavali (approved):
LGTM

zirain (approved):

PR #3725, opened 2026-06-11
fix: add UNSPECIFIED zero values to ConnectionSettings enums

cam-garrison:
Found a bug in recent PR I made #3707 while writing wiring code in istio/istio. HttpHeadersWithUnderscoresAction and HttpPathWithEscapedSlashesAction use bare enums whose zero values (ALLOW, KEEP_UNCHANGED) are indistinguishable from "unset". 

This adds Add `HEADERS_WITH_UNDERSCORES_UNSPECIFIED = 0` and `PATH_WITH_ESCAPED_SLASHES_UNSPECIFIED = 0` to the two `ConnectionSettings` enums introduced in above PR. 

Wire-breaking change to enum numbering, should be safe since ConnectionSettings has not shipped in any release.

### DIFF
--- a/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
+++ b/mesh/v1alpha1/istio.mesh.v1alpha1.pb.html
@@ -4875,6 +4875,14 @@ http_path_with_escaped_slashes_action:      UNESCAPE_AND_REDIRECT</p>
 </tr>
 </thead>
 <tbody>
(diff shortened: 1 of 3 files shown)


zirain:
is this released or not? if so, this seems to be a breaking change.

keithmattix:
Not released yet; just merged last week. LGTM

keithmattix (approved):

zirain (approved):

